PR #30, opened 2016-02-23
Update Client API docs - POST scrubbing not req'd

nateberkopec:
Was told by Armin this is the case


dcramer:
Let's keep the requirement. Even if we dont yet do it in SDKs, we should have strong guidelines. As long as its clearly documented how to enable sending POST data that's good enough. Secure by-default, and POST data is sensitive/complex.

PR #34, opened 2016-04-20
Document breadcrumbs interface

benvinegar:
Note I intentionally left off the "query" and "rpc" breadcrumbs because we haven't represented them in the UI yet.

I also left off the classifier stuff because I don't think we've fleshed that out either.

cc @mitsuhiko @dcramer @nateberkopec

### DIFF
--- a/docs/clientdev/interfaces.rst
+++ b/docs/clientdev/interfaces.rst
@@ -329,3 +329,175 @@ data related to the current user, the current HTTP request.
           "query": "SELECT 1",
           "engine": "psycopg2"
         }
+
+
+Breadcrumbs Interface
+---------------------
+
+**NOTE:** *Breadcrumbs are an experimental Sentry feature and may not yet be available.*
+
+The breadcrumbs interface specifies a series of application events, or "breadcrumbs",
+that occurred before the main event.
+
+.. describe:: sentry.interfaces.Breadcrumbs
+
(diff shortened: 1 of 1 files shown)


benvinegar:
Example of how the individual breadcrumb type descriptions are rendered inside docs:

image

docs/clientdev/interfaces.rst
```diff
@@ -329,3 +329,175 @@ data related to the current user, the current HTTP request.
...
+        }, {
+          "timestamp": 1461185753847,
+          "type": "navigation",
+          "data": {
+            "from": "/login",
+            "to": "/dashboard"
+          }
+        }]
+
+Below are descriptions of individual breadcrumb types, and what their ``data`` properties look like.
```
dcramer: I would add a Types header

docs/clientdev/interfaces.rst
```diff
@@ -329,3 +329,175 @@ data related to the current user, the current HTTP request.
...
+
+.. describe:: "message" breadcrumb
```
dcramer: just call it "message" here instead of saying its a '"message" breadcrumb'

benvinegar:
Updated.

@dcramer – did you want me to push this and deploy docs, or is it sufficient to just point people at the PR for now?

dcramer:
lgtm

PR #36, opened 2016-04-20
Add User Feedback to Learn

dcramer:
@getsentry/team

### DIFF
--- a/docs/learn/user-feedback.rst
+++ b/docs/learn/user-feedback.rst
@@ -1,0 +1,42 @@
+User Feedback
+=============
+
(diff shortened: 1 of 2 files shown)


docs/learn/user-feedback.rst
```diff
@@ -0,0 +1,42 @@
+User Feedback
+=============
+
+Sentry provides the ability to collect additional feedback form the user upon hitting
```
mattrobenolt: > form the user

dcramer:
Anyone know if we have prior usage of images in docs? Would be good to put something here.

mattrobenolt:
Seems ok to me.

PR #37, opened 2016-04-26
Publicize internal code guidelines.

JTCunning:
Shamelessly copied from internal docs and turned into RST.

@getsentry/infrastructure 

This is for the second bulletpoint of https://github.com/getsentry/sentry/issues/2899.

### DIFF
--- a/docs/serverdev/code.rst
+++ b/docs/serverdev/code.rst
@@ -1,0 +1,37 @@
+Code Review
+===========
+Code review is mandatory at Sentry. This adds overhead to each change, but ensures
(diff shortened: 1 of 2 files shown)


mattrobenolt:
:cake:

Are we going to do a condensed version of this for `CONTRIBUTING.md` or will that be entirely different?

JTCunning:
@mattrobenolt: I was thinking of a stock "Welcome to getsentry/something here's how we do things *link to docs" on every project but sentry, and then an actual TLDR+link on that one.

mattrobenolt:
Sounds good. Bring it.

PR #54, opened 2016-08-23
Update event rollup removal instructions

nate-opti:
Update event deletion instructions to match current UI on the hosted version, which has this "Delete" button rather than a "Remove Event Data" link:
event_rollup_delete_button

Unless I can't see the remove event data option (I'm an admin but not an owner)... in which case my bad :)

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-docs/54)


mattrobenolt:
thx 🍰 🍦

PR #67, opened 2016-09-26
Document time_spent core event attribute

nateberkopec:
This change is [Reviewable](https://reviewable.io/reviews/getsentry/sentry-docs/67)

### DIFF
--- a/docs/clientdev/attributes.rst
+++ b/docs/clientdev/attributes.rst
@@ -269,3 +269,13 @@ highly encouraged:
 
 
 For information about overriding grouping see :ref:`custom-grouping`.
+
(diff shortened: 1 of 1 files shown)


dcramer:
@nateberkopec I think we decided to just kill this for now until we actually have features around perf. Originally it was intended to be the "duration of this event" (so you could capture an issue that is duration based).

nateberkopec:
@dcramer `site` is also undocumented, what's that supposed to do?

dcramer:
@nateberkopec ``site`` is a deprecated attributed that we have more or less killed. It was a Django thing.

nateberkopec:
👍

PR #70, opened 2016-10-19
Recommend `brew services start <service>`

mattrobenolt:
Running `pg_ctl -D` will at least dump stderr/stdout into your current running console and this is less than ideal. We can leverage `brew services` here now since we suggest `brew` anyways for installation.

/cc @torarnv


JTCunning:
Do it.

torarnv:
"Fixes" https://github.com/getsentry/sentry/issues/4403 and https://github.com/getsentry/sentry/issues/4402 :smile:

torarnv:
Follow-up: https://github.com/getsentry/sentry-docs/pull/73

PR #72, opened 2016-10-19
Add note about adding the OpenSSL libraries to LDFLAGS when building

torarnv:


torarnv:
cc @mattrobenolt

docs/internal/environment.rst
```diff
@@ -78,6 +78,10 @@ and some extra pieces that hold the development environment together::
 
     make develop
 
+.. note:: Some libraries depend on OpenSSL. Install via ``brew install openssl``, and make
+  sure to ``export LDFLAGS=-L/usr/local/opt/openssl/lib`` while building, so that the
```
dcramer: the indentation here might need lined up with the start of the next line

torarnv: Hmm, I used the same indentation as the notes above and below it I think...

dcramer: ah ok then just my RST paranoia probably

PR #79, opened 2016-11-16
Update cocoa doc for 1.0.0 release of sentry-swift

HazAT:


docs/clients/cocoa/dsym.rst
```diff
@@ -132,7 +132,10 @@ Shell: `/bin/bash`
     export SENTRY_ORG=___ORG_NAME___
     export SENTRY_PROJECT=___PROJECT_NAME___
     export SENTRY_AUTH_TOKEN=YOUR_AUTH_TOKEN
-    sentry-cli upload-dsym
+    ERROR=$(sentry-cli upload-dsym 2>&1 >/dev/null)
+    if [ ! -z "$ERROR" ]; then
```
mattrobenolt: Wouldn't it be more accurate to check returncode here instead of checking this string? It seems that this string will always be empty since you're piping all output to devnull.

HazAT: hmm I guess you are right ... could improve that a bit ...
basically I am capturing the output of the command (which only returns an output if it fails) and show it to the user ...

HazAT: so I played around with a few other ways of capturing exit codes and stderr/stdout ...

We only want the shell script to output something when `sentry-cli` fails, and then we want to capture the `error` and show it as a warning in xcode ...
If we let `sentry-cli` return the exit code without capturing, the xcode build will fail because a shell script produced an error ...

So I would propose we leave it like this

docs/clients/cocoa/dsym.rst
```diff
@@ -133,7 +133,7 @@ Shell: `/bin/bash`
     export SENTRY_PROJECT=___PROJECT_NAME___
     export SENTRY_AUTH_TOKEN=YOUR_AUTH_TOKEN
     ERROR=$(sentry-cli upload-dsym 2>&1 >/dev/null)
-    if [ ! -z "$ERROR" ]; then
+    if [ ! $? -eq 0 ]; then
```
HazAT: @mattrobenolt I guess this should be fine now, right?

mattrobenolt: Yeah, this is better. 👍

PR #97, opened 2017-02-07
Add section on managing event volume to pricing doc

jessla:
Need to make sure the links are 👌


jessla:
cc @ehfeng for 👀 and feedback. I wanted to add this to the docs so we have a linkable way to share best practices with folks.

ehfeng:
Add a bullet point point for Javascript
- whitelistUrl
- ignoreErrors

ehfeng:
the rest LGTM

PR #104, opened 2017-03-10
Add sample rate to SDK docs.

bretthoerner:
This is short and sweet, I didn't know what else to say.

### DIFF
--- a/docs/clientdev/overview.rst
+++ b/docs/clientdev/overview.rst
@@ -32,6 +32,7 @@ Additionally, the following features are highly encouraged:
 * Non-blocking event submission
 * Basic data sanitization (e.g. filtering out values that look like passwords)
 * Context data helpers
+* Event sampling
 
 Usage for End-users
 -------------------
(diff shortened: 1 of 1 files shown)


docs/clientdev/overview.rst
```diff
@@ -132,6 +133,20 @@ The resulting POST request would then transmit to::
...
 
+Event sampling
+--------------
+
+SDKs should allow the user to configure what percentage of events are actually
+sent to the server (the rest should be silently ignored). For example:
+
+.. sourcecode:: python
+
+    sample_rate = options.get('sample_rate') or 1.0
```
LewisJEllis: if `options.get('sample_rate')` returns 0 this expression will evaluate to 1.0 instead, i think theres a legitimate case for setting sample_rate to 0?

docs/clientdev/overview.rst
```diff
@@ -132,6 +133,20 @@ The resulting POST request would then transmit to::
...
+
+    # assuming random() returns a value between 0.0 and 1.0
```
LewisJEllis: maybe clarify 0 inclusive 1.0 exclusive, details might technically matter as ted pointed out in https://github.com/getsentry/raven-python/pull/978

docs/clientdev/overview.rst
bretthoerner: Welp, you're the CEO now.

docs/clientdev/overview.rst
bretthoerner: Good call, will do.

PR #110, opened 2017-03-21
add react native link

MeredithAnya:
Added react-native link to the sidebar

@HazAT @mitsuhiko


HazAT:
Since this is still open, can we also change this link?
https://github.com/getsentry/sentry-docs/blob/a3d6b3904fad7ae80bf785b5f62b16c288955f64/design/templates/layout.html#L215
To https://docs.sentry.io/clients/react-native/

MeredithAnya:
@HazAT Do you think we should keep it at all?

HazAT:
I think its fine if its there, but there shouldn't be multiple docs about react-native ... we probably have to decide if our new SDK is where we point our users to

PR #121, opened 2017-04-05
Expand releases guide

dcramer:


docs/learn/releases.rst
```diff
@@ -94,21 +103,55 @@ look like this:
...
-or :ref:`CLI <sentry-cli-commit-integration>` docs.
+.. code-block::
+
+    Prevent empty queries on users
+
+    Fixes SENTRY-317
+
+
+When Sentry sees this commit, we'll automatically annotate the matching
+issue with a reference to the commit, and upon deploy, we'll mark the issue
```
macqueen: uhh i have to double check, but i think it is actually when release is created not when deploy is created if that matters

dcramer: im gonna leave it as itll eventually need to be on deploy, and we can fake it

PR #123, opened 2017-04-06
try to just add class to note on releases

macqueen:
@ckj soo this is almost there, but is spilling off the page. i think maybe it's just a css issue now if you wanna check it out tomorrow?

screenshot 2017-04-05 18 09 50


ckj:
Nice! I'll take a look first thing

PR #127, opened 2017-04-24
Put postgres role creation in the setup instructions

chrismwendt:
This replaces the note referring to StackOverflow (introduced in https://github.com/getsentry/sentry-docs/pull/57) with a command that creates the `postgres` role if it doesn't already exist. This should save new users a few minutes of time on installation ⏱ 

@JTCunning Is this the command you used to create the role?

@LewisJEllis Bringing this to your attention since you reviewed https://github.com/getsentry/sentry-docs/pull/57

### DIFF
--- a/docs/internal/environment.rst
+++ b/docs/internal/environment.rst
@@ -37,8 +37,9 @@ Run the following to install, configure, and execute PostgreSQL as a daemonized 
     brew install postgresql
     brew services start postgresql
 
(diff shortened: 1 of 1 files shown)


JTCunning:
I'm fine with not opening up a link to SO but I'd definitely keep the note in there justifying the command, especially since `createuser` isn't idempotent.

EvanPurkhiser:
I fixed this in https://github.com/getsentry/sentry-docs/commit/291a83f156266bf4a150515f28ff3aae83dd2ed8

PR #137, opened 2017-06-13
Update loglevel options for breadcrumbs from warn to warning, to match api

tomasklingen:
The log level options of 
https://docs.sentry.io/learn/breadcrumbs/ and https://docs.sentry.io/clientdev/interfaces/breadcrumbs/ are not matching.


mattrobenolt:
Thanks! 🍦 🍕

PR #138, opened 2017-06-16
Improve SDK docs, fold in feature information doc.

bretthoerner:


docs/clientdev/features.rst
```diff
@@ -0,0 +1,156 @@
...
+user configuration at startup, often declared as a package/module prefix.
+
+Surrounding source in stacktrace
+--------------------------------
+
+Lines of source code to provide context in stacktraces. This is easier in interpreted languages,
+may be hard or impossible in compiled ones.
+
+Stack variable values
+---------------------
```
dcramer: Should note this might be calling args

bretthoerner: Not sure I follow, you just mean note that local variable data should include arguments passed to the functions?

dcramer: Php for example is actually just the function args which I think is more common to what we can get

PR #140, opened 2017-06-28
update learn/membership

MeredithAnya:
Add Billing member to list and add table

screen shot 2017-06-28 at 3 26 09 pm


MeredithAnya:
@dcramer maybe you want to make sure I got the permissions correct? Or if there are additional actions to add

dcramer:
looks correct to me

maybe "Can add/remove projects", which is admin/manager/owner

maybe same with teams, which is manager/owner

PR #148, opened 2017-07-28
Resolving via commits requires releases with commits only

ndmanvar:


docs/learn/releases.rst
```diff
@@ -144,8 +144,8 @@ For example, a commit message might look like this:
 
 
 When Sentry sees this commit, we'll automatically annotate the matching
-issue with a reference to the commit, and upon deploy, we'll mark the issue
-as resolved.
+issue with a reference to the commit, and upon release creation, we'll mark the issue
+as resolved in that release. **Note:** You must specify commits within release.
```
macqueen: can we make this more specific and say something like "You must either specify `commits` or `refs` when creating the release"?

benvinegar: +1

PR #153, opened 2017-08-14
Fix grammar and add info on quotas on old/new plan

ndmanvar:


ndmanvar:
used grammarly.com to find grammar mistakes, as well as some my intuition/brains.

PR #164, opened 2017-10-04
Cookies and hubspot

cameronmcefee:
The marketing folks would like to add Hubspot to the docs. This adds cookie permission banner we're using elsewhere and a link to Hubspot.

### DIFF
--- a/design/js/main.js
+++ b/design/js/main.js
@@ -343,6 +343,31 @@ function initRigidSearch() {
   }
 }
 
+window.Cookies = {
+  get: function(cname) {
+    var name = cname + '=';
+    var decodedCookie = decodeURIComponent(document.cookie);
+    var ca = decodedCookie.split(';');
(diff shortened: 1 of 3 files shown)


design/js/main.js
```diff
@@ -343,6 +343,31 @@ function initRigidSearch() {
   }
 }
 
+window.Cookies = {
```
cameronmcefee: I see we're already doing some stuff with cookies manually, so this may not be desirable.

design/templates/layout.html
```diff
@@ -378,5 +391,6 @@
         else {window.attachEvent('onload', _onload)}
     }());
   </script>
+  <script type="text/javascript" id="hs-script-loader" async defer data-src="//js.hs-scripts.com/3344477.js" class="js-privacy-shield-deferred"></script>
```
mattrobenolt: Just make this explicitly https. No reason to use protocol relative.

PR #171, opened 2017-10-16
Replace global EPM with global EPH in rate-limit doc

ndmanvar:


docs/pricing.rst
```diff
@@ -53,11 +53,6 @@ On-demand spending cap
 
 You can set a maximum monthly on-demand bill amount by setting an on-demand spending cap. Once you reach the spending cap, you will still be able to access Sentry and prior event data. However, any additional events you send will be rejected.
 
-Per-minute rate limits
-``````````````````````
-
-You can set the maximum number of events per minute that Sentry will accept. This allows you to protect yourself from consuming all of your month’s capacity in the event of sudden spikes.
-
```
jessla: instead of removing it entirely, what do you think of something like:

# Rate limits

Every organization can set an hourly rate limit (1,000 - 1,000,000 events) to prevent massive spikes from consuming your monthly quota. 

Large and Enterprise plans have the option to configure rate limits for each project that let you specify the number of events and time interval for rate limiting.

ndmanvar:
@jessla Sounds good. Made that change. Could you please review again?

PR #174, opened 2017-10-20
change 'mkvirtualenv .' to be 'mkvirtualenv sentry'

Chrissy:
I'm new to python, and this line in the docs really got me.

image

image


Chrissy:
@mattrobenolt does this seem reasonable? 

My only thought was that maybe the `.`was there because the doc envisions a virtualenv in the project root? This obviously isn't happening when we set up but maybe could explain the funky command?

PR #175, opened 2017-10-26
Update internal docs to pin postgres to a specific version

mattrobenolt:


kamilogorek:
```

This formula is keg-only, which means it was not symlinked into /usr/local,
because this is an alternate version of another formula.

If you need to have this software first in your PATH run:
  echo 'export PATH="/usr/local/opt/postgresql@9.4/bin:$PATH"' >> ~/.bash_profile
```

PR #176, opened 2017-10-26
fix: Note that you must do brew tap to use version pinned Postgres

bretthoerner:


docs/internal/environment.rst
```diff
@@ -34,7 +34,9 @@ PostgreSQL
 PostgreSQL is the primary database that Sentry uses for all persistent storage.
 Run the following to install, configure, and execute PostgreSQL as a daemonized server::
 
+    brew tap petere/postgresql
```
mattrobenolt: Why is this some random person's tap?

Issue #179, opened 2017-11-02
API authentification example is not working


ablanchard:
Hi,

I started the api documentation by API authentification https://docs.sentry.io/api/auth/#
And I get a 404 on the first example `$ curl -H 'Authorization: Bearer {TOKEN}' https://sentry.io/api/0/projects/1/`
The trailing `1/` is causing the problem.
The example should be `$ curl -H 'Authorization: Bearer {TOKEN}' https://sentry.io/api/0/projects/`

PR #180, opened 2017-11-02
Remove project_id from curl example

ndmanvar:
`1` is not a valid project, so users are not getting a response back when running.
```
$ curl -H 'Authorization: Bearer XXXXXXX' https://sentry.io/api/0/projects/1
$
```


ndmanvar:
Fixes https://github.com/getsentry/sentry-docs/issues/179

ndmanvar:
👍

PR #183, opened 2017-11-20
docs: Add note about project root address

kamilogorek:
I'm not sure sure if it's called `root address of the server`, as I don't want to tear down my env for the 4th time :P


mattrobenolt:
cc @getsentry/javascript

PR #201, opened 2018-01-03
fix: Update node and yarn in docker file

HazAT:
Docs build is failing since a npm dep in the `sentry` repo needs `node >= 5.10.0`
See:
https://freight.getsentry.net/deploys/docs/production/1193
```
error extract-text-webpack-plugin@3.0.2: The engine "node" is incompatible with this module. Expected version ">= 4.8 < 5.0.0 || >= 5.10".1:33:52 pm
error Found incompatible module
```

Soooo this upgrades both deps in the docker file to the latest version.
Probably that's not the smartest idea but I don't know the node/yarn version Sentry requires which would be the wiser choice I guess.

To prevent such thing in the future we could add something like this 
```
"engines": {
    "node": ">= 4.8 < 5.0.0 || >= 5.10"
  },
```
in Sentrys `package.json` so no one installs a dep that breaks other builds.

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,19 +1,20 @@
 FROM python:2.7
 
 RUN set -x \
-    && export NODE_VERSION=9.3.0 \
+    && export NODE_VERSION=8.9.1 \
     && export YARN_VERSION=1.3.2 \
     && export GNUPGHOME="$(mktemp -d)" \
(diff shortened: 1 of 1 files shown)


mattrobenolt:
This won't be enough, unfortunately. I'll take it over and do the same thing we do in the other repos.

PR #212, opened 2018-03-14
fix: update docs to recommend Python 2

cameronmcefee:
Homebrew installs Python 3 by default, which doesn't work with Sentry. I'm not sure if 2.7 is required specifically, but I figured I'd err on the side of verbosity.


cameronmcefee:
I thought so, but I didn't have any issues doing it this way ¯\_(ツ)_/¯.

PR #221, opened 2018-04-06
feat: Replace ___MINIDUMP_URL___ with the actual API url

jan-auer:
This adds a new magic string for rendering the minidump endpoint in docs:

**NOTE**: That the `/api` and `?sentry-key=` portions are currently hard coded. I couldn't come up with a nice way to infer those. This might become relevant once we start ingestion via Relay.

## Example (Electron Docs)

screen shot 2018-04-06 at 09 05 20

### DIFF
--- a/design/js/main.js
+++ b/design/js/main.js
@@ -44,12 +44,12 @@ function dsnToHtml(parsedDsn, pub) {
 
 function tagInteractiveBlocks(parent) {
   parent.find('div.highlight pre').each(function() {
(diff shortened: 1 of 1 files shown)


jan-auer:
[This also has to be replaced in Sentry](https://github.com/getsentry/sentry/blob/234abbd7570e853db3df165b58e8359051d6e276/src/sentry/api/endpoints/project_docs_platform.py#L16-L20)

PR #233, opened 2018-05-15
Simplified OS X dev environment setup

JoshuaRLi:
The `Brewfile` updates in https://github.com/getsentry/sentry/pull/8453 allows for the development environment setup on Mac OS X to be simplified, which is reflected in these documentation changes.


JoshuaRLi:
`make develop` in the main repo runs `install-system-pkgs` which runs `brew bundle`, which interprets the `Brewfile`.

PR #237, opened 2018-05-21
Add Sentry 9 environment docs

saragilford:
Documentation for our new environment feature.


docs/learn/environments.rst
```diff
@@ -0,0 +1,50 @@
+Environments
+----------------
```
mattrobenolt: This is going to be invalid RST and will cause docs to fail to deploy.

Headings need to have a matching length line under it.

So like:

```rst
Environments
------------
```

PR #249, opened 2018-06-19
Add getting started

maheskett:


src/collections/_documentation/error-tracking/quickstart.md
```diff
@@ -0,0 +1,101 @@
...
+- The project ID which the authenticated user is bound to.
+
+You\'ll have a few options for plugging the DSN into the SDK, depending
+on what it supports. At the very least, most SDKs will allow you to set
+it up as the `SENTRY_DSN` environment variable or by passing it into the
+SDK\'s constructor.
+
+For example for the JavaScript SDK it works roughly like this:
+
+    import Raven from 'raven-js'
```
cameronmcefee: Where appropriate, let's do these as highlighted blocks.

```
{%- highlight javascript -%}
import Raven from 'raven-js'     
Raven.config('___PUBLIC_DSN___')
{%- endhighlight -%}
```

cameronmcefee:
Just to verify, since they are different from the PR title, are all the renames an intentional change? I don't *really* care, but it might be nice to keep them as a separate pr since they seem unrelated.

maheskett:
Ya, that was intentional but I'll move it to a separate PR

src/collections/_documentation/error-tracking/quickstart.md
maheskett: The plan is to replace these later down the line with a component to handle DSNs?

cameronmcefee: Yep, and it's already built, though it's currently broken if you don't have getsentry running locally. I have a fix that I'll push tomorrow that should resolve that.

ehfeng:
Let's update the links to use the "new links" (as mapped in the [spreadsheet](https://docs.google.com/spreadsheets/d/1f7D_izcmNcJ42FqovIsej95_65r4Yggl-k_u2QrEaRQ/edit#gid=0)).

We should probably update your python script (when processing links) to do the same automatically.

maheskett:
@ehfeng We decided against that format. I'll link you the thread offline

cameronmcefee:
> We decided against that format

If you're talking about me, I'm fine with doing regular links off-site, but we should use the `{% link %}` syntax for anything within the docs themselves. It'll protect us against things like accidental removals, typos, and will ensure the links work if we ever update permalinks.

maheskett:
Decided offline to keep with the markdown URL format but add a TODO to fix these later `[Python](TODO /clients/python/)`, since we don't have new URL paths finalized yet and the build fails if the files don't exist.

PR #250, opened 2018-06-20
style: update session stack screenshot

cameronmcefee:
Sessionstack requested we update this screenshot. The only difference I see is that the text field is focused. ¯\_(ツ)_/¯ I also compressed it by 90% so that's a win.


mattrobenolt:
Great work @cameronmcefee Truly fantastic.

cameronmcefee:
Can't wait to put this on my resume

PR #252, opened 2018-06-21
Add suspect commits info/docs and image for commits

ndmanvar:
Image shows suspect commits and suggested owners on Issue Details page.


docs/learn/releases.rst
```diff
@@ -10,7 +10,8 @@ What is a Release?
 ------------------
 
 Releases are used by Sentry to provide you with additional context
-when determining the cause of an issue. 
+when determining the cause of an issue, as well for
+`un-minification <https://docs.sentry.io/clients/javascript/sourcemaps/>`__.
```
mattrobenolt: I'm not sure what the correct word here is, but `un-minification` isn't a word. I get what it means, and I can't really think of a better word.

mattrobenolt: I don't particularly care, I'm not really educated enough to say how this should be written. I approved it already, so it's your call or ask someone who knows English better than me.

PR #256, opened 2018-06-26
add Integrations category with Asana page

mikellykels:
add Integrations category with Asana page

https://docs.sentry.io/integrations/asana/


ehfeng:
Check why build is failing?

mikellykels:
So it looks like all of ours are failing - not really sure how to fix that?

PR #259, opened 2018-06-27
Added Account/SSO

LSSangha:
Added Single Sign-On to the new docs.  Previously was (https://docs.sentry.io/learn/sso/) now (https://docs.sentry.io/account/sso).


ehfeng:
Hey, can you run this locally and include a screenshot? Because of the branch system, it's hard to be able to compare the two without running it locally myself.

PR #260, opened 2018-07-01
Remove referenced to public/private keys

timfish:
As far as I understand, only the public key is now used....


mitsuhiko:
Closed due to docs migration

PR #263, opened 2018-07-17
[WIP] feat: Docs for new SDK

HazAT:
This PR is meant to provide the basic structure for new SDK docs.
**Do not merge yet.**

- Add support for headers in dropdown

I was trying to add "headers" in the dropdown because while we mostly have languages depending on the SDK you run you still have different setup steps depending on the platform.

@cameronmcefee If you don't like that I mess with the base templates, I don't mind if you want to do it and I describe what kind of problem I am trying to solve.

docs inst mov

Things I've noticed so far:
- A lot of content is in the `installation_content:` section on top of the page and it's not really markdown highlighted. Which works but my syntax highlighting looks like this: 
screenshot 2018-07-17 15 43 15
Maybe there is a better solution for this since I think the majority of the content lives inside this.
- There are no anchor links, like in the old docs, where I can link to a headline of one page.
- We need some kind of version variable like: https://github.com/getsentry/sentry-docs/pull/263/files#diff-09c82791fbcc3ae9624c90bbea8367e8R24
- I did not find a way to do a "soft wrap" meaning create a new line without a paragraph and just putting the next sentence into a new line.
- If you click on an `<a href="#TODO" ...` the page goes crazy with some reload loops.

### DIFF
--- a/src/_data/categories.yml
+++ b/src/_data/categories.yml
@@ -1,4 +1,7 @@
 # This file dictates sort order and display properties for top-level categories
+- title: Setup
+  icon: releases
+  slug: setup
 
 - title: Error Tracking
   icon: error-tracking
--- a/src/_data/platforms.yml
+++ b/src/_data/platforms.yml
@@ -1,6 +1,18 @@
 - name: JavaScript
   slug: javascript
 
+- name: JavaScript / Browser
+  slug: javascript-browser
+
+- name: JavaScript / Node
+  slug: javascript-node
+
+- name: JavaScript / Electron
+  slug: javascript-electron
+
+- name: JavaScript / Cordova
+  slug: javascript-cordova
+
 - name: Ruby
   slug: ruby
 
--- a/src/_includes/components/platform_content.html
+++ b/src/_includes/components/platform_content.html
@@ -48,15 +48,27 @@ utilize their tokens across sites.
       </a>
 
       <div class="nav dropdown-menu" id="{{ __uniq }}-tab" role="tablist" aria-hidden="true">
+        {%- assign category = '' -%}
         {%- for __example in __platforms -%}
           {%- if __default[0] == __example[0] -%}
           {% assign __is_default = true %}
           {%- else -%}
           {% assign __is_default = false %}
           {%- endif -%}
+          {%- if __example[0] contains '/' -%}
+          {%- assign split = __example[0] | split: '/' -%}
+          {%- if split[0] != category -%}
+          <div class="dropdown-header">{{ split[0] }}</div>
+          {%- endif -%}
+          {%- assign category = split[0] -%}
+          {%- assign __name = split[1] -%}
+          {%- assign __slug = __example[0] | slugify -%}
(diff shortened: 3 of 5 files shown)


cameronmcefee:
### Organization

> A lot of content is in the installation_content: section on top of the page and it's not really markdown highlighted. 

This was a question I brought up early and we settled on the frontmatter approach to keep the documents together. We could also store the content separately in a collection which we can reference like `{{ site.installation_examples.javascript }}` or store them in a folder with the document and then use includes like `{% include_relative javascript.md %}` which will bring the highlighting back, but those methods will separate the content from the main document.

The current method is nice, in that it's in the form of Objects within Jekyll, so it's easier to do validation around. For example, I can test the key name against the platform list to ensure it's a supported platform and is named according to the token list that is referenced within getsentry.

On the other hand, the relative includes is a pretty nice organizational option because the structure then looks like:

```
installation /
   index.md
   javascript.md
   angular.md
```

Validation becomes a little less obvious, having to rely on the name of the file as the slug referenced. With that said, it might be a little more clear overall. I'll test it out when revising the platform picker.

### Deep links

> There are no anchor links, like in the old docs, where I can link to a headline of one page.

All H tags have an ID attribute on them that is the content of the heading, lowercase, with dashes instead of spaces. (let me know if I misunderstand or if this is not true for you). 

```
### My Heading

<h3 id="my-heading">My Heading</h3>
```

### Versions

> We need some kind of version variable like: https://github.com/getsentry/sentry-docs/pull/263/files#diff-09c82791fbcc3ae9624c90bbea8367e8R24

I assume you'll need multiples of these depending on the SDK, right? We should add a file like `versions.yml` to the *data* folder and then we can reference them like `{{ site.data.versions.raven_js }}`

### Softwrap

> I did not find a way to do a "soft wrap" meaning create a new line without a paragraph and just putting the next sentence into a new line.

Use a `>` instead of a `|` at the beginning of the yaml string. I believe that will [replace newlines with spaces.](https://stackoverflow.com/questions/3790454/in-yaml-how-do-i-break-a-string-over-multiple-lines/21699210)

### Reload loops

> If you click on an <a href="#TODO" ... the page goes crazy with some reload loops.

Good catch, it's probably an issue with the dynamic loading I'll look into that

HazAT:
### Organization
👍 let me know how we handle this, it was just my two cents. 
While the other approach might be more files I think it should be more manageable overall.

### Deep links

So if I do 
```
### My Heading

<h3 id="my-heading">My Heading</h3>
```
the result is:
screenshot 2018-07-18 16 41 51

None is clickable or linkable like in the old docs.

### Versions
👍 
Yeah, a simple file called `versions.yml` with arbitrary `string: string` mapping like `raven-js:3.2.1` would be sufficient.

### Softwrap
Thanks 👍 

### Reload loops
👍

cameronmcefee:
### Organization

I'll take a stab at the relative includes method. I didn't consider that option when doing the original planning, so it may work out ok with a little additional engineering to make the process cleaner.

### Deep links

Oh, I see. I don't think our parser supports wrapping the headings with links, but I could add some JavaScript to do it. I'll add that to my todo list

### Versions

No problem, I'll do that today.

cameronmcefee:
Closing in favor of #268 and other PRs to address the various feedback items.

PR #271, opened 2018-07-26
Fix broken links

cameronmcefee:


jan-auer:
@cameronmcefee turns out we were missing another `_` for the indirection 🤷‍♂️

PR #283, opened 2018-08-27
fix: restore old nginx config

cameronmcefee:


nginx.conf
```diff
@@ -2,25 +2,36 @@ server {
...
-  error_page 403 =404 /404/index.html;
   root /usr/share/nginx/html;
 
-  add_header X-Frame-Options sameorigin always;
-  add_header X-Content-Type-Options nosniff always;
-  add_header X-XSS-Protection "1; mode=block" always;
+  rewrite ^/(?:hosted|on-premise)/(.*)$ /$1$is_args$args redirect;
 
-  location / {
-    try_files $uri $uri.html $uri/ $uri/index.html =404;
```
mattrobenolt: We need this back.

And all the add_header blocks.

And the `error_page` blocks.

Let's try again, and purely copy over the redirects.

nginx.conf
```diff
@@ -2,25 +2,36 @@ server {
...
+
+  location = /api/releases/post-project-releases/ {
+    return 302 /api/releases/post-organization-releases/$is_args$args;
+  }
+
+  location = /api/releases/delete-project-release-details/ {
+    return 302 /api/releases/delete-organization-release-details/$is_args$args;
   }
 
-  location /_assets/ {
```
mattrobenolt: Instead of removing this, change it to `/assets/` since that's where the docs assets go.

nginx.conf
```diff
@@ -2,25 +2,36 @@ server {
   listen 80;
   server_name _;
 
+  gzip on;
   gzip_static on;
   gzip_vary on;
+  gzip_types application/javascript text/css;
 
-  error_page 404 /404/index.html;
```
mattrobenolt: Well, I guess this in a different location for docs?

PR #292, opened 2018-08-29
(fix) Remove Sprint.ly from the integrations page

ceorourke:
Fixes SE-87


mattrobenolt:
@ceorourke I switched review to @cameronmcefee.

I dunno how docs work.

PR #299, opened 2018-08-29
Fix: send reload pageview on pushstate

cameronmcefee:


src/_js/lib/DynamicLoad.js
```diff
@@ -71,9 +71,18 @@ class DynamicLoader {
 
     const { pathname, hash, search } = url;
     const done = function() {
+      // Send reload a pageview if this is a new page
+      if (pathname !== location.pathname) {
+        console.log('reload');
```
adhiraj: Remove

PR #309, opened 2018-09-03
Manual setup instructions in settings.gradle

pettersamuelsen:
The docs were missing a step in the manual setup for React Native Android. The only way I found out this missing step was to run `react-native link react-native-sentry` and look at the diff.


HazAT:
Thx! :)

PR #323, opened 2018-09-06
fix: run the header linker on page.didUpdate

cameronmcefee:
This fixes the header links that seemed to disappear after navigating to a new page

### DIFF
--- a/src/_js/lib/HeaderLinker.js
+++ b/src/_js/lib/HeaderLinker.js
@@ -1,4 +1,4 @@
-const init = function() {
+$(document).on('page.didUpdate', function() {
   $('h1,h2,h3,h4,h5,h6').each((i, el) => {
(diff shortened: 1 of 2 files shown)


src/_js/lib/HeaderLinker.js
```diff
@@ -1,4 +1,4 @@
-const init = function() {
+$(document).on('page.didUpdate', function() {
```
dcramer: til thats an event

lobsterkatie:
Cool - thanks for fixing, @cameronmcefee !

src/_js/lib/HeaderLinker.js
cameronmcefee: It's a thing I added myself that fires after the dynamic update. It's not native.

PR #325, opened 2018-09-07
ref: Use slugs for user data

cameronmcefee:


src/_js/lib/User.js
```diff
@@ -71,12 +72,12 @@ const formatAPIURL = function(dsn) {
...
 const formatProjectLabel = function(project = {}) {
-  const { projectName, teamName } = project;
-  if (!projectName && !teamName) return null;
+  const { projectSlug, organizationSlug } = project;
+  if (!projectSlug && !organizationSlug) return null;
 
-  return projectName.indexOf(teamName) === -1
-    ? `${teamName} / ${projectName}`
-    : projectName;
+  return projectSlug.indexOf(organizationSlug) === -1
```
dcramer: fwiw you can *always* show org + project slug, just like github

PR #331, opened 2018-09-11
fix: resolve min-height layout issues

cameronmcefee:


src/_assets/css/_includes/grid.scss
```diff
@@ -15,25 +15,32 @@
...
+
+.main-scroll {
+  overflow: scroll;
+}
 
-  .overflow-md-scroll {
-    overflow: scroll;
+.document-wrapper {
+  display: flex;
+  min-height: 100%;
```
markstory: :+1: This fixes the issue for me locally.

PR #342, opened 2018-09-18
fix: Remove overloaded name method

cameronmcefee:
Some docs are broken because the name switched to id

### DIFF
--- a/src/_js/lib/User.js
+++ b/src/_js/lib/User.js
@@ -103,7 +103,7 @@ const constructDSNObject = function(project = {}) {
   return {
     id: project.id || -1,
     group: escape(project.organizationName) || 'Example',
-    PROJECT_NAME: escape(formatProjectLabel(project)) || 'Your Project',
+    PROJECT_NAME: escape(project.name) || 'Your Project',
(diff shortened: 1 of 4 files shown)


src/_js/lib/User.js
```diff
@@ -103,7 +97,7 @@ const constructDSNObject = function(project = {}) {
   return {
     id: project.id || -1,
     group: escape(project.organizationName) || 'Example',
-    PROJECT_NAME: escape(formatProjectLabel(project)) || 'Your Project',
+    PROJECT_NAME: escape(project.name) || 'Your Project',
```
bretthoerner: OK, I see what you were asking about name vs id now. I didn't realize `PROJECT_ID` is the slug in this repo. That's actually what we need for the Java docs, and I guess `PROJECT_NAME` used to actually be the slug so it just worked? (I didn't author those Android docs so this is new to me)

Anyway, I'll just push to this branch and swap out for `PROJECT_ID` in the Java docs?

cameronmcefee: 👍Works for me

cameronmcefee:
I've inquired the same in #343

PR #346, opened 2018-09-18
feat(docs) Add a docker based api docs builder

markstory:
Add a docker container that will allow us to build API docs in a docker host. The generated tar file will be uploaded to GCS where it can be consumed and integrated into the sentry-docs project.

Docs can be built with

```bash
docker build -t sentry:apidocs -f Dockerfile.apidocs
docker run --rm -v $(pwd)/apidocs:/usr/src/output sentry:apidocs
```

The generated markdown files can then be moved in place with

```bash
cp -R apidocs/markdown/* _site/api/
```

For the time being updating the API docs will be a manual on-demand process.

Refs getsentry/sentry#9643

### DIFF
--- a/Dockerfile.apidocs
+++ b/Dockerfile.apidocs
@@ -10,10 +10,7 @@
 # The container will dump the generated documentation in markdown and JSON
 # formats into the /usr/src/output directory which you should mount as a volume
 #
-FROM python:2.7.11-slim
-
-RUN groupadd -r redis --gid=998 && useradd -r -g redis --uid=998 redis \
-    && groupadd -r postgres --gid=999 && useradd -r -g postgres --uid=999 postgres
+FROM python:2.7.15-slim-jessie
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
         clang \
@@ -27,9 +24,10 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         libxml2-dev \
         libxslt-dev \
         libyaml-dev \
-        llvm-3.5 \
+        llvm \
         bzip2 \
         make \
+        redis-server \
         unzip \
     && rm -rf /var/lib/apt/lists/*
 
@@ -37,10 +35,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 ENV PIP_NO_CACHE_DIR off
 ENV PIP_DISABLE_PIP_VERSION_CHECK on
 ENV PYTHONUNBUFFERED 1
-
-# Install Redis
-ENV DEBIAN_FRONTEND=noninteractive
-RUN apt-get update && apt-get install -y --no-install-recommends redis-server
 
 RUN mkdir -p /usr/src/bin
 COPY scripts/build-api-docs /usr/bin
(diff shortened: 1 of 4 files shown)


scripts/build-api-docs
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+echo "--> Fetching sentry source"
+curl https://codeload.github.com/getsentry/sentry/zip/master --silent --output sentry.zip
```
mattrobenolt: Is this url reliable?

I'd just use the canonical https://github.com/getsentry/sentry/archive/master.zip

It'll incur a redirect, but it's not some seemingly hidden API. But maybe you know something I don't.

Dockerfile.apidocs
```diff
@@ -0,0 +1,51 @@
...
+#
+#   Build the container:
+#     $ docker build -t sentry:apidocs -f Dockerfile.apidocs .
+#   Run the container:
+#     $ docker run --rm -v $(pwd):/usr/src/output sentry:apidocs
+#
+# The container will dump the generated documentation in markdown and JSON
+# formats into the /usr/src/output directory which you should mount as a volume
+#
+FROM python:2.7.11-slim
```
mattrobenolt: Bump this to `python:2.7.15-slim-jessie` since I think there will be weird issues with this version.

Dockerfile.apidocs
```diff
@@ -0,0 +1,51 @@
...
+
+RUN groupadd -r redis --gid=998 && useradd -r -g redis --uid=998 redis \
```
mattrobenolt: None of this matters here. You're also creating a postgres user and we don't install Postgres.

Dockerfile.apidocs
```diff
@@ -0,0 +1,51 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+# Sane defaults for pip
+ENV PIP_NO_CACHE_DIR off
+ENV PIP_DISABLE_PIP_VERSION_CHECK on
+ENV PYTHONUNBUFFERED 1
+
+# Install Redis
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get install -y --no-install-recommends redis-server
```
mattrobenolt: Would just add this into the list above. No reason to have it separate.

Dockerfile.apidocs
```diff
@@ -0,0 +1,51 @@
+# Sentry api documentation generation environment
+#
+# Instructions:
+#
+#   Build the container:
```
mattrobenolt: We should wrap all of this into a simple that thing is just `bin/update-api-docs` or whatever, that does these docker commands and mounts the volume, etc.

Dockerfile.apidocs
markstory: Good point.

Dockerfile.apidocs
markstory: Ok.

scripts/build-api-docs
markstory: No I was just skipping the redirect as I forgot about `-L` in curl earlier.

PR #356, opened 2018-09-19
fix: Broken link to forum

untitaker:


src/collections/_documentation/clients/react-native/sourcemaps.md
```diff
@@ -2,7 +2,7 @@
 title: 'Sourcemaps for Other Platforms'
 ---
 
-Currently automatic sourcemap handling is only implemented for iOS with Xcode and Android with gradle. If you manually invoke the [react-native packager](https://github.com/facebook/react-native/tree/master/packager) you can however get sourcemaps anyways by passing _–sourcemap-output_ to it.
+Currently automatic sourcemap handling is only implemented for iOS with Xcode and Android with gradle. If you manually invoke the [react-native packager](https://github.com/facebook/metro) you can however get sourcemaps anyways by passing _–sourcemap-output_ to it.
```
untitaker: @HazAT is this even the same thing? It's the only packager I found

src/collections/_documentation/server/plugins.md
```diff
@@ -82,5 +82,3 @@ The following plugins are available and maintained by members of the Sentry comm
 -   [sentry-youtrack](https://github.com/bogdal/sentry-youtrack)
 -   [sentry-zabbix](https://github.com/m0n5t3r/sentry-zabbix)
 -   [sentry-zendesk](https://github.com/ESSS/sentry-zendesk)
-
-Have an extension that should be listed here? Submit a [pull request](https://github.com/getsentry/sentry/edit/master/docs/plugins.rst) and we’ll get it added.
```
untitaker: Since we have this kind of link on every doc I felt that this is not necessary

.travis.yml
```diff
@@ -9,5 +9,10 @@ cache:
   yarn: true
   directories:
     - node_modules
-script:
-  - bin/test
+
+matrix:
```
cameronmcefee: So is this doubling our build times because we're building it twice? Is there a way we can run the test without building jekyll twice?

untitaker: We can do it in one build but then we have to run tests and linkcheck sequentially. I am not sure what's faster.

link check is definetly much slower than any build procedure though

untitaker: Ah, I mean I guess we can do one build and parallelize within the same travis job, but then the output is mixed

cameronmcefee: I guess we can see how it works out.

untitaker:
Hmm, I am not sure if we'll be able to run this without making the build either flaky or really slow.

untitaker:
Travis has persistent issues connecting to some hosts, so I removed the new build again

PR #357, opened 2018-09-19
feat: automated api generation

cameronmcefee:
@markstory has a PR open that dumps the api to markdown. This hooks up the dump he's generating with a template to render it so we can have up-to-date api docs again.


markstory:
@cameronmcefee I've made a few changes to add:

* Removed bullet points from parameter lists. The old documents omitted the bullet if the list contained only one element but displayed bullets when there was more than one element. I normalized on no bullets.
* Inline formatting in parameter descriptions.
* Display warning notices.
* Display an alertbox when `user-context` is required for authentication.

I reviewed all the generated documentation and compared it with the current site and it looks good. There are a few duplicate document titles but that is a problem in the source documentation, which I plan on fixing.

PR #363, opened 2018-09-20
fix: Remove uglifier

untitaker:
uglifier is unable to parse the JS changes commit
c784cdfda5da54feed6666abea3165adca6f0486. Particularly the variable name
`anchor` is somehow not accepted.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,6 @@ group :jekyll_plugins do
   }
   gem 'jekyll-assets', '3.0.11'
   gem 'terminal-notifier'
-  gem 'uglifier'
   gem 'autoprefixer-rails'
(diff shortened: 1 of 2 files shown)


untitaker:
Just adding @mattrobenolt and @mitsuhiko so they can have the satisfaction of hitting approve

untitaker:
FTR, uglifier chokes on this call: https://github.com/getsentry/sentry-docs/blob/e4007b328009695644c2c1447928e435834c4327/src/_js/lib/PlatformContent.js#L139-L141

Just giving `location.hash` a local name, and passing that local name fixes the build issue:

```
  const foo = location.hash;
  history.replaceState({}, '', updateUrlPlatform(location.href, slug, foo));
```

PR #368, opened 2018-09-21
fix: Fix anchors again

untitaker:
The original fix for this was under the assumption that `location.href`
does not contain the anchor.

### DIFF
--- a/src/_js/lib/PlatformContent.js
+++ b/src/_js/lib/PlatformContent.js
@@ -45,7 +45,8 @@ const updateUrlPlatform = function(url, slug) {
   const url_split = url.split("#");
   const { url: path, query } = qs.parseUrl(url_split[0]);
   query.platform = slug;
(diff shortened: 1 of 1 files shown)


src/_js/lib/PlatformContent.js
```diff
@@ -41,10 +41,11 @@ const verifyPlatformSlug = function(slug) {
...
 //
 // Returns a string
-const updateUrlPlatform = function(url, slug, anchor) {
-  const { url: origin, query } = qs.parseUrl(url);
+const updateUrlPlatform = function(url, slug) {
+  const url_split = url.split("#");
+  const { url: path, query } = qs.parseUrl(url_split[0]);
   query.platform = slug;
-  return `${origin}?${qs.stringify(query)}${anchor}`;
+  return `${path}?${qs.stringify(query)}#${url_split[1]}`;
```
markstory: Won't this append `#` even when there is no anchor?

PR #380, opened 2018-09-25
feat: Document new aws lambda integration in Python

untitaker:


src/_data/platform_icons.yml
```diff
@@ -70,6 +70,10 @@
   link: _documentation/platforms/python/celery.md
   platformicon: python
 
+- name: AWS Lambda (Python)
```
untitaker: I'd imagine we will have multiple AWS lambda "platforms" in the wizard for lack of a better format

PR #386, opened 2018-09-25
ref: simplify DynamicLoad

cameronmcefee:
My reviewer choices are arbitrary, please review if you feel relevant to do so.

DynamicLoad was beginning to feel too complicated and the confusion was causing bugs. This refactor simplifies the logic to try to make it a little easier to understand. Now:

- I removed the loaders concept (methods of handling different urls).
- Rather than listening for all anchor clicks, it only responds when a click:
  - Does not have a meta key pressed
  - Is not an external url
  - Is not on the same page
- In all the cases above, the default behavior is followed, that is, meta key links open in new tabs, external urls are followed, and links on the same page (like hashes) use the browser behavior or (like `#` links used as buttons) follow whatever their given event handler does.
- I moved the Reload page view and feedback stuff to event handlers on `page.didUpdate` to make them easier to understand in the lifecycle.

With something this crucial I want to get tests on it, but I couldn't manage to listen to the custom jquery events with jest. I'll swing back around on that at some point.

### DIFF
--- a/src/_js/lib/DynamicLoad.js
+++ b/src/_js/lib/DynamicLoad.js
@@ -1,112 +1,77 @@
 import { fullPath } from './Helpers';
 
-let currentPathName = document.location.pathname;
-
-const replaceContent = function(html) {
-  const $page = $(html.replace('<!DOCTYPE html>', ''));
-  const $new = $page.find('[data-dynamic-load]');
-  const $title = $page.filter('title').eq(0);
-  $('head title').text($title.text());
-
-  $new.each(function(i, el) {
-    const $el = $(el);
-    const slug = $el.data('dynamic-load');
-    const content = $el.html();
-    $(`[data-dynamic-load="${slug}"]`).html(content);
-  });
+const isDifferentDomain = function(here, there) {
+  return here.protocol !== there.protocol && here.host !== there.host;
 };
 
-const isSameDomain = function(here, there) {
-  return here.protocol === there.protocol && here.host === there.host;
+const isSamePage = function(here, there) {
+  return here.pathname === there.pathname && here.host === there.host;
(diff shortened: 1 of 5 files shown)


src/_js/lib/DynamicLoad.js
```diff
@@ -1,112 +1,77 @@
...
+$(window).on('popstate', event => {
+  load(document.location);
+});
+
+$(document).on('page.didUpdate', function(event) {
+  // Handle vertical scroll position of new content
+  const hash = window.location.hash.replace('#', '');
+  if (hash) {
+    const $el = $(hash).get(0);
+    if ($el) $el.scrollIntoView();
```
cameronmcefee: I'm fairly sure this doesn't work, but it's what was there. I'm going to take a stab at fixing it in the morning

src/_js/lib/DynamicLoad.js
```diff
@@ -1,112 +1,77 @@
...
-    const slug = $el.data('dynamic-load');
-    const content = $el.html();
-    $(`[data-dynamic-load="${slug}"]`).html(content);
-  });
-};
-
-const isSameDomain = function(here, there) {
-  return here.protocol === there.protocol && here.host === there.host;
+const isDifferentDomain = function(here, there) {
+  return here.protocol !== there.protocol && here.host !== there.host;
```
kamilogorek: `||` instead of `&&`, as this should return `false` when either or both are false.

cameronmcefee:
I've added tests with near complete coverage (I ignored a couple tests for lines covering scroll position so I don't hold this back longer while I figure that out). The major change is that I had to put this back into a class so I'd be able to mock the functions correctly.

PR #392, opened 2018-09-26
ref: Restructure documentation for before-send and before-breadcrumb

untitaker:


untitaker:
cc @mitsuhiko @dcramer I saw cramer briefly complain about "hint-based customization" being too prominent. I think this is a better title.

PR #394, opened 2018-09-26
fix: Only show platform query on pages with platform content

cameronmcefee:
It's kind of weird seeing a url like /platforms/go?platform=python. This disables the automatic addition of the platform slug on pages that don't have platform content, since the slug serves only as a helper to either force a platform to display or to provide a permalink to a page with a platform shown.


cameronmcefee:
Ah, I disabled prettier because it was modifying the whole document and let that one slip. Thanks for pointing it out

PR #402, opened 2018-09-27
Adds docs for enabled config key

kbremner:


untitaker:
This key is not available for the Python SDK which is why we neglected to document this, because this section applies to all new-style SDKs. Generally you can always disable the client by passing an empty string, `null` or something similar as DSN.

@bruno-garcia @HazAT is this available in your SDKs? I wonder if we should add it to Python. I don't really understand how this is more useful than passing an empty DSN

kbremner:
Hi, thanks for the feedback!

We found the `enabled` key when using the JavaScript SDK and I assumed it was in all SDKs.

I hadn't realised that passing a blank DSN would have the same effect, happy to update the PR to add docs for that behaviour instead?

untitaker:
Yes I think that would be better, but I need to confirm tomorrow if that actually works in all SDKs.

bruno-garcia:
It's not in the dotnet SDK. I believe the JS SDK is the only one which has it and the reason was to later define this via the server and push it to the client.

untitaker:
But an empty dsn disables the dotnet sdk, right?

bruno-garcia:
Dsn is a type. One needs to call the constructor and an empty string there is invalid. On integrations that bind config files (logging, aspnet core), empty string means disabled SDK.

untitaker:
Ok. We should align this f2f sometime, because I don't see a reason why it has to work differently in each SDK

kbremner:
Happy to update this PR, or close it.

Just for context, we were looking for ways to disable the JavaScript SDK during local development and came across the `enabled` option in the typescript definition. This seemed cleaner than wrapping the init call in a conditional so we went with that.

Opened the PR as we couldn't find any mention of how to disable the SDK during local development.

untitaker:
I'm just going to close this PR since it would need a redo anyway and we need to align about how we actually want to expose this. I apprechiate the effort though!

PR #404, opened 2018-09-28
Fix JS example for setExtra

TimPetricola:
The signature being `public setExtra(key: string, extra: any): Scope`


untitaker:
Thanks!

PR #416, opened 2018-10-01
fix (sidebar): Add trailing slash to sidebar urls

cameronmcefee:
This fixes an issue where navigating to a page from search results does not expand the right sidebar item.

### DIFF
--- a/src/_js/lib/Search.js
+++ b/src/_js/lib/Search.js
@@ -4,7 +4,7 @@ import Lunr from './Lunr';
 const renderResult = function(data) {
   const relativePath = data.path.replace(/^\/|\/$/g, '');
 
-  const url = `${window.location.origin}/${relativePath}`;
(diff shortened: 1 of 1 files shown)


src/_js/lib/Search.js
```diff
@@ -4,7 +4,7 @@ import Lunr from './Lunr';
 const renderResult = function(data) {
   const relativePath = data.path.replace(/^\/|\/$/g, '');
 
-  const url = `${window.location.origin}/${relativePath}`;
+  const url = `${window.location.origin}/${relativePath}/`;
```
adhiraj: Why does this fix it?

cameronmcefee: The sidebar determines what to show by matching the urls of the links to the current page. All the sidebar links end with a trailing slash.

PR #435, opened 2018-10-04
Export SentryErrorHandler class

NikolaJankovic:
This solves the 'references to a non-exported class are not supported in decorators...' error when building angular application.


kamilogorek:
Thanks @NikolaJankovic!

PR #437, opened 2018-10-04
Indicate that "attach_stacktrace" is not supported in python

Bazze:
Noticed that the `attach_stacktrace` option is not available in the Python SDK. Not sure if this is a correct edit, but wanted to make you aware.


untitaker:
Thanks and sorry for the trouble!

PR #444, opened 2018-10-08
Fix event_level default for logging integration

ahivert:
In [python file](https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/logging.py#L16), default value for ``event_level`` is not ``None`` but ``ERROR``.


untitaker:
Thanks, I forgot to change this!

PR #447, opened 2018-10-09
docs: Explain getHubFromCarrier in express

kamilogorek:
ref: https://github.com/getsentry/sentry-javascript/pull/1637/
ref: https://github.com/getsentry/sentry-javascript/issues/1632

### DIFF
--- a/src/collections/_documentation/platforms/javascript/express.md
+++ b/src/collections/_documentation/platforms/javascript/express.md
@@ -4,20 +4,21 @@ sidebar_order: 1010
 ---
 
 <!-- WIZARD -->
+
 Our Express integration only requires _@sentry/node_ to be installed then you can use it like this:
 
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/platforms/javascript/express.md
```diff
@@ -55,4 +56,36 @@ app.use(Sentry.Handlers.errorHandler());
...
+    next();
+  } else {
+    res.end();
+  }
+});
+```
+
+All `configureScope` calls that were done directly on initialized `Sentry` object, will be carried to each request, so it's a good way to set some data globally.
+
 <!-- ENDWIZARD -->
```
HazAT: Not sure, I think we should move the wizard tag.

kamilogorek:
updated

PR #448, opened 2018-10-09
Anonymize GA IP addresses

cameronmcefee:


cameronmcefee:
It had gone stale, and you did something on the GA side as well. I'm fine merging this just to be thorough.

PR #451, opened 2018-10-09
Update React.MD instructions to read better and add an init example a…

ericnakagawa:
…nd instructions

### DIFF
--- a/src/collections/_documentation/platforms/javascript/react.md
+++ b/src/collections/_documentation/platforms/javascript/react.md
@@ -10,7 +10,10 @@ If you’re using React 16 or above, Error Boundaries are an important tool for 
 
 ```jsx
 import * as Sentry from '@sentry/browser';
-// Sentry.init has to be called in the somewhere before
+// Sentry.init has to be called before sending an error to Sentry.
+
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/platforms/javascript/react.md
```diff
@@ -10,7 +10,10 @@ If you’re using React 16 or above, Error Boundaries are an important tool for
 
 ```jsx
 import * as Sentry from '@sentry/browser';
-// Sentry.init has to be called in the somewhere before
+// Sentry.init has to be called before sending an error to Sentry.
+
+// Uncomment the line below and copy your DSN key found in Settings > Data > Client Keys (DSN).
+// Sentry.init({ dsn: 'PASTE_YOUR_DSN_HERE' });
```
markstory: Doesn't this imply that each class that you would use sentry in needs to set the DSN which isn't what we should be recommending.

HazAT:
@ericnakagawa Thanks for your contribution, I updated the react docs here to make it clearer
https://github.com/getsentry/sentry-docs/pull/472

PR #463, opened 2018-10-15
fix: Add dismiss feedback functionality

denamwangi:
This PR aims to:
1) hide footer on page after feedback given
2) add an `x' out of feedback bar option. (Dismisses for 30minutes)

To Do
- [x] remove border around the button

feedback_footer

### DIFF
--- a/src/_js/lib/Feedback.js
+++ b/src/_js/lib/Feedback.js
@@ -1,4 +1,3 @@
-var obj = document.getElementsByClassName('feedback-footer')[0];
 
 // Send an event to reload whenever a feedback button is clicked
 $(document).on('change', '[data-feedback-toggle]', function(event) {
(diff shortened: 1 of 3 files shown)


src/_js/lib/Feedback.js
```diff
@@ -6,9 +8,29 @@ $(document).on('change', '[data-feedback-toggle]', function(event) {
   window.ra.event('docs.feedback-sent', {
     useful: parseInt($selected.val(), 10)
   });
+  obj.style.display="none";
```
adhiraj: will this also dismiss it for other pages the user navigates to in the same session?

src/_js/lib/Feedback.js
```diff
@@ -6,9 +8,29 @@ $(document).on('change', '[data-feedback-toggle]', function(event) {
...
+	if(dismissFeedback === 'true'){
+		obj.style.display = "none";
+	 } else {
+		obj.style.display = "";
+		$('[data-feedback-toggle] label').removeClass('active');
+	 }
 });
+
+var dismissFeedback = function() {
+	window.localStorage.setItem('dismissFeedback', 'true')
```
adhiraj: let's do the time bounded dismissed we discussed

src/_js/lib/Feedback.js
```diff
@@ -6,9 +7,35 @@ $(document).on('change', '[data-feedback-toggle]', function(event) {
...
+	 }
 });
+
+var dismissFeedback = function() {
+	window.localStorage.setItem('dismissTimestamp', Date.now())
+};
+
+// dismiss for 30mins
+var pastDismissWindow = function(dismissTimestamp) {
+  return ((Date.now() - dismissTimestamp)/ (1000*60)) > 30;
```
denamwangi: Should this be longer? shorter?

src/_js/lib/Feedback.js
denamwangi: Yes - for a period of 30 minutes

src/_js/lib/Feedback.js
```diff
@@ -6,9 +7,35 @@ $(document).on('change', '[data-feedback-toggle]', function(event) {
...
+  $('.feedback-footer').addClass('d-none');
+  dismissFeedback()
 });
 
 // Reset the feedback widget for the new page
 $(document).on('page.didUpdate', function(event) {
-  $('[data-feedback-toggle] label').removeClass('active');
+	const dismissTimestamp = window.localStorage.getItem('dismissTimestamp');
+
+	if(dismissTimestamp && !pastDismissWindow(dismissTimestamp)){
```
adhiraj: there's something wrong with the indentation in this block

src/_layouts/default.html
```diff
@@ -41,7 +41,7 @@ <h1 class="mb-3">{{ page.title }}</h1>
           </section>
         </div>
 
-        <footer class="border-top bg-white px-3 py-2 flex-grow-0">
+        <footer class="border-top bg-white px-3 py-2 flex-grow-0 feedback-footer">
```
adhiraj: where is the part where we're hiding by default and showing it if conditions are satisfied? i think we talked about this?

cameronmcefee: FWIW, the argument I was trying to make when we talked about this in person was that this feedback widget isn't functional without JavaScript, so a noscript fallback is moot, and having it visible would technically make a button which clicking would have no effect.

denamwangi: Ooh gotcha! That makes sense- thanks for clarifying

cameronmcefee:
Sorry, one more convention I missed mentioning earlier. For classes that are used for javascript dom references, I prefer to use `js-*` so in this case `js-feedback-footer` to signify the function of the class as JavaScript. The primary benefit is that it untangles style from function and prevents scenarios where the class gets deleted because it looks like it's just an empty class without a style.

PR #492, opened 2018-10-23
fix(tooling): Make prettier not autoformat markdown

lobsterkatie:
So that it stops breaking internal links and changing the bullets in lists from `-` to `*`.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,8 @@
 {
   "editor.trimAutoWhitespace": false,
   "files.trimTrailingWhitespace": false,
+  "prettier.disableLanguages": [
+    "vue", //default
+    "markdown"
+  ]
 }


.vscode/settings.json
```diff
@@ -1,4 +1,8 @@
 {
   "editor.trimAutoWhitespace": false,
   "files.trimTrailingWhitespace": false,
-}
\ No newline at end of file
+  "prettier.disableLanguages": [
+    "vue", //default
```
cameronmcefee: Is vue necessary? It isn't used, unless it's being used as a hack of some sort.

lobsterkatie: I just kept it because it's the default setting and didn't want to do anything other than add markdown. But I can remove it if you'd rather I do so.

cameronmcefee: Might as well, there risk of Vue is low

lobsterkatie: Done. Will merge once the lights turn green.

PR #497, opened 2018-10-24
fix: Document transaction style

untitaker:


src/collections/_documentation/platforms/python/django.md
```diff
@@ -39,6 +39,29 @@ This causes the following this to happen:
...
 
+## Options
+
+The following keyword arguments can be passed to `DjangoIntegration()`:
+
+* `transaction_style`:
+
+  ```python
+  @app.route("/myurl/<foo>")
+  def my_function(request):
```
bittner: This looks like [Flask code](flask.pocoo.org/docs/latest/quickstart/). Are you sure this is meant for the Django-related documentation?

If it is the code sample probably needs some explanation. Explain why the annotation is needed and what it does.

untitaker: The decorator is a copypaste error, yes. Thanks for bringing this to our attention.

untitaker: https://github.com/getsentry/sentry-docs/pull/623

PR #501, opened 2018-10-24
Update to whitelist + blacklist docs

tstirrat15:
## Motivation
The signature for those options is an array of either strings or regexes, while the documentation made it sound like it was a single pattern or string.

It may make sense generally for the docs to indicate the types accepted by each option.

## Changes
Clarify the language.


untitaker:
Thanks!

PR #506, opened 2018-10-25
fix: Add note about init order for RQ

untitaker:
Fix https://github.com/getsentry/sentry-python/issues/151


untitaker:
This is basically admitting defeat: The code for job functions is loaded so late that the SDK is not able to patch everything at that point.

PR #508, opened 2018-10-25
fix: update environments

maheskett:
cc: @DeadSuperHero


DeadSuperHero:
Yay! Thanks, @maheskett

PR #511, opened 2018-10-26
fix: document request data

untitaker:


src/_includes/platforms/python/request-data.md
```diff
@@ -0,0 +1 @@
+Request data is attached to all events: HTTP method, URL, headers, formdata, JSON payloads. Raw bodies and multipart file uploads are excluded and marked as excluded in the event. Personally identifiable information (such as user ids, usernames, cookies, authorization headers, ip addresses) is excluded unless ``send_default_pii`` is set to ``True``.
```
markstory:
```suggestion
Request data is attached to all events: HTTP method, URL, headers, formdata, JSON payloads. Raw bodies and multipart file uploads are excluded and marked as excluded from events. Personally identifiable information (such as user ids, usernames, cookies, authorization headers, ip addresses) are excluded unless ``send_default_pii`` is set to ``True``.
```

untitaker: So what I was trying to write here is that the event payload will still contain the filename of the uploaded file + a remark that the file content has been omitted from the event intentionally.

untitaker: I think it's fine now

PR #514, opened 2018-10-29
feat(minidump): Document event attachments for minidump requests

jan-auer:
screenshot 2018-10-29 at 22 46 12


dcramer:
Gave feedback on Slack, but documenting an example curl request would be super helpful to avoid "how do I name things?" and probably some other very basic affirmation style questions.

PR #523, opened 2018-11-01
(ref) Migrate browser compatibility chart to new SDK docs. Fixes SE-127

ceorourke:
If anyone has strong opinions about this being somewhere else lmk, this was the best page I could come up with.


mattrobenolt:
Add `saucelabs.com` after `img-src` in here: https://github.com/getsentry/sentry-docs/blob/master/nginx.conf#L16

PR #531, opened 2018-11-03
Fixes PHP send_callback sample code.

asantaniello:


untitaker:
I'm sorry, which issue were you seeing with the current example?

untitaker:
Makes sense, thanks! I think you can push to the same branch for this, no need to open another PR

untitaker:
Thanks again!

PR #539, opened 2018-11-06
(ref) Add table of contents to source map docs page

ceorourke:
screenshot 2018-11-06 15 08 46

### DIFF
--- a/src/collections/_documentation/platforms/javascript/sourcemaps.md
+++ b/src/collections/_documentation/platforms/javascript/sourcemaps.md
@@ -5,11 +5,16 @@ sidebar_order: 60
 
 Sentry supports un-minifying JavaScript via [Source Maps](http://blog.sentry.io/2015/10/29/debuggable-javascript-with-source-maps.html). This lets you view source code context obtained from stack traces in their original untransformed form, which is particularly useful for debugging minified code (e.g. UglifyJS), or transpiled code from a higher-level language (e.g. TypeScript, ES6).
 
-## Specify the release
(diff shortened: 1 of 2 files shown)


src/collections/_documentation/platforms/javascript/sourcemaps.md
```diff
@@ -5,15 +5,26 @@ sidebar_order: 60
...
 
-## Generating a Source Map
+If you are uploading source map artifacts yourself, you must specify the release in your SDK. Please refer to the [releases]({%- link _documentation/learn/releases.md -%}) page for more information on how to configure this.  Sentry will use the release name to associate digested event data with the files you’ve uploaded via the [releases API]({%- link _documentation/api/releases/index.md -%}), [sentry-cli]({%- link _documentation/learn/cli/index.md -%}) or [sentry-webpack-plugin](https://github.com/getsentry/sentry-webpack-plugin). This step is optional if you are hosting source maps on the remote server.
+
+## Generating a Source Map {#source-map-generation}
 
 Most modern JavaScript transpilers support source maps. Below are instructions for some common tools.
 
-### UglifyJS
+### UglifyJS {#source-map-uglifyjs}
```
cameronmcefee: Kramdown automatically adds ids to headings by slugifying the heading text (`#uglifyjs` in this case). Can we use those instead of adding the extra markup for an explicit id?

cameronmcefee:
The `jekyll-toc` plugin can do this automatically so we don't have to maintain this manually. Want me to drop that in here?

PR #545, opened 2018-11-09
Gitlab docs2

MimiDumpling:
this is the 2nd version of pull request #544 where we're merging into master


src/collections/_documentation/workflow/integrations/gitlab.md
```diff
@@ -0,0 +1,50 @@
...
+
+    b. Find the GitLab Group Path in your group’s GitLab page.
+
+      {% asset gitlab/gitlab-groups.png alt="GitLab group's page showing group path." %}
+
+    c. Find the GitLab Application ID and Secret in the Sentry app within GitLab.
+
+      {% asset gitlab/gitlab-app-id.png alt="GitLab application id and secret in Sentry app within GitLab." %}
+
+    d.  Use this information to fill out the GitLab Configuration and click ‘Submit.’gitlab
```
markstory: Does our markdown parser support letter based ordered lists? Other markdown parses I have used only support number based syntax and letters were added via css.

cameronmcefee:
Since @MimiDumpling is out today, I've made some changes myself:

- Markdown does not support lettered lists. Good catch @markstory. This explained why the indentation on images was behaving inconsistently, which was driving me a bit nuts. I converted the sublist to numbers as well, which I think is fine.

- Because markdown generates ordered lists which are automatically numbered, it's possible to prefix each list item with `1.` rather than explicitly numbering them. While this can potentially be confusing in  plain text, it makes reordering and inserting items much easier, as you don't have to renumber each item if you make a change.

PR #555, opened 2018-11-12
add browsernpm to support list for url lists

cameronmcefee:
@mitsuhiko This seems like it resolves DOCS-16. Or is it more complicated than I understand?


cameronmcefee:
If we need a more automated approach. I noticed in platforms.yml `browsernpm` falls back to `javascript` rather than `browser`. I'd need to take some time to figure out what the fallbacks are actually used for and how they work, but I would think that should be pointed to `browser`.

PR #556, opened 2018-11-12
feat: Embed fonts

cameronmcefee:
This puts the fonts into the repo rather than loading them from google.

@mattrobenolt Can you doublecheck my CSP edit. I just removed `style-src 'self' 'unsafe-inline' fonts.googleapis.com; font-src 'self' fonts.gstatic.com;`


mattrobenolt:
The CSP changes are wrong. You can't remove the directives entirely.

`style-src 'self' 'unsafe-inline'; font-src 'self';`

They're still needed, and `'self'` is required. Style also needs `unsafe-inline` yet. You just needed to remove the hostnames.

cameronmcefee:
A cool, I've put them back

PR #557, opened 2018-11-12
Move SDK filtering/scrubbing to top

ndmanvar:


ndmanvar:
Moving SDK/client-side scrubbing/filtering to top as that is the preferred/best place to scrub (i.e. so that it isn't sent over the wire)

PR #570, opened 2018-11-14
Add missing Sanic Import to Documentation

harshanarayana:
In the current helper document for `sanic`, the example code provided is missing an import for the `Sanic` class.  

This commit fixes that inconsistency.


markstory:
Thank you 👍

PR #571, opened 2018-11-14
GPU Interface

bruno-garcia:
@mitsuhiko Should this be a top level inf or under contexts?

### DIFF
--- a/src/collections/_documentation/development/sdk-dev/interfaces/gpu.md
+++ b/src/collections/_documentation/development/sdk-dev/interfaces/gpu.md
@@ -1,0 +1,59 @@
+---
+title: 'GPU Interface'
+sidebar_order: 13
+---
+
+An interface which describes the main GPU of the device.
+
+`name`
+
+: **Required**. The name of the graphics device.
+
+`version`
+
+: _Optional_. The Version of graphics device.
+
+`id`
+
+: _Optional_. The PCI Id of the graphics device.
+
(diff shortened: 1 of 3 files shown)


src/collections/_documentation/development/sdk-dev/interfaces/gpu.md
```diff
@@ -0,0 +1,59 @@
+---
+title: 'GPU Interface'
+sidebar_order: 13
+---
+
+An interface which describes the main GPU of the device.
+
+`name`
```
cameronmcefee: @jessla brought up that it'd be ideal if we could make these headings so they're linkable. Any reason to not do that?

bruno-garcia: @cameronmcefee I think it would be great if we could link to individual headings.
Something to be done in a separate PR I assume though?

cameronmcefee: I think it'd be fine to do it here. All it takes is switching these to headings. If you want to retain the code highlight, you could do 

```
### `name`
```

bruno-garcia: Oh, I thought you were referring to adding to all the `interfaces` docs since they are also like this.
I'll add the headings here as you suggested.

PR #573, opened 2018-11-15
fix(nginx): Handle trailing slashes (or lack thereof)

lobsterkatie:
A first pass at this, just to get it done and stop all our stuff from 404ing as quickly as possible. Will come back to the problem of cleaning up the file (nesting locations, etc) separately.

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -19,20 +19,20 @@ server {
   
   rewrite ^/clientdev/(.*)$ /development/sdk-dev/$1$is_args$args redirect;
   rewrite ^/integrations/(.*)$ /workflow/integrations/$1$is_args$args redirect;
-  rewrite ^/internal/(api|docs|contributing|environment)/$ /development/contribute/$1$is_args$args redirect;
+  rewrite ^/internal/(api|docs|contributing|environment)/$ /development/contribute/$1/$is_args$args redirect;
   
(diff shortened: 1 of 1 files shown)


nginx.conf
```diff
@@ -19,20 +19,20 @@ server {
   
   rewrite ^/clientdev/(.*)$ /development/sdk-dev/$1$is_args$args redirect;
```
lobsterkatie: For ones where the pattern ends in `(.*)`, my understanding is I didn't need to add the trailing slash in the destination's `$1$is_args$args` because either

a) there's nothing in the `(.*)` and so the slash before it becomes the trailing slash, or
b) whatever is in the `(.*)` already has a trailing slash, in which case we don't want to add another one, or
c) whatever is in the `(.*)` doesn't already has a trailing slash, in which case one will get added by the `/` prefix rule

lobsterkatie:
cc @maheskett 

This should obviate the direct need for https://github.com/getsentry/getsentry/pull/2355, though it's still not a bad idea.

nginx.conf
mattrobenolt: You’re right. But you can do `(.*?)/?$` then include the slash in the redirect. This would prevent it from being captured.

lobsterkatie:
Ah, that makes sense, thanks, @mattrobenolt ~!

PR #578, opened 2018-11-16
JavaScript: Made the example release version consistent with the suggested format in the releases docs

MartijnHols:
https://docs.sentry.io/workflow/releases/?platform=javascript: `Note that releases are global per organization so make sure to prefix them with something project specific if needed:`

### DIFF
--- a/src/collections/_documentation/clients/javascript/config.md
+++ b/src/collections/_documentation/clients/javascript/config.md
@@ -17,7 +17,7 @@ At this point, Raven is ready to capture any uncaught exception.
 
 ```javascript
 Raven.config('___PUBLIC_DSN___', {
-    release: '1.3.0'
+    release: 'myapp@1.3.0'
 }).install()
 ```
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/clients/javascript/config.md
```diff
@@ -43,7 +43,7 @@ Those configuration options are documented below:
   }
   ```
 
-  Can also be defined with `Raven.setRelease('721e41770371db95eee98ca2707686226b993eda')`.
+  Can also be defined with `Raven.setRelease('721e41770371db95eee98ca2707686226b993eda')`. See the [releases](https://docs.sentry.io/workflow/releases/?platform=javascript) page for more information about releases.
```
untitaker: Can you change this to use a less hardcoded link? You should find examples for this with `git grep '{% link'`

untitaker:
Thanks!

PR #591, opened 2018-11-22
Fixed user-feedback-example beforeSend

9renpoto:
Fixed beforeSend examples .
Please review it

refs: https://github.com/getsentry/sentry-javascript/blob/master/packages/core/src/interfaces.ts#L104-L116

### DIFF
--- a/src/collections/_documentation/enriching-error-data/user-feedback-example/browser.md
+++ b/src/collections/_documentation/enriching-error-data/user-feedback-example/browser.md
@@ -4,7 +4,7 @@
 <script>
   Sentry.init({
     dsn: '___PUBLIC_DSN___',
-    beforeSend(event) {
+    beforeSend: (event) => {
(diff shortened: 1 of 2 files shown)


HazAT:
Hmm, this shouldn't be necessary since we do not need a "this" binding in this function but it's also not wrong. So both cases should and will work, not sure if we should change this.

9renpoto:
Thank you for review!

`Sentry.Options interface` requires beforeSend as a `Function`, but this examples is an error.

screen shot 2018-11-27 at 20 27 47


In order to make it work you need to either:

```diff
-beforeSend(event) {
+beforeSend: (event) => {
```

or 

```diff
-beforeSend(event) {
+beforeSend: function (event) {
```

9renpoto:
Oops, I am sorry, it was a misunderstanding.
Was self resolved. Thank you review !

PR #593, opened 2018-11-26
fix: Add --trace to jekyll watch

HazAT:


src/_plugins/platform_api.rb
```diff
@@ -126,7 +126,11 @@ def initialize(site, base, dir, payload, name)
         end
 
         noko_doc.css("a").each do |node|
-          node["href"] = URI(node["href"]).host ? node["href"] : "#{site.config["url"]}#{node["href"]}"
+          begin
+            node["href"] = URI(node["href"]).host ? node["href"] : "#{site.config["url"]}#{node["href"]}"
```
cameronmcefee: Did you investigate why this was failing? This doesn't seem a like a very helpful solution since this doesn't handle the case and the error output doesn't explain what failed or why.

HazAT: I probably should do that 😅 
I wanted to quickly fix this to make the docs work, checking now.

HazAT: OK, I can't reproduce this anymore, I think it was because my nokogiri gem was broken.
nokogiri things 🤷‍♂️ 
But I've added the error to the message so in case someone runs into this ever again it doesn't break and he has more context.

cameronmcefee: If this is blocking something it's ok to ship, but I think we should still force this to throw, at least in production. If it's erroring, it's putting faulty links into the platform API stuff which is pretty invisible to us.

HazAT: You are right, like I said I can't repro this now.
Removed it altogether, but the other stuff in this PR `--trace` should be helpful right?

cameronmcefee: Yeah, it ought to be of some use.

PR #608, opened 2018-11-30
Raven celery logging

nitely:
Fixes https://github.com/getsentry/raven-python/issues/1004

Celery must be set explicitly in order to log unhandled exceptions.

### DIFF
--- a/src/collections/_documentation/clients/python/integrations/django.md
+++ b/src/collections/_documentation/clients/python/integrations/django.md
@@ -114,7 +114,7 @@ LOGGING = {
         }
     },
     'loggers': {
-        'celery': {  # Celery must be set explicitly
+        'celery': {  # Log unhandled exceptions from tasks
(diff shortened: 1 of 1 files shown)


untitaker:
The integration for celery is documented here: https://docs.sentry.io/clients/python/integrations/celery/

I don't see how you need this or why this is on the Django page.

nitely:
maybe you should setup a django project and try the integration. You will notice unhandled exceptions in tasks won't get logged. Not even if you add the celery integration code.

Whatever, close it.

untitaker:
Passive agressive comments like this don't help anybody. sentry.io itself
used Raven exactly like this until we switched to sentry-sdk.

untitaker:
One thing you *maybe* need to ensure is that your SDK is initialized early enough, not when the task's module is imported (which Celery *might* do lazily). I don't know Raven well enough to know.

PR #612, opened 2018-12-03
Fix Atexit import name

spascou:
The import name for the Atexit default integration was not the same as in the module source.


markstory:
Thank you :+1:

PR #620, opened 2018-12-05
(chore) Update membership table

ceorourke:
screenshot 2018-12-04 17 06 03


maheskett:
Could you add the update for add/remove projects as well?

PR #623, opened 2018-12-05
fix: Remove copypaste error in Django docs

untitaker:


bittner:
Alright, you removed the Flask-style annotation. But it's still unclear what this section is actually explaining.

- `transaction_style` is explained to be a keyword argument. So, why is there no example of how this is used? (This is not obvious!)
- What is the relation of the code sample and `DjangoIntegration()`?

bittner:
Maybe, if you move the paragraph below the code sample up, so that it's read before it, things would be clearer.

untitaker:
`transaction_style` is explained to be a keyword argument *to DjangoIntegration*. Meaning that you can instantiate `DjangoIntegration()` like this: `DjangoIntegration(transaction_style=...)`.

I'm open to suggestions on how the docs could be clearer but I don't think that adding an example to every option is a scalable solution.

bittner:
Note that the reputation of Sentry will suffer when the documentation is bad. This is not about "scalable". A bad reputation will easily scale across the whole planet.

I gave [my suggestion](https://github.com/getsentry/sentry-docs/pull/623#issuecomment-444587274) above. Thanks for listening!

untitaker:
@bittner My claim is that adding an example for each option will harm readability in the same way that verbose wording does. I don't understand what's unclear about this sentence:

>The following keyword arguments can be passed to `DjangoIntegration()`:

It says just what I described above: You take `transaction_style` and pass it as a keyword argument to `DjangoIntegration()`. Resulting in `DjangoIntegration(transaction_style=..)`. Where the result of `DjangoIntegration(..)` should be plugged into is explained at the top of the page.

PR #628, opened 2018-12-08
Update github enterprise docs

maheskett:


MimiDumpling:
I went through the configuration and it looks good to me! 👍

PR #634, opened 2018-12-11
feat(sidebar): Allow doc title in sidebar to differ from doc title on page

lobsterkatie:
As it stands, the `title` attribute of each page is used for two purposes: to create the link to the page n the sidebar, and to set the heading at the top of the page. Given the spatial constraints of the sidebar, this artificially limits the possible length of page headings.

This PR splits those two concerns, creating a new `sidebar_title` attribute which is used to create the sidebar link (if present). If it's not present (as will be the case for most pages), the `title` attribute is used, just as before.

Here's an example:
image
image
[NB: I get that the page title still should be short enough that it doesn't wrap - fixing that in a different branch]

I think the code change here is sufficient to make this work, but please let me know if there's something I've overlooked.


src/_includes/sidebar/default.html
```diff
@@ -1,3 +1,7 @@
 <a class="d-block{% if include.document.path == page.path %} active{% endif %}" href="{{ include.document.url }}" data-sidebar-link>
-  {{ include.document.title }}
+  {% if include.document.sidebar_title %}
```
cameronmcefee: This could be expressed more simply as:
```
{{ include.document.sidebar_title | default: include.document.title }}
```

lobsterkatie: Ah... good to know! Thanks! Will make the change.

PR #635, opened 2018-12-11
feat(integrations): Azure DevOps troubleshooting section.

lauryndbrown:


src/collections/_documentation/workflow/integrations/azure-devops.md
```diff
@@ -87,3 +87,17 @@ Fixes MYAPP-317
...
+#### No accounts during installation
+
+> [{% asset azure-add-account-no-accounts.png %}]({% asset azure-add-account-no-accounts.png @path %})  
+
+If you reach the account's page during Azure DevOps' installation process and there are no accounts to select please check the following possibilities:
+
+- Ensure you are logged into the account connected to your Azure DevOps organization.
+- Double check that your account is a Microsoft Account (MSA). At this time, we do not support Azure Active Directory (AAD) accounts. 
+
+#### Resolve via PR not working
```
saragilford: @lauryndbrown we don't need to explicitly say that Resolve via PR doesn't work -- we should just not mention it as a possibility. Users who use Azure Devops will likely not also be using Github/Gitlab etc, so its unlikely that they will be used tot doing one thing in GH that they can't do in AD.

src/collections/_documentation/workflow/integrations/azure-devops.md
```diff
@@ -87,3 +87,17 @@ Fixes MYAPP-317
...
 
 When Sentry sees this, we’ll automatically annotate the matching issue with a reference to the commit, and, later, when that commit is part of a release, we’ll mark the issue as resolved.
+
+## Troubleshooting
+
+#### No accounts during installation
+
+> [{% asset azure-add-account-no-accounts.png %}]({% asset azure-add-account-no-accounts.png @path %})  
+
+If you reach the account's page during Azure DevOps' installation process and there are no accounts to select please check the following possibilities:
```
saragilford: Can we change this to:

If you reach the account selection page during the Azure Devops installation process (step 4 in the instructions above) and see that there are no accounts available to select, please check the following possibilities:

lauryndbrown:
@saragilford let me know if this looks good.

PR #657, opened 2018-12-18
doc: Explain how to log 404s in django

untitaker:
lmk if you need this for other frameworks.

### DIFF
--- a/src/collections/_documentation/platforms/python/django.md
+++ b/src/collections/_documentation/platforms/python/django.md
@@ -66,3 +66,32 @@ The following keyword arguments can be passed to `DjangoIntegration()`:
 
 The user feedback feature can be used with this integration.  For more information
 see [User Feedback]({% link _documentation/enriching-error-data/user-feedback.md %}?platform=django).
+
+## Reporting other status codes
+
+In some situations it might make sense to report 404s and other errors besides 500 to Sentry. This can be achieved by writing your own Django view for those status codes. For example:
+
+```python
+
+# urls.py
+
+handler404 = 'mysite.views.my_custom_page_not_found_view'
+
+# views.py
+
+from django.http import HttpResponseNotFound
+from sentry_sdk import capture_message
+
+
+def my_custom_page_not_found_view(*args, **kwargs):
+    capture_message("Page not found!", level="error")
+
+    # return any response here
+    # return HttpResponseNotFound("Not found")
+```
+
+The error message you send to sentry will have the usual request data attached.
+
+Refer to [Customizing Error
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/platforms/python/django.md
```diff
@@ -66,3 +66,32 @@ The following keyword arguments can be passed to `DjangoIntegration()`:
 
 The user feedback feature can be used with this integration.  For more information
 see [User Feedback]({% link _documentation/enriching-error-data/user-feedback.md %}?platform=django).
+
+## Reporting other status codes
+
+In some situations it might make sense to report 404s and other errors besides 500 to Sentry. This can be achieved by writing your own Django view for those status codes. For example:
```
lobsterkatie: > other errors besides 500

Are 500 errors literally all we capture, or by default do we capture all 500-level errors (501, 502, etc)?

untitaker: Only 500. We don't have some code like `if status_code == 500` in there, but rely on Django to tell us what is an unhandled exception.

src/collections/_documentation/platforms/python/django.md
```diff
@@ -66,3 +66,32 @@ The following keyword arguments can be passed to `DjangoIntegration()`:
...
+# views.py
+
+from django.http import HttpResponseNotFound
+from sentry_sdk import capture_message
+
+
+def my_custom_page_not_found_view(*args, **kwargs):
+    capture_message("Page not found!", level="error")
+
+    # return any response here
```
lobsterkatie: How would you feel about this:

```
# return any response here, for ex:
return HttpResponseNotFound("Not found")
```
for the last two lines?

untitaker: can do

untitaker: done

src/collections/_documentation/platforms/python/django.md
untitaker: your call on this one, no opinion

untitaker:
@lobsterkatie pls review again

src/collections/_documentation/platforms/python/django.md
lobsterkatie: Yeah, I like "besides general unhandled exceptions" better - it's closer to the full story without getting too deep into the weeds.

lobsterkatie:
With that small change to the 500/unhandled errors language, I think it's good to go.

(Will be traveling part of tomorrow (before working the rest of the day remotely), so if I don't see the update right away, feel free to ping me in slack.)

PR #688, opened 2019-01-03
fix: Use loader only in prod

HazAT:
I was cleaning up issues in our sentry project and noticed a few issues that were emitted during local development, I think it doesn't make sense to send them to Sentry.

### DIFF
--- a/src/_includes/head.html
+++ b/src/_includes/head.html
@@ -34,7 +34,9 @@ the following value in the page frontmatter
 <link href="{{ assets['favicon.ico'].digest_path }}" rel="icon" type="image/png">
 <link rel="canonical" href="{{ page.url | site.url | replace:'index.html',''}}">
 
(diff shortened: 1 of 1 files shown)


src/_includes/head.html
```diff
@@ -34,7 +34,9 @@
 <link href="{{ assets['favicon.ico'].digest_path }}" rel="icon" type="image/png">
 <link rel="canonical" href="{{ page.url | site.url | replace:'index.html',''}}">
 
+{% if jekyll.environment != "development" %}
```
cameronmcefee: Let's use `{% if jekyll.environment == "production" %}` instead, otherwise you'll get exceptions from test runs.

cameronmcefee: Nevermind, I just looked through the code and we're only using `production`/`development`.

PR #697, opened 2019-01-07
Force bundler update

cameronmcefee:
Travis updated to Bundler 2.0 which has caused some recent builds to break on other Jekyll sites. This preemptively upgrades things so we don't get unexpected breakages the next time the gem cache updates.

https://docs.travis-ci.com/user/languages/ruby/#bundler-20


joshuarli:
Closing; not applicable anymore.

PR #715, opened 2019-01-14
make fingerprint example more clear

crhistianramirez:
It doesn't really make sense to use `configureScope` alongside fingerprints. The whole point of fingerprints is to be able to group one event differently than another and anything set in `configureScope` would be global to all events.

It also seems weird that the string example is `'my-view-function'`. This makes it seem like you need to pass the name of a function to be executed or something.

### DIFF
--- a/src/collections/_documentation/data-management/set-fingerprint/javascript.md
+++ b/src/collections/_documentation/data-management/set-fingerprint/javascript.md
@@ -1,5 +1,5 @@
 ```javascript
-Sentry.configureScope((scope) => {
-  scope.setFingerprint(['my-view-function']);
+Sentry.withScope((scope) => {
+  scope.setFingerprint(['group1']);
 });
 ```


untitaker:
`withScope` pushes a new scope to the "thread-local stack" (in Python, not sure how the equivalent terminology in JS looks like), runs a function inside of that scope, then **pops** the scope again. Your example will push a scope, add a fingerprint to that scope, pop the scope and with that the fingerprint will be gone again. The fingerprint will apply to nothing.

`configureScope` configures the *current* scope. Integrations usually wrap a request handler or background task for some job queue in a `withScope`, so the fingerprint will only apply to the current request.

I think the problem with this example is that fingerprints are most useful in `beforeSend`, where you can choose to set a fingerprint depending on other criteria (such as the error type), not just a "region" of code.

untitaker:
Since this came up twice now, let's write an example for exception-based fingerprints (python):

```python

def before_send(event, hint):
    if "exc_info" not in hint:
        return event
    if isinstance(hint['exc_info'][1], ZeroDivisionError):
        event['fingerprint'] = ...
    return event

init(..., before_send=before_send)
```

untitaker:
Closing for https://github.com/getsentry/sentry-docs/pull/720

PR #731, opened 2019-01-19
Updated Python docs spelling & grammar

MimiDumpling:
I've updated spelling and grammar/language in the Python docs. This will make communication more clear for the reader. :) @untitaker, will you take a look?


MimiDumpling:
@untitaker Yeah, let's hold off until tomorrow. I was going to go back through and take another look at the sentences where I changed passive to active voice. Going to be more specific about which software component takes the role of subject in a sentence.

PR #735, opened 2019-01-22
fix: Doubled config keys

untitaker:
screenshot 2019-01-22 at 14 49 10


cameronmcefee:
Ah, I think I broke this when I wrapped the titles in `code` tags

PR #759, opened 2019-02-05
[WIP] ref: Remove note of R8 unsupported

bruno-garcia:
WIP: To be merged after:

https://github.com/getsentry/symbolic/pull/125
And PRs to sentry and sentry-cli with new symbolic and gradle plugin.


bruno-garcia:
@bretthoerner I gave it a thought and decided not to at this point for a couple reasons:
R8 is supposed to be compatible with proguard. So supporting latest proguard should suffice. Users are not required to change their proguard rules files or anything else.
Also, If this no longer becomes truth, it possibly means R8 is ahead and we'd need to chance it and possibly split the docs/support code with an `R8` symbolic crate or something. 
For these reasons for the time being I just thought leaving it out would be the way to go.

But otherwise happy to add note on R8 being supported if you think it's better.

bretthoerner:
Yeah, that's fair.

PR #761, opened 2019-02-06
Replace request.ip with request.remote_ip

schm:
`request.ip` provides the IP address of the server/client, which requested the page. `request.remote_ip` in contrast will ignore known proxies and instead return for real clients IP address. So if the app server is deployed behind a proxy, the value in `request.remote_ip` will be more interesting. Since `request.remote_ip` will fall back to `request.ip`, this should not cause undesired side effects.

See Rails' documentation for more details: https://api.rubyonrails.org/classes/ActionDispatch/RemoteIp.html

### DIFF
--- a/src/collections/_documentation/clients/ruby/context.md
+++ b/src/collections/_documentation/clients/ruby/context.md
@@ -53,7 +53,7 @@ Raven.user_context(
   username: current_user.username, # "foo"
 
   # the actor's IP address, if available
-  ip_address: request.ip # '127.0.0.1'
+  ip_address: request.remote_ip # '127.0.0.1'
 )
 ```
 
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/clients/ruby/context.md
```diff
@@ -53,7 +53,7 @@ Raven.user_context(
   username: current_user.username, # "foo"
 
   # the actor's IP address, if available
-  ip_address: request.ip # '127.0.0.1'
+  ip_address: request.remote_ip # '127.0.0.1'
```
markstory: While rails does offer `request.remote_ip` the standard Rack interfaces do not. Adding this would result in broken examples for all non-rails use cases.

schm: D'uh. You're absolutely correct. I didn't realize, that this piece of documentation was targeted at all ruby users.

PR #770, opened 2019-02-12
fix: Remove back button from docs

maheskett:
We're removing the back button from the product, so updating the require 2fa docs to reflect that. Also, adding another image.

screen shot 2019-02-12 at 1 49 16 pm


maheskett:
Lol. Thanks!

PR #777, opened 2019-02-16
Update docs to reflect setUserContext instead of scope

ballenwillis:
I only updated the JavaScript version because it's the only one I could do confidently. I recommend someone with a better understanding of Sentry update the others in this directory as it will be significantly faster and more accurate.


HazAT:
This is using old raven-js SDK, the new syntax is with `configureScope`.

PR #784, opened 2019-02-19
Fixes to get both source map and bundle

prsauer:
To get both the source map and bundle uploaded I had to have them in the same folder for sentry-cli. To do this I ended up copying the bundle up a folder since if you create the source map in the same folder as the rest of the build the whole blob gets uploaded to CodePush and I didn't want source maps in my releases there.


PeloWriter:
Closing this PR as it's quite dated.

PR #788, opened 2019-02-21
Correct ThreadingIntegration import name

cakemanny:
Correcting a copy-paste error in the python default integrations docs


untitaker:
Thx. I can't count how many times this happend 😆

PR #793, opened 2019-02-25
Symfony v3

Jean85:
This is the first draft to document the new version of the Symfony bundle. I've marked this as under development, and it's not showing, but I'm not sure if it's the right thing to do...


HazAT:
Can we close this PR in favor of: https://github.com/getsentry/sentry-docs/pull/810 ?

Jean85:
Ok!

PR #796, opened 2019-02-26
Change $http_host to $host in example config

ogarrett:
See comment about $http_host and $host in http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_set_header


untitaker:
@ogarrett did you encounter an issue with the current config? I wonder if this makes a difference in practice

untitaker:
Closing due to lack of response. I think this change is unnecessary.

PR #798, opened 2019-02-26
Update local dev docs to reflect new reality

mattrobenolt:


src/collections/_documentation/development/contribute/environment.md
```diff
@@ -54,27 +54,25 @@ Run the following to install the Python and JavaScript libraries and database se
...
 ```bash
-psql template1 $(whoami) -c 'CREATE USER postgres SUPERUSER;'
+sentry init --dev
 ```
 
-## Running the Development Server
+This file ends up in `~/.sentry/sentry.conf.py`
 
-Before you are able to run the development server, you first must create a proper database for it to use. Running the following will create the proper database and fill it with example data:
+Next we need to start up our development services. This includes running services like Postgres, Redis, etc. `sentry devservices` depends on Docker, which should have gotten installed during `make develop`. If not, you can install with `brew cask install docker`. After installing Docker, make sure it is running by just launching `/Applications/Docker.app`.
```
MimiDumpling:
```suggestion
Next, we need to start up our development services. This includes running services like Postgres, Redis, etc. `sentry devservices` depends on Docker, which should have gotten installed during `make develop`. If not, you can install with `brew cask install docker`. After installing Docker, make sure it is running by just launching `/Applications/Docker.app`.
```

PR #814, opened 2019-03-05
Encourage safe usage of Sentry

lightswitch05:
I have come across several websites the immediately fail due to blocking Sentry. Its especially bad since it is often within the first few lines of their javascript. Encourage safe checking of the `Sentry` object before using it.

Here are a couple documented websites that fail completely if Sentry does not load:

* https://github.com/lightswitch05/hosts/issues/40
* https://github.com/lightswitch05/hosts/issues/77

### DIFF
--- a/src/collections/_documentation/error-reporting/getting-started-config/browser.md
+++ b/src/collections/_documentation/error-reporting/getting-started-config/browser.md
@@ -1,5 +1,7 @@
 You should `init` the Sentry Browser SDK as soon as possible during your page load:
 
 ```javascript
-Sentry.init({ dsn: '___PUBLIC_DSN___' });
+if (Sentry) {
+    Sentry.init({ dsn: '___PUBLIC_DSN___' });
+}
 ```


untitaker:
My personal opinion on this is that if you're going to mess with the internals of a site by blocking a certain domain, removing certain HTML tags or preventing certain functions from running, you should also be prepared to implement workarounds to prevent subsequent code from breaking. However, I think it's not reasonable to ask authors of the scripts you're breaking to deploy safeguards against cases that don't occur in regular execution.

You could argue that `Sentry` might be `undefined` if cdn.ravenjs.com is down (which would imply that an entire CDN provider is down) or just unreachable due to other local network issues, but I am fairly certain that this PR or anything like it would have never been opened in a world without adblocking. Even for its original intent, this PR is insufficient. The `Sentry` object is used many more times in people's codebases, and wrapping each of these uses in that if-statement is not something we will encourage anybody to do. This entire charade is also completely unnecessary for users who vendor our SDK into their own JS bundle, where your hosts file will have no effect anyway.

TL;DR: Patching code at runtime to not do certain things is harder than going to the network tab and copypasting hostnames that look weird to you.

lightswitch05:
Wow, I didn't expect this response. I thought showing safe usage check in the docs would have been a simple change, after all the goal is to track and fix errors and its just a documentation change. `cdn.ravenjs.com` would certainly cause the same error behavior. Anyways, thank you for your time reviewing the pull request

untitaker:
So this is not a new issue: https://github.com/getsentry/sentry-javascript/issues/1421. In the linked issue you have a customer of Sentry saying that this way of using the SDK is inacceptable (wrapping everything with `if(Sentry)`) which implies to me that this has to be solved at a different level.

Another thread here: https://github.com/StevenBlack/hosts/issues/568

PR #818, opened 2019-03-05
A few tweaks to Search doc 

MimiDumpling:


MimiDumpling:
Hi @wedamija! Will you take a look? Thx!

src/collections/_documentation/workflow/search.md
```diff
@@ -6,33 +6,37 @@ sidebar_order: 3
...
   title="Note"
   content=__alert_content
+  level="info"
 %}
 
+&nbsp;
 ## Syntax
 
-Queries are constructed using a `token:value` pattern:
+Queries are constructed using a `token:value` pattern, with an optional raw search at the end:
```
wedamija: It could make sense to describe this as a `key:value` pattern, and then each `key:value` element is described as a `token`? Then raw search is also an optional token included at the end.

MimiDumpling: Nice! Yeah. I like that. :)

PR #821, opened 2019-03-07
Fix/mimi search

MimiDumpling:
Updating Search doc. Clarifying how Syntax works and elaborating on how Custom Tags work.


src/collections/_documentation/workflow/search.md
```diff
@@ -218,12 +217,22 @@ Below is a list of Event-level tokens reserved and known to Sentry:
 
 : Restrict results to events with a matching stack property.
 
-&nbsp;
-### Custom Tags
+### Custom Tags {#custom-tags}
```
cameronmcefee: Let's leave out these guys: {#custom-tags}

They're automatic and only need to be included if you want an id that is different from the title, which we should avoid. The existing ones in the codebase were just there to prevent broken links from the old docs, but given that page scroll location isn't *that* critical, it might be worth just removing them universally.

src/collections/_documentation/workflow/search.md
```diff
@@ -218,12 +217,22 @@ Below is a list of Event-level tokens reserved and known to Sentry:
...
+
+{% include components/platform_content.html content_dir='set-tag' %}
+
+Several common uses for tags include:
+
+-   The hostname of the server
+-   The version of your platform (for example, iOS 5.0)
+-   The user’s language
+
+For more information, see [full documentation on Tagging Events]({%- link _documentation/enriching-error-data/context.md -%} #tagging-events).
```
cameronmcefee:
```suggestion
For more information, see [full documentation on Tagging Events]({%- link _documentation/enriching-error-data/context.md -%}#tagging-events).
```

While not strictly necessary because the `-` in `-%}` trims the trailing whitespace, I'd prefer we don't include the space before the hash.

MimiDumpling:
@lobsterkatie @cameronmcefee updated changes suggested in comments. :)

PR #829, opened 2019-03-11
fix(CLI releases docs): Add additional required scope for `set-commits`

lobsterkatie:
Currently, the [CLI authentication docs](https://docs.sentry.io/cli/configuration/) only list two required scopes:

image

But as it turns out, you can't use `set-commits` unless you also have `org:read`scope:

- calling `set-commits` calls `list_org_repos`: https://github.com/getsentry/sentry-cli/blob/d12a258c681bb57f00769ea962b4355fb2979637/src/commands/releases.rs#L400

- `list_org_repos` makes a get requtest to the `/organizations/{org-slug}/repos/` endpoint: https://github.com/getsentry/sentry-cli/blob/d12a258c681bb57f00769ea962b4355fb2979637/src/api.rs#L1210

- that endpoint inherits from `OrgEndpoint`: https://github.com/getsentry/sentry/blob/f5998e667adb39907fb58db9fe66f1b2482db263/src/sentry/api/endpoints/organization_repositories.py#L16

- `OrgEndpoint` uses `OrgPermissions`: https://github.com/getsentry/sentry/blob/f5998e667adb39907fb58db9fe66f1b2482db263/src/sentry/api/bases/organization.py#L122

- `OrgPermissions` needs at least `org:read` for `GET`ing: https://github.com/getsentry/sentry/blob/f5998e667adb39907fb58db9fe66f1b2482db263/src/sentry/api/bases/organization.py#L31

This PR adds the missing scope and reorganizes the text a bit:

image

### DIFF
--- a/src/collections/_documentation/cli/configuration.md
+++ b/src/collections/_documentation/cli/configuration.md
@@ -3,30 +3,43 @@ title: 'Configuration and Authentication'
 sidebar_order: 1
 ---
 
-For most functionality you need to authenticate with Sentry. To sign in via the CLI tool, you can use the _login_ command which will guide you through it:
+For most functionality you need to authenticate with Sentry. Setting this up can be done either automatically, using `sentry-cli`, or manually. Either way, you'll need a token with at least the following scopes:
+
+- `project:read`
+- `project:releases`
+- `org:read`
+
+##### To use the automatic option:
 
 ```bash
 $ sentry-cli login
 ```
 
-If you want to manually authenticate `sentry-cli` you can go to your to your auth token settings in your user account (User Icon -> API) and generate a new token with at least the following scopes:
-
--   `project:read`
--   `project:releases`
-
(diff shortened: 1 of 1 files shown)


lobsterkatie:
I put it as an alert because it's a "you gotta do me before any of the rest of this will work" kind of thing. Didn't want it to be missed by someone skimming, and other permissions stuff is called out this way:

image

lobsterkatie:
Another option is to just add `org:read` to the initial list of scopes we require (even though it's not required for all CLI commands).

markstory:
Including `org:read` as a default permission is a good option. It sets people up for success later and saves them having to backtrack to update permissions later on.

lobsterkatie:
Okay - pulled it out of the releases page and added it into the default config. Since I was there anyway, also made the edits I was planning to stick into a different PR. Lemme know what you think.

lobsterkatie:
Great, thanks. Caught one typo (in two spots) - fixed and will merge once the tests pass.

PR #830, opened 2019-03-11
fix: Change local hosting port to not conflict with clickhouse

lobsterkatie:
~Snuba's~ Clickhouse's docker container is blocking port 9000, so this moves the docs server to 9001.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -10,7 +10,7 @@ home_url: https://sentry.io
 release_service: https://release-registry.services.sentry.io
 
 # Build settings
-port: 9000
+port: 9001
 markdown: kramdown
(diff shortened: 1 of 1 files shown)


lobsterkatie:
Isn't snuba our clickhouse wrapper, though? So many names...

Anyway, fixed. :-)

mattrobenolt:
Snuba is just an API and other stuff that sits in front of ClickHouse. ClickHouse is still exposed and listens on port 9000 similar to how Postgres listens on 5432 even though Sentry is the primary interface/wrapper for it.

lobsterkatie:
Ah, got it. Thanks for the clarification!

PR #834, opened 2019-03-12
http:/mysite/leaflet.js -> http://mysite/leaflet.js

ttacon:
The lack of the double slash after the protocol caught me off guard while reading the docs 😄


untitaker:
Thanks!

ttacon:
Thank you for the quick merge!

PR #849, opened 2019-03-19
Ref: remove Relay docs

cameronmcefee:
Requested by @rsomething by way of @dcramer, this removes the Relay docs for the time being. We'll reinstate them when they're ready.


dcramer:
Also if there's concerns about people need these we can just move them to the internal section. I mostly dont want folks stumbling upon this and asking questions in which the answer is "dont use this".

cc @mitsuhiko

cameronmcefee:
If search is a concern, I can exclude these docs. We just may run into a future issue when the docs are good but we forget to un-exclude them.

PR #852, opened 2019-03-20
feat: Add javascript examples for fingerprinting

lobsterkatie:
Currently, the second and third examples of [custom fingerprinting](https://docs.sentry.io/data-management/rollups/?platform=browser#custom-grouping) only exist in Python and C#. This adds JavaScript versions.

Note to reviewers:

1) I tried to hew as closely as possible to the Python versions.

2) I tagged you both b/c I don't know which one of you knows the other's language better to be able to say if I did so successfully.

3) JavaScript is not my first/native language, so please feel free to point out places where I did things wrong or weirdly.

First snippet (existing Python example):

image

First snippet (new JavaScript example):

image

Second snippet (existing Python example):

image

Second snippet (new JavaScript example):

image

### DIFF
--- a/src/collections/_documentation/data-management/fingerprint-database-connection-example/javascript.md
+++ b/src/collections/_documentation/data-management/fingerprint-database-connection-example/javascript.md
@@ -6,10 +6,6 @@ Sentry.init({
   beforeSend: (event, hint) => {
     const exception = hint.originalException;
 
-    if (!(exception instanceof Error)) {
-      return event;
-    }
-
     if (exception instanceof DatabaseConnectionError) {
       event.fingerprint = ['database-connection-error'];
     }
--- a/src/collections/_documentation/data-management/fingerprint-database-connection-example/javascript.md
+++ b/src/collections/_documentation/data-management/fingerprint-database-connection-example/javascript.md
@@ -1,0 +1,20 @@
+```javascript
+class DatabaseConnectionError extends Error {}
+
+Sentry.init({
+  ...,
+  beforeSend: (event, hint) => {
+    const exception = hint.originalException;
+
+    if (!(exception instanceof Error)) {
+      return event;
+    }
+
+    if (exception instanceof DatabaseConnectionError) {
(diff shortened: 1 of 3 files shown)


untitaker:
I can't really comment on whether the JS examples are correct. I think oftentimes there's no errortype in JS?

Probably makes sense to request review from @kamilogorek as well

src/collections/_documentation/data-management/fingerprint-database-connection-example/javascript.md
```diff
@@ -0,0 +1,20 @@
+```javascript
+class DatabaseConnectionError extends Error {}
+
+Sentry.init({
+  ...,
+  beforeSend: (event, hint) => {
+    const exception = hint.originalException;
+
+    if (!(exception instanceof Error)) {
```
kamilogorek: This block can be removed as it wont have any effect now (we return `event` at the very bottom already.

src/collections/_documentation/data-management/fingerprint-rpc-example/javascript.md
```diff
@@ -0,0 +1,34 @@
...
+    this.errorCode = errorCode;
+  }
+}
+
+Sentry.init({
+  ...,
+  beforeSend: (event, hint) => {
+    const exception = hint.originalException;
+
+    if (!(exception instanceof Error)) {
```
kamilogorek: This one as well

kamilogorek:
> I can't really comment on whether the JS examples are correct. I think oftentimes there's no errortype in JS?

@untitaker if done in a way these snippets present, it'll work just fine. You can extend native Error object and it'll be distinguishable using `instanceof` call.

src/collections/_documentation/data-management/fingerprint-database-connection-example/javascript.md
lobsterkatie: Oh, good point.

Done.

PR #864, opened 2019-03-26
React Native Manual Setup: Ordering

krzysztof-miemiec:
"Upload debug symbols" script was misplaced with "Bundle React Native code and images"


HazAT:
Thanks, this has already been merged

PR #867, opened 2019-03-28
doc: New Python Bottle integration

shenek:
related to https://github.com/getsentry/sentry-python/pull/300#issuecomment-477344421 which is being supervised by @untitaker

### DIFF
--- a/src/_data/platforms.yml
+++ b/src/_data/platforms.yml
@@ -267,8 +267,7 @@
       doc_link: /clients/python/integrations/bottle/
       wizard_parent: python
       wizard:
-        - _documentation/clients/python/index.md
-        - _documentation/clients/python/integrations/bottle.md
+        - _documentation/platforms/python/bottle.md
 -
       slug: pylons
       support_level: production
--- a/src/_data/platforms.yml
+++ b/src/_data/platforms.yml
@@ -264,7 +264,7 @@
       support_level: production
       type: framework
       name: Bottle
-      doc_link: /clients/python/integrations/bottle/
+      doc_link: /platforms/python/bottle/
       wizard_parent: python
       wizard:
         - _documentation/platforms/python/bottle.md
--- a/src/collections/_documentation/platforms/python/bottle.md
+++ b/src/collections/_documentation/platforms/python/bottle.md
@@ -57,11 +57,7 @@ You can pass the following keyword arguments to `BottleIntegration()`:
 
   In the above code, you would set the transaction to:
 
-  * `/myurl/<foo>` if you set `transaction_style="url"`. This matches the behavior of the old Raven SDK.
+  * `/myurl/<foo>` if you set `transaction_style="url"`.
   * `myendpoint` if you set `transaction_style="endpoint"`
 
   The default is `"endpoint"`.
-
-## User Feedback
-
-You can use the user feedback feature with this integration.  For more information see [User Feedback]({% link _documentation/enriching-error-data/user-feedback.md %}?platform=bottle).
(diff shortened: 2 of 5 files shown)


src/_data/platforms.yml
```diff
@@ -267,8 +267,7 @@
       doc_link: /clients/python/integrations/bottle/
```
untitaker: This is still referring to the old Raven SDK while your updated lines refer to the new client.

It should be `/platforms/python/bottle/`

src/collections/_documentation/platforms/python/bottle.md
```diff
@@ -0,0 +1,67 @@
...
+
+  ```python
+  @app.route("/myurl/<foo>")
+  def myendpoint():
+      return "ok"
+  ```
+
+  In the above code, you would set the transaction to:
+
+  * `/myurl/<foo>` if you set `transaction_style="url"`. This matches the behavior of the old Raven SDK.
```
untitaker: Is this true though? I just took a quick look at the old Bottle integration in Raven, and it seems like we never set the transaction to anything.

src/collections/_documentation/platforms/python/bottle.md
```diff
@@ -0,0 +1,67 @@
...
+  * `myendpoint` if you set `transaction_style="endpoint"`
+
+  The default is `"endpoint"`.
+
+## User Feedback
+
+You can use the user feedback feature with this integration.  For more information see [User Feedback]({% link _documentation/enriching-error-data/user-feedback.md %}?platform=bottle).
```
untitaker: This link will not work. If you want to make an example for user feedback, there's the file `_site/documentation/enriching-error-data/user-feedback-example/flask.md` that you'd need to copy. But we didn't do this for all web frameworks either.

shenek:
> Looks good so far! Did you check locally how this looks?

Yes. The installation took more time than actual doc writing. Mostly caused by a slow train connection.


sentry-bottle-icons

sentry-bottle-page

src/_data/platforms.yml
shenek: Fixed. Thanks for noticing.

src/collections/_documentation/platforms/python/bottle.md
shenek: You're right. This part was originally taken from flask integration and I forgot to check whether it is also valid.

src/collections/_documentation/platforms/python/bottle.md
shenek: Ok, I removed it for now.

src/collections/_documentation/platforms/python/bottle.md
```diff
@@ -0,0 +1,63 @@
...
+sidebar_order: 5
+---
+
+{% version_added 0.7.9 %}
+
+
+<!-- WIZARD -->
+*Import name: `sentry_sdk.integrations.bottle.BottleIntegration`*
+
+The Bottle integration adds support for the [Bottle Web Framework](https://bottlepy.org/).
```
untitaker: Could you specify here which versions of bottle the integration supports? We should have a sentence like that for some of the existing integrations.

src/collections/_documentation/platforms/python/bottle.md
```diff
@@ -0,0 +1,63 @@
+---
```
untitaker: I almost overlooked it, you need to add a link to this page in `index.md` (in the same folder) for completeness

shenek: Added a link. Thanks for noticing.

src/collections/_documentation/platforms/python/bottle.md
shenek: A note was added. I also discovered that it doens't work well with the development version of bottle (0.13). Probably I need to redo some of the patching.

untitaker:
Great work, thanks! 0.7.9 is released and the docs deploy on merge.

PR #871, opened 2019-04-02
feat: Add v5 docs for integrations

HazAT:


kamilogorek:
Looks good. I'd also add some mention of `@sentry/integrations` to node docs.

PR #873, opened 2019-04-03
fix(user feedback): Pass event id to feedback dialog in JS examples

lobsterkatie:
Examples of using the user feedback dialog in other languages pass the event id to the method, but the JS examples (both vanilla and React) don't. Sometimes, race conditions mean that a different event id is associated with the feedback than that of the event itself. Passing the event id prevents that.

This makes that change, and also adds user feedback to the Angular usage example.

### DIFF
--- a/src/collections/_documentation/platforms/javascript/react.md
+++ b/src/collections/_documentation/platforms/javascript/react.md
@@ -45,7 +45,7 @@ class ExampleBoundary extends Component {
         if (this.state.error) {
             //render fallback UI
             return (
-              <a onClick={() => Sentry.showReportDialog(this.state.eventId)}>Report feedback</a>
+              <a onClick={() => Sentry.showReportDialog({ eventId: this.state.eventId })}>Report feedback</a>
             );
         } else {
             //when there's not an error, render children untouched
--- a/src/collections/_documentation/platforms/javascript/react.md
+++ b/src/collections/_documentation/platforms/javascript/react.md
@@ -6,7 +6,7 @@ sidebar_order: 30
 To use Sentry with your React application, you will need to use `@sentry/browser` (Sentry’s browser JavaScript SDK).  
 On its own, `@sentry/browser` will report any uncaught exceptions triggered from your application.
 
-If you’re using React 16 or above, Error Boundaries are an important tool for defining the behavior of your application in the face of errors. Be sure to send errors they catch to Sentry using `Sentry.captureException`, and optionally this is also a great opportunity to surface User Feedback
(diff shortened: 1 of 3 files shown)


src/collections/_documentation/platforms/javascript/react.md
```diff
@@ -24,29 +24,28 @@ import * as Sentry from '@sentry/browser';
...
 // });
 // should have been called before using it here
-// ideally before even rendering your react app 
+// ideally before even rendering your react app
 
 class ExampleBoundary extends Component {
     constructor(props) {
         super(props);
-        this.state = { error: null };
+        this.state = { error: null, eventId: null };
```
lobsterkatie: Is this necessary?

src/collections/_documentation/platforms/javascript/react.md
```diff
@@ -24,29 +24,28 @@ import * as Sentry from '@sentry/browser';
...
     }
 
     componentDidCatch(error, errorInfo) {
       this.setState({ error });
       Sentry.withScope(scope => {
-        Object.keys(errorInfo).forEach(key => {
-          scope.setExtra(key, errorInfo[key]);
-        });
-        Sentry.captureException(error);
+          scope.setExtras(errorInfo);
```
lobsterkatie: Since 5.0 is out now, switched to the cleaner syntax.

HazAT: Sweet 👍

src/collections/_documentation/platforms/javascript/react.md
```diff
@@ -24,29 +24,28 @@ import * as Sentry from '@sentry/browser';
...
+          this.setState({eventId})
       });
     }
 
     render() {
         if (this.state.error) {
             //render fallback UI
             return (
-              <a onClick={() => Sentry.showReportDialog()}>Report feedback</a>
+              <a onClick={() => Sentry.showReportDialog(this.state.eventId)}>Report feedback</a>
```
HazAT:
```suggestion
              <a onClick={() => Sentry.showReportDialog({eventId: this.state.eventId})}>Report feedback</a>
```
Doesn't this have to be this ^

src/collections/_documentation/platforms/javascript/react.md
HazAT: Yes, it's best practice to do it.

src/collections/_documentation/platforms/javascript/react.md
lobsterkatie: D'oh! Good catch.

Fixed.

PR #881, opened 2019-04-04
fix(apidocs) Update apidocs container to use postgres

markstory:
We're removing support for sqlite from sentry so this container needs to provide a postgres database.


markstory:
I went with a postgres base as it was less lifting to install python into postgres than to install postgres into python.

PR #887, opened 2019-04-09
fix: Fix wording of RQ docs for RQ 1.0

untitaker:
cc @selwin

### DIFF
--- a/src/collections/_documentation/platforms/python/rq.md
+++ b/src/collections/_documentation/platforms/python/rq.md
@@ -24,7 +24,7 @@ Start your worker with:
 ```shell
 rq worker \
     -c mysettings \  # module name of mysettings.py
-    --sentry-dsn=""  # Keep this empty! Disables RQ's own Sentry plugin
+    --sentry-dsn=""  # only necessary for RQ < 1.0
 ```
 
 <!-- ENDWIZARD -->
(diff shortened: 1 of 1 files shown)


selwin:
Well, it depends on whether you want to mention the easiest way to do this ;). The easiest way to send errors to Sentry with sentry-sdk's default options is to do:
```shell
rq worker --sentry-dsn "https://abc@sentry.io/1"
```

If it were me, I'd also change to the config section with `rq worker -c mysettings` because it looks much simpler.

And then have a section documenting RQ < 1.0 compatibility with setting `--sentry-dsn=""`

untitaker:
It's extremely unrealistic to assume that the default options are sufficient for most users beyond the short term.

I also don't expect people to adopt RQ 1.0 immediately. I expect that existing users of RQ will adopt Sentry, and those people might have any RQ version installed. I want to deemphasize the `--sentry-dsn=""` soon but RQ 1.0 has been out for less than a month.

selwin:
> It's extremely unrealistic to assume that the default options are sufficient for most users beyond the short term.

My apologies, I didn't mean to imply that the default options are sufficient for most users. I'm just pointing out that this one liner is the easiest way for RQ to send errors to Sentry.

Whether you put that in, it's up to you, it's your documentation after all :)

> I also don't expect people to adopt RQ 1.0 immediately. I expect that existing users of RQ will adopt Sentry, and those people might have any RQ version installed. I want to deemphasize the --sentry-dsn="" soon but RQ 1.0 has been out for less than a month.

I usually update the docs (when I have the time, which is to say rarely 😢) to emphasize the most recent landscape so I don't have to remember updating the docs again, but I understand your reasoning.

Should I open an issue about updating this docs again in the next few months so we don't forget?

untitaker:
I think we have a misunderstanding here: My argument is that the ease of getting started is not important compared to the long-term experience with Sentry. Sending errors to Sentry is not the hard part, but making those errors *valuable* is, which is, when it doesn't work out of the box, already painful enough for the user to configure.

IMO it doesn't matter if I have a really easy time embedding Sentry into my application if I later have a hard time tweaking fingerprints, environments, releases, etc because I took this shortcut with the CLI option at the beginning. So you just pay the cost later. Also the argument with multiple integrations is still the most important one to me. Not because of web frameworks, but even enabling something as trivial as the [GNU backtrace](https://docs.sentry.io/platforms/python/gnu_backtrace/) integration would require you to refactor your setup. If you say all of this doesn't matter as much as having a fast and easy onboarding experience, then to me it follows that you think the default options will be enough for many users *at least in the medium-term*. It's not a bad thing to believe, but I just don't think it's true.

I have set a reminder for updating the docs but feel free to make an issue (preferrably in sentry-python rather than here).

PR #889, opened 2019-04-09
Add docs analytics to another GA property

adhiraj:
Similar to https://github.com/getsentry/blog/pull/531

### DIFF
--- a/src/_js/lib/Page.js
+++ b/src/_js/lib/Page.js
@@ -9,4 +9,8 @@ export const logPageview = function() {
     url: canonicalUrl(location.search)
   });
   window.ra.page();
+  if (window.gtag) {
+    window.gtag('config', 'UA-30327640-1', { anonymize_ip: true });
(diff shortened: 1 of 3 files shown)


src/_js/lib/Page.js
```diff
@@ -9,4 +9,8 @@ export const logPageview = function() {
     url: canonicalUrl(location.search)
   });
   window.ra.page();
+  if (window.gtag) {
```
cameronmcefee: I don't think this is necessary. The tracker code does the config stuff on the first page load and that persists until you navigate off-site or the page refreshes.

adhiraj: we don't load the tracker code until the user consents, so window.gtag won't exist if they navigate to a dynamic page and window.gtag() will throws a JS error..

cameronmcefee: Ah, forgive me. I've been away from this one for too long.

PR #899, opened 2019-04-15
fix: Specific polyfill

HazAT:


kamilogorek:
I'm not sure about that tbh. There's a big chance we'll add something in the future, forget to update the docs and break it for users. I'd just leave it as is, as only IE need polyfills anyway.

HazAT:
You are right, still, if someone prefers to just get the polyfills that are necessary to run this is it.
The argument of adding all polyfills for <= IE 11 adds another ~40 kB which is a lot.

kamilogorek:
Ok, we just have to make sure it's always up to date. (and `includes` is easy to get rid of anyway)

PR #901, opened 2019-04-16
fix: Koa docs

HazAT:


mirague:
Note that when using TypeScript it will expect an `async` (promise) handler.

This works:
```js
scope.addEventProcessor(async (event) => Sentry.Handlers.parseRequest(event, ctx.request))
```

this triggers an error:
```js
scope.addEventProcessor((event) => Sentry.Handlers.parseRequest(event, ctx.request))
```

image

HazAT:
@mirague Are you using `@sentry/node` `>= 5.x` because we changed this there.
Since our docs are not versioned, we always want to support the latest version

mirague:
Well look at that, a new version! We are indeed on 4.5

Will have to look at the changelogs and see if we can upgrade :)

PR #903, opened 2019-04-16
Don't tell people to use a random string as a DSN

dangreenisrael:
`dsn` should look like a DSN, it can't be some random string. That will cause `SentryError: Invalid Dsn`

https://github.com/wix/sentry-testkit/blob/master/test/index.test.js#L17

### DIFF
--- a/src/collections/_documentation/platforms/javascript/sentry-testkit.md
+++ b/src/collections/_documentation/platforms/javascript/sentry-testkit.md
@@ -26,7 +26,7 @@ const { testkit, sentryTransport } = sentryTestkit()
 
 // initialize your Sentry instance with sentryTransport
 Sentry.init({
-  dsn: 'some_dummy_dsn',
+  dsn: '',
   transport: sentryTransport
   //... other configurations
 })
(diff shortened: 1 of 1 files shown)


untitaker:
I would recommend taking the approach from https://github.com/getsentry/sentry-docs/blob/master/src/collections/_documentation/platforms/python/flask.md, which renders as https://docs.sentry.io/platforms/python/flask/

dangreenisrael:
This PR is for the testing docs. They are wrong. I had to go read the tests from wix's sentry test kit

untitaker:
To clarify, I'm suggesting to use `___PUBLIC_DSN___` as placeholder because this will be replaced by the user's actual DSN within the docs when the user is logged into Sentry.

untitaker:
Ah sorry, I get what you mean now. I wonder what we want to recommend here. Setting it to a bogus DSN has the effect of the SDK trying to resolve a bogus hostname in the background.

kamilogorek:
It's not possible to provide `null` nor empty string in JS SDK right now, however I'd agree that should be the case (where null/empty disables transport)

dangreenisrael:
Closing as stale

PR #909, opened 2019-04-18
Fix typographic error in `enriching-error-data/scopes.md`

DannyNemer:
Fixes the following typographic error in `enriching-error-data/scopes.md`.

Before:
> So you can either set context information in there that you don’t want to be somewhere
> else **_or create do not_** attach any context information at all

After:
> So you can either set context information in there that you don't want to be somewhere
> else **_or not_** attach any context information at all


untitaker:
Thanks!

PR #943, opened 2019-04-29
Update/architecture go

MimiDumpling:
Changed Go docs to fit two tier architecture.


cameronmcefee:
Let's save this one for tomorrow morning. This one has an impact on the wizard, so I want to be sure we can be around in case something breaks in getsentry

PR #952, opened 2019-05-01
Update/architecture ruby

MimiDumpling:
Changed Ruby docs to fit two tier architecture.


nginx.conf
```diff
@@ -206,6 +206,10 @@ server {
   location = /clients/go/integrations/http/ {
     return 302 /clients/go/integrations/;
   }
+  
+  location /clients/ruby/integrations/ {
```
cameronmcefee:
```suggestion
  location ~ ^/clients/ruby/integrations/(.*)$ {
```

PR #956, opened 2019-05-02
filestore: document S3 parameters

### LINKED ISSUE #953: Document S3 permission requirements
The S3 configuration requires slightly non-obvious permissions because the default ACL setting is `public-read`, which means that `PutObject` will fail unless you also have the `PutObjectAcl` permission.

It looks like there may be a way to set the default ACL to private which would avoid this but I have not yet been able to confirm it works:

https://github.com/getsentry/sentry/issues/11033

acdha:
I ran into a couple of hitches switching to S3 storage and wanted to document the necessary permissions and how to secure it.

Closes #953

### DIFF
--- a/src/collections/_documentation/server/filestore.md
+++ b/src/collections/_documentation/server/filestore.md
@@ -30,7 +30,24 @@ filestore.options:
 ```yaml
 filestore.backend: 's3'
 filestore.options:
-  access_key: '...'
-  secret_key: '...'
   bucket_name: '...'
 ```
+
+* Sentry uses the [boto3](https://pypi.org/project/boto3/) library to access S3. Boto supports a number of ways to [retrieve configuration information](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#guide-configuration) including EC2/ECS IAM roles (recommended), the environment, its own configuration files, and you may also specify specify  `access_key` and `secret_key` values in your Sentry `filestore.options`:
+
+    ```yaml
+    filestore.backend: 's3'
+    filestore.options:
+        bucket_name: '...'
+        access_key: '...'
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/server/filestore.md
```diff
@@ -30,7 +30,24 @@ filestore.options:
 ```yaml
 filestore.backend: 's3'
 filestore.options:
-  access_key: '...'
-  secret_key: '...'
```
untitaker: I don't quite understand why you removed those lines?

acdha: Basically because authentication is complicated and there’s a separate discussion point. I didn’t want it to look like that’s required parameter.

src/collections/_documentation/server/filestore.md
```diff
@@ -30,7 +30,24 @@ filestore.options:
...
+        secret_key: '...'
+    ```
+
+* By default, S3 objects are created with the `public-read` ACL which means that the account used must have the `PutObjectAcl` permission in addition to `PutObject`, along with `GetObject` and `DeleteObject`. If you don't want your uploaded files to be publically-accessible you can specify the default ACL using a value accepted by [boto's S3 `put_object` method](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.put_object):
+
+    ```yaml
+    filestore.backend: 's3'
+    filestore.options:
+        bucket_name: '...'
+        default_acl: 'private'
```
gtkesh: This PR/issue saved the day! I spent hours debugging this. It was a matter of `default_acl: 'private'`.

Thanks for creating this PR! :)

PeloWriter:

PR #959, opened 2019-05-03
Update/architecture legacy js

MimiDumpling:


MimiDumpling:
@cameronmcefee, I think this one is ready. Didn't update every html slug in platforms.yml because some wizards were referring to platforms/javascript instead of clients/javascript. Snapshot tests were failing when I altered anything referring to platforms/javascript.

PR #963, opened 2019-05-06
fix: Raven does not support latest Django

untitaker:


untitaker:
@lobsterkatie just wanted you to know this in case any questions are incoming. I realized that the development version of Django was broken for a while now and made the decision to draw the line at the current latest Django version.

lmk if I should notify other people about this as well.

PR #965, opened 2019-05-06
Fix/java wizard slug

MimiDumpling:
Updated the connection between platforms.yml and the wizard.


src/collections/_documentation/clients/java/integrations.md
```diff
@@ -67,7 +67,7 @@ public class MainActivity extends Activity {
 
 You can optionally configure other values such as `environment` and `release`. [See the configuration page]({%- link _documentation/clients/java/config.md -%}#configuration) for ways you can do this.
 
-<!-- WIZARD -->
+<!-- WIZARD android -->
```
cameronmcefee: Is Android used twice on purpose?

MimiDumpling: Yes. Both of these sections fall under Android usage. Is there a better way to do this?

cameronmcefee: No that's correct, I just wanted to verify.

PR #968, opened 2019-05-07
send-before-hint/javascript: Fix "Can not access 'name' of undefined" error

Turbo87:
In some cases `originalException` can be `null` or `undefined` and then the destructuring operation will fail.


untitaker:
Thanks!

PR #974, opened 2019-05-08
Document `ignore_errors` option

aviv-ebates:
Found it in https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/client.py#L164


aviv-ebates:
I'm assuming (and hoping) this is a supported option that's missing from the docs, and not an vestigial piece of code that's being removed...

untitaker:
@aviv-ebates This is just unstable functionality at the moment, see https://github.com/getsentry/sentry-python/issues/149

Thanks though :)

PR #983, opened 2019-05-15
[WIP] ref(roles): Change "owner" to "org owner"

lobsterkatie:
Per SEN-598, rename "owner" role to "org owner." Also clarify places where "owner" actually means "issue owner," and do other language clean-up.

Still to do:

- check out changes visually once my docs dev env gets unborked
- clean up pass
- update API docs

~To be merged after https://github.com/getsentry/sentry/pull/13077 and https://github.com/getsentry/sentry/pull/13239.~ DONE 

Once those are merged, the API docs need to be regenerated and those changes added to this PR.


lobsterkatie:
@cameronmcefee Should I be worried that the search doesn't seem to be updated? Or is that something that can only happen once it's live?

image

cameronmcefee:
The index gets updated when the deploy goes out, so what you see in development is the live index from `master`.

PeloWriter:

PR #986, opened 2019-05-16
feat: Improve navigation structure

dcramer:
- Auto expand most top level categories
- Bring design into alignment with readme


dcramer:
image

cameronmcefee:
### Style

As a professional designer and person responsible for our brand, I'm not super thrilled about intentionally trying to copy Readme. May of these changes feel like personal taste rather than things that are objective improvements. I would presume there are ways to get the usability benefits without completely copying their flavor.

### Sidebar

- Expanded categories seem fine.

- Button states seem incomplete

- Contrast needs improvement.

### Search in the header

I'm into it. I've wanted to go that direction. Is it intentional that the search doesn't work in mobile?

### Misc

There seems to be a handful of arbitrary formatting changes, a handful of which make things harder to understand.

src/_assets/css/_includes/bootstrap-overrides.scss
```diff
@@ -49,4 +49,4 @@ $breadcrumb-margin-bottom: 0;
 $border-width: 2px;
 
 $btn-font-size: 1rem;
-$btn-font-size-sm: 0.875rem;
```
dcramer: Just realized this wasn’t base don’t size will revert

dcramer:
More fixes, and darkened text in sidebar

image

dcramer:
image

PR #987, opened 2019-05-16
Stop "Was this page Helpful to you" from floating around.

eurica:
I'm sure there's a better way to do it, but this will cause "Was this page Helpful to you" to stop following you around and only show at the bottom of the page.

Rational: it takes up space and adds minimal value.

### DIFF
--- a/src/_layouts/default.html
+++ b/src/_layouts/default.html
@@ -46,11 +46,11 @@
             {{content}}
 
           </section>
-        </div>
 
-        <footer class="border-top bg-white px-3 py-2 flex-grow-0 d-none js-feedback-footer">
-          {% include feedback_footer.html %}
-        </footer>
+          <footer class="border-top bg-white px-3 py-2 flex-grow-0 d-none js-feedback-footer">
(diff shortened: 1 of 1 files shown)


cameronmcefee:
This particular approach creates undesirable layouts when the page is shorter than the viewport. This can be resolved if it's desired. However, this is done this way for a reason. Have you discussed this with @adhiraj first?

Screen Shot 2019-05-16 at 12 59 57 PM

eurica:
It's not my top priority on the docs.  We can wait until @adhiraj is back to understand what he's using it for.

dcramer:
Kind of wonder if we should just hide this. Once you X on it it never comes back so the current implementation feels like it falls short a bit anyways.

adhiraj:
> Once you X on it it never comes back

It actually comes back after 30m. That said, no one's using this data right now so I'd be ok completely removing it

adhiraj:
bar is gone in https://github.com/getsentry/sentry-docs/pull/1086

PR #992, opened 2019-05-19
Correct the koa usage

mehunk:
The Koa usage of the official document is wrong. I'v corrected it.


dashed:
@mehunk thank you for correcting the Koa example. 👍

PR #993, opened 2019-05-20
feat: NLog support

bruno-garcia:
Docs to NLog integration brought in by @josh-degraw 
@josh-degraw Could you please also review it?

### DIFF
--- a/src/collections/_documentation/platforms/dotnet/nlog.md
+++ b/src/collections/_documentation/platforms/dotnet/nlog.md
@@ -126,7 +126,7 @@ The SDK can also be configured via `NLog.config` XML file:
 
 #### MinimumBreadcrumbLevel
 
-A `LogLevel` which indicates the minimum level a log message has to be included as a breadcrumb. By default this value is `Info`.
+A `LogLevel` which indicates the minimum level a log message needs to be, for Sentry to receive it as an event. By default, this value is `Info`.
 
 #### MinimumEventLevel
 
--- a/src/collections/_documentation/platforms/dotnet/nlog.md
+++ b/src/collections/_documentation/platforms/dotnet/nlog.md
@@ -30,7 +30,7 @@ This package extends `Sentry` main SDK. That means that besides the logging rela
 
 Two separate settings define the minimum log level to keep the log entry as a `Breadcrumb` and to send an `Event` to Sentry. The events include any stored breadcrumb on that [scope]({% link _documentation/enriching-error-data/scopes.md %}).
 
-By default, any message with log level `Info` or higher will be kept as a `Breadcrumb`.
+By default, Sentry will keep any message with log level `Info` or higher as a `Breadcrumb`.
 
 The default value to report a log entry as an event to Sentry is `Error`.
 
--- a/src/collections/_documentation/platforms/dotnet/nlog.md
+++ b/src/collections/_documentation/platforms/dotnet/nlog.md
@@ -156,7 +156,7 @@ Determines whether or not to include event-level data as data in breadcrumbs for
 
 Custom layout for breadcrumbs.
 
-#### layout
+#### Layout
 
 Configured layout for the NLog logger.
 
--- a/src/collections/_documentation/platforms/dotnet/nlog.md
+++ b/src/collections/_documentation/platforms/dotnet/nlog.md
@@ -1,0 +1,169 @@
+---
+title: NLog
+sidebar_order: 13
+---
+
+Sentry has an integration with `NLog` through the [Sentry.NLog NuGet package](https://www.nuget.org/packages/Sentry.NLog).
+
+## Installation
+
+Using package manager:
+
+```powershell
+Install-Package Sentry.NLog -Version {% sdk_version sentry.dotnet.nlog %}
+```
+
+Or using the .NET Core CLI:
+
+```sh
+dotnet add package Sentry.NLog -v {% sdk_version sentry.dotnet.nlog %}
+```
+
+This package extends `Sentry` main SDK. That means that besides the logging related features, through this package you'll also get access to all API and features available in the main `Sentry` SDK.
+
+> NOTE: Messages logged from assemblies with the name starting with `Sentry` will not generate events.
+
+## Features
+
+* Store log messages as breadcrumbs
+* Send events to sentry
+
+Two separate settings define the minimum log level to keep the log entry as a `Breadcrumb` and to send an `Event` to Sentry. The events include any stored breadcrumb on that [scope]({% link _documentation/enriching-error-data/scopes.md %}).
+
+By default, any message with log level `Info` or higher will be kept as a `Breadcrumb`.
+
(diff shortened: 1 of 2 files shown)


bruno-garcia:
Package still beta so build isn't happy about it since it can't find it in the registry.
I'll publish it next week so once the PR is approved I can merge it right after releasing it.

src/collections/_documentation/platforms/dotnet/nlog.md
```diff
@@ -0,0 +1,169 @@
...
+<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+>
+  <extensions>
+    <add assembly="Sentry.NLog" />
+  </extensions>
+
+  <targets>
+    <target xsi:type="Sentry" name="sentry"
+            dsn="___PUBLIC_DSN___"
```
josh-degraw: One thing I just thought of is that I specified that the `Dsn` is specified as a required parameter in the `SentryTarget`, which means that if a sentry target is defined in the `NLog.config` file without a dsn specified, it will probably not register the target correctly. 

I'd have to test it to verify that, but that could cause issues if people want to set the dsn another way. So that should either be documented here or else changed on the target, which I'm happy to do if you think that's the better option.

src/collections/_documentation/platforms/dotnet/nlog.md
```diff
@@ -0,0 +1,169 @@
...
+
+Determines whether event properties will be sent to sentry as Tags or not. Defaults to false.
+
+#### IncludeEventDataOnBreadcrumbs
+
+Determines whether or not to include event-level data as data in breadcrumbs for future errors. Defaults to false.
+
+#### BreadcrumbLayout
+
+Custom layout for breadcrumbs.
```
josh-degraw: I'd been assuming anyone who is planning on using this would be familiar with `NLog` layouts, but it might be good to put a link to the documentation for layouts, e.g. either https://nlog-project.org/config/?tab=layout-renderers or https://github.com/NLog/NLog/wiki/Layout-Renderers, just to make it easy if someone is coming from sentry to nlog, for example, instead of the other way around.

Might not be that big a deal, but it could come in handy for people new to NLog.

src/collections/_documentation/platforms/dotnet/nlog.md
```diff
@@ -0,0 +1,169 @@
...
+> NOTE: Messages logged from assemblies with the name starting with `Sentry` will not generate events.
+
+## Features
+
+* Store log messages as breadcrumbs
+* Send events to sentry
+
+Two separate settings define the minimum log level to keep the log entry as a `Breadcrumb` and to send an `Event` to Sentry. The events include any stored breadcrumb on that [scope]({% link _documentation/enriching-error-data/scopes.md %}).
+
+By default, any message with log level `Info` or higher will be kept as a `Breadcrumb`.
```
MimiDumpling:
```suggestion
By default, Sentry will keep any message with log level `Info` or higher as a `Breadcrumb`.
```

src/collections/_documentation/platforms/dotnet/nlog.md
```diff
@@ -0,0 +1,169 @@
...
+
+        // Send the logger name as a tag
+        o.AddTag("logger", "${logger}");
+
+        // All Sentry Options are accessible here.
+    });  
+```
+
+It's also possible to initialize the SDK through the NLog integration (as opposed to using `SentrySdk.Init`). 
+This is useful when the NLog is the only integration being used in your application. To initialize the Sentry SDK through the NLog integration, provide it with the DSN:
```
MimiDumpling:
```suggestion
This is useful when the NLog is the only integration you are using in your application. To initialize the Sentry SDK through the NLog integration, provide it with the DSN:
```

src/collections/_documentation/platforms/dotnet/nlog.md
```diff
@@ -0,0 +1,169 @@
...
+    <logger name="*" writeTo="sentry" />
+  </rules>
+</nlog>
+```
+
+### Options
+
+#### MinimumBreadcrumbLevel
+
+A `LogLevel` which indicates the minimum level a log message has to be included as a breadcrumb. By default this value is `Info`.
```
MimiDumpling:
```suggestion
A `LogLevel` which indicates the minimum level a log message needs to be, for Sentry to receive it as an event. By default, this value is `Info`.
```

josh-degraw: I think this might be a more appropriate change:

`LogLevel` that indicates the minimum level a log message needs to be in order for Sentry to receive it as a breadcrumb in an event. By default, this value is `Info`.

_(it should still specify that the message is being sent as a breadcrumb, not an event)_

src/collections/_documentation/platforms/dotnet/nlog.md
josh-degraw: Also might want to change 
```
This is useful when the NLog is...
```
to
```
This is useful when NLog is...
```

src/collections/_documentation/platforms/dotnet/nlog.md
bruno-garcia: Right. We should make this not required then. Firstly because the lack of `DSN` means _don't initialize Sentry at all_ but also because It's important to make it possible for integration *not* to initialize the SDK. The canonical way to init any Sentry SDK from the unified API is via `Sentry.Init` so allowing an integration initialize the SDK is just a convenience for those who use a single integration. It also helps making things idiomatic.

Could you please modify the integration to make `DSN` optional?

src/collections/_documentation/platforms/dotnet/nlog.md
bruno-garcia: I hit commit suggestion without scrolling all the way. @josh-degraw is right, it's to be included as breadcrumb, not event. So the original text was right although perhaps confusing as most of what I write.

src/collections/_documentation/platforms/dotnet/nlog.md
```diff
@@ -0,0 +1,169 @@
...
+      </options>
+
+      <!--Add any desired additional tags that will be sent with every message -->
+      <tag name="logger" layout="${logger}" />
+      <tag name="example" layout="sentry-nlog" />
+    </target>
+  </targets>
+
+  <rules>
+    <logger name="*" writeTo="sentry" />
```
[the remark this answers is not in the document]
bruno-garcia: I've added a note about this. I tried to be detailed on how it works but still seems confusing. Would appreciate some suggestions.
Or feel free to make a PR on this branch too if you wish :)

src/collections/_documentation/platforms/dotnet/nlog.md
josh-degraw: Can do. I should be able to send a PR this weekend.

PR #999, opened 2019-05-22
feat: Improve React docs

dcramer:


src/collections/_documentation/platforms/javascript/getting-started-dsn/react.md
```diff
@@ -1,42 +1,9 @@
-<!-- WIZARD -->
```
dcramer: i think this was copy paste - afaict it shouldnt be needed

src/collections/_documentation/platforms/javascript/react.md
```diff
@@ -4,17 +4,19 @@ sidebar_order: 30
 ---
 <!-- WIZARD -->
 To use Sentry with your React application, you will need to use `@sentry/browser` (Sentry’s browser JavaScript SDK).
```
dcramer: I would insert the install/configure instructions here, then error boundaries comes after they're fully setup

dcramer:
jsx support doesnt seem to exist/work

image

image

dcramer:
image

dcramer:
Updated to remove version

src/collections/_documentation/platforms/javascript/getting-started-dsn/react.md
cameronmcefee: Confirmed, this doc doesn't appear to be referenced in platforms.yml

dcramer:
Jk my jsx is wrong

dcramer:
To me this is an improvement if I stumble upon the React page itself, but something we should decide broadly: do we embed the setup docs on every framework, or should we just have a generic entry blurb that directs them to the core JavaScript setup. In some framework cases the setup will change so it can still make sense.

PR #1002, opened 2019-05-23
ref(javascript): Default to browsernpm over browser CDN

EvanPurkhiser:
By default in the sentry wizard clicking 'javascript' gives you the CDN version. I would argue that most people these days likely want the yarn/npm version.

### DIFF
--- a/src/_plugins/platform_api.rb
+++ b/src/_plugins/platform_api.rb
@@ -78,9 +78,9 @@ Jekyll::Hooks.register :site, :post_render, priority: :high do |site|
     is_self = group_slug === platform["slug"]
     grouped_file = group_slug === platform["slug"] ? file_name : "#{group_slug}/#{file_name}"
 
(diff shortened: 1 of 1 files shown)


dcramer:
im neutral on this as im not really sure thats true, and seems fairly subjective

certainly for certain platforms it is (e.g. react)

benvinegar:
I know that when I'm quickly messing around with Sentry I just whip together a static web page and use the CDN example.

I think copy/pasting is more approachable (a lot of websites/apps out there that don't use Webpack, Babel, etc.) and we should stick to that. We can always have a callout.

Otherwise I think we need to look at some kind of metric/stats to make an intelligent decision here.

EvanPurkhiser:
I think what we can do is have a 'ES6' option that lives right next to the Javascript option (or something line that) since right now we don't show both in the wizards 'popular' menu

image

PR #1007, opened 2019-05-23
ref(symfony): Add example of triggering an error

EvanPurkhiser:


src/collections/_documentation/clients/php/integrations.md
```diff
@@ -398,4 +398,18 @@ Add your DSN to `app/config/config.yml`:
 sentry:
     dsn: "___PUBLIC_DSN___"
 ```
+
+You can easily verify that Sentry is capturing errors in your Symfony application by creating a debug route that will throw an exception:
+
+```php
+/**
+ * @Route("/debug-sentry")
```
markstory: I am not sure that annotations are included in the default app setup for symfony4 applications. Defining routes in a yaml file is another option.

EvanPurkhiser: Ideally I'd like to not have two separate code blocks.

Would we be okay showing this as the example and letting users decide how they want to configure the routing themselves?

markstory: Symfony folks will have at least seen this style of routing before and know how to translate that int the syntax they use.

PR #1014, opened 2019-05-24
feat: Add styling to TOC

dcramer:


dcramer:
image

src/_js/lib/TableOfContents.js
```diff
@@ -0,0 +1,26 @@
+$(() => {
+  let lastMatch = null;
+  const updateLocation = () => {
+    const scrollTop = $('.main-scroll').scrollTop();
+    let curMatch;
+    $(':header').each(function() {
+      const id = $(this).attr('id');
+      if (id && scrollTop >= $(this).position().top) {
```
dcramer: had to use .position instead of .offset since the scrollbar is not on document (but instead on .main-scroll)

PR #1024, opened 2019-05-28
ref(python): Explain when to import and initialize

EvanPurkhiser:


src/collections/_documentation/error-reporting/getting-started-config/python.md
```diff
@@ -1,3 +1,5 @@
+Import and initialize the sentry SDK early in your applications startup:
```
MimiDumpling:
```suggestion
Import and initialize the Sentry SDK early in your application's setup:
```

PR #1025, opened 2019-05-28
ref(php): Better wording for composer usage

EvanPurkhiser:


src/collections/_documentation/error-reporting/getting-started-install/php.md
```diff
@@ -1,8 +1,6 @@
-To install the SDK you will need to be using `Composer` in your project. To install
-it please see the [docs](https://getcomposer.org/download/).
-
-If you just want to get started quickly you should run the following command:
```
EvanPurkhiser: Yes I do want to get started quickly.

PR #1029, opened 2019-05-29
fix(minidump): Do not cutoff `ul` in wizard html

EvanPurkhiser:
When the `<!-- WIZARD -->` tags are processed they will cutoff the `ul` at the end of a list, causing additional wizard content to be nested under the last list item.

This fixes this for minidumps

### DIFF
--- a/src/collections/_documentation/platforms/minidump/index.md
+++ b/src/collections/_documentation/platforms/minidump/index.md
@@ -13,12 +13,14 @@ In order to receive symbolicated stack traces, you have to upload debug informat
 <!-- WIZARD -->
 ## Platform and Language Support
 
-Minidumps are currently supported for **Windows, macOS and Linux**. There is no limitation as to which programming language can be used. Sentry can also demangle symbols from the following languages; other languages will show the mangled name instead:
+Minidumps are currently supported for **Windows, macOS and Linux**. There is no limitation as to which programming language can be used. Sentry can also demangle symbols from the following languages:
 
 -   C and C++
 -   ObjectiveC and ObjectiveC++
 -   Swift
 -   Rust
+
+Other languages will show the mangled name instead.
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/platforms/minidump/index.md
```diff
@@ -13,12 +13,14 @@ In order to receive symbolicated stack traces, you have to upload debug informat
 <!-- WIZARD -->
 ## Platform and Language Support
 
-Minidumps are currently supported for **Windows, macOS and Linux**. There is no limitation as to which programming language can be used. Sentry can also demangle symbols from the following languages; other languages will show the mangled name instead:
+Minidumps are currently supported for **Windows, macOS and Linux**. There is no limitation as to which programming language can be used. Sentry can also demangle symbols from the following languages:
```
MimiDumpling:
```suggestion
Sentry currently supports minidumps for **Windows, macOS and Linux**. There is no limitation as to which programming language you can use. Sentry can also demangle symbols from the following languages:
```

src/collections/_documentation/platforms/minidump/index.md
```diff
@@ -13,12 +13,14 @@ In order to receive symbolicated stack traces, you have to upload debug informat
...
 
 -   C and C++
 -   ObjectiveC and ObjectiveC++
 -   Swift
 -   Rust
+
+Other languages will show the mangled name instead.
```
MimiDumpling:
```suggestion
These other languages will show the mangled name instead.
```

EvanPurkhiser: Actually it is saying languages *not* in the list above

MimiDumpling: Aha. Okay, made a new suggestion.

PR #1048, opened 2019-06-06
Grey out unsupported options

ceorourke:
Old:
Screenshot 2019-06-06 12 06 54

New:
Screenshot 2019-06-06 12 05 10

### DIFF
--- a/src/collections/_documentation/error-reporting/configuration/index.md
+++ b/src/collections/_documentation/error-reporting/configuration/index.md
@@ -216,15 +216,17 @@ setup that requires proxy authentication.
 {:.config-key}
 ### `http-proxy`
 
+{% unsupported browser %}
 When set a proxy can be configured that should be used for outbound requests.  This is also used for
 HTTPS requests unless a separate `https-proxy` is configured.  Note however that not all SDKs
 support a separate HTTPS proxy.  SDKs will attempt to default to the system-wide configured proxy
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/error-reporting/configuration/index.md
```diff
@@ -216,15 +216,17 @@ setup that requires proxy authentication.
...
 HTTPS requests unless a separate `https-proxy` is configured.  Note however that not all SDKs
 support a separate HTTPS proxy.  SDKs will attempt to default to the system-wide configured proxy
 if possible.  For instance, on unix systems, the `http_proxy` environment variable will be picked up.
+{% endunsupported %}
 
 {:.config-key}
 ### `https-proxy`
 
-{% unsupported csharp aspnetcore php %}
+{% unsupported browser csharp aspnetcore php %}
```
lobsterkatie: Given these, you might check w the SDK folks if those SDK's should also be on `http-proxy`, or if those other ones support `http` but not `https`.

ceorourke: this note is more for me, but confirmed that csharp and aspnetcore don't need to change, awaiting confirmation about php

ceorourke: also confirmed w/ PHP that this doesn't need to change

PR #1049, opened 2019-06-07
Added clarification to controller location for Express

twolfson:
While setting up a new project, I spent a very long and sad amount of time trying to figure out why Sentry wasn't receiving my errors that I was clearly seeing locally. It turns out I put the error handler too high up. In this PR:

- Added clarification to controller location for Express

### DIFF
--- a/src/collections/_documentation/platforms/node/express.md
+++ b/src/collections/_documentation/platforms/node/express.md
@@ -26,6 +26,8 @@ Sentry.init({ dsn: '___PUBLIC_DSN___' });
 // The request handler must be the first middleware on the app
 app.use(Sentry.Handlers.requestHandler());
 
+// All controllers should live here
+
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/platforms/node/express.md
```diff
@@ -26,6 +26,8 @@ Sentry.init({ dsn: '___PUBLIC_DSN___' });
 // The request handler must be the first middleware on the app
 app.use(Sentry.Handlers.requestHandler());
 
+// All controllers should live here
+
```
dashed: @twolfson mind putting an example of an app-level middleware (e.g. `app.get()`) to drive the point home?

twolfson: Sure thing!

PR #1050, opened 2019-06-07
Add more detailed explanation on beforeBreadcrumb function

wsloth:
I was stuck on why my Sentry breadcrumbs were just described as `generic` with a timestamp attached to it. I could not clearly find out where the issue was, I thought it would be because I was adding the breadcrumb incorrectly.

Instead, I found that in the `beforeBreadcrumb` function you had to return the actual (modified) breadcrumb instead of `true` or `false`, which I could not clearly find in the docs.

Hence this PR! If you have a different idea on how it should be worded, let me know!

### DIFF
--- a/src/collections/_documentation/error-reporting/configuration/index.md
+++ b/src/collections/_documentation/error-reporting/configuration/index.md
@@ -195,9 +195,10 @@ sending.
 ### `before-breadcrumb`
 
 This function is called with an SDK specific breadcrumb object before the breadcrumb is added to the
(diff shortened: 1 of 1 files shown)


wsloth:
Can we merge it? 😄

untitaker:
ofc, was just waiting for travis over lunch break :)

untitaker:
Thanks!

wsloth:
Thank you for the quick merge!

PR #1060, opened 2019-06-11
feat: Add sampling documentation

dcramer:


dcramer:
As we inch towards the APM launch this kind of conversation (around sampling) is going to be increasingly important and visible inside of the product. This sets us up for a basis to expand and also hits an important keyword in search: "sampling"

dcramer:
Also just now realizing that a lot of this was under Quotas - id like to move away from Quotas/Rate Limits and focus on client side sampling. It's less costly for us and less prone to abuse.

dcramer:
(so we could consider combining the quotas doc into this, and maybe minimize the rate limits component?)

MimiDumpling:
> (so we could consider combining the quotas doc into this, and maybe minimize the rate limits component?)

I think this sounds reasonable. I'll sync up with Katie to see if she has an extra insights.

MimiDumpling:
> So has this been superseded by #1062 or are they both still live? (Left some comments there.)

Hi @lobsterkatie, yeah let's keep the conversation on #1062 for now -- since that's where most of the comments are. In the end, we'll go with @dcramer's suggestion and make it one page with a heavy emphasis on Sampling. Will definitely appreciate your help with that. :D

PR #1061, opened 2019-06-11
[WIP: do not merge] added use-cases for Hints using beforeBreadcrumb and beforeSend

MimiDumpling:
Hints section was unclear on how/what situation to use the hints. Added two use-cases to help clarify and provide some guidance on usage of hints.

### DIFF
--- a/src/collections/_documentation/platforms/javascript/index.md
+++ b/src/collections/_documentation/platforms/javascript/index.md
@@ -960,7 +960,7 @@ They are available in two places: `beforeSend`/`beforeBreadcrumb` and `eventProc
 
 The `before-send` callback is passed the event and a second argument `hint` which holds one or more hints. Typically this hint holds the original exception so that additional data can be extracted or grouping is affected. `before-send` will be called right before the event is sent to the server, so it’s the last place where you can edit its data.
 
-In this example the fingerprint is forced to a common value if an exception of a certain type has been caught:
+In this example, the fingerprint is forced to a common value if an exception of a certain type has been caught:
 
 ```javascript
 import * as Sentry from '@sentry/browser';
--- a/src/collections/_documentation/platforms/javascript/index.md
+++ b/src/collections/_documentation/platforms/javascript/index.md
@@ -956,6 +956,26 @@ They are available in two places: `beforeSend`/`beforeBreadcrumb` and `eventProc
 
 #### Hints for Events
 
+##### Use-Case: `before-send`
+
+The `before-send` callback is passed the event and a second argument `hint` which holds one or more hints. Typically this hint holds the original exception so that additional data can be extracted or grouping is affected. `before-send` will be called right before the event is sent to the server, so it’s the last place where you can edit its data.
+
+In this example the fingerprint is forced to a common value if an exception of a certain type has been caught:
+
+```javascript
+import * as Sentry from '@sentry/browser';
+
+init({
+  beforeSend(event, hint) {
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -956,6 +956,26 @@ They are available in two places: `beforeSend`/`beforeBreadcrumb` and `eventProc
 
 #### Hints for Events
 
+##### Use-Case: `before-send`
```
dcramer: could we just put all of these on a child page? its not core to usage and is more of "go much deeper on your integration" focus - same bucket as most of advanced usage

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -956,6 +956,26 @@ They are available in two places: `beforeSend`/`beforeBreadcrumb` and `eventProc
...
+
+The `before-send` callback is passed the event and a second argument `hint` which holds one or more hints. Typically this hint holds the original exception so that additional data can be extracted or grouping is affected. `before-send` will be called right before the event is sent to the server, so it’s the last place where you can edit its data.
+
+In this example, the fingerprint is forced to a common value if an exception of a certain type has been caught:
+
+```javascript
+import * as Sentry from '@sentry/browser';
+
+init({
+  beforeSend(event, hint) {
```
dcramer: i think we call this 'hints' elsewhere? if so we should match that in the code sample (I might be wrong though)

lobsterkatie: Everywhere I've seen the parameter used, it's been called `hint`. I think the heading **Hints** is plural only because calling it **Hint** would be weird.

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -956,6 +956,26 @@ They are available in two places: `beforeSend`/`beforeBreadcrumb` and `eventProc
 
 #### Hints for Events
 
+##### Use-Case: `before-send`
+
+The `before-send` callback is passed the event and a second argument `hint` which holds one or more hints. Typically this hint holds the original exception so that additional data can be extracted or grouping is affected. `before-send` will be called right before the event is sent to the server, so it’s the last place where you can edit its data.
```
lobsterkatie: > so that additional data can be extracted or grouping is affected
Not sure what you mean by "grouping is affected" here, but IMHO you get enough meaning by just leaving it off.

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -967,6 +987,27 @@ basic stack trace. This exception is stored here for further data extraction.
...
 
+##### Use-Case: `before-breadcrumb`
+
+Similar to `before-send` hints are also supported in `before-breadcrumb`. This is particularly useful when breadcrumbs should be augmented or changed with some extra data from log records or events. The following example shows a typical situation where you might want to extract extra information and how to do it:
+
+```javascript
+import * as Sentry from '@sentry/browser';
+
+Sentry.init({
+  dsn: 'https://<key>@sentry.io/',
```
lobsterkatie: `'https://<key>@sentry.io/'` <- should be `'https://<key>@sentry.io/project-id'

src/collections/_documentation/platforms/javascript/index.md
kamilogorek: It's singular in JS and Go as well.

PR #1062, opened 2019-06-12
[WIP: do not merge] add Sampling section to Quotas & Filtering

MimiDumpling:
Conversations around sampling will increase as Sentry approaches APM launch. This section will help kickoff discussion and raise visibility of sampling.

### DIFF
--- a/src/collections/_documentation/accounts/quotas.md
+++ b/src/collections/_documentation/accounts/quotas.md
@@ -68,6 +68,21 @@ Not all events count towards your monthly event quota. In general, when events a
     content="Delete and discard and per-project rate limits are only available on Business and Enterprise plans"
 %}
 
+## Sampling
+
+In some situations, you may want to sample data within Sentry. Sampling is a valuable tool if you're concerned about the cost of Sentry, or if some data is naturally very high volume and noisy (like [_Security Policy Reports_]({%- link _documentation/error-reporting/security-policy-reporting.md -%})).
+
+All of Sentry's SDKs provide a `sampleRate` configuration. For example, in JavaScript:
+
+```javascript
+Sentry.init({
+  // ...
+  sampleRate: 0.5, // send 50% of events
+});
+```
+
+The sample rate represents the percentage of events to be sent in the range of `0.0` to `1.0`. The default is `1.0`, which means that 100% of events are sent. If set to `0.1`, only 10% of events will be sent. Events are picked randomly.
+
 ## Increasing Quotas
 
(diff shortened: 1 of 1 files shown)


MimiDumpling:
@dcramer talked with @LSSangha and we feel like this is a good place for Sampling. :) Let us know what you think.

dcramer:
So the main thing here is this isnt just a documentation change I want to make. I want us to reframe how we think about "managing data you send to sentry". Sampling is the answer folks need to rely on. The rate limits and expectation of rate limits have been super abusive to the point where we're considering drastically reducing their effectiveness. I would rather the whole section be about Sampling as thats a core theme of Sentry going forward (and is why we added sampleRate). The framing of the paragraphs makes less sense when they're a subsection of the page and more sense when its "this is why you might not need or want all of your data"

MimiDumpling:
> So the main thing here is this isnt just a documentation change I want to make. I want us to reframe how we think about "managing data you send to sentry". Sampling is the answer folks need to rely on. The rate limits and expectation of rate limits have been super abusive to the point where we're considering drastically reducing their effectiveness. I would rather the whole section be about Sampling as thats a core theme of Sentry going forward (and is why we added sampleRate). The framing of the paragraphs makes less sense when they're a subsection of the page and more sense when its "this is why you might not need or want all of your data"

Okay. I can put together a plan on how to restructure the whole thing so that Sampling take precedence.

src/collections/_documentation/accounts/quotas.md
```diff
@@ -68,6 +68,21 @@ Not all events count towards your monthly event quota. In general, when events a
     content="Delete and discard and per-project rate limits are only available on Business and Enterprise plans"
 %}
 
+## Sampling
+
+In some situations, you may want to sample data within Sentry. Sampling is a valuable tool if you're concerned about the cost of Sentry, or if some data is naturally very high volume and noisy (like [_Security Policy Reports_]({%- link _documentation/error-reporting/security-policy-reporting.md -%})).
+
+All of Sentry's SDKs provide a `sampleRate` configuration. For example, in JavaScript:
```
lobsterkatie: > All of Sentry's SDKs provide a `sampleRate` configuration.

I believe this is true of all unified SDK's. Is it true of all the legacy ones as well?

LSSangha: I believe almost every SDK has some type of sampling but not 100% sure. For example, raven-JS has sample.

lobsterkatie: Right, but it's "spelled" differently in some of the legacy SDK's, I believe. Worth checking to know for sure.

MimiDumpling:
@dcramer, Katie and I will tackle this next week.

PR #1064, opened 2019-06-12
feat: Expose search keywords

dcramer:


dcramer:
This will allow us to better control search results as keywords are higher priority (lower than title), and dont have to exist in the content.

dcramer:
image

PR #1065, opened 2019-06-12
feat: Simplify quickstart

dcramer:
- People do not need to roll their own SDK - move that topic elsewhere
- None of the SDKs presented are secret-key based
- Fix some minor language


dcramer:
Was just reading this, saw the typo, then went down the rabbit hole.

dcramer:
image

PR #1081, opened 2019-06-26
Fix PHP docs to use Options class for transport

baohx2000:


ste93cry:
Superseded by #1276, so it can be closed

PR #1082, opened 2019-06-29
Adding support to Ionic v4 in docs

renansmoreira:
IonicErrorHandler doesn't exist anymore in Ionic 4+. Instead, we need to use ErrorHandler from angular/core


dashed:
@renansmoreira thanks for contributing! 👍

PR #1083, opened 2019-06-30
Typo when calling the SDK

B-Galati:
I just had this issue when I was setting up the SDK.

Let me know if I am  correct and I will fix it everywhere else.

Cheers!

### DIFF
--- a/src/collections/_documentation/error-reporting/getting-started-config/php.md
+++ b/src/collections/_documentation/error-reporting/getting-started-config/php.md
@@ -1,5 +1,5 @@
 To capture all errors, even the one during the startup of your application, you should initialize the Sentry PHP SDK as soon as possible.
 
 ```php
-Sentry\init(['dsn' => '___PUBLIC_DSN___' ]);
+\sentry\init(['dsn' => '___PUBLIC_DSN___' ]);
 ```


untitaker:
🤔 https://github.com/getsentry/sentry-docs/commit/e6c1252b13e7d0957ad2dcee36a96ae496ae0619

B-Galati:
Sorry I made a typo it's `Sentry` with an uppercase `S`.

Regarding the answer, at the moment if you copy/paste the given code from the doc it will not work without the leading `\` and that is a pity TBH.

ste93cry:
>if you copy/paste the given code from the doc it will not work without the leading \ and that is a pity TBH

The example supposes that you're code is not namespaced and so there is no need to prefix the `use` statement with `\`. However, if you are willing to update all the code samples to have the leading `\` it would be more than welcome since that way the code will work for both cases (namespaced and not)

src/collections/_documentation/error-reporting/getting-started-config/php.md
```diff
@@ -1,5 +1,5 @@
 To capture all errors, even the one during the startup of your application, you should initialize the Sentry PHP SDK as soon as possible.
 
 ```php
-Sentry\init(['dsn' => '___PUBLIC_DSN___' ]);
+\sentry\init(['dsn' => '___PUBLIC_DSN___' ]);
```
ste93cry: This is incorrect, the `S` must be uppercase

B-Galati:
I will thank you for your feedback

HazAT:
I closing this since it has become stale and it is already fixed.

PR #1085, opened 2019-07-01
docs: Go async delivery and goroutines hub

kamilogorek:


src/collections/_documentation/platforms/go/goroutines.md
```diff
@@ -0,0 +1,53 @@
+---
+title: Goroutines
+sidebar_order: 4
+---
+
+A goroutine is a lightweight thread managed by the Go runtime. It is run concurrently with other functions in your code. And because of this, it has to keep track of it's own data locally. Otherwise, there is a chance that you will override your informations stored in the Scope. More on this in [Scopes and Hubs]({%- link _documentation/enriching-error-data/scopes.md -%}?platform={{ include.platform }}) section.
+
+The easiest way to handle this, is to create a new `Hub`, however this would require you to rebind the current `Client` and handle `Scope` yourself. That is why we provide a helper method called `Clone`. It takes care of creating a `Hub`, cloning existing `Scope` and reassigning it alongside `Client` to newly create instance.
```
tonyo:
```suggestion
The easiest way to handle this, is to create a new `Hub` for every goroutine you start, however this would require you to rebind the current `Client` and handle `Scope` yourself. That is why we provide a helper method called `Clone`. It takes care of creating a `Hub`, cloning existing `Scope` and reassigning it alongside `Client` to newly create instance.
```

src/collections/_documentation/platforms/go/goroutines.md
```diff
@@ -0,0 +1,53 @@
...
+
+Once cloned, `Hub` is completly isolated and can be used safely inside concurrent call. However, instead of using globally exposed methods, they should be called directly on the `Hub`.
+
+Here are two examples, one that is non-deterministic, would leak information between threads and could trigger a concurrent-write panic and one that is totally safe, and should be used instead.
```
tonyo:
```suggestion
Here are two examples: one that is non-deterministic, would leak information between threads, and could trigger a concurrent-write panic; and one that is totally safe, and should be used instead.
```

src/collections/_documentation/platforms/go/index.md
```diff
@@ -42,12 +42,24 @@ To use `sentry-go`, you’ll need to import the `sentry-go` package and initiali
 
 More on this in [Configuration]({%- link _documentation/platforms/go/config.md -%}) section.
 
+## Usage {#usage}
+
+{% capture __alert_content -%}
+  By default, Sentry Go SDK uses asynchronous transport, which in the code example below, requires explicit awaiting for event delivery to be finished using `sentry.Flush` method. It is necessary, because Go programs does not wait for async HTTP calls to return a response and close the process immediately if it reaches the end of the `main` function. It would not be required inside a running go-routine or if you would use `HTTPSyncTransport`, which you can read about in `Transports` section.
```
tonyo:
```suggestion
  By default, Sentry Go SDK uses asynchronous transport, which in the code example below requires an explicit awaiting for event delivery to be finished using `sentry.Flush` method. It is necessary, because otherwise the program would not wait for the async HTTP calls to return a response, and exit the process immediately when it reached the end of the `main` function. It would not be required inside a running goroutine or if you would use `HTTPSyncTransport`, which you can read about in `Transports` section.
```

src/collections/_documentation/platforms/go/goroutines.md
```diff
@@ -0,0 +1,53 @@
+---
+title: Goroutines
+sidebar_order: 4
+---
+
+A goroutine is a lightweight thread managed by the Go runtime. It is run concurrently with other functions in your code. And because of this, it has to keep track of it's own data locally. Otherwise, there is a chance that you will override your informations stored in the Scope. More on this in [Scopes and Hubs]({%- link _documentation/enriching-error-data/scopes.md -%}?platform={{ include.platform }}) section.
```
tonyo:
```suggestion
A goroutine is a lightweight thread managed by the Go runtime. Goroutines can run concurrently, and because of this, every goroutine has to keep track of its own Sentry-related data locally. Otherwise, there is a chance that you will override your data stored in the Scope. More on this in [Scopes and Hubs]({%- link _documentation/enriching-error-data/scopes.md -%}?platform={{ include.platform }}) section.
```

PR #1093, opened 2019-07-02
docs: Update node frameworks doc to include shouldHandleError option

kamilogorek:
ref: https://github.com/getsentry/sentry-javascript/pull/2146

### DIFF
--- a/src/collections/_documentation/platforms/node/connect.md
+++ b/src/collections/_documentation/platforms/node/connect.md
@@ -48,4 +48,18 @@ connect(
 ).listen(3000);
 ```
 
+By default, `errorHandler` will capture only errors with status code of `500` or higher. If you want to change it, provide it with `shouldHandleError` callback, which accepts middleware error as its argument and decides, whether error should be sent or not, by returning an appropriate boolean value.
+
+```js
+app.use(Sentry.Handlers.errorHandler({
+  shouldHandleError(error) {
+    // Capture all 404 and 500 errors
+    if (error.status === 404 || error.status === 500) {
+      return true
+    }
+    return false
+  }
+}));
+```
+
 <!-- ENDWIZARD -->
--- a/src/collections/_documentation/platforms/node/connect.md
+++ b/src/collections/_documentation/platforms/node/connect.md
@@ -48,6 +48,37 @@ connect(
 ).listen(3000);
 ```
 
+`requestHandler` accepts some options that let you decide what data should be included in the event sent to Sentry.
+
+Possible options are:
+
+```js
+// request data (path, method, query params etc.)
+request?: boolean; // default: true
+// server name
+serverName?: boolean; // default: true
+// generate transaction name
+//   path == request.path (eg. "/foo")
+//   methodPath == request.method + request.path (eg. "GET|/foo")
+//   handler == function name (eg. "fooHandler")
+transaction?: boolean | 'path' | 'methodPath' | 'handler'; // default: true = 'methodPath'
+// keys to be extracted from req.user
+user?: boolean | string[]; // default: true = ['id', 'username', 'email']
+// node version
+version?: boolean; // default: true
+// timeout for fatal route errors to be delivered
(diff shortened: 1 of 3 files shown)


src/collections/_documentation/platforms/node/connect.md
```diff
@@ -48,4 +48,49 @@ connect(
 ).listen(3000);
 ```
 
+`requestHandler` accepts some options that let you decide what data should be included in the event sent to Sentry.
+
+Possible options are:
+
+```js
+// request data (path, method, query params etc.)
```
MimiDumpling:
```suggestion
// request data (path, method, query params, etc.)
```

src/collections/_documentation/platforms/node/connect.md
```diff
@@ -48,4 +48,49 @@ connect(
...
+transaction?: boolean | 'path' | 'methodPath' | 'handler'; // default: true = 'methodPath'
+// keys to be extracted from req.user
+user?: boolean | string[]; // default: true = ['id', 'username', 'email']
+// node version
+version?: boolean; // default: true
+// timeout for fatal route errors to be delivered
+flushTimeout?: number; // default: 2000
+```
+
+For example, if you want to skip server name and add just user, you would use `requestHandler` like this:
```
MimiDumpling:
```suggestion
For example, if you want to skip the server name and add just user, you would use `requestHandler` like this:
```

src/collections/_documentation/platforms/node/connect.md
```diff
@@ -48,4 +48,49 @@ connect(
...
+
+```js
+app.use(Sentry.Handlers.requestHandler({
+  serverName: false,
+  user: ['email']
+}));
+```
+
+By default, `errorHandler` will capture only errors with status code of `500` or higher. If you want to change it, provide it with `shouldHandleError` callback, which accepts middleware error as its argument and decides, whether error should be sent or not, by returning an appropriate boolean value.
```
MimiDumpling:
```suggestion
By default, `errorHandler` will capture only errors with a status code of `500` or higher. If you want to change it, provide it with the `shouldHandleError` callback, which accepts middleware errors as its argument and decides, whether an error should be sent or not, by returning an appropriate boolean value.
```

src/collections/_documentation/platforms/node/express.md
```diff
@@ -53,6 +53,51 @@ app.get('/debug-sentry', function mainHandler(req, res) {
...
+transaction?: boolean | 'path' | 'methodPath' | 'handler'; // default: true = 'methodPath'
+// keys to be extracted from req.user
+user?: boolean | string[]; // default: true = ['id', 'username', 'email']
+// node version
+version?: boolean; // default: true
+// timeout for fatal route errors to be delivered
+flushTimeout?: number; // default: 2000
+```
+
+For example, if you want to skip server name and add just user, you would use `requestHandler` like this:
```
MimiDumpling:
```suggestion
For example, if you want to skip the server name and add just user, you would use `requestHandler` like this:
```

src/collections/_documentation/platforms/node/express.md
```diff
@@ -53,6 +53,51 @@ app.get('/debug-sentry', function mainHandler(req, res) {
...
+
+```js
+app.use(Sentry.Handlers.requestHandler({
+  serverName: false,
+  user: ['email']
+}));
+```
+
+By default, `errorHandler` will capture only errors with status code of `500` or higher. If you want to change it, provide it with `shouldHandleError` callback, which accepts middleware error as its argument and decides, whether error should be sent or not, by returning an appropriate boolean value.
```
MimiDumpling:
```suggestion
By default, `errorHandler` will capture only errors with a status code of `500` or higher. If you want to change it, provide it with the `shouldHandleError` callback, which accepts middleware errors as its argument and decides, whether an error should be sent or not, by returning an appropriate boolean value.
```

MimiDumpling:
👍

PR #1066, opened 2019-06-17
Consolidate Python SDK doc

MimiDumpling:
Consolidating relevant Python information into one page to help reader/user receive all the information they need in one spot.

PR #1112, opened 2019-07-12
fix: Restore information about supported Python versions

untitaker:
As part of https://github.com/getsentry/sentry-docs/pull/1066 this
information was lost.

### DIFF
--- a/src/collections/_documentation/platforms/python/index.md
+++ b/src/collections/_documentation/platforms/python/index.md
@@ -12,7 +12,7 @@ Install our Python SDK using `pip`:
 $ pip install --upgrade sentry-sdk
 ```
 
-The SDK provides support for Python 2.7 and 3.4 or later.
+The SDK provides support for Python 2.7 and 3.4 or later. [Integrations](#integrations) with specific frameworks (particularly asynchronous ones) may impose additional requirements though.
 
 {% include components/alert.html
   title="Upgrading the SDK and want to understand what's new?"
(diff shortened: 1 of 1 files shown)


vladlutkov:
But `sentry_sdk/integrations/aiohttp.py`, `sentry_sdk/integrations/sanic.py` and `sentry_sdk/integrations/tornado.py` have python 3.5+ keyword `async`

untitaker:
@vladlutkov certain integrations have stricter version requirements. aiohttp supports 3.5+ so this integration only works on 3.5+. When somebody doesn't use the integration the file is not imported.

untitaker:
We are using this SDK with Python 2.7 and Django 1.8 (previously 1.6) in production for sentry.io :) This works because we never import tornado/sanic/aiohttp related integrations

vladlutkov:
In the case of aiohttp yes it's true, but tornado supports python2.7.  Can you write restrictions for integrations?

untitaker:
Sorry, yes, for Tornado we additionally require a minimum version of Python 3.6 because of contextvars support.

Integrations have their own requirements documented at e.g. https://docs.sentry.io/platforms/python/tornado/, I will note this here.

vladlutkov:
Thanks

untitaker:
I think it's possible that the strict version requirements on Tornado are not necessary or easily avoidable. But it would probably require code changes.

untitaker:
re-requesting review just to make sure

MimiDumpling:
👍

PR #1121, opened 2019-07-19
fix: Remove quotes around github-app.id.

YtvwlD:
The application expects an integer, but if we put quotes there it's a string.


dashed:
@YtvwlD Thanks for contributing! 👍

PR #1122, opened 2019-07-19
Fix HttpTransport code snippet on PHP platform

RageZBla:
This should be merged only if https://github.com/getsentry/sentry-php/pull/855 is merged.


RageZBla:
Related to https://github.com/getsentry/sentry-php/issues/787

PR #1125, opened 2019-07-24
docs: fix fingerprint rpc example on Go platform

spuf:
`fingerprint-rpc-example` contains examples with `{{ default }}`.
`fingerprint-database-connection-example` contains examples without `{{ default }}`.

But Go platform example doesn't contain `{{ default }}` in any example.


dashed:
@spuf thanks for contributing! 👍 

@kamilogorek thanks for looking it over.

PR #1133, opened 2019-07-26
docs: Remove panic warning from goroutines docs

kamilogorek:
It has been fixed in the most recent version

### DIFF
--- a/src/collections/_documentation/platforms/go/goroutines.md
+++ b/src/collections/_documentation/platforms/go/goroutines.md
@@ -9,7 +9,7 @@ The easiest way to handle this, is to create a new `Hub` for every goroutine you
 
 Once cloned, `Hub` is completly isolated and can be used safely inside concurrent call. However, instead of using globally exposed methods, they should be called directly on the `Hub`.
 
-Here are two examples: one that is non-deterministic, would leak information between threads, and could trigger a concurrent-write panic; and one that is totally safe, and should be used instead.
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/platforms/go/goroutines.md
```diff
@@ -9,7 +9,7 @@ The easiest way to handle this, is to create a new `Hub` for every goroutine you
 
 Once cloned, `Hub` is completly isolated and can be used safely inside concurrent call. However, instead of using globally exposed methods, they should be called directly on the `Hub`.
 
-Here are two examples: one that is non-deterministic, would leak information between threads, and could trigger a concurrent-write panic; and one that is totally safe, and should be used instead.
+Here are two examples: one that is non-deterministic, would leak information between threads, and one that is totally safe, and should be used instead.
```
MimiDumpling:
```suggestion
Here are two examples: 
- a recommended deterministic call on `Hub` that is safe
- a discouraged non-deterministic call on `Hub` that would leak information between threads
```

MimiDumpling: I'm assuming one is deterministic and the other is non-deterministic. :) If that's the wrong assumption, try:
"Here are two examples:
 - a recommended call on `Hub` that is safe
 - a discouraged non-deterministic call on `Hub` that would leak information between threads"

kamilogorek: You are correct, thanks! :)

PR #1134, opened 2019-07-26
feat: Use new release checksums feature to provide SRI

kamilogorek:
ref: https://github.com/getsentry/sentry-javascript/issues/966

### DIFF
--- a/src/_plugins/release_registry.rb
+++ b/src/_plugins/release_registry.rb
@@ -76,8 +76,14 @@ module Jekyll
 
     def render(context)
       file = @params[2]
-      algorithm = @params[3] || "sha384-base64"
-      get_info["files"][file]["checksums"][algorithm]
+      algorithm = "sha384-base64"
+      checksum = get_info["files"][file]["checksums"][algorithm]
+      
+      if !checksum || checksum.empty?
+        raise RuntimeError.new("WARN: Failed to fetch checksum for file '#{file}' from '#{name}' '#{version}' sdk: #{e}")
+      end
+      
+      return "#{algorithm.split("-").first()}-#{checksum}"
     end
   end
 end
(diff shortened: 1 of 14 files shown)


src/_plugins/release_registry.rb
```diff
@@ -68,8 +68,21 @@ def render(context)
...
+
+  class SdkCdnChecksumTag < ReleaseRegistryTag
+    def entity
+      "sdks"
+    end
+
+    def render(context)
+      file = @params[2]
+      algorithm = @params[3] || "sha384-base64"
+      get_info["files"][file]["checksums"][algorithm]
```
tonyo: I'd suggest we raise a RuntimeError here when the checksum is empty or doesn't exist (for any reason).
Otherwise, we can end up with broken integrity attribute.

src/collections/_documentation/enriching-error-data/user-feedback-example/aspnetcore.md
```diff
@@ -2,7 +2,7 @@ With ASP.NET Core MVC, the `Error.cshtml` razor page:
 
 
 ```html
-<script src="https://browser.sentry-cdn.com/{% sdk_version sentry.javascript.browser %}/bundle.min.js" crossorigin="anonymous"></script>
+<script src="https://browser.sentry-cdn.com/{% sdk_version sentry.javascript.browser %}/bundle.min.js" integrity="sha384-{% sdk_cdn_checksum sentry.javascript.browser latest bundle.min.js %}" crossorigin="anonymous"></script>
```
tonyo: Could we perhaps move "sha384-" to the tag return value? It will be much easier to switch to other hash types in the future.

HazAT:
Good change overall. TBH I am not familiar with what benefits this exactly brings but what about the loader? Does the script we inject in the page also need to have this?

kamilogorek:
@HazAT it gives users confidence that we won't ever modify code without their consent once they include it on the page. Thus having this in the loader doesn't have much sense, as we control it anyway :p

PR #1149, opened 2019-08-09
ref: Use Application instance (this) as context

bruno-garcia:


bruno-garcia:
@cameronmcefee FYI

bruno-garcia:
Thanks @kamilogorek @ninniuz

PR #1151, opened 2019-08-09
docs: Add section about adblockers in JS documentation

kamilogorek:
Tried my best ¯\_(ツ)_/¯

### DIFF
--- a/src/collections/_documentation/platforms/javascript/index.md
+++ b/src/collections/_documentation/platforms/javascript/index.md
@@ -60,6 +60,54 @@ Then, you can see the error in your dashboard:
 
 [{% asset js-index/error-message.png alt="Error in Unresolved Issues with title This is my fake error message" %}]({% asset js-index/error-message.png @path %})
 
+{% capture __alert_content -%}
+When you are using our CDN, it may happen, that some people will use ad-blocking or script-blocking extensions. When this happens, our SDK won't be fetched and initialized properly. Because of this, any call to SDKs API will fail and may cause your application to misbehave. If this could apply to your use-case, read the section below on how to mitigate this issue.
+{%- endcapture -%}
+{%- include components/alert.html
+  title="Dealing with Ad-Blockers"
+  content=__alert_content
+  level="warning"
+%}
+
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -45,6 +45,54 @@ Then, you can see the error in your dashboard:
...
+The second way is to download the SDK from our CDN and host it yourself, this way SDK will still be separated from the rest of your code, but you'll be certain that it won't be blocked, as it's origin will be the same as the origin of your website.
+
+You can easily fetch it using `curl` or any other similar tool:
+
+```sh
+$ curl https://browser.sentry-cdn.com/{% sdk_version sentry.javascript.browser %}/bundle.min.js -o sentry.browser.{% sdk_version sentry.javascript.browser %}.min.js -s
+```
+
+The last option, is to use `Proxy` guard, that will make sure that your code won't break, even when you will call our SDK which was blocked.
+`Proxy` is supported by all browser except Internet Explorer, however there are no extensions in these browsers anyway. Also if `Proxy` is not in any of your user's browser, it will be silently skipped, so you don't have to worry about it breaking anything.
```
MimiDumpling:
```suggestion
`Proxy` is supported by all browsers except Internet Explorer. However, there are no extensions in these browsers anyway. Also, if `Proxy` is not in any of your users' browsers, it will be silently skipped, so you don't have to worry about it breaking anything.
```

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -45,6 +45,54 @@ Then, you can see the error in your dashboard:
...
+        if (typeof cb === "function") return cb(window.Sentry);
+        return window.Sentry;
+      }, handler);
+    }
+  };
+  window.Sentry = new Proxy({}, handler);
+}
+```
+
+And here is more useful, minified snippet, that you can copy-paste directly:
```
MimiDumpling:
```suggestion
And here is a more useful, minified snippet, that you can copy-paste directly:
```

kamilogorek:
Almost a year old PR. Still holds true though! :D

PR #1152, opened 2019-08-12
Change RNSentry import comment

sahil290791:
Current import comment states that
```
#import <React/RNSentry.h> // versions of react < 0.40 
```

instead it should be
```
#import <React/RNSentry.h> // versions of  react-native < 0.40
```


dashed:
@sahil290791 thanks! 👍

PR #1153, opened 2019-08-12
fix: Update go WrapWithExtra example typo

ff4f01:
These are two simple revision to the golang example code syntax in the sentry-go migration document.


dashed:
@ff4f01 good catch! 👍  thanks for fixing those typos.

PR #1159, opened 2019-08-15
Fix code snippet

uglide:
WaitForHandlerStart can be used only with asynchronous_start=true


dashed:
@mitsuhiko @jan-auer adding you folks as reviewers since I know that you know cpp.

PR #1163, opened 2019-08-16
Update sourcemaps.md

karltaylor:
Add information about using `Sentry.setRelease(value)` when uploading sourcemaps.

This had be baffled for quite a while why my sourcemaps were not linking up to the correct release. I thought it would be better if we explicitly mention this in the Docs 😊


dashed:
@karltaylor thanks for contributing! 👍

PR #1175, opened 2019-08-22
Rearrange auth token info in Integration Platform docs

MimiDumpling:
This fix rearranges some of the content for more clarity and distinct categorization of information.


src/collections/_documentation/workflow/integrations/integration-platform/index.md
```diff
@@ -69,6 +73,65 @@ Start your build by implementing the Redirect URL endpoint, /setup — typically
         return redirect('https://sentry.io/settings/')
 ```
 
+#### Auth Tokens
+
+Sentry's Integration Platform uses Auth Tokens, which are a similar concept to Access Tokens.
```
mnoble: Auth Tokens are Access Tokens. The former is just what we call them, the latter is the generalized name used when describing OAuth systems, but they are the same thing (opposed to similar concepts).

I'm not sure what to put here, tbh. Maybe just something saying that "Sentry uses header-based authentication" or something along those lines?

PR #1185, opened 2019-08-29
Soft wrap on pre tags

dcramer:


dcramer:
Might be controversial, but imo the user experience of wrapping things like this mega long script tag is better:

https://docs.sentry.io/platforms/javascript/

dcramer:
image

MimiDumpling:
I'm okay with this, but I'd like @cameronmcefee's opinion on this, too.

EvanPurkhiser:
This might be weird for other things. Have you looked at the java example? IMO what’s better is to hard break the actual example code

dcramer:
It's not amazing on Java https://www.dropbox.com/s/zd0sitskfi7rmbw/Screen-Shot-2019-08-29-10-55-19.15.png?dl=0

cameronmcefee:
Pre wrap styles always fix one problem and create ten more. You kinda just have to decide which bad solution you prefer.

If you want a higher-maintenance solution, there is a zero-width space character that can be used to imply logical breakpoints in long strings. However, this requires manual usage, is invisible, and will be copy/pastable which can break code. There's also the `wbr` html tag that serves the same purpose, though use of that within `pre` can be tricky.

PR #1198, opened 2019-09-05
build(config): Use proper configuration instaed of symlinks

BYK:
This patch replaces our hidden symlinks for bootstrap with the proper configuration so we can build the docs on Windows too.


BYK:
> but that logical reason may have just been not understanding how to do this

Oh, trust me when I say it took _way_ longer than I'd have hope for me to figure out.

PR #1213, opened 2019-09-10
feat: Add ignore options docs for JS SDK

HazAT:


src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -606,6 +606,68 @@ Sentry.init({
 });
 ```
 
+#### Decluttering Sentry
+
+The first thing to do is to consider constructing a whitelist of domains in which might raise acceptable exceptions.
```
MimiDumpling:
```suggestion
The first thing to do is consider constructing a whitelist of domains which might raise acceptable exceptions.
```

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -606,6 +606,68 @@ Sentry.init({
...
+
+```javascript
+Sentry.init({
+  whitelistUrls: [
+    /https?:\/\/((cdn|www)\.)?example\.com/
+  ]
+});
+```
+
+There is also `blacklistUrls` if you only want to always block specifc urls.
```
MimiDumpling:
```suggestion
There is also `blacklistUrls` if you want to block specific URLs forever.
```

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -606,6 +606,68 @@ Sentry.init({
...
+
+The community has compiled a list of common ignore rules for common things, like Facebook, Chrome extensions, etc. So it’s recommended to at least check these out and see if they apply to you. [Check out the original gist](https://gist.github.com/impressiver/5092952). This is not the default value of our SDK, it just a highlight of an extensive example what it could be.
```
MimiDumpling:
```suggestion
The community has compiled a list of common ignore rules for everyday things, like Facebook, Chrome extensions, etc. So it’s recommended to at least check these out and see if they apply to you. [Here is the original gist](https://gist.github.com/impressiver/5092952). This is not the default value of our SDK; it's just a highlight of an extensive example of what it could be.
```

HazAT:
Thank you @MimiDumpling :>

PR #1218, opened 2019-09-11
Update react.md

dperetti:
Fixed Typo in code `setExtras` -> `setExtra`

### DIFF
--- a/src/collections/_documentation/platforms/javascript/react.md
+++ b/src/collections/_documentation/platforms/javascript/react.md
@@ -51,7 +51,7 @@ class ExampleBoundary extends Component {
 
     componentDidCatch(error, errorInfo) {
       Sentry.withScope(scope => {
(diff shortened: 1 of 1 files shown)


dashed:
@dperetti Hey thanks for the PR! Unfortunately, this isn't a typo. `errorInfo` is an object as per https://reactjs.org/docs/react-component.html#componentdidcatch

`setExtras` is analogue of `setExtra`, which can be fed an object of keys mapped to an "extra". It's akin to performing multiple `setExtra` calls.

see: 

https://github.com/getsentry/sentry-javascript/blob/c38d25e9217e52178ecd11417c1777c8c03995da/packages/hub/src/scope.ts#L140 
and https://github.com/getsentry/sentry-javascript/blob/c38d25e9217e52178ecd11417c1777c8c03995da/packages/hub/src/scope.ts#L152

PR #1225, opened 2019-09-18
fix: Add note about complications in Django Channels

untitaker:


untitaker:
Adding matt as person who knows Django

PR #1226, opened 2019-09-19
fix: Change the goroutine example order

ff4f01:
The order of the two examples and their descriptions are not the same.
It would read better if the descriptions are provided right before each example.
Also fixing `a Goroutines` -> `a Goroutine`


kamilogorek:
Thanks!

PR #1230, opened 2019-09-20
Add mention of which plans we support

ceorourke:
Screenshot 2019-10-01 17 41 11


MimiDumpling:
@ceorourke can we condense this to one Note? :)

PR #1238, opened 2019-09-25
one at the time -> one at a time

dx-pbuckley:
I think `one at a time` is idiomatic vs `one at the time`. I also clarified that `one of them` is `one cron container`.

One thing I don't know yet (at this point in the docs - I just started reading) is what `beat processes` are? Are beat processes something that run in the worker node or possibly a typo?


markstory:
Thank you 👍

PR #1242, opened 2019-09-30
fix: Improve docs of send-default-pii

### LINKED ISSUE #1240: send_default_pii: what do the docs really recommend?
In the Python docs:

```Sending users to Sentry will unlock many features,```

I want to unlock those features! I want to do this for Django, so I go there and read:

```
If you use django.contrib.auth and have set send_default_pii=True in your call to init, user data (current user id, email address, username) is attached to the event.
```

Awesome, all I have to do is turn on `send_default_pii` to True and I'm good to go. But:

```Sentry also excludes personally identifiable information (such as user ids, usernames, cookies, authorization headers, IP addresses) unless you set send_default_pii to True.```

Wait. I don't want to send authorization headers and cookies!

I go on and find this in the configuration section:

https://docs.sentry.io/error-reporting/configuration/?platform=python#send-default-pii

```
send_default_pii

If this flag is enabled, certain personally identifiable information is added by active integrations. Without this flag they are never added to the event, to begin with. If possible, it’s recommended to turn on this feature and use the server side PII stripping to remove the values instead.
```

Okay, what's server side PII stripping? I assume it the same as server-side PII *scrubbing* - I recommend you use that term consistently in the docs and link to it.

https://docs.sentry.io/data-management/sensitive-data/

```
We recommend filtering or scrubbing sensitive data within the SDK, so that data is not sent with the event, and also configuring server-side scrubbing to ensure the data is not stored.
```

This seems wise; better not send the information in the first place. But wait? The SDK documentation just told me to send this data... And I want to unlock features?

So are you recommending we turn *on* `send_default_pii` or are you recommending we turn it *off* to ensure data is not sent?

I think that would be best is to *only* send the minimum information (which user) to unlock the Django features, but *not* to send any other authorization headers like password tokens along. I guess I'm off on my own for that, but it would be helpful if the documentation clearly told me so.

### PR DESCRIPTION
**Pull request by untitaker**
Fix #1240

### DIFF
--- a/src/collections/_documentation/error-reporting/configuration/index.md
+++ b/src/collections/_documentation/error-reporting/configuration/index.md
@@ -93,10 +93,9 @@ This feature is `off` by default.
 ### `send-default-pii`
 
 {% unsupported browser javascript node %}
-If this flag is enabled, certain personally identifiable information is added by active
-integrations.  Without this flag they are never added to the event, to begin with.  If possible,
-it's recommended to turn on this feature and use the server side PII stripping to remove the
-values instead.
+If this flag is enabled, certain personally identifiable information is added by active integrations. By default no such data it sent.
+
+If possible, it's recommended to turn on this feature to send all such data by default, and manually remove what you don't want to send using our features for managing [_Sensitive Data_]({%- link _documentation/data-management/sensitive-data.md -%}).
 {% endunsupported %}
 
 {:.config-key}
--- a/src/collections/_documentation/error-reporting/configuration/index.md
+++ b/src/collections/_documentation/error-reporting/configuration/index.md
@@ -93,7 +93,7 @@ This feature is `off` by default.
 ### `send-default-pii`
 
 {% unsupported browser javascript node %}
-If this flag is enabled, certain personally identifiable information is added by active integrations. By default no such data it sent.
+If this flag is enabled, certain personally identifiable information is added by active integrations. By default no such data is sent.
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by faassen** [comment]
I suggest linking to the section https://docs.sentry.io/data-management/sensitive-data/#custom-event-processing-in-the-sdk directly, as that's really the mechanism the user would need to employ to hook into the scrubbing process in the SDK itself. I would also recommend naming 'before-send' directly within the docs for `send-default-pii`, as that sets users directly on the right path.

**Comment by untitaker** [comment]
@faassen so depending on what your privacy policy contains, it may be entirely permissible to scrub data on the serverside only, which is easier to set up

File: src/collections/_documentation/error-reporting/configuration/index.md [hunk 1]
```diff
@@ -93,10 +93,9 @@ This feature is `off` by default.
 ### `send-default-pii`
 
 {% unsupported browser javascript node %}
-If this flag is enabled, certain personally identifiable information is added by active
-integrations.  Without this flag they are never added to the event, to begin with.  If possible,
-it's recommended to turn on this feature and use the server side PII stripping to remove the
-values instead.
+If this flag is enabled, certain personally identifiable information is added by active integrations. By default no such data it sent.
```
dashed:
```suggestion
If this flag is enabled, certain personally identifiable information is added by active integrations. By default no such data is sent.
```

untitaker:
@faassen I believe this structure makes sense.. "Sensitive Data" contains all links now, and you can discover that page from send-default-pii.

PR #1256, opened 2019-10-07
Fix example on how to check source map

rchl:
`new sourceMap.SourceMapConsumer` returns a promise in the latest version
of `source-map` so it needs to be `await`ed.

### DIFF
--- a/src/collections/_documentation/platforms/javascript/sourcemaps.md
+++ b/src/collections/_documentation/platforms/javascript/sourcemaps.md
@@ -404,25 +404,22 @@ var fs        = require('fs'),
     path      = require('path'),
     sourceMap = require('source-map');
 
-async function run() {
-    // file output by Webpack, Uglify, etc.
-    var GENERATED_FILE = path.join('.', 'app.min.js.map');
+// file output by Webpack, Uglify, etc.
+var GENERATED_FILE = path.join('.', 'app.min.js.map');
 
-    // line and column located in your generated file (e.g. source of your error
-    // from your minified file)
-    var GENERATED_LINE_AND_COLUMN = {line: 1, column: 1000};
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/platforms/javascript/sourcemaps.md
```diff
@@ -404,21 +404,25 @@ var fs        = require('fs'),
...
-// from your minified file)
-var GENERATED_LINE_AND_COLUMN = {line: 1, column: 1000};
+    // line and column located in your generated file (e.g. source of your error
+    // from your minified file)
+    var GENERATED_LINE_AND_COLUMN = {line: 1, column: 1000};
 
-var rawSourceMap = fs.readFileSync(GENERATED_FILE).toString();
-var smc = new sourceMap.SourceMapConsumer(rawSourceMap);
+    var rawSourceMap = fs.readFileSync(GENERATED_FILE).toString();
+    var smc = await new sourceMap.SourceMapConsumer(rawSourceMap);
```
dashed: @rchl can you make this example use promises instead of top-level await?

rchl: I can do that but technically speaking, this is not a top-level await. It's inside a function.

dashed: @rchl Ah you're right 👍 I've meant if the example can be conveyed without this wrapping function.

rchl: It can but IMO using `then` form of promise is less readable. Are you sure you want me to change that?

dashed:
@rchl thanks! Apologizes for not getting back to you sooner.

PR #1267, opened 2019-10-09
Updated path to podspec

tunegov:


HazAT:
These are the old docs, which are only accessible through a deep link.
In the new one this is already fixed: https://docs.sentry.io/platforms/react-native/cocoapods/

Thank you nevertheless!

PR #1281, opened 2019-10-17
feat: Add php max_request_body_size

HazAT:


src/collections/_documentation/platforms/php/index.md
```diff
@@ -269,6 +269,12 @@ The number of attempts that should be made to send an event before erroring
 and dropping it from the queue.
 By default this option is set to `6`.
 
+
+##### max_request_body_size
+
+This option represents the size limit in bytes beyond which the body of the request is not captured. Default value is `medium`. Possible values are `none` = never send body, `small`, `medium`, `alawys` = always send the body, no matter how large it is.
```
kamilogorek:
```suggestion
This option represents the size limit in bytes beyond which the body of the request is not captured. The default value is `medium`. Possible values are `none` = never send body, `small`, `medium`, `always` = always send the body, no matter how large it is.
```

src/collections/_documentation/platforms/php/index.md
```diff
@@ -269,6 +269,12 @@ The number of attempts that should be made to send an event before erroring
...
+The values are chosen like this on purpose since the SDK internally can define what the exact limits are.
```
kamilogorek: Still not sure what those values mean. Maybe add some ranges? or current values?

bruno-garcia: I think this is an attempt to allow us to change this without 'breaking' customers.
Today you'd need to read the code to see what we consider `small`, `medium` etc.
i.e: 

https://github.com/getsentry/sentry-dotnet/blob/46bf5548d347ad1e206aefa3ea91cc46ab02004f/src/Sentry/Extensibility/RequestBodyExtractionDispatcher.cs#L45

src/collections/_documentation/platforms/php/index.md
bruno-garcia: Isn't this part of the unified API?

src/collections/_documentation/platforms/php/index.md
HazAT: ^ this

src/collections/_documentation/platforms/php/index.md
HazAT: Yes, kinda, but as always we are only going to introduce this reactively to SDKs once ppl ask for it and not proactively push it out to all SDKs. 

For PHP it was there in the old SDK so we needed to bring it back.

kamilogorek:
Merging so it won't get stuck.

PR #1282, opened 2019-10-18
Update csharp.md

nbarbettini:
`await` inside the lambda body doesn't work unless it begins with `async`.


bruno-garcia:
Good catch! That wouldn't compile otherwise.
Thanks @nbarbettini

PR #1285, opened 2019-10-23
Testing git push

dkrito:
Added a dot then removed it so I could test git push on my machine.


dkrito:
Ran docker tests and it looks like it worked!

PR #1290, opened 2019-10-24
Adds steps for installing/importing sentry-go/http for panic recovery

inoda:
Small tweak - but I had to jump from https://docs.sentry.io/platforms/go/migration to the GitHub pkg readme to figure out how to get this imported!


kamilogorek:
Thanks!

PR #1293, opened 2019-10-24
feat(elixir): Add sentence about `before_send_event` returning `nil`

lobsterkatie:
https://github.com/getsentry/sentry-elixir/pull/364 (which fixes https://github.com/getsentry/sentry-elixir/issues/363) brings the Elixir SDK's `before_send_event` config option in line with other, similar config options (`beforeSend` and `before_send`) in the unified SDKs, in that it now recognizes a falsy return value as a signal to drop the event. This PR adds a sentence explaining that behavior to the Elixir config docs.

### DIFF
--- a/src/collections/_documentation/clients/elixir/config.md
+++ b/src/collections/_documentation/clients/elixir/config.md
@@ -106,7 +106,7 @@ The backend can also be configured to capture Logger metadata, which is detailed
 
 `before_send_event`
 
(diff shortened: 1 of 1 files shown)


lobsterkatie:
@mitchellhenke - can't tag you as a reviewer, but would appreciate you taking a glance anyway.

mitchellhenke:
👍 👍

PR #1294, opened 2019-10-24
ref(docs): Clarify which data structures we scrub

ceorourke:
Fixes https://app.asana.com/0/1144304343179777/1146130459036962

### DIFF
--- a/src/collections/_documentation/data-management/sensitive-data.md
+++ b/src/collections/_documentation/data-management/sensitive-data.md
@@ -40,7 +40,7 @@ Within your project settings you’ll find a **Data Scrubber** option. By defaul
     -   card[number]
 -   Any keys which match values that you’ve added to the list of additional fields in your Project Settings.
 
-Sentry will scrub arrays and literal values, but not entire objects. For example:
+Sentry will redact arrays and literal values, but not entire objects. For example:
 
 ```bash
 credentials = {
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/data-management/sensitive-data.md
```diff
@@ -40,6 +40,20 @@ Within your project settings you’ll find a **Data Scrubber** option. By defaul
     -   card[number]
 -   Any keys which match values that you’ve added to the list of additional fields in your Project Settings.
 
+Sentry will scrub arrays and literal values, but not entire objects. For example:
```
lobsterkatie:
```suggestion
Sentry will redact arrays and literal values, but not entire objects. For example:
```

(see note below)

src/collections/_documentation/data-management/sensitive-data.md
```diff
@@ -40,6 +40,20 @@ Within your project settings you’ll find a **Data Scrubber** option. By defaul
...
+credentials = {
+  username: 'jane',
+  cats: ['pancake', 'maple', 'hellboy']
+  password: 'p4ssw0rd!',
+  lastLogin: 'yesterday',
+  ...
+}
+```
+
+using the default options would _not_ cause the contents of `credentials` to be scrubbed, but `password` would be scrubbed, and adding `cats`, `username`, and/or `lastLogin` to the list of additional fields will cause those values to be scrubbed.
```
lobsterkatie:
```suggestion
using the default options would not cause the _object_ `credentials` to be redacted in its entirety; rather, all of its entries would be subject to scrubbing. So `password` would be redacted by default, and adding `cats`, `username`, and/or `lastLogin` to the list of additional fields would cause those values to be redacted as well. 
```
Two (related) reasons for this suggestion:
- "scrubbed" vs "redacted": trying to make the distinction between "things that are run through the data scrubber" and "things that actually get taken out" 
- "object `credentials`" vs "contents of `credentials`": we actually _do_ clobber some of the contents, so this could be misleading; related to the first reason because the point you're trying to make is, _we do run objects through the scrubber, but we never redact the object itself_.

(Also just changing "will" to "would" to match the beginning of the sentence.)

PR #1295, opened 2019-10-25
Remove fixed sdk version from composer require command in docs

JanMikes:
It is misleading and hard to maintain, often leads to outdated and incorrect information.

For example right now it shows 2.0.3 and there is already 2.0.4 released.

As well if value is omitted, it uses latest stable version (if not prefered dev version by user's composer.json).


untitaker:
FWIW this is automatically updated when deploying the docs. Which happens at least once a week.

JanMikes:
Oh i see 😄 

In case this is not real problem, feel free to close this issue and reject PR.

untitaker:
Cool. If you feel like the lag is too high then I believe we can talk about auto-deploying docs on sdk release. That's probably doable.

JanMikes:
> If you feel like the lag is too high then I believe we can talk about auto-deploying docs on sdk release

I think this is completely okay. I just myself prefer and use `composer require` command without explicitly specifying exact version of package.

PR #1304, opened 2019-10-29
add max tag length

ethanmdavidson:
fixes documentation problem raised in https://github.com/getsentry/sentry/issues/6576


untitaker:
thanks!

PR #1311, opened 2019-11-01
Add mention of using aws lambda node

ceorourke:
Screenshot 2019-10-31 17 16 05

### DIFF
--- a/src/collections/_documentation/platforms/node/index.md
+++ b/src/collections/_documentation/platforms/node/index.md
@@ -20,7 +20,7 @@ $ node --unhandled-rejections=warn app.js
 ```
 
 {% capture __alert_content -%}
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/platforms/node/index.md
```diff
@@ -19,6 +19,13 @@ The default configuration of the SDK captures unhandled promise rejections, whic
 $ node --unhandled-rejections=warn app.js
 ```
 
+AWS's Node runtime doesn't include a global error handler in the same way Node does, so the Node.js SDK has no way of catching unhandled exceptions. The only way to send events to Sentry is manually, by using `Sentry.captureException()` or `Sentry.captureMessage()`.
```
dashed:
```suggestion
AWS's Node runtime doesn't include a global error handler in the same way Node does, so the Node.js SDK has no way of catching unhandled exceptions. The only way to send events to Sentry is to manually use `Sentry.captureException()` or `Sentry.captureMessage()`.
```

PR #1321, opened 2019-11-05
Fix configuration array append syntax

j15e:
I think this is was should be written, or : 

```ruby
config.processors += [MyJobProcessor]
```


dashed:
@j15e thanks for opening this PR. I'll be closing this in favour of https://github.com/getsentry/sentry-docs/pull/1313; which is now merged.

PR #1336, opened 2019-11-07
Add get-organization-users API doc page

LearningNerd:
Same format as:
https://docs.sentry.io/api/organizations/get-organization-projects/

Note: also add docstring here?:
https://github.com/getsentry/sentry/blob/master/src/sentry/api/endpoints/organization_users.py

I plan to link to this in the docs when we publish my blog post about [this sample internal integration](https://github.com/getsentry/sentry-round-robin), which uses this endpoint to get the users who can actually be assigned to issues in a given project.

### DIFF
--- a/__tests__/__snapshots__/documentation.js.snap
+++ b/__tests__/__snapshots__/documentation.js.snap
@@ -93,6 +93,7 @@ Array [
   "api/organizations/get-organization-repositories/index.html",
   "api/organizations/get-organization-repository-commits/index.html",
   "api/organizations/get-organization-stats/index.html",
+  "api/organizations/get-organization-users/index.html",
   "api/organizations/get-short-id-lookup/index.html",
   "api/organizations/index.html",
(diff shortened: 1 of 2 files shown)


LearningNerd:
Hm @MimiDumpling thoughts on how I can fix this / why it's not passing these checks? And any context on the reviewing/approval process for adding a page to the docs? My understanding is the script to autogenerate API docs isn't working, and I was hoping to be able to link to this page from a new blog post that references the `organizations/users/` endpoint which was previously undocumented. Thanks!!

untitaker:
We have a test that asserts that only a certain set of pages is created. You need to run `bin/test -u` to update this set.

scefali:
@LearningNerd Looks good. Probably just need to restart the builds to get them to pass, however.

MimiDumpling:
@LearningNerd let me know how it goes with `bin/test -u` :D I'll be in the office all next week.

PR #1337, opened 2019-11-07
Add Javascript Example to Releases Documentation

### LINKED ISSUE #1335: Add JavaScript Example to Releases Documentation
**What**
The current documentation states:
```
How you make the version available to your code is up to you. For example, you could use an environment variable that is set during the build process.
```

**Why**
It is my belief that most people using JavaScript will be using Sentry inside of a npm/node project that has a package.json with a `"version"`. For this use-case I can imagine that many people would prefer to tag their releases with that `"version"` from the package.json and, there is an easy way for them to do this. [`process.env.npm_package_version`](https://docs.npmjs.com/misc/scripts#packagejson-vars)

**Solution**
https://github.com/getsentry/sentry-docs/pull/1337

Pharserror:
Resolves #1335

### DIFF
--- a/src/collections/_documentation/workflow/release-examples/javascript.md
+++ b/src/collections/_documentation/workflow/release-examples/javascript.md
@@ -1,5 +1,0 @@
-```javascript
-Sentry.init({
-  release: 'my-repo-name@' + process.env.npm_pacakage_version
-});
-```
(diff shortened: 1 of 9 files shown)


Pharserror:
I'm re-reading the docs I linked in the issue and, I believe I've misread them and misunderstood the availability of `process.env.npm_package_*` variables in code itself. There is some language being used here:
```
You can access these variables in your code with process.env.npm_package_name and process.env.npm_package_version, and so on for other fields.
```
But, I am getting the impression these variables are available to code being run via `npm run`.

PR #1344, opened 2019-11-11
ref(integration-platform): Use term Token or Auth Token

MeredithAnya:
We don't use the term Access Token anywhere except for the Integration Platform so it makes more sense to just call them tokens (or Auth Tokens) since that's what they are.

### DIFF
--- a/src/collections/_documentation/workflow/integrations/integration-platform/index.md
+++ b/src/collections/_documentation/workflow/integrations/integration-platform/index.md
@@ -321,7 +321,7 @@ Sentry Integration apps are similar to OAuth apps, except Sentry Integration app
 
 #### How is the Sentry Integration different from personal Auth Tokens?
 
-Personal Auth Tokens are tokens a user can use to invoke APIs directly and have access to all the resources tied to that user. Sentry Integration Apps do not represent a single user, but rather make requests as “itself."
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/workflow/integrations/integration-platform/index.md
```diff
@@ -319,9 +319,9 @@ OAuth Applications allow an external developer to create an application which ca
 
 Sentry Integration apps are similar to OAuth apps, except Sentry Integration apps act as an independent entity. There is currently no way for Sentry Integration apps to act on behalf of a user.
 
-#### How is the Sentry Integration different from Auth Tokens?
+#### How is the Sentry Integration different from personal Auth Tokens?
 
-Auth Tokens are personal access tokens a user can use to invoke APIs directly. Sentry Integration Apps do not represent a single user, but rather make requests as “itself."
+Personal Auth Tokens are tokens a user can use to invoke APIs directly and have access to all the resources tied to that user. Sentry Integration Apps do not represent a single user, but rather make requests as “itself."
```
MimiDumpling:
```suggestion
Personal Auth Tokens are tokens a user can use to invoke APIs directly and have access to all the resources tied to that user. Sentry Integration Apps do not represent a single user, but rather make requests as "itself."
```

PR #1348, opened 2019-11-13
docs(go): Add section about handling panics

kamilogorek:
Closes https://github.com/getsentry/sentry-go/issues/61

### DIFF
--- a/src/collections/_documentation/platforms/go/panics.md
+++ b/src/collections/_documentation/platforms/go/panics.md
@@ -3,12 +3,12 @@ title: Handling Panics
 sidebar_order: 3
 ---
 
-The way to capture unhandled panics in our Go SDK is through `Recover` method. It can be either used directly through `defer` keyword, or as the part of your own implementation.
+The way to capture unhandled panics in our Go SDK is through the `Recover` method. It can be either used directly through the `defer` keyword or as part of your implementation.
 
 ## Usage
 
-When used directly as shown below, Sentry will recover from the panic and internally decide whether to use `CaptureException` or `CaptureMessage` method, based on the type of input it received.
-As it's not uncommon to panic with a string, it's recommended to use `AttachStacktrace` option during SDK initialization, which will try to provide an useful stack trace for messages as well.
+When used directly, as shown below, Sentry will recover from the panic and internally decide whether to use the `CaptureException` or `CaptureMessage` method, based on the type of input it received.
+As it's not uncommon to panic with a string, it's recommended to use the `AttachStacktrace` option during SDK initialization, which will try to provide a useful stack trace for messages as well.
 
 ```go
 func() {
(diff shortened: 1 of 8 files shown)


src/collections/_documentation/platforms/go/panics.md
```diff
@@ -0,0 +1,74 @@
...
+title: Handling Panics
+sidebar_order: 3
+---
+
+The way to capture unhandled panics in our Go SDK is through `Recover` method. It can be either used directly through `defer` keyword, or as the part of your own implementation.
+
+## Usage
+
+When used directly as shown below, Sentry will recover from the panic and internally decide whether to use `CaptureException` or `CaptureMessage` method, based on the type of input it received.
+As it's not uncommon to panic with a string, it's recommended to use `AttachStacktrace` option during SDK initialization, which will try to provide an useful stack trace for messages as well.
```
MimiDumpling:
```suggestion
As it's not uncommon to panic with a string, it's recommended to use the `AttachStacktrace` option during SDK initialization, which will try to provide a useful stack trace for messages as well.
```

src/collections/_documentation/platforms/go/panics.md
```diff
@@ -0,0 +1,74 @@
...
+
+```go
+func() {
+	defer sentry.Recover()
+	// do all of the scary things here
+}()
+```
+
+By default, Sentry Go SDK uses asynchronous transport, which in the code example below requires an explicit awaiting for event delivery to be finished using `sentry.Flush` method. It is necessary, because otherwise the program would not wait for the async HTTP calls to return a response, and exit the process immediately when it reached the end of the `main` function. It would not be required inside a running goroutine or if you would use `HTTPSyncTransport`, which you can read about in `Transports` section.
```
MimiDumpling:
```suggestion
By default, Sentry Go SDK uses asynchronous transport, which in the code example below requires an explicit awaiting for event delivery to be finished using `sentry.Flush` method. It is necessary, because otherwise the program would not wait for the async HTTP calls to return a response, and exit the process immediately when it reached the end of the `main` function. It would not be required inside a running goroutine or if you would use `HTTPSyncTransport`, which you can read about in the `Transports` section.
```

src/collections/_documentation/platforms/go/panics.md
```diff
@@ -0,0 +1,74 @@
...
+
+If you want to control the delivery for a single `defer` call, or do some other things before capturing, you have to use `Recovery` method on the `Hub` instance directly, as it can accept `err` itself.
```
MimiDumpling:
```suggestion
If you want to control the delivery for a single `defer` call, or do some other things before capturing, you have to use the `Recovery` method on the `Hub` instance directly, as it can accept `err` itself.
```

src/collections/_documentation/platforms/go/panics.md
```diff
@@ -0,0 +1,74 @@
...
+		}
+	}()
+
+	// do all of the scary things here
+}()
+```
+
+### Using Context
+
+Besides regular `Recover` method, there's one more that can be used for panics, namely `RecoverWithContext`.
```
MimiDumpling:
```suggestion
Besides the regular `Recover` method, there's one more that can be used for panics, namely `RecoverWithContext`.
```

src/collections/_documentation/platforms/go/panics.md
```diff
@@ -0,0 +1,74 @@
...
+It allows for passing an instance of `context.Context` as the first argument. This gives us two additional features.
+
+The first one being extracting the `Hub` instance from the context and using it instead of the global one - this is used for every http/server package integration, as it allows for execution context separation. You can see it in action in our `http` integration [source code](https://github.com/getsentry/sentry-go/blob/383614eaf2e038cf3a6d2022c56fb206589efe11/http/sentryhttp.go#L50-L91).
+
+And the second feature, an access to the `context.Context` itself inside `beforeSend` method, which can be used to extract any additional information about what happened during the panic:
```
MimiDumpling:
```suggestion
And the second feature, an access to the `context.Context` itself inside the `beforeSend` method, which can be used to extract any additional information about what happened during the panic:
```

PR #1350, opened 2019-11-13
[sentry-go] sentry.NewEvent already returns a pointer

fungiboletus:


kamilogorek:
Thanks! I forgot to change it from when I used `&sentry.Event{}` 😅

PR #1351, opened 2019-11-13
docs(node): Update unhandled rejection handler options

kamilogorek:
Introduced in the latest release.

### DIFF
--- a/src/collections/_documentation/platforms/node/default-integrations.md
+++ b/src/collections/_documentation/platforms/node/default-integrations.md
@@ -59,7 +59,7 @@ Available options:
 
 _Import name: `Sentry.Integrations.OnUnhandledRejection`_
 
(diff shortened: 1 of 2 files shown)


src/collections/_documentation/platforms/node/default-integrations.md
```diff
@@ -59,21 +59,17 @@ Available options:
...
-{%- endcapture -%}
-{%- include components/alert.html
-    title="Note"
-    content=__alert_content
-    level="info"
-%}
-
-```bash
-$ node --unhandled-rejections=warn app.js
+This integration attaches global unhandled rejection handlers. By default all unhandled rejections trigger a warning and log the error. You can change this behavior using `mode` option, which is in line with Node's CLI options: https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode
```
MimiDumpling:
```suggestion
This integration attaches global unhandled rejection handlers. By default, all unhandled rejections trigger a warning and log the error. You can change this behavior using the `mode` option, which is in line with Node's CLI options: https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode
```

PR #1359, opened 2019-11-21
Fix `node_modules` path to new @sentry packages

jeremywiebe:
I believe these need to be changed to match the new npm `@sentry` scope used by all Sentry packages.


HazAT:
These are the old react-native docs, they shouldn't be indexed by google anymore.
In the new docs this is already correct.

Thank you though!

PR #1360, opened 2019-11-21
Create new doc for Android SDK

MimiDumpling:
Sentry has a new Android SDK in alpha. These docs will help users get started with the alpha version.

image


src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
+// ADD SENTRY ANDROID AS A DEPENDENCY
+dependencies {
+    implementation 'io.sentry:sentry-android:{version}'
+}
+```
+
+For other dependency managers, see the [central Maven repository](https://search.maven.org/artifact/io.sentry/sentry-android/1.7.28/jar).
```
bruno-garcia: I don't understand what this means. Also, is this supposed to point to 1.7?

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+import android.os.Bundle;
+import androidx.appcompat.app.AppCompatActivity;
+import io.sentry.core.Sentry;
+
+public class MyActivity extends AppCompatActivity {
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        try{
+            throw new Exception("This is a test.");
+        }catch(Exception e){
```
bruno-garcia:
```suggestion
        } catch (Exception e) {
```

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+```java
+import android.app.Application;
+import io.sentry.android.core.SentryAndroid;
+import io.sentry.core.SentryLevel;
+
+public class SentryApplication extends Application {
+    public void onCreate() {
+        super.onCreate();
+        /**
+            * Manual Initialisation of the Sentry Android SDK
```
bruno-garcia:
```suggestion
            * Manual Initialization of the Sentry Android SDK
```

I suggest doing a find and replace for this so it's consistent.

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+            * @Context - Instance of the Android Context
+            * @Options - Call back function that you need to provide to be able to modify the options.
+            * The call back function is provided with the options loaded from the manifest.
+            *
+            */
+        SentryAndroid.init(this, options -> {
+            // Add a callback that will be used before the event is sent to Sentry.
+            // With this callback, you can modify the event or, when returning null, also discard the event.
+            options.setBeforeSend((event, hint) -> {
+                if (event.getLevel().equals(SentryLevel.DEBUG))
```
bruno-garcia:
```suggestion
                if (SentryLevel.DEBUG.equals(event.getLevel()))
```

NPE

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+
+For more details about the available configurations, see [Configuration options](#configuration-options).
+
+## Capturing Errors
+
+### Crash Reporting
+
+Your application will crash whenever a thrown exception goes uncaught. The Sentry SDK catches the exception right before the crash and builds a crash report that will persist to the disk. The SDK will try to send the report right after the crash, but since the environment may be unstable at the crash time, the report is guaranteed to send once the application is started again.
+
+If there is a fatal error in your native code, the process is the same. The difference is, a restart is always needed, as a hard crash immediately stops the app.
```
bruno-garcia: This is not accurate. On my tests we consistently sent the event before the app crashed.
so for native crashes we might send before restart

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+        * An Example of how to report an error and how to enrich the error context.
+        */
+    void logWithStaticAPI() {
+    /*
+    Record a breadcrumb in the current context, which will be sent
+    with the next event(s). By default, the last 100 breadcrumbs are kept.
+    */
+        Breadcrumb breadcrumb = new Breadcrumb();
+        breadcrumb.setMessage("User made an action.");
+        Sentry.addBreadcrumb(breadcrumb);
```
bruno-garcia: This a bit too verbose. We should have an overload that takes a string only

```suggestion
        Sentry.addBreadcrumb("the user performed an action");
```

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+
+        // Set the user in the current context.
+        Sentry.configureScope(
+        scope -> {
+            User user = new User();
+            user.setEmail("hello@sentry.io");
+                    scope.setUser(user);
```
bruno-garcia: Indentation is off

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+        });
+
+    /*
+    This sends a simple event to Sentry, with a simple String message.
+    */
+
+        Sentry.captureMessage("This is a test");
+
+        try {
+        unsafeMethod();
```
bruno-garcia:
```suggestion
            unsafeMethod();
```

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+
+```groovy
+buildscript {
+    dependencies {
+        classpath 'io.sentry:sentry-android-gradle-plugin:1.7.28'
+    }
+}
+```
+
+The plugin will automatically generate appropriate ProGuard mapping files and upload them when you run `gradle` `assembleSomething`. For example, `assembleRelease` — Release is the default, but the plugin works for others if you have enabled ProGuard. The credentials for the upload step are loaded via environment variables or from a `sentry.properties` file in your project root. The `sentry.properties` in your project root that configures `sentry-cli` is different than the one you include in your application resources to configure the Sentry SDK at runtime — as seen in the [Configuration documentation]({%- link _documentation/clients/java/config.md -%}).
```
bruno-garcia: We don't support properties file in the new sdk. It's only relevant for the gradle plugin now

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+
+If you want to use the SDK with the NDK but you still want to support the devices on the API level lower than 21 you can update your manifest as follows:
+
+```xml
+    <!-- Merging strategy for the imported manifests -->
+      <uses-sdk
+            tools:overrideLibrary="io.sentry.android, io.sentry.android.ndk"/>
+```
+
+With these changes, the NDK is going to be used only on the devices with the API level ≥ 21 and the rest of the devices with API level ≥14, but ≤21 will use just the SDK.
```
bruno-garcia:
```suggestion
With these changes, NDK is going to be used only on the devices with the API level ≥ 21 and the rest of the devices with API level ≥14, but ≤21 will use just the SDK.
```

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+```java
+Breadcrumb breadcrumb = new Breadcrumb();
+breadcrumb.setMessage("Https Call to Sentry");
+
+Map<String,String> data = new HashMap();
+data.put("url","https:\\sentry.io");
+
+breadcrumb.setData(data);
+
+Sentry.addBreadcrumb(breadcrumb);
```
bruno-garcia: Here an overload that takes the message and data would be nice too would become 3 lines of code

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+- [Full documentation on Breadcrumbs](https://docs.sentry.io/enriching-error-data/breadcrumbs/)
+- [Debug Issues Faster with Breadcrumbs](https://blog.sentry.io/2016/05/04/breadcrumbs).
+
+### Configuration options
+
+**NDK Integration**
+
+NDK is packed with the SDK and enabled by default with each start of the SDK. 
+
+Alternatively, you can disable the NDK by adding the following line into your manifest.
```
bruno-garcia: I'd say tis either 'the NDK integration' or just 'NDK'
But we are not enabling ndk. We are enabling out ndk integration. Maybe we reword a bit?

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+
+```xml
+<meta-data android:name="io.sentry.ndk.enable" android:value="false" />
+```
+
+**ANR** 
+
+Whenever the main UI thread of the application is blocked for more than four seconds, the SDK will report the problem to the server.  (Sentry does not report the ANR if the application is in debug mode.)
```
bruno-garcia:
```suggestion
Whenever the main UI thread of the application is blocked for more than four seconds, the SDK will report the problem to the server.  (Sentry does not report ANR if the application is in debug mode.)
```

bruno-garcia: I'd also suggest to write 'anr (application no responding)' the first time it shows up

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+If you want to specify how long the thread should be blocked before the ANR is reported, provide the duration in the attribute `io.sentry.anr.timeout-interval-mills` in your manifest.
+
+```xml
+<meta-data android:name="io.sentry.anr.timeout-interval-mills" android:value="2000" />
+```
+
+**“In Application” Stack Frames**
+
+Sentry differentiates stack frames that are directly related to your application (“in application”) from stack frames that come from other packages such as the standard library, frameworks, or other dependencies. The difference is visible in the Sentry web interface, where only the “in application” frames are displayed by default.
+
```
bruno-garcia: Worth mentioning:

We automatically mark your applications package as inApp and everything else as not in app.

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+The callback can be registered during the initialization of the SDK. 
+
+```java
+SentryAndroid.init(this,options -> {
+
+        // Add a callback that will be used before the event is sent to Sentry.
+        // With this callback, you can modify the event or, when returning null, also discard the event.
+        options.setBeforeSend((event, hint) -> {
+            String environment = event.getEnvironment();
+            if (environment == null | environment.equals("TEST"))
```
bruno-garcia:
```suggestion
            if (environment == null || environment.equals("TEST"))
```

src/collections/_documentation/platforms/android/index.md
MimiDumpling: ^ @marandaneto @fdrabek

src/collections/_documentation/platforms/android/index.md
MimiDumpling: @bruno-garcia How about:

If there is a fatal error in your native code, the process is similar. The crash report will send before the app crashes.

src/collections/_documentation/platforms/android/index.md
MimiDumpling: @bruno-garcia are you suggesting we change this code snippet? If so, will you make a suggested edit?

src/collections/_documentation/platforms/android/index.md
MimiDumpling: @bruno-garcia how about this:

The NDK integration is packed with the SDK and enabled out by default with each start of the SDK. 

Alternatively, you can debilitate the NDK integration by adding the following line into your manifest.

src/collections/_documentation/platforms/android/index.md
MimiDumpling: Excellent suggestion. 😄

src/collections/_documentation/platforms/android/index.md
fdrabek: I think that it should point to https://search.maven.org/artifact/io.sentry/sentry-android 
@bruno-garcia I think that it means that if you want to see syntax for Kotlin DSL, Bazel etc.. you should check it there.
Does it make sense?

bruno-garcia: You want to link to the maven repositories we push the sdk? Android devs use more jcenter I believe at right now we had issues with maven central so I'm not sure the latest is going there /cc @marandaneto

marandaneto: I'd just delete it till we get this published to maven central that has a nice way of showing how to integrate things using other DSLs.
let's focus on bintray/jcenter only right now.

bruno-garcia: Oh I see what you mean. I wouldn't leave this in. It's OK to just have the groovy one for now. We can add the Lang selector later and show it here.

I suggest we remove this bit

src/collections/_documentation/platforms/android/index.md
bruno-garcia: I'd just use might instead of will.
we will for sure send on restart but we'll attempt to sent it before the app goes away

marandaneto: for fatal errors on native-code, I've always needed to restart it, it was never time enough to send it over.

src/collections/_documentation/platforms/android/index.md
bruno-garcia: @marandaneto I believe you added the overload already?

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+    This sends a simple event to Sentry, with a simple String message.
+    */
+
+        Sentry.captureMessage("This is a test");
+
+        try {
+            unsafeMethod();
+        } catch (Exception e) {
+        // This sends an error report to the  Sentry
+        Sentry.captureException(e);
```
marandaneto: missing a tab here as `Sentry.captureException(e);` is inside of the `catch` block

src/collections/_documentation/platforms/android/index.md
MimiDumpling: Got it. Will remove.

src/collections/_documentation/platforms/android/index.md
bruno-garcia: Looks good. One thing: Debilitante - > disable.

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+```
+
+And declare a dependency in your top-level `build.gradle`:
+
+```groovy
+buildscript {
+    dependencies {
+        classpath 'io.sentry:sentry-android-gradle-plugin:1.7.28'
+    }
+}
```
marandaneto: missing:
```
    repositories {
        jcenter()
    }
```

marandaneto: if somebody only copies it over, the plugin won't be found.

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+```
+
+The plugin will automatically generate appropriate ProGuard mapping files and upload them when you run `gradle` `assembleSomething`. For example, `assembleRelease` — Release is the default, but the plugin works for others if you have enabled ProGuard. The credentials for the upload step are loaded via environment variables.
```
marandaneto: what's about `assemble{BuildVariant}` instead of `assembleSomething`?

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+### Tagging Events
+
+Tags are key/value pairs assigned to events that can be used for breaking down issues or quick access to finding related events.
+
+Most SDKs generally support configuring tags by configuring the scope:
+
+```java
+Sentry.configureScope(
+    scope -> {
+    scope.setTag("tagKey", "tagValue");
```
marandaneto: I'd give a tab on this line as its inside of the lambda block

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+For more information, see [Tagging Events]({%- link _documentation/enriching-error-data/context.md -%}#tagging-events) in Context.
+
+### Setting the Level
+
+You can set the severity of an event to one of five values: `fatal`, `error`, `warning`, `info`, and `debug`. `error` is the default, `fatal` is the most severe, and `debug` is the least severe.
+
+```java
+Sentry.configureScope(
+    scope -> {
+    scope.setLevel(SentryLevel.WARNING);
```
marandaneto: same here

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,597 @@
...
+        implementation 'io.sentry:sentry-android-core::2.0.+'
+}
+```
+
+If you want to use the SDK with the NDK but you still want to support the devices on the API level lower than 21 you can update your manifest as follows:
+
+```xml
+    <!-- Merging strategy for the imported manifests -->
+      <uses-sdk
+            tools:overrideLibrary="io.sentry.android, io.sentry.android.ndk"/>
```
marandaneto: `tools:overrideLibrary="io.sentry.android"/>` only, I've removed the necessary of `io.sentry.android.ndk`, will be available on version alpha08

src/collections/_documentation/platforms/android/index.md
marandaneto: that would need a change in our source code, we won't be able till tomorrow, let's keep as it is for now.

src/collections/_documentation/platforms/android/index.md
MimiDumpling: @marandaneto which is correct, 1 or 2:
1.  `gradle` `assemble{BuildVariant}`
2. `gradle assemble{BuildVariant}`

src/collections/_documentation/platforms/android/index.md
MimiDumpling: @marandaneto where does this go? What line? 😸 You can make it a suggested edit.

src/collections/_documentation/platforms/android/index.md
MimiDumpling: Okay. No changes here.

src/collections/_documentation/platforms/android/index.md
```diff
@@ -0,0 +1,599 @@
...
+        * An Example of how to report an error and how to enrich the error context.
+        */
+    void logWithStaticAPI() {
+    /*
+    Record a breadcrumb in the current context, which will be sent
+    with the next event(s). By default, the last 100 breadcrumbs are kept.
+    */
+        Breadcrumb breadcrumb = new Breadcrumb();
+        breadcrumb.setMessage("User made an action.");
+        Sentry.addBreadcrumb("the user performed an action");
```
bruno-garcia:
```suggestion
        Sentry.addBreadcrumb(breadcrumb);
```

PR #1361, opened 2019-11-21
Link to Round Robin integration on Internal Integrations page

LearningNerd:
Tiny update :) Added a link to:
https://docs.sentry.io/workflow/integrations/integration-platform/#internal-integrations

The link points to this blog post and the repo for its sample code:
https://blog.sentry.io/2019/11/21/customize-your-sentry-workflow-a-sample-internal-integration 

@MimiDumpling Is there an existing convention for example code links / related blog posts? For example, a note/alert box thing like the one at the top of [the Node platform docs page](https://docs.sentry.io/platforms/node/)? Or italic text? Or leave it as is?

I also wasn't sure where best to place the link... below the diagram about not needing OAuth? Or immediately below the "Internal Integrations" heading? Or at the very top of the Integration Platform page? Hrmm.


manuzope:
I like where you've placed it and like the language.

I'll let @MimiDumpling decide on the alert box. I do agree that it would be better with some different formatting because it's tangential to what is being discussed.

PR #1363, opened 2019-11-22
Add link from Integrations to Integration Platform

LearningNerd:
Updates https://docs.sentry.io/workflow/integrations/
  - Adds a heading at the top for Integrations Platform
  - ...with the first paragraph straight from https://docs.sentry.io/workflow/integrations/integration-platform/

### DIFF
--- a/src/collections/_documentation/workflow/integrations/index.md
+++ b/src/collections/_documentation/workflow/integrations/index.md
@@ -5,6 +5,10 @@ sidebar_order: 6
 
 Sentry integrates seamlessly with your favorite apps and services.
 
+### Integration Platform
+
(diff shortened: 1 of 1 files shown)


MeredithAnya:
@LearningNerd should we be putting the integration platform at the bottom since that's where it shows up in the sidebar?

MimiDumpling:
> @LearningNerd should we be putting the integration platform at the bottom since that's where it shows up in the sidebar?

Yeah, I agree with Meredith. :)

LearningNerd:
Alrighty moved it to the bottom of the page @MimiDumpling @MeredithAnya :) But maybe it _should_ be at the top / and top of the sidebar because the Integration Platform is the new hotness? Or find another way to highlight it. hmm. thoughts?

MeredithAnya:
@LearningNerd My feeling is that more people are going to care about knowing whether the integration they want is supported, and then after that might care about writing something custom, since the Integration Platform is geared towards developers not only using the integrations but actually building them. Not sure if @MimiDumpling has other thoughts

MimiDumpling:
@LearningNerd @MeredithAnya Yeah, when prioritizing order, the path of least resistance (in this case, the pre-built integrations) are highlighted before the path of customizability (in this case, the integration platform). UNLESS, someone in leadership explicitly tells us to highlight/re-prioritize something. 😸 Thank youuuu! I think it's ready to merge.

PR #1375, opened 2019-11-27
fix(test): document.output is already an HTML doc

BYK:
I'm guessing in older versions of Jekyll, `document.output` was a string but currently it is already a `Nokogiri::HTML` instance which broke my local builds. This PR removes the obsolete and problematic logic


BYK:
Okay, I give up. Just gonna assume my env is broken.

PR #1377, opened 2019-11-27
removes empty "h2"

fetzig:


MimiDumpling:
@fetzig Nice catch! Thanks for helping with Docs. 🎉

PR #1383, opened 2019-12-02
Updated self-hosted index page

MimiDumpling:
The Self-Hosted install page was recently updated. I went through and cleaned up some spelling/grammar/language.

### DIFF
--- a/src/collections/_documentation/server/installation/index.md
+++ b/src/collections/_documentation/server/installation/index.md
@@ -9,28 +9,28 @@ Sentry relies on multiple services to work, which are all orchestrated by [Docke
 
 -   [Docker](https://www.docker.com/) 17.05.0+
 -   [Docker Compose](https://docs.docker.com/compose/) 1.19.0+
--   A dedicated (sub)domain to host Sentry on (i.e. _sentry.yourcompany.com_).
+-   A dedicated (sub)domain to host Sentry on (for example, _sentry.yourcompany.com_).
 -   At least 2400MB memory
 -   2 CPU Cores
 
 ## Installing Sentry Server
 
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/server/installation/index.md
```diff
@@ -3,34 +3,34 @@ title: Installation
...
+Before starting the installation, we strongly recommend you to check out [how to configure your Sentry instance]({%- link _documentation/server/config.md -%}) as you'd need to rebuild your images (`docker-compose build`) if you ever want to change your configuration settings. You may copy and edit the example configs provided in the repository. If none exists, the install script will use these examples as the actual configurations.
 
-To start, just run the install script:
+To start, run the install script:
 ```bash
 ./install.sh
 ```
 
-If the `CI` environment variable is set to a non-empty value, the script will assume unattended installation and will skip some steps. You can look at [its source code](https://github.com/getsentry/onpremise/blob/master/install.sh) to figure out what those steps are, which will also be logged to standard output.
+If you've set the `CI` environment variable to a non-empty value, the script will assume unintended installation and skip some steps. You can look at [its source code](https://github.com/getsentry/onpremise/blob/master/install.sh) to figure out what those steps are, which will also be logged to standard output.
```
BYK: Let's revert the changes on this paragraph as it changes the meaning. The original one was accurate:

1. The `CI` environment variable is usually set by continuous-integration services such as TravisCI. We don't expect users to set this manually as the new wording suggests (nor we want to encourage this).
2. The script will assume _unattended_ installation, meaning this is fully automated and no keyboard input will be expected (and probably won't be available anyway).

MimiDumpling: @BYK what if we said, "If your continuous-integration service sets the `CI` environment variable to a non-empty value, the script will assume unattended installation and will skip some steps."

BYK: As discussed over Slack, we want to keep the dual meaning here: 
1. This variable is usually automatically set by services like TravisCI, so we want to inform you what will happen in that case
2. We also want to highlight the possibility of using `CI=1 ./install.sh` for an on-purpose unattended install locally or on a different environment.

MimiDumpling: Yup yup! You right. 😄

PR #1389, opened 2019-12-05
Update JS index page

MimiDumpling:
Went through the JS index page and removed any sections that feel "too noisy." The goal was to trim the page a bit because it's so massive. Let me know if any topics need to stay instead of being removed.


kamilogorek:
I'm fine with removing all marked sections. Only not sure about UglifyJS removal. WDYT @HazAT ?

MimiDumpling:
@HazAT have you had a chance to review?

HazAT:
Good job, this removes a lot of noise!

PR #1392, opened 2019-12-06
Add shouldSendEvent snippet in Cocoa docs

ceorourke:
Screenshot 2019-12-05 16 10 30

### DIFF
--- a/src/collections/_documentation/clients/cocoa/advanced.md
+++ b/src/collections/_documentation/clients/cocoa/advanced.md
@@ -163,7 +163,7 @@ Client.shared?.shouldSendEvent = { (event) in
 Here's an example in Objective C too:
 
 ```objc
-self.shouldSendEvent = ^BOOL(SentryEvent *_Nonnull event) {
+client.shouldSendEvent = ^BOOL(SentryEvent *_Nonnull event) {
     return (sampleRate >= ((double)arc4random() / 0x100000000));
 };
 ```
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/clients/cocoa/advanced.md
```diff
@@ -147,6 +147,16 @@ Client.shared?.snapshotStacktrace {
...
 
+## Determining whether to send events
+
+If you want to filter certain events, you can use `shouldSendEvent` which takes the event, returns a boolean, and decides whether or not to send the event.
+
+```swift
+    self.shouldSendEvent = ^BOOL(SentryEvent *_Nonnull event) {
+        return (sampleRate >= ((double)arc4random() / 0x100000000));
+    };
+```
```
HazAT: So this code is objective c and since most code samples on this page are swift we should also have a swift example here.

```swift
        Client.shared?.shouldSendEvent = { (event) in
            if (event.environment == "staging") {
                return false
            }
            return true
        }
```

src/collections/_documentation/clients/cocoa/advanced.md
```diff
@@ -147,6 +147,27 @@ Client.shared?.snapshotStacktrace {
...
+                return false
+            }
+            return true
+        }
+```
+
+Here's an example in Objective C too:
+
+```objc
+self.shouldSendEvent = ^BOOL(SentryEvent *_Nonnull event) {
```
HazAT:
```suggestion
client.shouldSendEvent = ^BOOL(SentryEvent *_Nonnull event) {
```

Issue #1393, opened 2019-12-06
maxBreadcrumbs was not defaults 100


77xi:
[Docs](https://github.com/getsentry/sentry-docs/blob/master/src/collections/_documentation/error-reporting/configuration/index.md):
- This variable controls the total amount of breadcrumbs that should be captured. This defaults to 100.

Problem:
- maxBreadcrumbs was not defaults 100，relate issue https://github.com/getsentry/sentry-javascript/pull/1295/commits showing that was defaults 30 ？

- my env was react-native，test result was defaults 30 too。

PR #1394, opened 2019-12-06
Update maxBreadcrumbs defaults 30

77xi:
fix: https://github.com/getsentry/sentry-docs/issues/1393


untitaker:
This value is not 30 for python. This documentation is applied for python and js though

77xi:
Add some explanation is necessary ，you can do it for yourself。

PR #1395, opened 2019-12-06
fix: Add noindex for docs that are already updated

HazAT:
This adds 
```html
 <meta name="robots" content="noindex">
```
to all old doc pages that already have been updated so google doesn't index them and ppl are not misguided.

Disclaimer: Only added so many ppl for visibility and raising awareness we should do this in the future for all relevant pages.

cc @stayallive

### DIFF
--- a/src/collections/_documentation/clients/csharp/index.md
+++ b/src/collections/_documentation/clients/csharp/index.md
@@ -1,5 +1,6 @@
 ---
 title: 'C#'
+robots: noindex
 ---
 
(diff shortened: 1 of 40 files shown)


cameronmcefee:
FYI, you can set this as a default for all docs in the `clients` folder if you like: https://github.com/getsentry/sentry-docs/blob/master/_config.yml#L50

HazAT:
@cameronmcefee Thanks yeah, thought about that but we don't want it on all `clients` since some of the SDKs aren't updated yet and some `clients` docs are still valid.

PR #1396, opened 2019-12-08
added destructuring-assignment.

divyanshu-rawat:
Replaced current `hasError` syntax with destructuring-assignment.


dashed:
I have no strong opinions on this. I'll let @kamilogorek make a decision.

kamilogorek:
I don't think a deconstruction for a single variable is worth it. It doesn't help readability in this case, so I'd prefer to stick with a regular dot-access. Thanks for the PR nonetheless 🙏

PR #1398, opened 2019-12-09
fix(self-hosted): Use https:// link to clone the onpremise repository

ProgVal:
Instead of ssh link, which requires a Github account and
having a public key configured


untitaker:
Thanks!

PR #1406, opened 2019-12-11
Guide - Grouping and Fingerprints

thinkocapo:
## Overview
This is a Guide. It has less info than https://docs.sentry.io/data-management/event-grouping/ but it's a more common set of practices that helps the user understand Grouping & Fingerprints faster.

I covered Server-Side Fingerprinting in this Guide.

This whole Guide was inspired by an account in the sales pipeline who requested a lot of assistance for merging some issues. I asked for examples, and they sent me Exceptions that looked like they needed to be grouped by `error.type` or `error.message` hence that's what my screenshots in this guide cover. I felt this was a good use case that a lot of people can connect with.

I hope the distinction is clear between sections **Merging Similar Issues** which means updating the fingerprints of historical issues already in your Issues Stream, and **Server-Side Fingerprinting** which means setting rules/config that will actually update fingerprints of new (future) issues coming in. Open to suggestion.

### History of Fingerprinting
Server-Side Fingerprint configuration was introduced first, then users wanted client-side (SDK) fingerprinting so that was delivered next. I chose to cover Server-Side in more detail because 1. It was demanded first 2. it involves clicking and typing out a few rules, as opposed to updating source code in your client side (people are naturally a bit more resistant to this, IMO. involves testing, CI/CD etc.)

I didn't go into detail on 'Custom Grouping Enhancements'. I thought these were more for native errors, and it's more efficient to stick with the regular documentation on this than reproduce it on a guide.

### Terminologies
In my sub section **What is a fingerprint** I ended up borrowing the definition from https://docs.sentry.io/data-management/event-grouping/ so that's kind of like duplicate documentation now? For weeks I had my own definition/explanation but realized the docs.sentry.io one was smoother. Hmmm.

the following terms sometimes get used interchangeably, even outside of this guide:
"fingerprint" 
"default grouping behavior"
"grouping algorithm"
"grouping technique"

Observation - the "fingerprint" is not the stack trace. the fingerprint is the hash of the stack trace. But when you open the JSON of an event and look at the Key `fingerprint` its value is set to `{{ default }}` meaning "the hash of the stack trace was used", rather than show the hash of the stack trace (i.e. value) itself. I find this very confusing, because if you set your own Server-Side Fingerprint then it shows the value itself and not something like `{{ type-of-algorithm }}`. I raised this in engineering channels. Really don't need to bring the user to this level of detail, but it does make it tricky when conversing about Fingerprints in documentation and with co-workers.


src/collections/_documentation/guides/grouping-and-fingerprints/index.md
```diff
@@ -0,0 +1,97 @@
...
+title: Grouping & Fingerprints
+sidebar_order: 0
+---
+
+Have you ever had a set of similar looking issues like this?
+![Issues Dashboard]({% asset guides/grouping-and-fingerprints/issues-dashboard.png @path %})
+
+Let's first understand what makes events group into an issue. testTypeIssue15 is an issue that had 4 events. These events grouped into the issue because they shared the same *fingerprint*
+
+**What is a fingerprint?** All events have a fingerprint and events with the same fingerprint are grouped together. By default, Sentry will run one of our built-in grouping algorithms to generate a fingerprint based on information available within the event such as stacktrace, exception, and message.
```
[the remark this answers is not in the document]
thinkocapo: @MimiDumpling yes, agreed. I think I used to wonder the same about the fingerprint. I'm committing your suggestion here and will do another commit to address the full concern.

src/collections/_documentation/guides/grouping-and-fingerprints/index.md
```diff
@@ -0,0 +1,97 @@
...
+
+**Do I need to do anything?** No. Stack trace is used by default if it's available. So if 2 events have the same stack trace, then they will group into 1 issue. If you have separate issues that you'd like to group together, then this guide will show you how.
+
+**How do I see the fingerprint?** If you're curious then open an issue, click the JSON link, and find the *fingerprint* property. If the default grouping was used, you'll see default written there. If a different grouping was used, you'll see the actual fingerprint value itself.
```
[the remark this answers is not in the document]
thinkocapo: good call

src/collections/_documentation/guides/grouping-and-fingerprints/index.md
```diff
@@ -0,0 +1,98 @@
...
+title: Grouping & Fingerprints
+sidebar_order: 0
+---
+
+Have you ever had a set of similar-looking issues like this?
+![Issues Dashboard]({% asset guides/grouping-and-fingerprints/issues-dashboard.png @path %})
+
+Let's first understand what makes events group into an issue. testTypeIssue15 is an issue that had 4 events. These events grouped into the issue because they shared the same *fingerprint*
+
+**What is a fingerprint?**
```
thinkocapo: @MimiDumpling I updated this paragraph to try and make clear Who is setting the Fingerprint. I guess at some point the reader is clued that they can set it themselves, eventually.

src/collections/_documentation/guides/grouping-and-fingerprints/index.md
```diff
@@ -0,0 +1,98 @@
...
+A fingerprint is a way to uniquely identify an event. Sentry sets a fingerprint by default for you. Sentry uses its own built-in grouping algorithms to generate a fingerprint based on information available within the event such as a stack trace, exception, and message. Events with the same fingerprint are grouped together.
+
+**Do I need to do anything?** No. A stack trace is used by default if it's available. So if two events have the same stack trace, then they will group into one issue. If you have separate issues that you'd like to group together, then this guide will show you how.
```
thinkocapo: @MimiDumpling "...`this guide will show you how`." kind of implies, "`this guide will show you how to set your own fingerprint`" I think it's clear if they've been reading along. If someone comes to this page and already knows "I just want to see how to set a fingerprint" then I think they would jump down to the "Solutions" section, and wouldn't end up lost.

a phrase like "Change the grouping behavior algorithm" can be spoken so interchangeably with "Set your own fingerprint rule" that it made it hard to decide when and where to write each 😆

src/collections/_documentation/guides/grouping-and-fingerprints/index.md
MimiDumpling: 👍

src/collections/_documentation/guides/grouping-and-fingerprints/index.md
MimiDumpling: 👍

PR #1413, opened 2019-12-14
build(gcr): Use Google Cloud Run for PR previews

BYK:


MimiDumpling:
:: claps hands in excitement :: 😄

MimiDumpling:
I dunno much about the technical part of this, so I'll defer approval to the other reviewers

BYK:
And finally, here's the preview link for this PR: https://sentry-docs-o2paie5ivq-uc.a.run.app/

cameronmcefee:
Damn, this is nice. How do you feel about setting `<meta name="robots" content="noindex, nofollow" />` in this environment just to be safe?

BYK:
>Damn, this is nice. How do you feel about setting `<meta name="robots" content="noindex, nofollow" />` in this environment just to be safe?

That's fine but the environment is the same as prod so not sure how doable is that (or at least easily doably). What's your main concern?

cameronmcefee:
Could you add an environment variable in that we check for in the template? While it's definitely overkill, I'd like to prevent google accidentally indexing not-technically-production content

BYK:
>Could you add an environment variable in that we check for in the template? While it's definitely overkill, I'd like to prevent google accidentally indexing not-technically-production content

I think this is a valid concern and using an env var is an efficient way to get this done. Thanks for the suggestion and update coming soon!

Dockerfile
```diff
@@ -1,3 +1,4 @@
 ARG VERSION=2019.05.08
 FROM getsentry/jekyll-base:builder-${VERSION} AS builder
 FROM getsentry/jekyll-base:runtime-${VERSION} AS runtime
+CMD ["/bin/sh", "-c", "sed -i \"s/\\([[:space:]]\\{1,\\}\\)listen 80;/\\1listen ${PORT:-80};/\" /etc/nginx/conf.d/default.conf && exec nginx -g 'daemon off;'"]
```
mattrobenolt: We should just bake this behavior into the `jekyll-base:runtime` image in the entrypoint. This is fine as is, but we can do better. The container should also include `envsubst(1)` which should make this pattern a bit easier.

BYK: I'll submit that as a fix up. Cannot use `envsubst(1)` as it doesn't support fallback values.

mattrobenolt: You shouldn't need it. :) I can show you what I'm talking about ifyou want. You'd do the fallback logic in the actual entrypoint. No need for it in the template.

BYK: Oh, right. Yeah, not sure why I didn't go with that as I tried that earlier.

mattrobenolt: :) Yeah, just do the variable stuff inside the bash script before calling envsubst.

BYK:
Don't think this is needed anymore with the new docs effort + Vercel.

PR #1415, opened 2019-12-17
feat(webpack): Add notes on configuring SENTRY_RELEASE

EvanPurkhiser:
Fixes https://app.asana.com/0/1144304343179777/1144580065042336

I didn't add anything about `SENTRY_DSN` since typically that is just a hardcoded static value, not something that needs to be configured through webpack.

### DIFF
--- a/src/collections/_documentation/error-reporting/configuration/index.md
+++ b/src/collections/_documentation/error-reporting/configuration/index.md
@@ -45,7 +45,8 @@ Release names are just strings but some formats are detected by Sentry and might
 differently.  For more information have a look at [the releases documentation]({% link _documentation/workflow/releases.md %}).
 
 By default the SDK will try to read this value from the `SENTRY_RELEASE` environment
-variable (except for the browser SDK where this is not applicable).
+variable (in the browser SDK, this will be read off of the `window.SENTRY_RELEASE` if
+available).
 
 {:.config-key}
 ### `environment`
--- a/src/collections/_documentation/platforms/javascript/index.md
+++ b/src/collections/_documentation/platforms/javascript/index.md
@@ -88,6 +88,8 @@ Most promise libraries have a global hook for capturing unhandled errors. You ma
 
 {% include platforms/configure-releases.md %}
 
+If you are using the [Webpack Plugin](#webpack) releases will be automatically configured for you using the plugin's release-lookup logic.
+
 ## Source Maps
 
 {% include components/alert.html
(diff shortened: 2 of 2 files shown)


src/collections/_documentation/error-reporting/configuration/index.md
```diff
@@ -54,7 +54,7 @@ Sets the environment.  This string is freeform and not set by default.  A releas
 with more than one environment to separate them in the UI (think `staging` vs `prod` or similar).
 
 By default the SDK will try to read this value from the `SENTRY_ENVIRONMENT`
-environment variable (except for the browser SDK where this is not applicable).
+environment variable (In the browser SDK, this will be read off of the `window` if available).
```
lobsterkatie: Do you want to mention here that the reason it would be available is because of the webpack plugin?

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -88,6 +88,8 @@ Most promise libraries have a global hook for capturing unhandled errors. You ma
 
 {% include platforms/configure-releases.md %}
 
+If you are using the [Webpack Plugin](#webpack) releases will be automatically configured for you using the plugins release lookup logic.
```
lobsterkatie:
```suggestion
If you are using the [Webpack Plugin](#webpack) releases will be automatically configured for you using the plugin's release lookup logic.
```

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -146,6 +150,8 @@ module.exports = {
 
 In case you use [SourceMapDevToolPlugin](https://webpack.js.org/plugins/source-map-dev-tool-plugin) for more fine-grained control of source map generation, leave `noSources` turned off, so Sentry can display proper source code context in event stack traces.
 
+Additionally the webpack plugin will also automatically configure your `window.SENTRY_RELEASE`, meaning your `Sentry.init` call will not need to be configured with the `release` key.
```
lobsterkatie:
```suggestion
Additionally, the webpack plugin will automatically set `window.SENTRY_RELEASE`, meaning your `Sentry.init` call will not need to be configured with the `release` key.
```

src/collections/_documentation/error-reporting/configuration/index.md
EvanPurkhiser: Oh this isn't even the right place to change this, but no, would rather keep webpack out of this, since there might be other reasons you would configure `window.SENTRY_RELEASE`.

lobsterkatie: Fair. I'd still vote for telling people what the variable is called here, though. `(In the browser SDK, this will be read off of `window.SENTRY_RELEASE` if defined)` or some such.

EvanPurkhiser: Yeah agreed, updated

PR #1419, opened 2019-12-18
Fix react-native codepush release value.

wali-s:
The source maps don't get associated without the bundle id.

What I believe happened is `Sentry.setVersion` got changed to `Sentry.setRelease` in the docs and `Sentry.setRelease` expects a bundle id which was not reflected here.


bruno-garcia:
@HazAT any input here?

PeloWriter:

PR #1421, opened 2019-12-19
Use SetExtras() on scope instead of package

Dalabad:
The called method SetExtras() can only be called on the scope, not as previously noted on the package level


dashed:
@Dalabad thanks!

PR #1422, opened 2019-12-19
Add webhook plugin page and payload documentation

MeredithAnya:
**Context:**
We don't document the Webhook Plugin anymore (that I could find) let alone the payload that gets sent to the configured urls. It's unclear to me if it was intentional because it is a Legacy Integration and we would prefer to push people to use the Integration Platform for webhooks. However people still write in asking about this, and while we can point to the source code, it would be nice for people to discover it through the docs. 

**Changes:**
* Link to the Integration Platform in the Integrations with Notifications section to notify people of that option
* Add Webhooks as an option in the Legacy Integration section
> Screen Shot 2019-12-19 at 10 18 31 AM
* Add Webhook Plugin page (I didn't want this to be in the tree structure so I made it independent, welcome to feedback on that front)

### DIFF
--- a/src/webhook-plugin.md
+++ b/src/webhook-plugin.md
@@ -4,60 +4,270 @@ title: 'Webhook Plugin'
 
 ### Webhook Plugin
 
+When the webhook plugin is enabled and configured with one or more urls, it can be used in the [action section]({%- link _documentation/workflow/notifications/alerts.md -%}#actions) of the alert rules. When an event triggers a rule that has the Webhook Plugin as one of the actions, the following attributes and payload will be sent to the url(s) specified in the plugin configuration.
+ 
 **Attributes**
 
-**data['id']**
+**`id`**
 - type: string
 - description: The issue id for the issue associated with the event
 
-**data['project']**
+**`project`**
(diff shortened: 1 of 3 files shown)


src/webhook-plugin.md
```diff
@@ -0,0 +1,273 @@
+---
+title: 'Webhook Plugin'
+---
+
+### Webhook Plugin
+
+When the webhook plugin is enabled and configured with one or more urls, it can be used in the [action section]({%- link _documentation/workflow/notifications/alerts.md -%}#actions) of the alert rules. When an event triggers a rule that has the Webhook Plugin as one of the actions, the following attributes and payload will be sent to the url(s) specified in the plugin configuration.
```
MimiDumpling:
```suggestion
When the webhook plugin is enabled and configured with one or more URLs, you can use it in the [action section]({%- link _documentation/workflow/notifications/alerts.md -%}#actions) of the alert rules. When an event triggers a rule that has the Webhook Plugin as one of the actions, the following attributes and payload will be sent to the URL(s) specified in the plugin configuration.
```

PR #1423, opened 2019-12-23
chore: Discontinue Java Agent

bruno-garcia:


bruno-garcia:
@MimiDumpling I suggest we keep it for a few weeks, then move the code (agent) to a branch and delete it from master. At that point we can delete the docs too.

PR #1428, opened 2020-01-02
build(nokogiri): Security upgrade for nokogiri

BYK:


cameronmcefee:
I'll investigate. Most of it is style guide violations but I see actual test errors that need debugging.

cameronmcefee:
Ok, this is actually failing during the build, so the tests are running on empty data. You can see the error by running `bin/server --trace`. It looks like something that got updated includes a breaking change. Looks to be somewhere in jekyll-assets, which tends to be aggravatingly fragile.

cameronmcefee:
This is introducing more issues than I can presently take the time to troubleshoot.

BYK:
@cameronmcefee I'll take a look at this later on.

joshuarli:
Closing; not applicable anymore.

PR #1431, opened 2020-01-07
Update express.md, flushTimeout default is not 2000

andidev:
The default value for flushTimeout is not 2000, it is undefined: 
https://github.com/getsentry/sentry-javascript/blob/645ff532941d9a23bb6531c6305e99544b356502/packages/node/src/handlers.ts#L293


dashed:
For reference, `flushTimeout ` was added in: https://github.com/getsentry/sentry-javascript/pull/2017

PR #1436, opened 2020-01-10
Clarify that releases are org wide

ceorourke:
Sometimes users are confused when they don't realize releases are scoped to the organization rather than a project. If you have 2 projects that use the same versioning like project A and B both have 1.5.6 it can be confusing, so we'd recommend they do `projectA-1.5.6` and `projectB-1.5.6` or something.

### DIFF
--- a/src/collections/_documentation/cli/releases.md
+++ b/src/collections/_documentation/cli/releases.md
@@ -16,7 +16,7 @@ Because releases work on projects you will need to specify the organization and 
 
 ## Creating Releases
 
-Releases are created with the `sentry-cli releases new` command. It takes at the very least a version identifier that uniquely identifies the relases. It can be arbitrary but for certain platforms recommendations exist:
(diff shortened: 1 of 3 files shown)


src/collections/_documentation/cli/releases.md
```diff
@@ -3,7 +3,7 @@ title: 'Release Management'
 sidebar_order: 2
 ---
 
-The `sentry-cli` tool can be used for release management on Sentry. It allows you to create, edit and delete releases as well as upload release artifacts for them.
+The `sentry-cli` tool can be used for release management on Sentry. It allows you to create, edit and delete releases as well as upload release artifacts for them. Note that releases are global per organization, so if you want the releases in different projects to be treated as separate entities, you should make the version name unique across the organization.
```
MimiDumpling: Should we add an example? Similar to the one you wrote in the PR description?

ceorourke: Should we? 🤔 I was trying to keep it less wordy, is this not clear enough? I'm open to changing it!

PR #1447, opened 2020-01-19
Add installation instructions for CLI via Scoop

alastairs:
Hi there! I created a [Scoop](https://scoop.sh) manifest for easy installation of sentry-cli on Windows, and it's PRd over at ScoopInstaller/Main#738. It would be great to see this documented once it's live!


dashed:
@BYK I'm adding you as a reviewer since you're one of the few that I know who uses Windows.

PR #1449, opened 2020-01-21
Added alert box to IP Ranges page

MimiDumpling:
Our IP Ranges page only applies to the SaaS product -- not on-prem or single tenant. Added the alert box to give readers the correct expectations.

image

### DIFF
--- a/src/ip-ranges.md
+++ b/src/ip-ranges.md
@@ -3,7 +3,7 @@ title: 'IP Ranges'
 ---
 
 {% capture __alert_content -%}
-The contents of this page only apply to Sentry's SaaS product. It **does not** apply to on-premise or single tenant.
+The contents of this page only apply to Sentry's SaaS product. It **does not** apply to On-Premise or Single Tenant.
 {%- endcapture -%}
 {%- include components/alert.html
     title="Note"
--- a/src/ip-ranges.md
+++ b/src/ip-ranges.md
@@ -2,6 +2,15 @@
 title: 'IP Ranges'
 ---
 
+{% capture __alert_content -%}
+The contents of this page only apply to Sentry's SaaS product. It does not apply to on-premise or single tenant.
+{%- endcapture -%}
+{%- include components/alert.html
+    title="Note"
+    content=__alert_content
+    level="warning"
+%}
+
 Sentry is served from a single IP address for all web traffic and all event accepting. The IP address is:
 
 ```python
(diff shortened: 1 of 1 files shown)


src/ip-ranges.md
```diff
@@ -2,6 +2,15 @@
 title: 'IP Ranges'
 ---
 
+{% capture __alert_content -%}
+The contents of this page only apply to Sentry's SaaS product. It **does not** apply to on-premise or single tenant.
```
mattrobenolt: @dcramer do we officially refer to `single tenant` as `single-tenant` or?

MimiDumpling: @dcramer just checkin' in on this

rhcarvalho:
In https://sentry.io/_/resources/about-sentry/ we have "Sentry Single Tenant". And we have "Self Hosted" instead of "on-prem".

image

MimiDumpling:
@PeloWriter I think we need a conversation with Marketing and Sales about communication around this topic.

MimiDumpling:
> In https://sentry.io/_/resources/about-sentry/ we have "Sentry Single Tenant". And we have "Self Hosted" instead of "on-prem".
> 
> image

@PeloWriter I think we need a conversation with Marketing and Sales about communication around this topic.

dcramer:
Its Single Tenant, Self-Hosted, and On-Premise

dcramer:
For clarity, On-Premise and Self-hosted aren't product names, Single Tenant is.

MimiDumpling:
> Its Single Tenant, Self-Hosted, and On-Premise

So, all three are synonymous with each other?

dcramer:
No they are different words with different meanings

dcramer:
You could swap on-premise in this language to be self-hosted. All the box says is "this only plays to our standard saas service, not single tenant".

PeloWriter:
All three terms are not synonymous, but the discussion here highlights an inconsistency that we should address, then correct, to avoid confusing our customers.

For now, yes, let's swap on-prem to say self-hosted. I will open a larger discussion on terminology, since this is something we need to get right (and by right, I mean consistently referred to) so our customers aren't confused.

MimiDumpling:
@rhcarvalho do you feel like your questions were answered? If not, we can continue the discussion in #discuss-docs. 😊 I'll merge this branch by end of day.

rhcarvalho:
> @rhcarvalho do you feel like your questions were answered?

Thanks for asking, I think I had no questions and just commented to get this old PR rolling again. Happy to see it resolved. Good job!

PR #1459, opened 2020-01-26
Description how capture errors from Spring tasks

MaksTuev:
Sentry by default not capture this errors but developers expect this.

### DIFF
--- a/src/collections/_documentation/clients/java/integrations.md
+++ b/src/collections/_documentation/clients/java/integrations.md
@@ -1012,6 +1012,8 @@ public HandlerExceptionResolver sentryExceptionResolver() {
 }
 ```
 
+
+
 Next, **you’ll need to configure your DSN** (client key) and optionally other values such as `environment` and `release`. [See the configuration page]({%- link _documentation/clients/java/config.md -%}#setting-the-dsn) for ways you can do this.
 
 #### Spring Boot HTTP Data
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/clients/java/integrations.md
```diff
@@ -1034,3 +1036,31 @@ public ServletContextInitializer sentryServletContextInitializer() {
 ```
 
 After that, your Sentry events should contain information such as HTTP request headers.
+
+#### Spring @Scheduled Tasks
+
+If you want to capture errors from tasks you need also add this configuration:
+
+```java
```
bruno-garcia: This is a Kotlin snippet though:

```suggestion
```kotlin
```

I reckon it would be better to have a Java snippet since it's much more common.
Could you provide a Java version, please? Or if someone has a Java spring app handy to prepare one. Maybe @lobsterkatie does?

src/collections/_documentation/clients/java/integrations.md
```diff
@@ -1012,6 +1012,8 @@ public HandlerExceptionResolver sentryExceptionResolver() {
 }
 ```
 
+
```
bruno-garcia: Are these 2 lines intentional?

src/collections/_documentation/clients/java/integrations.md
lobsterkatie: I do not, but maybe @ndmanvar does?

ndmanvar: @lobsterkatie -> https://github.com/sentry-demos/java-spring-boot-log4j
let me know if i can be of any help here.

PR #1460, opened 2020-01-26
Use Dependency Injection for User Feedback on ASP.NET Core

kanadaj:
The documentation should avoid using hard-coded DSNs in places where configuration settings are available as a single source of truth.

### DIFF
--- a/src/collections/_documentation/enriching-error-data/user-feedback-example/aspnetcore.md
+++ b/src/collections/_documentation/enriching-error-data/user-feedback-example/aspnetcore.md
@@ -16,7 +16,7 @@ With ASP.NET Core MVC, the `Error.cshtml` razor page:
 }
 ```
 
-If you configured Sentry without Dependency Injection (e.g. without `UseSentry()`) you can also configure the SDK using a hard-coded DSN in the template:
+If you configured Sentry without the Dependency Injection (for example, without `UseSentry()`) you can also configure the SDK using a hard-coded DSN in the template:
 
 ```js
 Sentry.init({ dsn: '___PUBLIC_DSN___' });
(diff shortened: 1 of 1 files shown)


untitaker:
`___PUBLIC_DSN___` is a placeholder that is dynamically replaced in the docs with the real DSN. I would keep that placeholder somewhere.

kanadaj:
@untitaker Ah I see. I was a bit on the fence about whether I should keep a note about setting up without DI. Will amend in a moment

src/collections/_documentation/enriching-error-data/user-feedback-example/aspnetcore.md
```diff
@@ -5,12 +5,19 @@ With ASP.NET Core MVC, the `Error.cshtml` razor page:
...
 {
     <script>
-        Sentry.init({ dsn: '___PUBLIC_DSN___' });
+        Sentry.init({ dsn: '@(SentryOptions.Value.Dsn)' });
         Sentry.showReportDialog({ eventId: '@SentrySdk.LastEventId' });
     </script>
 }
 ```
+
+If you configured Sentry without Dependency Injection (e.g. without `UseSentry()`) you can also configure the SDK using a hard-coded DSN in the template:
```
MimiDumpling:
```suggestion
If you configured Sentry without the Dependency Injection (for example, without `UseSentry()`) you can also configure the SDK using a hard-coded DSN in the template:
```

kanadaj:
@MimiDumpling thank you, I actually wasn't sure about consistency since I'm not a regular contributor for for the docs :) Added the suggestions as a commit!

bruno-garcia:
Thanks @kanadaj !

PR #1466, opened 2020-01-29
Update Visibility doc in preparation for Discover V2

MimiDumpling:
Discover V2 will impact the Visibility page. Docs will address these changes by:

1. Removing the Discover section in the Visibility docs
2. Renaming the page Dashboards instead of Visibility
3. Removing the Events section because that's changing with Discover V2
4. Removing the Tracing section because the information is already available here: [https://docs.sentry.io/enriching-error-data/tracing/](https://docs.sentry.io/enriching-error-data/tracing/)

image


dashed (approved):
Looks good to me 👍

PR #1474, opened 2020-01-31
(fix) Reformat defer note in JS docs

ceorourke:
Code formatting doesn't appear to work in alert box titles, so remove this formatting.

**Before**:
docs sentry io_platforms_javascript_

**After**:
0 0 0 0_9001_platforms_javascript_


ceorourke:
Turns out I forgot to merge this which is good news because I also missed a couple other places this happens.

PR #1475, opened 2020-01-31
ref(search): Clarify search syntax

adhiraj:
- Move the tags syntax out of Advanced (it's not an advanced feature)
- Don't encourage using reserved keywords as tags (you may wish to... -> if you do..)
- organization filter list -> organization saved search

### DIFF
--- a/src/collections/_documentation/workflow/search.md
+++ b/src/collections/_documentation/workflow/search.md
@@ -37,9 +37,9 @@ The tokens `is:resolved` and `user.username:"Jane Doe"` are standard search toke
 
 The token `example error` is utilizing the optional raw search and is passed as part of the issue search query (which uses a CONTAINS match similar to SQL). When using the optional raw search, you can provide _one_ string, and the query uses that entire string.
 
-### Tag Syntax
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/workflow/search.md
```diff
@@ -307,7 +305,7 @@ You can pin a premade search the same way you pin any other search. When you've
 
 Owners and managers can create a persistent view for their organization by creating custom saved searched. These saved searches will not be associated with a specific project, but with all projects (and users) across the org.
 
-1. Type a search into the search bar, then click the "add to Organization filter list" icon just to the right of it. Keeping in mind that you need to be an owner or manager within the org to use this feature.
+1. Type a search into the search bar, then click the "add to Organization saved searches" icon just to the right of it. Keeping in mind that you need to be an owner or manager within the org to use this feature.
```
MimiDumpling:
```suggestion
1. Type a search into the search bar, then click the "add to Organization saved searches" icon just to the right of it. Keep in mind that you need to be an owner or manager within the org to use this feature.
```

PR #1483, opened 2020-02-03
Generalize statement the -> a

rhcarvalho:
Depending on the platform, there might not be a concept of a single
background thread. Therefore, prefer the more general "a background
thread" instead of the specific "the background thread".


rhcarvalho:
I had this minor change lost somewhere in my disk, flushing it out here.

Reviewers few free to flag this as incorrect if the edit doesn't make sense and we'll drop this PR.

rhcarvalho:
Thanks for the validation! 😄

PR #1489, opened 2020-02-06
Update the PHP transport classes section

HazAT:


src/collections/_documentation/platforms/php/index.md
```diff
@@ -382,12 +422,19 @@ use Sentry\State\Hub;
...
+$transportFactory = new class implements TransportFactoryInterface {
+    public function create(\Sentry\Options $options): \Sentry\Transport\TransportInterface
+    {
+        return new SpoolTransport($spool);
+    }
+};
+
 $builder = ClientBuilder::create($options);
-$builder->setTransport($transport);
+$builder->setTransportFactory($transportFactory);
```
ste93cry: Somewhere after this call there should be a call to `$spool->flushQueue($httpTransport);`. This means that you should instantiate another transport (e.g. the HTTP transport) yourself and use it here

ste93cry:
As a general thing I would suggest to always import all classes using `use` statements instead of having a mix of imports and FQCN

src/collections/_documentation/platforms/php/index.md
HazAT: It's there, 3 lines below.

ste93cry: Ops, I didn't see it

PR #1490, opened 2020-02-06
Update information about ignore_logger usage

jaysonsantos:


untitaker:
That's not true though. The moment you call `ignore_logger` the logger is ignored, it has nothing to do with `init()`. What is the actual behavior you are seeing?

jaysonsantos:
Hey there, I guess you are right.
Trying to isolate the problem I found out that it won't ignore the children of root's logger.
If I ignore the root e.g. `test` it does not mean that `test.child` would be ignored.
From that, I thought that the problem was with the ordering but it was with the children.

jaysonsantos:
Hey there @untitaker, do you think that it is important to document that if you ignore the root logger it does not mean that the children will also be ignored?

untitaker:
Do you want to make a pr? Otherwise I'll do one later

PR #1487, opened 2020-02-05
fix(sso): Update and simplify SSO docs

BYK:
Fixes an [issue reported on the forum](https://forum.sentry.io/t/upgrade-9-0-with-github-sso-9-1-internal-server-error/8613?u=byk) and getsentry/onpremise#373.

PR #1492, opened 2020-02-06
fix(github): Fix GitHub SSO Auth URL

BYK:
Follow up to #1487, addressing https://git.io/JvZMx

### DIFF
--- a/src/collections/_documentation/server/integrations/github/index.md
+++ b/src/collections/_documentation/server/integrations/github/index.md
@@ -11,7 +11,7 @@ sidebar_order: 1
     | Setting | Value |
     | ---- | ---- |
     | Homepage URL | ${urlPrefix} |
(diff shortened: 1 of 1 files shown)


EvanPurkhiser:
There may be a reason it needs to be like this. It depends on if github shares the same app for integrations and SSO (which It might, I need to verify)

BYK:
>It depends on if github shares the same app for integrations and SSO (which It might, I need to verify)

It doesn't but it can and I already verified this myself. This was also mentioned in the original issue: https://github.com/getsentry/onpremise/issues/373#issue-558531728

EvanPurkhiser:
Did we actually change this URL? These are both valid URLs in Sentry.

BYK:
I only changed the user auth URL which makes sense here. The integration still uses the setup URL, well, as the setup URL.

PR #1498, opened 2020-02-13
fixed some android snippets and bumped sentry-java

marandaneto:


src/_data/platforms.yml
```diff
@@ -93,10 +93,10 @@
       support_level: production
       type: framework
       name: Android
-      doc_link: /clients/java/modules/android/
-      wizard_parent: java
+      doc_link: /platforms/android/
+      wizard_parent: android
       wizard:
-        - _documentation/clients/java/integrations.md#android
+        - _documentation/platforms/android/#integrating-the-sdk
```
HazAT: Look for places where wizard: true
I think this is how it works now.
Compare to php or something else.

Also there is <!— Wizard android —> double check so we are not missing something

marandaneto: I am a bit lost here, not sure how to test it either, all versions (other platforms) are a bit different and this one was a client/integration and now its own platform.
how do I test it?

cameronmcefee: My environment is in an inoperable state currently so I can't verify this for you, but to compile this, run:

```
JEKYLL_ENABLE_PLATFORM_API=true bin/server
```

Check the output of _site for a folder called _platforms and then look to see if your new stuff exists in there. This is one of the most opaque parts of the docs, so it's difficult to deal with. Code is here: https://github.com/getsentry/sentry-docs/blob/master/src/_plugins/platform_api.rb

MimiDumpling:
@marandaneto Let me know if Cameron's suggestion works. If not, we can still deploy the docs without the wizard and add the wizard later.

src/_data/platforms.yml
marandaneto: thanks for the tips, the command didn't work though :(
`NoMethodError: undefined method `content' for nil:NilClass`
I'll merge without the wizard and figure it out later on, thanks

PR #1503, opened 2020-02-18
ref: Simplify gradle docs

bruno-garcia:
Right now, after the advised configuration we immediately start talking about "manual config".
I believe it's confusing. In 90% of cases people should stop at that configuration and not even be aware of those manual steps and the _strongly discouraged_ options.


src/collections/_documentation/platforms/android/gradle-advanced.md
```diff
@@ -0,0 +1,67 @@
...
+Sentry calculates UUIDs for ProGuard files. For more information about how this works, see [ProGuard UUIDs]({%- link _documentation/workflow/debug-files.md -%}#proguard-uuids).
+
+`sentry-cli` can write the `sentry-debug-meta.properties` file for you:
+
+```bash
+sentry-cli upload-proguard \
+    --android-manifest app/build/intermediates/manifests/full/release/AndroidManifest.xml \
+    --write-properties app/build/intermediates/assets/release/sentry-debug-meta.properties \
+    --no-upload \
+    app/build/outputs/mapping/{buildVariant}/mapping.txt
```
marandaneto: those folders are looked up dynamically, not sure if we should add as a static path, it might create more questions.
eg `manifests` is called `merged_manifests` on newer versions
`assets` is called `merged_assets` on newer versions
where is `release`, it should be `{buildVariant}` because it's also dynamic.
I'd say that we should stick to a generic path and point them to the Gradle integration source code?

src/collections/_documentation/platforms/android/gradle-advanced.md
```diff
@@ -0,0 +1,67 @@
...
+                "${project.rootDir.toPath()}/app/build/intermediates/assets" +
+                    "/${variant.dirName}/sentry-debug-meta.properties",
+                variant.getMappingFile(),
+                "--no-upload"
+            ]
+        }
+        dexTask.dependsOn task
+        task.dependsOn proguardTask
+    }
+}
```
marandaneto: same as the comment above, everything is dynamic, task names, paths, tasks order, if this even works, might be only on very old versions.

src/collections/_documentation/platforms/android/gradle-advanced.md
```diff
@@ -0,0 +1,67 @@
...
+Alternatively, you can generate a UUID upfront yourself and then force Sentry to honor that UUID after upload. However, this is **strongly discouraged**.
+
+### Uploading ProGuard Files
+
+Finally, manually upload the ProGuard files with `sentry-cli` as follows:
+
+```bash
+sentry-cli upload-proguard \
+    --android-manifest app/build/intermediates/manifests/full/release/AndroidManifest.xml \
+    app/build/outputs/mapping/{buildVariant}/mapping.txt
```
marandaneto: same as https://github.com/getsentry/sentry-docs/pull/1503/files#r380807690

src/collections/_documentation/platforms/android/gradle-advanced.md
bruno-garcia: Should I just delete this whole `advanced-gradle` doc?
I believe we should point users to the plugin only. Is it a real use-case that ppl go straight to calling a CLI from gradle instead of the plugin?

src/collections/_documentation/platforms/android/gradle-advanced.md
```diff
@@ -0,0 +1,67 @@
+# Advanced gradle plugin configuration
+
+If for some reason the gradle plugin isn't an option for you, here are other ways to get the tasks done:
```
MimiDumpling:
```suggestion
If for some reason the Gradle plugin isn't an option for you, here are other ways to get the tasks done:
```

src/collections/_documentation/platforms/android/gradle-advanced.md
```diff
@@ -0,0 +1,67 @@
...
+                "--no-upload"
+            ]
+        }
+        dexTask.dependsOn task
+        task.dependsOn proguardTask
+    }
+}
+```
+
+Alternatively, you can generate a UUID upfront yourself and then force Sentry to honor that UUID after upload. However, this is **strongly discouraged**.
```
MimiDumpling: Maybe we want to add _why_ we strong discourage these steps.

src/collections/_documentation/platforms/android/index.md
```diff
@@ -242,69 +242,8 @@ sentry {
...
-
-Finally, manually upload the ProGuard files with `sentry-cli` as follows:
-
-```bash
-sentry-cli upload-proguard \
-    --android-manifest app/build/intermediates/manifests/full/release/AndroidManifest.xml \
-    app/build/outputs/mapping/{buildVariant}/mapping.txt
-```
+And that's it! Now when you build your app, the plugin will upload the proguard/R8 mappings,
+source bundle and native symbols, as you configured, to Sentry.
```
MimiDumpling:
```suggestion
source bundle, and native symbols, as you configured them to Sentry.
```

src/collections/_documentation/platforms/android/gradle-advanced.md
bruno-garcia: I rather remove this altogether tbh. The person who wrote left and for me it doesn't make sense to leave it in. Let's not give an option

src/collections/_documentation/platforms/android/gradle-advanced.md
marandaneto: I'd be in to kill it all, I really don't see a reason to have this manual steps.
a valid point would be using another build system instead of Gradle, but then again, those scripts won't work.
See: https://stackoverflow.com/questions/60282834/gradle-task-for-sentry-not-compiling
I just asked why people were trying to use those scripts instead of the Gradle integration and found no valid reason rather the Integration didn't work.
I think we should focus on making the integration to work smoothly rather than fixing those scripts.

PR #1507, opened 2020-02-20
docs: Replace Go config with formattable placeholder

kamilogorek:


mattrobenolt:
We’re going to need to roll that DSN I assume.

PR #1508, opened 2020-02-21
improve android docs, just little tweaks

marandaneto:


src/collections/_documentation/platforms/android/index.md
```diff
@@ -359,7 +359,8 @@ For more information, see [Tagging Events]({%- link _documentation/enriching-err
 
 ### Setting the Level
 
-You can set the severity of an event to one of five values: `fatal`, `error`, `warning`, `info`, and `debug`. `error` is the default, `fatal` is the most severe, and `debug` is the least severe.
+You can set the severity of an event to one of five values: `fatal`, `error`, `warning`, `info`, and `debug`.
+`error` is the default, `fatal` is the most severe, and `debug` is the least severe.
```
marandaneto: I just jumped a line because the sentence `error is the default` was mixed with other things and you might be confused on which level actually is the default one

PR #1509, opened 2020-02-21
docs: Add note regarding source-map-support package.

kamilogorek:


src/collections/_documentation/platforms/node/sourcemaps.md
```diff
@@ -37,6 +37,16 @@ module.exports = {
 };
 ```
 
+Make sure that your code is not using [source-map-support](https://www.npmjs.com/package/source-map-support) package if you want to rely on Sentry source maps resolution.
```
lobsterkatie:
```suggestion
If you want to rely on Sentry's source map resolution, make sure that your code is not using the [source-map-support](https://www.npmjs.com/package/source-map-support) package, as it overwrites the captured stacktrace in a way that makes it impossible for our processors to correctly parse it.
```

src/collections/_documentation/platforms/node/sourcemaps.md
```diff
@@ -37,6 +37,16 @@ module.exports = {
...
+The package is overwriting `stack` property and `Error` object, and we need the original stace traces in order to make source maps work correctly.
```
lobsterkatie:
```suggestion
```

src/collections/_documentation/platforms/node/sourcemaps.md
```diff
@@ -37,6 +37,15 @@ module.exports = {
 };
 ```
 
+If you want to rely on Sentry's source map resolution, make sure that your code is not using the [source-map-support](https://www.npmjs.com/package/source-map-support) package, as it overwrites the captured stacktrace in a way that makes it impossible for our processors to correctly parse it.
```
kamilogorek:
```suggestion
If you want to rely on Sentry's source map resolution, make sure that your code is not using the [source-map-support](https://www.npmjs.com/package/source-map-support) package, as it overwrites the captured stack trace in a way that makes it impossible for our processors to correctly parse it.
```

PR #1512, opened 2020-02-25
Revise Shutdown and Draining section

rhcarvalho:
The original motivation was a continuation to getsentry/sentry-go#106, to clarify the usage of `Flush` in the Go SDK.

I went ahead and revised the contents for other languages as well.

### DIFF
--- a/src/collections/_documentation/error-reporting/configuration/drain-example/go.md
+++ b/src/collections/_documentation/error-reporting/configuration/drain-example/go.md
@@ -1,11 +1,18 @@
-The client provides a `Flush` method that takes the time in `time.Duration` for how long it waits and will return a boolean that indicates whether everything was flushed or the timeout kicked in.
+To avoid unintentionally dropping events when the program terminates, arrange
+for `sentry.Flush` to be called, typically using `defer`.
+
+If you use multiple clients, arrange for each of them to be flushed as
+appropriate.
+
+`Flush` waits until any buffered events are sent to the Sentry server, blocking
+for at most the given timeout. It returns `false` if the timeout was reached. In
+that case, some events may not have been sent.
 
 ```go
-sentry.CaptureMessage("my message")
+func main() {
+	// err := sentry.Init(...)
+	defer sentry.Flush(2 * time.Second)
 
-if sentry.Flush(time.Second * 2) {
-	fmt.Println("All queued events delivered!")
-} else {
-	fmt.Println("Flush timeout reached")
+	sentry.CaptureMessage("my message")
 }
 ```
--- a/src/collections/_documentation/error-reporting/configuration/drain-example/javascript.md
+++ b/src/collections/_documentation/error-reporting/configuration/drain-example/javascript.md
@@ -1,12 +1,15 @@
-The client provides a close method that optionally takes the time in milliseconds for how long it
-waits and will return a promise that resolves when everything was flushed or the timeout kicked
+The `close` method optionally takes a timeout in milliseconds and returns a
+promise that resolves when all pending events are flushed, or the timeout kicks
 in.
 
 ```javascript
-const client = Sentry.getCurrentHub().getClient();
-if (client) {
-  client.close(2000).then(function() {
-    process.exit();
-  });
-}
+Sentry.close(2000).then(function() {
+  process.exit();
+});
 ```
+
+After a call to `close`, the current client cannot be used anymore. It's
+important to only call `close` immediately before shutting down the application.
+
+Alternatively, the `flush` method drains the event queue while keeping the
+client enabled for continued use.
--- a/src/collections/_documentation/error-reporting/configuration/drain-example/native.md
+++ b/src/collections/_documentation/error-reporting/configuration/drain-example/native.md
@@ -6,3 +6,6 @@ sentry_shutdown();
 
 Calling `sentry_shutdown()` before exiting the application is critical to avoid
 data loss.
+
+After shutdown, the client cannot be used anymore. It's important to only call
+`sentry_shutdown()` immediately before shutting down the application.
--- a/src/collections/_documentation/error-reporting/configuration/drain-example/python.md
+++ b/src/collections/_documentation/error-reporting/configuration/drain-example/python.md
@@ -8,3 +8,9 @@ client = Hub.current.client
 if client is not None:
     client.close(timeout=2.0)
 ```
+
+After a call to `close`, the client cannot be used anymore. It's important to
+only call `close` immediately before shutting down the application.
+
+Alternatively, the `flush` method drains the event queue while keeping the
+client enabled for continued use.
--- a/src/collections/_documentation/error-reporting/configuration/drain-example/rust.md
+++ b/src/collections/_documentation/error-reporting/configuration/drain-example/rust.md
@@ -2,7 +2,7 @@ When the Rust SDK initializes a guard is returned from the `init` function.  The
 will automatically wait `shutdown_timeout` seconds.  This means you just need to hold on to
 the guard and make sure it disposes on shutdown.  Alternatively the client can be closed:
 
-```javascript
+```rust
 use std::time::Duration;
 use sentry::Hub;
 
@@ -10,3 +10,6 @@ if let Some(client) = Hub.current().client() {
     client.close(Some(Duration::from_secs(2)));
 }
 ```
+
+After a call to `close`, the client cannot be used anymore. It's important to
+only call `close` immediately before shutting down the application.
--- a/src/collections/_documentation/error-reporting/configuration/draining.md
+++ b/src/collections/_documentation/error-reporting/configuration/draining.md
@@ -3,15 +3,9 @@ title: 'Shutdown and Draining'
 sidebar_order: 1
 ---
 
-Most SDKs use a background queue to send out events.  Because the queue sends asynchronously in the background
-it means that some events might be lost if the application shuts down unexpectedly.  To prevent this all SDKs
-provide mechanisms to cope with this.
-
-Typically SDKs provide two ways to shut down: a controlled shutdown where the system will wait up to about two
-seconds to flush out events (configurable) and an uncontrolled shutdown (also referred to as "killing" the
-client).
+The default behavior of most SDKs is to send out events over the network
+asynchronously in the background. This means that some events might be lost if
+the application shuts down unexpectedly. The SDKs provide mechanisms to cope
+with this.
 
 {% include components/platform_content.html content_dir='drain-example' %}
-
-After shutdown the client cannot be used any more so make sure to only do that right before you shut down
-the application.


rhcarvalho:
The only language I didn't touch was C#/.NET.

This was intentional, as I believe @bruno-garcia would rather not mention Flush, and the current state seems to be accurate.

The global changes would still take effect, and I think generally the edits make sense for C#:
- The "Typically SDKs provide two ways" paragraph was redundant.
- The "After shutdown the client cannot be used" paragraph doesn't make sense, since the SDK should not be used outside of the `SentrySdk.Init` block anyway.

---

**BEFORE**

image

**AFTER**

image

src/collections/_documentation/error-reporting/configuration/drain-example/go.md
```diff
@@ -1,11 +1,18 @@
...
 ```go
-sentry.CaptureMessage("my message")
+func main() {
+	// err := sentry.Init(...)
+	defer sentry.Flush(2 * time.Second)
 
-if sentry.Flush(time.Second * 2) {
-	fmt.Println("All queued events delivered!")
-} else {
-	fmt.Println("Flush timeout reached")
```
rhcarvalho: I intentionally omit checking on the return value of `sentry.Flush`.

Printing debug messages is redundant, and the same visibility can be achieved by enabling debug logging on the SDK.

There's nothing useful users can do if `Flush` times out. For instance, retrying is not an option because that could be replaced with a longer timeout to begin with, and typically Flush is called in a moment where the program should not hang indefinitely (and that's the very reason for having a timeout).

src/collections/_documentation/error-reporting/configuration/drain-example/javascript.md
```diff
@@ -1,12 +1,15 @@
...
 ```javascript
-const client = Sentry.getCurrentHub().getClient();
-if (client) {
-  client.close(2000).then(function() {
-    process.exit();
-  });
-}
+Sentry.close(2000).then(function() {
+  process.exit();
+});
```
rhcarvalho: @HazAT @kamilogorek was there a reason to manually get the client instead of using `Sentry.close` that does the same?

src/collections/_documentation/error-reporting/configuration/drain-example/javascript.md
```diff
@@ -1,12 +1,15 @@
...
 ```
+
+After calling `close` the current client cannot be used any more so make sure to
+only do that right before you shut down the application.
```
rhcarvalho: This is moved here from the global context, with one important edit: "the client" => "the **current** client".

src/collections/_documentation/error-reporting/configuration/drain-example/go.md
```diff
@@ -1,11 +1,18 @@
-The client provides a `Flush` method that takes the time in `time.Duration` for how long it waits and will return a boolean that indicates whether everything was flushed or the timeout kicked in.
+To avoid unintentionally dropping events when the program terminates, arrange
+for `sentry.Flush` to be called, typically using `defer`.
+
+If you use multiple clients, arrange for each of them to be flushed as
+appropriate.
```
rhcarvalho: I wonder if we want to add a similar note for other languages?

src/collections/_documentation/error-reporting/configuration/drain-example/javascript.md
```diff
@@ -1,12 +1,15 @@
...
+
+Alternatively, the `flush` method drains the event queue while keeping the
+client enabled for continued use.
```
rhcarvalho: I would understand if @HazAT or @kamilogorek would rather omit this note. Thoughts?

src/collections/_documentation/error-reporting/configuration/drain-example/native.md
```diff
@@ -6,3 +6,6 @@ sentry_shutdown();
 
 Calling `sentry_shutdown()` before exiting the application is critical to avoid
 data loss.
+
+After shutdown the client cannot be used any more so make sure to only do that
+right before you shut down the application.
```
rhcarvalho: This is moved in here from the global context with no changes.

As for the general contents of this file, @jan-auer @Swatinem is there anything else you'd like to add?
Is this still accurate and reflecting the best practice?

src/collections/_documentation/error-reporting/configuration/drain-example/python.md
```diff
@@ -8,3 +8,9 @@ client = Hub.current.client
 if client is not None:
     client.close(timeout=2.0)
 ```
+
+After shutdown the current client cannot be used any more so make sure to only
+do that right before you shut down the application.
```
rhcarvalho: This is moved here from the global context, with one important edit: "the client" => "the **current** client".

src/collections/_documentation/error-reporting/configuration/drain-example/python.md
```diff
@@ -8,3 +8,9 @@ client = Hub.current.client
...
+
+Alternatively, the `flush` method drains the event queue while keeping the
+client enabled for continued use.
```
rhcarvalho: I would understand if @untitaker would rather omit this note. Thoughts?

src/collections/_documentation/error-reporting/configuration/drain-example/rust.md
```diff
@@ -2,11 +2,14 @@ When the Rust SDK initializes a guard is returned from the `init` function.  The
 will automatically wait `shutdown_timeout` seconds.  This means you just need to hold on to
 the guard and make sure it disposes on shutdown.  Alternatively the client can be closed:
 
-```javascript
+```rust
```
rhcarvalho: Ooops :smile:

src/collections/_documentation/error-reporting/configuration/drain-example/rust.md
```diff
@@ -2,11 +2,14 @@ When the Rust SDK initializes a guard is returned from the `init` function.  The
...
 use std::time::Duration;
 use sentry::Hub;
 
 if let Some(client) = Hub.current().client() {
     client.close(Some(Duration::from_secs(2)));
 }
 ```
+
+After shutdown the current client cannot be used any more so make sure to only
+do that right before you shut down the application.
```
rhcarvalho: This is moved here from the global context, with one important edit: "the client" => "the **current** client".

src/collections/_documentation/error-reporting/configuration/draining.md
```diff
@@ -3,15 +3,9 @@ title: 'Shutdown and Draining'
 sidebar_order: 1
 ---
 
-Most SDKs use a background queue to send out events.  Because the queue sends asynchronously in the background
-it means that some events might be lost if the application shuts down unexpectedly.  To prevent this all SDKs
-provide mechanisms to cope with this.
-
-Typically SDKs provide two ways to shut down: a controlled shutdown where the system will wait up to about two
-seconds to flush out events (configurable) and an uncontrolled shutdown (also referred to as "killing" the
-client).
```
rhcarvalho: This doesn't sound useful to me. It is a generic statement, and unclear whether it applies to the SDK the user is interested in.

Each language/platform from the dropdown includes the relevant mechanisms that exist in that language/platform.

src/collections/_documentation/error-reporting/configuration/draining.md
```diff
@@ -3,15 +3,9 @@ title: 'Shutdown and Draining'
...
+The default behavior of most SDKs is to send out events over the network
+asynchronously in the background. This means that some events might be lost if
+the application shuts down unexpectedly. The SDKs provide mechanisms to cope
+with this.
 
 {% include components/platform_content.html content_dir='drain-example' %}
-
-After shutdown the client cannot be used any more so make sure to only do that right before you shut down
-the application.
```
rhcarvalho: Moved this to the languages/platforms where it applies.

src/collections/_documentation/error-reporting/configuration/draining.md
```diff
@@ -3,15 +3,9 @@ title: 'Shutdown and Draining'
 sidebar_order: 1
 ---
 
-Most SDKs use a background queue to send out events.  Because the queue sends asynchronously in the background
-it means that some events might be lost if the application shuts down unexpectedly.  To prevent this all SDKs
-provide mechanisms to cope with this.
```
rhcarvalho: Reworded.

src/collections/_documentation/error-reporting/configuration/drain-example/go.md
```diff
@@ -1,11 +1,18 @@
...
+
+`Flush` waits until any buffered events are sent to the Sentry server, blocking
+for at most the given timeout. It returns false if the timeout was reached. In
```
kamilogorek:
```suggestion
for at most the given timeout. It returns `false` if the timeout was reached. In
```

src/collections/_documentation/error-reporting/configuration/drain-example/javascript.md
kamilogorek: Not really, you are right it's the same.

src/collections/_documentation/error-reporting/configuration/drain-example/javascript.md
kamilogorek: I'm totally fine leaving it here.

rhcarvalho:
@MimiDumpling could you please have a look?

MimiDumpling:
> @MimiDumpling could you please have a look?

@rhcarvalho I'll try to have it done by EOD, if not tomorrow. :smile

src/collections/_documentation/error-reporting/configuration/drain-example/javascript.md
```diff
@@ -1,12 +1,15 @@
-The client provides a close method that optionally takes the time in milliseconds for how long it
-waits and will return a promise that resolves when everything was flushed or the timeout kicked
+The `close` method optionally takes a timeout in milliseconds and returns a
+promise that resolves when all pending events were flushed or the timeout kicked
```
MimiDumpling:
```suggestion
promise that resolves when all pending events are flushed, or the timeout kicks
```

src/collections/_documentation/error-reporting/configuration/drain-example/javascript.md
```diff
@@ -1,12 +1,15 @@
...
-  client.close(2000).then(function() {
-    process.exit();
-  });
-}
+Sentry.close(2000).then(function() {
+  process.exit();
+});
 ```
+
+After calling `close` the current client cannot be used any more so make sure to
```
MimiDumpling:
```suggestion
After calling `close`, the current client cannot be used anymore, so make sure to
```

src/collections/_documentation/error-reporting/configuration/drain-example/javascript.md
MimiDumpling:
```suggestion
 do that before you shut down the application.
```

src/collections/_documentation/error-reporting/configuration/drain-example/native.md
```diff
@@ -6,3 +6,6 @@ sentry_shutdown();
 
 Calling `sentry_shutdown()` before exiting the application is critical to avoid
 data loss.
+
+After shutdown the client cannot be used any more so make sure to only do that
```
MimiDumpling:
```suggestion
After shutdown, the client cannot be used anymore, so make sure to do that
```

src/collections/_documentation/error-reporting/configuration/drain-example/native.md
MimiDumpling:
```suggestion
 before you shut down the application.
```

src/collections/_documentation/error-reporting/configuration/drain-example/python.md
```diff
@@ -8,3 +8,9 @@ client = Hub.current.client
 if client is not None:
     client.close(timeout=2.0)
 ```
+
+After shutdown the current client cannot be used any more so make sure to only
```
MimiDumpling:
```suggestion
After shutdown, the current client cannot be used anymore, so make sure to
```

src/collections/_documentation/error-reporting/configuration/drain-example/python.md
MimiDumpling:
```suggestion
do that before you shut down the application.
```

src/collections/_documentation/error-reporting/configuration/drain-example/rust.md
```diff
@@ -2,11 +2,14 @@ When the Rust SDK initializes a guard is returned from the `init` function.  The
...
+```rust
 use std::time::Duration;
 use sentry::Hub;
 
 if let Some(client) = Hub.current().client() {
     client.close(Some(Duration::from_secs(2)));
 }
 ```
+
+After shutdown the current client cannot be used any more so make sure to only
```
MimiDumpling:
```suggestion
After shutdown, the current client cannot be used anymore, so make sure to
```

src/collections/_documentation/error-reporting/configuration/drain-example/rust.md
MimiDumpling:
```suggestion
do that before you shut down the application.
```

src/collections/_documentation/error-reporting/configuration/drain-example/javascript.md
rhcarvalho: Hmm, in this case here "only" was used to say "if you do it, you should only do it in this circumstance (when the app is shutting down)".

Without "only", it reads:

> After calling `close`, the current client cannot be used anymore, so make sure to do that before you shut down the application.

I may be misinterpreting, I read the above as a suggestion to *always* ("make sure") call `close` before shutting down the app. And that suggestion would be incorrect.

Shall we keep "only" and delete "right"?

src/collections/_documentation/error-reporting/configuration/drain-example/native.md
rhcarvalho: Same concern re: "only".

"make sure to do something _only_ [at this given time]"

vs

"make sure to do something [at this given time]"

src/collections/_documentation/error-reporting/configuration/drain-example/native.md
rhcarvalho: Thinking about "right before" vs "before".

In this case, "before" sounds like any time before -- not so precise.
Conversely, I think the intention of "right before" was to mean "call `close` and the very next thing is the end of the app". The latter is the correct, precise suggestion.

@MimiDumpling could you elaborate on why we would remove "right"?

rhcarvalho:
> Couple edits.  Thanks for updating the docs!

Thanks @MimiDumpling!!! Amazing, thanks for your help!

I do have a couple of questions before applying the suggestions, to make sure we stay both technically correct and with good prose :)

(inline comments re: "only" and re: "right before")

src/collections/_documentation/error-reporting/configuration/drain-example/javascript.md
MimiDumpling: How about:

After calling `close`, the current client cannot be used anymore. It's important to only call `close` immediately before shutting down the application.

MimiDumpling: I think since these steps are so dependent on each other, we should avoid using the word "that" and "it" and be as explicit as possible. And the word "right" is not quite precise enough, so I changed it to "immediately." Let me know if that works. If not, we can try rewriting the sentence, again.

src/collections/_documentation/error-reporting/configuration/drain-example/native.md
MimiDumpling: How about:

After shutdown, the client cannot be used anymore. It's important to only call `sentry_shutdown()` immediately before shutting down the application.

src/collections/_documentation/error-reporting/configuration/drain-example/native.md
MimiDumpling: Made a suggestion in the comment above. 😄 

The word "right" isn't quite precise enough in this instance. "Right" has many meanings. I think a better word for this case is "immediately."

MimiDumpling:
> > Couple edits.  Thanks for updating the docs!
> 
> Thanks @MimiDumpling!!! Amazing, thanks for your help!
> 
> I do have a couple of questions before applying the suggestions, to make sure we stay both technically correct and with good prose :)
> 
> (inline comments re: "only" and re: "right before")

Happy to help! 😄 Made some alternate suggestions. Let me know if they work well with the technical accuracy. If not, we can try other sentence structures.

src/collections/_documentation/error-reporting/configuration/drain-example/javascript.md
rhcarvalho: Suggestion applied, thanks!

src/collections/_documentation/error-reporting/configuration/drain-example/javascript.md
rhcarvalho: Makes sense. I've taken your suggestion and updated the relevant sentences in the multiple files. Thanks!

src/collections/_documentation/error-reporting/configuration/drain-example/native.md
rhcarvalho: Suggestion applied, thank you!

rhcarvalho:
Applied suggestions from Mimi, merging this in.

PR #1518, opened 2020-02-28
update before send doc with error spam prevention

caleb15:
Hopefully this makes it clear to the reader that before_send can be used to prevent error spam.
Related issue: https://github.com/getsentry/sentry-python/issues/149

Thanks to @untitaker for the tip


untitaker:
Thanks!

PR #1526, opened 2020-03-03
chore: Update android wizard link

HazAT:
I hope this is correct, goal should be that the onboarding wizard uses the new android docs.

### DIFF
--- a/src/_data/platforms.yml
+++ b/src/_data/platforms.yml
@@ -93,10 +93,10 @@
       support_level: production
       type: framework
       name: Android
-      doc_link: /clients/java/modules/android/
+      doc_link: /platforms/android/
       wizard_parent: java
       wizard:
-        - _documentation/clients/java/integrations.md#android
+        - _documentation/platforms/android/index.md
 -
       slug: log4j2
       support_level: production


src/_data/platforms.yml
```diff
@@ -93,10 +93,10 @@
       support_level: production
       type: framework
       name: Android
-      doc_link: /clients/java/modules/android/
+      doc_link: /platforms/android/
       wizard_parent: java
```
marandaneto: `wizard_parent: java` I don't know what it is used for, but moving away from the java client to its own platform, this line seems wrong

src/_data/platforms.yml
```diff
@@ -93,10 +93,10 @@
...
       wizard:
-        - _documentation/clients/java/integrations.md#android
+        - _documentation/platforms/android/index.md
```
marandaneto: this file didn't have any changes, and afaik, we'd need to add something like `<!— Wizard android —>` into it.
See: https://github.com/getsentry/sentry-docs/pull/1498#discussion_r378828245
so that's the way of how the wizard would extract the right piece of code from the whole md file

src/_data/platforms.yml
HazAT: I think this is not really relevant, only for the code parts that we have spread everywhere.

PR #1544, opened 2020-03-13
Update: distributed tracing

MimiDumpling:
Updated spacing in the JavaScript and Node.js code examples.


dashed:
@MimiDumpling the node examples would also probably need to be updated with the same indentation style.

MimiDumpling:
> @MimiDumpling the node examples would also probably need to be updated with the same indentation style.

👍 On it.

PR #1556, opened 2020-03-18
fix: Update JS Docs

HazAT:


src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -154,42 +154,42 @@ def process_item(item):
 
 ### JavaScript
 
-To access our tracing features, you will need to install our Tracing integration:
+To access our tracing features, you will need to install our Tracing package `@sentry/apm`:
 
 ```bash
 $ npm install @sentry/browser
```
rhcarvalho: Isn't this line unnecessary? `@sentry/apm` depends on `@sentry/browser`.

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -154,42 +154,42 @@ def process_item(item):
...
 $ npm install @sentry/apm
 ```
 
-**Sending Traces**
+**Sending Traces/Transactions/Spans**
 
-Tracing resides in the `@sentry/apm` package. You can add it to your `Sentry.init` call:
+`Tracing` Integration resides in the `@sentry/apm` package. You can add it to your `Sentry.init` call:
```
rhcarvalho:
```suggestion
The `Tracing` integration resides in the `@sentry/apm` package. You can add it to your `Sentry.init` call:
```

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -154,42 +154,42 @@ def process_item(item):
...
 
 ```javascript
 import * as Sentry from '@sentry/browser';
-import { Integrations as ApmIntegrations } from '@sentry/apm';
+import { Integrations as AmpIntegrations } from '@sentry/apm';
```
rhcarvalho: Typo?

```suggestion
import { Integrations as ApmIntegrations } from '@sentry/apm';
```

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -154,42 +154,42 @@ def process_item(item):
...
 
 Sentry.init({
     dsn: '___PUBLIC_DSN___',
     integrations: [
-        new ApmIntegrations.Tracing(),
+        new AmpIntegrations.Tracing(),
```
rhcarvalho:
```suggestion
        new ApmIntegrations.Tracing(),
```

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -154,42 +154,42 @@ def process_item(item):
...
+    tracesSampleRate: 1, // To enable sending traces
 });
 ```
 
 To send any traces, set the `tracesSampleRate` to a nonzero value. The following configuration will capture 10% of all your transactions:
 
 ```javascript
 import * as Sentry from '@sentry/browser';
-import { Integrations as ApmIntegrations } from '@sentry/apm';
+import { Integrations as AmpIntegrations } from '@sentry/apm';
```
rhcarvalho:
```suggestion
import { Integrations as ApmIntegrations } from '@sentry/apm';
```

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -154,42 +154,42 @@ def process_item(item):
...
 
 Sentry.init({
     dsn: '___PUBLIC_DSN___',
     integrations: [
-        new Integrations.Tracing({
-            tracesSampleRate: 0.1,
-        }),
+        new AmpIntegrations.Tracing(),
```
rhcarvalho:
```suggestion
        new ApmIntegrations.Tracing(),
```

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -207,27 +209,61 @@ The tracing integration will create a [transaction]({%- link _documentation/perf
 // This line starts an activity (and creates a span).
 // As long as you don't pop the activity, the transaction will not be finished and therefore not sent to Sentry.
 // If you do not want to create a span out of an activity, just don't provide the second arg.
-const activity = ApmIntegrations.Tracing.pushActivity(displayName, {
+const activity = AmpIntegrations.Tracing.pushActivity(displayName, {
```
rhcarvalho:
```suggestion
const activity = ApmIntegrations.Tracing.pushActivity(displayName, {
```

src/collections/_documentation/performance/distributed-tracing.md
HazAT: Nope we need it

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -186,10 +187,9 @@ import { Integrations as ApmIntegrations } from '@sentry/apm';
 Sentry.init({
     dsn: '___PUBLIC_DSN___',
     integrations: [
-        new Integrations.Tracing({
-            tracesSampleRate: 0.1,
-        }),
+        new ApmIntegrations.Tracing(),
     ],
+    tracesSampleRate: 0.1,
```
dashed: @HazAT Just curious, is this the preferred placement of the configuration? Wondering because of https://github.com/getsentry/sentry/blob/3672ca751d9306fc22761f1a52847787fcef7915/src/sentry/static/sentry/app/bootstrap.jsx#L37-L40

HazAT: This will be fixed in the next update, should have never been there in the first place.

dashed: Gotcha. Thanks for clarifying 👍

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -174,6 +174,7 @@ Sentry.init({
     integrations: [
         new ApmIntegrations.Tracing(),
     ],
+    tracesSampleRate: 1, // To enable sending traces
```
rhcarvalho:
```suggestion
    tracesSampleRate: 0.1,
```

I suggest:

- Using floats to help emphasize `tracesSampleRate` is a (rational) number in the [0.0, 1.0] range.
- Never set it to `1.0`, as it is often a bad idea for production traffic. Can be useful in development for debugging.

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -280,7 +316,7 @@ const Apm = require("@sentry/apm"); // This is required since it patches functio
 
 Sentry.init({
     dsn: "___PUBLIC_DSN___",
-    tracesSampleRate: 0.1
+    tracesSampleRate: 1
```
rhcarvalho: Why change it to `1`?

untitaker: Daniel, we cannot allow anybody to send 100% of their traffic to us. This burns immediately through their quota as long as the new billing is not live.

HazAT: I leave it in the responsibility of the user to set the correct value here.
1 = they see it works
0.1 = they write into support it's not working

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -201,6 +201,20 @@ You can pass many different options to tracing, but it comes with reasonable def
...
+*tracingOrigins Option*
+ 
+The default value of `tracingOrigins` is : `new ApmIntegrations.Tracing({tracingOrigins: ['localhost', /^\//]})`. This options will attach the `sentry-trace` header to all outgoing XHR/fetch requests that contain a string in the list or match the regex. In case your frontend is making requests to a different domain you might want to add it there so we attach the `sentry-trace`.
+
+*Example:*
+
+- Your frontend is running `mywebapp.com`
+- Your backend is running `api.mybackend.com`
+- You setup your frontend to make API calls to `api.mybackend.com`
+- Then the option need to be configured like this `new ApmIntegrations.Tracing({tracingOrigins: ['localhost', /^\//, 'api.mybackend.com']})`
```
rhcarvalho: What's the `/^\//` for in this case?
Also, I'd not include `localhost`, because what's the point of sending the header to `localhost` (the page visitor's loopback device). Users may need to arrange for this config to have different values depending on environment.

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -201,6 +201,20 @@ You can pass many different options to tracing, but it comes with reasonable def
 - If available: Browser Resources (Scripts, CSS, Images ...)
 - If available: Browser Performance API Marks
 
+*tracingOrigins Option*
+ 
+The default value of `tracingOrigins` is : `new ApmIntegrations.Tracing({tracingOrigins: ['localhost', /^\//]})`. This options will attach the `sentry-trace` header to all outgoing XHR/fetch requests that contain a string in the list or match the regex. In case your frontend is making requests to a different domain you might want to add it there so we attach the `sentry-trace`.
```
rhcarvalho:
```suggestion
The default value of `tracingOrigins` is : `new ApmIntegrations.Tracing({tracingOrigins: ['localhost', /^\//]})`. The JavaScript SDK will attach the `sentry-trace` header to all outgoing XHR/fetch requests that contain a string in the list or match a regex. In case your frontend is making requests to a different domain you might want to add it there in order to propagate the `sentry-trace` header to the backend services, required to link transactions together as part of a single trace.
```

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -201,6 +201,20 @@ You can pass many different options to tracing, but it comes with reasonable def
...
+
+*Example:*
+
+- Your frontend is running `mywebapp.com`
```
rhcarvalho:
```suggestion
- A frontend application is served from `example.com`
```

`example.com` is a reserved domain for documentation, I'd be cautious using other domains we don't own or control.

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -201,6 +201,20 @@ You can pass many different options to tracing, but it comes with reasonable def
...
+- Your backend is running `api.mybackend.com`
```
rhcarvalho:
```suggestion
- A backend service is served from `api.example.com`
```

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -201,6 +201,20 @@ You can pass many different options to tracing, but it comes with reasonable def
...
+- You setup your frontend to make API calls to `api.mybackend.com`
```
rhcarvalho:
```suggestion
- The frontend application makes API calls to the backend
```

src/collections/_documentation/performance/distributed-tracing.md
rhcarvalho:
```suggestion
- Then the option need to be configured like this `new ApmIntegrations.Tracing({tracingOrigins: ['api.example.com']})`
```

rhcarvalho: Do we have a similar configuration in Python, or do we just send the header regardless of destination domain? cc @untitaker

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -201,6 +201,20 @@ You can pass many different options to tracing, but it comes with reasonable def
...
+- Then the option need to be configured like this `new ApmIntegrations.Tracing({tracingOrigins: ['localhost', /^\//, 'api.mybackend.com']})`
+- All outgoing XHR/fetch request would get the `sentry-trace` header attached
```
rhcarvalho:
```suggestion
- Outgoing XHR/fetch requests to `api.example.com` would get the `sentry-trace` header attached
```

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -201,6 +201,20 @@ You can pass many different options to tracing, but it comes with reasonable def
...
+
+*NOTE:* You need to make sure you webserver CORS is configured to allow the `sentry-trace` header. The configuration might look like this: `"Access-Control-Allow-Headers: sentry-trace"`, this depends a lot on your configuration. But if you are not whitelisting the `sentry-trace` header the request might be blocked.
```
rhcarvalho:
```suggestion
*NOTE:* You need to make sure your web server CORS is configured to allow the `sentry-trace` header. The configuration might look like this: `"Access-Control-Allow-Headers: sentry-trace"`, this depends a lot on your configuration. But if you are not whitelisting the `sentry-trace` header the request might be blocked.
```

src/collections/_documentation/performance/distributed-tracing.md
untitaker: Okay, can we then add a comment here that they should be careful in increasing this towards 1 in prod? 1 makes total sense in dev.

src/collections/_documentation/performance/distributed-tracing.md
untitaker: I would suggest adding a new bullet point before this one:

```
- In development, your API is at `localhost`
```

To explain what the `localhost` is needed for. Then they can make an informed choice on which values to include in their config

untitaker: We do it regardless of destination domain and do not let the user configure this, but I think we will not get away with either in the long-term

src/collections/_documentation/performance/distributed-tracing.md
untitaker: Suggestion to the suggestion: `, required` should prob be something like `, which is required`

src/collections/_documentation/performance/distributed-tracing.md
untitaker: FWIW this is also inconsistent with other code examples.

src/collections/_documentation/performance/distributed-tracing.md
rhcarvalho: The choice of hosts is very env specific. If you're developing on local containers, `localhost` is probably the wrong value. We can't know what the right value is for everyone, but we can explain what it needs to match and how to detect errors/ fix it.

src/collections/_documentation/performance/distributed-tracing.md
rhcarvalho: > can we then add a comment here that they should be careful in increasing this towards 1 in prod? 1 makes total sense in dev.

+1

Maybe nobody cares, but I would also find `1.0` more indicative that this is not a `0` or `1` setting, but that typical values are floats in the [0.0, 1.0] range: `0.0`, `0.1`, `0.25`...

src/collections/_documentation/performance/distributed-tracing.md
thinkocapo: @rhcarvalho I think 
it's for checking that the string 'localhost' is validated or not by the regex /^\// ?

check out the function isMatchingPattern in string.ts
```
// tracing.ts
            const isWhitelisted = self._options.tracingOrigins.some((origin: string | RegExp) =>
              isMatchingPattern(url, origin),
            );
```

looks like you can also write ['localhost',  'string']?
```
// string.ts
export function isMatchingPattern(value: string, pattern: RegExp | string): boolean {
  if (isRegExp(pattern)) {
    return (pattern as RegExp).test(value);
  }
  if (typeof pattern === 'string') {
    return value.includes(pattern);
  }
  return false;
}
```

but i'm a little confused because:
```
reg = /^\// 
reg.test('foobar')
false
reg.test('foo/bar')
false
```

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -210,24 +226,58 @@ The tracing integration will create a [transaction]({%- link _documentation/perf
...
     data: {},
     op: 'react',
-    description: `<${displayName}>`,
+    description: `${displayName}`,
 });
 
 // Sometime later ...
 // When we pop the activity, the Integration will finish the span and after the timeout finish the transaction and send it to Sentry
-Integrations.ApmIntegrations.popActivity(activity);
+// Keep in mind, as long as you do not pop the activity, the transaction will be kept alive and not sent to Sentry
```
rhcarvalho: We have now two sentences ending "send to Sentry" and missing punctuation.
Maybe the two sentences could be combined together better?

What information are we trying to convey?

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -210,24 +226,58 @@ The tracing integration will create a [transaction]({%- link _documentation/perf
...
+ApmIntegrations.Tracing.popActivity(activity);
+```
+
+Keep in mind, if there is no active transaction you need to create one before otherwise nothing will happen.
```
rhcarvalho: I think here we need to be explicit about what "before" refers to.

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -210,24 +226,58 @@ The tracing integration will create a [transaction]({%- link _documentation/perf
...
+So given a different example where you want to create an Transaction for a user interaction on you page you need to do the following:
+
+
+```javascript
+// Let's say this function is invoked when a user clicks on the checkout button of your shop
+shopCheckout() {
+    // This will create a new Transaction for you
+    ApmIntegrations.Tracing.startIdleTransaction('shopCheckout');
```
rhcarvalho: Why here we use `ApmIntegrations.Tracing.startIdleTransaction` instead of `hub.startSpan`?

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -210,24 +226,58 @@ The tracing integration will create a [transaction]({%- link _documentation/perf
...
+        data: {
+            result
+        },
+        op: 'task',
+        description: `processing shopping cart result`,
+    });
+    processAndValidateShoppingCart(result);
+    ApmIntegrations.Tracing.popActivity(activity);
+        
+    ApmIntegrations.Tracing.finshIdleTransaction(); // This is optional
```
rhcarvalho:
```suggestion
    ApmIntegrations.Tracing.finishIdleTransaction(); // This is optional
```

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -210,24 +226,58 @@ The tracing integration will create a [transaction]({%- link _documentation/perf
...
+shopCheckout() {
+    // This will create a new Transaction for you
+    ApmIntegrations.Tracing.startIdleTransaction('shopCheckout');
+    
+    const result = validateShoppingCartOnServer(); // Consider this function making an xhr/fetch call
+    const activity = ApmIntegrations.Tracing.pushActivity('task', {
+        data: {
+            result
+        },
+        op: 'task',
```
rhcarvalho: Why do we need a `name` (first arg) for an Activity? Why would people need to write an activity name and then `op` (which is also some sort of activity/operation name)?!

src/collections/_documentation/performance/distributed-tracing.md
thinkocapo: @rhcarvalho nevermind, the regex is checking that the string doesn't begin with a forward slash.

src/collections/_documentation/performance/distributed-tracing.md
thinkocapo: @HazAT 
maybe keep it at 1.0 but then put a comment before or after the code snippet that says,
"We recommend 0.1 for when you put this to production [because...]"

maybe keep it at 0.1 but then put a comment before or after the code snippet that says,
"You may want this on 1.0 if you're testing it for the first time"
"You may want this on 1.0 if you're testing it for the first time or else you might think your transactions aren't coming through"

you're 100% correct about them writing into support, as I tricked myself on this the first time I set it up, and then again a month later when I re-visisted it. So that makes me prefer to see it at 1.0 by default.

src/collections/_documentation/performance/distributed-tracing.md
HazAT: Because we prefer ppl using the Integration over the `startSpan`
The integration also adds spans for XHR/fetch

src/collections/_documentation/performance/distributed-tracing.md
HazAT: It's there if you want to have an activity that is "blocking" the idle transaction but you do not want to create a span out of it.

PR #1558, opened 2020-03-18
fix(python) Use exception instead of error

markstory:
Fixes getsentry/sentry#17751


untitaker:
@markstory that should not matter though. `exception` is just an alias for `fatal` (one log level higher than error) with `exc_info=True`

PR #1560, opened 2020-03-18
doc(python): Add code sample for hub propagation

untitaker:


src/collections/_documentation/platforms/python/default-integrations.md
```diff
@@ -78,3 +78,47 @@ See [_Logging_]({% link _documentation/platforms/python/logging.md %})
...
+# We take all context data (the tags map and even the entire client
+# configuration), and pass it as explicit variable
+# into the thread.
+thread_hub = Hub(Hub.current)
+
+tr = threading.Thread(target=run, args=[thread_hub])
+tr.start()
+tr.join()
+
+### Example B: Automatic propagation
```
markstory: Do we often use headings inside code blocks? Or should the previous block and and a new one begin around this heading?

untitaker: fixed... wanted to share imports but it's cleaner this way

markstory (approved):

PR #1561, opened 2020-03-19
fix typo in react documentation

meain:
Local files use relative import


kamilogorek:
It's very common to configure module resolution so that it can resolve this path just fine, but I agree that docs should show the simplest possible scenario.

kamilogorek:
Although in this case it should rather say `./App` anyway, as `index.js` is usually in the same directory as `App.js`

meain:
Yeah, I think that makes sense. I will switch it to `./App` or should it be `./App.js`?

kamilogorek:
I prefer a form without an extension

meain:
Done!

kamilogorek:
Thanks!

PR #1562, opened 2020-03-19
ref(pii): Better explanation of selectors

untitaker:
Also remove some useless value types that I might want to remove: booleans are never stripped, ever, and the event value type does not work at the moment.

### DIFF
--- a/src/collections/_documentation/data-management/advanced-datascrubbing.md
+++ b/src/collections/_documentation/data-management/advanced-datascrubbing.md
@@ -125,13 +125,16 @@ You can combine selectors using boolean logic.
 
 ### Value Types
 
-Select subsections by JSON-type or semantic meaning using the following:
+Select subsections by JSON-type using the following:
 
 * `$string` matches any string value
 * `$number` matches any integer or float value
 * `$datetime` matches any field in the event that represents a timestamp
 * `$array` matches any JSON array value
 * `$object` matches any JSON object
+
+Select known parts of the schema using the following:
+
 * `$exception` matches a single exception instance in `{"exception": {"values": [...]}}`
 * `$stacktrace` matches a stack trace instance
 * `$frame` matches a frame
--- a/src/collections/_documentation/data-management/advanced-datascrubbing.md
+++ b/src/collections/_documentation/data-management/advanced-datascrubbing.md
@@ -93,7 +93,7 @@ Another example. Sentry knows about two kinds of error messages: The exception m
   "logentry": {
     "formatted": "Failed to roll out the dinglebop"
   },
-  "exceptions": {
+  "exception": {
     "values": [
       {
         "type": "ZeroDivisionError",
--- a/src/collections/_documentation/data-management/advanced-datascrubbing.md
+++ b/src/collections/_documentation/data-management/advanced-datascrubbing.md
@@ -70,6 +70,7 @@ Sentry does not know if a local variable that looks like a creditcard number act
 {%- include components/alert.html
   title="Sentry does not know what your code does"
   content=__alert_content
+  level="warning"
 %}
 
 
--- a/src/collections/_documentation/data-management/advanced-datascrubbing.md
+++ b/src/collections/_documentation/data-management/advanced-datascrubbing.md
@@ -64,7 +64,7 @@ Rules generally consist of three parts:
 
 {% capture __alert_content -%}
 
-Sentry does not know if a local variable that looks like a creditcard number actually is one. As such you need to expect false-positives but also false-negatives. [_Selectors_](#selectors) can help you in limiting the scope in which your rule runs.
+Sentry does not know if a local variable that looks like a credit card number actually is one. As such, you need to expect not only false-positives but also false-negatives. [_Selectors_](#selectors) can help you in limiting the scope in which your rule runs.
 
 {%- endcapture -%}
 {%- include components/alert.html
@@ -86,7 +86,7 @@ For example, what is called "Additional Data" in the UI is called `extra` in the
 [Remove] [Anything] from [extra.foo]
 ```
 
-Another example. Sentry knows about two kinds of error messages: The exception message, and the top-level log message. Here is an example of how such an event payload as sent by the SDK (and downloadable from the UI) would look like:
+Another example. Sentry knows about two kinds of error messages: the exception message, and the top-level log message. Here is an example of how such an event payload as sent by the SDK (and downloadable from the UI) would look like:
 
 ```json
 {
@@ -141,9 +141,9 @@ Select known parts of the schema using the following:
 * `$frame` matches a frame
 * `$request` matches the HTTP request context of an event
 * `$user` matches the user context of an event
-* `$logentry` (also applies to the `message` attribute)
+* `$logentry` matches both the `logentry` attribute of an event as well as the `message` attribute
 * `$thread` matches a single thread instance in `{"threads": {"values": [...]}}`
-* `$breadcrumb` matches a single breadcrumb in `{"breadcrumbs": [...]}`
+* `$breadcrumb` matches a single breadcrumb in `{"breadcrumbs": {"values": [...]}}`
 * `$span` matches a [trace span]({% link _documentation/performance/performance-glossary.md %}#span)
 * `$sdk` matches the SDK context in `{"sdk": ...}`
 
--- a/src/collections/_documentation/data-management/advanced-datascrubbing.md
+++ b/src/collections/_documentation/data-management/advanced-datascrubbing.md
@@ -133,7 +133,7 @@ Select subsections by JSON-type or semantic meaning using the following:
 * `$array` matches any JSON array value
 * `$object` matches any JSON object
 * `$exception` matches a single exception instance in `{"exception": {"values": [...]}}`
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/data-management/advanced-datascrubbing.md
```diff
@@ -62,17 +62,59 @@ Rules generally consist of three parts:
 - _MAC Addresses_
 - _Anything_: Matches any value. This is useful if you want to remove a certain JSON key by path using [_Selectors_](#selectors) regardless of the value.
 
+{% capture __alert_content -%}
+
+Sentry does not know if a local variable that looks like a creditcard number actually is one. As such you need to expect false-positives but also false-negatives. [_Selectors_](#selectors) can help you in limiting the scope in which your rule runs.
```
rhcarvalho:
```suggestion
Sentry does not know if a local variable that looks like a credit card number actually is one. As such, you need to expect not only false-positives but also false-negatives. [_Selectors_](#selectors) can help you in limiting the scope in which your rule runs.
```

src/collections/_documentation/data-management/advanced-datascrubbing.md
```diff
@@ -62,17 +62,59 @@ Rules generally consist of three parts:
...
+
+{%- endcapture -%}
+{%- include components/alert.html
+  title="Sentry does not know what your code does"
+  content=__alert_content
```
rhcarvalho:
```suggestion
  content=__alert_content
  level="warning"
```

Not sure about this one, but I've seen and used warning for notes like this. Here's the difference rendered:

image

src/collections/_documentation/data-management/advanced-datascrubbing.md
```diff
@@ -62,17 +62,59 @@ Rules generally consist of three parts:
...
+
+Data scrubbing always works on the raw event payload. Keep in mind that some fields in the UI may be called differently in the JSON schema. When looking at an event there should always be a link called "JSON" present that allows you to see what the data scrubber sees.
+
+For example, what is called "Additional Data" in the UI is called `extra` in the event payload. To remove a specific key called `foo`, you would write:
+
+```
+[Remove] [Anything] from [extra.foo]
+```
+
+Another example. Sentry knows about two kinds of error messages: The exception message, and the top-level log message. Here is an example of how such an event payload as sent by the SDK (and downloadable from the UI) would look like:
```
rhcarvalho:
```suggestion
Another example. Sentry knows about two kinds of error messages: the exception message, and the top-level log message. Here is an example of how such an event payload as sent by the SDK (and downloadable from the UI) would look like:
```

src/collections/_documentation/data-management/advanced-datascrubbing.md
```diff
@@ -62,17 +62,59 @@ Rules generally consist of three parts:
...
+
+```json
+{
+  "logentry": {
+    "formatted": "Failed to roll out the dinglebop"
+  },
+  "exceptions": {
```
rhcarvalho:
```suggestion
  "exception": {
```

I think unlike other "interfaces", this one uses the singular form for the key. Please double-check.

In any case, needs to match the rule below.

src/collections/_documentation/data-management/advanced-datascrubbing.md
```diff
@@ -83,32 +125,33 @@ You can combine selectors using boolean logic.
...
-* `$breadcrumb`
-* `$span`
-* `$sdk`
-
-Examples:
+Select subsections by JSON-type using the following:
+
+* `$string` matches any string value
+* `$number` matches any integer or float value
+* `$datetime` matches any field in the event that represents a timestamp
```
rhcarvalho: Would it make sense to talk about formats here?

A timestamp could be RFC3399, a numeric value (integer or float) representing time since the UNIX Epoch, or a myriad of actual formats.

src/collections/_documentation/data-management/advanced-datascrubbing.md
```diff
@@ -83,32 +125,33 @@ You can combine selectors using boolean logic.
...
+* `$object` matches any JSON object
+
+Select known parts of the schema using the following:
+
+* `$exception` matches a single exception instance in `{"exception": {"values": [...]}}`
+* `$stacktrace` matches a stack trace instance
+* `$frame` matches a frame
+* `$request` matches the HTTP request context of an event
+* `$user` matches the user context of an event
+* `$logentry` (also applies to the `message` attribute)
```
rhcarvalho: This line doesn't match the format used in the other lines around it. Suggestion:

```suggestion
* `$logentry` matches both the `logentry` attribute of an event as well as the `message` attribute
```

By the way, looking at the [source code in Relay](https://github.com/getsentry/relay/blob/31593da8c59876434406c6dd00007b9b71674847/relay-general/src/protocol/event.rs#L236-L239) I have the impression that `message` is an alias to `logentry`.

Perhaps we shall not document it here? That was the decision when I was updating dev docs on interfaces.

src/collections/_documentation/data-management/advanced-datascrubbing.md
```diff
@@ -83,32 +125,33 @@ You can combine selectors using boolean logic.
...
+* `$thread` matches a single thread instance in `{"threads": {"values": [...]}}`
+* `$breadcrumb` matches a single breadcrumb in `{"breadcrumbs": [...]}`
```
rhcarvalho:
```suggestion
* `$breadcrumb` matches a single breadcrumb in `{"breadcrumbs": {"values": [...]}}`
```

Since I looked at this recently, I remember that `"exception"`, `"threads"` and `"breadcrumbs"` are the 3 fields that serialize with a `"values"` list 😄

src/collections/_documentation/data-management/advanced-datascrubbing.md
```diff
@@ -83,32 +125,33 @@ You can combine selectors using boolean logic.
...
+* `$span` matches a [trace span]({% link _documentation/performance/performance-glossary.md %}#span)
```
rhcarvalho: It's a pity that Jekyll doesn't support checking the reference within the target page, we indeed have to construct the link to `#span` the way it is here, and accept it may become broken without breaking the build (I admit, I don't even know if a bad `{% link %}` would break the build either...).

https://jekyllrb.com/docs/liquid/tags/#link

src/collections/_documentation/data-management/advanced-datascrubbing.md
```diff
@@ -117,7 +160,7 @@ Examples:
   [Remove] [Anything] from [$frame.vars]
   ```
 
-### Escaping Specal Characters
+### Escaping Special Characters
```
rhcarvalho: I hope you don't mind I took a chance to fix an old typo here :)

src/collections/_documentation/data-management/advanced-datascrubbing.md
```diff
@@ -62,17 +62,59 @@ Rules generally consist of three parts:
...
 
-You can select a region of the event using JSON-path-like syntax. As an example, to delete a specific key in "Additional Data", you would configure:
+Selectors allow you to restrict rules to certain parts of the event. This is useful to unconditionally remove certain data by variable/field name from the event, but can also be used to conservatively test rules on real data.
+
+Data scrubbing always works on the raw event payload. Keep in mind that some fields in the UI may be called differently in the JSON schema. When looking at an event there should always be a link called "JSON" present that allows you to see what the data scrubber sees.
+
+For example, what is called "Additional Data" in the UI is called `extra` in the event payload. To remove a specific key called `foo`, you would write:
+
+```
+[Remove] [Anything] from [extra.foo]
```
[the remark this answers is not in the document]
untitaker: to all of them

src/collections/_documentation/data-management/advanced-datascrubbing.md
untitaker: That does not pattern-match but matches against *locations* we understand as timestamp (so behind the scenes it's a set of json paths, not a regex)

src/collections/_documentation/data-management/advanced-datascrubbing.md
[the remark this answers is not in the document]
rhcarvalho: Removing timestamps from spans would be catastrophic :)

src/collections/_documentation/data-management/advanced-datascrubbing.md
rhcarvalho: Oh, yes. And that's actually what the description says. Disregard my question.

src/collections/_documentation/data-management/advanced-datascrubbing.md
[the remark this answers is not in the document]
untitaker: We have some safety rails for this. Ideally they are so unobstrusive that we will never need to document them, but let's see :D

src/collections/_documentation/data-management/advanced-datascrubbing.md
untitaker: check out `bin/link-check`, unfortunately it is extremely slow as it goes over HTML output and is still Ruby

rhcarvalho: Interesting, thanks for sharing!

PR #1572, opened 2020-03-26
Added --dist argument

charliemday:
My understanding is if the --dist doesn't get explicitly declared then it'll just be None which can present problems with the deminification of source maps

see merge where this was added: https://github.com/getsentry/sentry-cli/commit/ee81633876c1f19b27724dcb70cadb88ff05e7e2


kamilogorek:
Thanks!

PR #1577, opened 2020-03-31
fix: Update datascrubbers documentation based on feedback

untitaker:
* Make writing more verbose to steer users into the right direction in settings. A tester expected Data Privacy Rules to be the sidebar tab's name (after skimming over the first section)
* Use new aliases from https://github.com/getsentry/relay/pull/532
* Bring back replace, see https://github.com/getsentry/relay/pull/542

Note: Do not deploy until relay has been released and bumped


untitaker:
Blocked on https://github.com/getsentry/sentry/pull/17999

PR #1580, opened 2020-04-02
fix(pii): Add more examples for selectors

untitaker:


src/collections/_documentation/data-management/advanced-datascrubbing.md
```diff
@@ -76,7 +76,22 @@ Sentry does not know if a local variable that looks like a credit card number ac
 
 ## Selectors
 
-Selectors allow you to restrict rules to certain parts of the event. This is useful to unconditionally remove certain data by variable/field name from the event, but can also be used to conservatively test rules on real data.
+Selectors allow you to restrict rules to certain parts of the event. This is useful to unconditionally remove certain data by variable/field name from the event, but can also be used to conservatively test rules on real data. A few examples:
+
+* `**` to scrub everywhere
```
rhcarvalho, line 81:
```suggestion
* `**` to scrub everything
```

Just like `extra.**` is 
> to scrub everything in "Additional Data"

src/collections/_documentation/data-management/advanced-datascrubbing.md
```diff
@@ -76,7 +76,22 @@ Sentry does not know if a local variable that looks like a credit card number ac
...
+* `$error.value` to scrub in the exception message
+* `$message` to scrub the event-level log message
+* `extra.'My Value'` to scrub the key `My Value` in "Additional Data"
```
rhcarvalho, line 84: Should we call out that the quotes are required to group together custom fields that have spaces or other special characters? (If that is actually the case)

src/collections/_documentation/data-management/advanced-datascrubbing.md
```diff
@@ -76,7 +76,22 @@ Sentry does not know if a local variable that looks like a credit card number ac
...
+* `extra.**` to scrub everything in "Additional Data"
```
rhcarvalho, line 85: Why is this `extra` and not `$extra` like `$error` and `$message`?

src/collections/_documentation/data-management/advanced-datascrubbing.md
```diff
@@ -76,7 +76,22 @@ Sentry does not know if a local variable that looks like a credit card number ac
...
+* `$http.headers.x-custom-token` to scrub the request header `X-Custom-Token`
+* `$user.ip_address` to scrub the user's IP address
+* `$frame.vars.foo` to scrub a stacktrace frame variable called `foo`
+* `contexts.device.timezone` to scrub a key from the Device context
+* `tags.server_name` to scrub the tag `server_name`
```
rhcarvalho, line 90: Similarly here, why not `$contexts` and `$tags`?

src/collections/_documentation/data-management/advanced-datascrubbing.md
```diff
@@ -76,7 +76,22 @@ Sentry does not know if a local variable that looks like a credit card number ac
 
 ## Selectors
 
-Selectors allow you to restrict rules to certain parts of the event. This is useful to unconditionally remove certain data by variable/field name from the event, but can also be used to conservatively test rules on real data.
+Selectors allow you to restrict rules to certain parts of the event. This is useful to unconditionally remove certain data by variable/field name from the event, but can also be used to conservatively test rules on real data. A few examples:
```
rhcarvalho, line 79:
```suggestion
Selectors allow you to restrict rules to certain parts of the event. This is useful to unconditionally remove certain data by attribute name from the event, but can also be used to conservatively test rules on real data. A few examples:
```

I see we use "attribute" a lot of times in the docs. I'm good with either "field" or "attribute", however I imagine that "variable" is not very precise. WDYT?

src/collections/_documentation/data-management/advanced-datascrubbing.md
untitaker: We do this further down (outside of this diff). This is only supposed to be a cheatsheet of some sort.

src/collections/_documentation/data-management/advanced-datascrubbing.md
untitaker: Because `$error` is a shortcut for `exception.*.values` and `$message` is a shortcut for `logentry.formatted` while `extra` is already short enough. There's a section further down which explains what `$error` expands to, but this section is just here for copy-pasting, not for deep understanding.

src/collections/_documentation/data-management/advanced-datascrubbing.md
untitaker: Let's do that

src/collections/_documentation/data-management/advanced-datascrubbing.md
rhcarvalho: Found it, sorry I should have looked at the whole file.

[#escaping-special-characters](https://github.com/getsentry/sentry-docs/blob/aeb3af72cf77460990506cc2e969677fcfb57fa1/src/collections/_documentation/data-management/advanced-datascrubbing.md#escaping-special-characters)

In some cases, like in the Breadcrumbs payload docs, we add intra-page links. In this case it might not be worth it, would probably mean the description here gets longer to include a descriptive link.

src/collections/_documentation/data-management/advanced-datascrubbing.md
rhcarvalho: Talked a bit off-GH about this.

Perhaps we can be more explicit about long forms and shortcuts ($ prefix).

Maybe we can turn this into a table?!


long form | short form | explanation of what it does
----------|------------|--------------------------
`exception.values.*.value` | `$error.value` | scrub error/exception message
... | ... | ...

PR #1581, opened 2020-04-02
docs explaining about current, crashed, handled and level fields

marandaneto:
it's often confusing to know either when a crash means your application has been terminated or not.
the terminology often means different things in different programming languages.
on this PR I try to make a bit more user friendly.

### DIFF
--- a/src/collections/_documentation/development/sdk-dev/event-payloads/exception.md
+++ b/src/collections/_documentation/development/sdk-dev/event-payloads/exception.md
@@ -85,7 +85,7 @@ mechanism-specific values.
 
 : Optional flag indicating whether the user has handled the exception
   (for example, via `try ... catch`).
-  When this flag is set to `false`, it probably means that the Application has been terminated.
+   Where applicable, when this flag is set to `false` it indicates the application terminated.
 
 `meta`:
 
--- a/src/collections/_documentation/development/sdk-dev/event-payloads/exception.md
+++ b/src/collections/_documentation/development/sdk-dev/event-payloads/exception.md
@@ -85,6 +85,7 @@ mechanism-specific values.
 
 : Optional flag indicating whether the user has handled the exception
   (for example, via `try ... catch`).
+  When this flag is set to `false`, it probably means that the Application has been terminated.
 
 `meta`:
 
--- a/src/collections/_documentation/development/sdk-dev/event-payloads/exception.md
+++ b/src/collections/_documentation/development/sdk-dev/event-payloads/exception.md
@@ -85,7 +85,7 @@ mechanism-specific values.
 
 : Optional flag indicating whether the user has handled the exception
   (for example, via `try ... catch`).
-   Where applicable, when this flag is set to `false` it indicates the application terminated.
+   Where applicable, when you set this flag to `false`, it indicates the application terminated.
 
 `meta`:
 
--- a/src/collections/_documentation/development/sdk-dev/event-payloads/index.md
+++ b/src/collections/_documentation/development/sdk-dev/event-payloads/index.md
@@ -143,7 +143,7 @@ highly encouraged:
   -   `info`
   -   `debug`
 
-  When this field is set to `fatal`, it probably means that the application has been terminated.
+  When this field is set to `fatal` it indicates the application terminated.
 
 `transaction`
 
--- a/src/collections/_documentation/development/sdk-dev/event-payloads/index.md
+++ b/src/collections/_documentation/development/sdk-dev/event-payloads/index.md
@@ -143,7 +143,7 @@ highly encouraged:
   -   `info`
   -   `debug`
 
-  When this field is set to `fatal`, it probably means that the Application has been terminated.
+  When this field is set to `fatal`, it probably means that the application has been terminated.
 
 `transaction`
 
--- a/src/collections/_documentation/development/sdk-dev/event-payloads/index.md
+++ b/src/collections/_documentation/development/sdk-dev/event-payloads/index.md
@@ -143,7 +143,7 @@ highly encouraged:
   -   `info`
   -   `debug`
 
-  When this field is set to `fatal` it indicates the application terminated.
+  When you set this field to `fatal`, it indicates the application terminated.
 
 `transaction`
 
--- a/src/collections/_documentation/development/sdk-dev/event-payloads/index.md
+++ b/src/collections/_documentation/development/sdk-dev/event-payloads/index.md
@@ -143,6 +143,8 @@ highly encouraged:
   -   `info`
   -   `debug`
 
+  When this field is set to `fatal`, it probably means that the Application has been terminated.
+
 `transaction`
 
 : The name of the transaction which caused this exception.
--- a/src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
+++ b/src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
@@ -34,7 +34,7 @@ When you set this flag to `true`, it's also used to preselect a thread in the th
 `current`:
 
 : _Optional_. A flag indicating whether the thread was in the foreground, often known as UI or main thread. Defaults to `false`.
-It may be seen as `was active: yes` in the Issues page.
+It may display as `was active: yes` on the Issues page.
 
 `name`:
 
--- a/src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
+++ b/src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
@@ -29,6 +29,7 @@ exception and Sentry will connect the two.
 
 : _Optional_. A flag indicating whether this thread has caused the error. It does not necessarily indicate the application terminated. Defaults to `false`.
 It may be seen as `crashed: yes` in the Issues page.
+When this flag is set to `true` it's also used to preselect a thread in the thread's selector on the Issues page.
 
 `current`:
 
--- a/src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
+++ b/src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
@@ -27,12 +27,13 @@ exception and Sentry will connect the two.
 
 `crashed`:
 
-: _Optional_. A flag indicating whether the thread crashed. Defaults to `false`.
+: _Optional_. A flag indicating whether the thread has caused the error, it doesn't mean that the Application has been terminated, but captured the error. Defaults to `false`.
+It may be seen as `crashed: yes` in the Issues page.
 
 `current`:
 
-: _Optional_. A flag indicating whether the thread was in the foreground.
-  Defaults to `false`.
+: _Optional_. A flag indicating whether the thread was in the foreground, often known as UI or main thread. Defaults to `false`.
+It may be seen as `was active: yes` in the Issues page.
 
 `name`:
 
--- a/src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
+++ b/src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
@@ -28,7 +28,7 @@ exception and Sentry will connect the two.
 `crashed`:
 
 : _Optional_. A flag indicating whether this thread has caused the error. It does not necessarily indicate the application terminated -- defaults to `false`.
-It may be seen as `crashed: yes` in the Issues page.
+It may display as `crashed: yes` on the Issues page.
 When this flag is set to `true` it's also used to preselect a thread in the thread's selector on the Issues page.
 
 `current`:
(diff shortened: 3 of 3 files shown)


src/collections/_documentation/development/sdk-dev/event-payloads/index.md
```diff
@@ -143,6 +143,8 @@ highly encouraged:
   -   `info`
   -   `debug`
 
+  When this field is set to `fatal`, it probably means that the Application has been terminated.
```
bruno-garcia, line 146:
```suggestion
  When this field is set to `fatal`, it indicates that the application terminated.
```

lobsterkatie, line 146:
```suggestion
  When this field is set to `fatal`, it probably means that the application has been terminated.
```

src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
```diff
@@ -27,12 +27,13 @@ exception and Sentry will connect the two.
 
 `crashed`:
 
-: _Optional_. A flag indicating whether the thread crashed. Defaults to `false`.
+: _Optional_. A flag indicating whether the thread has caused the error, it doesn't mean that the Application has been terminated, but captured the error. Defaults to `false`.
```
bruno-garcia, line 30:
```suggestion
: _Optional_. A flag indicating whether this thread has caused the error. It does not necessarily indicate the application terminated. Defaults to `false`.
```

src/collections/_documentation/development/sdk-dev/event-payloads/exception.md
```diff
@@ -85,6 +85,7 @@ mechanism-specific values.
 
 : Optional flag indicating whether the user has handled the exception
   (for example, via `try ... catch`).
+  When this flag is set to `false`, it probably means that the Application has been terminated.
```
bruno-garcia, line 88:
```suggestion
   Where applicable, when this flag is set to `false` it indicates the application terminated.
```
Perhaps we could go futher and give examples like, JS event doesn't mean the app crashed since it's a webpage on the browser. On a native C++ application, that would terminate the app.

src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
```diff
@@ -27,12 +27,13 @@ exception and Sentry will connect the two.
...
+It may be seen as `crashed: yes` in the Issues page.
```
lobsterkatie:
```suggestion
: _Optional_. A flag indicating whether this thread is the one which captured the event sent to Sentry. In spite of the name, it doesn't necessarily indicate that the application terminated. Defaults to `false`, and should only be set to `true` on one thread.
On the Issue Details page, appears as a `crashed: yes` tag on the thread in question, and causes that thread to be the one shown by default.
```

MimiDumpling:
@marandaneto thanks for updating. Either @PeloWriter or I will take a look when we get a chance. 🌻

src/collections/_documentation/development/sdk-dev/event-payloads/index.md
```diff
@@ -143,6 +143,8 @@ highly encouraged:
   -   `info`
   -   `debug`
 
+  When this field is set to `fatal` it indicates the application terminated.
```
rhcarvalho, line 146: I saw that @bruno-garcia suggested changing from `probably means` to `indicates`.

Since users are eventually in control of levels (*), I'd tend to something less definitive as this depends on the interpretation of parts of code that are out of our control.

I understand that for many SDKs there is no way to assert that the application terminated, because the SDK runs in the same process as the application. E.g. JS Node, JS Browser, Python, PHP, Rust, Go, ...
At most such SDKs may "think" that the application will terminate after the event is reported.

Other than changing icons in the UI, I find it hard to carry any more semantical value from levels. Both because we never explicitly said what each level is supposed to mean and implemented SDKs with that in mind, and because different languages and frameworks carry different semantics (and often choose to have other levels like "critical", "trace", etc)

What do you think?

---

(*) SDKs do set default levels depending on circumstances. E.g. the Go SDK uses `info` for `CaptureMessage`, `error` for `CaptureException` and `fatal` when it recovers from a `panic` (the program may or may not terminate -- depends on the configuration whether to "repanic" or not and whether something else recovers from the panic later).

untitaker: The added sentence is not true for Python at all, and also not the majority case. `fatal` is just another log level.

src/collections/_documentation/development/sdk-dev/event-payloads/exception.md
```diff
@@ -85,6 +85,7 @@ mechanism-specific values.
 
 : Optional flag indicating whether the user has handled the exception
   (for example, via `try ... catch`).
+   Where applicable, when this flag is set to `false` it indicates the application terminated.
```
MimiDumpling, line 88:
```suggestion
   Where applicable, when you set this flag to `false`, it indicates the application terminated.
```

src/collections/_documentation/development/sdk-dev/event-payloads/index.md
MimiDumpling, line 146:
```suggestion
  When you set this field to `fatal`, it indicates the application terminated.
```

src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
```diff
@@ -27,12 +27,14 @@ exception and Sentry will connect the two.
 
 `crashed`:
 
-: _Optional_. A flag indicating whether the thread crashed. Defaults to `false`.
+: _Optional_. A flag indicating whether this thread has caused the error. It does not necessarily indicate the application terminated. Defaults to `false`.
```
MimiDumpling, line 30:
```suggestion
: _Optional_. A flag indicating whether this thread has caused the error. It does not necessarily indicate the application terminated -- defaults to `false`.
```

src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
```diff
@@ -27,12 +27,14 @@ exception and Sentry will connect the two.
...
+It may be seen as `crashed: yes` in the Issues page.
```
MimiDumpling, line 31:
```suggestion
It may display as `crashed: yes` on the Issues page.
```

src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
```diff
@@ -27,12 +27,14 @@ exception and Sentry will connect the two.
...
+When this flag is set to `true` it's also used to preselect a thread in the thread's selector on the Issues page.
```
MimiDumpling, line 32:
```suggestion
When you set this flag to `true`, it's also used to preselect a thread in the thread's selector on the Issues page.
```

src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
```diff
@@ -27,12 +27,14 @@ exception and Sentry will connect the two.
...
 
 `current`:
 
-: _Optional_. A flag indicating whether the thread was in the foreground.
-  Defaults to `false`.
+: _Optional_. A flag indicating whether the thread was in the foreground, often known as UI or main thread. Defaults to `false`.
```
MimiDumpling, line 36:
```suggestion
: _Optional_. A flag indicating whether the thread was in the foreground, often known as UI or main thread -- defaults to `false`.
```

src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
```diff
@@ -27,12 +27,14 @@ exception and Sentry will connect the two.
...
+It may be seen as `was active: yes` in the Issues page.
```
MimiDumpling, line 37:
```suggestion
It may display as `was active: yes` on the Issues page.
```

src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
rhcarvalho: Both options here ("may be seen" and "may display") suggest that it may or may not display like that. Is there a case where the event has `crashed: true` and doesn't display `crashed: yes` in Sentry?!

Why do we say "may"?

rhcarvalho: If the usage of "may" was future-proofing in case of UI changes, then maybe we should not mention it altogether?

MimiDumpling: Great question! I was assuming that we included the word "may" because it was uncertain? Happy to take your lead on this and omit the word "may." @marandaneto do you know if there's a reason to emphasize "may?"

src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
```diff
@@ -27,12 +27,14 @@ exception and Sentry will connect the two.
 
 `crashed`:
 
-: _Optional_. A flag indicating whether the thread crashed. Defaults to `false`.
+: _Optional_. A flag indicating whether this thread has caused the error. It does not necessarily indicate the application terminated -- defaults to `false`.
+It may display as `crashed: yes` on the Issues page.
```
marandaneto, line 31:
```suggestion
It displays as `crashed: yes` on the Issues page.
```

src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
```diff
@@ -27,12 +27,14 @@ exception and Sentry will connect the two.
...
+: _Optional_. A flag indicating whether this thread has caused the error. It does not necessarily indicate the application terminated -- defaults to `false`.
+It displays as `crashed: yes` on the Issues page.
+When you set this flag to `true`, it's also used to preselect a thread in the thread's selector on the Issues page.
 
 `current`:
 
-: _Optional_. A flag indicating whether the thread was in the foreground.
-  Defaults to `false`.
+: _Optional_. A flag indicating whether the thread was in the foreground, often known as UI or main thread -- defaults to `false`.
+It may display as `was active: yes` on the Issues page.
```
marandaneto, line 37:
```suggestion
It displays as `was active: yes` on the Issues page.
```

src/collections/_documentation/development/sdk-dev/event-payloads/threads.md
marandaneto: @MimiDumpling tbh no, it's just a synonym of `can` that didn't work well, so the meaning would be: `can be seen/it can be viewed`, I've changed to `it displays as...` :)

marandaneto: @bruno-garcia true, we can talk about this on TSC again, I generally agree, but still think the current status of docs would need to be changed, either by this PR or other, this is a very recurrent problem.
I'll keep this PR open then.

src/collections/_documentation/development/sdk-dev/event-payloads/index.md
marandaneto: @rhcarvalho I generally agree, but I also think this should be a guideline to be followed, so I think it should be explicit what you need to set to differentiate a hardcrash and a non-hardcrash event.
Right now people don't know how to set nor to filter it.
Let's discuss this over TSC again.

src/collections/_documentation/development/sdk-dev/event-payloads/exception.md
marandaneto: @bruno-garcia I'm all for it, let's discuss it again on TSC about which languages should be the examples and happy to change.

src/collections/_documentation/development/sdk-dev/event-payloads/index.md
untitaker: What about saying "if this event was created from a log record, this indicates the level with which that message was logged". This delegates the meaning and is still helpful

rhcarvalho: > "if this event was created from a log record, this indicates the level with which that message was logged"

That is better indeed. Unfortunately, it is not always true, as depending on the logging framework the level has to be mapped from some original value to one of those values we accept.

Could be made always true if we accepted arbitrary strings as level, at the moment we don't:

image

The invalid value gets overwritten as `level: error`:

image

rhcarvalho: If we don't relax the requirement, we can relax the docs... tried to rewrite Markus' sentence, but then maybe the current docs is just enough?

image

> If this event was created from a log record, the `level` indicates the level with which that message was logged if the logging framework uses the same levels as above.

This rewrite doesn't sound useful to me because it is not assertive.

untitaker: Well oftentimes we can't be assertive, but we can always tell the user where the data came from

marandaneto: on Android, we have an `enum` type for levels and not a raw String, so it's up to the Logging integration to pass semantically the right value.

PR #1582, opened 2020-04-02
ref(dsn): Use organization subdomains in DSN docs

jan-auer:
Sentry is switching to subdomains per organization for event ingestion, soon. This PR ensures that the subdomains are handled properly and `___PUBLIC_DSN___` is used consistently across all docs.


src/_js/lib/User.js
```diff
@@ -118,7 +118,12 @@ const formatUnrealEngineURL = function(dsn) {
 };
 
 const formatAPIURL = function(dsn) {
-  return `${dsn.scheme}${dsn.host}/api`;
```
untitaker, line 121: Oof, so we showed the ingest host for the web API docs?

PR #1586, opened 2020-04-06
fix: capitalization mismatch for `testKit` vs `testkit`

b-zurg:
Currently the typed result of `sentryTestKit()` is the following:
```typescript
  export interface TestkitResult {
    testkit: Testkit;
    sentryTransport: {
      new (options: TransportOptions): Transport;
    };
    initNetworkInterceptor<T>(dsn: string, initCallback: (baseUrl: string, handleRequestBody: (requestBody: { [key: string]: any }) => void) => T): T;
    localServer: LocalServer
  }
```

Notice that `testkit` is all lower case.  

In the example above this is imported and used with inconsistent naming.  This is simply a fix for that.


markstory:
Thank you 👍

PR #1587, opened 2020-04-06
fix: Clarify accepted timestamp formats

rhcarvalho:
Fix examples that had numeric value as milliseconds instead of seconds (server only accepts seconds).

Update examples to use RFC3339 as a preferred format for readability.

Related to https://github.com/getsentry/sentry-go/issues/190.

### DIFF
--- a/src/collections/_documentation/development/sdk-dev/event-payloads/index.md
+++ b/src/collections/_documentation/development/sdk-dev/event-payloads/index.md
@@ -54,9 +54,10 @@ The following attributes are required for all events.
 
 `timestamp`
 
-: Indicates when the logging record was created (in the Sentry SDK). The Sentry
-  server assumes the time is in UTC. The timestamp should be in ISO 8601 format,
-  without a timezone.
+: Indicates when the event was created in the Sentry SDK. The format is either a
+  string as defined in [RFC 3339](https://tools.ietf.org/html/rfc3339) or a
+  numeric (integer or float) value representing the [number of seconds that have
+  elapsed since the Unix epoch](https://en.wikipedia.org/wiki/Unix_time).
 
   ```json
   {
@@ -64,12 +65,9 @@ The following attributes are required for all events.
   }
   ```
 
-  Alternatively, the timestamp can be specified as floating-point UNIX
-  timestamp.
-
   ```json
   {
-    "timestamp": 1304358096000
+    "timestamp": 1304358096.0
   }
   ```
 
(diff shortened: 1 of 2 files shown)


src/collections/_documentation/development/sdk-dev/event-payloads/index.md
```diff
@@ -54,22 +54,21 @@ The following attributes are required for all events.
...
   }
   ```
 
-  Alternatively, the timestamp can be specified as floating-point UNIX
-  timestamp.
-
   ```json
   {
-    "timestamp": 1304358096000
+    "timestamp": 1304358096.0
```
rhcarvalho, line 71: I intentionally chose to make it a float as it emphasizes this possibility. Quick readers that will not read the fine print (int or float) and will still get it at a glance.

src/collections/_documentation/development/sdk-dev/event-payloads/index.md
```diff
@@ -54,22 +54,21 @@ The following attributes are required for all events.
...
 `timestamp`
 
-: Indicates when the logging record was created (in the Sentry SDK). The Sentry
-  server assumes the time is in UTC. The timestamp should be in ISO 8601 format,
-  without a timezone.
+: Indicates when the logging record was created (in the Sentry SDK). The format
+  is either a string as defined in [RFC
+  3339](https://tools.ietf.org/html/rfc3339) or a numeric (integer or float)
+  value representing the [number of seconds that have elapsed since the Unix
+  epoch](https://en.wikipedia.org/wiki/Unix_time).
```
rhcarvalho, line 61: The format description is the same as for `Breadcrumb.timestamp`.

I intentionally did not change other timestamps that are not part of ingestion, e.g. multiple parts of the Web API, because I did not research what is accepted in each case.

src/collections/_documentation/development/sdk-dev/event-payloads/index.md
```diff
@@ -54,22 +54,21 @@ The following attributes are required for all events.
 
 `timestamp`
 
-: Indicates when the logging record was created (in the Sentry SDK). The Sentry
-  server assumes the time is in UTC. The timestamp should be in ISO 8601 format,
-  without a timezone.
+: Indicates when the logging record was created (in the Sentry SDK). The format
```
jan-auer, line 57: Since we're updating this paragraph already, should we rename this to "event" and put some more emphasis on that it was created in the SDK?

```suggestion
: Indicates when the vevent was created in the Sentry SDK. The format
```

rhcarvalho: Applied this change locally to also rewrap the paragraph as the line length changed.

PR #1589, opened 2020-04-07
ref(sdk-dev): Improve Contexts Interface docs

rhcarvalho:
Getting it better step by step.

### DIFF
--- a/src/collections/_documentation/development/sdk-dev/event-payloads/contexts.md
+++ b/src/collections/_documentation/development/sdk-dev/event-payloads/contexts.md
@@ -3,9 +3,9 @@ title: Contexts Interface
 sidebar_order: 9
 ---
 
-The context interfaces provide additional context data. Typically, this is data
-related to the current user and the environment. For example, the device or application version. Its canonical name is
-`contexts`.
+The Contexts Interface provides additional context data. Typically, this is data
+related to the current user and the environment. For example, the device or
+application version. Its canonical name is `contexts`.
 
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/development/sdk-dev/event-payloads/contexts.md
```diff
@@ -298,6 +298,10 @@ Example:
 
 ## Examples
 
+The following example illustrates the contexts part of the [event payload]({%-
+link _documentation/development/sdk-dev/event-payloads/index.md -%}) and omits
+other attributes for simplicity.
```
rhcarvalho, line 303: Took the same wording as we have in the Breadcrumbs page, after replacing s/breadcrumbs/contexts.

src/collections/_documentation/development/sdk-dev/event-payloads/contexts.md
```diff
@@ -3,9 +3,9 @@ title: Contexts Interface
 sidebar_order: 9
 ---
 
-The context interfaces provide additional context data. Typically, this is data
```
rhcarvalho, line 6: There was a typo here, `context interfaces` => `Contexts Interface`.
The rest is just rewrapping the paragraph.

jan-auer: It was actually meant as plural since there are multiple contexts, but you're right. Singular is more consistent, since it is actually just one interface.

PR #1598, opened 2020-04-08
ref(APM): Fix typos, minor wordsmithing

lobsterkatie:
Started out just fixing one typo, ended up proofreading the whole thing. 🤦🏻‍♀️

Things which happened here:

- Fix a handful of tiny typos
- Fix a few places where `each` was treated as a plural
- Fix JS code snippets so comments fit in the default viewport (no scrolling necessary) and so that indents are 2 spaces (the JS standard) rather than 4 spaces (the standard for Python)
- Make places with identical content (between node and browser) actually have identical content. In the long run might want to use includes, but there may be some bigger changes to the section anyway (given new integrations being built this quarter by the Client Infra/Web team), so I figured I'd wait until after that happens.
- General smoothing of language

### DIFF
--- a/src/collections/_documentation/performance/distributed-tracing.md
+++ b/src/collections/_documentation/performance/distributed-tracing.md
@@ -16,43 +16,43 @@ Enabling tracing data collection augments your existing error data to capture in
 
 ## Tracing & Distributed Tracing
 
-Applications (for example, web applications) typically consist of interconnected applications/components (also called services). For instance, let us assume a modern web application could be composed of the following components separated by the network boundaries:
+Applications (for example, web applications) typically consist of interconnected components (also called services). For instance, let us assume a modern web application could be composed of the following components separated by the network boundaries:
 
 - Frontend (Single-Page Application)
 - Backend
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -102,15 +102,15 @@ Click on a row to expand the details of the span. From here, you can see all att
 
 **Search by Trace ID**
 
-You can search using `trace id` by expanding any of the span details and click on "Search by Trace".
+You can search using `trace id` by expanding any of the span details and click on "Search by Trace."
```
MimiDumpling, line 105:
```suggestion
You can search using `trace id` by expanding any of the span details and click on "Search by Trace".
```

MimiDumpling: It's the name of the button so we put the punctuation outside the quotes.

MimiDumpling: Thank you! 😸

PR #1599, opened 2020-04-09
fix: Add missing semicolon

bruno-garcia:


samrobben:
The block above is also missing a semicolon :-).

bruno-garcia:
@samrobben the one above is followed by `.` so chaining

PR #1612, opened 2020-04-16
chore: Cocoa Bump

HazAT:


untitaker:
Stupid q, why don't we use the release registry for this one? I think we did so for the Py sdk, even for betas

PR #1619, opened 2020-04-17
node: Unify imports syntax across all pages

kamilogorek:
Some people may get confused without providing both ways. See: https://github.com/getsentry/sentry-javascript/issues/1647

I verified that frameworks are imported correctly as well:

Screenshot 2020-04-17 at 11 31 21
Screenshot 2020-04-17 at 11 31 30

Also changed how integrations are imported to make it more obvious that they are completely independent from each other.

### DIFF
--- a/src/collections/_documentation/platforms/javascript/getting-started-dsn/node.md
+++ b/src/collections/_documentation/platforms/javascript/getting-started-dsn/node.md
@@ -1,6 +1,8 @@
 
 ```javascript
 import * as Sentry from '@sentry/node';
+// or using CommonJS
+// const Sentry = require('@sentry/node');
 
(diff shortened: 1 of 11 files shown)


src/collections/_documentation/platforms/javascript/getting-started-dsn/node.md
```diff
@@ -1,6 +1,8 @@
 
 ```javascript
 import * as Sentry from '@sentry/node';
+// or using CommonJS
+// const Sentry = require('@sentry/node');
 
 // All integrations that come with an SDK can be found on the Sentry.Integrations object
```
kamilogorek, line 7: I'm not sure if we need this comment part at all.

PR #1621, opened 2020-04-17
docs: Clarify disabled SDK, and SDK Event Processing

Swatinem:
This still uses `may` a few times.

I would ideally like to have hard guarantees around `configure_scope` and `add_breadcrumb`, right now it would be up to the discretion of the SDK.

### DIFF
--- a/src/collections/_documentation/development/sdk-dev/features.md
+++ b/src/collections/_documentation/development/sdk-dev/features.md
@@ -31,7 +31,7 @@ Automatic addition of useful attributes such as `tags` or `extra` or specific `c
 
 ## Breadcrumbs
 
-Manually record application events (into the current context) during the lifecycle of an application. Needs to implement a ring buffer so as not to grow indefinitely. The most recent breadcrumbs should be attached to events as they are created.
+Manually record application events (into the current scope) during the lifecycle of an application. Needs to implement a ring buffer so as not to grow indefinitely. The most recent breadcrumbs should be attached to events as they are created.
 
 With deeper framework integration, the automatic recording of some breadcrumbs is possible and recommended.
 
--- a/src/collections/_documentation/development/sdk-dev/features.md
+++ b/src/collections/_documentation/development/sdk-dev/features.md
@@ -31,7 +31,9 @@ Automatic addition of useful attributes such as `tags` or `extra` or specific `c
 
 ## Breadcrumbs
 
-Manually record application events (into the current scope) during the lifecycle of an application. Needs to implement a ring buffer so as not to grow indefinitely. The most recent breadcrumbs should be attached to events as they occur.
+Manually record application events (into the current scope) during the
+lifecycle of an application. Implement a ring buffer so as not to grow
+indefinitely. The most recent breadcrumbs should be attached to events as they occur.
 
 With deeper framework integration, the automatic recording of some breadcrumbs is possible and recommended.
 
--- a/src/collections/_documentation/development/sdk-dev/features.md
+++ b/src/collections/_documentation/development/sdk-dev/features.md
@@ -31,7 +31,7 @@ Automatic addition of useful attributes such as `tags` or `extra` or specific `c
 
 ## Breadcrumbs
 
-Manually record application events (into the current scope) during the lifecycle of an application. Needs to implement a ring buffer so as not to grow indefinitely. The most recent breadcrumbs should be attached to events as they are created.
+Manually record application events (into the current scope) during the lifecycle of an application. Needs to implement a ring buffer so as not to grow indefinitely. The most recent breadcrumbs should be attached to events as they occur.
 
 With deeper framework integration, the automatic recording of some breadcrumbs is possible and recommended.
 
--- a/src/collections/_documentation/development/sdk-dev/overview.md
+++ b/src/collections/_documentation/development/sdk-dev/overview.md
@@ -77,8 +77,7 @@ Sentry.init({
 ```
 
 {% capture __alert_content -%}
-If an empty DSN is passed, you should treat it as valid option which signifies not sending any events to the server.
-The SDK will however still process events and submit them to the configured transport.
+If an empty DSN is passed, you should treat it as valid option.
 {%- endcapture -%}
 {%- include components/alert.html
   title="Note"
--- a/src/collections/_documentation/development/sdk-dev/overview.md
+++ b/src/collections/_documentation/development/sdk-dev/overview.md
@@ -77,7 +77,7 @@ Sentry.init({
 ```
 
 {% capture __alert_content -%}
-It is valid to configure the SDK with an empty DSN.
+The SDK should accept an empty DSN as valid configuration.
 {%- endcapture -%}
 {%- include components/alert.html
   title="Note"
--- a/src/collections/_documentation/development/sdk-dev/overview.md
+++ b/src/collections/_documentation/development/sdk-dev/overview.md
@@ -77,7 +77,7 @@ Sentry.init({
 ```
 
 {% capture __alert_content -%}
-If you pass an empty DSN, treat it as a valid option.
+It is valid to configure the SDK with an empty DSN.
 {%- endcapture -%}
 {%- include components/alert.html
   title="Note"
--- a/src/collections/_documentation/development/sdk-dev/overview.md
+++ b/src/collections/_documentation/development/sdk-dev/overview.md
@@ -77,7 +77,7 @@ Sentry.init({
 ```
 
 {% capture __alert_content -%}
-If an empty DSN is passed, you should treat it as valid option.
+If you pass an empty DSN, treat it as a valid option.
 {%- endcapture -%}
 {%- include components/alert.html
   title="Note"
--- a/src/collections/_documentation/development/sdk-dev/overview.md
+++ b/src/collections/_documentation/development/sdk-dev/overview.md
@@ -77,7 +77,8 @@ Sentry.init({
 ```
 
 {% capture __alert_content -%}
-If an empty DSN is passed, you should treat it as valid option which signifies disabling the SDK.
+If an empty DSN is passed, you should treat it as valid option which signifies not sending any events to the server.
+The SDK will however still process events and submit them to the configured transport.
 {%- endcapture -%}
 {%- include components/alert.html
   title="Note"
--- a/src/collections/_documentation/development/sdk-dev/unified-api.md
+++ b/src/collections/_documentation/development/sdk-dev/unified-api.md
@@ -54,16 +54,17 @@ meant that certain integrations (such as breadcrumbs) were often not possible.
 
 - **integration**: Code that provides middlewares, bindings or hooks into certain frameworks or environments, along with code that inserts those bindings and activates them. Usage for integrations does not follow a common interface.
 
-- **event processors**: Callbacks that run for every event.
+- **event processors**: Callbacks that run for every event.  
   They can either return a "new" event, which in most cases means just adding
   data OR return `null` in which case the event will be discarded and not
   processed further.
 
   See [Event Pipeline](#event-pipeline) for more information.
 
-- **disabled SDK / active Client**: Most of the SDK functionality depends on a
-  configured and active Client. Sentry considers the client active when it has a
-  *transport*. Otherwise, the SDK is considered "disabled".
+- **disabled SDK**: Most of the SDK functionality depends on a
+  configured and active Client.
+  Sentry considers the client active when it has a *transport*.
+  Otherwise, the client is inactive, and the SDK is considered "disabled".
   This means that certain callbacks, such as `configure_scope` or
   *event processors*, may not be invoked, and no breadcrumbs are being recorded.
 
@@ -99,9 +100,9 @@ Additionally it also setups all default integrations.
 - `capture_message(message, level)`: Reports a message. The level can be optional in language with default parameters in which case it should default to `info`.
 
 - `add_breadcrumb(crumb)`: Adds a new breadcrumb to the scope. If the total
-  number of breadcrumbs exceeds the `max_breadcrumbs` setting, Sentry removes
-  the oldest breadcrumb. This works like the Hub API with regards to what
-  `crumb` can be. If the SDK is disabled, Sentry may ignore the breadcrumb.
+  number of breadcrumbs exceeds the `max_breadcrumbs` setting, the SDK should
+  remove the oldest breadcrumb. This works like the Hub API with regards to what
+  `crumb` can be. If the SDK is disabled, it should ignore the breadcrumb.
 
 - `configure_scope(callback)`: Calls a callback with a scope object that can be reconfigured. This is used to attach contextual data for future events in the same scope.
 
@@ -145,7 +146,7 @@ The SDK maintains two variables: The *main hub* (a global variable) and the *cur
 
 - `Hub::pop_scope()` (optional): Only exists in languages without better resource management. Better to have this function on a return value of `push_scope` or to use `with_scope`.  This is also sometimes called `pop_scope_unsafe` to indicate that this method should not be used directly.
 
-- `Hub::configure_scope(callback)`: Invokes the callback with a mutable reference to the scope for modifications. This can also be a `with` statement in languages that have it (Python). If no active client is bound to this hub, Sentry may not invoke the callback.
+- `Hub::configure_scope(callback)`: Invokes the callback with a mutable reference to the scope for modifications. This can also be a `with` statement in languages that have it (Python). If no active client is bound to this hub, the SDK should not invoke the callback.
 
 - `Hub::add_breadcrumb(crumb, hint)`: Adds a breadcrumb to the current scope.
 
@@ -155,7 +156,7 @@ The SDK maintains two variables: The *main hub* (a global variable) and the *cur
     - a list of breadcrumbs (optional)
   - In languages where we do not have a basic form of overloading only a raw breadcrumb object should be accepted.
 
-  Sentry may ignore the breadcrumb if no active client is bound to this hub.
+  The SDK should ignore the breadcrumb if no active client is bound to this hub.
 
   For the hint parameter see [hints](#hints).
 
@@ -174,9 +175,9 @@ The SDK maintains two variables: The *main hub* (a global variable) and the *cur
 A scope holds data that should implicitly be sent with Sentry events. It can hold context data, extra parameters, level overrides, fingerprints etc.
 
 The user can modify the current scope (to set extra, tags, current user) through the global function `configure_scope`.  `configure_scope` takes a callback function to which it passes the current scope.
-If the SDK is disabled, Sentry may not invoke the callback.
 
-Here's an example from another place in the docs:
+The reason for this callback-based API is efficiency. If the SDK is disabled,
+it should not invoke the callback, thus avoiding unnecessary work.
 
 ```javascript
 Sentry.configureScope(scope =>
--- a/src/collections/_documentation/development/sdk-dev/unified-api.md
+++ b/src/collections/_documentation/development/sdk-dev/unified-api.md
@@ -54,17 +54,18 @@ meant that certain integrations (such as breadcrumbs) were often not possible.
 
 - **integration**: Code that provides middlewares, bindings or hooks into certain frameworks or environments, along with code that inserts those bindings and activates them. Usage for integrations does not follow a common interface.
 
-- **event processors**: Are callbacks that run for every event.
-  They can either return a "new" event which in most cases means just adding
-  data OR return `null` in case the event will be dropped and not sent.
+- **event processors**: Callbacks that run for every event.
+  They can either return a "new" event, which in most cases means just adding
+  data OR return `null` in which case the event will be discarded and not
+  processed further.
 
   See [Event Pipeline](#event-pipeline) for more information.
 
 - **disabled SDK / active Client**: Most of the SDK functionality depends on a
-  configured and active Client. A client is considered active when it has a
+  configured and active Client. Sentry considers the client active when it has a
   *transport*. Otherwise, the SDK is considered "disabled".
   This means that certain callbacks, such as `configure_scope` or
-  *event processors* may not be invoked and no breadcrumbs are being recorded.
+  *event processors*, may not be invoked, and no breadcrumbs are being recorded.
 
 
 ## "Static API"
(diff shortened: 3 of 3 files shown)


src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -54,7 +54,15 @@ meant that certain integrations (such as breadcrumbs) were often not possible.
 
 - **integration**: Code that provides middlewares, bindings or hooks into certain frameworks or environments, along with code that inserts those bindings and activates them. Usage for integrations does not follow a common interface.
 
-- **event processors:** Are callbacks that run for every event. They can either return a "new" event which in most cases means just adding data OR return `null` in case the event will be dropped and not sent.
+- **event processors**: Are callbacks that run for every event. They can either return a "new" event which in most cases means just adding data OR return `null` in case the event will be dropped and not sent.
+
+- **disabled SDK**: Before `init` has been called, or when no client is active,
```
untitaker, line 59: There are two disabled states: Having no client, or having a client without a transport. One can reach both states by either not calling init, or by calling init with an empty string as DSN.

If you call `init("")` you get breadcrumbs recorded.

Swatinem: > One can reach both states by either not calling init, or by calling init with an empty string as DSN.

as you can see below, the event pipeline should run in this case. IMO its a valid usecase to have no DSN when you are using a custom transport.

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -86,7 +94,7 @@ Additionally it also setups all default integrations.
 
 - `capture_message(message, level)`: Reports a message. The level can be optional in language with default parameters in which case it should default to `info`.
 
-- `add_breadcrumb(crumb)`: Adds a new breadcrumb to the scope. If the total number of breadcrumbs exceeds the `max_breadcrumbs` setting, the oldest breadcrumb should be removed in turn. This works like the Hub api with regards to what `crumb` can be.
+- `add_breadcrumb(crumb)`: Adds a new breadcrumb to the scope. If the total number of breadcrumbs exceeds the `max_breadcrumbs` setting, the oldest breadcrumb should be removed in turn. This works like the Hub api with regards to what `crumb` can be. The breadcrumb is ignored if the SDK is disabled.
```
HazAT, line 97: > The breadcrumb is ignored if the SDK is disabled.
Do we really want to define it that strictly?
Why wouldn't we want that even if disabled it will keep the breadcrumbs?

Swatinem: While I don’t like making public API based on implementation details in general, here, if we don’t have a client, we don’t have a `max_breadcrumbs` setting. So what should we do in that case? Maybe define a global limit somewhere?

Also, both this and `configure_scope` live on the scope. So we should skip both, to at least be consistent.

src/collections/_documentation/development/sdk-dev/unified-api.md
rhcarvalho, line 59: `when no client is active,` is ambiguous.

What we have is a "current hub" and each hub has a stack of clients (and scopes).

Here `active` probably refers to `no client on the top of the hub stack`, though maybe we never talk about hub stacks.

Swatinem: How would you phrase that ideally? since all the APIs essentially end up at a Hub, which may not have an active Client.

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -217,6 +227,22 @@ the event.  For instance if an exception was captured the hint might carry the
 original exception object.  Not all SDKs are required to provide this.  The
 parameter however is reserved for this purpose.
 
+## Event Pipeline
+
+Any event that is captured by `capture_event` is processed in the following
+order and can be discarded at any of the stages, at which point no further
+processing happens.
+
+1. If the SDK is disabled or there is no active client, the event is discarded
```
rhcarvalho, line 236: Given the definition of `disabled SDK` above, isn't `SDK is disabled or there is no active client` redundant?

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -217,6 +227,22 @@ the event.  For instance if an exception was captured the hint might carry the
...
+   right away. The client is active if it has a valid *transport* configured.
+2. The client does event sampling based on the configured sample rate.
+   Events will be discarded randomly, according to the sample rate.
+3. The scope is applied, using `apply_to_event`. The scopes *event processors*
```
rhcarvalho, line 240:
```suggestion
3. The scope is applied, using `apply_to_event`. The scope's *event processors*
```

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -217,6 +227,22 @@ the event.  For instance if an exception was captured the hint might carry the
...
+   are invoked in order and can possibly discard the event.
+4. The *before-send* hook is invoked, which can also discard the event.
+5. The event is passed to the configured transport. The transport can discard
+   the event in case it has no valid DSN, or due to rate limiting.
```
rhcarvalho, line 244: Well, rate limiting is a server concern.

The more general thing to say is that "The default transport does not retry to send events that are not accepted by the server" (and that can be 4xx, 5xx responses from the server).

WDYT

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -164,7 +174,7 @@ Sentry.configureScope(function(scope) {
 });
 ```
 
-Why not just have a `get_current_scope()` function instead of this indirection?  If the SDK is disabled (e.g. by not providing a DSN), modifying a scope is pointless because there will never be any events to send. In that situation `configure_scope` may choose not to call the callback.  This is also used to more efficiently flush out scope changes in some languages.  When a `with` statement is used that always executes the scope needs to be a dummy scope object.  It also helps to emphasize that under normal circumstances we do not want the user the do the scope handling.
+Why not just have a `get_current_scope()` function instead of this indirection?  If the SDK is disabled, modifying a scope is pointless because there will never be any events to send. In that situation `configure_scope` may choose not to call the callback.  This is also used to more efficiently flush out scope changes in some languages.  When a `with` statement is used that always executes the scope needs to be a dummy scope object.  It also helps to emphasize that under normal circumstances we do not want the user the do the scope handling.
```
rhcarvalho, line 177: This paragraph, even before change, sounds like a long rant.

Going on a sidetrack re: writing style, would be good to have @MimiDumpling's/ @PeloWriter's take on whether we want to ask questions like this, or write it differently simply to provide the answer.

src/collections/_documentation/development/sdk-dev/unified-api.md
rhcarvalho:
```suggestion
- **disabled SDK**: Before `init` has been called or when the current hub has no client
```

?

Perhaps not ideal yet...

src/collections/_documentation/development/sdk-dev/unified-api.md
marandaneto: `max_breadcrumbs` lives on SentryOptions though.

src/collections/_documentation/development/sdk-dev/unified-api.md
Swatinem: hm… at least for native, the transport eagerly discards anything in the `retry-after` window.

src/collections/_documentation/development/sdk-dev/unified-api.md
marandaneto: agreed

src/collections/_documentation/development/sdk-dev/unified-api.md
marandaneto: the server can also discard events down the pipeline even if it returned 200, 200 just means it enqueued right

src/collections/_documentation/development/sdk-dev/unified-api.md
bruno-garcia: In Android (/cc @marandaneto )and .NET it's all NoOp until init is called. If init is called with an empty DSN, nothing happens but it sounds it's the only SDK doing that.

src/collections/_documentation/development/sdk-dev/overview.md
```diff
@@ -77,7 +77,8 @@ Sentry.init({
 ```
 
 {% capture __alert_content -%}
-If an empty DSN is passed, you should treat it as valid option which signifies disabling the SDK.
+If an empty DSN is passed, you should treat it as valid option which signifies not sending any events to the server.
+The SDK will however still process events and submit them to the configured transport.
```
bruno-garcia, line 81: What is the advantage of passing event to the transport if the SDK is disabled?
Does this need to be a rule? I've taken the line above quite literally and I have looked up for a DSN on init:

https://github.com/getsentry/sentry-dotnet/blob/master/src/Sentry/SentrySdk.cs#L79-L92

If no DSN exists, I keep a NoOp hub bound to the static accessor.

src/collections/_documentation/development/sdk-dev/unified-api.md
bruno-garcia, line 97: @HazAT better like this? It's just a suggestion (to make it cheaper):

```suggestion
- `add_breadcrumb(crumb)`: Adds a new breadcrumb to the scope. If the total number of breadcrumbs exceeds the `max_breadcrumbs` setting, the oldest breadcrumb should be removed in turn. This works like the Hub api with regards to what `crumb` can be. The breadcrumb may ignored if the SDK is disabled.
```

src/collections/_documentation/development/sdk-dev/unified-api.md
bruno-garcia: IHMO we could remove this paragraph altogether.
Content like this could be a separate document for motivations of the current API or, FAQ:

Q: Why not expose the current scope?
A: The lifetime is controlled via Push/Pop and holding the reference by the user could lead to a use after free.

src/collections/_documentation/development/sdk-dev/unified-api.md
bruno-garcia, line 244:
```suggestion
   the event in case it has no valid DSN, max queued items have been reached, or due to rate limiting.
```

src/collections/_documentation/development/sdk-dev/unified-api.md
marandaneto: actually, if you create a `Hub` by yourself with a `null` or `empty` DSN you even get `IllegalArgumentException` on Android.
If you create a `SentryClient` with a `null` or `empty` DSN and no transport is set, its also throwing `IllegalArgumentException`
Only thru the Sentry static class with an empty/null DSN that becomes NoOp for everything on Android.
Yes, it seems to be different than others.

bruno-garcia: `null` IMHO should throw. The spec says empty string.

Throwing on initialization isn't that big deal since it should repro during the app start and during tests, shows the misconfiguration.

Something that is completely against the guideline for any public method **after** the SDK is initialized, which should never throw.

`Hub` and `SentryClient` are much simpler if can only be properly instantiated if a valid DSN is provided. So the guard for `""` happens outside (`Sentry.init` does that). The documentation never made clear that "" as a DSN in any code path should mean empty string and for that  reason that guideline was taken only at the public documented API `Sentry.init`. That said, it was by design.

src/collections/_documentation/development/sdk-dev/overview.md
Swatinem: I think its a valid usecase if you have a custom transport configured which does not depend on a DSN.

bruno-garcia: Could be a valid use-case even though it hasn't come up yet but honestly this sounds like an edge case and maybe better just left off of the spec.

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -217,6 +231,22 @@ the event.  For instance if an exception was captured the hint might carry the
...
+processing happens.
+
+1. If the SDK is disabled, the event is discarded right away.
+2. The client does event sampling based on the configured sample rate.
+   Events will be discarded randomly, according to the sample rate.
+3. The scope is applied, using `apply_to_event`. The scope’s *event processors*
+   are invoked in order and can possibly discard the event.
+4. The *before-send* hook is invoked, which can also discard the event.
+5. The event is passed to the configured transport. The transport can discard
+   the event in case it has no valid DSN, its internal queue is full,
```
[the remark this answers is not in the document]
Swatinem: do you really separate enumeration items by a full stop?

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -156,16 +173,16 @@ The SDK maintains two variables: The *main hub* (a global variable) and the *cur
 
 A scope holds data that should implicitly be sent with Sentry events. It can hold context data, extra parameters, level overrides, fingerprints etc.
 
-The user should be able to modify the current scope easily (to set extra, tags, current user), through a global function `configure_scope`.  `configure_scope` takes a callback function to which it passes the current scope. Here's an example from another place in the docs:
+The user can modify the current scope (to set extra, tags, current user) through the global function `configure_scope`.  `configure_scope` takes a callback function to which it passes the current scope.
+If the SDK is disabled, Sentry may not invoke the callback.
+
+Here's an example from another place in the docs:
```
rhcarvalho, line 179: This is strange. Why say "this is from another place"?

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -156,16 +173,16 @@ The SDK maintains two variables: The *main hub* (a global variable) and the *cur
...
 
 ```javascript
-Sentry.configureScope(function(scope) {
-  scope.setExtra("character_name", "Mighty Fighter");
-});
+Sentry.configureScope(scope =>
+  scope.setExtra("character_name", "Mighty Fighter"));
 ```
 
-Why not just have a `get_current_scope()` function instead of this indirection?  If the SDK is disabled (e.g. by not providing a DSN), modifying a scope is pointless because there will never be any events to send. In that situation `configure_scope` may choose not to call the callback.  This is also used to more efficiently flush out scope changes in some languages.  When a `with` statement is used that always executes the scope needs to be a dummy scope object.  It also helps to emphasize that under normal circumstances we do not want the user the do the scope handling.
```
rhcarvalho, line 167: Why remove this completely? In general it is good to have insights on why certain APIs even exist.

src/collections/_documentation/development/sdk-dev/overview.md
```diff
@@ -77,7 +77,7 @@ Sentry.init({
 ```
 
 {% capture __alert_content -%}
-If an empty DSN is passed, you should treat it as valid option which signifies disabling the SDK.
+If you pass an empty DSN, treat it as a valid option.
```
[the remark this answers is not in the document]
Swatinem: How can we better highlight the difference between *I want the default (=env.SENTRY_DSN)* vs *I want *no* dsn*? In JS you have `undefined` vs `null` exactly for this usecase. I think other languages offer something similar.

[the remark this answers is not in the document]
bruno-garcia: One approach here: https://github.com/getsentry/sentry-dotnet/blob/master/src/Sentry/SentrySdk.cs#L81-L91

src/collections/_documentation/development/sdk-dev/unified-api.md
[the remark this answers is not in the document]
MimiDumpling: Ah! Sorry. That's my mistake. I thought this was a run on sentence. The original version is fine. Thanks!

src/collections/_documentation/development/sdk-dev/features.md
```diff
@@ -31,7 +31,7 @@ Automatic addition of useful attributes such as `tags` or `extra` or specific `c
 
 ## Breadcrumbs
 
-Manually record application events (into the current context) during the lifecycle of an application. Needs to implement a ring buffer so as not to grow indefinitely. The most recent breadcrumbs should be attached to events as they are created.
+Manually record application events (into the current scope) during the lifecycle of an application. Needs to implement a ring buffer so as not to grow indefinitely. The most recent breadcrumbs should be attached to events as they occur.
```
PeloWriter, line 34: Clarify who or what "Needs to implement a ring buffer" - as it reads now it is confusing.

Suggest: "Implement a ring buffer" - developers understand why to use it, I think.

src/collections/_documentation/development/sdk-dev/overview.md
```diff
@@ -77,7 +77,7 @@ Sentry.init({
 ```
 
 {% capture __alert_content -%}
-If an empty DSN is passed, you should treat it as valid option which signifies disabling the SDK.
+It is valid to configure the SDK with an empty DSN.
```
PeloWriter, line 80: Suggest "Configure the SDK so that an empty DSN is a valid option"

Q - does this disable the SDK? If so, that is not captured in the new content (nor my suggested edit)

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -54,7 +54,20 @@ meant that certain integrations (such as breadcrumbs) were often not possible.
 
 - **integration**: Code that provides middlewares, bindings or hooks into certain frameworks or environments, along with code that inserts those bindings and activates them. Usage for integrations does not follow a common interface.
 
-- **event processors:** Are callbacks that run for every event. They can either return a "new" event which in most cases means just adding data OR return `null` in case the event will be dropped and not sent.
+- **event processors**: Callbacks that run for every event.  
+  They can either return a "new" event, which in most cases means just adding
```
PeloWriter, line 58: suggest using italics for _new_ here rather than quotes.

Can we make this more clear? The phrase "which in most cases means just adding data" suggests that other cases exist. 

Suggest "These either return a _new_ event or `null`. New events typically add data. `Null` is discarded and does not process further."

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -54,7 +54,20 @@ meant that certain integrations (such as breadcrumbs) were often not possible.
...
+  data OR return `null` in which case the event will be discarded and not
+  processed further.
+
+  See [Event Pipeline](#event-pipeline) for more information.
+
+- **disabled SDK**: Most of the SDK functionality depends on a
+  configured and active Client.
```
PeloWriter, line 65: lower case for client

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -54,7 +54,20 @@ meant that certain integrations (such as breadcrumbs) were often not possible.
...
+  Sentry considers the client active when it has a *transport*.
+  Otherwise, the client is inactive, and the SDK is considered "disabled".
+  This means that certain callbacks, such as `configure_scope` or
```
PeloWriter, line 68: Suggest "When a client is inactive and the SDK is considered disabled, certain callbacks, such as `configure_scope` or *event processors*, may not be invoked. As a result, breadcrumbs are not recorded.

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -188,7 +206,8 @@ Why not just have a `get_current_scope()` function instead of this indirection?
 
 - `scope.add_error_processor(processor)` (optional): Registers an error processor function.  It takes an event and exception object and returns a new event or `None` to drop it.  This can be used to extract additional information out of an exception object that the SDK cannot extract itself.
 
-- `scope.clear()`: Resets a scope to default values (prevents inheriting).  This never clears registered event processors.
+- `scope.clear()`: Resets a scope to default values however keeping all
```
PeloWriter, line 209: This is confusing, perhaps: "Resets a scope to default values while keeping registered event processors, which does not affect child or parent scopes."

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -217,6 +236,22 @@ the event.  For instance if an exception was captured the hint might carry the
 original exception object.  Not all SDKs are required to provide this.  The
 parameter however is reserved for this purpose.
 
+## Event Pipeline
+
+Any event that is captured by `capture_event` is processed in the following
```
PeloWriter, line 241: Strictly editorial, to tighten up the language: "An event captured by `capture_event` processes in the following order: **Note:** The event can be discarded at any of these stages, at which point it is no longer processed."

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -217,6 +236,22 @@ the event.  For instance if an exception was captured the hint might carry the
...
+order and can be discarded at any of the stages, at which point no further
+processing happens.
+
+1. If the SDK is disabled, Sentry discards the event right away.
+2. The client does event sampling based on the configured sample rate.
```
PeloWriter, line 246: The client samples as defined by the configured sample rate.

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -217,6 +236,22 @@ the event.  For instance if an exception was captured the hint might carry the
...
+   Events will be discarded randomly, according to the sample rate.
+3. The scope is applied, using `apply_to_event`. The scope’s *event processors*
```
PeloWriter, line 248: Given the note about discarding events at any stage, suggest removing the phrase "...and can possibly discard the event."

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -217,6 +236,22 @@ the event.  For instance if an exception was captured the hint might carry the
...
+   are invoked in order and can possibly discard the event.
+4. Sentry invokes the *before-send* hook, which can also discard the event.
```
PeloWriter, line 250: Suggest removing "...which can also discard the event" given the note at the start of this list.

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -217,6 +236,22 @@ the event.  For instance if an exception was captured the hint might carry the
...
+5. Sentry passes the event to the configured transport. The transport can discard
```
PeloWriter, line 251: Mild rephrase suggestion for clarity: "The transport can discard the event if it does not have a valid DSN; its internal queue is full; or due to rate limiting, as requested by the server."

src/collections/_documentation/development/sdk-dev/unified-api.md
Swatinem: I actually changed this completely, as the *new* was a bit misleading.
You return the event that you get after modifications (move semantics).
While it is technically possible to create a *completely new* event, I don’t think that is a valid usecase actually.

src/collections/_documentation/development/sdk-dev/unified-api.md
Swatinem: TBH, I think active voice is not the right thing here. The event does not process itself ;-)

src/collections/_documentation/development/sdk-dev/unified-api.md
```diff
@@ -217,6 +236,22 @@ the event.  For instance if an exception was captured the hint might carry the
...
+processing happens.
+
+1. If the SDK is disabled, Sentry discards the event right away.
+2. The client samples events as defined by the configured sample rate.
+   Events may be discarded randomly, according to the sample rate.
+3. The scope is applied, using `apply_to_event`. The scope’s *event processors*
+   are invoked in order.
+4. Sentry invokes the *before-send* hook.
+5. Sentry passes the event to the configured transport.
+   The transport can discard the event if it does no have valid DSN; its
```
rhcarvalho, line 252:
```suggestion
   The transport can discard the event if it does not have a valid DSN; its
```

PR #1624, opened 2020-04-17
Fix: Remove unnecessary redirect

MimiDumpling:
Removing a 302 from nginx.conf because it's unnecessary and incorrect.


mattrobenolt:
🌻

PR #1625, opened 2020-04-20
fix[sentry-cocoa]: Use new Start Syntax

HazAT:
- beta.4
- Session Health Changes


src/collections/_documentation/platforms/cocoa/enableAutoSessionTracking/objc.md
```diff
@@ -0,0 +1,6 @@
+```objc
```
marandaneto, line 1: is this the right one?  I think the other docs have `objective-c`

HazAT: `objective-c` is wrong I tested it and everywhere else it's `objc`

PR #1630, opened 2020-04-21
Docs: Fix Error in JavaScript SDK Guide (EmberJS)

jayjayjpg:
During a pairing session with a colleague who was integrating Sentry into an Ember app I realised that the

```js
import * as Sentry from '@sentry/browser'
```
statement mentioned in the Ember tutorial will error when following the guide, because the installation step for `@sentry/browser` is missing.

This PR adds installation instructions for installing `@sentry/browser` as an initial step, as well as more context for what ember-auto-import is used for during the integration process.


jayjayjpg:
Can anyone take a look at this one or comment if something is still missing? Thanks!

PeloWriter:
Thanks for the PR. We've recently changed the Ember SDK to an Ember addon - https://docs.sentry.io/platforms/javascript/guides/ember/

PR #1631, opened 2020-04-22
updated javascript.md snippet with comment

krunalg:
process.exit() should only be used in NodeJS application, it will throw an error in the browser.


krunalg:
not required

PR #1638, opened 2020-04-24
feat(protocol): Document image_vmaddr

jan-auer:
This was originally deprecated and thus undocumented. However, to render proper apple crash reports and addresses that can be fed into `addr2line`, it is useful to retain the original load address as `image_vmaddr`. 

Note that the description for the three platforms varies slightly.

### DIFF
--- a/src/collections/_documentation/development/sdk-dev/event-payloads/debugmeta.md
+++ b/src/collections/_documentation/development/sdk-dev/event-payloads/debugmeta.md
@@ -92,7 +92,7 @@ Attributes:
   still choose to place it at a different address. 
   
   If this value is non-zero, all symbols and addresses declared in the native
-  image start at this address, rather than `0`. In contrast, Sentry deals with
+  image start at this address, rather than `0`. By contrast, Sentry deals with
   addresses relative to the start of the image. For example, with
   `image_vmaddr: 0x40000`, a symbol located at `0x401000` has a relative address
   of `0x1000`.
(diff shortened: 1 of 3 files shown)


src/collections/_documentation/development/sdk-dev/event-payloads/debugmeta.md
```diff
@@ -85,6 +85,21 @@ Attributes:
...
   to locate the file if it is missing on Sentry.
 
+`image_vmaddr`:
+
+: _Optional_: Preferred load address of the image in virtual memory, as declared
+  in the headers of the image. When loading an image, the operating system may
+  still choose to place it at a different address. 
+  
+  If this value is non-zero, all symbols and addresses declared in the native
+  image start at this address, rather than `0`. In contrast, Sentry deals with
```
PeloWriter, line 95: By contrast 

(In contrast is typically followed by "with" or "to"; by contrast is typically followed by the subject of the contrast :) )

src/collections/_documentation/development/sdk-dev/event-payloads/debugmeta.md
```diff
@@ -165,6 +181,21 @@ Attributes:
...
```
PeloWriter, line 191: see previous edit

src/collections/_documentation/development/sdk-dev/event-payloads/debugmeta.md
jan-auer: Thanks, also updated other parts in docs where this was used.

cc @MimiDumpling

PR #1642, opened 2020-04-27
Consistency with other pages & better line length.

nitinnain:
Formatting changes that make the code sample consistent with other code samples, say https://docs.sentry.io/platforms/python/redis/


untitaker:
Thanks!

PR #1647, opened 2020-04-28
Update JavaScript capture errors sections

MimiDumpling:
Switched the automatic and manual sections on the JS page.

**BEFORE**:
image

**AFTER**:
image

### DIFF
--- a/src/collections/_documentation/platforms/javascript/index.md
+++ b/src/collections/_documentation/platforms/javascript/index.md
@@ -69,40 +69,6 @@ By including and configuring Sentry, the SDK will automatically attach global ha
 [{% asset js-index/automatically-capture-errors.png alt="Stack trace of a captured error" %}]({% asset js-index/automatically-capture-errors.png @path %})
 
 Browsers take security measures when serving script files from different origins. To ensure errors are routed to Sentry, configure CORS headers, and add appropriate script attributes. 
-
-**CORS Attributes and Headers**
-
-To gain visibility into a JavaScript exception thrown from scripts originating from different origins, do two things:
-
-1. Add a `crossorigin=”anonymous”` script attribute
-
-    ```javascript
-    <script src="http://another-domain.com/app.js" crossorigin="anonymous"></script>
-    ```
-
-    The script attribute tells the browser to fetch the target file "anonymously." Potentially user-identifying information like cookies or HTTP credentials won't be transmitted by the browser to the server when requesting this file.
-
-2. Add a Cross-Origin HTTP header
-
-    ```javascript
-    Access-Control-Allow-Origin: *
-    ```
-    Cross-Origin Resource Sharing (CORS) is a set of APIs (mostly HTTP headers) that dictate how files ought to be downloaded and served across origins.
-    
-    By setting `Access-Control-Allow-Origin: *`, the server is indicating to browsers that any origin can fetch this file. Alternatively, you can restrict it to a known origin you control:
-    
-    ```javascript
-    Access-Control-Allow-Origin: https://www.example.com
-    ```
-    
-    Most community CDNs properly set an Access-Control-Allow-Origin header.
-
-    ```bash
-    $ curl --head https://ajax.googleapis.com/ajax/libs/jquery/2.2.0/jquery.js | \
-    grep -i "access-control-allow-origin"
-
-    Access-Control-Allow-Origin: *
-    ```
     
 ### Manually Capture Errors
 In most situations, you can capture errors manually with `captureException()`.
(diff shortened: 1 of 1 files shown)


benvinegar:
This feels like an improvement. I think that "Keep in mind" block is ... a little overkill for this stage (and also the grammar seems rough). I'd guess less than 10% of people are going to have any trouble here (and probably not when they're fooling around w/ Sentry for the first time).

Maybe it should go in a "info" block or something. Also, the grammar could use some improving (the present tense feels odd, "browsers are taking some security measures" vs "browsers take security measures").

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -61,7 +61,19 @@ Then, you can see the error in your dashboard:
...
 ## Capturing Errors
-In most situations, you can capture errors automatically with `captureException()`.
+Automatically and manually capture errors, exceptions, and rejections.
+
+### Automatically Capture Errors
+By including and configuring Sentry, the SDK will automatically attach global handlers to capture uncaught exceptions and unhandled rejections.
+
+[{% asset js-index/automatically-capture-errors.png alt="Stack trace of a captured error" %}]({% asset js-index/automatically-capture-errors.png @path %})
+
+Keep in mind; browsers are taking some security measures when serving script files from different origins. For errors to always make their way to Sentry, make sure that you configure CORS headers and add appropriate script attributes.
```
PeloWriter, line 71: Agree with Ben that this reads oddly. Suggest "Browsers take security measure when serving script files from different original. To ensure errors are routed to Sentry, configure the CORS headers and add script attributes."

We should also add the content from the blog post, then remove this link since the blog is pretty dated. We can manage that change in a separate PR, if that's better.

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -61,7 +61,51 @@ Then, you can see the error in your dashboard:
...
+    
+    Most community CDNs properly set an Access-Control-Allow-Origin header.
+
+    ```bash
+    $ curl --head https://ajax.googleapis.com/ajax/libs/jquery/2.2.0/jquery.js | \
+    grep -i "access-control-allow-origin"
+
+    Access-Control-Allow-Origin: *
+    ```
+
```
MimiDumpling, line 106: @benvinegar the two steps above were taken from a [blog post](https://blog.sentry.io/2016/05/17/what-is-script-error#the-fix-cors-attributes-and-headers). I've omitted a couple of the ending sentences because they reference an example earlier in the blog post that I didn't add here. Does this section still make sense without those sentences? Here's what I omitted:

> Once you've completed these steps, any errors triggered by this script will report to window.onerror just like any regular same-domain script. So instead of "Script error," the onerror example from the beginning will yield:
> `"ReferenceError: bar is not defined", "http://another-domain.com/app.js", 2, 1, [Object Error]`

MimiDumpling: @PeloWriter will you review lines 73 through 105? Thanks 😄

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -68,10 +68,42 @@ By including and configuring Sentry, the SDK will automatically attach global ha
 
 [{% asset js-index/automatically-capture-errors.png alt="Stack trace of a captured error" %}]({% asset js-index/automatically-capture-errors.png @path %})
 
-Keep in mind; browsers are taking some security measures when serving script files from different origins. For errors to always make their way to Sentry, make sure that you configure CORS headers and add appropriate script attributes. 
+Browsers take security measures when serving script files from different origins. To ensure errors are routed to Sentry, configure the CORS headers, and add script attributes.
```
benvinegar, line 71: >  To ensure errors are routed to Sentry, configure the CORS headers, and add script attributes.

"The CORS headers" isn't correct; drop "the". "Appropriate script attributes" is more accurate than "add script attributes". The second half of this sentence was more accurate before edits.

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -61,7 +61,51 @@ Then, you can see the error in your dashboard:
...
+### Automatically Capture Errors
+By including and configuring Sentry, the SDK will automatically attach global handlers to capture uncaught exceptions and unhandled rejections.
+
+[{% asset js-index/automatically-capture-errors.png alt="Stack trace of a captured error" %}]({% asset js-index/automatically-capture-errors.png @path %})
+
+Browsers take security measures when serving script files from different origins. To ensure errors are routed to Sentry, configure the CORS headers, and add script attributes. 
+
+**CORS Attributes and Headers**
+
+To gain visibility into a JavaScript exception thrown from scripts originating from different origins, do two things:
```
benvinegar, line 75: So, I think this is too much information for the docs at this stage. Like, 95-99% of users won't have to deal with this at all. It should be reserved for a troubleshooting section.

src/collections/_documentation/platforms/javascript/index.md
MimiDumpling: @benvinegar, I updated the sentence. 😄 Let me know if there are any other changes you'd like.

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -987,7 +990,43 @@ basic stack trace. This exception is stored here for further data extraction.
 
 : For breadcrumbs created from HTTP requests done via the legacy `XMLHttpRequest` API. This holds the original xhr object.
 
+## Troubleshoot
```
benvinegar, line 993: This is typically called "Troubleshooting"; I'm not sure I've seen "Troubleshoot" as a heading before.

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -61,7 +61,17 @@ Then, you can see the error in your dashboard:
...
 ## Capturing Errors
-In most situations, you can capture errors automatically with `captureException()`.
+Automatically and manually capture errors, exceptions, and rejections.
+
+### Automatically Capture Errors
+By including and configuring Sentry, the SDK will automatically attach global handlers to capture uncaught exceptions and unhandled rejections.
+
+[{% asset js-index/automatically-capture-errors.png alt="Stack trace of a captured error" %}]({% asset js-index/automatically-capture-errors.png @path %})
+
+Browsers take security measures when serving script files from different origins. To ensure errors are routed to Sentry, configure CORS headers, and add appropriate script attributes.
```
benvinegar, line 71: I'd do something like this:

> Note: Browsers may take security measures when serving script files from different origins that can block error reporting. If you find errors are not being reported properly, please see this troubleshooting section on Cross Origin Resource Sharing (CORS).

... with that last bit linking to the troubleshooting guide you've added below.

PR #1650, opened 2020-04-29
fix(envelopes): Clarify what is an event

jan-auer:


rhcarvalho:
> We're missing at least one more

Unless we really mean that the storage is of error events, and never other things like sessions and transactions.

jan-auer:
> Maybe `s/events/items`? Unfortunate that we have to find synonyms since `"events"` means `"errors"`

@rhcarvalho the "events" here is very intentional, since event means both "error" and "default". The example is specifically for a common case of saving traditional events with meta data. I'd rather not mention "error" events however, since this terminology is not established and requires further standardization. This is a pending decision process which we rather not get ahead of.

rhcarvalho:
@jan-auer disregard my rant, but consider the line I linked to:

```
- Offline storage of events along with all their meta data for deferred sending.
```

I think `"events"` there is not accurate. Could you please confirm?

jan-auer:
@rhcarvalho that line is precisely what I meant with my previous comment. The issue is that we can't use "error" here, and the example specifically talks about saving events with attachments.

rhcarvalho:
> **Edit**: I could also leave that completely open and just say "Offline storage for deferred sending". Would you prefer that?

👍 

Yes, I understood that we can store any envelope, not just the ones with specific content.

PR #1652, opened 2020-04-30
chore: remove riak and cassandra nodestore backends

joshuarli:
The riak and cassandra nodestore backends don't exist anymore. They were removed:

- https://github.com/getsentry/sentry/pull/15072
- https://github.com/getsentry/sentry/pull/12965


MimiDumpling:
@PeloWriter Just a heads up in case this impacts any of your work

PR #1658, opened 2020-05-01
fix(migration): Fix export cmd on SaaS migration

BYK:


BYK:
No longer necessary after getsentry/sentry#18599.

PR #1670, opened 2020-05-07
added example code to React Native for setDist

MimiDumpling:
Customer wrote into Support about confusion over `setDist`. This PR gives a code example of how to use `setDist`.

Before:
image

After:
image

### DIFF
--- a/src/collections/_documentation/platforms/react-native/sourcemaps.md
+++ b/src/collections/_documentation/platforms/react-native/sourcemaps.md
@@ -44,3 +44,7 @@ Sentry.setRelease(RELEASE_NAME)
 `DISTRIBUTION_NAME`:
 
 : This is the version code or build ID depending on your platform. So for instance just set this to whatever is set in your _Info.plist_ or what your Gradle setup generates (For example, `52`).
+
+```js
+Sentry.setDist(`dist`);
+```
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/platforms/react-native/sourcemaps.md
```diff
@@ -44,3 +44,7 @@ Sentry.setRelease(RELEASE_NAME)
 `DISTRIBUTION_NAME`:
 
 : This is the version code or build ID depending on your platform. So for instance just set this to whatever is set in your _Info.plist_ or what your Gradle setup generates (For example, `52`).
+
+```js
+Sentry.setDist(`dist`);
```
billyvg, line 49: We mention in the previous paragraph that an example is `52`, should this match?

mbauer404: The example directly above this section for setting releases uses: Sentry.setRelease(RELEASE_NAME)`

Should we match that and use
Sentry.setDist(DISTRIBUTION_NAME)

instead?

billyvg: O yeah @mbauer404 's suggestion makes more sense since you would never hardcode a `52`

MimiDumpling: Great suggestion, @mbauer404 ! Let me know if there are other changes we should address. If not, feel free to approve. ;)

PR #1691, opened 2020-05-15
ref: Add simple migration guide for Python

untitaker:
This is https://forum.sentry.io/t/switching-to-sentry-python/4733 with some touch-ups. Paging cramer on this one because he's the other person who underwent a migration at sentry (for zeus)


bowencai8:
Sounds good, thanks Markus! I can send this to some folks for feedback.

PR #1692, opened 2020-05-15
Add `release:latest` to Search docs

MimiDumpling:
**BEFORE**:
image

**AFTER**:
image

### DIFF
--- a/src/collections/_documentation/workflow/search.md
+++ b/src/collections/_documentation/workflow/search.md
@@ -132,7 +132,7 @@ Below is a list of Issue-level tokens reserved and known to Sentry:
 
 : Restrict results to issues first seen within the given release.
 
-  Exact match on the version of a release.
+  Exact match on the version of a release, or `release:latest` to pick the most recent release.
 
 `has`
 
--- a/src/collections/_documentation/workflow/search.md
+++ b/src/collections/_documentation/workflow/search.md
@@ -199,6 +199,9 @@ Below is a list of Event-level tokens reserved and known to Sentry:
 
   Exact match on the version of a release.
 
+`release:latest`
+
+: Exact match on the most recent release.
 
 `user.id`
 `user.email`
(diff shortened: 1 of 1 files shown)


adhiraj:
@mimi `latest` is just a value they can use for the `release` keyword, so it should be within the `release` section. We can say "Exact match on the version of a release, or `latest` to pick the latest release". Same applies to the `first-release` keyword

MimiDumpling:
> @mimi `latest` is just a value they can use for the `release` keyword, so it should be within the `release` section. We can say "Exact match on the version of a release, or `latest` to pick the latest release". Same applies to the `first-release` keyword

Hi @adhiraj, `first-release` is under Issue Properties and `release` is under Event Properties. Looks like we should keep them separate to differentiate? So, I made your suggested changes to `release`, but kept `first-release` the same. Let me know if there's a better way to define the properties. 😸

adhiraj:
latest can be used with first-release too, to denote the latest release

MimiDumpling:
> `latest` value can be used with `first-release` keyword too, to denote the latest release

Nice! Thanks for the clarification. Made the changes.

src/collections/_documentation/workflow/search.md
```diff
@@ -197,8 +197,7 @@ Below is a list of Event-level tokens reserved and known to Sentry:
 
 : Restrict results to events tagged with the given release.
 
-  Exact match on the version of a release.
-
+  Exact match on the version of a release, or `release:latest` to pick the most recent release.
```
adhiraj, line 200: this should be `first-release:latest`

adhiraj: sorry I mean the one above should be

MimiDumpling: 🤦 My bad. Nice catch, Adi

PR #1695, opened 2020-05-19
docs(js): Rewrite Angular documentation and add TryCatch options

kamilogorek:
Merge only once sentry-javascript `5.16.0` is released.

### DIFF
--- a/src/collections/_documentation/platforms/javascript/angular.md
+++ b/src/collections/_documentation/platforms/javascript/angular.md
@@ -4,7 +4,7 @@ sidebar_order: 35
 ---
 
 <!-- WIZARD -->
-On its own, `@sentry/browser` will report any uncaught exceptions triggered from your application. Additionally, `@sentry/browser` can be configured to catch any Angular-specific (2.x) exceptions reported through the [@angular/core/ErrorHandler](https://angular.io/api/core/ErrorHandler) component. This is also a great opportunity to collect user feedback by using `Sentry.showReportDialog`.
+On its own, `@sentry/browser` will report any uncaught exceptions triggered from your application. Additionally, `@sentry/browser` can be configured to catch any Angular-specific exceptions reported through the [@angular/core/ErrorHandler](https://angular.io/api/core/ErrorHandler) component. This is also a great opportunity to collect user feedback by using `Sentry.showReportDialog`.
 
 First, install `@sentry/browser`:
 
@@ -16,23 +16,81 @@ yarn add @sentry/browser
 npm install @sentry/browser
 ```
 
+Then initialize new Sentry instance and configure Angular with `ErrorHandler` provided and explained below:
+
 ```typescript
 import { BrowserModule } from "@angular/platform-browser";
 import { NgModule, ErrorHandler, Injectable } from "@angular/core";
+import { HttpErrorResponse } from "@angular/common/http";
 
+import { environment } from "../environments/environment";
 import { AppComponent } from "./app.component";
 
 import * as Sentry from "@sentry/browser";
 
 Sentry.init({
-  dsn: "___PUBLIC_DSN___"
+  dsn: "___PUBLIC_DSN___",
+  // TryCatch has to be configured to disable XMLHttpRequest wrapping, as we are going to handle
+  // http module exceptions manually in Angular's ErrorHandler and we don't want it to capture same error twice.
+  // Please note that TryCatch configuration requires at least @sentry/browser v5.16.0.
+  integrations: [new Sentry.Integrations.TryCatch({
+    XMLHttpRequest: false,
+  })],
 });
 
 @Injectable()
 export class SentryErrorHandler implements ErrorHandler {
   constructor() {}
+
+  extractError(error) {
+    // Try to unwrap zone.js error.
+    // https://github.com/angular/angular/blob/master/packages/core/src/util/errors.ts
+    if (error && error.ngOriginalError) {
+      error = error.ngOriginalError;
+    }
+
+    // We can handle messages and Error objects directly.
+    if (typeof error === "string" || error instanceof Error) {
+      return error;
+    }
+
+    // If it's http module error, extract as much information from it as we can.
+    if (error instanceof HttpErrorResponse) {
+      // `error` property of http exception can be either `Error` which we can use.
+      if (error.error instanceof Error) {
+        return error.error;
+      }
+
+      // `ErrorEvent` which can provide us with the message but no stack.
+      if (error.error instanceof ErrorEvent) {
+        return error.error.message;
+      }
+
+      // Or the request body itself, which we can use as a message instead.
+      if (typeof error.error === "string") {
+        return `Server returned code ${error.status} with body "${error.error}"`;
+      }
+
+      // If we don't have any detailed information, fallback to the request message itself.
+      return error.message;
+    }
+
+    // Skip if there's no error, and let user decide what to do with it.
+    return null;
+  }
+
   handleError(error) {
-    const eventId = Sentry.captureException(error.originalError || error);
+    let extractedError = this.extractError(error) || "Handled unknown error";
+
+    // Capture handled exception and send it to Sentry.
+    const eventId = Sentry.captureException(extractedError);
+    
+    // When in development mode, log the error to console for immediate feedback.
+    if (!environment.production) {
+      console.error(extractedError);
+    }
+    
+    // Optionally show user dialog to provide details on what happened.
     Sentry.showReportDialog({ eventId });
   }
 }
@@ -45,6 +103,48 @@ export class SentryErrorHandler implements ErrorHandler {
 })
 
 export class AppModule {}
+```
+
+When using your own `ErrorHandler`, make sure that whenever you use `HttpInterceptor` alongside it,
+it doesn't modify the error captured originally.
+The same goes for writing your own API Services with built-in `http` methods.
+
+For example, the service below make it impossible for the SDK to extract correct value.
+The reason being that it's not passed down the error pipeline in it's entirety, but returning `error` property of the original instead.
+
+```js
+export class ApiService {
+  constructor(http) {
+    this.http = http;
+  }
+
+  formatErrors(e) {
+    console.log('Captured service error');
+    return throwError(e.error);
+  }
+
+  get(path) {
+    return this.http
+      .get(path)
+      .pipe(catchError(this.formatErrors));
+  }
+}
+```
+
+Instead, make sure that you always rethrow or pass-down directly original error, eg.:
+
+```js
+export class ApiService {
+  constructor(http) {
+    this.http = http;
+  }
+  
+  get(path) {
+    return this.http
+      .get(path)
+      .pipe(catchError((e) => throwError(e)));
+  }
+}
 ```
 
 ## AngularJS 1.x
--- a/src/collections/_documentation/platforms/javascript/angular.md
+++ b/src/collections/_documentation/platforms/javascript/angular.md
@@ -109,8 +109,7 @@ When using your own `ErrorHandler`, make sure that whenever you use `HttpInterce
 the interceptor doesn't modify the error captured originally.
 The same goes for writing your own API services with built-in `http` methods.
 
-For example, the service below make it impossible for the SDK to extract correct value.
-The reason being that it's not passed down the error pipeline in it's entirety, but returning `error` property of the original instead.
+For example, the service below makes it impossible for the SDK to extract the correct data, because the full, original error `e` is not propagated.
 
 ```js
 export class ApiService {
@@ -131,7 +130,7 @@ export class ApiService {
 }
 ```
 
-Instead, make sure that you always rethrow or pass-down directly original error, eg.:
+Instead, make sure that you always rethrow or directly pass the original error. For example:
 
 ```js
 export class ApiService {
--- a/src/collections/_documentation/platforms/javascript/angular.md
+++ b/src/collections/_documentation/platforms/javascript/angular.md
@@ -4,7 +4,7 @@ sidebar_order: 35
 ---
 
 <!-- WIZARD -->
-On its own, `@sentry/browser` will report any uncaught exceptions triggered from your application. Additionally, `@sentry/browser` can be configured to catch any Angular-specific exceptions reported through the [@angular/core/ErrorHandler](https://angular.io/api/core/ErrorHandler) component. This is also a great opportunity to collect user feedback by using `Sentry.showReportDialog`.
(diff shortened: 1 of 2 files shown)


src/collections/_documentation/platforms/javascript/angular.md
```diff
@@ -16,23 +16,81 @@ yarn add @sentry/browser
 npm install @sentry/browser
 ```
 
+Then initialize new Sentry instance and configure Angular with `ErrorHandler` provided and explained below:
```
lobsterkatie, line 19:
```suggestion
Then initialize a new Sentry instance and configure Angular with the `ErrorHandler` provided and explained below:
```

src/collections/_documentation/platforms/javascript/angular.md
```diff
@@ -16,23 +16,81 @@ yarn add @sentry/browser
...
+import { environment } from "../environments/environment";
 import { AppComponent } from "./app.component";
 
 import * as Sentry from "@sentry/browser";
 
 Sentry.init({
-  dsn: "___PUBLIC_DSN___"
+  dsn: "___PUBLIC_DSN___",
+  // TryCatch has to be configured to disable XMLHttpRequest wrapping, as we are going to handle
+  // http module exceptions manually in Angular's ErrorHandler and we don't want it to capture same error twice.
```
lobsterkatie, line 34:
```suggestion
  // http module exceptions manually in Angular's ErrorHandler and we don't want it to capture the same error twice.
```

src/collections/_documentation/platforms/javascript/angular.md
```diff
@@ -16,23 +16,81 @@ yarn add @sentry/browser
...
+  // Please note that TryCatch configuration requires at least @sentry/browser v5.16.0.
```
lobsterkatie, line 35: 💯

src/collections/_documentation/platforms/javascript/angular.md
```diff
@@ -16,23 +16,81 @@ yarn add @sentry/browser
...
+      error = error.ngOriginalError;
+    }
+
+    // We can handle messages and Error objects directly.
+    if (typeof error === "string" || error instanceof Error) {
+      return error;
+    }
+
+    // If it's http module error, extract as much information from it as we can.
+    if (error instanceof HttpErrorResponse) {
```
lobsterkatie, line 58: This is great. Have helped people do this piecemeal from time to time, but so much better to have it documented with a tested solution. 🎉

src/collections/_documentation/platforms/javascript/angular.md
```diff
@@ -16,23 +16,81 @@ yarn add @sentry/browser
...
+      // `error` property of http exception can be either `Error` which we can use.
```
lobsterkatie, line 59:
```suggestion
      // The `error` property of http exception can be either an `Error` object, which we can use directly...
```

src/collections/_documentation/platforms/javascript/angular.md
```diff
@@ -16,23 +16,81 @@ yarn add @sentry/browser
...
+      if (error.error instanceof Error) {
+        return error.error;
+      }
+
+      // `ErrorEvent` which can provide us with the message but no stack.
```
lobsterkatie, line 64:
```suggestion
      // ... or an`ErrorEvent`, which can provide us with the message but no stack...
```

src/collections/_documentation/platforms/javascript/angular.md
```diff
@@ -16,23 +16,81 @@ yarn add @sentry/browser
...
+      if (error.error instanceof ErrorEvent) {
+        return error.error.message;
+      }
+
+      // Or the request body itself, which we can use as a message instead.
```
lobsterkatie, line 69:
```suggestion
      // ...or the request body itself, which we can use as a message instead.
```

src/collections/_documentation/platforms/javascript/angular.md
```diff
@@ -16,23 +16,81 @@ yarn add @sentry/browser
...
-    const eventId = Sentry.captureException(error.originalError || error);
+    let extractedError = this.extractError(error) || "Handled unknown error";
+
+    // Capture handled exception and send it to Sentry.
+    const eventId = Sentry.captureException(extractedError);
+    
+    // When in development mode, log the error to console for immediate feedback.
+    if (!environment.production) {
+      console.error(extractedError);
+    }
```
lobsterkatie, line 91: If I'm reading this right, we're not actually skipping reporting, we're just reporting that we got something we didn't know how to deal with (`Handled unknown error`). In that case, can we attach the thing we did get as `extra`, for debugging purposes?

Could we do something like this? (GH won't let me put it as a suggestion, so for the purposes of reasonable syntax highlighting I've copied the whole class an elided parts I didn't change.)

```typescript
export class SentryErrorHandler implements ErrorHandler {
  constructor() {}

  extractError(error) {
    // ...

    // Nothing meaningful to extract from any of the standard spots
    // Handle this case in `handleError()`
    return null;
  }

  handleError(error) {
    // Attempt to extract an Error object or a message
    let extractedError = this.extractError(error);

    // Send error to Sentry
    let eventId;
    if (extractedError) {
      eventId = Sentry.captureException(extractedError);
    } 
    else {
      // If we weren't able to extract data from the usual spots, attach the
      // originally-captured object as `extra` for debugging purposes
      Sentry.withScope((scope) => {
        scope.setExtra("capturedError", JSON.stringify(error));
      });
      eventId = Sentry.captureException("Handled unknown error");
    }

    // When in development mode, log the error to console for immediate feedback.
    if (!environment.production) {
      console.error(
        extractedError || `Handled unknown error: ${JSON.stringify(error)}`
      );
    }

    // ...
  }
}
```

src/collections/_documentation/platforms/javascript/angular.md
```diff
@@ -47,6 +105,48 @@ export class SentryErrorHandler implements ErrorHandler {
 export class AppModule {}
 ```
 
+When using your own `ErrorHandler`, make sure that whenever you use `HttpInterceptor` alongside it,
+it doesn't modify the error captured originally.
```
lobsterkatie, line 109: > it doesn't modify the error captured originally.

I'm not 100% clear here if "it" is the custom error handler or the http interceptor. Can we specify which one?

src/collections/_documentation/platforms/javascript/angular.md
```diff
@@ -47,6 +105,48 @@ export class SentryErrorHandler implements ErrorHandler {
...
+The same goes for writing your own API Services with built-in `http` methods.
```
lobsterkatie, line 110:
```suggestion
The same goes for writing your own API services with built-in `http` methods.
```

(Unless `API Services` is a "thing" in the whole-concept, proper noun sense.)

src/collections/_documentation/platforms/javascript/angular.md
```diff
@@ -4,7 +4,7 @@ sidebar_order: 35
 ---
 
 <!-- WIZARD -->
-On its own, `@sentry/browser` will report any uncaught exceptions triggered from your application. Additionally, `@sentry/browser` can be configured to catch any Angular-specific (2.x) exceptions reported through the [@angular/core/ErrorHandler](https://angular.io/api/core/ErrorHandler) component. This is also a great opportunity to collect user feedback by using `Sentry.showReportDialog`.
+On its own, `@sentry/browser` will report any uncaught exceptions triggered from your application. Additionally, `@sentry/browser` can be configured to catch any Angular-specific exceptions reported through the [@angular/core/ErrorHandler](https://angular.io/api/core/ErrorHandler) component. This is also a great opportunity to collect user feedback by using `Sentry.showReportDialog`.
```
MimiDumpling, line 7:
```suggestion
On its own, `@sentry/browser` will report any uncaught exceptions triggered from your application. Additionally, you can configure `@sentry/browser` to catch any Angular-specific exceptions reported through the [@angular/core/ErrorHandler](https://angular.io/api/core/ErrorHandler) component. This is also a great opportunity to collect user feedback by using `Sentry.showReportDialog`.
```

src/collections/_documentation/platforms/javascript/angular.md
```diff
@@ -47,6 +105,48 @@ export class SentryErrorHandler implements ErrorHandler {
 export class AppModule {}
 ```
 
+When using your own `ErrorHandler`, make sure that whenever you use `HttpInterceptor` alongside it,
+it doesn't modify the error captured originally.
+The same goes for writing your own API Services with built-in `http` methods.
+
+For example, the service below make it impossible for the SDK to extract correct value.
```
MimiDumpling, line 112:
```suggestion
For example, the service below makes it impossible for the SDK to extract the correct value.
```

src/collections/_documentation/platforms/javascript/angular.md
```diff
@@ -47,6 +105,48 @@ export class SentryErrorHandler implements ErrorHandler {
...
+The reason being that it's not passed down the error pipeline in it's entirety, but returning `error` property of the original instead.
```
MimiDumpling, line 113:
```suggestion
The reason being that it's not passed down the error pipeline in its entirety, but returning `error` property of the original instead.
```

src/collections/_documentation/platforms/javascript/angular.md
```diff
@@ -47,6 +105,48 @@ export class SentryErrorHandler implements ErrorHandler {
...
+
+  get(path) {
+    return this.http
+      .get(path)
+      .pipe(catchError(this.formatErrors));
+  }
+}
+```
+
+Instead, make sure that you always rethrow or pass-down directly original error, eg.:
```
MimiDumpling, line 134:
```suggestion
Instead, make sure that you always rethrow or pass-down directly the original error. For example:
```

src/collections/_documentation/platforms/javascript/angular.md
lobsterkatie, line 113:
```suggestion
For example, the service below makes it impossible for the SDK to extract the correct data, because the full, original error `e` is not propagated.
```

src/collections/_documentation/platforms/javascript/angular.md
lobsterkatie, line 134:
```suggestion
Instead, make sure that you always rethrow or directly pass the original error, eg.:
```

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -725,7 +725,19 @@ This integration allows the SDK to provide original functions and method names,
...
+
+```javascript
+{
+  setTimeout: boolean;             // Wrap `setTimeout` callback functions
+  setInterval: boolean;            // Wrap `setInterval` callback functions
+  requestAnimationFrame: boolean;  // Wrap `requestAnimationFrame` callback functions
+  XMLHttpRequest: boolean;         // Wrap `XMLHttpRequest.send` callback functions
+  eventTarget: boolean | string[]; // Wrap EventTarget API callback functions. For a list of constructors wrapped by default see: https://github.com/getsentry/sentry-javascript/blob/master/packages/browser/src/integrations/trycatch.ts
+}
+```
```
lobsterkatie, line 740: Damn - how have we gone this long without this? 🤔

Good addition.

src/collections/_documentation/platforms/javascript/angular.md
lobsterkatie: @MimiDumpling - LOL, we edited the same sentence. Incorporated your edits into my suggestion so GH doesn't get confused, and am going to resolve this conversation.

src/collections/_documentation/platforms/javascript/angular.md
lobsterkatie: Same deal here, @MimiDumpling  - incorporating your fix into my larger suggestion and resolving.

src/collections/_documentation/platforms/javascript/angular.md
lobsterkatie: And here, @MimiDumpling.

(Sorry, this is probably wicked confusing, since I'm working my way up rather than down...)

lobsterkatie:
Good change, @kamilogorek!

Most of my comments are just proofreading.

src/collections/_documentation/platforms/javascript/angular.md
kamilogorek: Oh, this case happens only if `error` is `undefined`, which is pointless to provide in extras. I have no idea how this can happen, but it happened to me once 😅

src/collections/_documentation/platforms/javascript/angular.md
kamilogorek, line 109:
```suggestion
the interceptor doesn't modify the error captured originally.
```

kamilogorek: Updated

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -725,7 +725,19 @@ This integration allows the SDK to provide original functions and method names,
...
+
+Available options:
+
+```javascript
+{
+  setTimeout: boolean;             // Wrap `setTimeout` callback functions
+  setInterval: boolean;            // Wrap `setInterval` callback functions
+  requestAnimationFrame: boolean;  // Wrap `requestAnimationFrame` callback functions
+  XMLHttpRequest: boolean;         // Wrap `XMLHttpRequest.send` callback functions
+  eventTarget: boolean | string[]; // Wrap EventTarget API callback functions. For a list of constructors wrapped by default see: https://github.com/getsentry/sentry-javascript/blob/master/packages/browser/src/integrations/trycatch.ts
```
kamilogorek, line 738:
```suggestion
  eventTarget: boolean | string[]; // Wrap EventTarget API callback functions. Can provide an
                                   // array to only wrap callbacks on specific event targets. 
                                   // For a list of targets wrapped by default, see
                                   // https://github.com/getsentry/sentry-javascript/blob/master/packages/browser/src/integrations/trycatch.ts
```

src/collections/_documentation/platforms/javascript/index.md
kamilogorek: I used the first variant to make it consistent with  `Breadcrumbs` format. 
And I have no idea how we lived without it :D

kamilogorek:
Ready to merge whenever `5.16.0` is released :)

src/collections/_documentation/platforms/javascript/angular.md
lobsterkatie: What if they end up with some custom error class which doesn't inherit from `Error`? IDK how Angular works - could something like that end up coming through this path, or no?

src/collections/_documentation/platforms/javascript/index.md
lobsterkatie: 👍🏻

src/collections/_documentation/platforms/javascript/angular.md
kamilogorek: I don't think so, Angular errors all inherit. If someone hacks it and provides something that we don't support (I dont believe it'll ever happen), then they just need to update their code. The docs should stay generic imo.

lobsterkatie: 👍🏻

PR #1700, opened 2020-05-21
Add section about Performance Tab in Discover

MimiDumpling:
**BEFORE**:
image

**AFTER**:
image


lobsterkatie:
LGTM

Just to confirm: Is the result of what we discussed yesterday? If so, I think we miscommunicated - I was talking about adding something about the performance tab to the Distributed Tracing doc, rather than the Discover doc. If not, then you can ignore this. :-)

PR #1708, opened 2020-05-25
Update serverless.md

fumiki:
I found `return error` doesn't act same as default lambdaHandler. 
I think `throw error` acts as same.


simowaer:
Function should throw not fulfill: https://docs.aws.amazon.com/lambda/latest/dg/nodejs-exceptions.html

PR #1710, opened 2020-05-26
Update serverless.md

simowaer:
A lambda function should throw not fulfill with the error. cfr: https://docs.aws.amazon.com/lambda/latest/dg/nodejs-exceptions.html


simowaer:
duplicate of https://github.com/getsentry/sentry-docs/pull/1708

PR #1718, opened 2020-05-28
docs(api): Add session length info

slohmes:
Add information about session length to SSO docs. A customer wrote in asking about session length and whether it could be customized, so hopefully this'll help folks in the future self-service.

### DIFF
--- a/src/collections/_documentation/accounts/sso.md
+++ b/src/collections/_documentation/accounts/sso.md
@@ -32,7 +32,7 @@ Every member who creates a new account via SSO will be given global organization
 
 Our SSO implementation prioritizes security. We aggressively monitor linked accounts and will disable them within any reasonable sign that the account’s access may have been revoked. Generally this will be transparent to you, but if the provider is functioning in an unexpected way you may experience more frequent re-authorization requests.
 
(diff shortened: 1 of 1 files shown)


slohmes:
Lmk if there's a public docs contribution guideline doc I should be following... currently just following general sentry contribution guidelines. Thank you!

src/collections/_documentation/accounts/sso.md
```diff
@@ -32,6 +32,8 @@ Every member who creates a new account via SSO will be given global organization
 
 Our SSO implementation prioritizes security. We aggressively monitor linked accounts and will disable them within any reasonable sign that the account’s access may have been revoked. Generally this will be transparent to you, but if the provider is functioning in an unexpected way you may experience more frequent re-authorization requests.
 
+Sessions last for [Django's default session length](https://docs.djangoproject.com/en/3.0/topics/http/sessions/#using-cookie-based-sessions), which is 2 weeks. We do not currently support customizing this value.
```
PeloWriter, line 35: Suggest: "..., which is two weeks. We do not support customizing the session length."

nathan-heskia:
does it make sense to link to the docs for the Django version used in Sentry?

slohmes:
> does it make sense to link to the docs for the Django version used in Sentry?

Oh good catch, I forgot we aren't on latest -- thanks!

PR #1721, opened 2020-05-29
doc: Add information about our use of asgiref.local in the py sdk

untitaker:


untitaker:
@rhcarvalho this is just telling you the behavior of the SDK and enumerating all edge cases

This PR is obsolete as it seems we're not going to implement this hack anyway, if https://github.com/django/asgiref/pull/171 lands contextvars will just work fine.

PR #1722, opened 2020-05-29
Fix typo

mlissner:
It's on the homepage. Oh no!

PS - Thanks for the non-profit discount. You're really going to make a difference with it.


untitaker:
Thanks!

PR #1720, opened 2020-05-29
react-native release health

bruno-garcia:
Document the _Release Health_ feature for React Native users.

Please note I've removed the `Sentry.setRelease/setDist` docs since cannot affect the release used by the native SDKs doing release health. Values should be taken in during initialization. We might deprecate the methods in the future if possible.

PR #1724, opened 2020-05-30
fix spelling

bruno-garcia:
@MimiDumpling reviews from #1720

Please note a section there is copied over into 3 docs, maybe we can use an include instead?


MimiDumpling:
> @MimiDumpling reviews from #1720
> 
> Please note a section there is copied over into 3 docs, maybe we can use an include instead?

Yeah, I think using `include` would be great here.

PR #1728, opened 2020-06-03
feat(sourcemaps): Mention volume sharing across web & workers

daicoden:
Add something to the docs about how to get source maps to work with docker deployed environment.

### DIFF
--- a/src/collections/_documentation/platforms/javascript/sourcemaps.md
+++ b/src/collections/_documentation/platforms/javascript/sourcemaps.md
@@ -428,6 +428,10 @@ The Sentry API currently only works with source maps and source files that are u
 
 This sometimes occurs with build scripts and plugins that produce pre-compressed minified files. For example, Webpack’s [compression plugin](https://github.com/webpack/compression-webpack-plugin). You’ll need to disable such plugins and perform the compression _after_ the generated source maps / source files have been uploaded to Sentry.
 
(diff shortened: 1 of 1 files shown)


BYK:
Why do we need this, the on-premise repo ensures this already?

daicoden:
Apologies, I guess I'm running an older version of sentry. 9.0 I started everything manually. This was the fix for me, but it might be worth putting in the README something generic that for sourcemaps to work the worker and web need to share a volume - however that may be.

daicoden:
FYI @BYK

from: https://hub.docker.com/_/sentry/

```
SENTRY_FILESTORE_DIR
Directory where uploaded files will be stored. This defaults to /var/lib/sentry/files and is a VOLUME for persistent data.
```

BYK:
@daicoden 

>Apologies, I guess I'm running an older version of sentry. 9.0 I started everything 
manually.

No need to apologize, in fact _I apologize_ as you took the extra mile to submit a fix for an issue you encountered and reading my initial response, it is not very encouraging.

Thanks a lot for submitting this 🙂 

>from: hub.docker.com/_/sentry

This image is locked on 9.1.2 unfortunately and we are working with Docker folks to update it. We encourage everyone to use [the official on-premise repo](https://github.com/getsentry/onpremise) for installation in all channels but I guess we need to update this page too :)

PR #1731, opened 2020-06-04
Explicit Scope for JS SDKs

kamilogorek:
ref: https://github.com/getsentry/sentry-javascript/pull/2627
ref: https://app.asana.com/0/1169125731075107/1178479294693625

### DIFF
--- a/src/collections/_documentation/platforms/javascript/index.md
+++ b/src/collections/_documentation/platforms/javascript/index.md
@@ -437,6 +437,52 @@ Sentry.setExtra("character_name", "Mighty Fighter");
   level="warning"
 %}
 
+### Passing Context Directly
+
+Starting version `5.16.0` of our JavaScript SDKs, some of the contextual data can be provided directly to `captureException` and `captureMessage` calls.
+Provided data will be merged with the one that is already stored inside the current scope, unless explicitly cleared using a callback method.
+
+It works in three different variations:
+
+- plain object containing updatable attributes
+- scope instance that will extract the attributes from
+- callback function that will receive the current scope as an argument and allow for modifications
+
+Possible context keys allowed to be passed are: `tags`, `extra`, `contexts`, `user`, `level`, `fingerprint`.
+
+#### Example Usages
+
+```javascript
+Sentry.captureException(new Error("something went wrong"), {
+  tags: {
+    section: "articles",
+  }
+});
+```
+
+```javascript
+// Explicitly clear what has been already stored on the scope
+Sentry.captureException(new Error("clean as never"), (scope) => {
+  scope.clear();
+  scope.setTag("clean", "slate");
+  return scope;
+});
+```
+
+```javascript
+// Use Scope instance to store the data (it will still merge with the global scope)
+const scope = new Sentry.Scope();
+scope.setTag("section", "articles");
+Sentry.captureException(new Error("something went wrong"), scope);
+```
+
+```javascript
+// Use Scope instance to store the data and ignore global one completely
+const scope = new Sentry.Scope();
+scope.setTag("section", "articles");
+Sentry.captureException(new Error("something went wrong"), () => scope);
+```
+
 ### Unsetting Context
 Context is held in the current scope and thus is cleared out at the end of each operation --- request, etc. You can also push and pop your own scopes to apply context data to a specific code block or function.
 
--- a/src/collections/_documentation/platforms/javascript/index.md
+++ b/src/collections/_documentation/platforms/javascript/index.md
@@ -314,49 +314,6 @@ You can also set context when manually triggering events.
 
 {% include platforms/event-contexts.md %}
 
-### Extra Context {#extra-context}
-In addition to the structured context that Sentry understands, you can send arbitrary key/value pairs of data which the Sentry SDK will store alongside the event. These are not indexed, and the Sentry SDK uses them to add additional information about what might be happening:
-
-```javascript
-Sentry.setExtra("character_name", "Mighty Fighter");
-```
-
-{% capture __alert_content -%}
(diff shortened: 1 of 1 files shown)


src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -464,6 +421,95 @@ Sentry.withScope(function(scope) {
 });
 ```
 
+### Extra Context {#extra-context}
+In addition to the structured context that Sentry understands, you can send arbitrary key/value pairs of data which the Sentry SDK will store alongside the event. These are not indexed, and the Sentry SDK uses them to add additional information about what might be happening:
+
+```javascript
+Sentry.setExtra("character_name", "Mighty Fighter");
+```
```
HazAT, line 429: We decided to de-emphasize the use of guide users to call `setExtra` we always want them to use `setContext`

I know this was there before but let's fix it now

kamilogorek: Fixed

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -464,6 +421,95 @@ Sentry.withScope(function(scope) {
...
+
+### Passing Context Directly
+
+Starting version `5.16.0` of our JavaScript SDKs, some of the contextual data can be provided directly to `captureException` and `captureMessage` calls.
+Provided data will be merged with the one that is already stored inside the current scope, unless explicitly cleared using a callback method.
+
+It works in three different variations:
+
+- plain object containing updatable attributes
+- scope instance that will extract the attributes from
```
AbhiPrasad, line 448: > extract the attributes from

from what?

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -464,6 +421,95 @@ Sentry.withScope(function(scope) {
...
+
+```javascript
+// Use Scope instance to store the data (it will still merge with the global scope)
+const scope = new Sentry.Scope();
+scope.setTag("section", "articles");
+Sentry.captureException(new Error("something went wrong"), scope);
+```
+
+```javascript
+// Use Scope instance to store the data and ignore global one completely
```
AbhiPrasad, line 480:
```suggestion
// Use Scope instance to store the data and ignore the global one completely
```

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -464,6 +421,95 @@ Sentry.withScope(function(scope) {
...
+Sentry.withScope(function(scope) {
+  scope.setUser(someUser);
+  Sentry.captureException(error);
+});
+```
+
+If you want to remove globally configured data from the scope, you can call:
+
+```javascript
+Sentry.configureScope(scope => scope.clear())
```
AbhiPrasad, line 505: Will this affect stuff on scope set by integrations? (Or is that even something we need to worry about).

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -464,6 +421,95 @@ Sentry.withScope(function(scope) {
...
+### Unsetting Context
+Context is held in the current scope and thus is cleared out at the end of each operation --- request, etc. You can also push and pop your own scopes to apply context data to a specific code block or function.
+
+There are two different scopes for unsetting context --- a global scope which Sentry does not discard at the end of an operation, and a scope that can be created by the user.
+
+```javascript
+// This will be changed for all future events
+Sentry.setUser(someUser);
+
+// This will be changed only for the error caught inside and automatically discarded afterward
```
AbhiPrasad, line 495: > automatically discarded afterward

Seems to be unclear wording here, but I'm not sure how to better phrase it.

src/collections/_documentation/platforms/javascript/index.md
kamilogorek, line 448:
```suggestion
- scope instance that we will extract the attributes from
```

src/collections/_documentation/platforms/javascript/index.md
kamilogorek: If integration called `Sentry.configureScope` then yes. But we never had any issues with that wording.

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -464,6 +421,95 @@ Sentry.withScope(function(scope) {
...
+
+```javascript
+// Use Scope instance to store the data (it will still merge with the global scope)
+const scope = new Sentry.Scope();
+scope.setTag("section", "articles");
+Sentry.captureException(new Error("something went wrong"), scope);
+```
+
+```javascript
+// Use Scope instance to store the data and ignore the global one completely
```
PeloWriter, line 480: clarify what "one" means here - "ignore the global scope" perhaps?

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -464,6 +421,95 @@ Sentry.withScope(function(scope) {
...
+const scope = new Sentry.Scope();
+scope.setTag("section", "articles");
+Sentry.captureException(new Error("something went wrong"), () => scope);
+```
+
+### Unsetting Context
+Context is held in the current scope and thus is cleared out at the end of each operation --- request, etc. You can also push and pop your own scopes to apply context data to a specific code block or function.
```
PeloWriter, line 487: avoid use of Latin terms (etc., e.g., i.e.) - instead say something like "Context is held in the current scope, thus cleared out at the end of each operation - request and so forth."

src/collections/_documentation/platforms/javascript/index.md
```diff
@@ -464,6 +421,95 @@ Sentry.withScope(function(scope) {
...
+
+There are two different scopes for unsetting context --- a global scope which Sentry does not discard at the end of an operation, and a scope that can be created by the user.
```
PeloWriter, line 489: Avoid there is/there are constructions - it leads to dead writing.

Suggest "Sentry supports two different scopes for unsetting context:
-  a global scope, which Sentry does not discard at the end of an operation
- a scope created by the user 

Then add when/if Sentry discards the second type of scope so we have parallel information.

kamilogorek: When/if are explained in the code comments

kamilogorek:
Fixed all review comments. Thanks!

PR #1733, opened 2020-06-06
fix: Fix bad example for setContext in js SDK

jennmueng:
In the js SDK, the `Sentry.setContext` method's second parameter takes in
an object only and not a string as per http://getsentry.github.io/sentry-javascript/classes/hub.scope.html#setcontext.

Users will attempt to pass a string:
https://github.com/getsentry/sentry-react-native/issues/694#issuecomment-640069758

Attempting to pass a string in the second parameter will encounter an error when processing the event:
Screen Shot 2020-06-06 at 10 33 22 PM

### DIFF
--- a/src/collections/_documentation/platforms/javascript/index.md
+++ b/src/collections/_documentation/platforms/javascript/index.md
@@ -422,10 +422,14 @@ Sentry.withScope(function(scope) {
 ```
 
 ### Extra Context {#extra-context}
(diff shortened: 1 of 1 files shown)


untitaker:
Might wanna align with other platforms examples of the page but I am p sure you did (on mobile right now)

HazAT:
For more context, see latest entry
https://www.notion.so/sentry/Decision-Log-9c3a1f7d9f364724ade6962de9d1b22b

PR #1740, opened 2020-06-09
fix(md): html escaping issues in code block

untitaker:
Screenshot 2020-06-09 at 18 35 40


untitaker:
That breaks in github markdown rendering. something is broken fundamentally

untitaker:
We basically have no way to satisfy both md renderers so let's just go with our own one

PR #1747, opened 2020-06-12
docs(async-csv): Add instructions for checking SHA1 checksum on CSV

Zylphrex:


src/collections/_documentation/performance/discover/query-builder.md
```diff
@@ -314,7 +314,24 @@ A down arrow indicates sorting the column in descending order. An up arrow indic
 
 ### Download CSV
 
-Note, only the current values visible on the table are exported, not the paginated results. To export the table as CSV, click on the "Download CSV" button.
+If you want to take the data elsewhere, click on the "Export" button to start the export. The results are going to take a while, so sit back and relax. You'll get an email with the download link once they're ready.
```
MimiDumpling, line 317: :) Let's be more specific about how long this will take. Do we have an estimation of time?

MimiDumpling: Also, our docs have a more formal tone. Let's remove the informal voice here.
```suggestion
If you want to take the data elsewhere, click on the "Export" button to start the export. The results are going to take a while. You'll get an email with the download link once they're ready.
```

src/collections/_documentation/performance/discover/query-builder.md
```diff
@@ -314,7 +314,24 @@ A down arrow indicates sorting the column in descending order. An up arrow indic
...
+
+{% capture __alert_content -%}
+The results are limited to 10 million rows or 2GB whichever comes first.
```
MimiDumpling, line 320:
```suggestion
The results are limited to 10 million rows or 2GB, whichever comes first.
```

src/collections/_documentation/performance/discover/query-builder.md
```diff
@@ -314,7 +314,24 @@ A down arrow indicates sorting the column in descending order. An up arrow indic
...
+{%- endcapture -%}
+{%- include components/alert.html
+    title="Note"
+    content=__alert_content
+    level="info"
+%}
+
+#### Verifying the Download
+
+The download page contains a `SHA1` checksum of the file which you can use to verify its integrity with the following command.
```
MimiDumpling, line 330:
```suggestion
The download page contains a `SHA1` checksum of the file, which you can use to verify its integrity with the following command:
```

MimiDumpling (commented):
🌻 Thanks for updating the docs, Tony!

src/collections/_documentation/performance/discover/query-builder.md
Zylphrex: Unfortunately this can be anywhere from 1 second to over an hour, so I don't have a good estimate to put here.

src/collections/_documentation/performance/discover/query-builder.md
```diff
@@ -314,7 +314,24 @@ A down arrow indicates sorting the column in descending order. An up arrow indic
 
 ### Download CSV
 
-Note, only the current values visible on the table are exported, not the paginated results. To export the table as CSV, click on the "Download CSV" button.
+If you want to take the data elsewhere, click on the "Export" button to start the export. The results are going to take a while. You'll get an email with the download link once they're ready.
```
MimiDumpling, line 317: Can we say "depending on the file size, the results can take between one second to one hour"

Zylphrex: Would that be misleading if the results can take well over an hour? I've seen it take like 90 minutes, and there isn't really a strict upper bound as how long it'll take so it can potentially go above that.

MimiDumpling: How about:

Depending on the file size, the result times will vary.

MimiDumpling: Is that statement accurate? I'm assuming the results are dependent on the file size.

Zylphrex: Can we can say

Depending on the amount of data, the wait times will vary.

MimiDumpling: 👍 That's great! Thanks, Tony

MimiDumpling (approved):
🎉

src/collections/_documentation/performance/discover/query-builder.md
Zylphrex:
```suggestion
If you want to take the data elsewhere, click on the "Export" button to start the export. Depending on the amount of data, the wait times will vary. You'll get an email with the download link once they're ready.
```

dashed (approved):

k-fish (approved):

PR #1758, opened 2020-06-16
fix: move multidex support to the Advanced Usage section

marandaneto:
placing it in the right section.


src/collections/_documentation/platforms/android/index.md
```diff
@@ -624,6 +624,25 @@ If you want to have control over what breadcrumbs are attached to your data, you
 
 That allows the user to decide whether and how a breadcrumb should be sent.
 
+### Multi-Dex support
```
MimiDumpling, line 627:
```suggestion
### Multi-Dex Support
```

PR #1760, opened 2020-06-16
Update terminology

PeloWriter:
Removing outdated terminology - LMK if the update for the Node isn't quite complete and I can remove that change.


HazAT:
ref https://github.com/getsentry/sentry-docs/pull/1825

PR #1774, opened 2020-06-23
Update performance page to expected navigation and fix nav issues

PeloWriter:


src/collections/_documentation/performance/getting-started.md
```diff
@@ -12,6 +12,8 @@ Learn more about performance monitoring in:
 - [Metrics]({% link _documentation/performance/performance-metrics.md %}): provides insight into user experience of your application based on custom thresholds
 - [Discover]({% link _documentation/performance/discover/index.md %}): view data across environments with pre-built queries. Customers subscribed to the Team or Business plan can use Discover to view comprehensive information sent to Sentry.
 
+## Install
```
benvinegar, line 15: I'm assuming "Install" is the same terminology we use for this in similar sections.

benvinegar (approved):

src/collections/_documentation/performance/getting-started.md
PeloWriter: We use variations (Install the SDK, Installation), but Install is the heading used in the template (thus will become the standard).

PR #1775, opened 2020-06-23
feat: Add retrieving Transaction

HazAT:
This PR adds the last missing piece of how to retrieve a Transaction for SDKs.

Related:
- https://github.com/getsentry/sentry-javascript/pull/2668
- https://github.com/getsentry/sentry-python/pull/734


src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -380,7 +380,7 @@ import sentry_sdk
 while True:
   item = get_from_queue()
 
-  with sentry_sdk.start_span(op="task", transaction=item.get_transaction()):
+  with sentry_sdk.startTransaction(op="task", name=item.get_transaction()):
```
rhcarvalho, line 383:
```suggestion
  with sentry_sdk.start_transaction(op="task", name=item.get_transaction()):
```

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -530,6 +534,24 @@ function processItem(item, transaction) {
 }
 ```
 
+#### Retrieving a Transaction
+
+In cases where you want to attach Spans to an already ongoing Transaction you can use `Sentry.getCurrentHub().getScope().getTransaction()`. This function will return a `Transaction` in case there is a running Transaction otherwise it returns `undefined`. If you are using our Tracing integration by default we attach the Transaction to the Scope. So you could do something like this:
+
+```javascript
+function myJsFunction() {
+  const transaction = Sentry.getCurrentHub().getScope().getTransaction();
```
rhcarvalho, line 543: Wouldn't this be a good case to use `withScope`?


```js
func f() {
  Sentry.withScope((scope) => {
    const transaction = scope.getTransaction();
    let span;
    if (transaction) {
      span = transaction.startChild({
        op: "encode",
        description: "parseAvatarImages"
      });
    } else {
      span = hub.startTransaction({
        name: "f",
        op: "encode",
        description: "parseAvatarImages"
      });
    }
    try {
      // Do work
    } finally {
      span.finish();
    }
  });
}

// did not test, may contain typos; typed in GitHub directly.
```

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -616,6 +638,25 @@ app.use(function processItems(req, res, next) {
...
+```javascript
+app.get("/success", function successHandler(req, res) {
+  const transaction = Sentry.getCurrentHub().getScope().getTransaction();
+  if (transaction) {
+    let span = transaction.startChild({
+      op: "encode",
+      description: "parseAvatarImages"
+    });
+    // Do something
+    span.finish();
```
rhcarvalho, line 654: In Python we have context managers. The way they work guarantees that even if there is an exception in the middle of a span, the span will be finished.

Without `withTransaction` and `withSpan` wrappers, it might be a good idea to document using `try...finally` and calling `finish()` in the `finally` block.

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -530,6 +534,24 @@ function processItem(item, transaction) {
 }
 ```
 
+#### Retrieving a Transaction
+
+In cases where you want to attach Spans to an already ongoing Transaction you can use `Sentry.getCurrentHub().getScope().getTransaction()`. This function will return a `Transaction` in case there is a running Transaction otherwise it returns `undefined`. If you are using our Tracing integration by default we attach the Transaction to the Scope. So you could do something like this:
```
kamilogorek, line 539:
```suggestion
In cases where you want to attach a span to already ongoing transaction, you can use `Sentry.getCurrentHub().getScope().getTransaction()` call. This function will return a `Transaction` in case there is a running transaction or `undefined` otherwise. If you are using our Tracing integration, we attach the transaction to the scope by default, so you could do something like this:
```

src/collections/_documentation/performance/distributed-tracing.md
kamilogorek: `withScope` creates a new scope, but we could use `configureScope` here just fine imo.

```js
Sentry.configureScope(scope => {
  const transaction = scope.getTransaction();
  if (transaction) {
    let span = transaction.startChild({
      op: "encode",
      description: "parseAvatarImages"
    });
    // Do something
    span.finish();
  }
});

File: src/collections/_documentation/performance/distributed-tracing.md [hunk 5]
```diff
@@ -616,6 +638,25 @@ app.use(function processItems(req, res, next) {
 });
 ```
 
+#### Retrieving a Transaction
+
+In cases where you want to attach Spans to an already ongoing Transaction you can use `Sentry.getCurrentHub().getScope().getTransaction()`. This function will return a `Transaction` in case there is a running Transaction otherwise it returns `undefined`. If you are using our Express integration by default we attach the Transaction to the Scope. So you could do something like this:
```
kamilogorek, line 643:
```suggestion
In cases where you want to attach a span to already ongoing transaction, you can use `Sentry.getCurrentHub().getScope().getTransaction()` call. This function will return a `Transaction` in case there is a running transaction or `undefined` otherwise. If you are using our Tracing integration, we attach the transaction to the scope by default, so you could do something like this:
```

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -403,6 +403,10 @@ def process_item(item):
       span.set_data("http.foobarsessionid", get_foobar_sessionid())
 ```
 
+#### Retrieving a Transaction
+
+TODO
```
rhcarvalho, line 408: This is going to be along the lines of:

```py
from sentry_sdk import Hub

transaction = Hub.current.scope.transaction

# ... start child span or new transaction depending whether `transaction is None`
```

HazAT:
We need to rebase this, the SDK docs have moved into their own page.

src/collections/_documentation/performance/distributed-tracing.md
```diff
@@ -530,6 +534,24 @@ function processItem(item, transaction) {
 }
 ```
 
+#### Retrieving a Transaction
```
rhcarvalho, line 537: The same heading appearing multiple times in the same page (3x in this page) will all generate the same anchor link.

Do we have some helper to customize the anchor name? We could use that to make those 3 headings use different anchors.

HazAT:
I rebased the PR ontop of master

HazAT:
There is still a todo and we need to release python so we can't merge this yet

rhcarvalho:
Filled in Python TODOs.

PR #1778, opened 2020-06-23
Add initial Gatsby capabilities

dcramer:
This is a first pass at getting Gatsby available on the docs site, which will allow us to begin migrating content over. See the additions to `README.md` for an understanding of how this works.


.babelrc.js
```diff
@@ -0,0 +1,13 @@
...
+
+// Jest needs to compile this code, but generally we don't want this copied
+// to output folders
+if (process.env.NODE_ENV !== `test`) {
+  ignore.push(`**/__tests__`)
+}
+
+module.exports = {
+  sourceMaps: true,
+  presets: ["@babel/preset-env", "babel-preset-gatsby-package"],
```
dcramer, line 11: im not sure if preset-env is needed, but i didnt wanna spend any time to figure it out.

dcramer:
Got the start of the proxy + converting /meta/. Proxy isnt serving the route correctly yet (probably need to configure something else), but that should be quick to fix.

I havent yet deleted the pre-existing files as I'm pretty sure they're needed to generate the sidebar. I think what I'll instead do is try to generate more of the sidebar by hand, and then duplicate that for the Gatsby file.

dcramer:
Oh, I also messed around w/ the Jekyll config as I couldn't get it running (again). Hoping that this set of deps works fully for everyone from a current gen Ruby via Homebrew.

dcramer:
Also found something odd happening w/ Gatsby and the SEO component:

image

dcramer:
TODO: can we import the gatsby_src css assets via jekyll-assets to avoid replicating them? that'd still generate a new css.

Alternative idea: overwrite the jekyll css with the gatsby CSS - meaning you wouldnt get any styles if you didnt load both.

dcramer:
Also this disables the pjax-style fast loader for links on the jekyll site as I don't want to deal w/ fixing it. Problem wont exist with CDN speeds.

dcramer:
not sure how we're gonna deal w/ algolia in the transition

dcramer:
new plan: remove usage of jekyll tags as much as possible first, that will let us direct import some files (e.g. {% link %} can fully die)

dcramer:
next steps:

- merge link changes into master
- replace all alert includes with raw html (to master, not this branch)
- replace version_added tag with raw html

dcramer:
Aside, if there's a way to make InstantSearch use multiple indexes.. thats a really easy path to solve search.

dcramer:
I fixed the Algolia concerns - need to figure out best mechanism to deploy (ideally we can just pick up env vars, but im not sure we can in this case).

PR #1785, opened 2020-06-24
feat: Docs for react native touch events

jennmueng:
Add docs describing how to use the new touch events feature for `@sentry/react-native` soon to be introduced.

### DIFF
--- a/src/collections/_documentation/platforms/react-native/touchevents.md
+++ b/src/collections/_documentation/platforms/react-native/touchevents.md
@@ -39,10 +39,10 @@ const App = () => {
 export default AppRegistry.registerComponent("Your Amazing App", () => Sentry.withTouchEventBoundary(App));;
 ```
 
-#### How Touches are Tracked
-Each touch event that occurs will be automatically logged as a breadcrumb, and will be displayed on the dashboard when an event occurs along with the component tree that the touch event occurred in.
+## How Touches are Tracked Automatically
+Each touch event that occurs will be automatically logged as a breadcrumb, and will be displayed on the dashboard when an event occurs along with the component tree that the touch event occurred in. This component tree is logged using the `name` property of each component, which by default is the name given to it when it was defined or the name given to it by React Native such as "View" or "ScrollView".
 
-## Naming Specific Components
+## Tracking Specific Components
 You can let Sentry know which components to specifically track by setting the `displayName` property on them. If Sentry detects a component with a `displayName` within a touch's component tree, it will be logged nicely on the dashboard as having occurred in that component.
 ```jsx
 class YourCoolComponent extends React.Component {
--- a/src/collections/_documentation/platforms/react-native/touchevents.md
+++ b/src/collections/_documentation/platforms/react-native/touchevents.md
@@ -4,7 +4,7 @@ title: 'Tracking Touch Events'
 
 {% version_added 1.5.0 %}
 
-To track touch events with the React Native SDK you will need at least version `1.5.0` and will have to wrap your app with a `TouchEventBoundary`. This can be done in two ways:
+To track touch events with the React Native SDK, you will need at least version `1.5.0` and will have to wrap your app with a `TouchEventBoundary`. You can track touch events in two ways:
 
 ## Wrapping with `TouchEventBoundary`
 At the root of your app, usually `App.js`, wrap the app component with `Sentry.TouchEventBoundary`:
@@ -39,7 +39,7 @@ const App = () => {
 export default AppRegistry.registerComponent("Your Amazing App", () => Sentry.withTouchEventBoundary(App));;
 ```
 
-# How Touches are Tracked
+#### How Touches are Tracked
 Each touch event that occurs will be automatically logged as a breadcrumb, and will be displayed on the dashboard when an event occurs along with the component tree that the touch event occurred in.
 
 ## Naming Specific Components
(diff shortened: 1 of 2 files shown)


marandaneto:
@jennmueng just made it a draft, so we don't merge by accident before cutting a RN release.

src/collections/_documentation/platforms/react-native/touchevents.md
```diff
@@ -0,0 +1,106 @@
+---
+title: 'Tracking Touch Events'
```
marandaneto, line 2: what's about being a Breadcrumbs page instead of Touch events? I believe we'll be adding more of those and they would make sense to be aggregated IMO

src/collections/_documentation/platforms/react-native/touchevents.md
```diff
@@ -0,0 +1,106 @@
...
+---
+
+{% version_added 1.5.0 %}
+
+To track touch events with the React Native SDK you will need at least version `1.5.0` and will have to wrap your app with a `TouchEventBoundary`. This can be done in two ways:
```
AbhiPrasad, line 7: What happens if multiple `TouchEventBoundaries` are used? Should we put a warning against that?

src/collections/_documentation/platforms/react-native/touchevents.md
```diff
@@ -0,0 +1,106 @@
...
+        </View>
+    )
+}
+
+YourCoolComponent.displayName = "CoolComponent";
+```
+
+# Options
+
+Options can be passed to configure the boundary either as props to the `Sentry.TouchEventBoundary` component or as the second argument to the `Sentry.withTouchEventBoundary` wrapper.
```
MimiDumpling, line 74:
```suggestion
You can pass specific options to configure the boundary either as props to the `Sentry.TouchEventBoundary` component or as the second argument to the `Sentry.withTouchEventBoundary` wrapper.
```

src/collections/_documentation/platforms/react-native/touchevents.md
jennmueng: Hmm..yes it would make sense when we add more docs on breadcrumbs, but having this alone in there right now would be weird I'd think.

src/collections/_documentation/platforms/react-native/touchevents.md
```diff
@@ -0,0 +1,106 @@
...
+        <Sentry.TouchEventBoundary>
+            <RestOfTheApp />
+        </Sentry.TouchEventBoundary>
+    );
+}
+
+export default AppRegistry.registerComponent("Your Amazing App", () => App);;
+```
+
+## Using the `withTouchEventBoundary` Wrapper
```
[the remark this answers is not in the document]
jennmueng: Ah yes, definitely, I'll change it to use the HOC term.

src/collections/_documentation/platforms/react-native/touchevents.md
jennmueng: You can actually have multiple if you want to wrap specific parts of the app, but they should not be nested. Good catch, will add that.

src/collections/_documentation/platforms/react-native/touchevents.md
```diff
@@ -0,0 +1,106 @@
...
+    return (
+        <RestOfTheApp />
+    );
+}
+
+export default AppRegistry.registerComponent("Your Amazing App", () => Sentry.withTouchEventBoundary(App));;
+```
+
+# How Touches are Tracked
+Each touch event that occurs will be automatically logged as a breadcrumb, and will be displayed on the dashboard when an event occurs along with the component tree that the touch event occurred in.
```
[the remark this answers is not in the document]
jennmueng: It actually uses `name` to unless there is a `displayName` present. Should make this more clear then.

src/collections/_documentation/platforms/react-native/touchevents.md
```diff
@@ -0,0 +1,116 @@
...
+{%- endcapture -%}
+{%- include components/alert.html
+  title="Multiple Boundaries"
+  content=__alert_content
+  level="warning"
+  deep_link="multiple-boundaries"
+%}
+
+## How Touches are Tracked Automatically
+Each touch event that occurs will be automatically logged as a breadcrumb, and will be displayed on the dashboard when an event occurs along with the component tree that the touch event occurred in. This component tree is logged using the `name` property of each component, which by default is the name given to it when it was defined or the name given to it by React Native such as "View" or "ScrollView".
```
AbhiPrasad, line 53:
```suggestion
Each touch event that occurs will be automatically logged as a breadcrumb, and will be displayed on the dashboard when an event occurs along with the component tree that the touch event occurred in. This component tree is logged using the `name` property of a component. By default, React Native will set this property automatically on components.
```

> This component tree is logged using the `name` property of each component, which by default is the name given to it when it was defined or the name given to it by React Native such as "View" or "ScrollView".

Run on sentence, and we probably can just avoid going super detailed

As an aside did you use https://github.com/facebook/react-native/blob/3276563806966c16256181d9e71099f63f26e67f/Libraries/Renderer/implementations/ReactFabric-prod.js#L1020-L1060 when working on this?

jennmueng: Ah no I didn't, I'll see if we can get access to that method. Thanks!

jennmueng:
@MimiDumpling Are we good to merge this?

MimiDumpling:
> @MimiDumpling Are we good to merge this?

Yep yep! Feel free to take the glory of the merge! 🙌

PR #1795, opened 2020-06-26
feat(cocoa): Update sdk configuration

philipphofmann:


src/collections/_documentation/platforms/cocoa/init/objc.md
```diff
@@ -2,9 +2,9 @@
 @import Sentry;
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
-    [SentrySDK startWithOptions:@{
-        @"dsn": @"___PUBLIC_DSN___",
-        @"debug": @(YES)
+    [SentrySDK startWithConfigureOptions:^(SentryOptions *options) {
```
marandaneto, line 5: if they are on the older version, this will likely not work.

What's about doing something like:

Version <= do like this:
...
Version >= do like this:
...

src/collections/_documentation/platforms/cocoa/init/swift.md
```diff
@@ -5,10 +5,10 @@ import Sentry // Make sure you import Sentry
 
 func application(_ application: UIApplication, 
     didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
-    SentrySDK.start(options: [
-        "dsn": "___PUBLIC_DSN___",
-        "debug": true // Enabled debug when first installing is always helpful
-    ])
+    SentrySDK.start { options in
```
marandaneto, line 8: same as https://github.com/getsentry/sentry-docs/pull/1795/files#r446074337

src/collections/_documentation/platforms/cocoa/init/objc.md
philipphofmann: Very good point 👏

PR #1796, opened 2020-06-26
feat(python): Expand troubleshooting guide

rhcarvalho:
Document cloning hub when using asyncio and the general idea of having a Hub instance per concurrency unit.

Related: https://github.com/getsentry/sentry-python/issues/147

---

## Before

image

## After

image


rhcarvalho:
The test failure is due to changing URLs... `/platforms/python/contextvars/` => `/platforms/python/troubleshooting/`
Shall we add a symlink or a redirect to keep the old URL working?

src/collections/_documentation/platforms/python/troubleshooting.md
```diff
@@ -1,19 +1,58 @@
...
+```python
+with Hub(Hub.current):
+    # in this block Hub.current refers to a new clone
+    # of the original hub, with the same client and
+    # the same initial scope data.
+```
+
+Issues with `asyncio` have then an easy workaround: every `asyncio` coroutine
+that really does run concurrently with other coroutines needs to be made into a
+task, then the hub needs to be cloned and reassigned.
```
untitaker, line 43:
```suggestion
task, then the hub needs to be cloned as shown above.
```

untitaker:
If you want to add a redirect `nginx.conf` is the place to put it iirc

rhcarvalho:
> If you want to add a redirect `nginx.conf` is the place to put it iirc

Thanks. I've added a redirect, though looking at that config, I suspect we've not added redirects for mirror things like this one might be.

Also opened up the question why we do 302 redirects instead of 301 (supposedly better for SEO).

src/collections/_documentation/platforms/python/troubleshooting.md
rhcarvalho: Thanks. I was not sure what you meant with "reassigned" originally, but now I see that `Hub.__enter__` sets the current hub, so that should be enough.

I wanted to add a more complete example, similar to what is in https://github.com/getsentry/sentry-python/issues/147 but with your suggested correction applied. WDYT?

src/collections/_documentation/platforms/python/troubleshooting.md
```diff
@@ -1,19 +1,58 @@
...
+The general recommendation is to have one hub per "concurrency unit"
+(thread/coroutine/etc). The SDK ensures every thread has an independent hub. If
+you do concurrency with `asyncio` coroutines, clone the current hub for use
+within a block that runs concurrent code:
+
+```python
+with Hub(Hub.current):
+    # in this block Hub.current refers to a new clone
+    # of the original hub, with the same client and
+    # the same initial scope data.
```
rhcarvalho, line 38: @untitaker if you'd be able to help here, I wanted to add a more complete example, similar to what is in getsentry/sentry-python#147 but with your suggested correction applied.

src/collections/_documentation/platforms/python/index.md
```diff
@@ -291,8 +291,8 @@ The Python SDK provides some common [hints](/error-reporting/configuration/filte
 
 ## Troubleshooting and known issues
 
-This section contains various gotchas that are inherent to the Python SDK's
-design, do not necessarily affect the majority of users and do not really fit
-anywhere else.
+Sometimes a little advice from a friend is all it takes to solve a problem.
```
PeloWriter, line 294: I like the friendly tone, but perhaps we might simply say "The [Troubleshooting](/platforms/python/troubleshooting) section provides advice and tips to help solve common problems that may occur when using the Python SDK."

src/collections/_documentation/platforms/python/index.md
```diff
@@ -291,8 +291,8 @@ The Python SDK provides some common [hints](/error-reporting/configuration/filte
...
 
-- [Contextvars vs thread locals](/platforms/python/contextvars/)
+We have compiled a few tips to help you use the Python SDK even under the most
```
PeloWriter, line 296: remove this sentence if you adopt the edit above.

src/collections/_documentation/platforms/python/troubleshooting.md
```diff
@@ -1,19 +1,58 @@
...
+title: Troubleshooting
 sidebar_order: 11
 ---
 
-The Python SDK generally does its best to figure out how contextual data such
-as tags set with `sentry_sdk.set_tags` is supposed to flow along your control
-flow. In most usecases this is achieved perfectly fine using [thread
-locals](https://docs.python.org/3/library/threading.html#thread-local-data),
-but there are quite a few situations where this approach fails.
+This section contains various gotchas that are inherent to the Python SDK's
```
PeloWriter, line 6: Perhaps " Troubleshooting helps those using the Python SDK with issues inherent to the design; these issues do not necessarily affect most users. Use this section to help answer these questions:"

src/collections/_documentation/platforms/python/troubleshooting.md
```diff
@@ -1,19 +1,58 @@
...
+design, do not necessarily affect the majority of users and do not really fit
+anywhere else.
 
-Read this document if you cannot figure out why context data is leaking across
-your HTTP requests, or when data is missing or popping up at the wrong place
-and time.
+The information here will help you answer these questions:
```
PeloWriter, line 10: delete if you adopt the edit above

src/collections/_documentation/platforms/python/index.md
```diff
@@ -291,8 +291,8 @@ The Python SDK provides some common [hints](/error-reporting/configuration/filte
 
 ## Troubleshooting and known issues
```
PeloWriter, line 292: Please use Title Case for Titles - so Troubleshooting and Known Issues"

src/collections/_documentation/platforms/python/troubleshooting.md
```diff
@@ -1,19 +1,58 @@
...
 
-## Python 2: thread locals and gevent
```
PeloWriter, line 16: Use Title case (though do not capitalize technical terms that aren't capitalized:) )

src/collections/_documentation/platforms/python/troubleshooting.md
```diff
@@ -1,19 +1,58 @@
...
+- "What do I do if scope data is leaking between requests?"
+- "What do I do if my transaction has nested spans when they should be parallel?"
+
+The short answer to those: check if your `contextvars` work and you clone the
+hub where needed.
+
+## Addressing concurrency issues
```
PeloWriter, line 18: Title case here

untitaker:
@rhcarvalho 

>I wanted to add a more complete example, similar to what is in getsentry/sentry-python#147 but with your suggested correction applied. WDYT?

Maybe we want to move this codesample into this new page: https://docs.sentry.io/platforms/python/default-integrations/#manual-propagation

src/collections/_documentation/platforms/python/index.md
rhcarvalho: That works for me too, except perhaps for the "common problems". At least for the current content the page is more about the "esoteric" side of problems, problems that only happen under uncommon circumstances / combinations of configurations.

rhcarvalho:
> Maybe we want to move this codesample into this new page: https://docs.sentry.io/platforms/python/default-integrations/#manual-propagation

Oh, good, I didn't know it existed. We can just link to https://docs.sentry.io/platforms/python/default-integrations/#threading for the more complete example. Thanks!

src/collections/_documentation/platforms/python/index.md
rhcarvalho: I made it short and direct:

```
The Troubleshooting page provides advice for solving known problems.
```

src/collections/_documentation/platforms/python/troubleshooting.md
rhcarvalho: I've rewritten to be short and direct:

```
We expect most users of the Python SDK not to run into any of the problems documented here.

Use the information in this page to help answer these questions:
```

rhcarvalho:
Updated all relevant titles to Title Case.

rhcarvalho:
@PeloWriter thank you very much for the feedback, I've incorporated your suggestions and tried to make sentences shorter and direct. Would you mind giving a second pass of eyes? Thanks!

PR #1798, opened 2020-06-26
Restructuring serverless files and updating setup instructions

ajjindal:
The PR includes python. and node docs changes for the Lambda and the Cloud Functions instructions. Also, added Serverless as a first class entity to the left nav.


src/collections/_documentation/platforms/python/generic.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Serverless
```
untitaker, line 2: I would keep this named serverless (I mean the filename). We have this naming pattern established with WSGI already, which is also a "generic" integration.

src/collections/_documentation/serverless/index.md
```diff
@@ -0,0 +1,26 @@
+---
+title: 'Serverless'
```
untitaker, line 2: We do not have a similar top-level page for other kinds of platforms. If we think serverless integrations deserve this kind of prominence I would say this should become a marketing page in the style of https://sentry.io/for/gaming/ rather than a documentation page.

src/collections/_documentation/platforms/python/gcp_functions.md
```diff
@@ -0,0 +1,54 @@
+---
+title: GCP Functions
```
untitaker, line 2: Please use WIZARD/ENDWIZARD tags like in other integrations and add it to `platforms.yml` and `platform_icons.yml`:

* https://github.com/getsentry/sentry-docs/blob/93777fd7c3a3d4d8ce4b84c21a32953bb58f51fa/src/_data/platform_icons.yml#L91-L94
* https://github.com/getsentry/sentry-docs/blob/93777fd7c3a3d4d8ce4b84c21a32953bb58f51fa/src/_data/platforms.yml#L220-L227

All of those measures will allow the user to search for serverless in the setup wizard within the Sentry UI and get the tagged excerpt from this page.

src/collections/_documentation/platforms/python/gcp_functions.md
```diff
@@ -0,0 +1,54 @@
...
+sidebar_order: 8
+---
+
+Install our Sentry SDK in the `requirement.txt` section:
```
untitaker, line 6:
```suggestion
Install our Sentry SDK in the `requirements.txt` section:
```

src/collections/_documentation/platforms/python/gcp_functions.md
```diff
@@ -0,0 +1,54 @@
...
+
+@serverless_function
+def my_function(...): ...
+```
+
+Use the generic integration by calling the `serverless_function` decorator. Decorators wrap a function and modify its behavior. Then, deploy and test the function. Checkout Sentry's [gcp sample apps](https://github.com/getsentry/examples/tree/master/gcp-cloud-functions/python) for detailed examples.
+
+{% capture __alert_content -%}
+
+If you are using another web framework inside of AWS Lambda, the framework may catch exceptions before they are sent to Sentry. To avoid this, enable the framework-specific integration as well, if one exists. See [*Integrations*](/platforms/python/#integrations) for more information
```
untitaker, line 26: I think this alert about using a different, more specific integration is not relevant in this context.

src/collections/_documentation/platforms/node/aws_lambda.md
```diff
@@ -0,0 +1,47 @@
+---
+title: AWS Lambda
+sidebar_order: 7
+---
+
+Create a deployment package on your local machine and install the required dependencies in the deployment package. For more information, see [Building an AWS Lambda deployment package for Node.js](https://aws.amazon.com/premiumsupport/knowledge-center/lambda-deployment-package-nodejs/).
+
+Install our Node.js SDK using `pip`:
```
silent1mezzo, line 8: Should be 'using npm'

src/collections/_documentation/platforms/node/aws_lambda.md
```diff
@@ -0,0 +1,47 @@
...
+
+module.exports.hello = sentryHandler(async event => {
+  throw new Error("asd");
+});
+```
+You can obtain the DSN using your Sentry account from your organization's *Settings -> Projects -> Client Keys (DSN)* in the Sentry web UI.
+
+Note: You need to call both `CaptureException` and `Flush` to report errors into Sentry.
+
+Create the deployment package in .zip format, then upload it to AWS Lambda as a Lambda function. Checkout Sentry’s [aws sample apps](http://tbd/) for detailed examples. Use  [JavaScript docs](/platforms/javascript/) for more configuration options.
```
PeloWriter, line 45: nit: Extra space prior the JS docs link

src/collections/_documentation/serverless/index.md
PeloWriter: I agree. The page content is a helpful overview (similar to what's been added for Performance), but isn't this another type of platform? So, this intro page should be nested alongside the other platforms. The files are maintained here:
src/collections/_documentation/platforms/<platform-name>/index.md

ajjindal: I do think that we should have a standalone serverless page. I can do what Fiona is suggesting but don't think we should get rid of the page. @untitaker thoughts?

untitaker: If we really want this type of content in the docs (which I really don't agree with -- Performance introduces an entirely new product to the user, with SDK APIs documented which are documented nowhere else) I would like to get rid of the mention of Additional Data there. First the example is incomplete, second it's duplicating existing docs.

src/collections/_documentation/platforms/python/generic.md
ajjindal: Suggestion taken

src/collections/_documentation/serverless/index.md
ajjindal: Suggestion taken.

src/collections/_documentation/platforms/python/gcp_functions.md
ajjindal: Taken

ajjindal:
I am going to go ahead and merge based on the PR approvals.

PR #1816, opened 2020-06-30
fix(apm): Polish dependency installation instructions for tracing getting started

dashed:


MimiDumpling (approved):
LGTM 🌻 Thanks!

AbhiPrasad (commented):
LGTM.

I think there are some cases of 
> To get started with performance monitoring **on** LIBRARY/LANGUAGE

that should be:
> To get started with performance monitoring **with/for** LIBRARY/LANGUAGE

not sure if with or for is better here, so I'll leave it do docs team to say.

MimiDumpling:
> LGTM.
> 
> I think there are some cases of
> 
> > To get started with performance monitoring **on** LIBRARY/LANGUAGE
> 
> that should be:
> 
> > To get started with performance monitoring **with/for** LIBRARY/LANGUAGE
> 
> not sure if with or for is better here, so I'll leave it do docs team to say.

I'd default to "with" because the tools are used together, right?

PR #1830, opened 2020-07-01
Various features to support Gatsby migration

dcramer:
- Dont index files with `gatsby: true` yaml header
- Add initial PlatformContent helper
- Cleanup abstraction of core components
- Remove /meta/ from jekyll index
- Don't load jekyll content into Gatsby (causes file path duplication issues)
- Add pre-commit
- Use default prettier configuration (no more singleQuote)


gatsby/gatsby-config.js
```diff
@@ -98,16 +99,30 @@ const getPlugins = () => {
...
+      }
+    },
     {
       resolve: `gatsby-source-filesystem`,
       options: {
         name: `pages`,
         path: `${__dirname}/src/pages`
       }
+    },
+    {
```
dcramer, line 120: this makes all yaml files available as a graphql collection. you can see the platform example in the PlatformContent component.

gatsby/plugins/gatsby-plugin-code-tabs/index.js
```diff
@@ -1,4 +1,4 @@
-const visit = require("unist-util-visit");
+const visit = require('unist-util-visit');
```
dcramer, line 1: all of this is just prettier, not sure how its diff here than in the other repo but i didnt care to figure it out

gatsby/src/components/platformContent.js
```diff
@@ -0,0 +1,102 @@
...
+
+        return (
+          <div className="platform-specific-content">
+            <div className="nav pb-1 flex">
+              <div className="dropdown mr-2 mb-1">
+                <button className="btn btn-sm btn-secondary dropdown-toggle">
+                  {defaultPlatform.name}
+                </button>
+
+                <div className="nav dropdown-menu" role="tablist">
```
dcramer, line 64: oh these dont actually load yet, as we dont have bootstrap support here, needs resolved

src/collections/_documentation/meta/soc2.md
```diff
@@ -1,8 +1,5 @@
 ---
 title: 'SOC2 Report'
 sidebar_order: 2
+gatsby: true
```
dcramer, line 4: this will likely become the proposal for migrating pages now

- add gatsby: true
- delete content (makes it obvious its not used0

PR #1831, opened 2020-07-01
Various fixes for code blocks

dcramer:
- Correct error handling on XHR requests
- Ensure defaults are always present
- Restore pre-existing behavior with contextual variables


gatsby/src/components/codeBlock.js
```diff
@@ -30,8 +31,8 @@ function makeKeywordsClickable(children) {
       }
       arr.push(
         Selector({
-          group: match[1],
-          keyword: match[2],
+          group: "PROJECT",
```
dcramer, line 34: I think I get what you were going for here @mitsuhiko but not sure we need it? I figure we can extend this functionality later and it keeps the core payload future proof.

PR #1838, opened 2020-07-02
Fix inclusive language for JS URLs

PeloWriter:


src/collections/_documentation/error-reporting/configuration/index.md
```diff
@@ -110,14 +110,14 @@ will attempt to auto-discover this value.
 {% endunsupported %}
 
 {:.config-key}
-### `allow-urls`
+### `deny-urls`
 
 {% supported browser browsernpm %}
 A list of strings or regex patterns that match error URLs which should not be sent to Sentry.  By default, all errors will be sent.
```
MimiDumpling, line 116:
```suggestion
A list of strings or regex patterns that match error URLs which should be sent to Sentry.  By default, all errors will be sent.
```

MimiDumpling: @LSSangha Is this the mistake you pointed out earlier?

PeloWriter: It is. This is the fix. Thanks, Mimi.

PR #1875, opened 2020-07-09
Remove: duplicated section in Transaction Summary

MimiDumpling:
Removed duplicated section in Transaction Summary and updated first two sentences in Getting Started.


src/collections/_documentation/performance-monitoring/performance/transaction-summary.md
```diff
@@ -41,7 +41,3 @@ _Note_: Currently, only transaction data - the transaction name and any attribu
 ## Related Issues
```
bowencai8, line 41: The content in these sections is different. The 2nd Related Issues paragraph with ("this table will show you...") is accurate.

the first one we just need to change the Title from "Related Issues" to "Sidebar"

PR #1880, opened 2020-07-10
feat: Add documentation about Apache Airflow

AbhiPrasad:
As promised, here is the Apache Airflow documentation.

The idea here is that we can get them started sending errors, and then point them to the actual Airflow docs, which will be more update to date about configuration, usage, tags/breadcrumbs tracked etc.

Although there is no Python code here, Airflow uses the python SDK internally to send errors, so I figured it would be best to put it under Python for now. Please let me know if there is a better way to do this, I'd be happy to oblige.

screenshot of new apache airflow docs

### DIFF
--- a/src/collections/_documentation/platforms/python/airflow.md
+++ b/src/collections/_documentation/platforms/python/airflow.md
@@ -1,0 +1,29 @@
+---
+title: Apache Airflow
+sidebar_order: 9
+---
+
+[Apache Airflow](https://airflow.apache.org/) **1.10.6** and above can be setup to send errors to Sentry.
+
+## Installation
+
+Install the `apache-airflow` package with the `sentry` requirement.
+
+```py
+pip install 'apache-airflow[sentry]'
+```
+
+You can then add your Sentry DSN to your configuration file (ex. `airflow.cfg`) under the `[sentry]` field.
+
+```ini
+[sentry]
+sentry_dsn = ___PUBLIC_DSN___
+```
+
+Now, Airflow should report errors to Sentry automatically. Airflow will also generate custom tags and breadcrumbs based on the current dag and tasks at time of error.
+
+Please see the official [Apache Airflow documentation](https://airflow.apache.org/docs/stable/errors.html) for more details.
+
+### Configuration
+
+Please see the official [Apache Airflow documentation](https://airflow.apache.org/docs/stable/configurations-ref.html#sentry) for the full list of configuration options available.
--- a/src/collections/_documentation/platforms/python/airflow.md
+++ b/src/collections/_documentation/platforms/python/airflow.md
@@ -20,7 +20,7 @@ Then, add your Sentry DSN to your configuration file (ex. `airflow.cfg`) under t
 sentry_dsn = ___PUBLIC_DSN___
 ```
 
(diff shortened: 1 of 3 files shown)


src/collections/_documentation/platforms/python/airflow.md
```diff
@@ -0,0 +1,29 @@
+---
+title: Apache Airflow
+sidebar_order: 9
+---
+
+[Apache Airflow](https://airflow.apache.org/) **1.10.6** and above can be setup to send errors to Sentry.
```
MimiDumpling, line 6:
```suggestion
[Apache Airflow](https://airflow.apache.org/) **1.10.6** and above can be set up to send errors to Sentry.
```

src/collections/_documentation/platforms/python/airflow.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Installation
+
+Install the `apache-airflow` package with the `sentry` requirement.
+
+```py
+pip install 'apache-airflow[sentry]'
+```
+
+You can then add your Sentry DSN to your configuration file (ex. `airflow.cfg`) under the `[sentry]` field.
```
MimiDumpling, line 16:
```suggestion
Then, add your Sentry DSN to your configuration file (ex. `airflow.cfg`) under the `[sentry]` field.
```

src/collections/_documentation/platforms/python/airflow.md
```diff
@@ -0,0 +1,29 @@
...
+
+```ini
+[sentry]
+sentry_dsn = ___PUBLIC_DSN___
+```
+
+Now, Airflow should report errors to Sentry automatically. Airflow will also generate custom tags and breadcrumbs based on the current dag and tasks at time of error.
```
MimiDumpling, line 23:
```suggestion
Now, Airflow should report errors to Sentry automatically. Airflow will also generate custom tags and breadcrumbs based on the current dag and tasks at time of error.
```
```suggestion
Now, Airflow should report errors to Sentry automatically. Airflow will also generate custom tags and breadcrumbs based on the current dag and tasks at the time of the error.
```

MimiDumpling: What's a dag? 😄

AbhiPrasad: A dag is a direct acyclic graph - https://en.wikipedia.org/wiki/Directed_acyclic_graph

> In Airflow, a DAG – or a Directed Acyclic Graph – is a collection of all the tasks you want to run, organized in a way that reflects their relationships and dependencies.
https://airflow.apache.org/docs/stable/concepts.html#dags

It seems in Airflow docs and code they choose to capitalize `DAG`, do you think that we should do the same here?

MimiDumpling: Yup. 😸 I updated the suggested edit. When using an acronym, write out the words explicitly the first time you mention it with the acronym in parentheses. After that initial definition, feel free to use just the acronym.

AbhiPrasad: Cool thank you, will keep that in mind.

PR #1883, opened 2020-07-13
Update: Release Health details

MimiDumpling:
Updated with definitions for healthy, crashed, errored, and abnormal. 

image


src/collections/_documentation/workflow/releases/health.md
```diff
@@ -41,6 +41,26 @@ The release details page focuses on an individual release. Elements of the relea
 
 [{% asset releases/release_details.png alt="View of the release details page showing an overview graph, files changed, and other details." %}]({% asset releases/release_details.png @path %})
 
+Hover over specific dates and times in the overview graph for a quick perspective of events that crashed, behaved abnormally, errored, or are healthy.
+
+Crashed 
+
+: The app had an explicit hard crash. You’ll likely see a corresponding Sentry issue for the crash.
```
PeloWriter, line 48: Please check with @marandaneto whether we should use the term "unhandled error" rather than crash in the description.

marandaneto: I'd go with unhandled error as well because of https://develop.sentry.dev/sdk/event-payloads/exception/#attributes-1

for the mobile world, the term hard crash is known, but in JS this might be confusing, as we use the `mechanism.handled` to tell either if it's a hard crash or not, I'd stick to the term handled/unhandled for keeping it semantic.

jan-auer: This depends on how we're planning to implement this in other SDKs. For example, will the Python SDK set the session to "crashed", if the web server continues to run, but there was an unhandled exception in the web request? I'd say, probably not.

This is an excellent point to think about. For now, I'd stick with the current description, however.

src/collections/_documentation/workflow/releases/health.md
```diff
@@ -41,6 +41,26 @@ The release details page focuses on an individual release. Elements of the relea
...
+
+: The app went Out of Memory (OOM) and was forced to quit by the OS/timed out/froze. There’s usually no corresponding Sentry issue because this is a passive action.
+
+Errored 
+
+: The app shut down normally, but there were non-fatal errors in the session. Most likely, there are issues/errors in Sentry.
+
+Healthy 
+
+: The app shuts down (or goes into the background) normally, and there were no errors in that session.
```
jan-auer, line 60: Errored and Healthy potentially are too specific as well. A session may end without the app shutting down. Especially in the new platforms outside of mobile, a session will not necessarily correspond to the application lifetime.

How about:

> The session ends normally and no errors occurred during its lifetime.

src/collections/_documentation/workflow/releases/health.md
```diff
@@ -41,6 +41,26 @@ The release details page focuses on an individual release. Elements of the relea
 
 [{% asset releases/release_details.png alt="View of the release details page showing an overview graph, files changed, and other details." %}]({% asset releases/release_details.png @path %})
 
+Hover over specific dates and times in the overview graph for a quick perspective of events that crashed, behaved abnormally, errored, or are healthy.
+
+Crashed 
+
+: The app had an explicit unhandled error or hard crash. You'll typically be able to see the corresponding issue in Sentry that captures this event.
```
MimiDumpling, line 48: @marandaneto @jan-auer How about this? Might be a good compromise for now?

marandaneto: sounds good to me

jan-auer: Sounds good, thanks!

PR #1887, opened 2020-07-14
fix: Remove {% raw %} calls in Gatsby mdx

dcramer:


dcramer:
Missed these in the conversion.

PR #1895, opened 2020-07-16
feat: Convert enriching-error-data to gatsby

dcramer:


gatsby/src/components/jsCdnTag.js
```diff
@@ -1,19 +1,34 @@
 import React, { useState } from "react";
 
+let cachedVersionData = null;
```
dcramer, line 3: realized it was doing a ton of requests - this stops that

gatsby/src/components/jsCdnTag.js
```diff
@@ -1,19 +1,32 @@
...
+
 export default ({ apm = false }) => {
-  const [versionData, setVersionData] = useState(null);
+  const [versionData, setVersionData] = useState(cachedVersionData);
 
-  fetch(
-    "https://release-registry.services.sentry.io/sdks/sentry.javascript.browser/latest"
-  ).then(data => setVersionData(data));
+  if (!versionData) {
```
dcramer, line 8: im confused as i thought i tested this but this clearly never worked

gatsby/src/includes/set-user/javascript.mdx
```diff
@@ -0,0 +1,5 @@
+```javascript
```
PeloWriter, line 1: if this renders fine, no change. Or should it be "```js"?

dcramer: i believe behind the scenes its just an alias, so you can use either

PR #1908, opened 2020-07-20
feat: Add docs on how to handle minified production builds for rn touch events

jennmueng:
Document how to have Touch Events work with minified component names in React Native production builds.

https://github.com/getsentry/sentry-react-native/issues/976


src/collections/_documentation/platforms/react-native/touchevents.md
```diff
@@ -114,3 +114,31 @@ The max number/depth of components to display when logging a touch's component t
...
+            mangle: {
+                keep_classnames: true, // Preserve class names
+                keep_fnames: true, // Preserve function names
+            }
+        }
+    }
+}
+```
+{% capture __alert_content -%}
+As this applies to **all** class and function names—not just React components—this could increase your js bundle size.
```
rhcarvalho, line 137: We use `—` in docs seldom, when we do we include spaces around it.

nit: js => JS

bruno-garcia, line 137:
```suggestion
As this applies to **all** class and function names—not just React components—this could increase your JS bundle size. 
```

PR #1913, opened 2020-07-21
ref: Bump and document attachment size limits

jan-auer:
Size limits for attachments are now at 100MB. Additionally, this PR adds more information on size limits to the Native SDK and Attachments page.

See https://github.com/getsentry/relay/pull/671
See https://github.com/getsentry/develop/pull/90

### DIFF
--- a/src/collections/_documentation/platforms/native/index.md
+++ b/src/collections/_documentation/platforms/native/index.md
@@ -178,6 +178,30 @@ sentry_value_t event = sentry_value_new_message_event(
 );
 ```
 
+### Application Crashes
+
+By default, the Native SDK intercepts crash signals and unhandled exceptions and
+sends crash reports to Sentry. Depending on the [backend](#integrations), the
+crash report is sent as a conventional event, or as a native crash report file.
+On Windows, Linux, and macOS, the default backends send binary Minidump files.
+Sentry processes them to extract stack traces and exception information into a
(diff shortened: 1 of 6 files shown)


src/collections/_documentation/platforms/native/index.md
```diff
@@ -178,6 +178,30 @@ sentry_value_t event = sentry_value_new_message_event(
...
+into a readable crash report.
+
+For more information on Minidumps and the limits that apply, see
+[What is a Minidump](/platforms/native/minidumps#what-is-a-minidump).
+
+{% capture __alert_content -%}
+The size of Minidumps can vary between a few kilobytes and many megabytes.
+Contributing factors are the number of threads, size of stack space, and the
+number of heap memory regions referenced from the stack. Sentry limits Minidump
+uploads to _100MB_ and drops all larger requests.
```
Swatinem, line 197: is it worth mentioning decompressed size here as well?

tonyo: 100MB is decompressed limit, and it probably makes sense to link to the compressed limit too somehow? (here and in `enriching-error-data/attachments/index.mdx`)

jan-auer: Ah, you're right. We should probably mention both limits. Let me update this.

PR #1918, opened 2020-07-22
Update: CLI for creating source bundles

MimiDumpling:
We need to update the CLI docs to improve this statement in creating source bundles:

Currently:
"Alternatively, specify the --include-sources parameter to the upload command, which generates source bundles on the fly during the upload. This requires that the upload is performed on the same machine as the application build."

It should say: 
"Alternatively, specify the --include-sources parameter to use the upload-dif command, which generates source bundles on the fly during the upload. This requires that the upload is performed on the same machine as the application build."

### DIFF
--- a/gatsby/src/docs/cli/dif.mdx
+++ b/gatsby/src/docs/cli/dif.mdx
@@ -59,13 +59,23 @@ To create a source bundle, use the `difutil bundle-sources` command on a list of
 debug information files:
 
 ```bash
+# on the build machine:
 $ sentry-cli difutil bundle-sources /path/to/files...
+
+# at any later time:
+$ sentry-cli upload-dif --type sourcebundle /path/to/bundles...
 ```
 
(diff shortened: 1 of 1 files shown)


jan-auer:
Might just be me, but this now sounds like you just need to specify `--include-sources` instead of passing it to `upload-dif`. What if we add a usage example to clarify this?

MimiDumpling:
> Might just be me, but this now sounds like you just need to specify `--include-sources` instead of passing it to `upload-dif`. What if we add a usage example to clarify this?

Sounds good. Will you provide a usage example? I'm happy to add it. Or, feel free to push directly to the branch. 😄 Thanks!

jan-auer:
@MimiDumpling Added, what do you think of this?

MimiDumpling:
> @MimiDumpling Added, what do you think of this?

LGTM. 🌻 Thanks, Jan!

PR #1929, opened 2020-07-25
Add GitHub Actions to release automation docs

jonesphillip:
Add GitHub Actions to release automation docs.
screencapture-localhost-9002-workflow-releases-release-automation-github-actions-2020-07-24-17_35_46


PeloWriter:
Love that we've got these started. Thanks! Could we possibly add the goodness that is available here - https://github.com/marketplace/actions/sentry-release so we have an authoritative set of content in the docs repo?

PR #1930, opened 2020-07-27
feat: change async default to false

scefali:
Goes with PR https://github.com/getsentry/sentry/pull/20078

The default value of `async` is now `true` so the docs need to be updated


scefali:
Closing for now

PR #1935, opened 2020-07-28
Add Gatsby-powered wizard components

dcramer:
- Automatically generate markdown wizard files in `gatsby/src/wizard` from Jekyll content
- Move JSON generation from Jekyll into Gatsby (achieved at build time via `gatsby build`)


dcramer:
I have verified _index.json, but will write a script to verify the child files yet.

You'll see that some of the generated wizard files are a bit wonky yet - they didnt convert the origin markdown as you'd prefer. We can fix those by hands before we move over. For now we'll assume 100% of those files are generated.

I will also look to see if its safe to remove these and only do them at build time for now.

dcramer:
need to scrub platform content

dcramer:
Decided we'll just generate these automatically from Jekyll for now. Once we move to drop jekyll we'll convert them over, remove .gitignore, and manually clean up the odd ones.

dcramer:
Still trying to figure out what the deal is w/ the failing test as I haven't been able to repro locally and the log I believe is misleading..

PR #1937, opened 2020-07-28
fix: Clarify how scopes work in Flask/Django

untitaker:
Fix https://github.com/getsentry/sentry-python/issues/769

### DIFF
--- a/src/_includes/platforms/python/request-data.md
+++ b/src/_includes/platforms/python/request-data.md
@@ -1,3 +1,3 @@
 Request data is attached to all events: **HTTP method, URL, headers, form data, JSON payloads**. Sentry excludes raw bodies and multipart file uploads. Sentry also excludes personally identifiable information (such as user ids, usernames, cookies, authorization headers, IP addresses) unless you set ``send_default_pii`` to ``True``.
 
-Scope data that is set from within a request handler/view or middleware will not be shared between requests. E.g. setting a tag during import time will set a global tag, while setting a tag within a view will attach the tag only to events sent as part of that request lifecycle.
+Each request has a separated scope. Setting a tag during import time will set a global tag, while setting a tag within a view will attach the tag only to events sent as part of that request lifecycle.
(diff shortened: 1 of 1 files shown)


src/_includes/platforms/python/request-data.md
```diff
@@ -1 +1,3 @@
 Request data is attached to all events: **HTTP method, URL, headers, form data, JSON payloads**. Sentry excludes raw bodies and multipart file uploads. Sentry also excludes personally identifiable information (such as user ids, usernames, cookies, authorization headers, IP addresses) unless you set ``send_default_pii`` to ``True``.
+
+Scope data that is set from within a request handler/view or middleware will not be shared between requests. E.g. setting a tag during import time will set a global tag, while setting a tag within a view will attach the tag only to events sent as part of that request lifecycle.
```
kam193, line 3: Maybe would be better to start like this: `Each request has a separated scope, this means...`

kam193:
It looks great, but are you sure it is true for all places when this is included? It's not only Flask/Django

untitaker:
@kam193 all of our integrations work like that if that's what you mean, and yeah this snippet is supposed to be included in all of our webframework integrations.

src/_includes/platforms/python/request-data.md
```diff
@@ -1 +1,3 @@
 Request data is attached to all events: **HTTP method, URL, headers, form data, JSON payloads**. Sentry excludes raw bodies and multipart file uploads. Sentry also excludes personally identifiable information (such as user ids, usernames, cookies, authorization headers, IP addresses) unless you set ``send_default_pii`` to ``True``.
+
+Each request has a separated scope. Setting a tag during import time will set a global tag, while setting a tag within a view will attach the tag only to events sent as part of that request lifecycle.
```
rhcarvalho, line 3:
```suggestion
Each request has a separate scope. Setting a tag during import time will set a global tag, while setting a tag within a view will attach the tag only to events sent as part of handling an individual request.
```

> Setting a tag during import time

This seems more confusing than clarifying to me. What does it mean to set a tag during import time?

Maybe a code snippet would be clearer?

In both cases it is going to be `scope.set_tag`, the difference being where the `scope` is coming from. On the script's top-level, that's going to be likely the scope attached to the global hub, and in the view code that is going to be the scope created specifically for the execution of the view.

PR #1944, opened 2020-07-29
Remove deprecated integrations from index

eurica:
I believe these integrations are gone (or in PD's case, not community/per-project)

Long term we should still restructure this whole section, but let's keep it true until then.


MimiDumpling:
Thanks for getting this started, @eurica. 😊 I'll double check to see if any other integrations need removing from this list.

FYI, I'm currently creating a restructuring plan for Integrations and working with @jonesphillip on the project.

MimiDumpling:
@jonesphillip will you double check the list of integrations we're removing? Pretty sure I got 'em all.

MimiDumpling:
@eurica do you want the glory of the merge? 🏆

eurica:
@MimiDumpling nope! send it out whenever works

PR #1945, opened 2020-07-29
fix(docs): Fix custom inbound filter plan

maheskett:
`custom-inbound-filters` and `discard-groups` are available on the Business plan


PeloWriter:
Thank you, Megan!

PR #1961, opened 2020-08-03
Restore API docs

dcramer:
Uses an api-docs collection in Gatsby


dcramer:
will merge on green

dcramer:
Aside none of the api generator tools were working. Brought them back just because we may need them in the future.

I pulled these in from jekyll and converted md -> json.

PR #1964, opened 2020-08-03
Update JS documentation

PeloWriter:
This review includes: 
- updating the Source Maps documentation (merging and editing the current doc that's kept in two places - sourcemaps.mdx and the index file for JS)
- moving the security policy documentation into the new repo
- Adding the missing info on `ignoreErrors` to the Filtering content


PeloWriter:
FYI - will fix the source maps redirect in a separate PR.

PR #1966, opened 2020-08-04
fix: Fix Sentry docs

dcramer:
Sentry queries our docs index, and updating it is non trivial. This adapts our docs schema to match Sentry's expected schema.


dcramer:
I already indexed docs and verified this fix in prod

PR #1968, opened 2020-08-04
ref: Compress pngs using pngquant to 85%

dcramer:


dcramer:
h/t @joshuarli

dcramer:
Yeah - I can't even tell the difference

PR #1970, opened 2020-08-06
fix: Patch up PHP docs to remove version pins and outdated/duplicate wizards

untitaker:
The monolog integration setup guide was for an older version and we currently don't have resources to produce content for the new one. Let's remove it for now such that people click on the regular PHP one.

cc @stayallive @ste93cry 

@dcramer did the version number get vendored when moving to gatsby, possibly?


dcramer:
Yeah it got rendered in.

Still need to figure out a global solution to solving these outside of React components.

PR #1972, opened 2020-08-06
Update sidebar to add product folders

PeloWriter:
This PR begins the shift of our content to organize by platform and by product. 

We'll have more work here to ensure that what needs to be configured is documented with its respective SDK. In addition, the platform content is part of the new SDK new format, so enriching error data and error monitoring content is (mostly) not yet migrated to the new structure.


nginx.conf
```diff
@@ -469,5 +469,69 @@ location = /guides/manage-event-stream/ {
...
+  
+  location = /enriching-error-data/environments/ {
+    return 302 /product/sentry-basics/environments/;
+  } 
+  
+  location = /enriching-error-data/error-tracing/ {
+    return 302 /product/sentry-basics/guides/;
+  }
+  
+  location = /guides/ {
```
dcramer, line 518: should we even bother moving this? if ultimately these should live somewhere else (e.g. imagine another section similar to 'api docs' or 'developer docs') then changing the url now just to potentially change it in the future wont be very useful.

nginx.conf
```diff
@@ -469,5 +469,69 @@ location = /guides/manage-event-stream/ {
...
+    return 302 /product/guides/;
+  }
+  
+  location = /platforms/javascript/sourcemaps {
```
dcramer, line 522: does this page actually cover the sourcemaps component well enough?

src/components/sidebar.js
```diff
@@ -159,20 +159,19 @@ const Sidebar = () => {
...
             <DynamicNav
-              root="performance-monitoring"
-              title="Performance Monitoring"
-              tree={tree}
-            />
-            <DynamicNav
-              root="workflow"
-              title="Workflow and Integrations"
+              root="product"
+              title="Product Guide"
```
dcramer, line 174: id probably just call it Product

nginx.conf
PeloWriter: We get a lot of asks from customers to provide best practices, and I've provided front matter to each of these guides for better context. 

I've moved it here to provide a primer to users as part of the concept of "basic info you should know". We can easily move it again if we think it needs a stronger location.

nginx.conf
PeloWriter: It should. I worked with Kamil on integrating the sourcemaps content that was published in the JavaScript index file with the content that was published (and slightly different) in the sourcemaps file, creating one source of truth. I'm open to input, if you think something is missing.

src/components/sidebar.js
PeloWriter: Will do. Thanks.

AndreKR:
This MR introduced [perf-homepage.png](https://github.com/getsentry/sentry-docs/blob/master/src/docs/product/performance/perf-homepage.png). I have a quick question if you don't mind... how did you manage to set project names with uppercase characters and spaces in them?

image

When I try to do the same, they're always mangled into lowercase with hyphens.

PR #1979, opened 2020-08-10
fix: Migrate PlatformContent to MDX

dcramer:
There is an issue with Gatsby and how it replaces content w/ SSR that uses the HTML API. This issue is mitigated in our use of PlatformContent by switching to MDX which uses their internal renderer.


dcramer:
Related conversation: https://github.com/gatsbyjs/gatsby/issues/12413

PR #1981, opened 2020-08-10
fix(jira): Link should be link, not markdown

manuzope:
From: https://docs.sentry.io/product/integrations/jira/

**Before**
Screen Shot 2020-08-10 at 3 52 55 PM

**After**
Screen Shot 2020-08-10 at 3 34 03 PM


dcramer:
:+1:

you can use markdown in these blocks but with syntax changes. see README for the giant "ugh" caveat

PR #1986, opened 2020-08-11
Add gatsby redirect support

dcramer:


gatsby-node.js
```diff
@@ -66,6 +66,9 @@ exports.createSchemaCustomization = ({ actions, schema }) => {
         draft: {
           type: "Boolean",
         },
+        redirect_from: {
```
untitaker, line 69: This looks suspicious but that's probably the missing piece and reason why I didn't get it to work

dcramer: this just makes it always exist in the graphql query - so it will never error on a missing field

untitaker: is this the way to do it in gatsby? I wrote a completely backwards query in develop that just never errors.

dcramer: i would say you dont _need_ this, because gatsby auto populates the graphql schema

the only time you need this is if a field isnt present. so for example, if no docs had "redirect_from" in them, and i tried to query on that field, the queries would error

PR #1992, opened 2020-08-12
Fix table formatting on /api/permissions

WFT:
This is my first PR here, so please let me know if I'm missing anything!

I believe that these rows are meant to be tables. They didn't get formatted as such because they were missing a header row. So I added a header row. Additionally, I made the methods under the **Projects** heading bold to match the rest of the documentation.

The [current state](https://docs.sentry.io/api/permissions/) looks like this: 
current: unformatted tables

This PR fixes the table formatting so that it looks like this:
new: formatted tables

### DIFF
--- a/src/docs/api/permissions.mdx
+++ b/src/docs/api/permissions.mdx
@@ -14,15 +14,19 @@ If you're looking for information on membership roles please visit the
 
 ### Organizations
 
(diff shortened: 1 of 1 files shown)


WFT:
Happy to help!

PR #1993, opened 2020-08-12
feat: Add SDK docs for React

AbhiPrasad:
This is WIP.

Open questions/thoughts:

1) How does react stay in "sync" with javascript? Like what happens if a new error section is added?

   Update: we still don't have a good solution, but, the way it is looking, the SDK specific folders should really only be touching the index and adding their own custom pages, so it might not be that big of a problem.

2) Links don't update, so you'll have to change the URLs (ex. from `/sdks/javascript/` -> `/sdks/react/`). This is a good use case for an eslint plugin, might write one if it's enough of a pain point.

   Actually we can probably lint against most concerns in docs. @PeloWriter @HazAT if y'all can collect a list of things we need to watch out for, we can write eslint rules that will lint against them. I also enjoy writing them, so I'll take any excuse 😄 

   ~~Edit: it's actually a bigger problem.~~

   ~~If we symlink all the files like we do right now, those files all link to `sdks/javascript`, while we want them to link to `sdks/react`. This means I end up copying over every file anyway, which kinda defeats the purpose of the symlink.~~

   ~~In addition, it makes 1) even more of a concern, the fact that stuff will get out of date.~~

   Update: we're solving this for now by using relative URLs.

3) It feels awkward to include framework specific stuff in the `index`. I tried to give a good balance, but would appreciate any feedback. Specifically here was how much Error Boundary stuff to have. Previously I had a whole example, and included the error boundary options right below the example. Now that there is so much more content, I had to move a lot of it into a different page.

4) ~~We should probably divide the docs into issue owners. I guess this is more of a meta level thing, but I think making sure the teams are at least pinged on PR is useful. I assume web platform would "own" the JS SDK for now.~~

   Update: see #2008.


src/docs/sdks/react/index.mdx
```diff
@@ -0,0 +1,223 @@
...
+
+- **[Review and Manage Integrations](/sdks/react/integrations/)**
+
+  Integrations extend the functionality of our JavaScript SDK to cover common libraries and environments automatically. Learn more about our default and pluggable integrations.
+
+- **[Troubleshooting](/sdks/react/troubleshooting/)**
+
+  If you need help solving issues with Sentry's JavaScript SDK, you can read the edge cases documented here.
+
+For a deep dive into our JavaScript SDK, [check out our GitHub repo](http://getsentry.github.io/sentry-javascript/).
```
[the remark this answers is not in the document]
rhcarvalho: This link is specific to the auto-generated API docs. Could consider pointing to the react specific docs (just noticed they are actually missing). Going to be similar to https://getsentry.github.io/sentry-javascript/modules/gatsby.html

src/docs/sdks/javascript/config/index.mdx
```diff
@@ -7,26 +7,26 @@ redirect_from:
 
 Sentry's JavaScript SDK includes many configuration options that are automatically set. You can configure your SDK using the options outlined in these pages:
 
-- **[Basic Options](/sdks/javascript/config/basics)**
+- **[Basic Options](basics)**
```
rhcarvalho, line 10: **1. relative links**
Relative links like this works fine. I noticed that in some places you preferred adding a `./` prefix. Commenting in case you want to do the same in headers, would make grepping relative links easier (?!).

**2. canonical form, `/` suffix**
https://docs.sentry.io/sdks/javascript/config/basics redirects to https://docs.sentry.io/sdks/javascript/config/basics/, so it would be good if our links used the canonical form to avoid a redirect.

src/docs/sdks/react/index.mdx
```diff
@@ -0,0 +1,179 @@
...
+
+You may also need to manage your configuration if you are using a third-party library to implement promises. Also, keep in mind that browsers may take security measures when serving script files from different origins that can block error reporting.
+
+Learn more about how to manually capture errors or enable message capture with Sentry's JavaScript SDK in [Capture Errors](./errors/capture).
+
+### Automatically Enrich Error Data
+
+Events sent by the JavaScript SDK to Sentry are enriched with data that helps identify the source of the event. Much of this data is sent automatically - including the error context and environment - as well as the trail of events that led up to the event, which we call breadcrumbs. You don't need to configure these, though you may modify them.
+
+Learn more about the data sent to with events in [Enrich Event Data](./errors//).
```
rhcarvalho, line 101:
```suggestion
Learn more about the data sent to with events in [Enrich Event Data](./errors/).
```

rhcarvalho: Also, "data sent to with events" -- something doesn't sound right there, typo? Maybe:

```suggestion
Learn more about the data sent with events in [Enrich Event Data](./errors/).
```

PR #2002, opened 2020-08-14
docs: Update Angular docs to use @sentry/angular package

kamilogorek:
This PR is dependent on https://github.com/getsentry/sentry-docs/pull/1993
The content is done and ready to review, but once PR above is merged, I'll rebase it on top of it, to make sure it's using the new system.

Note: `src/wizard/javascript/angular.md` and `src/docs/platforms/javascript/angular.mdx` are the same content, so no need to review them both.

TODO:
- [ ] Copy over the content of `angular.mdx` file to `angular.md` once it's approved
- [ ] Wait for #1993 to be merged
- [ ] Rebase on top of #1993 and verify everything works (mostly all links/navigation)
- [ ] Ship it.

### DIFF
--- a/src/docs/platforms/javascript/angular.mdx
+++ b/src/docs/platforms/javascript/angular.mdx
@@ -5,143 +5,241 @@ redirect_from:
   - /clients/javascript/integrations/angular2/
 ---
 
-On its own, `@sentry/browser` will report any uncaught exceptions triggered from your application. Additionally, you can configure `@sentry/browser` to catch any Angular-specific exceptions reported through the [@angular/core/ErrorHandler](https://angular.io/api/core/ErrorHandler) component. This is also a great opportunity to collect user feedback by using `Sentry.showReportDialog`.
+To use Sentry with your Angular application, you will need to use `@sentry/angular` (Sentry’s Browser Angular SDK).
 
-First, install `@sentry/browser`:
+<Alert level="info" title="Note"><markdown>
+
+`@sentry/angular` is a wrapper around the `@sentry/browser` package, with added functionality related to Angular. All methods available in the `@sentry/browser` package can also be imported from `@sentry/angular`.
+
+</markdown></Alert>
+
+Add the Sentry SDK as a dependency using yarn or npm:
 
 ```bash
 # Using yarn
-yarn add @sentry/browser
+$ yarn add @sentry/angular
 
 # Using npm
-npm install @sentry/browser
+$ npm install @sentry/angular
 ```
 
-Then initialize a new Sentry instance and configure Angular with the `ErrorHandler` provided and explained below:
+### Connecting the SDK to Sentry
 
-```typescript
-import { BrowserModule } from "@angular/platform-browser";
-import { NgModule, ErrorHandler, Injectable } from "@angular/core";
-import { HttpErrorResponse } from "@angular/common/http";
+After you've completed setting up a project in Sentry, Sentry will give you a value which we call a _DSN_ or _Data Source Name_. It looks a lot like a standard URL, but it’s just a representation of the configuration required by the Sentry SDKs. It consists of a few pieces, including the protocol, public key, the server address, and the project identifier.
 
-import { environment } from "../environments/environment";
-import { AppComponent } from "./app.component";
+You should `init` the Sentry browser SDK as soon as possible during your application load up, before initializing Angular:
 
-import * as Sentry from "@sentry/browser";
+```javascript
+import { enableProdMode } from '@angular/core';
+import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
+import { init } from '@sentry/angular';
 
-Sentry.init({
-  dsn: "___PUBLIC_DSN___",
-  // TryCatch has to be configured to disable XMLHttpRequest wrapping, as we are going to handle
-  // http module exceptions manually in Angular's ErrorHandler and we don't want it to capture the same error twice.
-  // Please note that TryCatch configuration requires at least @sentry/browser v5.16.0.
-  integrations: [
-    new Sentry.Integrations.TryCatch({
-      XMLHttpRequest: false,
-    }),
-  ],
+import { AppModule } from './app/app.module';
+
+init({
+  dsn: '___PUBLIC_DSN___',
+  // ...
 });
 
-@Injectable()
-export class SentryErrorHandler implements ErrorHandler {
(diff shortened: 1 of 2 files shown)


src/docs/platforms/javascript/angular.mdx
```diff
@@ -5,150 +5,253 @@ redirect_from:
...
+      new TracingIntegrations.BrowserTracing({
+        tracingOrigins: ['localhost', 'https://yourserver.io/api'],
+        routingInstrumentation: routingInstrumentation,
+      }),
+    ],
+    tracesSampleRate: 1,
+  });
+  ```
+
+2. Register `SentryTrace` as a provider in Angular's DI system, with a `Router` as its dependency:
```
PeloWriter, line 110: I don't see `SentryTrace` in the following code example, should this read "Register Sentry's Tracing integration as a provider..."?

src/docs/platforms/javascript/angular.mdx
```diff
@@ -174,7 +277,7 @@ Sentry.init({
 angular.module("yourApplicationModule", ["ngSentry"]);
 ```
 
-In case you're using the CDN version or the Loader, Sentry provides a standalone file for every integration:
+If you're using the CDN version or the Loader, Sentry provides a standalone file for every integration:
```
PeloWriter, line 280: Is the Loader referring to Lazy Loading Sentry? If so, is it worth a link to that content?

src/docs/platforms/javascript/angular.mdx
```diff
@@ -5,150 +5,253 @@ redirect_from:
...
 export class AppModule {}
 ```
 
-When using your own `ErrorHandler`, make sure that whenever you use `HttpInterceptor` alongside it,
-the interceptor doesn't modify the error captured originally.
-The same goes for writing your own API services with built-in `http` methods.
+Additionally, `createErrorHandler` accepts a set of options that allows you to configure its behavior. For more details see the `ErrorHandlerOptions` interface in [our repository](https://github.com/getsentry/sentry-javascript/blob/master/packages/angular/src/errorhandler.ts).
 
-For example, the service below makes it impossible for the SDK to extract the correct data, because the full, original error `e` is not propagated.
+## Integrating Tracing
```
PeloWriter, line 82: alternate title to this (which would be consistent) is "Monitor Performance", but I was uncertain if tracing is a subset of performance.

src/docs/platforms/javascript/angular.mdx
kamilogorek: Good catch, this is actually an incorrect statement. You cannot use the loader with old Angular.

kamilogorek:
> Are we missing a verification step?

It was there all along, left by Evan over a year ago, so I left it as is. Maybe it's already outdated? cc @EvanPurkhiser

src/docs/platforms/javascript/angular.mdx
kamilogorek: It should be `TraceService` instead. Fixed.

src/docs/platforms/javascript/angular.mdx
kamilogorek: Yes, yes it is. Updated the title then.

EvanPurkhiser:
The verification example TODO is for adding an example of how to trigger an error when setting up the sdk.

The TODO will explicitly create an alert in the onboarding wizard saying something like “looks like we haven’t added a verification example yet!” Or something like that.

Would be great to add an example.

src/wizard/javascript/angular.md
```diff
@@ -161,41 +257,44 @@ npm install @sentry/browser @sentry/integrations
...
+></script>
 
 <script>
   Sentry.init({
-    dsn: '___PUBLIC_DSN___',
-    integrations: [
-      new Sentry.Integrations.Angular(),
-    ],
+    dsn: "___PUBLIC_DSN___",
+    integrations: [new Sentry.Integrations.Angular()],
```
rhcarvalho, line 295: Shouldn't `Angular` be a default integration when using `@sentry/angular`?

kamilogorek: It's an old Angular 1.x integration, not `@sentry/angular`.

rhcarvalho: Aha! Got it.

src/wizard/javascript/angular.md
```diff
@@ -161,41 +257,44 @@ npm install @sentry/browser @sentry/integrations
...
 ```
 
 In case you're using the CDN version or the Loader, Sentry provides a standalone file for every integration:
 
 ```html
 <!-- Note that we now also provide a es6 build only -->
 <!-- <script src="https://browser.sentry-cdn.com/5.20.1/bundle.es6.min.js" integrity="sha384-vX2xdItiRzNmed/VJFb8J4h2p35hYqkdTI9+xNOueKEcr7iipZy17fplNS0ikHL0" crossorigin="anonymous"></script> -->
-<script src="https://browser.sentry-cdn.com/5.20.1/bundle.min.js" integrity="sha384-O8HdAJg1h8RARFowXd2J/r5fIWuinSBtjhwQoPesfVILeXzGpJxvyY/77OaPPXUo" crossorigin="anonymous"></script>
+<script
+  src="https://browser.sentry-cdn.com/5.20.1/bundle.min.js"
```
rhcarvalho, line 281: This and below should use the variables introduced in #2020

src/wizard/javascript/angular.md
```diff
@@ -4,148 +4,244 @@ doc_link: https://docs.sentry.io/platforms/javascript/angular/
...
 
-import { environment } from "../environments/environment";
-import { AppComponent } from "./app.component";
+You should `init` the Sentry browser SDK as soon as possible during your application load up, before initializing Angular:
 
-import * as Sentry from "@sentry/browser";
+```javascript
+import { enableProdMode } from '@angular/core';
+import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
+import { init } from '@sentry/angular';
```
rhcarvalho, line 35: Is there some smart idea behind departing from the `import * as Sentry from` form?

I thought `Sentry.init` was clearer than `init` in intent and eventually people copy pasting the code will also want to do something like `Sentry.captureMessage`, etc.

HazAT:
Moving to this PR instead: https://github.com/getsentry/sentry-docs/pull/2049

PR #2005, opened 2020-08-17
Fix: consistent naming of tracingOrigins option

birkirpalmason:
The `tracingOrigins` option is referenced as `traceOrigins` in the performance "getting started" documentation.


dcramer:
Thanks!

PR #2008, opened 2020-08-17
misc: Create CODEOWNERS

rhcarvalho:
Extracted from https://github.com/getsentry/sentry-docs/pull/1993#issue-466982710

> @AbhiPrasad 
> We should probably divide the docs into issue owners. I guess this is more of a meta level thing, but I think making sure the teams are at least pinged on PR is useful. I assume web platform would "own" the JS SDK for now.

---

For starters we can have the Web Team for all SDKs and override with more specific owners as needed.

This will cause the team to be pinged on PRs that touch files under the given path.


rhcarvalho:
> is it worth adding the src/docs/platforms/ path here as well?

Done.

PR #2011, opened 2020-08-17
feat: Add pre-commit (optional)

dcramer:


dcramer:
Aside i intentionally don't make it sound optional in our README as we'd prefer it, but I dont want to block basic commits

Makefile
```diff
@@ -1,6 +1,12 @@
-develop:
+develop: setup-git
 	[ ! -f .env.develpoment ] && cp .env.example .env.development
 	yarn
 
+setup-git:
+ifneq (, $(shell which pre-commit))
+	pre-commit install
+endif
+	git config branch.autosetuprebase always
```
dcramer, line 9: pulled this in since its standard in our repos

PR #2015, opened 2020-08-17
feat: Template variables in markdown

dcramer:
This primarily adds support for template variables (or otherwise javascript functions) executable within any markdown content. An example is found in the ``README.md``. Generally speaking, its the same as Django or Liquid: ``{{ scoped_variable }}``.

Using the system it implements the PackageRegistry and adds it to the scope as ``packages``, thus making them callable with e.g. ``packages.version()``.

Additionally this makes a few small changes to other various Gatsby configurations to improve performance.

### DIFF
--- a/plugins/gatsby-plugin-sentry-wizard/gatsby-node.js
+++ b/plugins/gatsby-plugin-sentry-wizard/gatsby-node.js
@@ -27,7 +27,7 @@ exports.createPages = async ({ graphql }, { source, output }) => {
       }
     `,
     {
-      source: source
+      source: source,
     }
   );
 
@@ -58,7 +58,7 @@ const writeJson = async (path, nodes) => {
       type: n.frontmatter.type,
       details: sub ? `${main}/${sub}.json` : `${main}.json`,
       doc_link: n.frontmatter.doc_link,
-      name: n.frontmatter.name
+      name: n.frontmatter.name,
     };
     if (sub) platforms[main][sub] = data;
     else platforms[main]["_self"] = data;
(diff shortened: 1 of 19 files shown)


dcramer:
I realize this is a big, but hopefully its easy to follow..

dcramer:
Aside there's probably a lot more places that need this tag updated, but i only did it in two as an example.

dcramer:
Updated to include ``page`` in the scope, which gives you access to frontmatter, and somewhat replicates what we had with Jekyll.

plugins/gatsby-remark-variables/index.js
```diff
@@ -0,0 +1,68 @@
...
+    markdownAST,
+    () => true,
+    async node => {
+      // XXX(dcramer): by far the worst template language of all time
+
+      if (!node.value) {
+        return;
+      }
+
+      // TODO(dcramer): this could be improved by parsing the string piece by piece so you can
```
dcramer, line 37: this is the main thing to note here.. likely won't be a big deal in practice, but would be ideal to support. problem for another day unless someones bored.

plugins/gatsby-remark-variables/index.js
```diff
@@ -0,0 +1,68 @@
+const visit = require("unist-util-visit");
+
+function scopedEval(expr, context = {}) {
```
dcramer, line 3: based on some research from stackoverflow - might still be possible to leak scope but we're not allowing user input so hopefully its not a significant concern

plugins/gatsby-plugin-sentry-wizard/gatsby-node.js
```diff
@@ -68,26 +68,28 @@ const writeJson = async (path, nodes) => {
   fs.mkdirSync(path, { recursive: true });
   fs.writeFileSync(`${path}/_index.json`, JSON.stringify({ platforms }));
 
-  nodes.forEach(node => {
-    const pathMatch = node.fields.slug.match(/^\/([^\/]+)(?:\/([^\/]+))?\/$/);
-    const [_, main, sub] = pathMatch;
+  Promise.all(
```
dcramer, line 71: only change here is moving this to concurrent ``Promise.all``

PR #2020, opened 2020-08-18
ref: Replace hardcoded versions of JavaScript package

dcramer:


src/docs/platforms/ember/index.mdx
```diff
@@ -36,16 +36,17 @@ like this:
 
 ```html
 <!-- Note that we now also provide a es6 build only -->
-<!-- <script src="https://browser.sentry-cdn.com/5.20.1/bundle.es6.min.js" integrity="sha384-vX2xdItiRzNmed/VJFb8J4h2p35hYqkdTI9+xNOueKEcr7iipZy17fplNS0ikHL0" crossorigin="anonymous"></script> -->
+<!-- <script src="https://browser.sentry-cdn.com/{{ packages.version('sentry.javascript.browser') }}/bundle.es6.min.js" integrity="sha384-{{ packages.checksum('sentry.javascript.browser', 'bundle.es6.min.js', 'sha384-base64') }}" crossorigin="anonymous"></script> -->
```
AbhiPrasad, line 39: We can use the `<JsCdnTag />` component for these right?

dcramer: Sort of - you cant embed a tag in a tag and its limited to .mdx files, so its arguably better for us to use this format going forward.

dcramer: for clarity, what i mean here is you can't embed ``<JsCdnTag />`` inside of the code block, which means its only useful for _only_ showing a single tag. It's also missing the syntax highlighting, which is more minor.

AbhiPrasad: Got it 👍 - I'll keep that in mind.

PR #2021, opened 2020-08-18
feat: Migrate from @sentry/apm to @sentry/tracing

AbhiPrasad:
Switch all references from `@sentry/apm` to `@sentry/tracing`.

Also add an upgrade guide detailing the changes.

### DIFF
--- a/src/components/jsCdnTag.js
+++ b/src/components/jsCdnTag.js
@@ -5,7 +5,7 @@ import CodeTabs from "./codeTabs";
 
 let cachedVersionData = null;
 
-export default ({ apm = false }) => {
+export default ({ tracing = false }) => {
   const [versionData, setVersionData] = useState(cachedVersionData);
 
   if (!versionData && typeof fetch !== "undefined") {
@@ -19,7 +19,7 @@ export default ({ apm = false }) => {
     });
   }
 
-  const packageName = apm ? "bundle.apm.min.js" : "bundle.min.js";
+  const packageName = tracing ? "bundle.tracing.min.js" : "bundle.min.js";
   const packageData = versionData
     ? versionData
     : {
--- a/src/docs/product/performance/apm-to-tracing.mdx
+++ b/src/docs/product/performance/apm-to-tracing.mdx
@@ -16,7 +16,7 @@ For most cases, just replacing your `@sentry/tracing` import is enough to switch
 +import "@sentry/tracing";
 ```
 
-If you were using the CDN use the tracing snippet instead:
+If you were using the Browser CDN bundle, switch to using the tracing snippet instead:
 
 ```html
 <script
@@ -25,6 +25,24 @@ If you were using the CDN use the tracing snippet instead:
   crossorigin="anonymous"
 ></script>
 ```
+
+### Node
+
+If you were using the Express integration for automatic instrumentation, be sure to update your imports.
+
+```diff
+import * as Sentry from "@sentry/node";
+-import { Integrations } from "@sentry/apm";
++import { Integrations } from "@sentry/tracing";
+
+ Sentry.init({
+   integrations: [
+     new Integrations.Express(),
+   ]
+ });
+```
+
+### Browser
 
 If you are using the automatic instrumentation for the _browser_, you should switch your imports from `Tracing` to `BrowserTracing`.
 
--- a/src/docs/product/performance/apm-to-tracing.mdx
+++ b/src/docs/product/performance/apm-to-tracing.mdx
@@ -18,7 +18,13 @@ For most cases, just replacing your `@sentry/tracing` import is enough to switch
 
 If you were using the CDN use the tracing snippet instead:
 
-<JsCdnTag tracing />
+```html
+<script
+  src="https://browser.sentry-cdn.com/{{ packages.version('sentry.javascript.browser') }}/bundle.tracing.min.js"
+  integrity="sha384-{{ packages.checksum('sentry.javascript.browser', 'bundle.tracing.min.js', 'sha384-base64') }}"
+  crossorigin="anonymous"
+></script>
+```
 
 If you are using the automatic instrumentation for the _browser_, you should switch your imports from `Tracing` to `BrowserTracing`.
 
--- a/src/docs/product/performance/apm-to-tracing.mdx
+++ b/src/docs/product/performance/apm-to-tracing.mdx
@@ -1,0 +1,101 @@
+---
+title: "Switching to @sentry/tracing"
+sidebar_order: 20
+---
+
+Our performance JavaScript SDK has been renamed from [`@sentry/apm`](https://www.npmjs.com/package/@sentry/apm) to [`@sentry/tracing`](https://www.npmjs.com/package/@sentry/tracing). The behaviour between `@sentry/apm` and `@sentry/tracing` is _identical_, but some imports and API have changed slightly.
+
+Although `@sentry/apm` is deprecated for `@sentry/tracing`, we will continue to support `@sentry/apm` with bug fixes for all `5.x.x` versions.
+
+## Migrating from @sentry/apm to @sentry/tracing
+
+For most cases, just replacing your `@sentry/tracing` import is enough to switch packages.
+
+```diff
+-import "@sentry/apm";
++import "@sentry/tracing";
+```
+
+If you were using the CDN use the tracing snippet instead:
+
+<JsCdnTag tracing />
+
+If you are using the automatic instrumentation for the _browser_, you should switch your imports from `Tracing` to `BrowserTracing`.
+
+```diff
+import * as Sentry from "@sentry/browser";
+-import { Integrations } from "@sentry/apm";
++import { Integrations } from "@sentry/tracing";
+
+ Sentry.init({
+   integrations: [
+-    new Integrations.Tracing(),
++    new Integrations.BrowserTracing(),
+   ]
+ });
+```
+
+If you were using the `beforeNavigate` option from the `Tracing` integration, the API for the `beforeNavigate` option in `BrowserTracing` has changed slightly. Instead of passing in a location and returning a string representing transaction name, `beforeNavigate` now accepts a transaction context and is expected to return transaction context. You can now add extra tags or change the op based on different parameters. If you want to access the location like before, you will have to manually look at `window.location`.
+
+For example, if you had a function like so that computed the a custom transaction name,
+
+```javascript
+import * as Sentry from "@sentry/browser";
+import { Integrations } from "@sentry/apm";
+
+Sentry.init({
+  integrations: [
+    new Integrations.Tracing({
+      beforeNavigate: location => {
+        return getTransactionName(location);
+      },
+    }),
+  ],
+});
+```
+
+You would know leverage the context to do the same thing.
+
+```javascript
+import * as Sentry from "@sentry/browser";
+import { Integrations } from "@sentry/tracing";
+
+Sentry.init({
+  integrations: [
+    new Integrations.BrowserTracing({
+      beforeNavigate: context => {
+        return {
+          ...context,
+          // Can even look at context tags or other data to adjust
+          // transaction name
+          name: getTransactionName(window.location),
+        };
+      },
+    }),
+  ],
+});
+```
+
+For the full diff:
+
+```diff
+import * as Sentry from "@sentry/browser";
+-import { Integrations } from "@sentry/apm";
++import { Integrations } from "@sentry/tracing";
+
+ Sentry.init({
+   integrations: [
+-    new Integrations.Tracing({
+-      beforeNavigate: (location) => {
+-        return getTransactionName(location)
++    new Integrations.BrowserTracing({
++      beforeNavigate: (ctx) => {
++        return {
++          ...ctx,
++          name: getTransactionName(ctx.name, window.location)
++        }
+       }
+     }),
+   ]
+});
+```
--- a/src/docs/product/performance/apm-to-tracing.mdx
+++ b/src/docs/product/performance/apm-to-tracing.mdx
@@ -26,6 +26,15 @@ If you were using the Browser CDN bundle, switch from the old
 ></script>
 ```
 
+And then update `Sentry.init`:
+
+```diff
+ Sentry.init({
+-  integrations: [new Sentry.Integrations.Tracing()]
++  integrations: [new Sentry.Integrations.BrowserTracing()]
+ });
+```
+
 ### Browser (npm package)
 
 If you were using automatic instrumentation, update the import statement and
--- a/src/docs/product/performance/apm-to-tracing.mdx
+++ b/src/docs/product/performance/apm-to-tracing.mdx
@@ -3,20 +3,20 @@ title: "Switching to @sentry/tracing"
 sidebar_order: 20
 ---
 
-Our performance JavaScript SDK has been renamed from [`@sentry/apm`](https://www.npmjs.com/package/@sentry/apm) to [`@sentry/tracing`](https://www.npmjs.com/package/@sentry/tracing). The behaviour between `@sentry/apm` and `@sentry/tracing` is _identical_, but some imports and API have changed slightly.
+The tracing integration for JavaScript SDKs has moved from
+[`@sentry/apm`](https://www.npmjs.com/package/@sentry/apm) to
+[`@sentry/tracing`](https://www.npmjs.com/package/@sentry/tracing). While the
+two packages are similar, some imports and APIs have changed slightly.
 
-Although `@sentry/apm` is deprecated for `@sentry/tracing`, we will continue to support `@sentry/apm` with bug fixes for all `5.x.x` versions.
+The old package `@sentry/apm` is deprecated in favor of `@sentry/tracing`.
+Future support for `@sentry/apm` is limited to bug fixes only.
 
 ## Migrating from @sentry/apm to @sentry/tracing
 
-For most cases, just replacing your `@sentry/tracing` import is enough to switch packages.
+### Browser (CDN bundle)
 
-```diff
--import "@sentry/apm";
-+import "@sentry/tracing";
-```
-
-If you were using the Browser CDN bundle, switch to using the tracing snippet instead:
+If you were using the Browser CDN bundle, switch from the old
+`bundle.apm.min.js` to the new tracing bundle:
 
 ```html
 <script
@@ -26,28 +26,13 @@ If you were using the Browser CDN bundle, switch to using the tracing snippet in
 ></script>
 ```
 
-### Node
+### Browser (npm package)
 
-If you were using the Express integration for automatic instrumentation, be sure to update your imports.
-
-```diff
-import * as Sentry from "@sentry/node";
--import { Integrations } from "@sentry/apm";
-+import { Integrations } from "@sentry/tracing";
-
- Sentry.init({
-   integrations: [
-     new Integrations.Express(),
-   ]
- });
-```
-
-### Browser
-
-If you are using the automatic instrumentation for the _browser_, you should switch your imports from `Tracing` to `BrowserTracing`.
+If you were using automatic instrumentation, update the import statement and
+update `Sentry.init` to use the new `BrowserTracing` integration:
 
 ```diff
-import * as Sentry from "@sentry/browser";
+ import * as Sentry from "@sentry/browser";
 -import { Integrations } from "@sentry/apm";
 +import { Integrations } from "@sentry/tracing";
 
@@ -59,9 +44,16 @@ import * as Sentry from "@sentry/browser";
  });
 ```
 
-If you were using the `beforeNavigate` option from the `Tracing` integration, the API for the `beforeNavigate` option in `BrowserTracing` has changed slightly. Instead of passing in a location and returning a string representing transaction name, `beforeNavigate` now accepts a transaction context and is expected to return transaction context. You can now add extra tags or change the op based on different parameters. If you want to access the location like before, you will have to manually look at `window.location`.
+If you were using the `beforeNavigate` option from the `Tracing` integration,
+the API in `BrowserTracing` has changed slightly. Instead of passing in a
+location and returning a string representing transaction name, `beforeNavigate`
+now accepts a transaction context and is expected to return a transaction
+context. You can now add extra tags or change the `op` based on different
+parameters. If you want to access the location like before, you can get it from
+`window.location`.
 
-For example, if you had a function like so that computed the a custom transaction name,
+For example, if you had a function like so that computed a custom transaction
+name:
 
 ```javascript
 import * as Sentry from "@sentry/browser";
@@ -78,7 +70,7 @@ Sentry.init({
 });
 ```
 
-You would now leverage the context to do the same thing.
+You would now leverage the context to do the same thing:
 
 ```javascript
 import * as Sentry from "@sentry/browser";
@@ -103,7 +95,7 @@ Sentry.init({
 For the full diff:
 
 ```diff
-import * as Sentry from "@sentry/browser";
+ import * as Sentry from "@sentry/browser";
 -import { Integrations } from "@sentry/apm";
 +import { Integrations } from "@sentry/tracing";
 
@@ -121,5 +113,22 @@ import * as Sentry from "@sentry/browser";
        }
      }),
    ]
-});
+ });
+```
+
+### Node
+
+If you were using the Express integration for automatic instrumentation, the
+only necessary change is to update the import statement:
+
+```diff
+ import * as Sentry from "@sentry/node";
+-import { Integrations } from "@sentry/apm";
++import { Integrations } from "@sentry/tracing";
+
+ Sentry.init({
+   integrations: [
+     new Integrations.Express(),
+   ]
+ });
 ```
--- a/src/docs/product/performance/apm-to-tracing.mdx
+++ b/src/docs/product/performance/apm-to-tracing.mdx
@@ -54,7 +54,7 @@ Sentry.init({
 });
 ```
 
-You would know leverage the context to do the same thing.
+You would now leverage the context to do the same thing.
 
 ```javascript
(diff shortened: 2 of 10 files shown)


src/includes/performance-monitoring/configuration/javascript.mdx
```diff
@@ -130,23 +131,33 @@ Sentry.init({
...
 span.finish(); // Remember that only finished spans will be sent with the transaction
 transaction.finish(); // Finishing the transaction will send it to Sentry
 ```
 
+If you want the transaction to be associated with any global errors, or have integrations automatically add spans to your transaction, the transaction should be set on the scope.
+
+```javascript
+const transaction = Sentry.startTransaction({ name: "test-transaction" });
+
+Sentry.configureScope(scope => scope.setSpan(transaction));
```
AbhiPrasad, line 149: This section is new - added as I figured this would be a common concern to bring up.

src/docs/product/performance/apm-to-tracing.mdx
```diff
@@ -0,0 +1,101 @@
...
+sidebar_order: 20
+---
+
+Our performance JavaScript SDK has been renamed from [`@sentry/apm`](https://www.npmjs.com/package/@sentry/apm) to [`@sentry/tracing`](https://www.npmjs.com/package/@sentry/tracing). The behaviour between `@sentry/apm` and `@sentry/tracing` is _identical_, but some imports and API have changed slightly.
+
+Although `@sentry/apm` is deprecated for `@sentry/tracing`, we will continue to support `@sentry/apm` with bug fixes for all `5.x.x` versions.
+
+## Migrating from @sentry/apm to @sentry/tracing
+
+For most cases, just replacing your `@sentry/tracing` import is enough to switch packages.
```
lobsterkatie, line 12:
```suggestion
For most cases, just replacing your `@sentry/apm` import is enough to switch packages.
```

src/docs/product/performance/apm-to-tracing.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+ Sentry.init({
+   integrations: [
+-    new Integrations.Tracing(),
++    new Integrations.BrowserTracing(),
+   ]
+ });
+```
+
+If you were using the `beforeNavigate` option from the `Tracing` integration, the API for the `beforeNavigate` option in `BrowserTracing` has changed slightly. Instead of passing in a location and returning a string representing transaction name, `beforeNavigate` now accepts a transaction context and is expected to return transaction context. You can now add extra tags or change the op based on different parameters. If you want to access the location like before, you will have to manually look at `window.location`.
```
lobsterkatie, line 38: This will need to change once we sunset `beforeNavigate` in favor of `tracesSampler`. Do we want to hold off publishing this until that happens?

src/includes/performance-monitoring/configuration/javascript.mdx
```diff
@@ -1,41 +1,42 @@
...
+
+</markdown></Alert>
+
 Next, initialize the integration in your call to `Sentry.init`:
 
 ```jsx
-import * as Sentry from '@sentry/browser';
-import { Integrations as ApmIntegrations } from '@sentry/apm';
```
lobsterkatie, line 21: Did you change it to just `Integrations` for simplicity? It had been `ApmIntegrations` in order that people be able to import the integrations package as `Integrations` also. You might consider `TracingIntegrations`.

src/includes/performance-monitoring/configuration/javascript.mdx
```diff
@@ -199,34 +210,37 @@ When Sentry captures transactions, they are assigned a transaction name. This na
...
   if (event.type === "transaction") {
-	event.transaction = sanitizeTransactionName(event.transaction);
+    event.transaction = sanitizeTransactionName(event.transaction);
   }
   return event;
 });
 ```
 
-For browser JavaScript applications using the `Tracing` integration, the `beforeNavigate` option  can be used to better group navigation/pageload transactions together based on URL.
+For browser JavaScript applications using the `BrowserTracing` integration, the `beforeNavigate` option can be used to better group navigation/pageload transactions together based on URL.
```
lobsterkatie, line 224: @HazAT - this is a use of `beforeNavigate` which isn't covered by `tracesSampler`. We may need to rethink the idea of deprecating the former in favor of the latter.

lobsterkatie, line 224: Could we roll this example into the nearly identical one above, by just mentioning grouping there? Are there any other differences between that example and this one?

src/includes/performance-monitoring/configuration/javascript.mdx
dcramer: could we just improve the exports here?

```
import { Tracing } from "..."
```

What else would this even expose? Why do we need it to be contained in Integrations?

src/includes/performance-monitoring/configuration/javascript.mdx
```diff
@@ -235,15 +249,17 @@ Sentry.init({
 
 **Retrieving a Transaction**
 
-In cases where you want to attach Spans to an already ongoing Transaction you can use `Sentry.getCurrentHub().getScope().getTransaction()`. This function will return a `Transaction` in case there is a running Transaction otherwise it returns `undefined`. If you are using our Tracing integration by default we attach the Transaction to the Scope. So you could do something like this:
+In cases where you want to attach Spans to an already ongoing Transaction you can use `Sentry.getCurrentHub().getScope().getTransaction()`. This function will return a `Transaction` in case there is a running Transaction on the scope, otherwise it returns `undefined`. If you are using our BrowserTracing integration by default we attach the Transaction to the Scope. So you could do something like this:
```
lobsterkatie, line 252:
```suggestion
In cases where you want to attach Spans to an already ongoing Transaction you can use `Sentry.getCurrentHub().getScope().getTransaction()`. This function will return a `Transaction` in case there is a running Transaction on the scope, otherwise it returns `undefined`. If you are using our BrowserTracing integration, by default we attach the Transaction to the Scope. So you could do something like this:
```

Also - do we attach all transactions to the scope automatically when using the `BrowserTracing` integration, or just the `pageload` and `navigation` ones?

src/includes/performance-monitoring/configuration/koa.mdx
```diff
@@ -85,8 +91,8 @@ app.use(tracingMiddleWare);
 
 // usual error handler
 app.on("error", (err, ctx) => {
-  Sentry.withScope(function (scope) {
-    scope.addEventProcessor(function (event) {
+  Sentry.withScope(function(scope) {
```
lobsterkatie, line 94: Is there a reason we're doing this with anonymous functions whereas we did the one above with arrow functions?

src/includes/performance-monitoring/configuration/react.mdx
```diff
@@ -35,7 +41,7 @@ ReactDOM.render(<App />, rootNode);
 
 **`withProfiler` Higher-Order Component**
 
-`@sentry/react` exports a `withProfiler` higher-order component that leverages the `@sentry/apm` Tracing integration to add React related spans to transactions. If the Tracing integration is not enabled, the Profiler will not work.
+`@sentry/react` exports a `withProfiler` higher-order component that leverages the `@sentry/tracing` add React related spans to transactions.
```
lobsterkatie, line 44:
```suggestion
`@sentry/react` exports a `withProfiler` higher-order component that leverages the `@sentry/tracing` to add React-related spans to transactions.
```

src/includes/performance-monitoring/configuration/vue.mdx
```diff
@@ -1,49 +1,54 @@
...
+
+</markdown></Alert>
+
 The Vue Tracing Integration allows you to track rendering performance during an initial application load.
 
 Sentry injects a handler inside Vue's `beforeCreate` mixin, providing access to a Vue component during its life cycle stages.
-When Sentry encounters a component named `root`, which is a top-level Vue instance (as in `new Vue({})`), we use our AM Tracing integration,
-and create a new activity named `Vue Application Render`. Once the activity has been created, it will wait until all of its child components render, and there aren't new rendering events triggered within the configured `timeout`, before marking the activity as completed.
+When Sentry encounters a component named `root`, which is a top-level Vue instance (as in `new Vue({})`), we use our AM BrowserTracing integration,
+and create a new span named `Vue Application Render`. Once the span has been created, it will wait until all of its child components render, and there aren't new rendering events triggered within the configured `timeout`, before marking the span as finished.
```
lobsterkatie, line 23: Spans don't have timeouts, do they? Is this actually a transaction?

src/includes/performance-monitoring/configuration/vue.mdx
```diff
@@ -1,49 +1,54 @@
...
+Once you've configured the BrowserTracing integration, move on to configuring the Vue integration itself.
 Sentry built the new tracing capabilities into the original Vue error handler integrations, so there is no need to add any new packages. You only need to provide an appropriate configuration.
 
 The most basic configuration for tracing your Vue app, which would track only the top-level component, looks like this:
 
 ```js
 import { Vue as VueIntegration } from "@sentry/integrations";
-import { Integrations } from "@sentry/apm";
```
lobsterkatie, line 39: If we're importing the entire tracing package here so that the `Sentry.init()` can look the same as it does everywhere else, perhaps we should do the same for the Vue integration? It might not be a bad idea to have at least one example where we're gracefully handling the `Integrations` naming conflict.

(Alternatively, we could handle every instance of this the other way, and only ever import the `BrowserTracing` integration instead of the whole package, which would make the question of naming conflicts moot.)

src/includes/performance-monitoring/configuration/vue.mdx
```diff
@@ -1,49 +1,54 @@
...
+and create a new span named `Vue Application Render`. Once the span has been created, it will wait until all of its child components render, and there aren't new rendering events triggered within the configured `timeout`, before marking the span as finished.
 
 The described instrumentation functionality will give you very high-level information about the rendering performance of the Vue instance. However, the integration can also provide more fine-grained details about what actually happened during a specific activity.
 To do that, you need to specify which components to track and what hooks to listen to (you can find a list of all available hooks [here](https://vuejs.org/v2/api/#Options-Lifecycle-Hooks)). You can also turn on tracking for all the components. However, it may be rather noisy if your app consists of hundreds of components. We encourage being more specific. If you don't provide hooks, Sentry will track a component's `mount` and `update` hooks.
 
 Note that we don't use `before` and `-ed` pairs for hooks, and you should provide a simple verb instead. For example, `update` is correct. `beforeUpdate` and `updated` are incorrect.
```
lobsterkatie, line 28:
```suggestion
Note that when specifying hooks we use the simple verb rather than `before` and `-ed` pairs. For example, `destroy` is correct. `beforeDestroy` and `destroyed` are incorrect.
```

NB that I'm only pattern matching - I don't know for sure if `destroy` has a `before` and an `-ed`. I just thought that if possible it might be nice to have one that isn't the default since they already have examples of us correctly formatting the default ones.

src/includes/performance-monitoring/configuration/vue.mdx
```diff
@@ -1,49 +1,54 @@
...
 
-To set up the Vue Tracing Integration, you will first need to configure the AM Tracing integration itself. For details on how to do this, see the [JavaScript](/performance-monitoring/distributed-tracing/#javascript) section above.
-Once you've configured the Tracing integration, move on to configuring the Vue integration itself.
+To set up the Vue Tracing Integration, you will first need to configure the AM BrowserTracing integration itself. For details on how to do this, see the [JavaScript](/performance-monitoring/distributed-tracing/#javascript) section above.
+Once you've configured the BrowserTracing integration, move on to configuring the Vue integration itself.
 Sentry built the new tracing capabilities into the original Vue error handler integrations, so there is no need to add any new packages. You only need to provide an appropriate configuration.
 
 The most basic configuration for tracing your Vue app, which would track only the top-level component, looks like this:
```
lobsterkatie, line 34:
```suggestion
The most basic configuration for tracing your Vue app, which would track only the `mount` and `update` hooks for the top-level component, looks like this:
```

src/includes/performance-monitoring/configuration/vue.mdx
```diff
@@ -78,10 +89,16 @@ new VueIntegration({
...
-  }
-})
+    trackComponents: [
+      "App",
+      "RwvHeader",
+      "RwvFooter",
+      "RwvArticleList",
+      "Pagination",
+    ],
+    hooks: ["mount", "update", "destroy"],
```
lobsterkatie, line 99: IDK if a comment in the code here is the right place - perhaps it belongs in the explanatory text above - but it might be nice to mention that specifying the hooks yourself _replaces_ rather than adds to the default list, so `'mount'` and `'update'` need to be in there if they're still going to be tracked.

src/includes/performance-monitoring/configuration/javascript.mdx
lobsterkatie: > could we just improve the exports here?

I eventually came to the same idea as I read on, so I'm upgrading my "we could" below to a "I think we should." (Provided it doesn't break anything, which I don't believe it would.)

src/docs/product/performance/apm-to-tracing.mdx
AbhiPrasad: This is a good point. I will hold off until we make a decision in grooming tomorrow.

src/includes/performance-monitoring/configuration/javascript.mdx
AbhiPrasad: `import { Integrations } from "@sentry/tracing";`

`Integrations` is an object containing `{ Express, BrowserTracing }`

We could separate this yes, I just chose to make the behaviour consistent with `@sentry/apm`. It wouldn't break anything to change the imports, we would just have to update the upgrade process.

src/includes/performance-monitoring/configuration/javascript.mdx
AbhiPrasad: There are no differences, we can combine the examples together yes.

src/includes/performance-monitoring/configuration/javascript.mdx
AbhiPrasad: We only attach `pageload` and `navigation` transaction to the scope. I can mention that.

src/includes/performance-monitoring/configuration/koa.mdx
AbhiPrasad: No reason. I think this documentation was community contributed, but I can switch to using arrow functions.

src/includes/performance-monitoring/configuration/vue.mdx
AbhiPrasad: Ah this is based on `idleTransaction`. Maybe it is better just to remove this paragraph.

src/includes/performance-monitoring/configuration/vue.mdx
AbhiPrasad: @kamilogorek will know.

src/includes/performance-monitoring/configuration/vue.mdx
AbhiPrasad: > only ever import the BrowserTracing integration instead of the whole package

We can make this change.

src/includes/performance-monitoring/configuration/vue.mdx
AbhiPrasad: This is a good point, but something we can address in a future PR.

src/includes/performance-monitoring/configuration/vue.mdx
kamilogorek: They all do, except `activate`. That's why we use one simple verbs instead. Here's the full mapping list - https://github.com/getsentry/sentry-javascript/blob/5ed313546dd687b637dfb20a3a28a7823278d21a/packages/integrations/src/vue.ts#L116-L123

src/docs/product/performance/apm-to-tracing.mdx
AbhiPrasad: Update after grooming: have decided to document `beforeNavigate`, but not to mention filtering.

src/components/jsCdnTag.js
```diff
@@ -19,7 +19,7 @@ export default ({ apm = false }) => {
     });
   }
 
-  const packageName = apm ? "bundle.apm.min.js" : "bundle.min.js";
+  const packageName = tracing ? "bundle.tracing.min.js" : "bundle.min.js";
```
rhcarvalho, line 22: I think we need to update `src/platforms/javascript/index.mdx` and `src/wizard/javascript/index.md`, they reference `bundle.apm.min.js`.

rhcarvalho: Also `src/wizard/javascript/react.md`.

rhcarvalho (changes requested):
Marking here that some references are still pointing to `@sentry/apm`.

```
$ git grep --line-number apm
src/platforms/javascript/index.mdx:136:  src="https://browser.sentry-cdn.com/{{ packages.version('sentry.javascript.browser') }}/bundle.apm.min.js"
src/platforms/javascript/index.mdx:137:  integrity="sha384-{{ packages.checksum('sentry.javascript.browser', 'bundle.apm.min.js', 'sha384-base64') }}"
src/wizard/javascript/index.md:12:  src="https://browser.sentry-cdn.com/{{ packages.version('sentry.javascript.browser') }}/bundle.apm.min.js"
src/wizard/javascript/index.md:13:  integrity="sha384-{{ packages.checksum('sentry.javascript.browser', 'bundle.apm.min.js', 'sha384-base64') }}"
src/wizard/javascript/index.md:22:import { Integrations as ApmIntegrations } from '@sentry/apm';
src/wizard/javascript/react.md:17:$ yarn add @sentry/react @sentry/apm
src/wizard/javascript/react.md:20:$ npm install @sentry/react @sentry/apm
src/wizard/javascript/react.md:33:import { Integrations } from '@sentry/apm';
```

AbhiPrasad:
@rhcarvalho I removed all references to `apm`. Can you please take another look at this PR? 

Also, I kept `import { Integrations } from "@sentry/tracing";` because we combine `Integrations` in the CDN bundle. To keep behaviour consistent, I changed this. We can come back to review this when we look at integrations in general for v6.

src/docs/product/performance/apm-to-tracing.mdx
```diff
@@ -0,0 +1,107 @@
...
+sidebar_order: 20
+---
+
+Our performance JavaScript SDK has been renamed from [`@sentry/apm`](https://www.npmjs.com/package/@sentry/apm) to [`@sentry/tracing`](https://www.npmjs.com/package/@sentry/tracing). The behaviour between `@sentry/apm` and `@sentry/tracing` is _identical_, but some imports and API have changed slightly.
+
+Although `@sentry/apm` is deprecated for `@sentry/tracing`, we will continue to support `@sentry/apm` with bug fixes for all `5.x.x` versions.
+
+## Migrating from @sentry/apm to @sentry/tracing
+
+For most cases, just replacing your `@sentry/tracing` import is enough to switch packages.
```
rhcarvalho, line 12: Looking at the examples, it seems that more realistically at least 2 changes are required (for the majority of users):

```diff
 -import "@sentry/apm";
 +import "@sentry/tracing";
 ```

and

```diff
 Sentry.init({
    integrations: [
 -    new Integrations.Tracing(),
 +    new Integrations.BrowserTracing(),
    ]
  });
```


Suggestion:

Break this down into `Browser` and `Node`. For browser the change is as above or the new CDN snippet (replacing the `bundle.apm.min.js` one).

For Node.js, is the import change enough?

src/docs/product/performance/apm-to-tracing.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+ Sentry.init({
+   integrations: [
+-    new Integrations.Tracing(),
++    new Integrations.BrowserTracing(),
+   ]
+ });
+```
+
+If you were using the `beforeNavigate` option from the `Tracing` integration, the API for the `beforeNavigate` option in `BrowserTracing` has changed slightly. Instead of passing in a location and returning a string representing transaction name, `beforeNavigate` now accepts a transaction context and is expected to return transaction context. You can now add extra tags or change the op based on different parameters. If you want to access the location like before, you will have to manually look at `window.location`.
```
rhcarvalho, line 44: I'm missing a mention here about the rename `beforeNavigation` -> `beforeNavigate` (or was it just the docs that were wrong before?)

AbhiPrasad: Docs were wrong before, it was always called `beforeNavigate`.

rhcarvalho: I think bad docs before.

src/docs/product/performance/apm-to-tracing.mdx
AbhiPrasad: Yes for Node the import change is enough. I guess if you were using Express, there is an added step. I will change.

AbhiPrasad: Wait, actually we should be fine because we didn't change the name for Express.

src/includes/performance-monitoring/configuration/javascript.mdx
```diff
@@ -1,36 +1,42 @@
...
+$ yarn add @sentry/browser @sentry/tracing
 
 # Using npm
-$ npm install @sentry/browser @sentry/apm
+$ npm install @sentry/browser @sentry/tracing
 ```
 
+<Alert level="warning" title="Note"><markdown>
+
+We switched our package name from `@sentry/apm` to `@sentry/tracing`. If you were previously using `@sentry/apm`, please see [Switching to the tracing package](/product/performance/apm-to-tracing/)
```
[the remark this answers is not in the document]
rhcarvalho: I'm going to apply Fiona's suggestion to match the page title.

src/includes/performance-monitoring/configuration/javascript.mdx
```diff
@@ -1,36 +1,42 @@
...
+$ yarn add @sentry/browser @sentry/tracing
 
 # Using npm
-$ npm install @sentry/browser @sentry/apm
+$ npm install @sentry/browser @sentry/tracing
 ```
 
+<Alert level="warning" title="Note"><markdown>
+
+We switched our package name from `@sentry/apm` to `@sentry/tracing`. If you were previously using `@sentry/apm`, please see [Switching to @sentry/tracing](/product/performance/apm-to-tracing/)
```
rhcarvalho, line 13:
```suggestion
We switched our package name from `@sentry/apm` to `@sentry/tracing`. If you were previously using `@sentry/apm`, please see [Switching to @sentry/tracing](/product/performance/apm-to-tracing/).
```

src/includes/performance-monitoring/configuration/javascript.mdx
AbhiPrasad: After revising this, I decided to not change the imports. In the CDN bundle, we keep everything together in `Sentry.Integrations`, hence why we chose to keep it the same way in the npm package. I think we should for now keep the same behaviour in `@sentry/apm` and we can revisit integrations in general in v6.

src/includes/performance-monitoring/configuration/javascript.mdx
rhcarvalho: Hmm, I don't like sneaking this in "Migrate to @sentry/tracing" PR.

We could as well decide that startTransaction always updates the scope, making it less of a burden to associate errors and transactions.

rhcarvalho (changes requested):

src/includes/performance-monitoring/configuration/react.mdx
```diff
@@ -32,7 +38,7 @@ ReactDOM.render(<App />, rootNode);
 
 **`withProfiler` Higher-Order Component**
 
-`@sentry/react` exports a `withProfiler` higher-order component that leverages the `@sentry/apm` Tracing integration to add React related spans to transactions. If the Tracing integration is not enabled, the Profiler will not work.
+`@sentry/react` exports a `withProfiler` higher-order component that leverages the `@sentry/tracing` to add React-related spans to transactions.
```
rhcarvalho, line 41:
```suggestion
`@sentry/react` exports a `withProfiler` higher-order component that leverages the `@sentry/tracing` package to add React-related spans to transactions.
```

src/includes/performance-monitoring/configuration/vue.mdx
```diff
@@ -4,39 +4,45 @@ To get started with performance monitoring with Vue.js, first install these pack
...
+We switched our package name from `@sentry/apm` to `@sentry/tracing`. If you were previously using `@sentry/apm`, please see [Switching to @sentry/tracing](/product/performance/apm-to-tracing/).
+
+</markdown></Alert>
+
 The Vue Tracing Integration allows you to track rendering performance during an initial application load.
 
 Sentry injects a handler inside Vue's `beforeCreate` mixin, providing access to a Vue component during its life cycle stages.
-When Sentry encounters a component named `root`, which is a top-level Vue instance (as in `new Vue({})`), we use our AM Tracing integration,
-and create a new activity named `Vue Application Render`. Once the activity has been created, it will wait until all of its child components render, and there aren't new rendering events triggered within the configured `timeout`, before marking the activity as completed.
+When Sentry encounters a component named `root`, which is a top-level Vue instance (as in `new Vue({})`), we use our AM BrowserTracing integration,
```
rhcarvalho, line 22: > and there aren't new rendering events triggered within the configured `timeout`

Did this behavior change?

AbhiPrasad: I changed this because I felt that we didn't need to expose activities/`idleTransaction` concepts here.

If we really want to do this, we can do it in `BrowserTracing`.

In addition, with `@sentry/tracing`, the Vue integration now can work without `BrowserTracing` and `idleTransactions`, so referring to activities/`timeout` is not always correct.

rhcarvalho (approved):
💯

PR #2026, opened 2020-08-18
feat: New platform layout and JavaScript conversion

dcramer:
image


dcramer:
Will update README to detail how this works and (at least) fully convert Python before we merge this.

dcramer:
I think we're in a good spot. I documented the layout in README.md, so you should treat that as the source of truth. If it doesnt make sense please leave feedback, as thats the only description you're getting.

This converts JavaScript only, as this layout doesn't really work with our historical "describe everything techincally" approach.

We ideally would have breadcrumbs in place before merging this.

dcramer:
Also, needless to say this includes a bunch of unrelated things, including TypeScript.

dcramer:
for anyone reviewer: ``src/gatsby/createPages.ts`` is the bulk of the work here

HazAT (approved):
I tried it out locally and it seems to work really well.
This is basically exactly what we envisioned for the new SDK docs 🥇 
I also like that we call it guides now, makes a lot of sense.

I skimmed over the code and couldn't find anything obvious so I approve this from a: "tried it locally and it works perspective".

dcramer:
Breadcrumbs and logo are updated - will merge on green.

dcramer:
image

PR #2030, opened 2020-08-19
build: Add test for platform json

dcramer:


dcramer:
my bash fu is weak.. will revisit this later

PR #2036, opened 2020-08-19
Feat: Global Algolia Search

cameronmcefee:
This rips out the old search implementation, which relies on the complicated algolia instantsearch module, in favor of https://github.com/getsentry/sentry-global-search, which provides a normalized api and yields results across all sentry static sites.

This does not address indexing itself, which still warrants improvement.

A question to be considered in this PR is whether this method of bringing in the module is suitable. I opted to include it via a repo rather than publishing a package.

Screen Shot 2020-08-19 at 3 27 54 PM


dcramer:
Does this no longer show the path in the search results? That was super useful to different pages that are named the same but different languages/contexts.

cameronmcefee:
Cool, wasn't sure if that was applicable since we control the content. Will do.

Correct, no schema changes at this point. I want to do that eventually but figured it'd be better to do that in a separate pull.

PR #2043, opened 2020-08-20
fix: Rewrite intro to Go integrations

rhcarvalho:
Main goal is to add context and make links clickable.


dcramer:
Heads up I just merged a large change that will conflict w/ this. If you (locally) do ``git fetch && git merge origin/master`` it should be clean as it mostly just moves stuff around

PR #2070, opened 2020-08-25
fix: behaviour => behavior

rhcarvalho:
We use US spelling in docs.

Note that for Elixir 'behaviour' is used as a technical term and thus preserved with a link to its definition.


untitaker (approved):

PeloWriter:
I think this is quite an elegant compromise - US spelling in the body of the content, "Elixirisms" with pointers to the rationale behind the extra "u". Nice.

PR #2073, opened 2020-08-25
fix: Use proper headers instead of bold text

rhcarvalho:
Proper header give semantic structure, are more accessible and generate
anchors we can link people to.


dashed:
@rhcarvalho Pre-gatsby, these were bolded instead of headers because they would show up on the navigation on right hand side. And it was inconsistent if every configuration had these headers. Wondering if that's not the case anymore?

Screen Shot 2020-08-26 at 3 51 59 AM

rhcarvalho:
> @rhcarvalho Pre-gatsby, these were bolded instead of headers because they would show up on the navigation on right hand side. And it was inconsistent if every configuration had these headers. Wondering if that's not the case anymore?

Good point. Honestly I thought we had fixed this to be headers, and discussed also the depth of header levels we use.

Looking at the Vercel deployment from this PR: **only the top level header appears in the right hand nav bar**

<details>

https://sentry-docs-git-rhcarvalho-md-headers.sentry.dev/product/performance/getting-started/
> image

</details>

And apparently we're able to link directly to an anchor as well:

https://sentry-docs-git-rhcarvalho-md-headers.sentry.dev/product/performance/getting-started/#tracingorigins-option

For some reason the platform dropdown is not working in the Vercel deployment.

AbhiPrasad:
I am planning to move this page, thanks for the update. Feel free to merge this in right now, but if it's not merged, I'll cherry pick the commit when I start.

PR #2079, opened 2020-08-26
fix(connect): Minor edits to Connect docs

lobsterkatie:
A few small fixes:

- remove SDK version so it's not immediately out of date
- break up installation code blocks so they both show (rather than being two tabs)
- space out options to make them easier to read and clarify option descriptions
- other random nits

### DIFF
--- a/src/platforms/node/connect.mdx
+++ b/src/platforms/node/connect.mdx
@@ -9,11 +9,12 @@ If you are using `yarn` or `npm`, you can add our package as a dependency:
 
 ```bash
 # Using yarn
-$ yarn add @sentry/node@5.20.1
+$ yarn add @sentry/node@5.21.4
 
 # Using npm
-$ npm install @sentry/node@5.20.1
+$ npm install @sentry/node@5.21.4
 ```
+<Break />
 
 ```javascript
 import connect from "connect";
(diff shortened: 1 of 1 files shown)


src/platforms/node/connect.mdx
```diff
@@ -9,11 +9,12 @@ If you are using `yarn` or `npm`, you can add our package as a dependency:
 
 ```bash
 # Using yarn
-$ yarn add @sentry/node@5.20.1
+$ yarn add @sentry/node@5.21.4
```
AbhiPrasad, line 12: I don’t think we should specify version numbers to make sure we don’t have to keep updating this

lobsterkatie: Fair. And agreed. It felt like more than I was going to do in the GH UI, LOL... but I suppose there's no reason we couldn’t just take the version off entirely. Will do that.

lobsterkatie, line 12:
```suggestion
$ yarn add @sentry/node
```

src/platforms/node/connect.mdx
```diff
@@ -9,11 +9,12 @@ If you are using `yarn` or `npm`, you can add our package as a dependency:
...
 
 # Using npm
-$ npm install @sentry/node@5.20.1
+$ npm install @sentry/node@5.21.4
```
lobsterkatie, line 15:
```suggestion
$ npm install @sentry/node
```

PR #2080, opened 2020-08-26
docs(vue): Update Vue wizard for tracing

benvinegar:
It additionally attempts to show two separate blocks for installing via npm/yarn vs. CDN, since the installation code snippet is different depending on which you choose.

I also dropped the es6 alternative use case since that feels unnecessary to me (we're already giving people 2 choices).

### DIFF
--- a/src/wizard/javascript/vue.md
+++ b/src/wizard/javascript/vue.md
@@ -5,43 +5,87 @@ support_level: production
 type: framework
 ---
 
-To use Sentry with your Vue application, you will need to use Sentry’s browser JavaScript SDK: `@sentry/browser`.
+## Instrument your application
+
+To begin collecting error and performance data from your Vue application, you'll need these packages:
+
+* `@sentry/browser` (Sentry's core browser SDK)
+* `@sentry/integrations` (contains Sentry's Vue integration)
+* `@sentry/tracing` (instruments performance data)
+
+Below are instructions for using your favorite package manager, or alternatively loaded directly from our CDN.
+
+### Using yarn or npm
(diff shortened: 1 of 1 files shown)


src/wizard/javascript/vue.md
```diff
@@ -5,81 +5,97 @@ support_level: production
 type: framework
 ---
 
-To use Sentry with your Vue application, you will need to use Sentry’s browser JavaScript SDK: `@sentry/browser`.
+## Instrument your application
 
-```bash
-# Using yarn
-$ yarn add @sentry/browser
+To begin collecting error and performance data from your Vue application, you'll need these packages:
```
AbhiPrasad, line 10:
```suggestion
To begin collecting error and performance data from your Vue application, you'll need the following packages:
```

src/wizard/javascript/vue.md
```diff
@@ -5,81 +5,97 @@ support_level: production
...
-yarn add @sentry/integrations
+$ yarn add @sentry/browser @sentry/integrations @sentry/tracing
 
 # Using npm
-npm install @sentry/integrations
+$ npm install @sentry/browser @sentry/integrations @sentry/tracing
 ```
 
-Then add this to your `app.js`:
+Next, initialize Sentry in your `app.js`:
```
AbhiPrasad, line 30: I would prefer we take a similar approach to react here.

```suggestion
Next, initialize Sentry in your app entry point before you initialize your root component. 
```

dcramer:
Shippin this for you ben

PR #2087, opened 2020-08-26
ref: Implement state-based HTML content to fix PlatformContent

dcramer:


dcramer:
Doesn't appear to fully reconcile the issue we're seeing in prod

PR #2097, opened 2020-08-26
feat: Overhaul Python docs

dcramer:
- Break up frameworks
- Move instrumentation/configuration into integrations


dcramer:
image

image

PR #2099, opened 2020-08-26
feat: Overhaul RN docs

dcramer:
- Various language
- Move manual setup to advanced setup


dcramer:
fwiw i think we can improve these in their structure a lot yet, im just going through and checking off things to make sure they are somewhat similar in structure

PR #2100, opened 2020-08-26
ref(node): Small tweaks to Default Integrations page

lobsterkatie:
The only substantive changes here are:

1) To list out the options for the `http` integration
2) To provide default option values as comments in the code
3) To demonstrate how to disable a single integration

Other than that, it's just cleaning up the language a bit.

### DIFF
--- a/src/platforms/node/common/integrations/default-integrations.mdx
+++ b/src/platforms/node/common/integrations/default-integrations.mdx
@@ -5,12 +5,18 @@ redirect_from:
   - /platforms/node/default-integrations/
 ---
 
-System integrations are integrations enabled by default that combine into the
+System integrations are integrations enabled by default that hook into the
 standard library or the interpreter itself. They're documented so you can see
-what they do and that they can be disabled if they cause issues. To disable
-system integrations set `defaultIntegrations: false` when calling `init()`.
-To override their settings, provide a new instance with your config
-to `integrations` option. For example: to change fatal error handler
+what they do and know how to disable them if they cause issues. 
+
+To disable _all_ default integrations, set `defaultIntegrations: false` when calling `init()`.
+
+To disable just one integration, provide a filtered list to the `integrations` option when calling `init()`.
+For example, to disable the `Console` integration:
+`integrations: defaults => defaults.filter(integration => integration.name !== 'Console')`.
+
+To override an integration's settings, provide a new instance to the `integrations` option when calling `init()`.
+For example, to change the fatal error handler:
 `integrations: [new Sentry.Integrations.OnUncaughtException({ onFatalError: () => { /** your implementation */ } })]`.
 
 ## Core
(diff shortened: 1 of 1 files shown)


src/platforms/node/common/integrations/default-integrations.mdx
```diff
@@ -62,27 +78,27 @@ Available options:
...
-This integration allows you to configure linked errors. They'll be recursively read up to a specified limit and lookup will be performed by a specific key. By default, the limit sets to 5 and the key used is `cause`.
+This integration allows you to configure linked errors. They'll be recursively read up to a specified limit and lookup will be performed by a specific key. By default, the limit is set to 5 and the key used is `"cause"`.
 
 Available options:
 
 ```js
 {
-  key: string;
-  limit: number;
+  key: string; // default: "cause"
```
dcramer, line 101: could use typescript syntax here

```typescript
{
  key: string = "cause"
```

lobsterkatie: Can you do that in TS? Everything I've read says you can't provide defaults to an interface.

dcramer: technically its:

```typescript
{
  key: string
}: {
  key = "cause"
}
```

but the version i gave i think is readable enough

kamilogorek: We don't have a standardized syntax for TS/JS options, but if I'd have to choose, I prefer the former version, the one that Katie used above. It' just more "obvious" imo.

kamilogorek (approved):

src/platforms/node/common/integrations/default-integrations.mdx
```diff
@@ -5,12 +5,18 @@ redirect_from:
...
 standard library or the interpreter itself. They're documented so you can see
-what they do and that they can be disabled if they cause issues. To disable
-system integrations set `defaultIntegrations: false` when calling `init()`.
-To override their settings, provide a new instance with your config
-to `integrations` option. For example: to change fatal error handler
+what they do and know how to disable them if they cause issues. 
+
+To disable _all_ default integrations, set `defaultIntegrations: false` when calling `init()`.
+
+To disable just one integration, provide a filtered list to the `integrations` option when calling `init()`.
```
rhcarvalho, line 14: Maybe a bit pedantic, but the example here shows that what has to be provided is a `function ([]integrations) => []integrations`, not a list/array directly.

```suggestion
To disable one or more of the default integrations, pass in a function to the
`integrations` option when calling `init()`. The function should take an array
of default integrations and return a filtered array, excluding the integrations
you do not want to use.
```

src/platforms/node/common/integrations/default-integrations.mdx
```diff
@@ -5,12 +5,18 @@ redirect_from:
...
+For example, to disable the `Console` integration:
+`integrations: defaults => defaults.filter(integration => integration.name !== 'Console')`.
+
+To override an integration's settings, provide a new instance to the `integrations` option when calling `init()`.
+For example, to change the fatal error handler:
```
rhcarvalho, line 19: There are so many ways to use the same `integrations` option -- `bool`, `func`, `array`.

We document using each way for one specific need. What's missing here is how to combine the needs, how to both provide specific integration settings AND remove/add integrations.

I'm guessing the answer is to use the `function` form, which seems to be the most powerful.
In fact, it could be used for all three examples:

```
// remove all integrations:
integrations: () => null; // or if we don't support this, then empty array

// remove one or more filtering by name:
integrations: defaults => defaults.filter(...)

// replace an integration with a custom version of the same integration:
integrations: defaults => defaults.map(integration =>
  integration.name === 'OnUncaughtException' ?
    new Sentry.Integrations.OnUncaughtException({ onFatalError: () => { /** your implementation */ } }) :
    integration)

// filter and update:
integrations: defaults => ...
```

kamilogorek: > // remove all integrations:

It should be just `defaultIntegrations: false` then, which is already described slightly before I guess.

> // replace an integration with a custom version of the same integration:

We always tell people to use `integrations: [new VersionWithOptions({})]` as it's more readable for this basic usecase

lobsterkatie: @rhcarvalho I think that you're right, and perhaps we should revisit this again in the future (and unify our explanation here with the one on the main JS SDK page), but for the moment I'm going to merge this and we can come back to it.

src/platforms/node/common/integrations/default-integrations.mdx
lobsterkatie: Good point, thanks.

src/platforms/node/common/integrations/default-integrations.mdx
dcramer: there is standardized HTML for options that we use everywhere (definition lists) so thats also a path forward. i didnt look at the context of what these are specifically, but its possible they dont belong in a code block to begin with.

lobsterkatie: It is possible. I was just going with the formatting that was already there. I'll let docs folks make that call. In any case, work for a future PR.

PR #2101, opened 2020-08-26
feat: Overhaul native docs

dcramer:
- Break up framework integrations
- Move native-specific stuff to top level
- Move debug information to shared
- Break up index into multiple referencable files


dcramer:
@jan-auer i could use your feedback here to see if what im doing actually makes sense

src/platforms/native/common/minidumps.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: Minidump
+title: Minidumps
```
jan-auer, line 2: This page contains two pieces of information:
- Briefly, what's a minidump in principle
- And then mostly, how to manually integrate minidumps without the SDK.

I'd suggest moving the majority of this file to "Guides", since it is a "custom" integration similar to Breakpad and Crashpad. We could consider leaving the part with "what's a minidump" here, but it might be easier to keep everything together.

src/platforms/native/common/debug-information.mdx
```diff
@@ -0,0 +1,23 @@
...
+instructions to retrieve them.
+
+In addition to Debug Information Files, Sentry needs _Call Frame Information_
+(CFI) to extract accurate stack traces from minidumps of optimized release
+builds. CFI is usually part of the executables and not copied to debug symbols.
+Unless you are uploading Breakpad symbols, be sure to also include the binaries
+when uploading files to Sentry.
+
+For more information on uploading debug information and their supported formats,
+see [Debug Information Files](/workflow/debug-files/).
```
jan-auer, line 23: @PeloWriter In JavaScript, we started to duplicate the relevant information for uploading sourcemaps into the platform docs. Should we do that here, too?

src/platforms/native/index.mdx
```diff
@@ -93,7 +88,7 @@ event data may be lost.
 Alternatively, the DSN can be passed as `SENTRY_DSN` environment variable during
 runtime. This can be especially useful for server applications.
 
-### Verifying Your Setup
+### Verify
 
 Now that SDK setup is complete, verify that all configuration is correct. Start
 by capturing a manual event:
```
jan-auer, line 94: Two things here:
1. I'd move "Capturing Events" and everything below to a separate pages. Ironically in native, this is an advanced use case and usually users just capture crashes (auto) initially. 
2. I'd also add a "What's Next" section at the bottom of this page similar to JavaScript.

src/platforms/native/advanced-config.mdx
```diff
@@ -0,0 +1,20 @@
...
+sentry_options_set_environment(options, "Production");
+sentry_options_set_release(options, "5fd7a6cd");
+sentry_options_set_debug(options, 1);
+
+sentry_init(options);
+```
+
+For more information, see:
+
+- [Sentry’s complete list of Common Options across SDKs](/error-reporting/configuration/)
```
jan-auer, line 19: For a follow-up: @PeloWriter, should we inline the configuration options here similar to what we do in JavaScript?

src/platforms/native/integrations/index.mdx
```diff
@@ -0,0 +1,16 @@
+---
+title: Integrations
```
jan-auer, line 2: Let's rename this to "Backends" (to not confuse it with the actual integrations mentioned in "Guides").

src/platforms/native/integrations/crashpad.mdx
```diff
@@ -0,0 +1,55 @@
...
+```
+
+The crashpad handler executable must be shipped alongside your application so
+that it can be launched when initializing the SDK. The path is evaluated
+according to shell lookup rules at runtime.
+
+It is advised to specify both the path to the `crashpad_handler` executable
+as well as the path to the database directory manually.
+
+## Event Attachments
```
jan-auer, line 40: Let's move this into a separate page one level up.

jan-auer (changes requested):
Thanks, @dcramer! This looks great to me and the general structure makes a lot of sense. I think, I can follow-up with @PeloWriter with tweaks in future PRs. 

There are a few suggestions below, I'm happy to apply them.

src/platforms/native/common/debug-information.mdx
dcramer: future duplication for sure will happen, but def makes sense in this case

src/platforms/native/index.mdx
dcramer: whats next is out of scope for this (mostly moving content) but i moved the other stuff

dcramer:
gonna ship we can come back and improve language/fill in more sections/etc. just trying to get all these updated

PR #2105, opened 2020-08-27
fix: android version placeholder

marandaneto:
the expression implementation 'io.sentry:sentry-android:`{{% packages.version('sentry.android', '{version}') }}'` is not being executed, so the dependency is wrong like the plain text mentioned above, the syntax looks good though, not sure what's happening, it should take the 2nd argument which is `{version}`.


HazAT (approved):

bruno-garcia:
preview: https://sentry-docs-git-fix-android-dep-version-placeholder.sentry.dev/

marandaneto:
@dcramer ideas why the expression is not being executed? the syntax looks good though.

PR #2107, opened 2020-08-27
Clean up JS integrations code samples and intro lang

PeloWriter:


src/platforms/node/common/integrations/pluggable-integrations.mdx
```diff
@@ -53,7 +53,7 @@ This integration extracts all non-native attributes from the Error object and at
 
 Available options:
 
-```js
+```JavaScript
 {
   depth: number; // limit of how deep the object serializer should go. Anything deeper than the limit will be replaced with standard Node.js REPL notation of [Object], [Array], [Function] or primitive value. Defaults to 3.
```
lobsterkatie, line 58:
```suggestion
   // limit of how deep the object serializer should go. Anything deeper than limit will 
   // be replaced with standard Node.js REPL notation of [Object], [Array], [Function] or 
   // a primitive value. Defaults to 3.
  depth: number;
```

src/platforms/node/common/integrations/pluggable-integrations.mdx
```diff
@@ -70,7 +70,7 @@ For example `C:\\Program Files\\Apache\\www` won't work, however, `/Program File
 
 Available options:
 
-```js
+```JavaScript
 {
   root: string; // root path that will be appended to the basename of the current frame's url
```
lobsterkatie, line 75:
```suggestion
  // root path that will be appended to the basename of the current frame's url 
  root: string; 
  
```

src/platforms/node/common/integrations/pluggable-integrations.mdx
```diff
@@ -70,7 +70,7 @@ For example `C:\\Program Files\\Apache\\www` won't work, however, `/Program File
...
   iteratee: (frame) => frame); // function that takes the frame, applies any transformation on it and returns it back
```
lobsterkatie, line 76:
```suggestion
  // function that takes the frame, applies a transformation, and returns it
  iteratee: (frame) => frame); 
```

src/platforms/node/guides/express/index.mdx
```diff
@@ -63,7 +63,7 @@ app.get("/debug-sentry", function mainHandler(req, res) {
 
 Possible options are:
 
-```js
+```JavaScript
 // keys to be extracted from req
 request?: boolean | string[]; // default: true = ['cookies', 'data', 'headers', 'method', 'query_string', 'url']
 // server name
```
lobsterkatie, line 69:
```suggestion

// server name
```

I'd do the rest here but they're not close enough to what you changed for it to be willing to let me make suggestions there. Personally I think spacing it out makes it much easier to read, but you'll have to do it locally if you want to do it.

PeloWriter: Done.

PR #2115, opened 2020-08-28
feat: Add tracing to angular wizard

HazAT:
Tried the code out and it works.

### DIFF
--- a/src/wizard/javascript/angular.md
+++ b/src/wizard/javascript/angular.md
@@ -11,10 +11,10 @@ Add the Sentry SDK as a dependency using `yarn` or `npm`:
 
 ```bash
 # Using yarn
-$ yarn add @sentry/angular
+$ yarn add @sentry/angular @sentry/tracing
 
 # Using npm
-$ npm install @sentry/angular
+$ npm install @sentry/angular @sentry/tracing
 ```
 
 You should `init` the Sentry browser SDK as soon as possible during your application load up, before initializing Angular:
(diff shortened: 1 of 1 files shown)


src/wizard/javascript/angular.md
```diff
@@ -23,11 +23,19 @@ You should `init` the Sentry browser SDK as soon as possible during your applica
...
 
 Sentry.init({
   dsn: '___PUBLIC_DSN___',
+  integrations: [
+    new Integrations.BrowserTracing({
+      tracingOrigins: ["localhost", "https://yourserver.io/api"],
+      routingInstrumentation: Sentry.routingInstrumentation,
+    }),
+  ],
+  tracesSampleRate: 1.0,
```
AbhiPrasad, line 38: We need to add the message about traces sample rate like we do for react/python wizard

HazAT: good catch
added

src/wizard/javascript/angular.md
```diff
@@ -23,11 +23,19 @@ You should `init` the Sentry browser SDK as soon as possible during your applica
...
 import * as Sentry from '@sentry/angular';
+import { Integrations } from "@sentry/tracing";
 
 import { AppModule } from './app/app.module';
 
 Sentry.init({
   dsn: '___PUBLIC_DSN___',
+  integrations: [
+    new Integrations.BrowserTracing({
+      tracingOrigins: ["localhost", "https://yourserver.io/api"],
```
silent1mezzo, line 34: @HazAT Is there a reason we have `["localhost", "https://yourserver.io/api"]` here instead of `['localhost', /^\//]` this threw me off when I was trying to implement it for the first time (I am not an Angular expert by any means).

benvinegar:
Thanks @HazAT 🙏

PR #2119, opened 2020-08-28
feat: New Platform Portals

dcramer:
This takes the platform content effort and elevates it to all shared SDK content, as well as implements the core of the new portal constructs.

- New homepage gives you persona-based portals (e.g. a specific platform, product guides, self-hosted, support).
- Selecting a platform keeps the existing focused sidebar, but now duplicates content from enriching-error-data and data-management.
- A huge swath of fixes for new and old platform content to clean it up and correct it, as well as missing content.
- A few new helper components, like PlatformSection and PageGrid.
- Moved platform detection logic into `usePlatform` hook.
- Added platformicons package and implementation. Still needs updated for homepage when new icons are available.
- Improvements to devenv (test suite, eslint).

The README is updated with a number of the user-facing changes (including core new components).

PR #2137, opened 2020-08-31
Create Open API docs v1

NisanthanNanthakumar:
## Objective
We want to create API docs from an OpenAPI document. First we created a gatsby-plugin to convert `openapi-derefed.json` from `getsentry/sentry-api-schema` into graphql nodes. Next, we created a new docs page under the route `/development-api/*`. This is temporary and we will be taking over the `/api/*` routes. Next we created a robots.txt file and disallowed crawling `/development-api/*`. 

Note: We resolve a specific commit of `getsentry/sentry-api-schema` into graphql nodes. Currently, this SHA needs to be updated manually.

## UI
Screen Shot 2020-08-31 at 10 33 24 AM

### The TODOs will be addressed in different PRs
- We can reuse the Authentication docs page that currently exists.
- We need to include the schema of the response in the docs.
- We need to include some visual sign of which parameters are required/optional.
- We need to create index directory pages for each tag.


NisanthanNanthakumar:
> https://sentry-docs-git-create-openapi-docs.sentry.dev

https://sentry-docs-git-create-openapi-docs.sentry.dev/development-api/update-team

dcramer:
@NisanthanNanthakumar how do you feel about swapping out the sidebar in api.js (see how its done in platform.js) and just relying on breadcrumbs to go back to the homeage? would let you be much more flexible with navigation that way

NisanthanNanthakumar:
> @NisanthanNanthakumar how do you feel about swapping out the sidebar in api.js (see how its done in platform.js) and just relying on breadcrumbs to go back to the homeage? would let you be much more flexible with navigation that way

yeah that looks good to me! we could even outline all the requests in the sidebar, similar to stripe docs

gatsby-config.js
```diff
@@ -170,6 +172,23 @@ const getPlugins = () => {
...
       },
     },
+    {
+      resolve: "./plugins/gatsby-plugin-openapi",
+      options: {
+        name: "openapi",
+        resolve: async () => {
+          try {
+            const response = await axios.get(
+              "https://raw.githubusercontent.com/getsentry/sentry-api-schema/68bb79acfbbee062bd8d2f71ee3a07d43dc934c9/openapi-derefed.json"
```
dcramer, line 182: is this temporary? should it be master? does it need updated regularly? at the very least its probably worht a comment explaining the url

NisanthanNanthakumar: @dcramer I wrote a small note in the Objective addressing this. Currently, the plan is to fix it to a specific SHA from `getsentry/sentry-api-schema` and update it manually. Eventually we can have a bot/github action do it automatically.

plugins/gatsby-plugin-openapi/gatsby-node.js
```diff
@@ -0,0 +1,114 @@
...
+      }, []);
+
+    const rootNode = {
+      ...parsedContent,
+      paths: data,
+      id: createNodeId(`openAPI-root`),
+      children: [],
+      parent: null,
+      internal: {
+        contentDigest: createContentDigest("openAPI"),
```
dcramer, line 105: should the digest be hashing the data? 

here's how ive done it:

https://github.com/getsentry/sentry-docs/pull/2119/files#diff-c03e26a205ea6e8099f39ca510c6f61dR220-R236

robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
```
dcramer, line 1: dont see how this file would get used anywhere, I think you'd need to put it into the 'static' folder for it to get copied over

src/gatsby/createPages/createDevelopmentApiPages.ts
```diff
@@ -0,0 +1,76 @@
...
+  const component = require.resolve(`../../templates/developmentAPI.js`);
+  await Promise.all(
+    data.allOpenApi.nodes.map(async node => {
+      await Promise.all(
+        node.paths.map(async p => {
+          actions.createPage({
+            path: `/development-api/${p.readableUrl}`,
+            component,
+            context: {
+              ...p,
```
dcramer, line 68: its possible this _might_ balloon gatsbys stuff and make it angry

generally, and ive been learning this as im overhauling platforms, its recommended to create a Node that contains this data, and then read that node in the template itself otherwise you end up adding all the fields to every node of the same type (so now all pages have all these API attributes as context)

NisanthanNanthakumar: ahh interesting! ok I'll make the necessary changes.

src/gatsby/createPages/createDevelopmentApiPages.ts
```diff
@@ -0,0 +1,72 @@
+import { getDataOrPanic } from "./helpers";
+
+export default async ({ actions, graphql, reporter }) => {
+  const data = await getDataOrPanic(
+    `
+      query {
+        allOpenApi {
```
dcramer, line 7: nice! the cool thing is this makes it easy to build the sidebar or any other dynamic navigation

PeloWriter:
Nice update. This looks really good.

PR #1784, opened 2020-06-24
feat(apm): Add markers to be used for inline tracing docs for the product

dashed:
Mark sections of the tracing SDK docs to be used for inline tracing docs on the product.

The corresponding PR in `sentry`: https://github.com/getsentry/sentry/pull/19527

-----

Node.js docs:

Screen Shot 2020-06-24 at 7 37 20 AM

Python docs:

Screen Shot 2020-06-24 at 7 36 04 AM

PR #2138, opened 2020-08-31
Remove Python tracing wizard…

silent1mezzo:
This causes confusion as it's not for a language/framework.

These docs are included https://docs.sentry.io/product/performance/getting-started/?platform=python

Screen Shot 2020-08-26 at 9 10 55 AM


silent1mezzo:
@dcramer I can't find any references to it. I'll wait for @dashed to respond but it's confusing on create new project page.

dashed:
@silent1mezzo they're used here https://github.com/getsentry/sentry/pull/19527 and added to sentry-docs (pre-gatsby) https://github.com/getsentry/sentry-docs/pull/1784 .

We just need to somehow hide these tracing entries on the platform list for the project setup in Sentry. I didn't anticipate this when I added this.

The wizard docs was being piggybacked for now until there's a more formal system to pull these into the product.

PR #2149, opened 2020-09-01
fix: Various npm installation instructions

dcramer:


dcramer:
@kamilogorek mentioned that current generation NPM (since 2017) does not require the ``--save`` flag. For now we'll keep it, and we can decide at a point in time to remove it based on legacy support (and remove it fully).

Of note this also simplifies the Vue instructions, which for some reason we thought you should run two different npm install commands.

PR #2155, opened 2020-09-02
cocoa sdk ios v9

bruno-garcia:
We've bumped the min version to iOS v9 for the next release of Sentry cocoa (6.0

https://github.com/getsentry/sentry-cocoa/pull/669

### DIFF
--- a/src/platforms/cocoa/index.mdx
+++ b/src/platforms/cocoa/index.mdx
@@ -15,7 +15,7 @@ The SDK can be installed using [CocoaPods](http://cocoapods.org), [Carthage](htt
 
 Support for
 
-- iOS >= 8.0
+- iOS >= 9.0
 - tvOS >= 9.0
 - macOS >= 10.10
 - watchOS >= 2.0 with limited symbolication support and no hard crashes


marandaneto:
> We've bumped the min version to iOS v10 for the next release of Sentry cocoa (6.0
> 
> [getsentry/sentry-cocoa#669](https://github.com/getsentry/sentry-cocoa/pull/669)

you have a typo, it's v9 and not v10, I've fixed the PR's title.

src/platforms/cocoa/index.mdx
```diff
@@ -15,7 +15,7 @@ The SDK can be installed using [CocoaPods](http://cocoapods.org), [Carthage](htt
 
 Support for
 
-- iOS >= 8.0
+- iOS >= 9.0
```
marandaneto, line 18: would make sense to explain that Cocoa v5 is iOS 8 and Cocoa v6 is iOS 9? we have a single page for 2 SDK versions, this is confusing IMO, at some point either we'll be misleading users or even losing them due to broken docs

bruno-garcia: The reason we dropped is that no one uses iOS 8 anymore anyway.
I'd say lets just document the current version (in this case 6.0)

bruno-garcia: Granted we need to review API docs too once GA lands

philipphofmann: Don't we have the possibility to have docs for different SDK versions?

bruno-garcia: That's ideal but generally more work. We just point users to upgrade at any issue which is less maintenance burden for us.

Lets try to stick to this for as long as we can.

PR #2161, opened 2020-09-02
feat: Add fallbackPlatform support to base languages

dcramer:
This is still partial support, and currently only works with PlatformContent.

It also adds the node fallback to javascript.


dcramer:
I wanted to add test coverage here but for whatever reason when I try to add PlatformContent to the test suite it explodes with an error that I was unable to diagnose.

PR #2183, opened 2020-09-03
feat: Replace notes with more subtle (controlled) styled component

dcramer:


src/components/alert.tsx
```diff
@@ -8,22 +9,26 @@ type Props = {
...
+`;
+
 export default ({
   title,
   children,
   level,
   deepLink,
   dismiss = false,
 }: Props): JSX.Element => {
-  let className = "alert";
```
dcramer, line 18: whoops unrelated change backing out

PR #2185, opened 2020-09-03
feat: Overhaul sourcemaps documentation

dcramer:
- Simplify documentation to only the recommended workflow
- Optimize for Angular by showing SystemJS as default over Webpack
- Move various extra content to other pages


dcramer:
image

PR #2187, opened 2020-09-03
bumping hero up b/c of what I think is aliasing cuasing weird space, …

stevenplewis:
…also centering logo on mobile.


dcramer (approved):

dcramer:
unrelated, but we probably should remove some bottom margin on icons to consolidate whitespace

PR #2195, opened 2020-09-04
fix: Remove language about @sentry/apm

dcramer:
Most folks don't know what this means, and folks who might can probably figure it out.


dcramer:
The general approach to documentation should be:

"Does a user need to care about this?"

And the bias should always be towards new users.

Something something Marie Kondo your docs. This sentence provides users zero joy.

PR #2198, opened 2020-09-04
fix: Use Link component instead of bare markdown links

untitaker:


untitaker:
Planning to do link checking on build output

PR #2199, opened 2020-09-04
fix: Link to DSN config option

joergi:
the original link was only a 404, this should fix it

before: https://docs.sentry.io/platforms/java/guides/spring/config/#setting-the-dsn
after: https://docs.sentry.io/platforms/java/configuration/options/#setting-the-dsn

### DIFF
--- a/src/platforms/java/guides/spring/index.mdx
+++ b/src/platforms/java/guides/spring/index.mdx
@@ -49,7 +49,7 @@ public HandlerExceptionResolver sentryExceptionResolver() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


rhcarvalho:
Thanks @joergi

joergi:
@rhcarvalho - I have seen a similar error somewhere else...
is the old way of linking to it somewhere still available?
or was there a complete change to the new one?

so was everything changed from
`config/#....`
to 
`configuration/options/#....`
?

Or are there still some valid links for `config/#....` ?

rhcarvalho:
@joergi there was a big change recently, sorry for the disruption and thanks for reporting. I went ahead and fixed links following that pattern. Yesterday we [added a link checker to CI](https://github.com/getsentry/sentry-docs/pull/2209) and should be fixing the rest of the broken links in the next days.

PR #2200, opened 2020-09-04
Index page for Getting Started

PeloWriter:
We're working toward an index page for Getting Started for our SDKs. I've edited to the JS index page further, which I think can serve as a model (also evaluated PHP as part of this creating this PR).

This is closer, though the section in Capture Errors regarding global handlers may not be common.


PeloWriter:
We have another idea for how to make this page better; closing this PR.

PR #2203, opened 2020-09-07
fix: Child markdown node interception

jan-auer:
Fixes onCreateNode to skip nodes that do not have an absolute file path set. For
such nodes, the `gatsby-source-filesystem` has a buggy resovler that throws an
error while trying to retrieve file information.

It seems that the intention of this hook was to inject attributes on markdown
pages. However, markdown snippet nodes, such as in Relay metrics, do not need 
the `slug` and `legacy` fields.

Along with this fix, the relationship between `RelayMetric` and
`RelayMetricDescription` was changed to many-to-one by adding a reference to the
`parent` id.

### DIFF
--- a/src/components/relayMetrics.tsx
+++ b/src/components/relayMetrics.tsx
@@ -11,7 +11,7 @@ const query = graphql`
         type
         features
 
-        childrenRelayMetricDescription {
+        childRelayMetricDescription {
           childMarkdownRemark {
             html
           }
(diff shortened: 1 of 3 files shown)


jan-auer:
Fixes onCreateNode to skip nodes that do not have an absolute file path
set. For such nodes, the `gatsby-source-filesystem` has a buggy resovler
that throws an error while trying to retrieve file information.

It seems that the intention of this hook was to inject attributes on
markdown pages. However, markdown snippet nodes, such as in Relay
metrics, do not need the `slug` and `legacy` fields.

Along with this fix, the relationship between `RelayMetric` and
`RelayMetricDescription` was changed to many-to-one by adding a
reference to the `parent` id.

PR #2206, opened 2020-09-08
Update filtering.mdx - Fix breadcrumbs link

badsyntax:


src/platforms/common/configuration/filtering.mdx
```diff
@@ -59,7 +59,7 @@ All Sentry SDKs support the `beforeSend` callback method. `before-send` is calle
 
 <PlatformContent includePath="configuration/before-send" />
 
-Note also that breadcrumbs can be filtered, as discussed in [our Breadcrumbs documentation](/docs/product/error-monitoring/breadcrumbs/).
+Note also that breadcrumbs can be filtered, as discussed in [our Breadcrumbs documentation](/platforms/javascript/enriching-error-data/breadcrumbs/).
```
rhcarvalho, line 62:
```suggestion
Note also that breadcrumbs can be filtered, as discussed in [our Breadcrumbs documentation](/product/error-monitoring/breadcrumbs/).
```

Thanks for noticing the broken link and sending out a PR.

The correct link is `/product/error-monitoring/breadcrumbs/`, without the `/docs` prefix.

https://docs.sentry.io/product/error-monitoring/breadcrumbs/

rhcarvalho (changes requested):

src/platforms/common/configuration/filtering.mdx
badsyntax: I'll make this change now. 
I'm going through the docs, trying to figure out how to use the SDK, and i'm finding many broken links. I don't have time to report them all. I'd suggest doing a review of your documentation to fix all the broken links.

rhcarvalho: Sorry you had this bad experience. We're iterating fast improving our docs.

As part of the project we will [automate link-checking](https://github.com/getsentry/sentry-docs/projects/1#card-44769135) to avoid this kind of problem altogether in the future.
Current experiment is https://github.com/getsentry/sentry-docs/pull/2209.

rhcarvalho (approved):

src/platforms/common/configuration/filtering.mdx
badsyntax: Sorry for not finding the time to make the suggested change. Thanks for doing it 👍 

Good to hear you have a solution in place to check the links. I want to also point at that I had encountered a couple broken backlinks from Google. I was getting 404's when visiting docs pages from google search results. (Sadly I can't remember the search terms I was using.) It would be great if you could implement 301 redirects when changing urls too. I understand the technical difficulties with regards, but it's important to not "break the web" like this.

PR #2209, opened 2020-09-08
fix: Run linkchecker in CI

untitaker:
This is very slow on my machine, let's see how long github actions takes.


untitaker:
This takes 15 min, which is acceptable, and doesn't block anything.

PR #2211, opened 2020-09-08
feat: DevelopmentAPI Sidebar

NisanthanNanthakumar:
## Objective:
We want to customize the sidebar of the OpenAPI docs and show something similar to the [Stripe docs](https://stripe.com/docs).
If a user wants to exit the OpenAPI Reference then they can use the breadcrumbs navigation.

## UI
Screen Shot 2020-09-08 at 11 45 16 AM
Screen Shot 2020-09-08 at 11 45 27 AM

### DIFF
--- a/src/components/developmentApiSidebar.tsx
+++ b/src/components/developmentApiSidebar.tsx
@@ -1,9 +1,9 @@
 import React from "react";
-import { graphql, useStaticQuery, withPrefix } from "gatsby";
+import { graphql, useStaticQuery } from "gatsby";
 import { useLocation } from "@reach/router";
 
 import SidebarLink from "./sidebarLink";
-import DynamicNav, { toTree, EntityTree } from "./dynamicNav";
+import DynamicNav, { toTree } from "./dynamicNav";
 
 const query = graphql`
   query DevelopmentApiNavQuery {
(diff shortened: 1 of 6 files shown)


src/components/developmentApiSidebar.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  let endpoints = tree[0].children.reduce((acc: EntityTree[], curr: EntityTree) => {
+    if (curr.children.length > 1) {
+      acc.push(curr);
+    }
+    return acc;
+  }, []);
+
+  const location = useLocation();
+  const isActive = path =>
```
mgaeta, line 36: Is `.startsWith` available?

src/components/developmentApiSidebar.tsx
```diff
@@ -0,0 +1,75 @@
...
+export default () => {
+  const data = useStaticQuery(query);
+  const tree = toTree(data.allSitePage.nodes.filter(n => !!n.context));
+
+  let endpoints = tree[0].children.reduce((acc: EntityTree[], curr: EntityTree) => {
+    if (curr.children.length > 1) {
+      acc.push(curr);
+    }
+    return acc;
+  }, []);
```
mgaeta, line 33: I think you should use a `.filter`: 
```
const endpoints = tree[0].children.filter(child => child.children.length > 1);
```

mgaeta (approved):
Cool 👍

manuzope:
Should https://docs.sentry.io/development-api/ show the new sidebar?

PR #2217, opened 2020-09-09
Fix: Broken Links

MimiDumpling:
Checked the docs 404s in Discover and created redirects for each broken link.

### DIFF
--- a/src/platforms/php/guides/symfony/config.mdx
+++ b/src/platforms/php/guides/symfony/config.mdx
@@ -1,7 +1,5 @@
 ---
 title: Config
-redirect_from:
-  - /clients/php/integrations/monolog/
 ---
 
 In addition to the [configuration available in the PHP SDK](/platforms/php/config/), there are some Symfony-specific options.
--- a/src/platforms/php/guides/symfony/config.mdx
+++ b/src/platforms/php/guides/symfony/config.mdx
@@ -1,5 +1,7 @@
 ---
 title: Config
+redirect_from:
+  - /clients/php/integrations/monolog/
 ---
 
 In addition to the [configuration available in the PHP SDK](/platforms/php/config/), there are some Symfony-specific options.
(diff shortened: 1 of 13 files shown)


MimiDumpling:
Dear reviewers, will someone help with a few issues I ran into:
1. adding a redirect to any page nested under /platforms/common/ will default to the Android version
2. adding a redirect to /platforms/php/guides/symfony/config/ resulted in an error
3. adding a redirect to /platforms/python/redis/ resulted in an error

HazAT (approved):
🎉

src/platforms/php/guides/symfony/config.mdx
```diff
@@ -1,5 +1,7 @@
 ---
 title: Config
+redirect_from:
+  - /clients/php/integrations/monolog/
```
untitaker, line 4: Why do we redirect to Symfony from Monolog?

src/platforms/python/guides/rq/index.mdx
```diff
@@ -3,6 +3,7 @@ title: RQ (Redis Queue)
 redirect_from:
   - /clients/python/integrations/rq/
   - /platforms/python/rq/
+  - /platforms/python/redis/
```
untitaker, line 6: we should redirect to `/platforms/python/integrations/redis/`, rq is not redis

src/platforms/ruby/index.mdx
```diff
@@ -4,6 +4,7 @@ caseStyle: snake_case
 supportLevel: production
 redirect_from:
   - /clients/ruby/
+  - /platforms/ruby/config/
```
untitaker, line 7: I think we may want to redirect to https://docs.sentry.io/platforms/ruby/configuration/ instead of the index page (same for others)

untitaker (changes requested):

PR #2222, opened 2020-09-09
fix: Markdown for image instead of link

rhcarvalho:
Also add an 'alt' for the image.

See https://docs.sentry.io/platforms/javascript/configuration/lazy-load-sentry/#additional-configuration

---

### Before

image

---

### After

image


rhcarvalho:
Obvious fix, going to merge it.

PR #2225, opened 2020-09-09
fix: Run single request for SDKs

untitaker:


src/gatsby/utils/packageRegistry.ts
```diff
@@ -45,28 +43,16 @@ export default class PackageRegistry {
...
-    return this.cache[name];
+    await this.getList();
+    return this.indexCache[name]
   };
 
-  version = async (name, defaultValue: string = "") => {
+  version = async (name: string, defaultValue: string = "") => {
     const data = (await this.getData(name)) as VersionData;
-    return data.version || defaultValue;
+    return data && data.version || defaultValue;
```
BYK, line 52: Are you TS yet?

```suggestion
    return data?.version || defaultValue;
```

BYK (approved):
gogogo

src/gatsby/utils/packageRegistry.ts
untitaker: i'm in fact not ts yet, doing this breaks

i think we're running an old ts

BYK: or maybe I don't know how to ts 😁

untitaker: no I tried before, and all the stuff we can do in sentry frontend doesn't work here

PR #2227, opened 2020-09-09
Revert "fix: Run single request for SDKs"

untitaker:
Reverts getsentry/sentry-docs#2225

the registry is instantiated multiple times (per-call), so this doesn't actually cache as much


PeloWriter:
Closing after brief convo.

PR #2228, opened 2020-09-09
fix: Show custom sidebar on "/development-api"

NisanthanNanthakumar:
## Objective
OpenAPI docs sidebar not showing up on `/development-api/` in production

## UI
Screen Shot 2020-09-09 at 10 46 56 AM

### DIFF
--- a/src/templates/doc.tsx
+++ b/src/templates/doc.tsx
@@ -3,10 +3,19 @@ import { graphql } from "gatsby";
 
 import BasePage from "~src/components/basePage";
 import Content from "~src/components/content";
+import DevelopmentApiSidebar from "~src/components/developmentApiSidebar";
 
 export default props => {
+  console.log({ props });
   return (
-    <BasePage {...props}>
+    <BasePage
+      sidebar={
+        props.path.startsWith("/development-api") ? (
+          <DevelopmentApiSidebar />
+        ) : null
+      }
+      {...props}
+    >
       <Content file={props.data.file} />
     </BasePage>
   );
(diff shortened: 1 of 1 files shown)


PeloWriter (approved):
LGTM - one q: on the current docs site, these display under Endpoints:
- Events & Issues
- Releases
- Projects
- Teams 
- Organizations

In the screenshot, I see only Teams.

mgaeta:
I think you could avoid the ternary with null if you do something like:
```
export default props => {
  if (props.path.startsWith("/development-api")) {
    props.sidebar = <DevelopmentApiSidebar/>;
  }
  return (
    <BasePage {...props}>
      <Content file={props.data.file} />
    </BasePage>
  );
};
```

NisanthanNanthakumar:
> LGTM - one q: on the current docs site, these display under Endpoints:
> 
> * Events & Issues
> * Releases
> * Projects
> * Teams
> * Organizations
> 
> In the screenshot, I see only Teams.

We are on a specific SHA of https://github.com/getsentry/sentry-api-schema which we use to generate the GraphQL nodes for the OpenAPI docs. Right now, not all endpoints have been migrated. In the near future, we will have endpoint parity.

NisanthanNanthakumar:
> if (props.path.startsWith("/development-api")) {
>     props.sidebar = <DevelopmentApiSidebar/>;
>   }
>   return (
>     <BasePage {...props}>
>       <Content file={props.data.file} />
>     </BasePage>
>   );

@mgaeta When I tried that, I got the following error:
Screen Shot 2020-09-09 at 11 26 55 AM

I guess we can't directly mutate `props`.

However, this works:
```
props = { ...props, sidebar: <DevelopmentApiSidebar /> }
```

I'm ok with `null` because we default to the main sidebar if there is no valid custom sidebar component, which the desired behaviour.

PR #2234, opened 2020-09-09
feat: Tell Vercel to use trailing slash on URLs

rhcarvalho:
That's what we had pre-Vercel and Gatsby, so keep canonical URLs with a
trailing slash.

Reference and explanation why the default value 'undefined' is harmful
(double indexing harming SEO):

https://vercel.com/docs/configuration#project/trailingslash/true

### DIFF
--- a/vercel.json
+++ b/vercel.json
@@ -22,6 +22,7 @@
       ]
     }
   ],
+  "trailingSlash": true,
   "redirects": [
     { "source": "/(hosted|on-premise)/(.*)", "destination": "/$2" },
     { "source": "/internal/(.*)", "destination": "https://develop.sentry.dev" },


rhcarvalho:
## Before (example)

For every page in the website, crawlers may find pairs of pages with identical content:

```console
$ curl -Iso /dev/null -w '%{http_code} %{url_effective}\n' 'https://docs.sentry.io/platforms/cocoa{,/}'
200 https://docs.sentry.io/platforms/cocoa
200 https://docs.sentry.io/platforms/cocoa/
```

## After

Site always redirects to the page with the trailing slash (note the 308 status code below): single URL per content == better SEO.

```console
$ curl -Iso /dev/null -w '%{http_code} %{url_effective}\n' 'https://sentry-docs-git-rhcarvalho-vercel-trailingslash.sentry.dev/platforms/cocoa{,/}'
308 https://sentry-docs-git-rhcarvalho-vercel-trailingslash.sentry.dev/platforms/cocoa
200 https://sentry-docs-git-rhcarvalho-vercel-trailingslash.sentry.dev/platforms/cocoa/
```

BYK (approved):

PR #2253, opened 2020-09-11
fix: Remove redirects

rhcarvalho:
It does not make sense to have the same URL in the 'redirect_from' list
of multiple pages, because visiting the given URL can only redirect to
one location, not multiple.

AFAIK the has never been content in those URLs we are redirecting from,
so either we need different redirects or none (the content under the
"common" directory is used to generate pages, but "common" is not part
of the resulting URLs).

Updates 84d8a9375bfa997632fe792655a6a1bc004c2a46

### DIFF
--- a/src/platforms/javascript/common/lazy-load-sentry.mdx
+++ b/src/platforms/javascript/common/lazy-load-sentry.mdx
@@ -1,8 +1,6 @@
 ---
 title: Lazy Loading Sentry
 sidebar_order: 11
-redirect_from:
-  - /platforms/javascript/common/
 excerpt: ""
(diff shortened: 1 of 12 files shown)


src/platforms/javascript/common/lazy-load-sentry.mdx
```diff
@@ -1,8 +1,6 @@
 ---
 title: Lazy Loading Sentry
 sidebar_order: 11
-redirect_from:
-  - /platforms/javascript/common/
```
rhcarvalho, line 5: https://docs.sentry.io/platforms/javascript/common/ should not redirect to https://docs.sentry.io/platforms/javascript/lazy-load-sentry/

Similar for the other redirects.

PR #2254, opened 2020-09-11
chore(VSCode): Update extension recommendations

lobsterkatie:
We don't use Liquid anymore, but we do use mdx.


PeloWriter (approved):
Thank you.

PR #2255, opened 2020-09-11
feat(relay): Add a section with best practices for external relays

tonyo:
Here's how it looks like:

image

### DIFF
--- a/src/docs/product/relay/best-practices.mdx
+++ b/src/docs/product/relay/best-practices.mdx
@@ -1,0 +1,63 @@
+---
+title: Self-hosting - Best Practices
+sidebar_order: 5
+---
+
+This page contains some advices and best practices when self-hosting external Relays, i.e. Relays that run on your hardware and forward events to `sentry.io`.
+
+## General Considerations
+
+* While Relay can be built from source, the recommended way to run Relay is via the officially provided Docker image (`getsentry/relay`), which can be found on [DockerHub]. The images are tagged with their Git revision identifier.
+
+* We recommend to run at least two Relay instances (containers) with a reverse proxy (such as [HAProxy] or [Nginx]) in front of them for improved availability and to simplify Relay updates.
+
+* To monitor your Relay setup, you can configure [Logging] and [Metrics].
+
+## System Requirements and Recommendations
+
+- CPU
+    - x86-64 (amd64) CPU architecture
+    - Relay is a multi-threaded application that tries to leverage all available CPU cores, thus it is highly recommended to run Relay on multi-core CPUs. If your setup is expected to handle more than 100 requests per second, it is recommended to run Relay on at least 4 CPU cores.
+        - By default, every Relay instance will use the total number of available cores to adjust the sizes of its thread pools. `limits.max_thread_count` configuration option can be set to adjust this behavior.
+- Memory
+    - At least 2GB of RAM per Relay container is recommended.
+- Storage
+    - Relay does not need any disk storage at the moment.
+
+## Example Configuration
+
+```yaml
+---
+relay:
+  upstream: https://o<YOUR_ORG_ID>.ingest.sentry.io
+  host: 0.0.0.0
+  port: 3000
+logging:
+  level: info
+  format: json
+metrics:
+  statsd: 127.0.0.1:8126
+  prefix: relay
+limits:
+  # Base size of various internal thread pools. Defaults to the number of logical CPU cores
+  max_thread_count: 8
+  # Number of concurrent upstream requests for event submission
+  max_concurrent_requests: 500
+cache:
+  # How many events relay buffers before it starts dropping new events
+  event_buffer_size: 10000
+  # How long a project configuration can still be used after having expired (in seconds).
+  # This might help when the upstream is unreachable (e.g., in case of network issues).
+  project_grace_period: 3600
+  # The cache timeout for buffered events before dropping them (in seconds)
+  event_expiry: 1800
+```
+
+See the [Configuration Options] page for detailed descriptions of all available options.
+
+[metrics]: /product/relay/metrics/
+[logging]: /product/relay/#logging-and-healthcheck
+[dockerhub]: https://hub.docker.com/r/getsentry/relay/
+[haproxy]: https://www.haproxy.org/
+[nginx]: https://nginx.org
+[configuration options]: /product/relay/options/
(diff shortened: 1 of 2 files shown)


src/docs/product/relay/best-practices.mdx
```diff
@@ -0,0 +1,65 @@
+---
+title: Self-hosting - Best Practices
+sidebar_order: 5
+---
+
+This page contains some advices and best practices when self-hosting external Relays, i.e. Relays that run on your hardware and forward events to `sentry.io`.
```
PeloWriter, line 6: please don't use Latin terms in Docs 
```suggestion
This page contains suggested best practices when self-hosting external Relays, that is, Relays that run on your hardware and forward events to `sentry.io`.
```

src/docs/product/relay/best-practices.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+## General Considerations
+
+* While Relay can be built from source, the recommended way to run Relay is via the officially provided Docker image (`getsentry/relay`), which can be found on [DockerHub]. The images are tagged with their Git revision identifier.
```
PeloWriter, line 10:
```suggestion
* We recommend running Relay using the officially provided Docker image (`getsentry/relay`) found on [DockerHub] and tagged with its Git revision identifier, rather than building from source.
```

src/docs/product/relay/best-practices.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+* We recommend to run at least two Relay instances (containers) with a reverse proxy (such as [HAProxy] or [Nginx]) in front of them for improved availability and to simplify Relay updates.
```
PeloWriter, line 12:
```suggestion
* We recommend running at least two Relay instances (containers) with a reverse proxy (such as [HAProxy] or [Nginx]) in front of them for both improved availability and simplified Relay updates.
```

src/docs/product/relay/best-practices.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+* To monitor your Relay setup, you can configure [Logging] and [Metrics].
+
+## System Requirements and Recommendations
```
PeloWriter, line 16: I would build this out at a table, with the first column CPU, Memory, and Storage, and the second column our recommendation.

src/docs/product/relay/best-practices.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+- CPU
+    - x86-64 (amd64) CPU architecture
+    - Relay is a multi-threaded application that tries to leverage all available CPU cores, thus it is highly recommended to run Relay on multi-core CPUs. If your setup is expected to handle more than 100 requests per second, it is recommended to run Relay on at least 4 CPU cores.
```
PeloWriter, line 20:
```suggestion
    - Relay is a multi-threaded application that tries to leverage all available CPU cores. As a result, Sentry highly recommends running Relay on multi-core CPUs. If your set up is expected to handle more than 100 requests per second, we recommend running running Relay on at least four (4) CPU cores.
```

src/docs/product/relay/best-practices.mdx
```diff
@@ -0,0 +1,65 @@
...
+        - By default, every Relay instance will use the total number of available cores to adjust the sizes of its thread pools. `limits.max_thread_count` configuration option can be set to adjust this behavior.
```
PeloWriter, line 21:
```suggestion
        - By default, every Relay instance will use the total number of available cores to adjust the sizes of its thread pools. Adjust this behavior by configuring the `limits.max_thread_count`.

File: src/docs/product/relay/best-practices.mdx [hunk 7]
```diff
@@ -0,0 +1,65 @@
...
+- Memory
+    - At least 2GB of RAM per Relay container is recommended.
+- Storage
+    - Relay does not need any disk storage at the moment.
```
PeloWriter, line 25:
```suggestion
    - Relay does not currently require disk storage.
```

src/docs/product/relay/best-practices.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+## Example Configuration
+
+This example configuration contains a few important values that can be tweaked to change Relay's concurrency settings and buffering behavior.
```
PeloWriter, line 29:
```suggestion
This following configuration example contains values that you can adjust to change Relay's concurrency settings and buffering behavior.
```

PeloWriter (approved):
Thank you.

src/docs/product/relay/best-practices.mdx
jan-auer: Agreed. We can also have multiple columns for what we would consider low, medium, and high load profiles. For instance <100rps, <1000rps, >1000rps. All detail information, for instance the description of multi-threading can move into footnotes.

Things to consider adding in addition to CPU cores:
- clock speed
- instances
- network speed

What if we also add some of the example config values you mention below to this list?
- max_thread_count
- max_concurrent_requests
- event_buffer_size
- http.encoding

src/docs/product/relay/best-practices.mdx
```diff
@@ -0,0 +1,65 @@
...
+  # Number of concurrent upstream requests for event submission
+  max_concurrent_requests: 500
+cache:
+  # How many events relay buffers before it starts dropping new events
+  event_buffer_size: 10000
+  # How long a project configuration can still be used after having expired (in seconds).
+  # This might help when the upstream is unreachable (e.g., in case of network issues).
+  project_grace_period: 3600
+  # The cache timeout for buffered events before dropping them (in seconds)
+  event_expiry: 1800
```
jan-auer, line 55: _nit_: Should we maybe remove them here and ensure we have sensible defaults? While the event buffer size totally depends on load (and ideally can be inferred in future), customers' expectation is probably that grace periods and expiry just work out-of-the-box. WDYT?

src/docs/product/relay/best-practices.mdx
tonyo: Not super sure that we want network/clock speeds here, for various reasons.
Network: do we just say "have enough bandwidth"?
Clock speeds, CPU types, cache values: those vary too much and there's no linear correlation to make it useful.

tonyo:
Thanks for the suggestions 👍 
Applied the proposed ones; let me merge it now, and I'll work on bigger ones in a follow-up PR.

src/docs/product/relay/best-practices.mdx
tonyo: I'm +1000 for having sensible defaults that we don't have to tweak for every new event pattern.

jan-auer: In that case you can remove. The current defaults are sensible for external Relays. It's only our PoPs and internal Relays that need overriding.

PR #2260, opened 2020-09-11
feat: Convert to typescript

NisanthanNanthakumar:
## Objective
Some files for to generate the OpenAPI docs were written in Javascript for quick prototyping. Now that everything is pretty much finalized, we can convert to Typescript. The scope of the migration includes:
- the OpenAPI to GraghQL gatsby plugin
- the React page templates


src/gatsby/plugins/gatsby-plugin-openapi/types.ts
```diff
@@ -0,0 +1,89 @@
...
+            };
+          };
+        };
+      };
+    };
+  };
+};
+
+export type ResponseContent = {
+  content_type: string;
```
scefali, line 54: @NisanthanNanthakumar from the code, it looked like `content-type` should have a dash and not an underscore.

src/templates/developmentApi.tsx
```diff
@@ -40,8 +45,9 @@ const strFormat = str => {
 };
 
 export default props => {
-  const data = props.data?.openApi?.path || {};
-  const parameters =
+  const openApi: OpenAPI = props.data?.openApi;
```
scefali, line 48: @NisanthanNanthakumar shouldn't the type here be `OpenAPI | undefined` since data might not have the `openApi` field?

src/gatsby/plugins/gatsby-plugin-openapi/types.ts
NisanthanNanthakumar: @scefali 👍 apparently GraphQL fields cannot have hyphens, so I'll covert & keep it with the underscore.

mgaeta (approved):

PR #2262, opened 2020-09-11
ref(serverless): Improve serverless wizard slugs

EvanPurkhiser:
As far as I can tell, there's no need for this prefix like this. I'll update in the product as well


untitaker (approved):

ajjindal (commented):
@EvanPurkhiser appending at the end may get cut off during display on UI. I liked how you had it prefixed before. However, if the UI shows well, we are good

EvanPurkhiser:
I'll fix in the UI

EvanPurkhiser:
This is a bit more consistent with what we had done before.

PR #2271, opened 2020-09-14
fix: Should be able to render despite missing data fields

NisanthanNanthakumar:
## Objective
Some pages was not rendering properly on docs.sentry.io/development-api/. This is because we assumed some data fields would be present, so in this PR we just made them conditionally render on the existence of those fields.

Also this PR will deploy the latest SHA of `getsentry/sentry-api-schema`

### DIFF
--- a/src/templates/developmentAPI.js
+++ b/src/templates/developmentAPI.js
@@ -41,7 +41,7 @@ const strFormat = str => {
 
 export default props => {
   const data = props.data?.openApi?.path || {};
-  const parameters =
+  const bodyParameters =
     (data.requestBody?.content?.schema &&
       JSON.parse(data.requestBody.content.schema)) ||
     null;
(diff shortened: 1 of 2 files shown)


manuzope:
Which fields were these?

NisanthanNanthakumar:
> Which fields were these?

`data.parameters ` and `bodyParameters.required`.

src/templates/developmentAPI.js
```diff
@@ -82,35 +82,48 @@ export default props => {
               <p>{data.description}</p>
             </div>
           )}
-          {!!data.parameters.filter(param => param.in === "path").length && (
```
mgaeta, line 85: I think it'd be clearer to define:
```
const pathParameters = (data.parameters || []).filter(param => param.in === "path");
const queryParameters = (data.parameters || []).filter(param => param.in === "query");
```
at the top.

mgaeta (changes requested):

mgaeta (approved):
Looks great 👍

PR #2272, opened 2020-09-14
Remove stale sentry/apm package content

PeloWriter:
The APM package was renamed to sentry/tracing at GA of performance monitoring. The content here is swiftly becoming stale, thus is confusing - sentry/tracing is the correct package name. 

We've preserved this content in the internal support wiki.


rhcarvalho:
@PeloWriter could you please add a redirect from the removed page pointing to the main JS contents? That would save people following old links from hitting a 404.

PR #2275, opened 2020-09-15
feat: Internal docs

dcramer:


src/docs/internal/pages/components.mdx
```diff
@@ -0,0 +1,176 @@
...
+}
+```
+
+```python
+def foo():
+  return 'bar'
+```
+````
+
+Some times you may not want this behavior. To solve this you can either break up the code blocks with some additional text, or you can use the `<Break />` component.
```
PeloWriter, line 78:
```suggestion
Sometimes you may not want this behavior. To solve this, you can either break up the code blocks with some additional text, or use the `<Break />` component.
```

src/docs/internal/pages/components.mdx
```diff
@@ -0,0 +1,176 @@
...
+
+- When the current platform comes from the page context and no matching include is found, the content will be hidden.
+
+- When the current platform comes from the page context path (not the querystring) the platform selector dropdown will be hidden.
+
+- Similar to `PlatformSection`, you can embed content in the block which will render _before_ the given include, but only when an include is available.
+
+- A file named `_default` will be used if no other content matches.
+
+Note: This currently causes issues with tableOfContents generation, so its recommended to disable the TOC when using it.
```
PeloWriter, line 144:
```suggestion
Note: This currently causes issues with tableOfContents generation, so we recommend disabling the TOC when using it.
```

src/docs/internal/pages/components.mdx
```diff
@@ -0,0 +1,176 @@
...
+</PlatformSection>
+```
+
+Attributes:
+
+- `platform` (string) - defaults to the `platform` value from the page context or querystring
+- `supported` (string[])
+- `notSupported` (string[])
+
+Note: This currently causes issues with tableOfContents generation, so its recommended to disable the TOC when using it.
```
PeloWriter, line 166:
```suggestion
Note: This currently causes issues with tableOfContents generation, so we recommend disabling the TOC when using it.
```

src/docs/internal/pages/components.mdx
```diff
@@ -0,0 +1,176 @@
...
+
+## PlatformLink
+
+Useful for linking to platform-specific content when there's not a specific platform already selected.
+
+```markdown {tabTitle:Example}
+<PlatformLink to="/enriching-error-data/" />
+```
+
+This will direct them to a page where they can choose the platform, and then to the appropriate link. If they're within a page that already has an active platform, it will simply link to the appropriate page and skip the redirect.
```
PeloWriter, line 176:
```suggestion
This will direct users to a page where they can choose the platform, and then to the appropriate link. If they're within a page that already has an active platform, it will simply link to the appropriate page and skip the redirect.
```

src/docs/internal/pages/frontmatter.mdx
```diff
@@ -0,0 +1,49 @@
+---
+title: Frontmatter
+noindex: true
+---
+
+Frontmatter a YAML-formatted blob defined at the top of every markdown or mdx file:
```
PeloWriter, line 6:
```suggestion
Frontmatter is a YAML-formatted blob defined at the top of every markdown or mdx file:
```

src/docs/internal/platforms/index.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+This is generated by inheriting all content with the `common/` directory, then adding (or overriding with) content from the siblings (or children as we'd call them). In the above example, you'll see `aSharedPage` is loaded at two different URLs, and `childTwo` is overwritten by `guideName`.
+
+The sidebar on platform pages (handled by `<PlatformSidebar>`) will generate with the Guide, or the Base Platform being the primary section, in addition to a list of other guides available in a section below it. This means that all content is focused on the current guide (usually a framework) they're in, ensuring ease of navigation.
+
+## Globally Shared Content
+
+In addition to platform-shared content (via `common/`) you can also defined globally shared content (shared by all platforms and guides). This is done by placing the content into the top-level `/platforms/common/` path. It works very much the same as the platform-level common content.
+
+## Platform-specific components
```
PeloWriter, line 46:
```suggestion
## Platform-Specific Components
```

src/docs/internal/platforms/index.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+There are a number of components which aid the development of platform specific pages, as well as behave differently when a platform is active. For more information, see [components](../components/).
```
PeloWriter, line 48:
```suggestion
There are a number of components that aid the development of platform specific pages, as well as behave differently when a platform is active. For more information, see [components](../components/).
```

src/docs/internal/platforms/index.mdx
```diff
@@ -0,0 +1,48 @@
+---
+title: Platform Portals
+noindex: true
+---
+
+Platform content lives in `src/platforms` and follows some specific rules to generate content. The content is rendered using the same features as all other [pages](../), but has some additional functionality exposed to clone content and encapsulate platforms as portals.
+
```
PeloWriter, line 7: Perhaps add in:

Platform content is hierarchical. When common content is present, it will be used only if the platform or the child of a primary platform has not provided an override. In sum, guide content is used when present. If not present, platform/common content is used. If not present, /common content is used.

PeloWriter (approved):
thank you.

src/docs/internal/platforms/index.mdx
dcramer: imo this probably needs a larger follow up, as i think adding it here isnt in depth enough

PR #2277, opened 2020-09-15
Update: Make Integrations a top tier section in Main Nav

MimiDumpling:
This PR is now superseded by https://github.com/getsentry/sentry-docs/pull/2294

This PR put Integrations at the same level as Product in the docs main nav, condenses the integrations under /third-party-integrations, and highlights the Integration Platform more prominently.


dcramer:
Why dont we avoid changing URLs. A lot of the mess we're in is having infinite redirects because we often change URLs without considering the longer term consequences. We can restructure content and the sidebar and keep them all in the same palce.

src/components/sidebar.tsx
```diff
@@ -46,6 +46,7 @@ export const Sidebar = ({ data }: ChildProps): JSX.Element => {
         tree={tree}
         exclude={["/product/relay/"]}
       />
+      <DynamicNav root="integrations" title="Integrations" tree={tree} />
```
dcramer, line 49: so you can do root=product/integrations, and then in the product one you can use exclude (see platformSidebar.tsx for an example)

```
exclude={[`/product/integrations/`]}
```

MimiDumpling: Nice! I'll make that change today.

dcramer:
So generally same feedback as before. Why are we changing all the links? Every time we do this we break SEO ranking, we create more 404s (almost guaranteed) and we ultimately end up just doing it again a few months later.

scefali (commented):
@MimiDumpling I think you need to rebase this PR with the Microsoft Teams changes that were just merged and then move those files to the new location

MimiDumpling:
> So generally same feedback as before. Why are we changing all the links? Every time we do this we break SEO ranking, we create more 404s (almost guaranteed) and we ultimately end up just doing it again a few months later.

The original goal was to highlight the Integration Platform more. So, all the third party integrations went into a separate directory. In the main nav sidebar, this reduces visual "noise" and highlights the Integration Platform. Is there a way to preserve the links _and_ keep the integrations under a new directory? Maybe another change in sidebar.tsx?

PR #2284, opened 2020-09-16
Docs: Fix Ruby `current_environment` config name

cooperka:
Relevant URL: https://docs.sentry.io/platforms/ruby/guides/rails/configuration/environments/

`environment` isn't a valid config and causes the app to crash; I looked into the code and it seems like it's called `current_environment`.

### DIFF
--- a/src/includes/set-environment/ruby.mdx
+++ b/src/includes/set-environment/ruby.mdx
@@ -1,5 +1,5 @@
 ```ruby
 Raven.configure do |config|
-  config.environment = 'production'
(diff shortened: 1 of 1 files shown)


dashed:
@st0012 Can you confirm if this update is correct?

st0012 (approved):
yeah this is correct. good catch 👍

dashed (approved):
Thanks!

dashed:
@st0012 Thanks for confirming 👍

PR #2283, opened 2020-09-15
feat: Refactor Cocoa into Apple

dcramer:
- cocoa => apple
- add guides for ios, macos, watchos, tvos
- add various setup snippets
- merge swift/objc code
- remove old version instructions
- suggest session tracking by default

PR #2286, opened 2020-09-16
feat: Add code snippets for Apple

philipphofmann:
Add missing code snippets for Cocoa, like for example breadcrumbs, before-send, etc.

This PR is going to have conflicts with https://github.com/getsentry/sentry-docs/pull/2283. So we need to decide first if we are going to merge this change before we merge this PR.


dcramer (approved):
:100: I just merged the other related change so will need rebased/nouns udpated

PR #2288, opened 2020-09-16
typo apple docs

bruno-garcia:


src/platforms/apple/guides/ios/index.mdx
```diff
@@ -8,7 +8,7 @@ categories:
 
 <PlatformContent includePath="getting-started-install" />
 
-For other installation methods, please see our [documentation](https://docs.sentry.ip/platforms/apple/install/).
+For other installation methods, please see our [documentation](https://docs.sentry.io/platforms/apple/install/).
```
rhcarvalho, line 11:
```suggestion
For other installation methods, please see our [documentation](/platforms/apple/install/).
```

We don't normally include the domain on internal links.

Among other things, makes it easier to verify when PR branch is deployed to vercel.

dcramer:
yeah i dont know why this had the domain in it. might have already been there or i added it for some really dumb reason

src/platforms/apple/guides/ios/index.mdx
dcramer: also you should just use "install/" as the path here as it's relative

dcramer:
Could we fix this in all of them? There's a number of duplicated versions of this. I reverted my change that just fixed the domain.

PR #2291, opened 2020-09-16
feat: index keywords

cameronmcefee:
this will let us add searchable terms to docs that don't appear within the content


Robi-Maker:
Punda

Robi-Maker:
Pukimak ngkau

PR #2293, opened 2020-09-16
feat: Clarify pagination doc page

NisanthanNanthakumar:
## Objective

I found that the first two lines added confusion about pagination, so I removed them. I also reworded the last line of the paragraph.


MimiDumpling (approved):
🚀 LGTM

PR #2305, opened 2020-09-18
feat: Enable webpack plugin and update doc

dcramer:


dcramer:
will need to install deps yet - im using yarn link locally right now to test

PR #2310, opened 2020-09-21
feat: Move platform configuration into generic util

dcramer:
- Shares platform configuration between linter (ts-node) and Gatsby
- Adds support for `config.yml` to allow templatizing the `index.mdx` (future work)
- Default page title from configuration
- Move apple to using shared index
- Turn on linter (wizard is now a warning)


dcramer:
I updated this to remove the duplicate pages in Apple to test the new config behavior.

PR #2313, opened 2020-09-21
Add: note about stack frame order

MimiDumpling:
We needed a small note about how the stack frame order lists the oldest frame to the most recent.


rhcarvalho (approved):
Contents lgtm. This can be a paragraph (as it is now) or make use of <Note> or <Alert> as we have elsewhere, affecting how the note is rendered.

MimiDumpling:
> Contents lgtm. This can be a paragraph (as it is now) or make use of or as we have elsewhere, affecting how the note is rendered.

Nice catch. I updated the Note for consistent styling.

PR #2319, opened 2020-09-22
fix JS syntax for IE 11 & Safari < 8 on beforeSend examples

irvin:


HazAT:
Thanks for the PR. 
I am still going to close it because this would require that we update all of our code samples to be compatible with es5? or this <es5?

For now we keep this syntax

irvin:
to be compatible with ES5 is something I believe most of your customer will need on JS. IE still had enough marketshare and when people implement Sentry, they will especially want to log the error from those edge case such as older browser. it's different than backend codes, as a front-end engineer we can't just ignore our users with legacy devices.

PR #2322, opened 2020-09-22
relay: Refactor best-practices page

tonyo:
* Turn the list into a table
* Add "Network" section
* Update Docker instructions
* A few other minor fixes

The first PR: #2255 

---
Preview:

image

### DIFF
--- a/src/docs/product/relay/best-practices.mdx
+++ b/src/docs/product/relay/best-practices.mdx
@@ -10,31 +10,93 @@ This page reviews our guidelines for operation when self-hosting external Relays
 
 ## General Considerations
 
-- We recommend running Relay using the officially provided Docker image (`getsentry/relay`) found on [DockerHub](https://hub.docker.com/r/getsentry/relay/) and tagged with its Git revision identifier, rather than building from source.
+- We recommend [running Relay](/product/relay/getting-started/#running-relay-in-docker) using the officially provided Docker image (`getsentry/relay`) found on [DockerHub](https://hub.docker.com/r/getsentry/relay/) and tagged with its Git revision identifier, rather than building from source.
 
 - We recommend running at least two Relay instances (containers) with a reverse proxy (such as [HAProxy](https://www.haproxy.org/) or [Nginx](https://nginx.org)) in front of them for both improved availability and simplified Relay updates.
 
-- To monitor your Relay setup, configure both [Logging](/product/relay/#logging-and-healthcheck) and [Metrics](/product/relay/metrics/).
+- To monitor your Relay setup, configure [Logging](/product/relay/#logging-and-healthcheck), [Metrics](/product/relay/metrics/), and [Health Checks](/product/relay/getting-started/#health-checks).
 
 ## System Requirements and Recommendations
 
-- CPU
-  - x86-64 (amd64) CPU architecture
-  - Relay is a multi-threaded application that tries to leverage all available CPU cores. As a result, Sentry highly recommends running Relay on multi-core CPUs. If your set up is expected to handle more than 100 requests per second, we recommend running running Relay on at least four (4) CPU cores.
-    - By default, every Relay instance will use the total number of available cores to adjust the sizes of its thread pools. Adjust this behavior by configuring the `limits.max_thread_count`.
-- Memory
-  - At least 2GB of RAM per Relay container is recommended.
-- Storage
-  - Relay does not currently require disk storage.
+<table>
+  <tbody>
+  <tr>
+    <th>
+      Resource
+    </th>
+    <th>
+      Recommendations
+    </th>
+  </tr>
+  <tr>
+    <td>
+      CPU
+    </td>
+    <td>
+      <ul>
+        <li>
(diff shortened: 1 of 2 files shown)


src/docs/meta/relay/best-practices.mdx
```diff
@@ -46,21 +108,15 @@ limits:
   max_thread_count: 8
   # Number of concurrent upstream requests for event submission
   max_concurrent_requests: 500
```
tonyo, line 110: @jan-auer Do you think we should `max_concurrent_requests` too?

tonyo:
@jan-auer Let's not add more things here before getting at least any (even internal) feedback.

src/docs/meta/relay/best-practices.mdx
```diff
@@ -32,8 +92,10 @@ This example configuration contains values you can adjust to change both Relay's
 ```yaml
 ---
 relay:
-  upstream: https://o<YOUR_ORG_ID>.ingest.sentry.io
+  upstream: https://sentry.io.
```
jan-auer, line 95: What if we leave this on the ingest domain and instead describe where to get this? I'm thinking we can add this URL to the Relay settings page. cc @matejminar 

Otherwise, Relays circumvent org domain protection.

src/docs/meta/relay/best-practices.mdx
jan-auer: Do you mean `max_concurrent_queries`? I'd say no, since Relays will only run for a single organization, so the number of projects will stay in reasonable bounds. This is in contrast to the event volume, which varies greatly from organization to organization, so it is worth thinking about this.

src/docs/meta/relay/best-practices.mdx
```diff
@@ -46,21 +108,15 @@ limits:
...
   # Number of concurrent upstream requests for event submission
   max_concurrent_requests: 500
-cache:
-  # How many events relay buffers before it starts dropping new events
-  event_buffer_size: 10000
-  # How long a project configuration can still be used after having expired (in seconds).
-  # This might help when the upstream is unreachable (e.g., in case of network issues).
-  project_grace_period: 3600
-  # The cache timeout for buffered events before dropping them (in seconds)
-  event_expiry: 1800
```
jan-auer, line 56: I'm on board with removing these. However, I think we should call out these two:

1. `event_buffer_size`, since this depends on the event volume. Of course, until we have better buffering in place, which can still take some time.
2. `event_expiry`. Even if we have a good default, it can be a conscious business decision to change the value.

src/docs/meta/relay/best-practices.mdx
tonyo: What I meant to ask was, do we want to keep or remove `max_concurrent_requests` in this example?
I think we should remove it: because if we encourage to tweak it, what stops users from setting it to 100000 to get it out of the way

src/docs/meta/relay/best-practices.mdx
tonyo: So does "call out" mean that you want to keep them?
Or move them to a separate section?

jan-auer: What if we mention them in the table above under something like "settings"? And then you can keep them in the example here, for copy paste, but with the default values.

src/docs/meta/relay/getting-started.mdx
```diff
@@ -161,13 +161,13 @@ directory in which it can find the configuration and credentials files
 For example, you can start the latest version of `relay` as follows:
 
 ```shell
-❯ docker run -v $(pwd)/configs/:/work/.relay/ getsentry/relay run
+# Generate configuration
+docker run -it -v $(pwd)/configs/:/work/.relay/ getsentry/relay config init
```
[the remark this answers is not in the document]
tonyo: I find it non-intuitive that we don't provide an instruction to generate configuration when going the Docker route. 

It's much more straightforward to provide both commands together (here or elsewhere), it reduces the getting-started with Docker to a few seconds instead of minutes. 

> Separately, if we remove the ❯, then we should remove it in every example.

Ok, but can it be perhaps made non-selectable? 

> For my education, is -it necessary here?

It's necessary for `config init`, otherwise stdin/tty are not connected properly, and you cannot use the interactive menu.

src/docs/meta/relay/best-practices.mdx
tonyo: Reverted.

src/docs/meta/relay/best-practices.mdx
tonyo: The table above is for our "recommendations", but what I think you want is a description of some common settings that might be relevant for scaling (and again, default values should cover most of the normal cases). I don't think that the table is a right place for them, then.

src/docs/meta/relay/best-practices.mdx
jan-auer: Good point, this makes me think that we need to limit this anyway. I still think it makes sense to leave it there, but I'm following your lead here.

jan-auer (approved):

PR #2325, opened 2020-09-22
user-feedback for web

bruno-garcia:


dcramer:
ah sorry we could make this work but what you actually gotta do is go into the user-feedback mdx page itself and put notSupported: - native

bruno-garcia:
Oh I see. OK, will that cover the platforms such as `log4net` which are just under `dotnet`?

dcramer:
@bruno-garcia yeah so it does inheritance - if you say "dotnet" isnt supported, then itll hide all of dotnet unless you put e.g. dotnet.log4net in supported

bruno-garcia:
@dcramer two platforms under dotnet should have it though: `aspnetcore` and `aspnet`.
Can I override at the specific child and set it's supported?

src/platforms/common/enriching-error-data/user-feedback.mdx
```diff
@@ -4,6 +4,8 @@ sidebar_order: 5
 redirect_from:
   - /learn/user-feedback/
 description: "Learn more about the embeddable JavaScript widget, which requests and collects the user information when an event occurs. Sentry pairs the feedback with the original event, giving you additional insight into issues."
+notSupported: [dotnet, android, cocoa, native]
```
dcramer, line 7: you should confirm this syntax works fwiw.. probably does, ive just only used native yaml lists (e.g. like redirect_from)

also cocoa -> apple

bruno-garcia: The syntax I confirmed by checking this: https://github.com/getsentry/sentry-docs/blob/6e56af9bfe52e97ff8bcb199e6cca508a1a17185/src/platforms/common/configuration/index.mdx#L3

WIll fix `apple`.

bruno-garcia, line 7:
```suggestion
notSupported: [dotnet, android, apple, native]
```

bruno-garcia: ops, quotes

bruno-garcia:
@dcramer works like a charm!

dcramer (approved):

PR #2330, opened 2020-09-22
Move Relay back into product

PeloWriter:
Added redirects from meta as a failsafe as well. Removed frontmatter that prevents the publication in the sidebar.


PeloWriter:
Closing. Will cut a new branch for this.

PR #2340, opened 2020-09-23
feat(analytics): Add Google Analytics back after Gatsby migration

BYK:


src/gatsby/config.ts
```diff
@@ -79,6 +79,22 @@ const getPlugins = () => {
     "gatsby-plugin-sharp",
     "gatsby-plugin-sass",
     "gatsby-plugin-zeit-now",
+    {
+      resolve: "gatsby-plugin-google-gtag",
+      options: {
+        // You can add multiple tracking ids and a pageview event will be fired for all of them.
+        trackingIds: [
+          "UA-30327640-1", // Sentry
+          "UA-30327640-3", // www
```
dcramer, line 88: same question - is www desirable? was that how we had it before?

dcramer (approved):

src/gatsby/config.ts
BYK: This is how it was done before. Happy to remove if we don't need it.

PR #2341, opened 2020-09-23
feat: Support platform tag in redirect_from

dcramer:


vercel.json
```diff
@@ -48,26 +48,6 @@
       "source": "/platforms/(.*)/configuration/install-cdn/",
       "destination": "/platforms/$1/install/cdn/"
     },
-    {
-      "source": "/platforms/(.*)/enriching-error-data/set-level/",
```
dcramer, line 52: pretty sure i had this redirect wrong and it was sitting under additional data, but its a minor page so doubt anyone would complain

dcramer:
Hitting max redirects (afaict) in Vercel. Probably time to come up with a better solution.

PR #2343, opened 2020-09-23
add unsupported to frontmatter for breakpad, crashpad, and minidumps

PeloWriter:
This PR ensures that breakpad, crashpad, and minidumps don't display information that is not pertinent to their audience.

### DIFF
--- a/src/platforms/common/enriching-events/attachments/index.mdx
+++ b/src/platforms/common/enriching-events/attachments/index.mdx
@@ -6,6 +6,10 @@ notoc: true
 supported:
   - native
   - javascript
+notSupported:
+  - native.breakpad
+  - native.crashpad
+  - native.minidumps
 ---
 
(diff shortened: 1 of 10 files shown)


src/platforms/common/enriching-events/attachments/index.mdx
```diff
@@ -6,6 +6,10 @@ notoc: true
 supported:
   - native
   - javascript
+notSupported:
+  - native.breakpad
```
jan-auer, line 10: The attachments page is complicated. We do support attachments for all "variants", but only the root (SDK) can add *custom* attachments. 

@PeloWriter is there a way to exclude just a section rather than the entire document?

jan-auer (commented):
`native.ue4` needs to be excluded from "usage" and all "enriching-events" pages other than "Attachments". See comment below for details.

Separately, is there a way to move UE4's "Advanced Configuration" page under "Enriching Events"?

src/platforms/common/enriching-events/attachments/index.mdx
PeloWriter: With includes, yes. But not with the frontmatter.

PR #2345, opened 2020-09-23
feat: Migrate /development-api/* into /api/*

NisanthanNanthakumar:
## Objective
Replace docs.sentry.io/api/ with docs.sentry.io/development-api/

As apart of this PR, I have deleted outdated files associated with the old `/api` page creation and renamed files with `*developmentApi*` as `*api*` to be more consistent with the new routing paths.

Also fixed a small UI bug where API paths in the request example were cutoff. Fixed with css `nowrap` and implementing a scrollbar.
Screen Shot 2020-09-23 at 4 17 33 PM


mgaeta (approved):
Looks good.

PR #2349, opened 2020-09-24
fix: suggest importing just init rather than all

pathurs:
To ensure proper treeshaking, do not import all from `@sentry/angular`


dcramer:
@HazAT we have this kind of behavior everywhere in docs - do we need to rethink exported APIs so we can improve this? I dont love if we have to tell folks to alias method calls.. when 100% of people will have to do that.

pathurs:
@dcramer That's a good idea. I also realised that I only changed one of the spots, when there are a couple more on that page, and probably many more on other pages.

HazAT:
Thanks for the PR, if we are going to do this we'll do it across all imports.
It doesn't really help with tree-shaking since our SDK is built in a way where it doesn't really can shake away anything.

We will revisit this in future SDK updates.

PR #2356, opened 2020-09-24
ref(docker): More optimized Docker stuff

BYK:
Not sure where this is used but I wanted to fix while I'm working on something related.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
+.cache
 .git
 .github
 .gitignore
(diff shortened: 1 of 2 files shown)


dcramer:
Not used anymore we could delete

dashed (approved):

BYK:
>Not used anymore we could delete

Oh, alright then. I'll need the dockerfile for something else soon so I'll move it instead then.

PR #2368, opened 2020-09-25
feat: Parse backticks as html code elements

NisanthanNanthakumar:
## Objective
All code formatted strings are formatted properly.

Naive parsing function should be O(n). **Can this be improved?**

## UI
*BEFORE*
Screen Shot 2020-09-25 at 1 01 23 PM

*AFTER*
Screen Shot 2020-09-25 at 1 03 05 PM

### DIFF
--- a/src/templates/apiPage.tsx
+++ b/src/templates/apiPage.tsx
@@ -114,7 +114,7 @@ export default props => {
 
           {data.description && (
             <div className="pb-3 content-flush-bottom">
-              <p>{data.description}</p>
+              <p>{parseBackticks(data.description)}</p>
             </div>
           )}
 
--- a/src/templates/apiPage.tsx
+++ b/src/templates/apiPage.tsx
@@ -2,6 +2,7 @@ import React, { useState, useEffect } from "react";
 import { graphql } from "gatsby";
 import Prism from "prismjs";
 
+import {parseBackticks} from "~src/utils";
 import BasePage from "~src/components/basePage";
 import SmartLink from "~src/components/smartLink";
 import ApiSidebar from "~src/components/apiSidebar";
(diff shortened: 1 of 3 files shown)


src/templates/apiPage.tsx
```diff
@@ -13,6 +13,37 @@ import {
 
 import "prismjs/components/prism-json";
 
+const parseBackticks = (str: string) => {
+  let arr = str.split("");
+  let i = 0;
+  let j = arr.length - 1;
+
+  const forward = () => {
+    for (i <= j; i++; ) {
```
mgaeta, line 22:
```
for (; i <= j; i++) {
```

mgaeta (changes requested):
`parseBackticks` needs tests.

mgaeta:
This will fail with:
```
`foo` `foo`
```

mgaeta:
```
const parseBackticks = (str) => {
  let i = 0;
  return str.split("").map((c) => {
    return c === "`" ? (i++ % 2 ? "</code>": "<code>") : c;
  }).join("");
};
```

NisanthanNanthakumar:
> const parseBackticks = (str) => {
>   let i = 0;
>   return str.split("").map((c) => {
>     return c === "`" ? (i++ % 2 ? "</code>": "<code>") : c;
>   }).join("");
> };

lol I didn't realize it would break for consecutive, this a good one!!

manuzope:
Hmmm, I don't know if docs even have tests 🤔

NisanthanNanthakumar:
> Hmmm, I don't know if docs even have tests 🤔

Not alot, but the structure is there

mgaeta (approved):
👍

mgaeta (approved):
Ok this looks good but there is one more edge case: `` ` `` (a code block containing a single backtick). I don't think we'll actually see that in the docs but I think you should add a comment describing why we squash backticks and why this case isn't covered.

PR #2372, opened 2020-09-26
fix: wrong link to context documentation

id0Sch:
additional-data is 404, so I'm guessing these links should lead to the context page?

link to the direct page:
https://sentry-docs-git-fork-id0sch-patch-1.sentry.dev/platforms/javascript/enriching-events/scopes/


untitaker:
Thx I think this has been fixed. We now check for this in CI

PR #2377, opened 2020-09-28
feat: Remove unused code related to nginx

NisanthanNanthakumar:
## Objective
We don't use nginx for redirects. Redirects are handled through: https://docs.sentry.io/contributing/pages/redirects/

So in this PR, unused code related to nginx and its' redirects should be removed.

cc: @BYK since we are not generating a nginx.out.conf file, I removed it from the Dockerfile

Consensus is that the Dockerfile is not necessary, so it will be also removed in this PR.

## Test Plan
Check the Vercel preview build to make sure redirects did not break.


dcramer:
tbqh we should remove the dockerfile and everything else that I forgot to clean up from pre-vercel era

dcramer (approved):

dcramer (changes requested):
wait you cant remove the nginx stuff from gatsby, thats used by vercel

NisanthanNanthakumar:
> wait you cant remove the nginx stuff from gatsby, thats used by vercel

The preview build's redirects seem to be working

dcramer:
its because you're seeing the client side redirects - the nginx.conf thing is a vercel compatible redirect solution as far as I remember. it's possible I'm making that up but i would verify that its actually literally unused (curl the page you should see a 302 not a 200).

NisanthanNanthakumar:
> its because you're seeing the client side redirects - the nginx.conf thing is a vercel compatible redirect solution as far as I remember. it's possible I'm making that up but i would verify that its actually literally unused (curl the page you should see a 302 not a 200).

Screen Shot 2020-09-28 at 2 33 23 PM

@dcramer I checked against the preview build and current live docs and both have the same results.

dcramer:
deferring here as its possible i dont remember correctly

Dockerfile
```diff
@@ -29,7 +29,4 @@ RUN gatsby build
 
 FROM gatsbyjs/gatsby
```
BYK, line 30: We should just remove this file, it is not used by anything.

src/gatsby/config.ts
```diff
@@ -1,9 +1,9 @@
 import axios from "axios";
 
 import queries from "./utils/algolia";
-import PackageRegistry from "./utils/packageRegistry";
+// import PackageRegistry from "./utils/packageRegistry";
```
BYK, line 4: No commented-out code please. Either delete or uncomment.

src/gatsby/config.ts
```diff
@@ -17,7 +17,7 @@ if (process.env.DISABLE_THUMBNAILS === "1") {
 
 const getPlugins = () => {
   const remarkPlugins = [
-    {
+   /* {
```
BYK, line 20: Same as above.

BYK (changes requested):
Thanks for the clean up @NisanthanNanthakumar!

@dcramer don't we use `vercel.json` now for all redirects: https://github.com/getsentry/sentry-docs/blob/master/vercel.json

Rejecting for commented-out code and for the removal of `Dockerfile`, LGTM otherwise.

BYK:
There are some link failures here: https://github.com/getsentry/sentry-docs/runs/1181514785?check_suite_focus=true

Should we be adding more redirects to the `vercel.json` file?

src/gatsby/config.ts
dcramer: guessing this was an accident - this code cannot be removed

NisanthanNanthakumar: whoops! this is my workaround to make gatsby build faster locally. I'm going to undo.

Dockerfile
BYK: Let's remove this file as a whole as w/o Nginx config this is useless.

BYK:
Linkchecker does not fail on master so I think those issues are somehow related to this change. Can you look into it as I don't feel confident with merging in this state?

BYK (changes requested):
See my comment above.

BYK (approved):
Great job, thanks a lot @NisanthanNanthakumar!

PR #2378, opened 2020-09-28
Fix: bolded headers --> H3 or H4

MimiDumpling:
Bolded headers need to be changed to H3 headings. This PR gets started on the fix. However, bolded headers are prevalent throughout Docs. Most likely means multiple PRs over time.


dashed (approved):
I'll allow it.

PR #2382, opened 2020-09-29
fix(urls): Use relative URLs instead of linking to docs.sentry.io

benvinegar:
Replaces any absolute URL in a markdown file that pointed to docs.sentry.io and instead uses a relative URL. This way I can navigate links on a local development build without being transported to the production website.

### DIFF
--- a/src/docs/accounts/quotas/index.mdx
+++ b/src/docs/accounts/quotas/index.mdx
@@ -15,7 +15,7 @@ Sentry’s flexibility means you can exercise fine-grained control over which ev
 Let’s clarify a few terms to start:
 
 - Event - an event is one instance of you sending Sentry data. Generally, this data is an error. Every event has a set of characteristics, called its fingerprint.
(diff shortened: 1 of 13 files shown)


dcramer (approved):
isnt it amazing in 2020 we still have to write code to change these things

mitsuhiko (approved):

benvinegar:
Interestingly, this fails a build because now that the links are relative, it is discovering that some of the document fragment links (#) no longer point anywhere. I guess this check wasn't happening for externally linked content.

https://github.com/getsentry/sentry-docs/pull/2382/checks?check_run_id=1180346161

benvinegar:
> Interestingly, this fails a build because now that the links are relative, it is discovering that some of the document fragment links (#) no longer point anywhere. I guess this check wasn't happening for externally linked content.

Oops, nope, I brought in a 404 because `/api/` isn't available locally.

PR #2385, opened 2020-09-29
fix(apple): CocoaPods Installation Podfile

philipphofmann:
The Podfile for how to install Sentry with CocoaPods displayed the placeholder
for the version instead of the version. This is fixed now and the Podfile is
cleaned up so it only contains the line to add Sentry instead of a template
for a Podfile.

### DIFF
--- a/src/includes/getting-started-install/apple.ios.mdx
+++ b/src/includes/getting-started-install/apple.ios.mdx
@@ -1,12 +1,7 @@
 We recommend installing the SDK with CocoaPods. To integrate Sentry into your Xcode project, specify it in your _Podfile_:
 
 ```ruby {filename:Podfile}
-platform :ios, '8.0'
-use_frameworks! # This is important
-
-target 'YourApp' do
-  pod 'Sentry', :git => 'https://github.com/getsentry/sentry-cocoa.git', :tag => '{{ packages.version('sentry.cocoa') }}'
-end
+pod 'Sentry', '~> {{ packages.version('sentry.cocoa') }}'
 ```
 
 Afterwards run `pod install`.
--- a/src/includes/getting-started-install/apple.macos.mdx
+++ b/src/includes/getting-started-install/apple.macos.mdx
@@ -1,12 +1,7 @@
 We recommend installing the SDK with CocoaPods. To integrate Sentry into your Xcode project, specify it in your _Podfile_:
 
 ```ruby {filename:Podfile}
-platform :macos, '8.0'
-use_frameworks! # This is important
-
-target 'YourApp' do
-  pod 'Sentry', :git => 'https://github.com/getsentry/sentry-cocoa.git', :tag => '{{ packages.version('sentry.cocoa') }}'
-end
+pod 'Sentry', '~> {{ packages.version('sentry.cocoa') }}'
 ```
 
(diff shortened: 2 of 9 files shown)


src/includes/getting-started-install/apple.ios.mdx
```diff
@@ -1,12 +1,7 @@
 We recommend installing the SDK with CocoaPods. To integrate Sentry into your Xcode project, specify it in your _Podfile_:
 
 ```ruby {filename:Podfile}
-platform :ios, '8.0'
-use_frameworks! # This is important
-
-target 'YourApp' do
-  pod 'Sentry', :git => 'https://github.com/getsentry/sentry-cocoa.git', :tag => '{{ packages.version('sentry.cocoa') }}'
-end
+pod 'Sentry', '~> {{ packages.version('sentry.cocoa') }}'
```
bruno-garcia, line 4: I'm unsure about this change. It seems, based on:

```
 use_frameworks! # This is important 
```

That because of folks not adding this, they had issues and we added this here because of that.

@HazAT could you please confirm?

src/includes/getting-started-install/apple.macos.mdx
```diff
@@ -1,12 +1,7 @@
 We recommend installing the SDK with CocoaPods. To integrate Sentry into your Xcode project, specify it in your _Podfile_:
 
 ```ruby {filename:Podfile}
-platform :macos, '8.0'
-use_frameworks! # This is important
-
-target 'YourApp' do
-  pod 'Sentry', :git => 'https://github.com/getsentry/sentry-cocoa.git', :tag => '{{ packages.version('sentry.cocoa') }}'
-end
+pod 'Sentry', '~> {{ packages.version('sentry.cocoa') }}'
```
bruno-garcia, line 4: Also, since we should optimize for beginners, it's good to clarify where the `pod` goes. Given that `target '..' do` is important. When I was trying to add this to Flutter it was helpful since Flutter buries it into their own generated podfile IIRC.

bruno-garcia (changes requested):

src/includes/getting-started-install/apple.ios.mdx
HazAT: Yes, so back at the time you needed if you used Swift.
Also if you don't have this I think the way you import changes, so I would leave it unless you know 100% we don't need it anymore.

HazAT: Also another thing, the reason it was `pod 'Sentry', :git => 'https://github.com/getsentry/sentry-cocoa.git', :tag => '{{ packages.version('sentry.cocoa') }}'` before is that.
At one point Cocoapods said you need to define a source like `source 'https://github.com/CocoaPods/Specs.git'` with having `:git` in there it will fetch it from the repo which is better I think?

So yeah, `use_frameworks!` can be dropped potentially (if it's default already) but may we keep to longer version of installing the Pod like:

```
pod 'Sentry', :git => 'https://github.com/getsentry/sentry-cocoa.git', :tag => '{{ packages.version('sentry.cocoa') }}'
```

philipphofmann:
Thanks for explaining these reasons @HazAT and @bruno-garcia. Then I close this PR and we keep it as it is.

PR #2387, opened 2020-09-29
feat: Switch to fixed navs with scrolling

dcramer:


dcramer:
- need to deal w/ responsive design yet (we are using bootstrap helpers, cant do that for some of this)
- need to fix width of navbar
- need to correct some display issues with right-side TOC

PR #2390, opened 2020-09-29
Fix: broken fragment URLs (anchor tags)

MimiDumpling:
Docs has many broken/dead fragment URLs because an anchored section of a page moved or was renamed.


PeloWriter (approved):
Thank you.

PR #2398, opened 2020-09-30
feat: Add example of how to sample rrweb

dcramer:


src/platforms/javascript/common/integrations/rrweb.mdx
```diff
@@ -48,3 +48,26 @@ For more information on configuration, see the [@sentry/rrweb project on GitHub]
 </markdown></Note>
 
 Once a replay is captured with an event, you'll find it visible within Issue Details under the Replay section of the event.
+
+## Sampling
+
+You may find - due to the size and cost - that you want to sample replays. The easiest way to do this is to make the sampling decision when you initialize the Sentry SDK. For example, here's how Sentry itself uses sampling to only capture these for employees:
```
PeloWriter, line 54:
```suggestion
To suit your organization's needs, you may prefer to sample replays. The easiest way to do this is to make the sampling decision when you initialize the Sentry SDK. For example, here's how Sentry itself uses sampling to capture these for only employees:
```

src/platforms/javascript/common/integrations/rrweb.mdx
```diff
@@ -48,3 +48,26 @@ For more information on configuration, see the [@sentry/rrweb project on GitHub]
...
+
+Sentry.init({
+  dsn: "___PUBLIC_DSN___",
+  integrations,
+});
+
+Sentry.setTag("rrweb.active", hasReplays ? "yes" : "no");
+```
+
+You'll note that we also set the `rrweb.active` tag here. That helps us identify events which have a replay attached, because otherwise we'd not be able to find them. Once configured, you'll be able simply use `rrweb.active:yes` in your search queries.
```
PeloWriter, line 73:
```suggestion
You'll note we also set the `rrweb.active` tag, which helps us identify events that have a replay attached, since otherwise we'd not be able to find them. Once configured, you'll be able simply use `rrweb.active:yes` in your search queries.
```

PeloWriter (approved):
Minor smoothing.

PR #2399, opened 2020-09-30
feat: Improve Product Guides structure

dcramer:
- Expose dropdown for Product navigation
- Move all product pages into /product/
- Re-order sidebar to push relevant (vs marketing/discovery) docs to the top


dcramer:
I subjectively pushed integrations to the bottom (its huge, not super relevant most of the time, searchable), and moved CLI/Relay up.

PR #2401, opened 2020-09-30
fix(links): Move performance sampling page to common

lobsterkatie:
Switch to using a `PlatformLink`, and move the sampling page to `common` to fix a broken link on the filtering page.

Now that the file is common, more parts will need to be "platform-ized" but this should fix the broken links for now and that can come in a follow-up PR.


PeloWriter (approved):
thank you

PR #2403, opened 2020-10-01
fix(performance): Pull platform-specific content into includes

lobsterkatie:
Follow up to https://github.com/getsentry/sentry-docs/pull/2401.

Now that the sampling page is in `/common`, anything language-specific needs to be pulled out into includes so it can be swapped out.


PeloWriter (approved):
Thank you

PR #2404, opened 2020-10-01
Update index.mdx

octomonoli1:
Error binding javascript breadcrumbs instead of SpringBoot


HazAT:
Thank you, this has already been resolved

PR #2415, opened 2020-10-01
feat: Hide framework list from guides

dcramer:
Introduce `noGuides` attribute on `PlatformContent` to enable hiding content from guide pages.


dcramer:
If someone has an alternative name that means "only base platforms" thats short I'm open to ideas. Want to keep it concise so its easy to understand what it means.

PR #2419, opened 2020-10-01
ref: Cleanup various old urls

dcramer:


dcramer:
failing test is unrelated

PR #2425, opened 2020-10-01
ref: Move resolve function and implement support for local OpenAPI paths

NisanthanNanthakumar:
## Objective
We want to work locally with API docs so that people can make changes to the OpenAPI spec in Sentry but also be able to see what it would look like in docs.sentry.io/api/.

In this PR, I have implemented support for developers to set an environment variable `OPENAPI_LOCAL_PATH` that can be lead to their locally generated `openapi-derefed.json`. 

If it is able to resolve the path, `gatsby-plugin-openapi` will generate its' graphql nodes from the local `openapi-derefed.json`, else it defaults to fetching the `openapi-derefed.json` from the versioned commit in `getsentry/sentry-api-schema`.

ex: `OPENAPI_LOCAL_PATH=/Users/nisanthan.nanthankumar/Documents/sentry-api-schema/openapi-derefed.json yarn start`

## Future Improvements (Not this PR)
 - Support Hot-Reload. Maybe we can watch the file from the env variable. **Need to look into this**

### DIFF
--- a/src/gatsby/utils/resolveOpenAPI.ts
+++ b/src/gatsby/utils/resolveOpenAPI.ts
@@ -5,7 +5,7 @@ const activeEnv =
   process.env.GATSBY_ENV || process.env.NODE_ENV || "development";
 
 export default async () => {
-  if (activeEnv && process.env.OPENAPI_LOCAL_PATH) {
+  if (activeEnv === "development" && process.env.OPENAPI_LOCAL_PATH) {
     try {
       console.log(`Fetching from ${process.env.OPENAPI_LOCAL_PATH}`);
       let data = await fs.readFile(process.env.OPENAPI_LOCAL_PATH, "utf8");
--- a/src/gatsby/utils/resolveOpenAPI.ts
+++ b/src/gatsby/utils/resolveOpenAPI.ts
@@ -1,0 +1,24 @@
+import axios from "axios";
+import { promises as fs } from "fs";
+
+const activeEnv =
+  process.env.GATSBY_ENV || process.env.NODE_ENV || "development";
+
+export default async () => {
+  if (activeEnv && process.env.OPENAPI_LOCAL_PATH) {
+    try {
+      console.log(`Fetching from ${process.env.OPENAPI_LOCAL_PATH}`);
+      let data = await fs.readFile(process.env.OPENAPI_LOCAL_PATH, "utf8");
+      return data;
+    } catch (error) {
+      console.log(
+        `Failed to connect to  ${process.env.OPENAPI_LOCAL_PATH}. Continuing to fetch versioned schema from Github.
+        ${error}`
+      );
+    }
+  }
+  const response = await axios.get(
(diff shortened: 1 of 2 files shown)


ceorourke (commented):
This is great! I just tried it out with my path to `sentry-api-schema` and then in `sentry`. Hot reload would definitely be a cool future addition. I doubt you want me reviewing the JS though.

MimiDumpling:
Thanks for updating the API docs! Gonna leave the technical review to the Eco team. 🌻

NisanthanNanthakumar:
> This is great! I just tried it out with my path to `sentry-api-schema` and then in `sentry`. Hot reload would definitely be a cool future addition. I doubt you want me reviewing the JS though.

@ceorourke Haha awesome! But you should definitely review the JS!!

src/gatsby/utils/resolveOpenAPI.ts
```diff
@@ -0,0 +1,24 @@
...
+      console.log(`Fetching from ${process.env.OPENAPI_LOCAL_PATH}`);
+      let data = await fs.readFile(process.env.OPENAPI_LOCAL_PATH, "utf8");
+      return data;
+    } catch (error) {
+      console.log(
+        `Failed to connect to  ${process.env.OPENAPI_LOCAL_PATH}. Continuing to fetch versioned schema from Github.
+        ${error}`
+      );
+    }
+  }
```
MeredithAnya, line 19: I think if you wanted you could add another `console.log` indicating that you are fetching from the `sentry-api-schema`. I forgot to put `export` in front of the `OPENAPI_LOCAL_PATH` and was confused for a min about why I didn't see the console.log message.

MeredithAnya (approved):
Had one small comment. But looks good. 

I think that we shouldn't really be using the local `sentry-api-schema` repo unless we are saying that the flow to changing and testing out schema changes is to have the sentry command `build-derefed-docs` output the derefed to your local version of that repo, right? But seems like people shouldn't have to have 3 repos to test out the docs. 

So I think the next step would be to figure out where does the local openapi-derefed.json go.

src/gatsby/utils/resolveOpenAPI.ts
NisanthanNanthakumar: @MeredithAnya hmm should we log defaults? I don't think we currently do that for anything else.

NisanthanNanthakumar:
> Had one small comment. But looks good.
> 
> I think that we shouldn't really be using the local `sentry-api-schema` repo unless we are saying that the flow to changing and testing out schema changes is to have the sentry command `build-derefed-docs` output the derefed to your local version of that repo, right? But seems like people shouldn't have to have 3 repos to test out the docs.
> 
> So I think the next step would be to figure out where does the local openapi-derefed.json go.

I used my `sentry-api-schema` as an example. I expect people to connect to their `sentry` derefed. But its generalized to be flexible for any use case.

manuzope:
@NisanthanNanthakumar can we update the reference here: https://develop.sentry.dev/docs/api/#how-to-make-an-endpoint-public.

Once we're done with developer experience improvements, might also be a good idea to add a blurb on that page.

PR #2426, opened 2020-10-02
fix: ruby environment config

shouichi:
`environment=` does not exist. Use `current_environment=` instead. See
https://github.com/getsentry/sentry-ruby/blob/94138234bdca8bd1cb44b36990b84b0ab5c0ff02/lib/raven/configuration.rb#L24.


dcramer (approved):
thanks!

PR #2427, opened 2020-10-02
feat(apple): Set installation id as user id

philipphofmann:
Add a comment to set-user on Apple about setting the installation id of
the SDK as the userId of an event before queuing it for submission.

Only merge when https://github.com/getsentry/sentry-cocoa/pull/757 is released.

### DIFF
--- a/src/includes/set-user/apple.mdx
+++ b/src/includes/set-user/apple.mdx
@@ -9,3 +9,6 @@ SentryUser *user = [[SentryUser alloc] init];
 user.email = @"john.doe@example.com";
 [SentrySDK setUser:user];
 ```
+
+Since 6.0.1: If no user is set the SDK uses the installation id of the SDK as the
+user id of the event before queuing it for submission.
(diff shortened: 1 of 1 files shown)


src/includes/set-user/apple.mdx
```diff
@@ -9,3 +9,6 @@ SentryUser *user = [[SentryUser alloc] init];
 user.email = @"john.doe@example.com";
 [SentrySDK setUser:user];
 ```
+
+Since 6.0.1: If no user is set the SDK uses the installation id of the SDK as the
```
marandaneto, line 13:
```suggestion
Since 6.0.1: Sentry will by fallback to `installationId` if you do not provide a user identity.
```

marandaneto: so it's similar to https://docs.sentry.io/platforms/android/enriching-events/identify-user/

src/includes/set-user/apple.mdx
```diff
@@ -9,3 +9,6 @@ SentryUser *user = [[SentryUser alloc] init];
...
+user id of the event before queuing it for submission.
```
marandaneto, line 14:
```suggestion
Since 6.0.1: Sentry will by fallback to `installationId` if you do not provide a user identity.
```

PR #2431, opened 2020-10-02
feat: Dynamic TOC

dcramer:


dcramer:
Unfinished - doesn't render on initial load, but seems to work well after that

PR #2432, opened 2020-10-02
feat: Remove duplicate content for Ruby guides

dcramer:


dcramer:
Need to fix common content as it's incorrectly calling the configuration stage "Initialize", as well as referring to `init` which is not present in all SDKs.

dcramer:
made shared terminology change as part of this PR

PeloWriter (approved):
Thanks for spotting the inaccuracy in the common content and fixing it.

PR #2446, opened 2020-10-05
Add missing Java examples.

maciejwalkowiak:


src/includes/set-fingerprint/basic/java.mdx
```diff
@@ -0,0 +1,14 @@
+```java
+import io.sentry.Sentry;
+import java.sql.SQLException;
+import java.util.Arrays;
+
+Sentry.init(options -> {
+    options.setBeforeSend((event, hint) -> {
+        if (event.getThrowable() instanceof SQLException) {
```
maciejwalkowiak, line 8: `SentryEvent#getThrowable` is marked as `@ApiStatus.Internal` but the alternative check for exception type is so cumbersome that I don't know if we should put it as an example:

```java
Sentry.init(options -> {
      options.setDsn("https://key@sentry.io/proj");
      options.setBeforeSend(new SentryOptions.BeforeSendCallback() {
        @Override
        public @Nullable
        SentryEvent execute(@NotNull SentryEvent event, @Nullable Object hint) {
          if (containsException(event, MyException.class)) {
            return null;
          } else {
            return event;
          }
        }

        private <T> boolean containsException(@NotNull SentryEvent event, Class<T> clazz) {
          List<SentryException> exceptions = event.getExceptions();
          return exceptions != null && exceptions.stream()
            .map(SentryException::getType)
            .anyMatch(type -> type.equals(toSentryExceptionType(clazz)));
        }

        private <T> String toSentryExceptionType(Class<T> clazz) {
          return clazz.getName().replace(clazz.getPackage().getName() + ".", "");
        }
      });
    });
```

maciejwalkowiak:
cc @bruno-garcia

src/includes/set-context/java.mdx
```diff
@@ -0,0 +1,9 @@
+```java
+import io.sentry.Sentry;
+
+Sentry.configureScope(scope -> {
+    scope.setContexts("name", "Mighty Fighter");
```
bruno-garcia, line 5: This won't work I'm afraid. A context takes an object, just a string will get Sentry to yell

src/includes/set-fingerprint/basic/java.mdx
bruno-garcia: Agree this is cumbersome.
Exposing the throwable from the event is not ideal either since it's confusing the relationship between SentryException and that. Which is in fact used to create the SentryException during processing.

Ideally we'd expose a hook for that. In .NET we have something called `IExceptionProcessor` which takes the event and the exception as a parameter. 

This way we're not exposing the reference to the exception through the event which is anyway not ideal. In other SDKs that doesn't exist. You can only capture an exception through `captureException` which there and then convert to SentryEvent already with the right data based on the exception.

I don't have a simple answer for that. It's easy to make this public, I'm just not sure we'll regret this later because we didn't think this through.

@marandaneto thoughts?
10 bucks says @mitsuhiko will ignore this @.

marandaneto: what's about offering a public getter for the exception type? like:

`Class<?> getExceptionType()`

so people could:

```
if (MyType.class.isAssignableFrom(event.getExceptionType())) {
... // filter it
}

so we don't make the `getThrowable` public, but rather its type which is enough.
```

src/includes/set-context/java.mdx
marandaneto: it does, I guess, you'd just need to cast it when you do a get for this key

src/includes/set-fingerprint/basic/java.mdx
marandaneto: the other way would be to match Strings like:

```
if ("MyException".equals(event.getExceptions().get(0).getType())) {
... // filter it                    
}
```

but the downside is we'd need to check all the inherited types as well and `instanceof` does it automatically + it's type-safe

src/includes/set-context/java.mdx
maciejwalkowiak: I checked, it works on the SDK level but indeed shown in Sentry as an error. Wondering how is it in other SDKs as this piece of code is a direct translation from Python or dotnet sample.

PR #2448, opened 2020-10-05
Update Java wizard pages to SDK 3.0.

maciejwalkowiak:


maciejwalkowiak:
cc @bruno-garcia

src/wizard/java/index.md
```diff
@@ -13,131 +13,127 @@ Using Maven:
...
 
 ```scala
-libraryDependencies += "io.sentry" % "sentry" % "{{ packages.version('sentry.java', '1.7.30') }}"
+libraryDependencies += "io.sentry" % "sentry" % "{{ packages.version('sentry.java', '3.0.0') }}"
 ```
 
 ## Capture an Error
 
-To report an event manually you need to initialize a `SentryClient`. It is recommended that you use the static API via the `Sentry` class, but you can also construct and manage your own `SentryClient` instance. An example of each style is shown below:
+To report an event manually you need to initialize a `Sentry` class. It is recommended that you use the static API via the `Sentry` class, but you can also construct and manage your own `IHub` instance. An example of each style is shown below:
```
bruno-garcia, line 34: Not sure this change makes sense.
The doc here is trying to say you can do it either with SentryClient or with the static Sentry class. We can't "initialize a Sentry class", it talks about Sentry init in the second phrase.

Maybe we delete the whole sentence?

bruno-garcia (commented):
Lets do this on a call

maciejwalkowiak:
Lets start over again.

PR #2452, opened 2020-10-05
feat: Add support for sidebar_title

dcramer:
Allow specifying sidebar_title in frontmatter to override the value in generated sidebar trees.


lobsterkatie (approved):
Can confirm it works in my testing

PR #2454, opened 2020-10-05
Update releases.mdx

rohitkataria-sentryio:
It appears that forward slash is the only one that isn’t allowed. All others characters can easily make their way into a release name.
Screen Shot 2020-10-05 at 3 52 29 PM

### DIFF
--- a/src/docs/product/cli/releases.mdx
+++ b/src/docs/product/cli/releases.mdx
@@ -15,7 +15,9 @@ Because releases work on projects you will need to specify the organization and 
 
 Releases are created with the `sentry-cli releases new` command. It takes at the very least a version identifier that uniquely identifies the releases. There are a few restrictions -- the release name cannot:
 
-- use a forward slash ("/")
+- be (".") ,  (".."), or forward slash ("/") 
+- be the string ("latest")
+- contain a forward slash ("/")
 - exceed 200 characters
 
 The value can be arbitrary, but for certain platforms, recommendations exist:
--- a/src/docs/product/cli/releases.mdx
+++ b/src/docs/product/cli/releases.mdx
@@ -15,8 +15,7 @@ Because releases work on projects you will need to specify the organization and 
 
 Releases are created with the `sentry-cli releases new` command. It takes at the very least a version identifier that uniquely identifies the releases. There are a few restrictions -- the release name cannot:
 
-- contain newlines or spaces
-- use a forward slash ("/"), back slash ("\"), period ("."), or double period ("..")
+- use a forward slash ("/")
 - exceed 200 characters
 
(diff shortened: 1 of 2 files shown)


MimiDumpling:
Interesting... @matejminar or @bruno-garcia will you verify that forward slash is the only prohibited character?

bruno-garcia:
It might break the parsing for release based on `package@version+build`.
@mitsuhiko ?

matejminar:
The release parser limits the length of the name to 250 chars (not 200).

Restricted release names are `.`, `..`, and `latest`.

These characters make the compiler sad: `/`, `\r`, `\n`.

rohitkataria-sentryio:
@matejminar 

Re release name length, anything over 200 is giving me error when using the sentry-cli to create a release i.e 200 characters work but 201 does not as that gives me:
```
error: API request failed
  caused by: sentry reported an error: request failure (http status: 400)
  Object({"version": Array([String("Ensure this field has no more than 200 characters.")])})
```
Further, the senty-cli does not prohibit if the release name is just `\n` or `just \r` or contains `\n` or contains `\r`

sentry-cli is does block me if the release name is just  `.`  or just `..`  or just  `latest`

src/docs/product/cli/releases.mdx
```diff
@@ -15,8 +15,7 @@ Because releases work on projects you will need to specify the organization and
 
 Releases are created with the `sentry-cli releases new` command. It takes at the very least a version identifier that uniquely identifies the releases. There are a few restrictions -- the release name cannot:
 
-- contain newlines or spaces
-- use a forward slash ("/"), back slash ("\"), period ("."), or double period ("..")
+- use a forward slash ("/")
```
rohitkataria-sentryio, line 18:
```suggestion
- be (".") or ("..") or the word ("latest")
- contain a forward slash ("/")
```

PeloWriter: Please also update the common file for Releases common/configuration/releases.mdx with this clarification.

PeloWriter (commented):
Please also update the information for Releases for the SDKs. The file is in common/configuration/releases.mdx

Thank you.

matejminar:
@rohitkataria-sentryio 

> Re release name length, anything over 200 is giving me error when using the sentry-cli to create a release

Ah, let's keep 200 then 👍

> Further, the senty-cli does not prohibit if the release name is just \n or just \r or contains \n or contains \r

The CLI will create these releases, however, `/`, `\r`, `\n` are listed as invalid characters in the release parser.
The parser will literally error out and it works only thanks to some raw fallbacks.
I think we should list the characters as prohibited and reach out to CLI folks to add validation for these.

PeloWriter:
Fixed in #2856.

PR #2463, opened 2020-10-06
fix(apple): Configuring additional context

philipphofmann:


bruno-garcia (approved):
Good catch, since the old docs would cause a processing error in Sentry

PR #2467, opened 2020-10-06
ref: Minor optimization for PlatformContent

dcramer:
- Remove unused fallbackPlatform
- Remove internal type from queries


dcramer:
Was hoping to completely find a way to kill returning the mdx content via page-data since its literally static here.. but no luck.

PR #2474, opened 2020-10-07
Django updates

PeloWriter:
Refined the code sample and removed preceding text
Moved keyword arguments into Basic Options
Created a new page "Recording Status Codes"


dcramer:
This now includes a larger change that we discussed on Slack which moves integrations into configuration and specifically pulls that Django config snippet into that page as well.

untitaker:
reverting move of android integrations into `/configuration/integrations` because of merge conflicts

PR #2475, opened 2020-10-07
Update index.mdx

oizik:
Without the `traces_sample_rate` explicitly set, there won't be any output in Performance Monitor as the default rate is 0


dashed (approved):
Thanks! 👍

PR #2479, opened 2020-10-07
fix: Typos

rhcarvalho:
Note that we use `pageload` when referring to a specific type of transaction (by
its "op" value), and page load otherwise.


lobsterkatie (approved):
Thanks for catching!

PR #2480, opened 2020-10-07
ref: Move the data scrubbing docs to product location

flub:
None of this is platform-specific, as this is a feature provided by relay, being in the same hierarchy as relay makes sense.  I made it a sibling based on the existing links from sentry itself which considered it a top-level page.  Now it is nested under product instead which seems more inline with the new layout.


flub:
Giving up on this, perhaps people wanted in here anyway.  Still weird to me.

PR #2481, opened 2020-10-07
Update Java Wizard pages.

maciejwalkowiak:
cc @bruno-garcia


src/wizard/java/index.md
```diff
@@ -5,139 +5,53 @@ support_level: production
...
-    }
+```java
+try {
+  System.out.println(1/0);
+} catch (Exception e) {
+  Sentry.captureException(e);
 }
 ```
+
+Once you've verified the library is initialized properly and sent a test event, consider visiting our [complete Java docs](https://docs.sentry.io/platforms/java/).
```
marandaneto, line 57: should we just`/platforms/java/`? and let the engine create the redirect?

src/wizard/java/spring-boot.md
```diff
@@ -0,0 +1,79 @@
...
+
+  @GetMapping("/")
+  void hello() {
+    LOGGER.error("Event triggered by Logback integration");
+    throw new IllegalArgumentException("Event triggered by Spring integration");
+  }
+}
+```
+
+Once you've verified the library is initialized properly and sent a test event, consider visiting our [complete Spring Boot docs](https://docs.sentry.io/platforms/java/guides/spring-boot/).
```
marandaneto, line 79: same https://github.com/getsentry/sentry-docs/pull/2481/files#r501517301

marandaneto:
I'm trying to solve the `Authorization required to deploy.` thingie.

src/wizard/java/index.md
maciejwalkowiak: The problem is that wizards are shown in different domain so I don't think it will create a correct absolute link to docs..?

src/wizard/java/index.md
```diff
@@ -5,139 +5,53 @@ support_level: production
...
-
-```scala
-libraryDependencies += "io.sentry" % "sentry" % "{{ packages.version('sentry.java', '1.7.30') }}"
-```
-
-## Capture an Error
+## Configure
 
-To report an event manually you need to initialize a `SentryClient`. It is recommended that you use the static API via the `Sentry` class, but you can also construct and manage your own `SentryClient` instance. An example of each style is shown below:
+Initialize Sentry class with an `init` method, configure the DSN and any other [_settings_](/platforms/java/configuration/#options) you need:
```
marandaneto, line 30: whats about here?

src/wizard/java/index.md
marandaneto: but then it'd not work here too: https://github.com/getsentry/sentry-docs/pull/2481/files#r501609138

bruno-garcia:
Authorized. Let's see if it works

PR #2483, opened 2020-10-07
feat: Support full fallback path on platform content

dcramer:
This resolves an issue for guides which their base platform also includes a fallbackPlatform. For example, the `node.express` guide should fallback to `node`, then to `javascript`.


dcramer:
We need to clean up some JS samples that are wrong in Node-land

PR #2485, opened 2020-10-07
fix(tracing): Remove incorrect references to response headers

lobsterkatie:
We only send `sentry-trace` headers in requests.


src/platforms/javascript/common/performance/index.mdx
```diff
@@ -96,11 +96,11 @@ Without sampling, our atomatic instrumenation will send a transaction any time a
...
 
 ### Pageload
 
-When enabling tracing in both frontend and backend and taking advantage of the automatic frontend instrumentation, you can connect the automatically-generated `pageload` transaction on the frontend with the request transaction that serves the page on the backend. Because JavaScript code running in a browser cannot read the response headers of the current page, the `trace_id` must be transmitted in the response itself, specifically in a `<meta>` tag in the `<head>` of the HTML sent from your backend.
+When enabling tracing in both frontend and backend and taking advantage of the automatic frontend instrumentation, you can connect the automatically-generated `pageload` transaction on the frontend with the request transaction that serves the page on the backend. Because JavaScript code running in a browser cannot read the headers of the current page, the `trace_id` must be transmitted in the response itself, specifically in a `<meta>` tag in the `<head>` of the HTML sent from your backend.
```
rhcarvalho, line 103: This one was there on purpose to explain we use meta instead of response headers -- it obviously cannot be in the request header, since the info comes from the backend.

Explaining why the use of a meta tag is perhaps not crucial, so please either revert this line or remove the justification altogether.

rhcarvalho (changes requested):

src/platforms/javascript/common/performance/index.mdx
lobsterkatie: Yeah, I wondered about that one... 

Fixed.

src/platforms/javascript/common/performance/index.mdx
```diff
@@ -119,7 +119,7 @@ Once the data is included in the `<meta>` tag, our `BrowserTracing` integration
 
 ### Navigation and Other XHR Requests
 
-Once a page is loaded, any requests it makes and any responses your backend generates as a result are linked through a request/response header.
+Once a page is loaded, any requests it makes (and any requests your backend makes as a result) are linked through a request header.
```
rhcarvalho, line 122: This statement can be misleading. Trace propagation via request headers only happens when there is an "active" transaction / span.

Since we're at it, I'd suggest:

```suggestion
Once a page is loaded, every navigation event triggers a `navigation` transaction, and any Fetch or XHR requests made within this transaction will automatically propagate the trace information.
```

Feel free to tweak it.

src/platforms/javascript/common/performance/index.mdx
```diff
@@ -96,7 +96,7 @@ Without sampling, our atomatic instrumenation will send a transaction any time a
 
 ## Connecting Backend and Frontend Transactions
 
-To connect backend and frontend transactions into a single coherent trace, Sentry uses a `trace_id` value which is propagated between frontend and backend. Depending on the circumstance, this id is transmitted either in a request/response header or in an HTML `<meta>` tag. Linking transactions in this way makes it possible for you to navigate between them in the Sentry UI, so you can better understand how the different parts of your system are affecting each other.
+To connect backend and frontend transactions into a single coherent trace, Sentry uses a `trace_id` value which is propagated between frontend and backend. Depending on the circumstance, this id is transmitted either in a request header or in an HTML `<meta>` tag. Linking transactions in this way makes it possible for you to navigate between them in the Sentry UI, so you can better understand how the different parts of your system are affecting each other.
```
rhcarvalho, line 99:
```suggestion
To connect transactions from different software components into a single coherent trace, Sentry uses a `trace_id` value which is typically propagated via a request header. Linking transactions in this way makes it possible for you to navigate between them in the Sentry UI, so you can better understand how the different parts of your system are affecting each other.
```

The use of `trace_id` to connect transactions is not limited to what this section was originally about: the specific backend => frontend flow.

The suggestion makes it a more general introduction, leaving the details for the next two subsections.

rhcarvalho (approved):
This is 👍  wrt removing ref. to response headers.

src/platforms/javascript/common/performance/index.mdx
lobsterkatie: I like it, but in this case this page lives specifically in the browser (and eventually node) docs, so the front/back idea applies. I'm going to keep this in my back pocket, though, in case we ever try to pull this page into `/common` (which we probably will at some point).

PR #2488, opened 2020-10-07
fix: Add missing Django redirect

untitaker:


untitaker:
generic integration redirects broke platforms where integration pages lived somewhere else still :(

PR #2489, opened 2020-10-07
ref: Rework configure_scope examples in Python

untitaker:


src/includes/set-context/python.mdx
```diff
@@ -1,10 +1,9 @@
 ```python
-from sentry_sdk import configure_scope
+from sentry_sdk import set_context
```
rhcarvalho, line 2: These `from sentry_sdk import *` examples look non-idiomatic to me. I expect most users to be writing `import sentry_sdk` and `sentry_sdk.FOO()` instead of bringing parts of the SDK into the global namespace.

src/includes/configure-scope/python.mdx
```diff
@@ -1,7 +1,12 @@
...
-with configure_scope() as scope:
+with sentry_sdk.configure_scope() as scope:
     scope.set_tag("my-tag", "my value")
-    scope.user = {'id': 42, 'email': 'john.doe@example.com'}
+    scope.user = {"id": 42, "email": "john.doe@example.com"}
+
+# Or:
+
+sentry_sdk.set_tag("my-tag", "my value")
+sentry_sdk.set_user({"id": 42, "email": "john.doe@example.com"})
```
rhcarvalho, line 11: Maybe better examples here would be methods that don't have global helpers:

```py
import logging
import sentry_sdk

with configure_scope() as scope:
    scope.clear()
    scope.set_level(logging.DEBUG)
```

rhcarvalho (approved):
I lack the context behind this change, but looks good. Removing lots of unnecessary use of `configure_scope` 👍

src/includes/configure-scope/python.mdx
untitaker: I added clear() but set_level already exists. I think we should strive for exposing more stuff as global api rather than documenting the differences.

PR #2492, opened 2020-10-08
Update: downloading 2FA codes

MimiDumpling:
Updated wording to encourage users to download their 2FA codes before they're lost.


sentrypri (approved):
Looks good

PR #2497, opened 2020-10-09
Feat/docs fixed navs layout

stevenplewis:


dcramer:
Mobile view with a large vertical height is a bit wonky. sidebar expands but not to 100% height so you have two scrollbars.

dcramer:
Also Chrome on Windows has the sidebar scrollbar always visible. Seems like we wouldn't want that.

stevenplewis:
@dcramer I changed the mobile sidebar behavior to just be like how it is now, where it just adds it to the top of the page and isn't fixed. I removed one `overflow-y: auto` rule that _maybe_ was causing the always-there scroll on windows but I'm unsure. Also may "Depends on the user agent." according to [Mozilla's read of "auto" for scroll overflow](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-y). I can keep looking but don't have a way to test on Windows set up atm.

dcramer:
gonna merge this on green

PR #2502, opened 2020-10-12
fix: migration apple common

bruno-garcia:
Currently migrations only show up when selecting `apple`:

image

If you select `iOS` for example, it goes away:

image


marandaneto (approved):
LGTM

PR #2506, opened 2020-10-12
Performance support for Node - AWS and GCP

ajjindal:
Support for Performance on Serverless - Node
- Docs update include updates for both AWS and GCP.


PeloWriter (commented):
I think the event-detail.mdx and relay/index.mdx files got picked up somehow for this PR, but aren't actually meant as part of it. LMK if you have merge conflicts.

PR #2524, opened 2020-10-14
feat(API docs): Parse markdown links

ceorourke:
In order for the API docs to be able to use links in the path and body param descriptions, we must parse markdown links into html links

Screen Shot 2020-10-14 at 4 32 08 PM


ceorourke:
closed in favor of https://github.com/getsentry/sentry-docs/pull/2530

PR #2528, opened 2020-10-15
feat: React Native SDK 2.x migration docs

jennmueng:
Docs for users migrating to the 2.x version of the React Native SDK.

- Changes android package name from `io.sentry.RNSentryPackage` to `io.sentry.react.RNSentryPackage`
- Migration docs for users moving from 1.x to 2.x.

Linked PR:
https://github.com/getsentry/sentry-react-native/pull/1131

### DIFF
--- a/src/platforms/react-native/migration.mdx
+++ b/src/platforms/react-native/migration.mdx
@@ -1,8 +1,39 @@
 ---
-title: Migrating from react-native-sentry
+title: Migration
 ---
 
-If you are upgrading from an earlier version of react-native-sentry you should unlink the package to ensure the generated code is updated to the latest version:
+## From 1.x to 2.x
+
+In `@sentry/react-native` version `2.x`, release health and native stack traces are now enabled by default.
+
+### iOS/MacOS
+
+There are no immediate breaking changes for iOS/MacOS on the React Native side, just make sure you run `pod install` after the upgrade.
+
+If you use the `sentry-cocoa` SDK directly, you should follow its [migration guide for 5.x to 6.x](/platforms/apple/migration/#migrating-from-5x-to-6x).
+
+### Android
+
+If you are on React Native <0.60, you will need to update this line in your `MainApplication.java`:
+
+
+Old:
+
+```java
+import io.sentry.RNSentryPackage;
+```
+
+New:
+```java
+import io.sentry.react.RNSentryPackage;
+```
+
+Otherwise, there should be no other immediate breaking changes on the React Native side. If you use the `sentry-andoid` SDK directly, you should follow its [migration guide for 2.x to 3.x](/platforms/android/migration/#migrating-from-sentry-android-2x-to-sentry-android-3x).
+
+
+## From 0.x to 1.x
+
+If you are upgrading from an earlier version of `react-native-sentry` you should unlink the package to ensure the generated code is updated to the latest version:
 
 ```bash
 react-native unlink react-native-sentry
--- a/src/platforms/react-native/migration.mdx
+++ b/src/platforms/react-native/migration.mdx
@@ -1,33 +1,34 @@
 ---
 title: Migration
+description: "Learn more about migrating from 1.x to 2.x to enable release health tracking and native stack traces by default."
 ---
 
 ## From 1.x to 2.x
 
-In `@sentry/react-native` version `2.x`, release health and native stack traces are now enabled by default.
+Sentry's most recent version of our React Native SDK enables release health tracking and native stack traces by default.
 
-This version uses the [envelope endpoint](https://develop.sentry.dev/sdk/envelopes/).
-If you are using an on-premise installation it requires Sentry version
-`>= v20.6.0` to work. If you are using sentry.io nothing will change and
-no action is needed.
+This version of the SDK uses the [envelope endpoint](https://develop.sentry.dev/sdk/envelopes/). If you are using an on-premise installation, the SDK requires Sentry version 20.6.0 and above. If you are using our SaaS product (`sentry.io`), there is no change and no action is required.
 
 ### iOS/MacOS
 
-There are no immediate breaking changes for iOS/MacOS on the React Native side, just make sure you run `pod install` after the upgrade.
+While the migration does not introduce breaking changes for iOS/MacOS on the React Native side, we recommend that you run `pod install` after the upgrade.
 
-If you use the `sentry-cocoa` SDK directly, you should follow its [migration guide for 5.x to 6.x](/platforms/apple/migration/#migrating-from-5x-to-6x).
+<Note>
+<markdown>
 
-On iOS/MacOS, we now cache events in envelopes on the disk.
-We decided not to take the effort to migrate these few cached events from 5.x to 6.x into
-envelopes. Instead we remove them from the disk. This means you might lose a few cached
-events of your users when upgrading.
+If you are using our Cocoa SDK directly, follow that [migration guide for 5.x to 6.x](/platforms/apple/migration/#migrating-from-5x-to-6x).
+
+</markdown>
+</Note>
+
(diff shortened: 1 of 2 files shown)


HazAT:
@jennmueng can we move this into the repo? 
We have it also in other repos https://github.com/getsentry/sentry-javascript/blob/master/MIGRATION.md

So we can remove the migration.mdx

philipphofmann:
> @jennmueng can we move this into the repo?
> We have it also in other repos https://github.com/getsentry/sentry-javascript/blob/master/MIGRATION.md
> 
> So we can remove the migration.mdx

I moved the migration from the repo to the docs for Cocoa. I thought we should have almost everything in the docs. So which one is it now? @HazAT

HazAT:
damn 😅 

OK, only thing I want to achieve is that we have it in two places and they go out of sync. Since we don't have migration in RN in the repo so far, lets put it into the docs.

src/platforms/react-native/migration.mdx
```diff
@@ -1,8 +1,39 @@
 ---
-title: Migrating from react-native-sentry
+title: Migration
 ---
 
-If you are upgrading from an earlier version of react-native-sentry you should unlink the package to ensure the generated code is updated to the latest version:
+## From 1.x to 2.x
+
+In `@sentry/react-native` version `2.x`, release health and native stack traces are now enabled by default.
```
philipphofmann, line 7: I think we should also mention the envelope endpoint, see https://docs.sentry.io/platforms/apple/guides/ios/migration/#store-endpoint.

HazAT: Good point

jennmueng: Good catch. Will do.

jennmueng:
Ah I made it in the docs as I saw on cocoa and Android it was in the docs. For react native let’s keep it in the docs then so we don’t have two fragmented parts.

bruno-garcia:
Also would prefer the docs to be honest. For migrations

bruno-garcia (approved):

philipphofmann (approved):
LGTM from Apple/Cocoa

src/platforms/react-native/migration.mdx
```diff
@@ -1,8 +1,49 @@
 ---
-title: Migrating from react-native-sentry
+title: Migration
 ---
```
PeloWriter, line 3:
```suggestion
description: "Learn more about migrating from 1.x to 2.x to enable release health tracking and native stack traces by default."
---
```

src/platforms/react-native/migration.mdx
```diff
@@ -1,8 +1,49 @@
...
 
-If you are upgrading from an earlier version of react-native-sentry you should unlink the package to ensure the generated code is updated to the latest version:
+## From 1.x to 2.x
+
+In `@sentry/react-native` version `2.x`, release health and native stack traces are now enabled by default.
```
PeloWriter, line 7:
```suggestion
Sentry's most recent version of our React Native SDK enables release health tracking and native stack traces by default.
```

src/platforms/react-native/migration.mdx
```diff
@@ -1,8 +1,49 @@
...
+
+This version uses the [envelope endpoint](https://develop.sentry.dev/sdk/envelopes/).
+If you are using an on-premise installation it requires Sentry version
+`>= v20.6.0` to work. If you are using sentry.io nothing will change and
+no action is needed.
```
PeloWriter, line 12:
```suggestion
This version of the SDK uses the [envelope endpoint](https://develop.sentry.dev/sdk/envelopes/). If you are using an on-premise installation, the SDK requires Sentry version 20.6.0 and above. If you are using our SaaS product (`sentry.io`), there is no change and no action is required.
```

src/platforms/react-native/migration.mdx
```diff
@@ -1,8 +1,49 @@
...
+
+### iOS/MacOS
+
+There are no immediate breaking changes for iOS/MacOS on the React Native side, just make sure you run `pod install` after the upgrade.
```
PeloWriter, line 16:
```suggestion
While the migration does not introduce breaking changes for iOS/MacOS on the React Native side, we recommend that you run `pod install` after the upgrade.
```

src/platforms/react-native/migration.mdx
```diff
@@ -1,8 +1,49 @@
...
+
+If you use the `sentry-cocoa` SDK directly, you should follow its [migration guide for 5.x to 6.x](/platforms/apple/migration/#migrating-from-5x-to-6x).
```
PeloWriter, line 18:
```suggestion
<Note>
<markdown>

If you are using our Cocoa SDK directly, follow that [migration guide for 5.x to 6.x](/platforms/apple/migration/#migrating-from-5x-to-6x).

</markdown>
</Note>
```

src/platforms/react-native/migration.mdx
```diff
@@ -1,8 +1,49 @@
...
+
+On iOS/MacOS, we now cache events in envelopes on the disk.
+We decided not to take the effort to migrate these few cached events from 5.x to 6.x into
+envelopes. Instead we remove them from the disk. This means you might lose a few cached
+events of your users when upgrading.
```
PeloWriter, line 23:
```suggestion
On iOS/MacOS, we now cache events in envelopes on the disk. As a result, you might lose a few cached events during the migration. Due to the effort involved, the migration from 5.x to 6.x does not move these few cached events into envelopes.
```

src/platforms/react-native/migration.mdx
```diff
@@ -1,8 +1,49 @@
...
+
+### Android
+
+If you are on React Native <0.60, you will need to update this line in your `MainApplication.java`:
```
PeloWriter, line 27:
```suggestion
If you are using our React Native SDK earlier than version 0.60, you will need to update this line in your `MainApplication.java`:
```

src/platforms/react-native/migration.mdx
```diff
@@ -1,8 +1,49 @@
...
+
+
+Old:
```
PeloWriter, line 30:
```suggestion
From (earlier version):
```

src/platforms/react-native/migration.mdx
```diff
@@ -1,8 +1,49 @@
...
+```java
+import io.sentry.RNSentryPackage;
+```
+
+New:
+```java
+import io.sentry.react.RNSentryPackage;
+```
+
+Otherwise, there should be no other immediate breaking changes on the React Native side. If you use the `sentry-andoid` SDK directly, you should follow its [migration guide for 2.x to 3.x](/platforms/android/migration/#migrating-from-sentry-android-2x-to-sentry-android-3x).
```
PeloWriter, line 41:
```suggestion
Other than the one line change noted above, the migration should not cause breaking changes on the React Native side. 

<Note>
<markdown>

If you use our Android SDK directly, you should follow its [migration guide for 2.x to 3.x](/platforms/android/migration/#migrating-from-sentry-android-2x-to-sentry-android-3x).

</markdown>
</Note>
```

src/platforms/react-native/migration.mdx
```diff
@@ -1,8 +1,49 @@
...
+
+
+## From 0.x to 1.x
+
+If you are upgrading from an earlier version of `react-native-sentry` you should unlink the package to ensure the generated code is updated to the latest version:
```
PeloWriter, line 46:
```suggestion
If you are upgrading from an earlier version of Sentry's React Native SDK, you should unlink the package to ensure the generated code is updated to the latest version:
```

PeloWriter (commented):
Some edits to smooth and clarify.

src/platforms/react-native/migration.mdx
jennmueng: Ah, this is actually referring to users on React Native less than 0.60 not our SDK.

PR #2530, opened 2020-10-15
fix: Render markdown formatting in description

NisanthanNanthakumar:
## Objective
Original Problem:
On the API docs, the description field no longer formats backticks correctly, instead rendering them as "<code>id</code>" for example.

In this PR, I made the `gatsby-plugin-openapi` create child nodes for each API path's parameters and description. For this child nodes, I tagged as `text/markdown` so that the `gatsby-plugin-mdx` and `gatsby-transformer-remark` automatically parse the strings. Then we can render the result in the `<Markdown>` component. This removes the need to `dangerouslySetInnerHTML` for the params and the description.

I removed `parseBacticks` bc we don't need it anymore.

## UI
**Before**
Screen Shot 2020-10-15 at 11 51 31 AM

**After**
Screen Shot 2020-10-15 at 11 51 45 AM

### DIFF
--- a/src/templates/apiPage.tsx
+++ b/src/templates/apiPage.tsx
@@ -2,7 +2,7 @@ import React, { useState, useEffect } from "react";
 import { graphql } from "gatsby";
 import Prism from "prismjs";
 
-import { parseBackticks } from "~src/utils";
+import Markdown from "~src/components/markdown.tsx";
 import BasePage from "~src/components/basePage";
 import SmartLink from "~src/components/smartLink";
 import ApiSidebar from "~src/components/apiSidebar";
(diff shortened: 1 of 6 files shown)


src/templates/apiPage.tsx
```diff
@@ -113,7 +113,11 @@ export default props => {
 
           {data.description && (
             <div className="pb-3 content-flush-bottom">
-              <p>{parseBackticks(data.description)}</p>
```
dcramer, line 116: is it possible for the description to contain other html tags? such as a `<script>` tag? if so we'll need to do something differently here to avoid xss

NisanthanNanthakumar: @dcramer so far, we only use <a></a> and <code></code> equivalents in the OpenAPI schema.

dcramer: yeah but if something else exists will it just insert it? if so lets use dompurify here to restrict what html tags are possible

also could you not just render data.description as raw markdown somehow? e.g. how we did it before where the remark processor created a markdown node

mgaeta (approved):
👍

PR #2531, opened 2020-10-15
fix: Remove repeated sentence

rhcarvalho:
The same sentence makes up a generic paragraph before the include.

---

https://docs.sentry.io/platforms/node/guides/aws-lambda/performance/sampling/#default-sampling-context-data
image


lobsterkatie (approved):
Good catch, thanks!

PR #2535, opened 2020-10-16
fix: state the fixed version about the native pointers on android

marandaneto:
related https://github.com/getsentry/sentry-java/pull/977

### DIFF
--- a/src/platforms/android/usage/advanced-usage.mdx
+++ b/src/platforms/android/usage/advanced-usage.mdx
@@ -162,7 +162,8 @@ when using a Android Gradle Plugin >= v3.6.0, and it needs to be enabled with th
 
 Apart from this, `sentry-native` also uses [tagged pointers](https://source.android.com/devices/tech/debug/tagged-pointers)
 internally, and will therefore not work out of the box on newer Android devices which have their own conflicting pointer tagging.
-To prevent this, disable the [`allowNativeHeapPointerTagging`](https://developer.android.com/guide/topics/manifest/application-element#allowNativeHeapPointerTagging) configuration option. You only need to do that if you are using `sentry-android` < 3.1, as this is fixed in `sentry-android` >= 3.1.
+If you are using `sentry-android` version earlier than 3.1, you must manually disable the [`allowNativeHeapPointerTagging`](https://developer.android.com/guide/topics/manifest/application-element#allowNativeHeapPointerTagging) configuration option because `sentry-native` uses [tagged pointers](https://source.android.com/devices/tech/debug/tagged-pointers)
+internally. As a result, this option will not work out of the box on newer Android devices that have their own conflicting pointer tagging. This is fixed in `sentry-android` versions 3.1 and above.
 
 **Example AndroidManifest.xml**:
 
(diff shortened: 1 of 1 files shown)


src/platforms/android/usage/advanced-usage.mdx
```diff
@@ -162,7 +162,7 @@ when using a Android Gradle Plugin >= v3.6.0, and it needs to be enabled with th
 
 Apart from this, `sentry-native` also uses [tagged pointers](https://source.android.com/devices/tech/debug/tagged-pointers)
 internally, and will therefore not work out of the box on newer Android devices which have their own conflicting pointer tagging.
-To prevent this, disable the [`allowNativeHeapPointerTagging`](https://developer.android.com/guide/topics/manifest/application-element#allowNativeHeapPointerTagging) configuration option.
+To prevent this, disable the [`allowNativeHeapPointerTagging`](https://developer.android.com/guide/topics/manifest/application-element#allowNativeHeapPointerTagging) configuration option, you only need to do that if you are using `sentry-android` <= 3.0, `sentry-android` >= 3.1 already fixed it.
```
Swatinem, line 165:
```suggestion
To prevent this, disable the [`allowNativeHeapPointerTagging`](https://developer.android.com/guide/topics/manifest/application-element#allowNativeHeapPointerTagging) configuration option. You only need to do that if you are using `sentry-android` < 3.1, as this is fixed in `sentry-android` >= 3.1.
```

src/platforms/android/usage/advanced-usage.mdx
```diff
@@ -162,7 +162,7 @@ when using a Android Gradle Plugin >= v3.6.0, and it needs to be enabled with th
 
 Apart from this, `sentry-native` also uses [tagged pointers](https://source.android.com/devices/tech/debug/tagged-pointers)
 internally, and will therefore not work out of the box on newer Android devices which have their own conflicting pointer tagging.
-To prevent this, disable the [`allowNativeHeapPointerTagging`](https://developer.android.com/guide/topics/manifest/application-element#allowNativeHeapPointerTagging) configuration option.
+To prevent this, disable the [`allowNativeHeapPointerTagging`](https://developer.android.com/guide/topics/manifest/application-element#allowNativeHeapPointerTagging) configuration option. You only need to do that if you are using `sentry-android` < 3.1, as this is fixed in `sentry-android` >= 3.1.
```
philipphofmann, line 165: `l`: Maybe we can change the wording a bit to make it clearer:
```suggestion
 If you are using `sentry-android` < 3.1 you have to disable the [`allowNativeHeapPointerTagging`](https://developer.android.com/guide/topics/manifest/application-element#allowNativeHeapPointerTagging) configuration option manually, because `sentry-native` uses [tagged pointers](https://source.android.com/devices/tech/debug/tagged-pointers)
internally and will therefore not work out of the box on newer Android devices that have their own conflicting pointer tagging. This is fixed in `sentry-android` >= 3.1.
```

philipphofmann (approved):
LGTM

src/platforms/android/usage/advanced-usage.mdx
PeloWriter: I agree that more detail makes it more clear. Suggest:
```suggestion
If you are using `sentry-android` version earlier than 3.1, you must manually disable the [`allowNativeHeapPointerTagging`](https://developer.android.com/guide/topics/manifest/application-element#allowNativeHeapPointerTagging) configuration option because `sentry-native` uses [tagged pointers](https://source.android.com/devices/tech/debug/tagged-pointers)
internally. As a result, this option will not work out of the box on newer Android devices that have their own conflicting pointer tagging. This is fixed in `sentry-android` versions 3.1 and above.
```

PeloWriter (approved):
LTGM

src/platforms/android/usage/advanced-usage.mdx
marandaneto: thanks @philipphofmann and @PeloWriter :)

bruno-garcia (approved):

PR #2537, opened 2020-10-16
Refactor Node Getting Started

PeloWriter:
This refactor focuses on the Node Getting Started page, with minor knock on effects to other areas of Node content.


src/includes/capture-error/node.mdx
```diff
@@ -0,0 +1,9 @@
+You can pass an error object to `captureException()` to get it captured as event. It's possible to throw strings as errors in which case no traceback can be recorded.
```
kamilogorek, line 1:
```suggestion
You can pass an error object to `captureException()` to get it captured as event. It's possible to throw strings as errors in which case no stacktrace can be recorded.
```

`traceback` is term used in Python, where `stacktrace` in JavaScript.

src/includes/getting-started-config/node.mdx
```diff
@@ -1,9 +1,29 @@
-You need to inform the Sentry Node SDK about your DSN:
+Once this is done, Sentry's Node SDK captures all unhandled exceptions and transactions.
```
kamilogorek, line 1: It kinda sounds like transactions are "unhandled" as well 🤔  Maybe switching the order to `transactions and unhandled exceptions` would make it more explicit?

src/includes/getting-started-config/node.mdx
```diff
@@ -1,9 +1,29 @@
...
 
-```javascript
-const Sentry = require("@sentry/node");
+```javascript {tabTitle:ESM}
+import * as Sentry from "@sentry/node";
+import * as Tracing from "@sentry/tracing";
 // or use es6 import statements
 // import * as Sentry from '@sentry/node';
```
kamilogorek, line 7: These 2 commented lines can be removed, as we have separate `esm/commonjs` tabs now.

src/includes/getting-started-config/node.mdx
kamilogorek: Also, code samples below won't capture any transactions. It requires either manual setup + HTTP integration, or additional all-in-one integration, like the one we have for Express.

src/includes/getting-started-primer/node.mdx
```diff
@@ -0,0 +1,9 @@
+<Note>
+<markdown>
+
+_Sentry's Node SDK enables automatic reporting of errors, exceptions, and transactions._
+
+Our Node SDK supports all recent versions, and integrates well with a variety of popular frameworks and packages. It gives developers the ability to see Node source code at each frame instead and get proper asynchronous context tracking — like thread-local storage — in a way that fits Node’s concurrency model.
```
kamilogorek, line 6: I'd remove "— like thread-local storage —" part, as it's not accurate for node. There are no thread-locals in this language.

src/platforms/node/common/typescript.mdx
```diff
@@ -2,6 +2,7 @@
 toc: true
 title: TypeScript
 sidebar_order: 500
+description: "Learn more about using your Sentry SDK and SourceMaps with TypeScript."
```
kamilogorek, line 5: Source Maps*, it should be written as 2 separate words.

kamilogorek (changes requested):

src/includes/getting-started-config/node.mdx
PeloWriter: I've switched xns and unhandled exceptions. 

@HazAT - can I get some help with a correct code sample that covers capturing xns?

src/includes/getting-started-config/node.mdx
PeloWriter: 👍

src/includes/getting-started-primer/node.mdx
PeloWriter: 👍

src/platforms/node/common/typescript.mdx
PeloWriter: thanks.

HazAT (approved):

dashed (approved):

PR #2538, opened 2020-10-17
ref: Use linkchecker github action

untitaker:


untitaker:
https://github.com/actions/runner/issues/646

PR #2539, opened 2020-10-17
ref: Use linkchecker github action

untitaker:
follow-up to https://github.com/getsentry/sentry-docs/pull/2538


untitaker:
Docker caching is also broken

PR #2550, opened 2020-10-19
Add: note about `stack.package` for Native 

MimiDumpling:
Customer wrote in about confusion between `stack.package` and `stack.module`. Added a note to clarify the difference.

### DIFF
--- a/src/docs/product/sentry-basics/search/index.mdx
+++ b/src/docs/product/sentry-basics/search/index.mdx
@@ -293,7 +293,7 @@ The following comparative operators are available:
 `stack.stack_level`<br />
 `stack.lineno`<br />
 
-: Restrict results to events with a matching stack property. Note: For Native SDK users, `stack.package` is the Native equivalent for `stack.module`. `stack.module` isn't typically used in Native code situations.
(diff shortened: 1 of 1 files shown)


untitaker (approved):
ok. I think this may speak for this doc becoming a platform doc cc @PeloWriter

src/docs/product/sentry-basics/search/index.mdx
```diff
@@ -293,7 +293,7 @@ The following comparative operators are available:
 `stack.stack_level`<br />
 `stack.lineno`<br />
 
-: Restrict results to events with a matching stack property. Note: For Native SDK users, `stack.package` is the Native equivalent for `stack.module`. `stack.module` isn't typically used in Native code situations.
+: Restrict results to events with a matching stack property. Note: For Native SDK users, `stack.package` is the Native equivalent for `stack.module`. `stack.module` isn't typically used in Native.
```
PeloWriter, line 296:
```suggestion
: Restrict results to events with a matching stack property. Note: For Native SDK users, `stack.package` is the Native equivalent for `stack.module`.
```

PeloWriter (commented):
The key to the location is where users conduct their search - the search ability is part of the web UI, thus the location of this content as part of product. I think we should add links back to the SDK content for language-specific queries @MimiDumpling.

PR #2553, opened 2020-10-20
ref(ip-ranges): Clarify domains

tonyo:
Small clarification

### DIFF
--- a/src/docs/product/security/ip-ranges.mdx
+++ b/src/docs/product/security/ip-ranges.mdx
@@ -28,13 +28,13 @@ Sentry's Event Ingestion respects two domains within a Data Source Name (DSN):
 1. The [organization subdomain](https://forum.sentry.io/t/organization-subdomains-in-dsns/9360)
 2. Sentry's apex domain
 
-Sentry's apex domain accepts events from the same IP address as the Dashboard and API:
+Sentry's apex domain (`sentry.io`) accepts events from the same IP address as the Dashboard and API:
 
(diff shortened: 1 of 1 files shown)


src/docs/product/security/ip-ranges.mdx
```diff
@@ -28,13 +28,13 @@ Sentry's Event Ingestion respects two domains within a Data Source Name (DSN):
...
 
-Sentry's apex domain accepts events from the same IP address as the Dashboard and API:
+Sentry's apex domain (`sentry.io`) accepts events from the same IP address as the Dashboard and API:
 
 ```python
 35.188.42.15/32
 ```
 
-Starting December 1st, 2020, Sentry's organization subdomains will migrate from the prior IP address to the following single address:
+Starting December 1st, 2020, Sentry's organization subdomains (`o<YOUR_ORG_ID>.ingest.sentry.io`) will migrate from the prior IP address to the following single address:
```
JTCunning, line 37: We're referred to this as (`o<number>.ingest.sentry.io`) in other places around our world of text since it's one of the only instances where customers are exposed to their organization IDs, meaning they aren't too familiar with what their ID actually is.

JTCunning (approved):
🍪 
One correction, otherwise we're good.

src/docs/product/security/ip-ranges.mdx
tonyo: Cool, updated

PR #2557, opened 2020-10-20
Update: Event Grouping

MimiDumpling:
Small improvements to the Issue Grouping section of docs to alleviate Support tickets on the topic.

### DIFF
--- a/src/platforms/common/data-management/event-grouping/grouping-enhancements.mdx
+++ b/src/platforms/common/data-management/event-grouping/grouping-enhancements.mdx
@@ -3,13 +3,9 @@ title: "Custom Grouping Enhancements"
 sidebar_order: 2
 ---
 
-If you use stack traces for grouping, the enhancements rules influence the data that feed into that algorithm.
-These rules can be configured on a per-project basis under _Project Settings > General Settings > Grouping Settings_.
+If you use stack traces for grouping, the enhancements rules influence the data that feed into that algorithm. These rules can be configured on a per-project basis under **Settings > [organization] > [project] > General: Grouping Settings**.
 
-Enhancements are rules written in a pretty straightforward way. Each line is a
-single enhancement rule. It's one or multiple match expressions followed by one
-or multiple actions to be executed when all expressions match. All rules are
-executed from top to bottom on all frames in the stack trace.
+Enhancements are straightforward rules. Each line is a single enhancement rule; one or multiple match expressions are followed by one or multiple actions to be executed when all expressions match. All rules are executed from top to bottom on all frames in the stack trace.
 
 The syntax for grouping enhancements is similar to:
 
@@ -17,8 +13,7 @@ The syntax for grouping enhancements is similar to:
 matcher-name:expression other-matcher:expression ... action1 action2 ...
 ```
 
-The syntax follows the syntax from discover queries. If you want to negate
-the match, you can prefix the expression with an exclamation mark (`!`).
+The syntax follows the syntax from [Discover queries](/product/discover-queries/). If you want to negate the match, you can prefix the expression with an exclamation mark (`!`).
 
 The following is a practical example of how this looks:
 
@@ -35,15 +30,13 @@ stack.abs_path:**/generated/**.js         -group
 
 ## Matchers
 
-These **matchers** are available; multiple matchers can be defined in a line:
+Multiple matchers can be defined in a line. The following matchers are available:
 
 <DefinitionList>
 
 ### `family`
 
-Matches on the general platform family, which currently includes `javascript`,
-`native` and `other`. Comma separate rules to apply them to multiple platforms.
-separate them.
+Matches on the general platform family, which currently includes `javascript`, `native` and `other`. Comma separate rules to apply them to multiple platforms.
 
 ```discover {tabTitle:Enhancement Rules}
 family:javascript,native stack.abs_path:**/generated/**  -group
@@ -53,9 +46,7 @@ family:javascript,native stack.abs_path:**/generated/**  -group
 
 alias: `path`
 
-This matches case insensitive with Unix glob behavior on a path in a stack trace.
-The path separators are normalized to `/`. As a special rule, if the filename is
-relative, it still matches on `**/`.
+This matcher is case insensitive with Unix glob behavior on a path in a stack trace. The path separators are normalized to `/`. As a special rule, if the filename is relative, it still matches on `**/`.
 
 ```discover {tabTitle:Enhancement Rules}
 # match on all files under `project` with a `.c` extension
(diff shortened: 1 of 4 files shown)


src/platforms/common/data-management/event-grouping/grouping-enhancements.mdx
```diff
@@ -92,30 +80,28 @@ stack.function:malloc      -group
...
 
 </DefinitionList>
 
 ## Actions
 
-There are two types of **actions**: flag setting and setting variables.
+There are two types of **actions**: flag and variables setting.
 
-- **flag**: flags identify the action to be taken if all matchers match. A flag needs to be prefixed with `+` to set it or `-` to unset it. If this expression is prefixed with a `^`, it applies to frames above the frame (toward the crash). If prefixed with `v` it applies to frames below the frame (away from the crash). For instance, `-group ^-group` removes the matching frame and all frames above it from the grouping.
+- **flag**: Flags identify the action to be taken if all matchers match. A flag needs to be prefixed with `+` to set it or `-` to unset it. If this expression is prefixed with a `^`, it applies to frames above the flag (toward the crash). If prefixed with `v` it applies to frames below the flag (away from the crash). For instance, `-group ^-group` removes the matching frame and all frames above it from the grouping.
```
MimiDumpling, line 99: @rbisol I changed the word "frame" to "flag" in a few places because I assumed "frame" was used incorrectly. Will you confirm?

rbisol: Frame in this context was the correct word but unclear. I suggest changing it to "matching frame":

```suggestion
- **flag**: Flags identify the action to be taken if all matchers match. A flag needs to be prefixed with `+` to set it or `-` to unset it. If this expression is prefixed with a `^`, it applies to frames above the matching frame (toward the crash). If prefixed with `v` it applies to frames below the matching frame (away from the crash). For instance, `-group ^-group` removes the matching frame and all frames above it from the grouping.
```

src/platforms/common/data-management/event-grouping/grouping-enhancements.mdx
```diff
@@ -3,22 +3,17 @@ title: "Custom Grouping Enhancements"
 sidebar_order: 2
 ---
 
-If you use stack traces for grouping, the enhancements rules influence the data that feed into that algorithm.
-These rules can be configured on a per-project basis under _Project Settings > General Settings > Grouping Settings_.
+If you use stack traces for grouping, the enhancements rules influence the data that feed into that algorithm. These rules can be configured on a per-project basis under **Settings > [organization] > [project] > General: Grouping Settings**.
```
georgekallilas, line 6: I suggest using the exact path order that a user would have to follow in order to reach those settings.
```suggestion
If you use stack traces for grouping, the enhancements rules influence the data that feed into that algorithm. These rules can be configured on a per-project basis under **Settings > Projects > [project] > General Settings > Grouping Settings**.
```

src/platforms/common/data-management/event-grouping/index.mdx
```diff
@@ -14,31 +14,33 @@ All events have a fingerprint. Events with the same fingerprint are grouped toge
 
 ## Grouping Algorithms
 
-Each time the default behavior is modified, Sentry releases it as a new version so it does not affect how existing issues are grouped. When you create a Sentry project, the latest and greatest version of the grouping algorithm is automatically selected. This means that the grouping behavior is consistent within a project. If you want to upgrade an existing project to a new grouping algorithm version, you can do so in the project settings. When upgrading you will very likely see new groups being created.
+Each time the default behavior is modified, Sentry releases it as a new version to not affect how existing issues are grouped. When you create a Sentry project, the latest and greatest version of the grouping algorithm is automatically selected. This means that the grouping behavior is consistent within a project. If you want to upgrade an existing project to a new grouping algorithm version, you can do so in **Settings > [organization] > [project] > General: Grouping Settings**. When upgrading, you will very likely see new groups being created.
```
georgekallilas, line 17: Natural path order is recommended (as mentioned above).
```suggestion
Each time the default behavior is modified, Sentry releases it as a new version to not affect how existing issues are grouped. When you create a Sentry project, the latest and greatest version of the grouping algorithm is automatically selected. This means that the grouping behavior is consistent within a project. If you want to upgrade an existing project to a new grouping algorithm version, you can do so in **Settings > Projects > [project] > General Settings > Grouping Settings**. When upgrading, you will very likely see new groups being created.
```

src/platforms/common/data-management/event-grouping/sdk-fingerprinting.mdx
```diff
@@ -5,8 +5,8 @@ redirect_from:
   - ../../../enriching-events/context/fingerprint-override/
 ---
 
-In supported SDKs, you can override the Sentry default grouping passing the `fingerprint` attribute an array of strings. This works similar to the [server side functionality](../server-side-fingerprinting/) which is always available and
-can achieve similar results.
+In supported SDKs, you can override the Sentry default grouping passing the fingerprint attribute an array of strings. The length of a fingerprints array is not restricted. This works similarly to the [server-side functionality](../server-side-fingerprinting/), which is always available and can achieve similar results.
```
georgekallilas, line 8:
```suggestion
In supported SDKs, you can override the Sentry default grouping passing the fingerprint attribute an array of strings. The length of a fingerprint's array is not restricted. This works similarly to the [server-side functionality](../server-side-fingerprinting/), which is always available and can achieve similar results.
```

src/platforms/common/data-management/event-grouping/server-side-fingerprinting.mdx
```diff
@@ -3,31 +3,20 @@ title: "Server-side Fingerprinting"
 sidebar_order: 1
 ---
 
-Server-side fingerprinting is also configured with a config similar to
-[grouping enhancements](../grouping-enhancements/),
-but the syntax is slightly different. The matchers are the same
-but instead of flipping flags, a fingerprint is assigned that overrides the
-default grouping entirely.
+Server-side fingerprinting is also configured with a config similar to [grouping enhancements](../grouping-enhancements/), but the syntax is slightly different. The matchers are the same, but instead of flipping flags, a fingerprint is assigned, and it overrides the default grouping entirely.
```
georgekallilas, line 6:
```suggestion
Server-side fingerprinting is also configured with a config similar to [grouping enhancements](../grouping-enhancements/), but the syntax is slightly different. The matchers are the same, but instead of flipping flags, a fingerprint is assigned and it overrides the default grouping entirely.
```

src/platforms/common/data-management/event-grouping/server-side-fingerprinting.mdx
```diff
@@ -3,31 +3,20 @@ title: "Server-side Fingerprinting"
...
 
-These rules can be configured on a per-project basis under _Project Settings > General Settings > Grouping Settings_. You are presented with an input field where you can write custom
-fingerprinting rules. The syntax follows the syntax from Discover queries. If
-you want to negate the match, you can prefix the expression with an
-exclamation mark (`!`).
+These rules can be configured on a per-project basis under **Settings > [organization] > [project] > General: Grouping Settings**. Group Settings has input fields where you can write custom fingerprinting rules. The syntax follows the syntax from [Discover queries](/product/discover-queries/). If you want to negate the match, prefix the expression with an exclamation mark (`!`).
```
georgekallilas, line 8:
```suggestion
These rules can be configured on a per-project basis under **Settings > Projects > [project] > General Settings > Grouping Settings**. Group Settings has input fields where you can write custom fingerprinting rules. The syntax follows the syntax from [Discover queries](/product/discover-queries/). If you want to negate the match, prefix the expression with an exclamation mark (`!`).
```

georgekallilas (changes requested):
Good job! Left some comments for review.

rbisol (approved):

georgekallilas (approved):

PR #2558, opened 2020-10-21
ref(breadcrumbs): Add to and tweak common SDK breadcrumb content

lobsterkatie:
This PR does a few things:

- Adds a list of the keys which can be used in a manually-captured breadcrumb, along with a warning that sending keys other than those listed will result in the data being lost.

- Moves the `beforeBreadcrumb` example into the correct section.

- Explains how `beforeBreadcrumb` can be used to drop a breadcrumb.

- Does general wordsmithing.

### DIFF
--- a/src/platforms/common/enriching-events/breadcrumbs.mdx
+++ b/src/platforms/common/enriching-events/breadcrumbs.mdx
@@ -25,14 +25,16 @@ You can manually add breadcrumbs whenever something interesting happens. For exa
 
 <PlatformContent includePath="breadcrumbs-example" />
 
+The available breadcrumb keys are "type", "category", "message", "level", "timestamp" (which many SDKs will set automatically for you), and "data", which is the place to put any information you'd like the breadcrumb to include. Using keys other than these six won't cause an error, but will result in the data being dropped when the event is processed by Sentry.
+
 ## Automatic Breadcrumbs
 
-SDKs will automatically start recording breadcrumbs by enabling integrations. For instance, the browser JavaScript SDK will automatically record all location changes.
-
-<PlatformContent includePath="before-breadcrumb" />
+SDKs and their associated integrations will automatically record many types of breadcrumbs. For instance, the browser JavaScript SDK will automatically record all location changes.
 
 ## Customize Breadcrumbs
 
-SDKs customize breadcrumbs through the `before_breadcrumb` hook. This hook passes an already assembled breadcrumb and, in some SDKs, an optional hint. The function can modify the breadcrumb or decide to discard it entirely:
+SDKs allow you to customize breadcrumbs through the `before_breadcrumb` hook. This hook is passed an already assembled breadcrumb and, in some SDKs, an optional hint. The function can modify the breadcrumb or decide to discard it entirely by returning `null`:
+
+<PlatformContent includePath="before-breadcrumb" />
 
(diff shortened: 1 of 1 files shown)


src/platforms/common/enriching-events/breadcrumbs.mdx
```diff
@@ -25,14 +25,16 @@ You can manually add breadcrumbs whenever something interesting happens. For exa
 
 <PlatformContent includePath="breadcrumbs-example" />
 
-## Automatic Breadcrumbs
+The available breadcrumb keys are "type", "category", "message", "level", "timestamp" (which many SDKs will set automatically for you), and "data", which is the place to put any information you'd like the breadcrumb to include. Using keys other than these six won't cause an error, but will result in the data being dropped when the event is processed by Sentry.
```
jan-auer, line 28: We may want to call out that this is relevant mostly for untyped languages like JavaScript. In all typed languages, Breadcrumbs and their fields are typed out, too. Maybe, this could be shown conditionally based on the platform? 

Should we also wrap the field names in code?

jan-auer: The data inside breadcrumbs can also be partially typed. The canonical documentation for this lives at:
https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/

lobsterkatie: > We may want to call out that this is relevant mostly for untyped languages like JavaScript. In all typed languages, Breadcrumbs and their fields are typed out, too. Maybe, this could be shown conditionally based on the platform?

Oh, right, that makes sense. So it would be shown for JS/Node/RN, Python, PHP, Ruby, Elixir, and Perl, but not for Java, Native, C#, Dart, Go, or Rust? What about "Apple," which is both Obj-C (dynamically typed) and Swift (statically typed)?

jan-auer: If something like this is possible, yes. 

Only exception: For native, we can show it since we haven't typed out the protocol.

lobsterkatie: Okay, should be good now.

src/platforms/common/enriching-events/breadcrumbs.mdx
```diff
@@ -25,7 +25,13 @@ You can manually add breadcrumbs whenever something interesting happens. For exa
 
 <PlatformContent includePath="breadcrumbs-example" />
 
-The available breadcrumb keys are `type`, `category`, `message`, `level`, `timestamp` (which many SDKs will set automatically for you), and `data`, which is the place to put any information you'd like the breadcrumb to include. Using keys other than these six won't cause an error, but will result in the data being dropped when the event is processed by Sentry.
+<!-- TODO: Native shouldn't really be on this list, since c++ is statically typed. Remove it once types have been added to the native SDK. -->
```
jan-auer, line 28: The Native SDK will never receive types, so it needs to stay here.

lobsterkatie: > For native, we can show it since we haven't typed out the protocol.

Ah - for some reason I read an implied "yet" in there. Will take out the todo.

jan-auer (approved):

PR #2560, opened 2020-10-21
fix: Force redirect for Ember

k-fish:
This adds a redirect for Ember to fix search results in google not bringing you to the new guides.


PeloWriter (approved):
LGTM. Thanks.

PR #2562, opened 2020-10-21
Refactor the Getting Started for Java

PeloWriter:
Initial work to refactor the GS for Java. To do:

- [x] confirm the primer content so it provides information about version and value

- [x] evaluate the code samples in manual usage - should we use the install examples from there? And, is the Configuration sample overly complex?

There's more work to do here, but this initial PR focuses on just the Getting Started content.


src/includes/getting-started-install/java.mdx
```diff
@@ -1,4 +1,4 @@
-To install the Android SDK, add it your `build.gradle` file:
+To install the SDK, add it your `build.gradle` file:
```
bruno-garcia, line 1: This install is rather bias towards Android usage (assumes Gradle.)

Under "usage" we have install instructions with gradle plus other options:

https://sentry-docs-git-update-java-getting-started.sentry.dev/platforms/java/usage/#installation

Ideally that would be merged with this (that gradle instruction doesn't suggest adding a repository so could take the approach used here instead.

src/includes/getting-started-primer/java.mdx
```diff
@@ -0,0 +1,13 @@
+<Note>
+<markdown>
+
+_Sentry's Java SDK enables reporting messages and errors._
```
bruno-garcia, line 4: This kind of leaves out Sessions (release health) and Performance (which indeed we lack but it's landing in a few weeks).
What about we just leave this line out altogether?

src/includes/getting-started-primer/java.mdx
```diff
@@ -0,0 +1,13 @@
...
+
+Sentry for Java is a collection of modules provided by Sentry. At its core, Sentry for Java provides a raw client for sending events to Sentry. To begin, we **highly recommend** you use one of the logging libraries or framework integrations. We also provide documentation for [manual usage](./usage/)
+
+</markdown>
+</Note>
+
+<!-- TODO: add version and value information, similar to what we have for Go:
```
bruno-garcia, line 11: Good point!

We can state we support Java 1.8 and later here

bruno-garcia (approved):
Thanks for improving this. I left a couple suggested notes or improvements for maybe a follow up PR

src/includes/getting-started-install/java.mdx
PeloWriter: Yes, I saw those more detailed instructions and wondered if they might be a good source. But we can't make it confusing by having too many options. For JS, for example, we document how to install with yarn or npm on the main Getting Started page, but we pulled the CDN instructions and lazy loading into separate pages.

So, I think our best plan is to tell developers what we anticipate is the most used form(s), with a link to other install methods. Do you think that would work here?

src/includes/getting-started-primer/java.mdx
PeloWriter: Let's add Sessions (release health) - we an add performance when we update the page to show how to set that up in Java, since we'll need to update the code sample at that time.

src/includes/getting-started-primer/java.mdx
PeloWriter: Great. I will add that. Is there anything else we'd like to add?

src/includes/getting-started-install/java.mdx
```diff
@@ -13,3 +11,22 @@ dependencies {
     implementation 'io.sentry:sentry:3.1.0'
 }
 ```
+
+```xml {tabTitle:Maven}
```
bruno-garcia, line 15: Is it possible to add a file name too? `pom.xml` is the file name here. I dunno about SBT though
```suggestion
```xml {tabTitle:Maven}
```

src/includes/getting-started-install/java.mdx
```diff
@@ -13,3 +11,22 @@ dependencies {
...
+```scala {tabTitle:SBT}
+libraryDependencies += "io.sentry" % "sentry" % "3.1.0"
+```
+<Note>
+<markdown>
+
+We also support [manual usage](/platforms/java/usage/).
+
+</markdown>
+</Note>
```
bruno-garcia, line 32: Given that this page is covering the manual setup maybe we can drop this line?


```suggestion
```

bruno-garcia (approved):

src/includes/getting-started-install/java.mdx
PeloWriter: I'll give it a go.

src/includes/getting-started-install/java.mdx
PeloWriter: Sure.

PR #2563, opened 2020-10-21
feat(actions): Create new workflow to bump sentry-api-schema SHA

NisanthanNanthakumar:
## Objective
Refactor our existing bump SHA action in `sentry-api-docs`, so that we have something similar to `./bin/bump-sentry` in getsentry.

This workflow listens to a workflow dispatch event and takes in a SHA. Then it will search and replace the with this latest SHA.

Concurrent PR: https://github.com/getsentry/sentry-api-schema/pull/32

### DIFF
--- a/.github/workflows/bump-openapi.yml
+++ b/.github/workflows/bump-openapi.yml
@@ -1,0 +1,31 @@
+name: bump-openapi
+on:
+  workflow_dispatch:
+    inputs:
+      sentry-api-schema-sha:
+        description: 'The "sentry-api-schema" SHA to run against'
+        required: false
+  push:
+    branches:
+      - master
+
+jobs:
+  bump:
+    name: Bump SHA referencing "sentry-api-schema" to ${{ github.event.inputs.sentry-api-schema-sha }}
+    runs-on: ubuntu-16.04
+    if: ${{ github.event.inputs.sentry-api-schema-sha != '' }}
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Bump latest SHA
+        sed -i "" 's|^const SENTRY_API_SCHEMA_SHA =.*$|const SENTRY_API_SCHEMA_SHA = "${{ github.event.inputs.sentry-api-schema-sha }}"|g' src/gatsby/utils/resolveOpenAPI.ts
+
+    - name: Git Commit & Push
+      uses: stefanzweifel/git-auto-commit-action@v4
+      with:
+        repository: sentry-docs
+        branch: master
+        commit_message: Latest SHA from getsentry/sentry-api-schema
+        commit_user_email: bot@getsentry.com
+        commit_user_name: openapi-getsentry-bot
--- a/.github/workflows/bump-openapi.yml
+++ b/.github/workflows/bump-openapi.yml
@@ -1,4 +1,6 @@
 name: bump-openapi
+# Deploys depend on the workflow file name: 'bump-openapi.yml'
+# You should not independently update it without updating the corresponding workflow in getsentry/sentry-api-schema
 on:
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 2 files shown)


.github/workflows/bump-openapi.yml
```diff
@@ -0,0 +1,31 @@
+name: bump-openapi
+on:
+  workflow_dispatch:
+    inputs:
+      sentry-api-schema-sha:
+        description: 'The "sentry-api-schema" SHA to run against'
+        required: false
+  push:
+    branches:
+      - master
```
billyvg, line 10: Should this run on pushes to master?

src/gatsby/utils/resolveOpenAPI.ts
```diff
@@ -1,6 +1,8 @@
 import axios from "axios";
 import { promises as fs } from "fs";
 
+const SENTRY_API_SCHEMA_SHA = "03ccef5d80c6e636994e0594312778e1186ba41c"
```
billyvg, line 4: I would add a comment here re: workflow

.github/workflows/bump-openapi.yml
```diff
@@ -0,0 +1,31 @@
...
+        sed -i "" 's|^const SENTRY_API_SCHEMA_SHA =.*$|const SENTRY_API_SCHEMA_SHA = "${{ github.event.inputs.sentry-api-schema-sha }}"|g' src/gatsby/utils/resolveOpenAPI.ts
+
+    - name: Git Commit & Push
+      uses: stefanzweifel/git-auto-commit-action@v4
+      with:
+        repository: sentry-docs
+        branch: master
+        commit_message: Latest SHA from getsentry/sentry-api-schema
+        commit_user_email: bot@getsentry.com
+        commit_user_name: openapi-getsentry-bot
```
billyvg, line 31: 1) We should either fork or use a commit hash for external actions (that are not by github), the reason being that `v4` here refers to either a tag or branch in github, both are mutable, so can potentially break our builds.

2) We don't need this action at all, you can use native git commands if you configure the name/email first see https://github.com/getsentry/sentry/blob/cc7b9015a7f43c9e5a909e28c69502c627dd6c00/.github/workflows/release.yml#L49-L50 for an example

3) This won't work if master is a protected branch

.github/workflows/bump-openapi.yml
```diff
@@ -0,0 +1,31 @@
...
+  bump:
+    name: Bump SHA referencing "sentry-api-schema" to ${{ github.event.inputs.sentry-api-schema-sha }}
+    runs-on: ubuntu-16.04
+    if: ${{ github.event.inputs.sentry-api-schema-sha != '' }}
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Bump latest SHA
+        sed -i "" 's|^const SENTRY_API_SCHEMA_SHA =.*$|const SENTRY_API_SCHEMA_SHA = "${{ github.event.inputs.sentry-api-schema-sha }}"|g' src/gatsby/utils/resolveOpenAPI.ts
```
billyvg, line 22: you're missing a `run:`

.github/workflows/bump-openapi.yml
NisanthanNanthakumar: hmm my mistake, it should only run on the workflow dispatch trigger

.github/workflows/bump-openapi.yml
```diff
@@ -0,0 +1,30 @@
...
+
+      - name: Bump latest SHA
+        run: |
+          sed -i "" 's|^const SENTRY_API_SCHEMA_SHA =.*$|const SENTRY_API_SCHEMA_SHA = "${{ github.event.inputs.sentry-api-schema-sha }}"|g' src/gatsby/utils/resolveOpenAPI.ts
+
+      - name: Git Commit & Push
+        run: |
+          git config user.name openapi-getsentry-bot
+          git config user.email bot@getsentry.com
+          git add .
```
billyvg, line 28: Should this be more specific in case somehow something else gets modified?

billyvg (approved):
LGTM

PR #2571, opened 2020-10-22
fix(dev): Add note about `supported`/`notSupported` for subtrees

lobsterkatie:
All kinds of crazy things happen if you don't do this.


PeloWriter (approved):
thanks.

PR #2577, opened 2020-10-23
feat: Add release/tag examples for minidumps

dcramer:


jan-auer (approved):
Examples are fine, but it doesn't work yet. 

https://sentry-docs-git-feat-improve-minidumps.sentry.dev/platforms/native/guides/minidumps/enriching-events/tags/ shows the C++ example and not the minidumps one. I guess because minidumps is not it's own platform:

https://sentry-docs-git-feat-improve-minidumps.sentry.dev/platforms/native/guides/minidumps/enriching-events/tags/

Also, there's no place where releases are rendered. I guess we need to enable the release health page?

dcramer:
@jan-auer mm yeah good catch i named the files wrong

re: releases.. pretty sure thats still on the backlog to fix it.. it seems to be missing in a lot of platforms cc @PeloWriter

PeloWriter:
Yes, we've held off due to changes that are expected to impact this area of content.

dcramer:
@PeloWriter unlikely to have any change from any SDK point of view so no need to block

dcramer:
aside we have this https://docs.sentry.io/platforms/native/guides/minidumps/configuration/releases/ (which is wrong today)

dcramer:
also the above isnt discoverable via any links just fyi

jan-auer:
I'll look into moving minidumps out into a top-level platform as per recent discussion next week. Thanks, @dcramer!

PR #2579, opened 2020-10-23
ref(apple): Use tabTitle for code snippets

philipphofmann:
Use tabTitle for both Swift and Objective-C so code snippets display
Swift and Objective-C instead of Swift and Objc.


dcramer:
What are they displaying currently? They should by default display the language name, which can be tweaked in the CodeTabs component. Would rather we fix it there vs changing this in every code sample

bruno-garcia:
Before: 
image

After:
image

bruno-garcia (approved):
LGTM

PR #2582, opened 2020-10-23
ref: Remove markdown tags

dcramer:
Unclear where we received the idea that this was required, but it seems to no-op and pass through the tag.

### DIFF
--- a/src/docs/contributing/pages/index.mdx
+++ b/src/docs/contributing/pages/index.mdx
@@ -12,16 +12,16 @@ Documentation is written in Markdown (via Remark) and MDX.
 
 :pray: that MDX v2 fixes this.
 
-MDX has its flaws. When rendering components, any text inside of them is treated as raw text (_not_ markdown). To work around this you can use the `<markdown>` tag, but it also has its issues. Generally speaking, put an empty line after the opening tag, and before the closing tag.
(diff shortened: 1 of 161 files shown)


dcramer:
Few minor unrelated tweaks in here (alert -> note, shuffle some content to be in more appropriate position)

src/docs/contributing/pages/index.mdx
```diff
@@ -12,16 +12,16 @@ Documentation is written in Markdown (via Remark) and MDX.
 
 :pray: that MDX v2 fixes this.
 
-MDX has its flaws. When rendering components, any text inside of them is treated as raw text (_not_ markdown). To work around this you can use the `<markdown>` tag, but it also has its issues. Generally speaking, put an empty line after the opening tag, and before the closing tag.
+MDX has its flaws. When rendering components, any text inside of them is treated as raw text (_not_ markdown). To work around this you can use the `` tag, but it also has its issues. Generally speaking, put an empty line after the opening tag, and before the closing tag.
```
PeloWriter, line 15: Does the second sentence need to be removed?
```suggestion
MDX has its flaws. When rendering components, any text inside of them is treated as raw text (_not_ markdown). . Put an empty line after the opening tag, and before the closing tag.
```

PeloWriter (approved):
One update to the contributor pages, all else looks good. Thanks.

PR #2585, opened 2020-10-23
feat: Next.js documentation

dcramer:


src/includes/getting-started-install/javascript.nextjs.mdx
```diff
@@ -0,0 +1,7 @@
+```bash {tabTitle:npm}
+npm install --save @next/plugin-sentry @next/sentry-source-maps
```
dcramer, line 2: save is here because we still keep it everywhere in docs - future item will be to remove if determined never applicable

src/platforms/common/index.mdx
```diff
@@ -18,7 +18,7 @@ Sentry captures data by using an SDK within your application’s runtime.
 
 ## Configure
 
-Configure Sentry as early as possible in your app.
+Configuration should happen as early as possible in your application's lifecycle.
```
dcramer, line 21: this language fits a lot better in a generic context

dcramer:
This is a draft. not ready to merge.

dcramer:
1. we should probably hide the configuration options page, but thats going to be poor UX. ideally vercel lets us hijack the Sentry.init call somehow.

2. we need to be able to do something like `import Sentry from "@next/plugin-sentry"` and have that proxy to browser/node so they can set tags/etc

3. need to expand sourcemaps documentation with next/vercel-specific instruction overrides

dcramer:
pretty sure the sourcemaps example is wrong, but ideally we wouldnt need it at all. its just another place for users to make a mistake.

added wizard content for onboarding - we need a next.js icon but i pinged design

src/platforms/javascript/common/sourcemaps/troubleshooting.mdx
```diff
@@ -2,9 +2,9 @@
 title: Troubleshooting
 redirect_from:
   - /platforms/javascript/sourcemaps/troubleshooting/
----
-
-Source maps can sometimes be tricky to get going. If you’re having trouble:
+notSupported:
+  - javascript.nextjs
+---Source maps can sometimes be tricky to get going. If you’re having trouble:
```
PeloWriter, line 7:
```suggestion
---
Source maps can sometimes be tricky to get going. If you’re having trouble:
```

src/includes/getting-started-config/javascript.nextjs.mdx
```diff
@@ -0,0 +1,33 @@
+If you're using [Vercel](https://vercel.com) you just need to install [Sentry's Vercel integration](/product/integrations/vercel/) and we'll automatically configure your environment for you.
```
PeloWriter, line 1: minor smoothing
```suggestion
If you're using [Vercel](https://vercel.com) you only need to install [Sentry's Vercel integration](/product/integrations/vercel/), then we'll automatically configure your environment for you.
```

PeloWriter (commented):
one minor edit; one bit of smoothing.

src/platforms/javascript/common/sourcemaps/troubleshooting.mdx
dcramer: oops good catch.. auto formatter broke it

dcramer:
We can now add code samples for contexts:

`import * as Sentry from "@next/plugin-sentry"`

PR #2591, opened 2020-10-24
ref(tracing): Make transaction sampling docs more generic

lobsterkatie:
This is a first step to enabling more than just JS to use the `tracesSampler` content.


PeloWriter (approved):
Thank you.

PR #2592, opened 2020-10-24
feat(tracing): Add python includes for transaction sampling

lobsterkatie:
This add includes for both the filtering config page and the performance/sampling page for Python, and adds Python to the list of supported platforms for the latter, so that it will show up in the Python docs.


HazAT (approved):
Good refactors

HazAT:
Next step, add PHP in this structure (as mentioned).

PR #2593, opened 2020-10-26
fix(PHP): Correctly case PHP on code tabs

lobsterkatie:
s/Php/PHP

Also, I alphabetized the relevant array, to make it easier to find stuff quickly.


PeloWriter (approved):
Thank you.

PR #2598, opened 2020-10-27
Initial refactor

PeloWriter:
Refactor of Getting Started page. This refactor enables use of the common content. In addition, the content:

- [ ] uses Java fallbacks for the Configuration and Verification step

- [ ] moves usage (which appears to be about configuration) into an example configuration page as a subpage of configuration - feedback particularly welcome on this change.

### DIFF
--- a/src/includes/getting-started-install/java.log4j2.mdx
+++ b/src/includes/getting-started-install/java.log4j2.mdx
@@ -1,0 +1,17 @@
+```xml {tabTitle:Maven}
+<dependency>
+    <groupId>io.sentry</groupId>
+    <artifactId>sentry-log4j2</artifactId>
+    <version>{{ packages.version('sentry.java', '3.1.1') }}</version>
+</dependency>
+```
+
+```groovy {tabTitle:Gradle}
+implementation 'io.sentry:sentry-log4j2:{{ packages.version('sentry.java', '3.1.1') }}'
+```
+
+```scala {tabTitle: SBT}
+libraryDependencies += "io.sentry" % "sentry-log4j2" % "{{ packages.version('sentry.java', '3.1.1') }}"
+```
+
+For other dependency managers see the [central Maven repository](https://search.maven.org/artifact/io.sentry/sentry-log4j2).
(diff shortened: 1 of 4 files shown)


bruno-garcia:
There's an issue introduced by this PR. The "usage" part seem to be suboptimal but it does have 1 thing that is special to log4j2 so we need to keep it, and with this change it doesn't seem to be there anymore.

-> https://docs.sentry.io/platforms/java/guides/log4j2/#usage

Basically this "usage" thing is a log4j2 specific to init the SDK.

Important:

> If the DSN is not present in the log4j2.xml configuration, Sentry will attempt to read it from the system property sentry.dsn, environment variable SENTRY_DSN or the dsn property in sentry.properties file. See the configuration page for more details on external configuration.

Basically this integration has some specific configuration (the stuff under _usage_) and it optionally can init the SDK for you. You do that by defining a `dsn` in the `appender.xml` file. If you choose to add the `dsn` to that config, then you don't need to call `Sentry.init` by yourself.

This means: This integration inherits everything that Java docs includes, but it introduces its own way to initialize (optionally). Besides that, it has its own config things (besides the default Java configuration information).

This section: https://docs.sentry.io/platforms/java/guides/log4j2/#minimum-log-level

bruno-garcia:
These things now show up on "Example configuration" but that's not ideal since this should be the first thing users see here.

The bottom part, "mapped tags" could be moved to the "Advanced usage": https://docs.sentry.io/platforms/java/guides/log4j2/#mapped-tags

But the main appender config should be the landing page for log4j2 imo.

src/includes/getting-started-config/java.log4j2.mdx
```diff
@@ -0,0 +1,56 @@
...
+            <AppenderRef ref="Sentry"/>
+            <AppenderRef ref="Console"/>
+        </Root>
+    </Loggers>
+</Configuration>
+```
+
+### DSN Configuration
+
+Note that **you need to configure your DSN** (client key).
```
bruno-garcia, line 30:
```suggestion
Note that **you need to configure your DSN** (client key) only if you wish to initialize the SDK through the log4j2 integration. If you're planning to use `Sentry.init` in order to provide configuration such as the `beforeSend` callback, you **should not** provide in both places. Just leave it out of the appender configuration in that case.
```

bruno-garcia (approved):
👍 
One clarification added

src/includes/getting-started-config/java.log4j2.mdx
PeloWriter:
```suggestion
Note that **you need to configure your DSN** (client key) only if you wish to initialize the SDK through the log4j2 integration. If you're planning to use `Sentry.init` to provide configuration, such as by using the `beforeSend` callback, you **should not** provide the DSN in both `Sentry.init` and the appender configuration; just leave it out of the appender configuration in this case.
```

PeloWriter:
Thanks. Modified the language slightly for clarity, but we're good to go with this refactor.

PR #2603, opened 2020-10-27
Update wizard content for JavaScript

PeloWriter:
This PR uses the updated content from the Getting Started page for JS and inserts it into both the browser and JS wizard pages.

### DIFF
--- a/src/wizard/javascript/browser.md
+++ b/src/wizard/javascript/browser.md
@@ -31,12 +31,15 @@ Sentry.init({
   tracesSampleRate: 1.0,
 });
 ```
+
 We recommend adjusting the value of `tracesSampleRate` in production. Learn more about configuring sampling in our [full documentation](https://docs.sentry.io/platforms/javascript/performance/sampling/).
 
-This snippet includes an intentional error, so you can test that everything is working as soon as you set it up:
+Then create an intentional error, so you can test that everything is working:
 
 ```js
 myUndefinedFunction();
 ```
 
-Sentry will send an email alerting you to this intentional error; follow the instructions in the email to resolve the error.
+If you're new to Sentry, use the email alert to access your account and complete a product tour.
+
+If you're an existing user and have disabled alerts, you won't receive this email.  
--- a/src/wizard/javascript/browser.md
+++ b/src/wizard/javascript/browser.md
@@ -5,28 +5,39 @@ support_level: production
 type: language
 ---
 
-The quickest way to get started is to use the CDN hosted version of the JavaScript browser SDK:
+Install our JavaScript browser SDK using either `yarn` or `npm`:
 
-```html
-<script
-  src="https://browser.sentry-cdn.com/{{ packages.version('sentry.javascript.browser') }}/bundle.min.js"
-  integrity="sha384-{{ packages.checksum('sentry.javascript.browser', 'bundle.min.js', 'sha384-base64') }}"
-  crossorigin="anonymous"
-></script>
+```bash {tabTitle: ESM}
+# Using yarn
+yarn add @sentry/browser @sentry/tracing
+# Using npm
+npm install --save @sentry/browser @sentry/tracing
 ```
 
-You should `init` the Sentry Browser SDK as soon as possible during your page load:
+We also support alternate [installation methods](/platforms/javascript/install/).
+
+`init` the Sentry Browser SDK as soon as possible during your page load:
 
 ```javascript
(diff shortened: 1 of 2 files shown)


src/wizard/javascript/browser.md
```diff
@@ -5,28 +5,39 @@ support_level: production
...
+import * as Sentry from "@sentry/browser";
+import { Integrations } from "@sentry/tracing";
+
+Sentry.init({
+  dsn: '___PUBLIC_DSN___',
+  integrations: [
+    new Integrations.BrowserTracing(),
+  ],
+
+  // We recommend adjusting this value in production, or using tracesSampler
```
benvinegar, line 31: > or using tracesSampler

This seems tricky to introduce at this stage. What is `tracesSampler`? How do I use it? Where do I find out more about it?

I'd sooner put a note that says (perhaps outside of the code snippet) that says: "It is recommended to lower this value in production. You can find out more about configuring sampling in our full documentation."

src/wizard/javascript/browser.md
```diff
@@ -5,28 +5,39 @@ support_level: production
...
-
-Calling an undefined function will throw an exception:
+This snippet includes an intentional error, so you can test that everything is working as soon as you set it up:
 
 ```js
 myUndefinedFunction();
 ```
 
-You can verify the function caused an error by checking your browser console.
+To view and resolve this intentional error, log into sentry.io and open your project. Clicking on the error's title will open a page where you can see detailed information and mark it as resolved.
```
benvinegar, line 43: I feel like we're introducing concepts that people aren't ready to wrap their heads around just yet. What is an `error title`? I'm already logged in; why am I being instructed to open sentry.io and log in? 

Note that if they send this error, they will get an email that will bring them into their account, we don't need to give them instructions that they'll forget.

benvinegar (changes requested):

src/wizard/javascript/browser.md
PeloWriter: I agree in this context. Will edit to:
"Sentry will send you an email alerting you to this intentional error; follow the instructions in the email to resolve the error."

src/wizard/javascript/browser.md
PeloWriter: In our docs, we want everything in the code snippet. But here, I think it makes sense to pull it out. As you note, too much isn't useful in the absence of supporting content.

Will edit to:
"We recommend lowering this value in production. Learn more about configuring sampling in our full documentation." with a link to https://docs.sentry.io/platforms/javascript/performance/sampling/

src/wizard/javascript/browser.md
```diff
@@ -28,16 +28,15 @@ Sentry.init({
...
+We recommend adjusting the value of `tracesSampleRate` in production. Learn more about configuring sampling in our [full documentation](https://docs.sentry.io/platforms/javascript/performance/sampling/).
 
 This snippet includes an intentional error, so you can test that everything is working as soon as you set it up:
 
 ```js
 myUndefinedFunction();
 ```
 
-To view and resolve this intentional error, log into sentry.io and open your project. Clicking on the error's title will open a page where you can see detailed information and mark it as resolved.
+Sentry will send an email alerting you to this intentional error; follow the instructions in the email to resolve the error.
```
benvinegar, line 42: I don't know that we should be recommending that people "mark it as resolved". I mean, the test event, I guess?

I just don't know that we should be describing the workflow right now. The app does a tour once you land on the Issues page. We just want them to send an event.

> Sentry will send an email alerting you to this intentional error; follow the instructions in the email to resolve the error.

Note this wizard is for both new users and existing users creating a new project (and they can disable alerts). So I don't think we should communicate "you will get an email". New users will, existing users may not.

benvinegar (changes requested):

src/wizard/javascript/browser.md
PeloWriter: Yes, the test event. You're correct that the language is unclear here - thanks for a second set of eyes on this.  

I am a little unclear about the email. In the previous comment, you mentioned that "Note that if they send this error, they will get an email that will bring them into their account, we don't need to give them instructions that they'll forget."

So, perhaps what we mean to say here is:
"If you're new to Sentry, we'll send you an email to access your account where you can complete a tour of the Issues page." And we don't need more for existing users since we can assume they know the flow here. What do you think?

benvinegar (approved):

PR #2604, opened 2020-10-27
fix: missing options from .NET platform. 

lucas-zimerman:
In-app-include and in-app-exclude are supported by .NET SDK so this fix enabled the description of those two fields when filtering by the .NET documentation.


bruno-garcia (approved):
Thanks @lucas-zimerman

bruno-garcia:
@PeloWriter I think the issue here is taht it was named `csharp` and now it's `dotnet`?

bruno-garcia:
Happens that both are in there now though

PR #2605, opened 2020-10-27
Update Java Wizard

PeloWriter:
Update the wizard for Java using the updated Getting Started content.

### DIFF
--- a/src/wizard/java/index.md
+++ b/src/wizard/java/index.md
@@ -6,52 +6,65 @@ type: language
 ---
 
 <Alert level="info">
-    If you use <strong>Spring Boot</strong>, <strong>Logback</strong> or <strong>Log4j2</strong>, we recommend using a dedicated integration. Please visit <a href="https://docs.sentry.io/platforms/java/">Sentry Java</a> documentation for installation instructions.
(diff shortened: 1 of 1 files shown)


src/wizard/java/index.md
```diff
@@ -6,52 +6,65 @@ type: language
 ---
 
 <Alert level="info">
-    If you use <strong>Spring Boot</strong>, <strong>Logback</strong> or <strong>Log4j2</strong>, we recommend using a dedicated integration. Please visit <a href="https://docs.sentry.io/platforms/java/">Sentry Java</a> documentation for installation instructions.
+    Sentry for Java is a collection of modules provided by Sentry; it supports Java 1.8 and above. At its core, Sentry for Java provides a raw client for sending events to Sentry. If you use <strong>Spring Boot</strong>, <strong>Spring</strong><strong>, Logback</strong>, or <strong>Log4j2</strong>, we recommend using a dedicated integration. Please visit <a href="https://docs.sentry.io/platforms/java/">Sentry Java</a> documentation for installation instructions.
```
bruno-garcia, line 9: Would be nice to link to another wizard instead. Or maybe have a toggle in this wizard and bundle them all in this page?

Spring boot + logback is different than those integrations alone.

bruno-garcia (approved):
Nice improvement. I left one note

src/wizard/java/index.md
PeloWriter: Spring Boot wizard exists, but it's not updated yet. The others don't have wizards (yet). 
So, let's do this for now:
```suggestion
    Sentry for Java is a collection of modules provided by Sentry; it supports Java 1.8 and above. At its core, Sentry for Java provides a raw client for sending events to Sentry. If you use <strong>Spring Boot</strong>, <strong>Spring</strong><strong>, Logback</strong>, or <strong>Log4j2</strong>, we recommend visiting our <a href="https://docs.sentry.io/platforms/java/">Sentry Java</a> documentation for installation instructions.
```
and we can sort out how/whether we can link to other wizards once they're available.

PR #2607, opened 2020-10-28
ref(tracing): Improve note about sampling when testing on Python page

lobsterkatie:
This removes the `<Note>` component (so they text doesn't show up in grey) and pulls in more content from the JS version of the page, to bring them more in line.


PeloWriter (approved):
Thank you.

PR #2608, opened 2020-10-28
Point Android users to Gradle

bruno-garcia:


src/platforms/common/data-management/debug-files.mdx
```diff
@@ -5,7 +5,6 @@ redirect_from:
   - /workflow/debug-files/
 supported:
   - native
```
bruno-garcia, line 7:
```suggestion
  - native
  - android
```

marandaneto (approved):

src/platforms/common/data-management/debug-files.mdx
```diff
@@ -214,6 +214,12 @@ traces.
 
 ### ProGuard Mappings
 
+<PlatformSection supported={["android"]}>
+<Note>
+  The recommended method for [Android users is to use the Gradle plugin](platforms/android/proguard/).
+</Note>
+</PlatformSection>
```
marandaneto, line 221:
```suggestion
<PlatformSection supported={["android"]}>

<Note>

  The recommended method for [Android users is to use the Gradle plugin](/platforms/android/proguard/).
  
</Note>

</PlatformSection>
```

marandaneto: I guess we'd need spaces between them because the preview is broken, trying this out

PeloWriter: yep, you need spaces above and below the <Note></Note> syntax. Otherwise, it doesn't work as expected.

PR #2618, opened 2020-10-29
[JS / Tracing] Reverse the order of pathname replacements

stayallive:
Otherwise the digits replacement happens inside hashes (since they also contain digits). 

Also using `replaceAll` otherwise only the first match will be replaced.

I've also added a `/` prefix to make sure you get full path segments and not random numbers in slugs for example.

### DIFF
--- a/src/platforms/javascript/common/performance/index.mdx
+++ b/src/platforms/javascript/common/performance/index.mdx
@@ -217,8 +217,8 @@ Sentry.init({
           // route template here. We don't have one right now, so do some basic
           // parameter replacements.
           name: location.pathname
-            .replace(/\d+/g, "<digits>")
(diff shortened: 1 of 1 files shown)


stayallive:
^ I think that message means I'm trying to deploy to the review branch... have not actually tried anything except added a commit to the PR.

stayallive:
I still believe this would be a good change.

lobsterkatie (approved):
Looks good - thanks!

PeloWriter:
Thank you for this fix. Since we've been moving this content into our common content model, it made fixing the conflicts on this PR a bit tricky. As a result, I opened  https://github.com/getsentry/sentry-docs/pull/2906 to capture this change.

PR #2628, opened 2020-10-30
Fix confusing grammar

cTxplorer:
I had to read that particular sentence thrice to understand what it is the intended meaning. 
Based on my understanding, I have now updated the wording. Cheers!


PeloWriter (approved):
Thank you.

PR #2634, opened 2020-10-30
Fix mention of releases not allowed to have single period

torarnv:


matejminar:
Related to: https://github.com/getsentry/sentry-docs/pull/2454

torarnv:
Aha! So the real meaning of that line is that release can't be === to "." or ".." etc? Not that it can't contain one? The "use a" part threw me off there. Sounds like #2454 solves this already 😊

PR #2636, opened 2020-10-30
Update data management

PeloWriter:
This PR 

- [ ] splits data scrubbing information into SDK content and product settings content
- [ ] *maintains* the current structure for grouping (since it works in combination, so makes sense to keep together)
- [ ] creates <PlatformContent includePath=> structure for grouping examples (with defaults to display in the absence of an example)
- [ ] edits the affected files to smooth and clarify

Feedback welcome.


src/docs/product/data-management-settings/server-side-scrubbing.mdx
```diff
@@ -0,0 +1,55 @@
...
+  - auth_token
+  - credentials
+  - mysql_pwd
+  - stripetoken
+  - card[number]
+- Any keys which match values that you’ve added to the list of additional fields in your Project Settings.
+
+Sentry will redact arrays and literal values, but not entire objects. For example:
+
+```bash
```
untitaker, line 29:
```suggestion
```javascript
```

untitaker (approved):
i have no way of quantifying how many links from external sites this will break. It does not have a vercel redirect...

untitaker:
That said one major concern I have is that we will have to end up doing something similar to server-side scrubbing as this PR introduces for grouping enhancements: platform-specific examples.

PeloWriter:
I will add a Vercel redirect for server-side-scrubbing.mdx (thanks for alerting me to the lack of it). I can also use this PR to build out platform-specific example structure for server-side scrubbing

PeloWriter:
Actually looks like vercel.json has it. And the platform-specific examples out of product raise an intriguing design issue. Will deploy these changes and we can re-visit as/when needed.

PR #2648, opened 2020-11-02
Fix: public dsn

MimiDumpling:
Changed the example to public dsn.


tonyo (commented):
I think we want the same change on line 86 (Bash Hook section)

PR #2658, opened 2020-11-05
add missing this (context) for android getting started config

marandaneto:
I guess a few more are missing the `this` or `context`, need to search for it, it's a typo from copying the examples from Java which doesn't require the context.


philipphofmann (approved):
LGTM

PR #2664, opened 2020-11-05
fix(tracing): Add missing `return`s to snippet

### LINKED ISSUE #2660: Dynamic Sampling Function intro - missing returns on Python documentation
In the document:

https://docs.sentry.io/platforms/python/performance/sampling/

for Python instead of:
``` 
def traces_sampler(sampling_context):
    # Examine provided context data (including parent decision, if any)
    # along with anything in the global namespace to compute the sample rate
    # or sampling decision for this transaction

    if "...":
        # These are important - take a big sample
        0.5
    elif "...":
        # These are less important or happen much more frequently - only take 1%
        0.01
    elif "...":
        # These aren't something worth tracking - drop all transactions like this
        0
    else:
        # Default sample rate
        0.1

sentry_sdk.init(
    # ...

    traces_sampler=traces_sampler,
)
```

shall be

```
def traces_sampler(sampling_context):
    # Examine provided context data (including parent decision, if any)
    # along with anything in the global namespace to compute the sample rate
    # or sampling decision for this transaction

    if "...":
        # These are important - take a big sample
        return 0.5
    elif "...":
        # These are less important or happen much more frequently - only take 1%
        return 0.01
    elif "...":
        # These aren't something worth tracking - drop all transactions like this
        return 0
    else:
        # Default sample rate
        return 0.1

sentry_sdk.init(
    # ...

    traces_sampler=traces_sampler,
)
```

It is stated in documentation that:
`For convenience, the function can also return a boolean.`, so returns shall be in place.

However, I could not find where to drop PR for Python example

lobsterkatie:
Somehow all the `return`s in the return statements went missing.

Fixes #2660.

### DIFF
--- a/src/includes/performance/traces-sampler-as-sampler/python.mdx
+++ b/src/includes/performance/traces-sampler-as-sampler/python.mdx
@@ -6,16 +6,16 @@ def traces_sampler(sampling_context):
 
     if "...":
         # These are important - take a big sample
-        0.5
+        return 0.5
     elif "...":
         # These are less important or happen much more frequently - only take 1%
-        0.01
+        return 0.01
     elif "...":
         # These aren't something worth tracking - drop all transactions like this
-        0
+        return 0
     else:
         # Default sample rate
-        0.1
+        return 0.1
 
 sentry_sdk.init(
     # ...


untitaker (approved):

PeloWriter (approved):
Thank you.

PR #2665, opened 2020-11-05
Update: bolded headers to h3, h4, h5

MimiDumpling:
Updated bolded headers for consistency with h3, h4, and h5. H5s used sparingly.

### DIFF
--- a/src/docs/product/accounts/pricing.mdx
+++ b/src/docs/product/accounts/pricing.mdx
@@ -54,7 +54,7 @@ If your projects have a high volume of events, you can control how many errors S
 
 ## FAQ
 
-**Is there detailed information for each plan?**
+#### Is there detailed information for each plan?
 
 Yes, the following table details costs for errors, transactions, and attachments.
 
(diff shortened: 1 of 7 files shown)


src/docs/product/accounts/pricing.mdx
```diff
@@ -54,7 +54,7 @@ If your projects have a high volume of events, you can control how many errors S
 
 ## FAQ
 
-**Is there detailed information for each plan?**
+#### Is there detailed information for each plan?
```
philipphofmann, line 57: Shouldn't this be `###` if FAQ has `##`?

src/docs/product/releases/health.mdx
```diff
@@ -102,25 +102,31 @@ Data filters are only partially supported for session events, and they might be
 
 ### Equivalent Filters
 
-**release filter**
-This filter discards events of a particular release. If sessions are started or updated and would match this filter, sessions are silently discarded entirely. This is consistent with error events.
+Release Filter
```
philipphofmann, line 105: The changes in this file make it way cleaner. I like it a lot 👏

philipphofmann (approved):
One small comment from my side, but I like the changes. Makes the docs a bit better IMO 👏

src/docs/product/accounts/pricing.mdx
PeloWriter: Structurally, yes. But in this case, the change was made to fix a bolding issue, so an H4 is fine. Thanks for reviewing.

PR #2666, opened 2020-11-05
fix(ci): Invalidate caches when node version changes

untitaker:
Getting the following error in CI:

```
Error: Missing binding /home/runner/work/sentry-docs/sentry-docs/node_modules/node-sass/vendor/linux-x64-83/binding.node
Node Sass could not find a binding for your current environment: Linux 64-bit with Node.js 14.x

Found bindings for the following environments:
  - Linux 64-bit with Node.js 12.x

This usually happens because your environment has changed since running `npm install`.
Run `npm rebuild node-sass` to download the binding for your current environment.
    at module.exports (/home/runner/work/sentry-docs/sentry-docs/node_modules/node-sass/lib/binding.js:15:13)
    at Object.<anonymous> (/home/runner/work/sentry-docs/sentry-docs/node_modules/node-sass/lib/index.js:14:35)
    at Module._compile (/home/runner/work/sentry-docs/sentry-docs/node_modules/gatsby/node_modules/v8-compile-cache/v8-compile-cache.js:178:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (/home/runner/work/sentry-docs/sentry-docs/node_modules/gatsby/node_modules/v8-compile-cache/v8-compile-cache.js:159:20)
    at getDefaultSassImpl (/home/runner/work/sentry-docs/sentry-docs/node_modules/sass-loader/dist/index.js:198:10)
    at Object.loader (/home/runner/work/sentry-docs/sentry-docs/node_modules/sass-loader/dist/index.js:80:29)
error Generating JavaScript bundles failed
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,8 @@ jobs:
         with:
           path: ${{ github.workspace }}/node_modules
           key: ${{ runner.os }}-${{ hashfiles('**/yarn.lock') }}
-      - run: yarn install
+      # Added --force as a workaround for https://github.com/yarnpkg/yarn/issues/4867
+      - run: yarn install --force
         if: steps.cache.outputs.cache-hit != 'true'
 
       - uses: actions/cache@v2


.github/workflows/test.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           path: |
             .cache
             public
-          key: ${{ runner.os }}-${{ hashfiles('**/yarn.lock') }}-gatsby-build
+          key: ${{ runner.os }}-${{ hashfiles('**/yarn.lock') }}-gatsby-build-
```
Zylphrex, line 30:
```suggestion
          key: ${{ runner.os }}-${{ hashfiles('**/yarn.lock') }}-gatsby-build-${{ env.NODE_VERSION }}
```

untitaker:
@BYK I am adding you as a reviewer here because this appears to be a yarn issue or an issue with how we cache yarn stuff, and I have really no idea what I am doing.

I basically wiped caches but at the second push to the PR it turned out that even the new caches were created broken.

I found a yarn bugreport that tells me to `--force` install and so that's what I am trying out.

.github/workflows/test.yml
```diff
@@ -19,7 +19,8 @@ jobs:
         with:
           path: ${{ github.workspace }}/node_modules
           key: ${{ runner.os }}-${{ hashfiles('**/yarn.lock') }}
```
BYK, line 21:
```suggestion
          key: ${{ runner.os }}-${{ hashfiles('**/package.json')-${{ hashfiles('**/yarn.lock') }}
```

BYK (changes requested):
The real fix is making the cache depend on node and yarn versions. `node_modules` consistency is only guaranteed across the same node and yarn versions with _both_ `package.json` and `yarn.lock` files. You can read more at https://classic.yarnpkg.com/blog/2017/05/31/determinism/

Since we use `volta` which uses the fields in `package.json` to get yarn and node versions, the proper fix is to simply include the hash of the `package.json` file in the cache key.

untitaker:
@BYK thanks for your input. We don't currently seem to use volta in gha but I think we should. The error happened because GitHub decided to upgrade the default node in their env. I will fix that as well. I still don't quite understand why my fix wasn't sufficient but pinning node via volta probably will fix it.

BYK:
Oh, your fix was completely sufficient it was just overkill and obliterating all the benefits of caching 😃

untitaker:
>Oh, your fix was completely sufficient

Apparently not... I still get the same error after pushing an empty commit to this PR. That's why I am confused.

BYK:
>Apparently not... I still get the same error after pushing an empty commit to this PR. That's why I am confused.

Well, that's a bit interesting but I really don't feel like digging into old Yarn code right now 😁 If you insist, I may do that though.

untitaker:
Thanks and no worries, ignore then. I'll figure it out in a new pr

PR #2667, opened 2020-11-06
wpf: subscribe to event

bruno-garcia:
Changing title (the original is too long)
adding explicit event subscription
/cc @Tyrrrz

### DIFF
--- a/src/platforms/dotnet/guides/wpf/index.mdx
+++ b/src/platforms/dotnet/guides/wpf/index.mdx
@@ -1,5 +1,5 @@
 ---
-title: Windows Presentation Foundation
+title: WPF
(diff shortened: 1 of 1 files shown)


Tyrrrz (approved):

Tyrrrz:
Some WPF/MVVM frameworks also subscribe to that event and might do so first. Hopefully, the developers who use those frameworks are aware of that.

MimiDumpling:
Changing the title to WPF sounds good because I think that version of the name is the industry standard.

MimiDumpling (approved):

bruno-garcia:
@PeloWriter I've tried rerunning CI but it fails consistently. Any idea what's going on?

PR #2668, opened 2020-11-09
chore: Use volta for main build

untitaker:
We will likely have to change the cache key to include package.json.

Follows up on #2666


untitaker:
Let's do the cache key change if master is still red.

PR #2675, opened 2020-11-11
User keys are actually listed _below_ the lines

Bathlamos:
Unless I'm mistaking, it seems like the user properties are listed below, not above the line mentioned


dashed (approved):
Thanks!

PR #2688, opened 2020-11-15
Change @sentry/serverless to @sentry/node

SpencerRiddering:
Honestly not sure what the correct documentation should be here. Changing to @sentry/node clears all the compile time errors for me. I couldn't find the  Sentry.GCPFunction.wrapCloudEventFunction(..) so I handle it manually in my project. 

I'm using Firebase Functions.


ajjindal:
@SpencerRiddering let us know if the latest version did not work.

PR #2695, opened 2020-11-18
fix: PHP request body

HazAT:
Once Python SDK implemented this correctly we can remove the dupe.
https://github.com/getsentry/sentry-python/issues/926

### DIFF
--- a/src/platforms/common/configuration/options.mdx
+++ b/src/platforms/common/configuration/options.mdx
@@ -115,7 +115,7 @@ This option can be overridden using `in-app-include`.
 
 </ConfigKey>
 
-<ConfigKey name="request-bodies" supported={["python", "php"]}>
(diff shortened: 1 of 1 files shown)


untitaker (approved):

src/platforms/common/configuration/options.mdx
```diff
@@ -126,6 +126,18 @@ This parameter controls if integrations should capture HTTP request bodies. It c
 
 </ConfigKey>
 
+<ConfigKey name="max-request-body-size" supported={["php"]}>
+
+This parameter controls if integrations should capture HTTP request bodies. It can be set to one of the following values:
+
+- `never`: request bodies are never sent
```
guilliamxavier, line 133: Thanks, but as already pointed out, PHP uses `none` instead of `never`. Xref https://github.com/getsentry/sentry-php/issues/1029

PR #2704, opened 2020-11-20
Update dsym.mdx

dirtyhenry:
Using `api_host` will output the following error: 

```
Could not find option 'api_host' in the list...
```


philipphofmann (approved):
Thanks a lot, @dirtyhenry.

PR #2706, opened 2020-11-20
{WIP} Release Health SDK Definitions

PeloWriter:
This PR provides a page that defines terms that are compatible for Release Health, per platform to minimize confusion in sentry.io. 

Because each platform has its own concept of what the term means, the content became too lengthy to include on the primary Health page, so this PR adds a brief explainer of the terms with a link to the Terms page from the Health page.


PeloWriter:
We will handle these definitions as tooltips in the UI.

PR #2707, opened 2020-11-20
fix(integrations): Update Splunk self-service HEC url

NisanthanNanthakumar:
## Objective
Our Splunk integration docs are out of date. When we try to setup an HEC with the self-service Splunk plan and our current docs, we fail to connect with a valid url. 

Screen Shot 2020-11-16 at 3 22 01 PM

Our current docs provide the same instructions as the Splunk docs:
https://docs.splunk.com/Documentation/Splunk/8.1.0/Data/UsetheHTTPEventCollector#Send_data_to_HTTP_Event_Collector_on_Splunk_Cloud_instances

However, when we remove the `input` prefix from the url, we are able to connect to the Splunk instance and forward events. This has only been confirmed with the self-service Splunk Cloud plan.

Apparently for other plans, Splunk need to use `https-input` prefix, but we need to confirm whether this is still true.

### DIFF
--- a/src/docs/product/integrations/splunk/index.mdx
+++ b/src/docs/product/integrations/splunk/index.mdx
@@ -94,10 +94,10 @@ After navigating to **Data Forwarding**, enable the Splunk integration:
 
 ![](splunk-data-forwarding-setting.png)
 
-Your instance URL is going to vary based on the type of Splunk service you’re using. If you’re using self-service Splunk Cloud, the instance URL will use the `input` prefix:
+Your instance URL is going to vary based on the type of Splunk service you’re using. If you’re using self-service Splunk Cloud, the instance URL will be <host> the url of the Splunk instance that runs HEC:
 
 ```
-https://input-<host>:8088
+https://<host>:8088
 ```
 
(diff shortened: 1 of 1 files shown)


src/docs/product/integrations/splunk/index.mdx
```diff
@@ -94,10 +94,10 @@ After navigating to **Data Forwarding**, enable the Splunk integration:
 
 ![](splunk-data-forwarding-setting.png)
 
-Your instance URL is going to vary based on the type of Splunk service you’re using. If you’re using self-service Splunk Cloud, the instance URL will use the `input` prefix:
+Your instance URL is going to vary based on the type of Splunk service you’re using. If you’re using self-service Splunk Cloud, the instance URL will be <host> the url of the Splunk instance that runs HEC:
```
ceorourke, line 97: this reads a little strangely to me, maybe something like:

"Your instance URL is going to vary based on the type of Splunk service you’re using. If you’re using self-service Splunk Cloud, the instance URL will be the URL of the Splunk instance running the HTTP Event Collector (HEC):"

ceorourke (approved):

PR #2709, opened 2020-11-23
Fix typo in rails platform guide

jaredpalmer:
Found a typo


PeloWriter (changes requested):
This is a correct fix of a typo (thank you).

One note - if updating the file, the index.mdx file can be removed altogether, as documented in these contributor guidelines - https://docs.sentry.io/contributing/approach/write-getting-started

PR #2715, opened 2020-11-25
Add Java & Spring Boot performance docs.

maciejwalkowiak:
First batch of Java performance docs. Spring integration will come next.

Python Performance reference was used as the starting point.

Note: these docs use `Sentry#getSpan` method, which is currently under review: https://github.com/getsentry/sentry-java/pull/1070

### DIFF
--- a/src/includes/performance/default-sampling-context-platform/java.mdx
+++ b/src/includes/performance/default-sampling-context-platform/java.mdx
@@ -1,16 +1,1 @@
-For Java-based SDKs, it includes at least the following:
-
-```javascript
-// contents of `samplingContext`
-{
-  transactionContext: {
-    name: string; // human-readable identifier, like "GET /users"
-    parentSampled: boolean; // if this transaction has a parent, its sampling decision
-  }
-  customSamplingContext: {
-    data: {
-        ... // custom context as passed to `startTransaction`
-    }
-  }
-}
-```
+For Java-based SDKs, it includes a [Transaction Context](https://getsentry.github.io/sentry-java/io/sentry/TransactionContext.html) and a [Custom Sampling Context](https://getsentry.github.io/sentry-java/io/sentry/CustomSamplingContext.html).
--- a/src/includes/performance/default-sampling-context-platform/java.mdx
+++ b/src/includes/performance/default-sampling-context-platform/java.mdx
@@ -1,0 +1,16 @@
+For Java-based SDKs, it includes at least the following:
+
+```javascript
+// contents of `samplingContext`
+{
+  transactionContext: {
+    name: string; // human-readable identifier, like "GET /users"
+    parentSampled: boolean; // if this transaction has a parent, its sampling decision
+  }
+  customSamplingContext: {
+    data: {
+        ... // custom context as passed to `startTransaction`
+    }
+  }
+}
+```
--- a/src/includes/performance/enable-tracing/java.mdx
+++ b/src/includes/performance/enable-tracing/java.mdx
@@ -1,0 +1,15 @@
+```java
+import io.sentry.Sentry;
+
+Sentry.init(
+  options -> {
+    options.setDsn("___PUBLIC_DSN___");
+    // To set a uniform sample rate
+    options.setTracesSampleRate(1.0);
+    // Determine traces sample rate based on the sampling context
+    options.setTracesSampler(
+        context -> {
+          // return a number between 0 and 1
+        });
+  });
+```
--- a/src/includes/performance/enable-tracing/java.mdx
+++ b/src/includes/performance/enable-tracing/java.mdx
@@ -6,7 +6,7 @@ Sentry.init(
     options.setDsn("___PUBLIC_DSN___");
     // To set a uniform sample rate
     options.setTracesSampleRate(1.0);
-    // Determine traces sample rate based on the sampling context
+    // OR if you prefer, determine traces sample rate based on the sampling context
     options.setTracesSampler(
         context -> {
           // return a number between 0 and 1
--- a/src/includes/performance/enable-tracing/java.spring-boot.mdx
+++ b/src/includes/performance/enable-tracing/java.spring-boot.mdx
@@ -1,0 +1,30 @@
+To get started automatically sending traces you must enable tracing in `application.properties`:
+
+```properties
+sentry.enable-tracing=true
+```
+
+Next, you must choose one of two options to configure sampling:
+
+- Set a uniform sample rate for all transactions, by setting the <PlatformIdentifier name="traces-sample-rate" /> option in your SDK config to a number between `0` and `1`. (For example, to send 20% of transactions, set <PlatformIdentifier name="traces-sample-rate" /> to `0.2`.)
+- Control the sample rate dynamically, based on the transaction itself and the context in which it's captured, by providing a function to the <PlatformIdentifier name="traces-sampler" /> config option.
+
+Or alternatively:
+
+Configure sample rate in `application.properties` or `application.yml` file:
+
+```properties
+sentry.traces-sample-rate=0.3
+```
+
+Or through providing a bean of type `SentryOptions#TracesSamplerCallback`:
+
+```java
+@Component
+class CustomTracesSamplerCallback {
+  @Override
+  public @NotNull Double sample(@NotNull SamplingContext samplingContext) {
+    // return a number between 0 and 1
+  }
+}
+```
--- a/src/includes/performance/traces-sampler-as-sampler/java.mdx
+++ b/src/includes/performance/traces-sampler-as-sampler/java.mdx
@@ -6,7 +6,7 @@ Sentry.init(
     options.setDsn("___PUBLIC_DSN___");
     // To set a uniform sample rate
     options.setTracesSampleRate(1.0);
-    // Determine traces sample rate based on the sampling context
+    // OR: Determine traces sample rate based on the sampling context
     options.setTracesSampler(
       context -> {
         if (...) {
--- a/src/includes/performance/traces-sampler-as-sampler/java.mdx
+++ b/src/includes/performance/traces-sampler-as-sampler/java.mdx
@@ -9,14 +9,14 @@ Sentry.init(
     // OR: Determine traces sample rate based on the sampling context
     options.setTracesSampler(
       context -> {
-        if (...) {
+        if ("/payment".equals(context.get("url"))) {
           // These are important - take a big sample
           return 0.5;
-        } else if (...) {
-          // These are less important or happen much more frequently - only take 1%
+        } else if ("/search".equals(context.get("url"))) {
+          // Search is less important and happen much more frequently - only take 1%
           return 0.01;
-        } else if (...) {
-          // These aren't something worth tracking - drop all transactions like this
+        } else if ("/health".equals(context.get("url"))) {
+          // The health check endpoint is just noise - drop all transactions
           return 0;
         } else {
           // Default sample rate
--- a/src/includes/performance/traces-sampler-as-sampler/java.mdx
+++ b/src/includes/performance/traces-sampler-as-sampler/java.mdx
@@ -1,0 +1,27 @@
+```java
+import io.sentry.Sentry;
+
+Sentry.init(
+  options -> {
+    options.setDsn("___PUBLIC_DSN___");
+    // To set a uniform sample rate
+    options.setTracesSampleRate(1.0);
+    // Determine traces sample rate based on the sampling context
+    options.setTracesSampler(
+      context -> {
+        if (...) {
+          // These are important - take a big sample
+          return 0.5;
+        } else if (...) {
+          // These are less important or happen much more frequently - only take 1%
+          return 0.01;
+        } else if (...) {
+          // These aren't something worth tracking - drop all transactions like this
+          return 0;
+        } else {
+          // Default sample rate
+          return 0.1;
+        }
+      });
+  });
+```
--- a/src/platforms/common/performance/sampling.mdx
+++ b/src/platforms/common/performance/sampling.mdx
@@ -7,6 +7,7 @@ supported:
   - javascript
   - node
   - python
+  - java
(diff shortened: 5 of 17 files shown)


maciejwalkowiak:
cc @bruno-garcia

src/includes/performance/enable-tracing/java.mdx
```diff
@@ -0,0 +1,15 @@
+```java
+import io.sentry.Sentry;
+
+Sentry.init(
+  options -> {
+    options.setDsn("___PUBLIC_DSN___");
+    // To set a uniform sample rate
+    options.setTracesSampleRate(1.0);
+    // Determine traces sample rate based on the sampling context
```
bruno-garcia, line 9:
```suggestion
    // OR if you prefer, determine traces sample rate based on the sampling context
```

src/includes/performance/default-sampling-context-platform/java.mdx
```diff
@@ -0,0 +1,16 @@
+For Java-based SDKs, it includes at least the following:
+
+```javascript
```
bruno-garcia, line 3: JS?

src/includes/performance/traces-sampler-as-sampler/java.mdx
```diff
@@ -0,0 +1,27 @@
+```java
+import io.sentry.Sentry;
+
+Sentry.init(
+  options -> {
+    options.setDsn("___PUBLIC_DSN___");
+    // To set a uniform sample rate
+    options.setTracesSampleRate(1.0);
+    // Determine traces sample rate based on the sampling context
```
bruno-garcia, line 9:
```suggestion
    // OR: Determine traces sample rate based on the sampling context
```

src/includes/performance/traces-sampler-as-sampler/java.mdx
```diff
@@ -0,0 +1,27 @@
...
+      context -> {
+        if (...) {
+          // These are important - take a big sample
+          return 0.5;
+        } else if (...) {
+          // These are less important or happen much more frequently - only take 1%
+          return 0.01;
+        } else if (...) {
+          // These aren't something worth tracking - drop all transactions like this
+          return 0;
```
bruno-garcia, line 20: I think it would be nice to show some actual example here to help clarify the use case:

// something like this (That compiles)
```suggestion
        } else if (context.get("url") != null && context.get("url") == "/health")) {
          // The health check endpoint is just noise - drop all transactions
          return 0;
```

src/platforms/common/performance/sampling.mdx
```diff
@@ -49,7 +50,7 @@ Whatever a transaction's sampling decision, that decision will be passed to its
 
 If the transaction currently being created is one of those subsequent transactions (in other words, if it has a parent transaction), the upstream (parent) sampling decision will always be included in the sampling context data, so that your <PlatformIdentifier name="traces-sampler" /> can choose whether and when to inherit that decision. (In most cases, inheritance is the right choice, so that you don't end up with partial traces.)
 
-For convenience, the <PlatformIdentifier name="traces-sampler" /> function can return a boolean, so that a parent's decision can be returned directly if that's the desired behavior.
+In some SDKs, for convenience, the <PlatformIdentifier name="traces-sampler" /> function can return a boolean, so that a parent's decision can be returned directly if that's the desired behavior.
```
bruno-garcia, line 53: Is this something we diverged? Should we align (one way or the other?)

src/platforms/java/common/performance/index.mdx
```diff
@@ -0,0 +1,103 @@
+---
+title: Performance Monitoring
+description: "Learn more about how to configure our Performance integrations to get the best experience out of it."
+---
+
+<Alert level="warning" title="Note">
+Performance Monitoring is available for the Sentry Java SDK version ≥ 4.0.
+</Alert>
+
+Sentry allows you to monitor the performance of your application, showing you how latency in one service may affect another service, and letting you pinpoint exactly which parts of a given operation may be responsible. To do this, it captures distributed traces consisting of transactions and spans, which measure individual services and individual operations within those services, respectively. You can learn more about this model in our [Distributed Tracing](/product/performance/distributed-tracing/) docs. Performance Monitoring is available for the Sentry Java SDK version ≥ 4.0.
```
bruno-garcia, line 10: We added the note above

```suggestion
Sentry allows you to monitor the performance of your application, showing you how latency in one service may affect another service, and letting you pinpoint exactly which parts of a given operation may be responsible. To do this, it captures distributed traces consisting of transactions and spans, which measure individual services and individual operations within those services, respectively. You can learn more about this model in our [Distributed Tracing](/product/performance/distributed-tracing/) docs.
```

src/platforms/java/common/performance/index.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+## Enabling Tracing
+
+To get started automatically sending traces you can:
+
+- Set a uniform sample rate for all transactions, by setting the <PlatformIdentifier name="traces-sample-rate" /> option in your SDK config to a number between `0` and `1`. (For example, to send 20% of transactions, set <PlatformIdentifier name="traces-sample-rate" /> to `0.2`.)
+- Control the sample rate dynamically, based on the transaction itself and the context in which it's captured, by providing a function to the <PlatformIdentifier name="traces-sampler" /> config option.
```
bruno-garcia, line 17: I got confused if we need both. Below we point out only 1 is enough.
I feel we need to make it more clear:


```suggestion
- Control the sample rate dynamically, based on the transaction itself and the context in which it's captured, by providing a function to the <PlatformIdentifier name="traces-sampler" /> config option.

Or alternatively:
```

src/platforms/java/common/performance/index.mdx
```diff
@@ -0,0 +1,103 @@
...
+```java
+import io.sentry.Sentry;
+import io.sentry.SentryTransaction;
+import io.sentry.SpanStatus;
+
+SentryTransaction transaction = Sentry.startTransaction("transaction name");
+transaction.setOperation("task");
+try {
+    processItem();
+} catch (Exception e) {
```
bruno-garcia, line 42: Maybe a sort of "unique" name and point out it's important the transaction name is relevant/helps you understand what this is about. Shows in the UI in a prevalent way.

Also `processItem` is used below on the span example.

```suggestion
The following example creates a transaction for a scope that contains an expensive operation (for example, `processOrderBatch`), and sends the result to Sentry:

```java
import io.sentry.Sentry;
import io.sentry.SentryTransaction;
import io.sentry.SpanStatus;

// A good name for the transaction is key, to help identify what this is about
SentryTransaction transaction = Sentry.startTransaction("processOrderBatch()");
transaction.setOperation("task");
try {
    procesOrderBatch();
} catch (Exception e) {
```

src/platforms/java/common/performance/index.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+You can choose the value of `operation` and `description`.
+
+```java
+import io.sentry.ISpan;
+import io.sentry.Span;
+import io.sentry.SpanStatus;
+
+void processItem(Item item):
+    ISpan span = Sentry.getSpan();
```
bruno-garcia, line 64: This means the user would need to have bound the transaction to the scope, according to what I got from @HazAT so far. And we haven't documented that.

Do we need to show that on the previous example? Or if we're binding it automatically, I wonder how this works on Android: Not possible to have more than 1 transaction at the same time in the app?

src/platforms/java/common/performance/index.mdx
```diff
@@ -0,0 +1,103 @@
...
+    Span innerSpan = span.startChild();
+    innerSpan.setOperation("task");
+    innerSpan.setDescription(item.getName());
```
bruno-garcia, line 67: Should we have an overload for `startChild`? This is consistently taking 3 lines of code for all operations.

```suggestion
    Span innerSpan = span.startChild(item.getName(), "task");
```

src/platforms/java/common/performance/index.mdx
```diff
@@ -0,0 +1,103 @@
...
+### Retrieving a Transaction
+
+In cases where you want to attach Spans to an already ongoing Transaction you can use `Sentry#getSpan`. This method will return a `SentryTransaction` in case there is a running Transaction or a `Span` in case there is already a running Span, otherwise it returns `null`.
+
+```java
+import io.sentry.ISpan;
+import io.sentry.Span;
+import io.sentry.SpanStatus;
+
+ISpan span = Sentry.getSpan();
```
bruno-garcia, line 88: This won't really work with asynchronous code (this relies on TLS) so should we add a section to the docs about that?

What happens if I want to capture a span for an outgoing request that's happening on a background thread?

src/platforms/common/performance/sampling.mdx
maciejwalkowiak: This I believe was written with Python and JS in mind, where a function can return different data types. In Java to achieve the same we would have to return a wrapper object. I don't think it's worth it..?

src/includes/performance/default-sampling-context-platform/java.mdx
maciejwalkowiak: Just for the sake of formatting. But I agree that it could be confusing. I replace it with just links to javadocs.

src/platforms/java/common/performance/index.mdx
maciejwalkowiak: Addressed in: https://github.com/getsentry/sentry-java/pull/1073

src/platforms/java/common/performance/index.mdx
maciejwalkowiak: Yes there is a single active transaction per `Hub`. I would imagine background thread should use a separate hub instance

src/platforms/java/common/performance/index.mdx
maciejwalkowiak: Transactions are bound to the scope https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Scope.java#L24

src/includes/performance/traces-sampler-as-sampler/java.mdx
maciejwalkowiak: Fixed

src/platforms/java/common/performance/index.mdx
maciejwalkowiak: @marandaneto maybe you can jump in and tell if/how this will work on Android

marandaneto: tbh, I think it'd be a nice addition to be able to have multiple transactions at once for Mobile as well, but I don't see this as a huge blocker, at the end, it's a single user doing all the things (for most of the Apps), we can just document this limitation and tell people to go around with Spans if they need to measure different things (this is possible, right)?

In most of the cases, there will be a single transaction per screen anyway and if they want to create their own Hub for having multiple transactions, so they do.

The only concern I have right now is, a Hub is not ready for usage the way how it is, when creating a new Hub out of a fresh SentryOptions, a few things are not set like the Serializer, it's NoOp by default, it's only set on Sentry.init static class, what's about fixing this limitation, so at least we give a useful alternative for people?

src/platforms/java/common/performance/index.mdx
marandaneto: so if I want to have a transaction per App. screen, also being able to create spans and its children for background op (like outgoing requests), etc... having a Single Hub which is the case on Android, I don't see this as a problem unless I am missing something, is this described scenario possible to achieve with a single Hub?

maciejwalkowiak:
PR updated with Spring Boot docs.

src/platforms/java/common/performance/index.mdx
maciejwalkowiak: The limitation is that there can be single active transaction per hub

marandaneto: the question still remains:

> is the described scenario above achievable with a single Hub?

the question is probably yes, but I'd like a confirmation.

maciejwalkowiak: It depends. If the transaction in the parent thread is finished before the child thread's span is finished, child thread's span won't be sent together with transaction.

src/platforms/java/common/performance/index.mdx
maciejwalkowiak: Regarding Hub initialization, since I don't know the background behind it I can't tell what are the side-effects, but perhaps @bruno-garcia can shine some light.

src/platforms/java/common/performance/index.mdx
marandaneto: I see, that makes sense, but for now, the support for performance on Android is "manual", so documenting this is just fine:

"finish your transaction only after your spans have finished".

src/platforms/java/common/performance/index.mdx
marandaneto: well it's pretty simple, we've discussed this offline recently actually.

do Hub(SentryOptions(dsn)).captureEvent(SentryEvent()), nothing is sent due to missing default options on SentryOptions, it should work ootb, eg a Serializer is not set (NoOp) by default.

src/platforms/java/common/performance/index.mdx
maciejwalkowiak: Thanks, updated!

src/includes/performance/enable-tracing/java.spring-boot.mdx
```diff
@@ -0,0 +1,30 @@
...
+Configure sample rate in `application.properties` or `application.yml` file:
+
+```properties
+sentry.traces-sample-rate=0.3
+```
+
+Or through providing a bean of type `SentryOptions#TracesSamplerCallback`:
+
+```java
+@Component
```
bruno-garcia, line 23: We could add Kotlin here too (on a new PR)

bruno-garcia (approved):
Great stuff lets merge

PR #2716, opened 2020-11-25
Add ColdFusion to Community Supported list

GiancarloGomez:


bruno-garcia:
Thank you for your PR but somehow we've already got it in:

https://github.com/getsentry/sentry-docs/pull/2720/files

PR #2723, opened 2020-11-28
fix(cli): Update all mention of --rewrite

jan-auer:
`upload-sourcemaps --rewrite` has been removed in https://github.com/getsentry/sentry-cli/pull/853 and will now cause an error when provided. This updates documentation to remove all mention of `--rewrite`.


jan-auer:
This only fixes a few places where obviously incorrect information was given. The document is pretty outdated, however, as multiple options (e.g. `--bundle`) are missing. @kamilogorek could you update this and also improve the language of the descriptions, as well as align them with what we show in the CLI?

bruno-garcia (approved):
LGTM

PR #2724, opened 2020-11-28
meta: Improve code owners for SDKs and CLI

marandaneto:


.github/CODEOWNERS
```diff
@@ -5,14 +5,18 @@
...
+/src/platforms/java/ @bruno-garcia @marandaneto
 
 /src/platforms/native/ @getsentry/owners-native
 
 /src/platforms/rust/ @Swatinem
 
 /src/docs/product/discover-queries/ @getsentry/visibility
 /src/docs/product/performance/ @getsentry/visibility
+
+/src/docs/product/cli/ @getsentry/team-webplatform
```
jan-auer, line 21: This is probably a bit much, how about we restrict this to the releases commands?

If you don't mind, I would add the native team in the same PR?

marandaneto: @jan-auer sure, go for it.
pay attention that the sorting here matters, the last match is the one that will be assigned as reviewer.

jan-auer: Ok so the problem is that I cannot select @getsentry/owners-native as a reviewer team in sentry-docs. Probably a permission issue. 

@PeloWriter to you have admin on this repo to look into this?

bruno-garcia (approved):
From my end: 👍

.github/CODEOWNERS
PeloWriter: @jan-auer - owners-native needs to be granted permissions to the docs repo, which should be handled by an IT request (for compliance/security reasons).

jan-auer: We're probably going to address this by moving the @getsentry/codeowners team under @getsentry/engineering, which should take care of this.

PeloWriter: 👍

jan-auer (approved):
Requested review from @kamilogorek since he is now added as owner for CLI.

kamilogorek (approved):

kamilogorek:
image

PR #2729, opened 2020-11-30
Update legacy plan callout to use better channels for customers

silent1mezzo:


jeremybourque (approved):
I don't know for certain that sales@sentry.io is valid but assuming that's correct this looks good to me.

markstory (approved):

PR #2741, opened 2020-12-02
ref: Promote nicer API in performance for python

untitaker:


untitaker:
@rhcarvalho apparently we deprecated those methods here: https://github.com/getsentry/develop/pull/138

please decide what to do with this

src/platforms/python/common/performance/index.mdx
```diff
@@ -48,12 +48,12 @@ Alternatively, instead of adding to the top-level transaction, you can make a ch
...
 
 ```python
-from sentry_sdk import Hub
+import sentry_sdk
 
 def process_item(item):
     transaction = Hub.current.scope.transaction
     # omitted code...
-    with transaction.start_child(op="http", description="GET /") as span:
+    with sentry_sdk.start_span(op="http", description="GET /") as span:
```
rhcarvalho, line 56: @untitaker could you describe what is the intended behavior?
What does starting a span without a reference to a transaction do?

untitaker: I can add this to the document but basically `start_span` would return a noop context manager. This is not what is actually happening under the hood: In practice a span is constructed and destroyed on the scope, but never put into any sort of array or other datastructure to be put into a transaction event (I'd leave that up to implementation)

rhcarvalho: 👍 

Okay, if I got you right, it would have the side-effect of changing the scope of the current hub.
This is only useful if at times there would also be a transaction in the same scope, in which case the span would not be discarded, but instead would become part of the ongoing transaction -- is that accurate?

untitaker: It's useful to have this noop in case there is no wrapping transaction. The otehr two alternatives I see is:

* throw error that there is no transaction -- functions would have to guard against this at all times
* start own transaction -- if there are 100 db spans in an endpoint and the endpoint for some reason did not start a transaction, this can lead to performance and billing problems

src/platforms/python/common/performance/index.mdx
```diff
@@ -48,12 +48,12 @@ Alternatively, instead of adding to the top-level transaction, you can make a ch
 You can choose the value of `op` and `description`.
 
 ```python
-from sentry_sdk import Hub
+import sentry_sdk
 
 def process_item(item):
     transaction = Hub.current.scope.transaction
```
rhcarvalho, line 54:
```suggestion
```

rhcarvalho (approved):
Once `start_transaction` was introduced, there were then two ways to start spans:

- `[sentry_sdk | hub].start_span`
- `Transaction.start_child`

IIRC, the only reason to "pick a favorite" was trying to bring back the idea of "there should be only one obvious way to do things".

The second form was preferred for clarity, though it does come with an onus of having to grab a reference to the current transaction when you didn't start it yourself; and checking for `None` before calling a method, so definitely less ergonomic.

I'm okay with the current change, as it documents both forms where they make sense.

untitaker:
@rhcarvalho what do we do with the development docs? Uncomment the shortcut description?

untitaker:
@rhcarvalho bump

bruno-garcia:
@untitaker Rodolfo is back on January 11th

untitaker:
Going to merge this as its affecting new users. I don't know what we want to do with development docs but something has to happen here.

PR #2744, opened 2020-12-04
feat: Add node.js db docs

HazAT:
image

### DIFF
--- a/src/platforms/node/common/performance/database.mdx
+++ b/src/platforms/node/common/performance/database.mdx
@@ -1,0 +1,57 @@
+---
+title: Database integrations
+---
+
+Node.js has a few integrations that support tracking database queries as Spans. For now you only need to use the latest version of our SDK and add the specific integration to `Sentry.init`:
+
+### MongoDB
+
+```javascript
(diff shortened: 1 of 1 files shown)


src/platforms/node/common/performance/database.mdx
```diff
@@ -0,0 +1,57 @@
+---
+title: Database integrations
+---
+
+Node.js has a few integrations that support tracking database queries as Spans. For now you only need to use the latest version of our SDK and add the specific integration to `Sentry.init`:
```
PeloWriter, line 5: I'm a little confused about what we mean by "for now" - do we mean "currently"? Or, are we saying that now users just need to upgrade to the latest version and add the MongoDB integration?

PeloWriter (approved):
One question to clarify, otherwise, lgtm.

src/platforms/node/common/performance/database.mdx
HazAT: ha :)
We can remove "for now", actually we want that this works out of the box without the users needing to configure anything. The internal architecture of how this works in JS doesn't make this easy yet so that's why.

PR #2746, opened 2020-12-04
fix(express): Group comments with the options to which they pertain

lobsterkatie:
Some newlines had gone on walkabout and landed in the wrong place.


PeloWriter (approved):
Thank you.

PR #2748, opened 2020-12-05
fix: include path of gatsby integration

yamitzky:
Gatsby's dist path is "public". When `include` is `'.'`, source maps of http://example.com/blah-blah-blah.js are uploaded as "~/public/blah-blah-blah.js.map". This is misconfigured.

image

When `"include"` is `"public"`, they will be "~/blah-blah-blah.js.map" and source mapping would work.

image

### DIFF
--- a/src/platforms/javascript/common/sourcemaps/index.mdx
+++ b/src/platforms/javascript/common/sourcemaps/index.mdx
@@ -62,7 +62,7 @@ module.exports = {
         project: "___PROJECT_SLUG___",
 
         // webpack specific configuration
-        include: ".",
+        include: "public",
         ignore: ["node_modules", "webpack.config.js"],
       }),
     ],


lobsterkatie (changes requested):
Hi, @yamitzky.

This content is shared across multiple JS frameworks, so making a gatsby-specific change would mess it up for our react/angular/etc packages.

Instead of this change, can you please add a section like the one we have for vue (directly underneath it), saying that the `include` value should be `"./public"`?

```
<PlatformSection supported={["javascript.vue"]}>

In Vue 2.x, you should use `vue.config.js` instead of `webpack.config.js`, and use `include: "./dist"` instead of `include: "."`

</PlatformSection>
```

P.S. As an aside, the issue isn't whether or not `public` is included in the filepath of both the minified scripts and their associated source maps - you can see that everything in the `public` directory is actually having its source map detected successfully.  It's what's getting examined in the first place. Without `public` in the path, webpack ends up catching your source files, which of course have no maps. With `public` in the path, it only looks at built code.

PR #2756, opened 2020-12-08
Logback Getting Started refactor and minor IA changes

PeloWriter:
Refactor of Getting Started page. This refactor enables use of the common content. In addition, the content:

- uses Java fallbacks for the Verification step
- creates Advanced Usage as a subpage of Usage to cover the MDC content
- create Configuration Example to capture the previously-named section "Advanced Usage"


marandaneto (approved):
LGTM

marandaneto:
cc @maciejwalkowiak

bruno-garcia (approved):
LGTM.

One thing though, the "verify" part should be a logback call to `logError`, instead of using `Sentry.something`

https://sentry-docs-git-update-logback-getting-started-refactor.sentry.dev/platforms/java/guides/logback/#verify

@maciejwalkowiak could you please send a code snippet for this?

Also noticed that breadcrumbs and scope use our Sentry API and it would be nice if there was a logback API that modifies the scope. This exists in .NET so I wonder here, @maciejwalkowiak knows?

Configuration shows how to init with `Sentry.init`. Here it should be as it's described in the getting started, and only if users want to pass a callback or do some other configuration that can't be done with the appender.xml, we'll point them to `Sentry.init`.

So good follow up PR.
https://sentry-docs-git-update-logback-getting-started-refactor.sentry.dev/platforms/java/guides/logback/configuration/

maciejwalkowiak:
@PeloWriter please take a look at a follow-up PR: https://github.com/getsentry/sentry-docs/pull/2779.

As Bruno mentioned, ideally we should replace configuration page for Logback (https://sentry-docs-git-update-logback-getting-started-refactor.sentry.dev/platforms/java/guides/logback/configuration/) with information how to configure Logback through `logback.xml` file and for more advanced cases how to initialize SDK in static block. I am not sure where to put it in the docs sources so that configuration sections are different for parent Java and children like Logback or Spring.

Same applies to "Scope & Breadcrumbs" - some of the information on how to do it with Logback is already in getting started section (https://sentry-docs-git-update-logback-getting-started-refactor.sentry.dev/platforms/java/guides/logback/#minimum-log-level).

PR #2763, opened 2020-12-10
fix(native): Update set-context documentation

jan-auer:
Documentation for `set_context` still mentioned the deprecated `set_extra` from
when `set_context` was not supported in the SDK.


jan-auer:
cc @PeloWriter

src/platforms/common/enriching-events/context.mdx
```diff
@@ -10,35 +10,33 @@ notSupported:
...
 
 </Note>
 
-To configure additional context:
+## Structured Context
 
-<PlatformContent includePath="set-context" />
-
-Note that the context's outer value has to be a dictionary/map/object, while the inner values can be arbitrary.
+The best practice to attach custom data is via structured contexts. A context must always be a dictionary or map, and its values can be arbitrary. Use <PlatformIdentifier name="set-context" /> and give the context a unique name:
```
flub, line 25: This reads weird, because you refer to `context` twice, but once it's the map of all context values and once it's a single value

flub (approved):

PR #2775, opened 2020-12-11
fix: Correct wording of basic options intro

jan-auer:
Adds a missing noun to the description of passing basic options to the init function or method. Whether it is a function or method depends on the platform.

Note that I had to remove the code backticks from descriptions as they are not being rendered by our page grid component:

image


PeloWriter (approved):
Thank you.

PR #2784, opened 2020-12-16
Update breadcrumbs docs to include types link

keeprock:
It was inconvenient to read about types and not being able to actually look what kind of types are available. This fix it.


PeloWriter:
This is a helpful fix. Thank you.

Issue #2785, opened 2020-12-16
Add docs for attachments preview for Java and Apple


philipphofmann:
Add docs about https://github.com/getsentry/sentry-cocoa/issues/871 and https://github.com/getsentry/sentry-java/pull/1082 mentioning this is still in alpha.

PR #2786, opened 2020-12-16
feat(apple/java): Add docs on attachments preview

philipphofmann:
Fixes GH-2785


src/platforms/android/common/enriching-events/attachments.mdx
```diff
@@ -0,0 +1,54 @@
...
+__Attention: This is a preview API, available since 4.0.0-alpha.2. This API is not in stable state yet. It may be renamed, changed or even removed in a future version.__
+
+Attachments provide the ability to to store additional files to an event. Be aware that attachments don't work yet with crashes.
+
+
+## Uploading Attachments
+
+You can either pass bytes or use a pathname to a file to create an attachment.
+
+```java
```
bruno-garcia, line 17: We're trying to double down on Kotlin examples going forward.
Would you add it after the Java ones please?

Once we have all examples for both, we could swap them so Kotlin comes first for Android, what do you think?

src/platforms/apple/common/enriching-events/attachments.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+The attachments live on the <PlatformLink to="/enriching-events/scopes/">Scope</PlatformLink>. You can either add an attachment on the global scope to be sent with every event or add it on the <PlatformLink to="/enriching-events/scopes/#local-scopes">local Scope</PlatformLink> to just send the attachment with one specific event.
+
+```swift {tabTitle:Swift}
+import Sentry
+
+let fileAttachment = Attachment(path: "your/path/file.log")
+
+// Global Scope
+SentrySDK.configureScope { scope in
```
bruno-garcia, line 42: Sounds like this code could move into the `includes` by platform (so apple.mdx here, and android.mdx and java.mdx for the other page) and share the main content which is now duplicate.

@PeloWriter will know how this works better

src/platforms/android/common/enriching-events/attachments.mdx
philipphofmann: I agree fully, good idea 😀. I added Kotlin samples.

src/platforms/apple/common/enriching-events/attachments.mdx
philipphofmann: Makes sense.

src/platforms/android/common/enriching-events/attachments.mdx
```diff
@@ -0,0 +1,78 @@
+---
+title: "Attachments Preview"
+sidebar_order: 106
+
+description: "Learn more about attachments, which provide the ability to store additional files to an event."
```
marandaneto, line 5:
```suggestion
description: "Learn more about attachments, which provides the ability to store additional files to an event."
```

src/platforms/android/common/enriching-events/attachments.mdx
```diff
@@ -0,0 +1,78 @@
...
+__Attention: This is a preview API, available since 4.0.0-alpha.2. This API is not in stable state yet. It may be renamed, changed or even removed in a future version.__
+
+Attachments provide the ability to to store additional files to an event. Be aware that attachments don't work yet with crashes.
+
+
+## Uploading Attachments
+
+You can either pass bytes or use a pathname to a file to create an attachment.
+
+```java
```
marandaneto, line 17: can we use the Tabs feature instead of adding a code snippet below?

src/platforms/java/common/enriching-events/attachments.mdx
```diff
@@ -0,0 +1,78 @@
+---
+title: "Attachments Preview"
```
marandaneto, line 2: Android and Java and even Apple could reuse the includes

src/platforms/android/common/enriching-events/attachments.mdx
philipphofmann: What exactly is the Tabs feature?

marandaneto: oh looks like it works by default?!

Screenshot 2020-12-17 at 10 50 34

if you look at https://raw.githubusercontent.com/getsentry/sentry-docs/master/src/platforms/java/usage.mdx

we needed to add this before: `{tabTitle:LanguageX}` so it'd group otherwise not.

philipphofmann: I think `{tabTitle:LanguageX}` only makes sense when the language specification for the code snippet deviates from the actual language name, for example `objc {tabTitle:Objective-C}`.

src/platforms/java/common/enriching-events/attachments.mdx
philipphofmann: I could make use of `<PlatformContent includePath="add-attachment" />`, but then native and Javascript also need to change it, but they have slightly different API, so we can't unify it yet.

src/platforms/android/common/enriching-events/attachments.mdx
marandaneto: they probably fixed this, as I stated before, without the `tabTitle`, they were not grouping at all.

philipphofmann: Ah okay didn't know that.

src/platforms/java/common/enriching-events/attachments.mdx
marandaneto: add-attachment is only the snippets, what's about creating its own mobile snippets and add support to the comments attachments?

https://github.com/getsentry/sentry-docs/blob/master/src/platforms/common/enriching-events/attachments/index.mdx#uploading-attachments

philipphofmann: Sorry @marandaneto I don't fully understand what you mean.

marandaneto: the common content already exists, so I don't see why we can't use it.

https://docs.sentry.io/platforms/javascript/enriching-events/attachments/

philipphofmann: I know added includes. Better now @marandaneto ?

src/includes/attachment-init/android.mdx
```diff
@@ -0,0 +1,14 @@
+```java
```
marandaneto, line 1: H: The Android bits when it's exactly the same, fallback to the Java ones, so we don't actually need to have the android.mdx here, it's exactly the same as Java and will pick up the java.mdx automatically.

src/includes/attachment-upload/android.mdx
```diff
@@ -0,0 +1,37 @@
+```java
```
marandaneto, line 1: H: Also not here, we could just remove this file if its exactly the same as Java

src/includes/attachment-init/android.mdx
philipphofmann: That's new to me. Didn't know that we fixed this now. Thanks for pointing it out.

bruno-garcia (approved):
Looks great!

PR #2787, opened 2020-12-16
Initial Kotlin additions

PeloWriter:
This WIP provides initial pieces of adding Kotlin. Note this fails during a local build.

### DIFF
--- a/src/platforms/kotlin/config.yml
+++ b/src/platforms/kotlin/config.yml
@@ -1,0 +1,5 @@
+title: Kotlin
+caseStyle: camelCase
+supportLevel: production
+sdk: "sentry.kotlin"
+categories:
--- a/src/data/platforms.yml
+++ b/src/data/platforms.yml
@@ -22,6 +22,9 @@
 - slug: java
   type: language
   name: Java
+- slug: kotlin
+  type: language
+  name: Kotlin
 - slug: android
   type: framework
   name: Android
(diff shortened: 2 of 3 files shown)


bruno-garcia:
@marandaneto when you added Dart, was there anything fancy you had to do?
build here isn't going through for some reason

src/platforms/kotlin/config.yml
```diff
@@ -0,0 +1,5 @@
+title: Kotlin
+caseStyle: camelCase
+supportLevel: production
+sdk: "sentry.kotlin"
```
marandaneto, line 4: such SDK does not exist as it falls back to `sentry.java`, should we skip it or set it as `sentry.java`?

src/platforms/kotlin/config.yml
```diff
@@ -0,0 +1,5 @@
...
+categories:
```
marandaneto, line 5: could be `mobile` and `server`, do we have a category for `desktop` as well?

marandaneto:
> @marandaneto when you added Dart, was there anything fancy you had to do?
> build here isn't going through for some reason

no, but there's no kotlin platform folder nor index page, I guess that would be the bare min.

src/platforms/kotlin/config.yml
marandaneto: yeah `desktop` also exists, so these 3 would be nice

PeloWriter:
Closing as we fixed this with https://github.com/getsentry/sentry-docs/pull/3600

PR #2796, opened 2020-12-20
fix: Hide sidebar entries without CSS

untitaker:
This is *not* logically equivalent but I can't find a case in the sidebar where it makes a difference.

This is continuation from #2776 and removes another 300MB from `public/`, bringing down the total to 284MB.


untitaker:
@dcramer any thoughts on this?

PR #2802, opened 2020-12-22
bump dart and flutter version

bruno-garcia:
Dart and Flutter still not using craft or release registry

### DIFF
--- a/src/includes/getting-started-install/dart.mdx
+++ b/src/includes/getting-started-install/dart.mdx
@@ -1,4 +1,4 @@
 ```yml {filename:pubspec.yaml}
 dependencies:
-  sentry: ^4.0.0
+  sentry: ^4.0.1
 ```
--- a/src/includes/getting-started-install/flutter.mdx
+++ b/src/includes/getting-started-install/flutter.mdx
@@ -1,4 +1,4 @@
 ```yml {filename:pubspec.yaml}
 dependencies:
-  sentry_flutter: ^4.0.0
+  sentry_flutter: ^4.0.1
 ```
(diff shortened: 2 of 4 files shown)


marandaneto:
I didn't do it purposefully, as exposed to Gradle, using the `^` operator is recommended on dependency management.
so, even if they copy-paste the current version, it downloads the latest one.
closing this, feel free to reopen if you disagree.

bruno-garcia:
@marandaneto but why wouldn't we already pin the latest?

marandaneto:
> @marandaneto but why wouldn't we already pin the latest?

tbh no special reason other than one more thing to do manually (while we don't do craft) for something that does not bring any value.

bruno-garcia:
@marandaneto it does bring value as in we know the customer is on that version (or later). Vs an issue opens and they say: I'm on `^4.0.0.

ALso, I already did the bump here, it just needed to be merged. Is there a reason not to merge this?

marandaneto:
> @marandaneto it does bring value as in we know the customer is on that version (or later). Vs an issue opens and they say: I'm on `^4.0.0.

version should be looked over the `lockfile` (version specific) and not the dependency line, so this would not be an issue.

> ALso, I already did the bump here, it just needed to be merged. Is there a reason not to merge this?

not really, feel free to reopen & merge, round numbers are nicer :D

bruno-garcia:
> version should be looked over the lockfile (version specific) and not the dependency line, so this would not be an issue.

We should point that out on the issue template then to make sure we don't get back and forth with users reporting bugs

PR #2669, opened 2020-11-09
.NET Sentry SDK v3.0.0 - ASP.NET integration

bruno-garcia:
To be merged once .NET 3.0 comes out

PR #2804, opened 2020-12-22
aspnet wizard

bruno-garcia:
An ASP.NET Classic wizard for the current version (which requires only `Sentry`).

#2669 will introduce docsonce the SDK v3 is released. That'll require installing `Sentry.AspNet` so this wizard install command will need to change then.

We're over a month away though.

### DIFF
--- a/src/wizard/dotnet/aspnet.md
+++ b/src/wizard/dotnet/aspnet.md
@@ -19,7 +19,7 @@ Install-Package Sentry -Version {{ packages.version('sentry.dotnet') }}
     </div>
 </div>
 
-Add Sentry to `Global.asax.cs`:
+You should `init` the Sentry SDK as soon as possible during your application load by adding Sentry to `Global.asax.cs`:
 
 ```csharp
 using System.Web;
--- a/src/wizard/dotnet/aspnet.md
+++ b/src/wizard/dotnet/aspnet.md
@@ -1,0 +1,64 @@
+---
+name: ASP.NET
+doc_link: https://docs.sentry.io/platforms/dotnet/guides/aspnet/
+support_level: production
+type: framework
+---
+
+Install the **NuGet** package:
+
+Package Manager:
+
+```shell
+Install-Package Sentry -Version {{ packages.version('sentry.dotnet') }}
+```
+
+<div class="alert alert-info" role="alert"><h5 class="no_toc">Using .NET Framework prior to 4.6.1?</h5>
+    <div class="alert-body content-flush-bottom">
+        <a href="https://docs.sentry.io/clients/csharp/">Our legacy SDK</a> supports .NET Framework as early as 3.5.
+    </div>
+</div>
+
+Add Sentry to `Global.asax.cs`:
+
+```csharp
+using System.Web;
+using Sentry;
+
+public class MvcApplication : HttpApplication
+{
+    private IDisposable _sentry;
+
+    protected void Application_Start()
+    {
+        // Initialize Sentry to capture AppDomain unhandled exceptions and more.
+        _sentry = SentrySdk.Init("___PUBLIC_DSN___");
+    }
+
+    protected void Application_Error()
+    {
+        var exception = Server.GetLastError();
+        // Capture the server errors.
+        SentrySdk.CaptureException(exception);
+    }
+
+    protected void Application_End()
+    {
+        // Flushes out events before shutting down.
+        _sentry?.Dispose();
+    }
+}
+```
+
+### Documentation
+
+Once you've verified the package is initialized properly and sent a test event, consider visiting our [complete ASP.NET docs](https://docs.sentry.io/platforms/dotnet/guides/aspnet/).
+
+### Samples
+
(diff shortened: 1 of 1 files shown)


src/wizard/dotnet/aspnet.md
```diff
@@ -0,0 +1,64 @@
+---
+name: ASP.NET
+doc_link: https://docs.sentry.io/platforms/dotnet/guides/aspnet/
+support_level: production
+type: framework
+---
+
+Install the **NuGet** package:
+
+Package Manager:
```
PeloWriter, line 10: something seems odd here. Do we mean to say "Insatll the **NuGet** package manager:"?

src/wizard/dotnet/aspnet.md
```diff
@@ -0,0 +1,64 @@
...
+```shell
+Install-Package Sentry -Version {{ packages.version('sentry.dotnet') }}
+```
+
+<div class="alert alert-info" role="alert"><h5 class="no_toc">Using .NET Framework prior to 4.6.1?</h5>
+    <div class="alert-body content-flush-bottom">
+        <a href="https://docs.sentry.io/clients/csharp/">Our legacy SDK</a> supports .NET Framework as early as 3.5.
+    </div>
+</div>
+
```
PeloWriter, line 21: I'm uncertain if this will render properly (and the preview of the wizard can be tricky) so may I suggest:

<Note>

If you are using .NET Framework prior to 4.6.1, our [legacy SDK](https://docs.sentry.io/clients/csharp/) supports .NET Framework as early as 3.5.

src/wizard/dotnet/aspnet.md
```diff
@@ -0,0 +1,64 @@
...
+Add Sentry to `Global.asax.cs`:
```
PeloWriter, line 22:
```suggestion
You should `init` the Sentry SDK as soon as possible during your application load by adding Sentry to `Global.asax.cs`:
```

src/wizard/dotnet/aspnet.md
```diff
@@ -0,0 +1,64 @@
...
+    {
+        // Flushes out events before shutting down.
+        _sentry?.Dispose();
+    }
+}
+```
+
+### Documentation
+
+Once you've verified the package is initialized properly and sent a test event, consider visiting our [complete ASP.NET docs](https://docs.sentry.io/platforms/dotnet/guides/aspnet/).
```
PeloWriter, line 55: We can delete this. 

I don't see an intentional being sent? For other SDKs, the next lines are (pulling from Browser JS here):

I don't think this applies --> We recommend adjusting the value of `tracesSampleRate` in production. Learn more about configuring sampling in our [full documentation](https://docs.sentry.io/platforms/javascript/performance/sampling/).

But I think we still need this --> Then create an intentional error, so you can test that everything is working:

```js
myUndefinedFunction();
```

If you're new to Sentry, use the email alert to access your account and complete a product tour.

If you're an existing user and have disabled alerts, you won't receive this email.

src/wizard/dotnet/aspnet.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Samples
+
+You can find an example ASP.NET MVC 5 app with Sentry integrated [on this GitHub repository.](https://github.com/getsentry/examples/tree/master/dotnet/AspNetMvc5Ef6)
+
+See the following examples that demonstrate how to integrate Sentry with various frameworks.
+
+- [Multiple samples in the `dotnet` SDK repository](https://github.com/getsentry/sentry-dotnet/tree/main/samples) (**C#**)
+- [Basic F# sample](https://github.com/sentry-demos/fsharp) (**F#**)
```
PeloWriter, line 64: You know the audience better than me, but I'd suggest removing these lines. They should be in our docs, but not on the wizard page.

src/wizard/dotnet/aspnet.md
bruno-garcia: like this:

image


https://www.nuget.org/packages/Sentry/3.0.0-alpha.9

src/wizard/dotnet/aspnet.md
bruno-garcia: I tested out and it works. We have that currently in Sentry for the .NET SDK

src/wizard/dotnet/aspnet.md
bruno-garcia: Our docs don't really have any samples. I and many like me really like to look at code to learn how to use things so I tend to point to samples.

The busiest parts of the SDK repository are actually the samples.

Should we track clicks on those to see if it makes sense to have them or not?

PeloWriter: Completely agree that samples are gold for our audience. Yes, if we could leave this and track, it would be useful data. Thanks, Bruno!

src/wizard/dotnet/aspnet.md
bruno-garcia: > We can delete this.

Isn't it a good chance to tell the users about the docs? Once they've set the basics up, they should be ready to go.

I'll add a verify step.

PR #2807, opened 2020-12-23
meta: Implement stalebot

chadwhitacre:
https://app.asana.com/0/1199602781063150/1198192573081727

- Labels un-triaged issues with `Status: Stale` after three weeks of inactivity.
- Closes issues so labeled after one more week of inactivity.
- Use `Status: Accepted` to hold open indefinitely.


BYK (approved):
Verified that this is exactly the same as getsentry/onpremise#784 which was already approved.

PR #2810, opened 2020-12-27
fix(native): Suggest release build

bruno-garcia:
On Windows it builds in Debug mode by default.

See also:

#439
#442

### DIFF
--- a/src/includes/getting-started-install/native.mdx
+++ b/src/includes/getting-started-install/native.mdx
@@ -8,7 +8,7 @@ For example, `CMake` can be used like this (on macOS):
 # configure the cmake build into the `build` directory, with crashpad (on macOS)
 cmake -B build -D SENTRY_BACKEND=crashpad
 # build the project
-cmake --build build --parallel
+cmake --build build --config Release --parallel
 # install the resulting artifacts into a specific prefix
 cmake --install build --prefix install
 # which will result in the following (on macOS):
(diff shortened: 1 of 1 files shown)


jan-auer:
Re-assigning this. @Swatinem there's probably more we could be updating in the documentation from the readme. Particularly, builds, backends and configuration options have extensive info in the README that should rather go in docs.

src/includes/getting-started-install/native.mdx
```diff
@@ -6,9 +6,9 @@ For example, `CMake` can be used like this (on macOS):
 
 ```shell
 # configure the cmake build into the `build` directory, with crashpad (on macOS)
-cmake -B build -D SENTRY_BACKEND=crashpad
+cmake -B build -D SENTRY_BACKEND=crashpad --config RelWithDebInfo 
 # build the project
-cmake --build build --parallel
+cmake --build build --config RelWithDebInfo --parallel
```
Swatinem, line 11: can you also re-generate the file tree? now that we install debug symbols properly? (or to check if we actually do)

bruno-garcia: Are you asking if the pdb gets copied over on install?

bruno-garcia: bump?

Swatinem (approved):

PR #2812, opened 2020-12-29
Add default trigger note

rbisol:
Adding the default trigger "When an event occurs" note to alert triggers session as this question comes up quite frequently in support.


src/docs/product/alerts-notifications/issue-alerts.mdx
```diff
@@ -26,6 +26,10 @@ Triggers specify what type of activity you'd like monitored. The following trigg
 - The issue is seen more than {**value**} times in {**interval**}. **Value**: a positive integer. **Interval**: one min, one hour, one day, one week, or 30 days.
 - The issue is seen by more than {**value**} users in {**interval**}. **Value**: a positive integer. **Interval**: one min, one hour, one day, one week, or 30 days.
 
+<Alert>
+  If no trigger is selected, the alert will be triggered by every event, subject to rate limits.
+</Alert>
```
PeloWriter, line 31:
```suggestion
<Note>

If no trigger is selected, the alert will be triggered by every event, subject to rate limits.
  
</Note>
```

PeloWriter (approved):
Thank you. 
Minor change to align with other content (note also that we need lines prior to and after using <Note></Note> to ensure it works properly.

PR #2813, opened 2020-12-29
feat(apple/java): max attachment size

philipphofmann:
Add docs for max attachment size on Apple. This feature is currently
only supported on Apple, but more platforms are going to follow.
Therefore, we hide the feature for other platforms, but already use
PlatformContent.

### DIFF
--- a/src/platforms/common/enriching-events/attachments/index.mdx
+++ b/src/platforms/common/enriching-events/attachments/index.mdx
@@ -124,7 +124,7 @@ If you chose to limit the number of crash reports per issue, you can show _Only 
 ## Maximum Attachment Size
 
 You can set the maximum size for each attachment with `maxAttachmentSize` on the `SentryOptions`.
-The scale is MiB and the default is 20 MiB. Please also check the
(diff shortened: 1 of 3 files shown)


src/platforms/common/enriching-events/attachments/index.mdx
```diff
@@ -118,3 +118,16 @@ Alternately, attachments also appear in the _Attachments_ tab on the **Issue Det
...
 
 </PlatformSection>
+
+<PlatformSection supported={["android", "apple", "java"]}>
+
+## Maximum Attachment Size
+
+The maximum size for each attachment is set on `SentryOptions.maxAttachmentSize`.
+The scale is bytes and the default is `20 MiB`. Please also check the
+[maxium attachment size of Relay](https://docs.sentry.io/product/relay/options/)
```
bruno-garcia, line 128:
```suggestion
[maximum attachment size of Relay](https://docs.sentry.io/product/relay/options/)
```

bruno-garcia (approved):

PR #2815, opened 2020-12-30
fix: Update attach-stacktrace default

philipphofmann:
The attach-stacktrace on the options is default on for Android, Apple, and Java. This is now updated for the docs.

@bruno-garcia did we also set the default to `true` for .NET? If so ping me, so I update the PR.


bruno-garcia:
It's not yet: https://github.com/getsentry/sentry-dotnet/blob/fd271982fdceab2b2a0b4ac425bbf0b50c18db18/src/Sentry/SentryOptions.cs#L130

Thanks for checking. We'll change what stacktraces look like on v3.0 and it'll be quite a bit more expensive to build them (so they look legible), for that reason I want to test it out before setting it to `true` in there.

bruno-garcia (approved):
Nice!

PR #2818, opened 2020-12-31
Java: Connecting Errors with Spans

maciejwalkowiak:
Related to https://github.com/getsentry/sentry-dotnet/issues/663#issuecomment-75284005

### DIFF
--- a/src/platforms/java/common/performance/index.mdx
+++ b/src/platforms/java/common/performance/index.mdx
@@ -221,4 +221,8 @@ try {
 }
 ```
 
-Please note, that in Spring Boot integration, errors are linked with spans automatically and there is no need to follow these steps.
+<Note>
+
+In our Spring Boot integration, errors are linked with spans automatically.
+
+</Note>
--- a/src/platforms/java/common/performance/index.mdx
+++ b/src/platforms/java/common/performance/index.mdx
@@ -223,6 +223,6 @@ try {
 
 <Note>
 
-In our Spring Boot integration, errors are linked with spans automatically.
+In our Spring and Spring Boot integrations, errors are linked with spans automatically.
 
 </Note>
--- a/src/platforms/java/common/performance/index.mdx
+++ b/src/platforms/java/common/performance/index.mdx
@@ -173,7 +173,7 @@ Sentry errors can be linked with transactions and spans.
 
 Errors reported to Sentry while transaction or span is running are linked automatically:
 
-```java
+```java {tabTitle: Java}
 ISpan span = Sentry.startTransaction("task");
 try {
     processItem();
@@ -184,9 +184,20 @@ try {
 }
 ```
 
+```kotlin {tabTitle: Kotlin}
+val span = Sentry.startTransaction("task")
+try {
+    processItem()
+} catch (e: Exception) {
+    Sentry.captureException(e)
+} finally {
+    span.finish()
+}
+```
+
 In cases where the decision if thrown exception should be reported to Sentry happens later in the future after span is already finished, to connect error with a span you must link the exception with a Span by calling `setThrowable` method:
 
-```java
+```java {tabTitle: Java}
 ISpan span = Sentry.getSpan();
 try {
     processItem();
@@ -195,6 +206,18 @@ try {
     throw e;
 } finally {
     span.finish();
+}
+```
+
+```kotlin {tabTitle: Kotlin}
+val span = Sentry.getSpan()
+try {
+    processItem()
+} catch (e: Exception) {
+    span.setThrowable(e)
+    throw e
+} finally {
+    span.finish()
 }
 ```
 
--- a/src/platforms/java/common/performance/index.mdx
+++ b/src/platforms/java/common/performance/index.mdx
@@ -195,7 +195,7 @@ try {
 }
 ```
 
-In cases where the decision if thrown exception should be reported to Sentry happens later in the future after span is already finished, to connect error with a span you must link the exception with a Span by calling `setThrowable` method:
+In future, if a thrown exception should be reported to Sentry after the span has already finished, you must link the exception with the span by calling `setThrowable` method to connect the error:
 
 ```java {tabTitle: Java}
 ISpan span = Sentry.getSpan();
--- a/src/platforms/java/common/performance/index.mdx
+++ b/src/platforms/java/common/performance/index.mdx
@@ -184,9 +184,7 @@ try {
 }
 ```
 
-In cases where the decision if thrown exception should be reported to Sentry happens later in the future after span is already finished, to connect error with a span you must:
-
-1. Link exception with a Span by calling `setThrowable` method:
+In cases where the decision if thrown exception should be reported to Sentry happens later in the future after span is already finished, to connect error with a span you must link the exception with a Span by calling `setThrowable` method:
 
 ```java
 ISpan span = Sentry.getSpan();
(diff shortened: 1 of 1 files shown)


maciejwalkowiak:
cc @bruno-garcia

src/platforms/java/common/performance/index.mdx
```diff
@@ -166,3 +166,50 @@ if (span != null) {
...
+    span.finish();
+}
+
+```
+
+2. Retrieve span context associated with exception when reporting an error with `IHub#getSpan`:
+
+```java
+SentryEvent event = new SentryEvent(throwable);
+SpanContext spanContext = HubAdapter.getInstance().getSpan(ex);
```
bruno-garcia, line 208: If we're going for `HubAdapter` it means we need to rethink our API here.

Is `getSpan` supposed to be on the main `Sentry` class? 

Better:

Can we delete this section and simply have that in the docs? When we capture an exception we lookup for a span in the weakref? don't think this should be a user concern

bruno-garcia (changes requested):

src/platforms/java/common/performance/index.mdx
```diff
@@ -166,3 +166,50 @@ if (span != null) {
...
+if (spanContext != null) {
+  event.getContexts().setTrace(spanContext);
+}
+hub.captureEvent(event);
+```
```
bruno-garcia, line 213:
```suggestion
```

src/platforms/java/common/performance/index.mdx
maciejwalkowiak: This means that we would need to move `WeakHashMap<Throwable, SpanContext> throwableToSpanContext` from `Hub` to `SentryClient` so that it can be applied for `captureEvent`, `captureEvent` and `captureException`. Which is fine for me but I am not sure if it does not violate any rules regarding what can be inside `SentryClient` cc @marandaneto

marandaneto: `HubAdapter` is only meant for easing unit tests, not to be used externally, so definitely we'd need to rethink the API.

https://develop.sentry.dev/sdk/unified-api/
Hub should manage the state and `throwableToSpanContext` looks like managing the state.
Client is responsible for building the events only, no state at all afaik.

I didn't have the chance to read/test the whole transaction feature, not sure if I'm already able to give a better option, but working on it.

bruno-garcia (commented):
image

Seems like my comments where not sent, sorry

src/platforms/java/common/performance/index.mdx
bruno-garcia: Ignore this, lets discuss how to add this OOTB first

src/platforms/java/common/performance/index.mdx
```diff
@@ -166,3 +166,50 @@ if (span != null) {
...
 }
 ```
+
+### Connecting Errors with Spans
+
+Sentry errors can be linked with transactions and spans.
+
+Errors reported to Sentry while transaction or span is running are linked automatically:
+
+```java
```
marandaneto, line 176: should we add kotlin snippets as well?

src/platforms/java/common/performance/index.mdx
```diff
@@ -166,3 +166,50 @@ if (span != null) {
...
+ISpan span = Sentry.startTransaction("task");
+try {
+    processItem();
+} catch (Exception e) {
+    Sentry.captureException(e);
+} finally {
+    span.finish();
+}
```
marandaneto, line 184: should we split these 2 code snippets into the include folder? they are gonna be reused by Android and right now they are under `platforms/java/common/performance/index.mdx`.

otherwise, every fix should be copied over 2 MDX files

src/platforms/java/common/performance/index.mdx
```diff
@@ -166,3 +166,50 @@ if (span != null) {
...
+```java
+ISpan span = Sentry.getSpan();
+try {
+    processItem();
+} catch (Exception e) {
+    span.setThrowable(e);
+    throw e;
+} finally {
+    span.finish();
+}
```
marandaneto, line 200: take into consideration

https://github.com/getsentry/sentry-docs/pull/2818/files#r553794887
https://github.com/getsentry/sentry-docs/pull/2818/files#r553834912

src/platforms/java/common/performance/index.mdx
marandaneto: > Sorry but I don't follow:
> Why does it need to go to the client? The API entry point is `Sentry.captureX` which goes through the Hub. We don't want to have `getSpan` there? Is this method not in the spec?
> 
> Lets just remove this from the docs for now, this was created to support our Spring integration and doesn't need to be documented yet.

yeah I also don't follow, don't really understand why we need that, I'm blind here.

src/platforms/java/common/performance/index.mdx
maciejwalkowiak: 👍

maciejwalkowiak:
Docs adjusted to changes made in PR: https://github.com/getsentry/sentry-java/pull/1156

src/platforms/java/common/performance/index.mdx
```diff
@@ -166,3 +166,59 @@ if (span != null) {
...
+try {
+    processItem()
+} catch (e: Exception) {
+    Sentry.captureException(e)
+} finally {
+    span.finish()
+}
+```
+
+In cases where the decision if thrown exception should be reported to Sentry happens later in the future after span is already finished, to connect error with a span you must link the exception with a Span by calling `setThrowable` method:
```
PeloWriter, line 198: This is confusing. I think this means (and feel free to modify if I've got this wrong):
```suggestion
In future, if a thrown exception should be reported to Sentry after the span has already finished, you must link the exception with the span by calling `setThrowable` method to connect the error:
```

src/platforms/java/common/performance/index.mdx
```diff
@@ -166,3 +166,59 @@ if (span != null) {
...
+    processItem()
+} catch (e: Exception) {
+    span.setThrowable(e)
+    throw e
+} finally {
+    span.finish()
+}
+```
+
+Please note, that in Spring Boot integration, errors are linked with spans automatically and there is no need to follow these steps.
```
PeloWriter, line 224: Move this up earlier in the page so Spring Boot users don't get to the end, only to learn they didn't need to read this far :).
```suggestion
<Note>

In our Spring Boot integration, errors are linked with spans automatically.

</Note>
```

PeloWriter (approved):
Couple of changes (one of which is an edit that I _think_ captures what we mean to explain, but please review to ensure it doesn't change the meaning)

src/platforms/java/common/performance/index.mdx
bruno-garcia, line 224: Isn't it "on our Spring Integration"?
Is this exclusive to Spring Boot? We've got customers asking for this that use Spring but not Spring Boot.

bruno-garcia (approved):
Small note on Spring Boot vs Spring but other than that LGTM

src/platforms/java/common/performance/index.mdx
maciejwalkowiak: Correct, updated.

src/platforms/java/common/performance/index.mdx
maciejwalkowiak: I am not sure if "In future" here is understandable for everyone. Maybe we can just leave "If a thrown exception ..." ?

maciejwalkowiak:
Since these changes can be merged after next 4.0 is released, I moved changes from this PR to https://github.com/getsentry/sentry-docs/pull/2833.

PR #2819, opened 2021-01-01
Prefix PHP tracing examples with namespaces

stayallive:
This is to "fix" copy pasta errors where users paste in the examples but they won't work because of missing imports or being inside their own namespace causing unneeded confusion.

Fixes getsentry/sentry-php#1159.

/cc @HazAT


stayallive:
@HazAT can you give a OK/NOT OK here?

HazAT (approved):

Issue #2821, opened 2021-01-04
Identifying users from their IP Address


MisterTea:
https://docs.sentry.io/platforms/native/enriching-events/identify-user/

According to ^ users are identified by their ip address by default.  I have a test branch with Sentry that is logging events but the user count is zero.  Is there something I have to do to turn on ip address identification for users?  If I want Sentry to use the incoming IP address, what should I put when setting up the user to specify this?  Something like:

```
sentry_value_t user = sentry_value_new_object();
sentry_value_set_by_key(user, "ip_address", NULL);  // This probably doesn't work but ideally it would let Sentry guess the ip from the http connection.
sentry_set_user(user);
```

MisterTea:
@jan-auer I tried that and it didn't work until I put the sentry_set_user after the sentry_init.  Please add that to the documentation as well, thanks.

jan-auer:
@MisterTea do you mean as opposed to putting it before `sentry_init`?

MisterTea:
Right, it didn't work when I had it before sentry_init.

jan-auer:
`sentry_init` always has to be first, before you initialize you cannot use any of the scope or capture methods. 

@PeloWriter I'd love your take on this particular question. Which would you prefer:

- We mention this on _Getting Started_ (under #configure) 
- We mention this on each of the pages under "Enriching Events", "Configuration" and "Usage" where this applies
- Both of the above

PR #2822, opened 2021-01-04
fix(native): Document ip_address {{auto}}

jan-auer:
Documents the special `{{auto}}` value for `user.ip_address` for all SDKs.
Client-side SDKs set this by default, whereas in server-side SDKs this must be
set manually.

Since the Native SDK is used in both contexts, we document `{{auto}}` explicitly
in the code example.

---

image

---

Fixes https://github.com/getsentry/sentry-docs/issues/2821


Swatinem (approved):

untitaker:
K lgtm

PR #2825, opened 2021-01-05
unity-lite

bruno-garcia:
Sentry Unity Lite isn't a full Sentry SDK so lots of the current docs don't apply here.

How do I get the icons in?

They exist in the `platformicons`:
https://github.com/getsentry/platformicons/blob/master/svg/unity.svg
https://github.com/getsentry/platformicons/blob/master/svg_80x80/unity.svg

Seems we need to bump it


bruno-garcia:
Build is broken:

>   platforms/unity/configuration/options#send-default-pii

Can't find a way to opt out of `configuration`

src/platforms/unity/index.mdx
```diff
@@ -0,0 +1,32 @@
+---
+title: Unity
```
PeloWriter, line 2: Can you not adopt the common content here? https://docs.sentry.io/contributing/approach/write-getting-started/

PeloWriter (commented):
It would be easier to adopt the common content for the Getting Started page, I've provided a link to those guidelines. But the Tl;dr is that you need only snap in the code samples, little to no prose required.

src/platforms/unity/index.mdx
bruno-garcia: I'll try. This SDK has nothing common with any other but I assume a "install, configure, test" I anyway added here so I can split it in 3 files.

Issue #2824, opened 2021-01-04
Apple Attachment docs doesn't mention `content-type`


bruno-garcia:
Came up the forum: https://forum.sentry.io/t/ios-screenshot-attachments/12285/3

It's possible that the Android/Java/.NET Docs are also lacking since this served as the base docs for the other.

PR #2826, opened 2021-01-05
fix(apple/java): Improve docs for attachments

philipphofmann:
The docs for attachments didn't mention content type. Now the docs are
restructured so they give more insight on how to use attachments.

Fixes GH-2824


src/platforms/common/enriching-events/attachments/index.mdx
```diff
@@ -41,6 +41,31 @@ To receive symbolicated stack traces, you have to upload debug information to Se
...
+<PlatformContent includePath="attachment-init-with-path" />
+
+Another way to initialize an attachment is to pass the `bytes`. When doing so you also need to specify a filename.
+
+<PlatformContent includePath="attachment-init-with-bytes" />
+
+If your SDK supports offline caching, which is typical for mobile, each attachment is stored to disk for each event or transaction you capture. When using large
+attachments this can eat up the disk space if the device is offline for a longer time period. You can use [maximum attachment size](#maximum-attachment-size) to drop large attachments.
+
+Additional parameters:
```
PeloWriter, line 61:
```suggestion
In addition, you can set these parameters:
```

src/platforms/common/enriching-events/attachments/index.mdx
```diff
@@ -41,6 +41,31 @@ To receive symbolicated stack traces, you have to upload debug information to Se
...
+
+**`filename`**: The filename is the name of the file to display in Sentry. When using bytes you have to specify a filename, whereas with a path you don't as the SKD is going to use the last path component.
+
+**`contentType`**: The `contentType` of the attachment. The default is `application/octet-stream`.
```
PeloWriter, line 65:
```suggestion
`filename`

: The filename is the name of the file to display in Sentry. When using bytes you have to specify a filename, whereas with a path you don't as the SKD is going to use the last path component.

`contentType`

: The type of content stored in this attachment. The default is `application/octet-stream`.
```

PeloWriter (approved):
Nicely done. A few edits to fix format for `filename` and `contentType` and smooth the prose.

borninla:
Not sure if it belongs in this specific article, but it would be nice to talk about some (if not most) of the options that can be specified for `contentType`.

philipphofmann:
Thanks a lot @PeloWriter for the feedback 👏🏼

philipphofmann:
@borninla all media types specified here: https://www.iana.org/assignments/media-types/media-types.xhtml. I'm going to add this to the docs.

philipphofmann:
I opened up two PRs @borninla https://github.com/getsentry/develop/pull/238 and https://github.com/getsentry/sentry-docs/pull/2843.

PR #2827, opened 2021-01-05
feat(apple): Not use versions 6.0.2 to 6.0.8

philipphofmann:


src/platforms/apple/common/usage.mdx
```diff
@@ -53,4 +53,6 @@ This integration will swizzle some methods to create breadcrumbs e.g.: for `UIAp
 
 ## Known Limitations
 
+- We strongly recommend you to not use versions 6.0.2 to 6.0.8. With [#786](https://github.com/getsentry/sentry-cocoa/pull/786) released in 6.0.2, we introduced a severe bug,
+that could lead to wrong stacktraces and error messages. We fixed this in 6.0.9.
```
bruno-garcia, line 57: What about a bit less alarming, like:

```suggestion
- If you're currently using a version between 6.0.2 and 6.0.8 we strongly recommend you to upgrade it to the latest. In those versions the [SDK had a regression](https://github.com/getsentry/sentry-cocoa/pull/786) that affected the quality of the stacktraces and error messages. We fixed this in 6.0.9.
```

src/platforms/apple/common/usage.mdx
```diff
@@ -53,4 +53,5 @@ This integration will swizzle some methods to create breadcrumbs e.g.: for `UIAp
 
 ## Known Limitations
 
+- If you're currently using a version between 6.0.2 and 6.0.8 we strongly recommend you to upgrade it to the latest. In those versions the [SDK had a regression](https://github.com/getsentry/sentry-cocoa/pull/786) that affected the quality of the stacktraces and error messages. We fixed this in 6.0.9.
```
PeloWriter, line 56:
```suggestion
- If you're currently using a version between 6.0.2 and 6.0.8, we **highly recommend** you upgrade to the latest version to avoid [a regression](https://github.com/getsentry/sentry-cocoa/pull/786) that affected the quality of the stacktraces and error messages, which was fixed in 6.0.9.
```

PeloWriter (approved):
Thanks.

PR #2828, opened 2021-01-05
Remove border and drop shadow from platform icons in guide

torarnv:
The platform icons from the 'platformicons' package do not have any
border or drop shadow, but type.scss adds them for any image inside
the 'prose' class, which the guide icons are automatically part of,
since they art part of the main content block.

A similar workaround is already added in platformGrid.tsx, while the
icons in the platform navigation dropdown do not need it as they are
not part of the main content (prose section).

### DIFF
--- a/src/components/guideGrid.tsx
+++ b/src/components/guideGrid.tsx
@@ -23,7 +23,7 @@ export default ({ platform }: Props): JSX.Element => {
             <PlatformIcon
               size={16}
               platform={guide.key}
-              style={{ marginRight: "0.5rem" }}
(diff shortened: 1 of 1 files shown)


torarnv:
Ideally the border and drop shadow would be removed in a single place,
but I did not find any way to do that. Let me know if it's possible.

torarnv:
cc @bruno-garcia

torarnv:
Before:
image

After:
image

bruno-garcia:
Oh that's awesome! Thanks @torarnv

bruno-garcia (approved):

PR #2835, opened 2021-01-06
Fix useplatform

torarnv:


torarnv:
Moved to https://github.com/getsentry/sentry-docs/pull/2842

PR #2840, opened 2021-01-07
frontmatter Android categories

marandaneto:


philipphofmann (commented):
I don't really understand what these changes are doing, but should I also apply them for Apple?

PeloWriter (approved):

bruno-garcia:
honestly I also don't know what `sdk` is for. There was a thread before where I tried to rename one and things broke. I know they don't really map 1 to 1 with the `event.sdk.name`. @PeloWriter are there any docs on the front matter we can catch up with?

Also, what's the category used for? Is it freeform? Can we use `desktop` for example when it makes sense?

PeloWriter:
Yes, SDK pages are discussed in our Contributor Guidelines - here's the link to the [frontmatter](https://docs.sentry.io/contributing/pages/frontmatter/) explainer.

PeloWriter:
In reading that page, though, it doesn't address the items you have questions about. @dcramer - can we update this page to more fully describe the components of the yaml frontmatter?

marandaneto:
> Also, what's the category used for? Is it freeform? Can we use `desktop` for example when it makes sense?

yes, https://github.com/getsentry/sentry-docs/blob/dd7208a0cd9d82fa742a6847fcb1f2c8d4d423ed/src/gatsby/createSchemaCustomization/platformSchema.ts#L16-L22

marandaneto:
> I don't really understand what these changes are doing, but should I also apply them for Apple?

I'm not sure what exactly it does but the `categories` I was assuming the wizard, as we have similar categories in there.
I added them because the Java one does have it and they were missing on the Android config.

Looks like the Apple one is missing `categories` only

marandaneto:
follow up for Java #2848

bruno-garcia (approved):

PR #2848, opened 2021-01-08
frontmatter Java categories

marandaneto:


src/platforms/java/config.yml
```diff
@@ -2,3 +2,6 @@ title: Java
 caseStyle: camelCase
 supportLevel: production
 sdk: sentry.java
+categories:
+  - desktop
+  - server
```
marandaneto, line 7: we could also add `serverless` but we don't have official docs about it yet, same for Dart, so I guess its not a good idea.

philipphofmann (approved):
LGTM.

src/platforms/java/config.yml
philipphofmann: I wouldn't add serverless then.

bruno-garcia (approved):
LGTM

PR #2851, opened 2021-01-08
Performance docs for Android

### LINKED ISSUE #2849: Missing Java/Kotlin imports
Most of the Java/Kotlin code snippets are missing the imports, I noticed most of them in the performance docs

We need them because of https://develop.sentry.dev/sdk/philosophy/#assume-novice-developers

so a plain copy-paste would work.

cc @maciejwalkowiak

marandaneto:
Add Performance docs for Android
Also fixes #2849


src/docs/product/performance/getting-started.mdx
```diff
@@ -17,12 +17,17 @@ Customers on legacy plans must add transaction events to their subscription in o
...
 
+- [Java](/platforms/java/performance/)
+
 - [JavaScript](/platforms/javascript/performance/)
   - [Angular](/platforms/javascript/guides/angular/performance/)
   - [Ember](/platforms/javascript/guides/ember/performance/)
   - [React](/platforms/javascript/guides/react/#monitor-performance)
   - [Vue](/platforms/javascript/guides/vue/#monitor-performance)
 
+- Mobile
```
marandaneto, line 28: @bruno-garcia not sure about a Mobile section here but under Java would be kinda weird if we now have a Kotlin language, wdyt?

bruno-garcia: Make sense to me to categorize like you did. We could rename Java to [Java/Kotlin/JVM] ?

src/docs/product/performance/getting-started.mdx
```diff
@@ -17,12 +17,17 @@ Customers on legacy plans must add transaction events to their subscription in o
 
 ## Supported SDKs
 
+- [Java](/platforms/java/performance/)
```
bruno-garcia, line 20:
```suggestion
- [Java/Kotlin/JVM](/platforms/java/performance/)
```

src/includes/performance/enable-tracing/android.mdx
```diff
@@ -0,0 +1,38 @@
+```xml {filename:AndroidManifest.xml}
+<application>
+  <meta-data android:name="io.sentry.traces.sample-rate" android:value="1.0" />
+</application>
+```
+
+Or, if you are manually instrumenting Sentry:
```
bruno-garcia, line 7: Opting in to tracing with the sampling decision isn't really related to manual vs auto instrumentation (we can start capturing transactions automatically if we are able to if the users set a sampling decision in):

```suggestion
Or, if you are initializing the SDK programatically:
```

src/includes/performance/enable-tracing/android.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+```java {tabTitle:Java}
```
bruno-garcia, line 9: What about having Kotlin first?

Maybe better as a single PR across all of docs, right?

src/includes/performance/traces-sample-rate/android.mdx
```diff
@@ -0,0 +1,26 @@
+```xml {filename:AndroidManifest.xml}
+<application>
+  <meta-data android:name="io.sentry.traces.sample-rate" android:value="0.2" />
+</application>
+```
+
+Or, if you are manually instrumenting Sentry:
```
bruno-garcia, line 7:
```suggestion
Or, if you are initializing the SDK programatically:
```

src/includes/performance/traces-sample-rate/android.mdx
```diff
@@ -0,0 +1,26 @@
...
+
+```java {tabTitle:Java}
+import io.sentry.android.core.SentryAndroid;
+
+SentryAndroid.init(this,
+  options -> {
+    ...
```
bruno-garcia, line 14:
```suggestion
```

src/includes/performance/traces-sampler-as-sampler/android.mdx
```diff
@@ -0,0 +1,62 @@
...
+      context -> {
+        CustomSamplingContext ctx = context.getCustomSamplingContext();
+        if (ctx != null) {
+          if ("/payment".equals(ctx.get("url"))) {
+            // These are important - take a big sample
+            return 0.5;
+          } else if ("/search".equals(ctx.get("url"))) {
+            // Search is less important and happen much more frequently - only take 1%
+            return 0.01;
+          } else if ("/health".equals(ctx.get("url"))) {
```
bruno-garcia, line 20: Maybe something more "mobile"?

```suggestion
          if ("LoginActivity".equals(ctx.get("ActivityName"))) {
            // These are important - take a big sample
            return 0.5;
          } else if ("SearchActivity".equals(ctx.get("ActivityName"))) {
            // Search is less important and happen much more frequently - only take 1%
            return 0.01;
          } else if ("HealthCheckActivity".equals(ctx.get("ActivityName"))) {
```

src/includes/performance/enable-tracing/android.mdx
marandaneto: this was the wording defined by the docs team about "auto init mode" vs "initializing the SDK by yourself".
I'd stick to it or we'd need to change everywhere, otherwise, only this page would differ.

it's not really related to tracing, but if you have the auto init mode or not

src/includes/performance/enable-tracing/android.mdx
marandaneto: I'd say so, also would need checking if docs somehow would order alphabetically or not

src/includes/performance/traces-sample-rate/android.mdx
marandaneto: this is in all the performance Java/Kotlin examples, if we decide to remove it, I'd do a single PR.

src/includes/performance/traces-sampler-as-sampler/android.mdx
marandaneto: good catch, makes sense

marandaneto:
@maciejwalkowiak I've changed a few scripts to be compatible with the new code like using interfaces instead of the implementation class, also nullability checks, added imports, and returning Doubles instead of Ints otherwise it'd never compile.
Please check, thanks

maciejwalkowiak (approved):
Looks good to me.

src/includes/java/performance/getting-started/java.mdx
```diff
@@ -0,0 +1,168 @@
...
+} finally {
+    transaction.finish();
+}
+```
+
+### Adding More Spans to the Transaction
+
+The next example contains the implementation of the hypothetical `processItem` function called from the code snippet in the previous section. Our SDK can determine if there is currently an open transaction and add all newly created spans as child operations to that transaction. Keep in mind that each individual span also needs to be manually finished; otherwise, spans will not show up in the transaction.
+
+In cases where you want to attach Spans to an already ongoing Transaction you can use `Sentry#getSpan`. This method will return a `SentryTransaction` in case there is a running Transaction or a `Span` in case there is already a running Span, otherwise it returns `null`.
```
marandaneto, line 69: @maciejwalkowiak should we still keep `SentryTransaction` and `Span` or just `ISpan`?

src/includes/java/performance/getting-started/java.mdx
```diff
@@ -0,0 +1,168 @@
...
+        }
+    }
+}
+```
+
+Spans are sent together with their parent transaction when the transaction is finished. Make sure to call `finish()` on transaction once all the child spans have finished.
+
+### Retrieving a Transaction
+
+In cases where you want to attach Spans to an already ongoing Transaction you can use `Sentry#getSpan`. This method will return a `SentryTransaction` in case there is a running Transaction or a `Span` in case there is already a running Span, otherwise it returns `null`.
```
marandaneto, line 124: same as https://github.com/getsentry/sentry-docs/pull/2851/files#r557241698

dashed (approved):

src/includes/performance/enable-tracing/android.mdx
bruno-garcia: @PeloWriter maybe has an opinion here

bruno-garcia (approved):
🚀

PR #2852, opened 2021-01-08
feat(apple): Add categories to config

philipphofmann:
Opened because https://github.com/getsentry/sentry-docs/pull/2848 and https://github.com/getsentry/sentry-docs/pull/2840

### DIFF
--- a/src/platforms/apple/config.yml
+++ b/src/platforms/apple/config.yml
@@ -4,3 +4,7 @@ caseStyle: canonical
 supportLevel: production
 aliases:
   - cocoa
+categories:
+  - mobile
+  - desktop
+  - server
--- a/src/platforms/apple/config.yml
+++ b/src/platforms/apple/config.yml
@@ -7,4 +7,3 @@ aliases:
 categories:
   - mobile
   - desktop
-  - server


src/platforms/apple/config.yml
```diff
@@ -4,3 +4,7 @@ caseStyle: canonical
 supportLevel: production
 aliases:
   - cocoa
+categories:
+  - mobile
+  - desktop
+  - server
```
marandaneto, line 10: does sentry cocoa support server? TIL

PeloWriter (approved):

src/platforms/apple/config.yml
bruno-garcia, line 10: Wouldn't really work on server due to single hub

bruno-garcia, line 10:
```suggestion
```

src/platforms/apple/config.yml
```diff
@@ -4,3 +4,6 @@ caseStyle: canonical
 supportLevel: production
 aliases:
   - cocoa
+categories:
+  - mobile
+  - desktop
```
bruno-garcia, line 9: For watchOS, tvOS we could add wearable and .. iot?

```suggestion
  - desktop
  - wearable
```

marandaneto: it'd be an invalid value though, I'd stick to mobile desktop/mobile only

https://github.com/getsentry/sentry-docs/blob/dd7208a0cd9d82fa742a6847fcb1f2c8d4d423ed/src/gatsby/createSchemaCustomization/platformSchema.ts#L16-L22

bruno-garcia: TIL, thanks

bruno-garcia (approved):

src/platforms/apple/config.yml
philipphofmann: Whoops you are both right @marandaneto and @bruno-garcia. It should work on server but it doesn't really.

PR #2855, opened 2021-01-08
fix: Check in lockfile changes and update CI

untitaker:
Make CI prevent updating package.json without updating lockfile.


bruno-garcia:
@untitaker funny thing is that the platformicons package did update, and we see the new icons on the docs

bruno-garcia (approved):

untitaker:
yes @bruno-garcia you bumped package.json which is used by both package managers

PR #2860, opened 2021-01-09
fix(react): Fix docs for ErrorBoundary fallback prop

devinsm:
The docs previously showed passing a component to the fallback prop. However the fallback prop takes a React _element_ not a React _component_:
https://github.com/getsentry/sentry-javascript/blob/a833f0ec0a186f0672a536ea8f7d3b129a71a4fe/packages/react/src/errorboundary.tsx#L111-L121

You can additionally verify that by checking the TypeScript types:
https://github.com/getsentry/sentry-javascript/blob/a833f0ec0a186f0672a536ea8f7d3b129a71a4fe/packages/react/src/errorboundary.tsx#L30

Since the difference between React elements and React components can be confusing, here's [a link to the React glossary for more information](https://reactjs.org/docs/glossary.html#elements).

### DIFF
--- a/src/platforms/javascript/guides/react/components/errorboundary.mdx
+++ b/src/platforms/javascript/guides/react/components/errorboundary.mdx
@@ -45,7 +45,7 @@ function FallbackComponent() {
 class App extends React.Component {
   render() {
     return (
-      <Sentry.ErrorBoundary fallback={FallbackComponent} showDialog>
+      <Sentry.ErrorBoundary fallback={<FallbackComponent />} showDialog>
         <Example />
       </Sentry.ErrorBoundary>
     );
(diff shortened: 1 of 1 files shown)


src/platforms/javascript/guides/react/components/errorboundary.mdx
```diff
@@ -45,7 +45,7 @@ function FallbackComponent() {
 class App extends React.Component {
   render() {
     return (
-      <Sentry.ErrorBoundary fallback={FallbackComponent} showDialog>
+      <Sentry.ErrorBoundary fallback={<FallbackComponent />} showDialog>
```
lobsterkatie, line 48:
```suggestion
      <Sentry.ErrorBoundary fallback={myFallback} showDialog>
```

or

```suggestion
      <Sentry.ErrorBoundary fallback={showThisInstead} showDialog>
```

lobsterkatie (commented):
Thanks for doing this, and for the helpful link!

I think the prose changes are good, but given that the `fallback` value can be either an element or a function, I'd rather make this more generic and give it a variable name. (See my suggestions below.) What do you think?

src/platforms/javascript/guides/react/components/errorboundary.mdx
devinsm: @lobsterkatie Works for me. I updated the example to use a variable (and show what you might assign that variable to). Let me know what you think.

lobsterkatie: Nice. I like the examples.

lobsterkatie (approved):
Thanks for doing this!

PR #2861, opened 2021-01-11
Update release version restrictions - CLI

rbisol:


src/docs/product/cli/releases.mdx
```diff
@@ -18,8 +18,8 @@ Because releases work on projects you will need to specify the organization and
 
 Releases are created with the `sentry-cli releases new` command. It takes at the very least a version identifier that uniquely identifies the releases. There are a few restrictions -- the release name cannot:
 
-- contain newlines or spaces
-- use a forward slash ("/"), back slash ("\"), period ("."), or double period ("..")
```
kamilogorek, line 22: It also cannot start or end with a spaces.

kamilogorek (changes requested):

src/docs/product/cli/releases.mdx
rbisol: I think for starting and ending with string is safer to trim the string on the CLI side. This change is mostly to keep the documentation consistent with the SDK release guide.

kamilogorek (approved):

PR #2865, opened 2021-01-12
Update azure-sso.mdx

georgekallilas:
Removed "Azure" from Step 9 since it does not exist and instead "Active Directory" is used.


PeloWriter (approved):
Thank you.

PR #2866, opened 2021-01-12
fix: Update docs to the most recent Rust SDK version

Swatinem:
This uses the current release-registry version of the Rust SDK.
It also updates some examples to use Releases, and updates the Actix docs to the current version.

### DIFF
--- a/src/wizard/rust/index.md
+++ b/src/wizard/rust/index.md
@@ -9,20 +9,26 @@ To add Sentry to your Rust project you just need to add a new dependency to your
 
 ```toml
 [dependencies]
-sentry = "0.19.0"
+sentry = "{{ packages.version('sentry.rust') }}"
 ```
 
 `sentry.init()` will return you a guard that when freed, will prevent process exit until all events have been sent (within a timeout):
 
 ```rust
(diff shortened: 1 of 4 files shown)


src/wizard/rust/index.md
```diff
@@ -9,20 +9,26 @@ To add Sentry to your Rust project you just need to add a new dependency to your
 
 ```toml
 [dependencies]
-sentry = "0.19.0"
+sentry = "{{ packages.version('sentry.rust') }}"
```
untitaker, line 12: fairly certain the tags won't work here. Have you tried this?

Swatinem: Works fine for me locally

untitaker (approved):

src/wizard/rust/index.md
Swatinem: Also see https://sentry-docs-git-fix-rust-version-update.sentry.dev/platforms/rust/#install ;-)

untitaker: This is for the wizard when creating a new project in Sentry. It's not the same as the src/platforms content

PR #2882, opened 2021-01-13
fix(python): Make default sampling context comments more helpful

lobsterkatie:
In particular, this points people to look in `wsgi_environ` for (all the parts of) the request URL.

Fixes https://github.com/getsentry/sentry-python/issues/969.


untitaker (approved):
🥚-celent

PR #2884, opened 2021-01-14
fix(ui): added styles for active tab in code blocks

robinrendle:
### Before

Screen Shot 2021-01-13 at 4 03 02 PM

### After

Screen Shot 2021-01-13 at 4 02 16 PM


EvanPurkhiser (approved):
Thanks for this

Issue #2870, opened 2021-01-12
Fix Data Filtering IA


PeloWriter:
In Release Health, we have a section regarding Data Filtering. It's not the right place:

- [x] Move this content to be part of our content in Product > Error Monitoring > Inbound Data Filters @PeloWriter 
- [x] Move Inbound Data Filters to Product > Data Management
- [x] In the Error Monitoring section of the filters page, develop content that clarifies each of our filtering types (perhaps a page for each, including examples of usage) - the content has been moved, but we need to further develop this content
- [ ] Explain how sampling works @RaduW

PeloWriter:
PR here that captures the first item here - https://github.com/getsentry/sentry-docs/pull/2885

doralchan:
@PeloWriter These filters are also applicable under transactions

PeloWriter:
Based on @doralchan 's comment, we should move this out of Error Monitoring once filters apply to xns, as well.

jan-auer:
@PeloWriter updated the issue description, to reflect our last discussion. This must've mixed up somewhere in note taking. We discussed to move this under "Data Management".

PeloWriter:
ah, great catch. Thanks. I was going to make that suggestion, so glad we are aligned. I will incorporate this location change also since the content work has begun in https://github.com/getsentry/sentry-docs/pull/2885

PeloWriter:
For this iteration, we can close the issue. Data Filtering IA is current for feature set.

Issue #2869, opened 2021-01-12
Fix Release Health IA


PeloWriter:
We'd like our Release Health page to be more consumable as we expand this feature across SDKs:

- [x] Remove the Releases Index topic from the page to either add to the Releases Index main page or remove entirely (if redundant) @PeloWriter 
- [x] Move Release Details onto its own page and expand the content @PeloWriter to create the structure
- [x] Remove duplication of definitions (Crashed, Abnormal, Errored, Healthy particularly) @PeloWriter 
- [x] Blend the glossary terms into the content so they are defined with context rather in a never-read glossary @PeloWriter 
- [x] Modify the Getting Started to mirror the approach we use in Performance @PeloWriter 
- [x] Add the minimum version information for release health into the appropriate SDK @PeloWriter 
- [ ] Review definitions for accuracy.

PeloWriter:
https://github.com/getsentry/sentry-docs/pull/2885

PeloWriter:
@jan-auer - should we close, and pull out the definitions items separately?

PR #2885, opened 2021-01-14
Structural changes and edits for Release Health

PeloWriter:
Initial changes, as captured in https://github.com/getsentry/sentry-docs/issues/2869 and part of https://github.com/getsentry/sentry-docs/issues/2870.


src/includes/configuration/auto-session-tracking/android.mdx
```diff
@@ -1,3 +1,5 @@
+To benefit from the health data, you must use at least version 2.1.0 of the Android SDK, and enable the collection of release health metrics in the AndroidManifest.xml file. Release health is captured, by default, in our Android version 3.0 and higher, unless you specifically disabled collecting it in the initialization options for the SDK.
```
marandaneto, line 1: should we stick to either `in the AndroidManifest.xml file` or `in the initialization options`? because they are 2 different ways of configuring the Android SDK.

marandaneto, line 1:
```suggestion
To benefit from the health data, you must use at least version 2.1.0 of the Android SDK, and enable the collection of release health metrics in the initialization options for the SDK. Release health is captured, by default, in our Android version 3.0 and higher, unless you specifically disabled collecting it in the initialization options for the SDK.
```

src/docs/product/releases/health/setup.mdx
```diff
@@ -0,0 +1,15 @@
+---
+title: Set Up
+sidebar_order: 40
+description: "Learn about setting up your SDK to enable Release Health."
+---
+
+To enable reporting of release health, use the link below to quickly configure your SDK.
+
+## Supported SDKs
```
marandaneto, line 9: Rust and Python should be here too

src/docs/product/error-monitoring/filtering/index.mdx
```diff
@@ -23,3 +25,35 @@ Inbound filters include:
 Once applied, you can track the filtered events (numbers and cause) using the graph provided at the top of the Inbound Data Filters view.
 
 ![](builtin-inbound-filters.png)
+
+## Filtering for Session Events
+
+Data filters are only partially supported for session events, and they work slightly differently:
+
+### Equivalent Filters
```
jan-auer, line 33: I checked again and I'm afraid that filters do not apply to sessions at all. We will revisit this in the future, so lets just state that inbound data filters do not impact sessions.

At the same time, _most_ filters will apply to transactions just like error events. That is, as long as they make sense. The one exception is Error Message filters. They obviously cannot apply to transactions since those do not have an error message.

src/docs/product/releases/health/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+Monitor the health of releases by observing user adoption, usage of the application, percentage of [crashes](#crash), and [session data](#session-data). Release health will provide insight into the impact of crashes and bugs as it relates to user experience, and reveal trends with each new issue through the release details, graphs, and filters.
+
+## Enabling Release Health
+
+Many SDKs automatically manage the start and end of sessions when the SDK is initialized, but release health configuration is key to ensuring you are receiving data that is useful, so check our information on [which SDKs support release health](/product/releases/health/setup/) for links to configuring the SDK, as needed.
+
+## Data Filtering
+
+Sentry provides [inbound data filters](/product/error-monitoring/filtering/) to filter data into your project. Data filters are only partially supported for session events.
```
jan-auer, line 18: Let's remove the data filtering section as per my comment above.

src/docs/product/releases/health/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+## What Release Health Monitors
```
jan-auer, line 20: May I ask, where did you get the term "monitor" from? We do have a concept of monitors in Sentry, but it does not relate to sessions at all.

src/docs/product/releases/health/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+Once you configure your SDK, Sentry connects the metrics to the specific release of your application and the associated code.
+
+### Session Data
+
+The primary component Sentry uses to monitor health is a _session_, which represents the interaction between the user and the application.
+
+A session begins with the start of the application. Or, it begins with bringing the already started application back from background to the foreground.
```
jan-auer, line 28: Let's keep this on our backlog for later: The definition of what a session is varies by platform. The original content was written for mobile, so it may no longer be accurate, especially for JavaScript. 

I would propose we follow up with this and give a general notion of what a session is (along the lines of "a continuous interaction of a user with your application") and then give examples for various platforms.

src/docs/product/releases/health/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+Sessions are submitted to Sentry so you can track the usage and adoption of your application. When a user of your application experiences a crash, error, or abnormal exit, the session will be flagged accordingly, and Sentry calculates derived metrics. The metrics include data such as the number of users that didn't experience a crash in the specified time range.
+
+### Active Users
+
+Number of users that started the application at least once in the last 24 hours.
+
+### Crash
+
+The fatal error that caused the crash of the application. Errors that did not cause the end of the application should not be included.
```
jan-auer, line 40: Let's please use the definitions from [here](https://docs.sentry.io/product/releases/health/#release-details) instead, or at least align the description:

> The app had an explicit unhandled error or hard crash. You'll typically be able to view the corresponding issue in Sentry that captures this event. To search for them in Discover or on the Issues page, filter by `handled:no`.


Particularly, the part "fatal error  that caused the crash" is incorrect. Any unhandled error, whether or not actually crashing the application, sets the session to "crashed". Really, the only time we should be mentioning the word crash is in the title of this state.

src/docs/product/releases/health/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+Sentry considers every error event with the field `handled` set to `false` as a Crashed state.
+
+### Crash Free User
+
+Percentage of users who did not experience a crash during the specified time period.
```
jan-auer, line 46:
```suggestion
Percentage of distinct users who did not experience a crash during the specified time period.
```

Do you think we should add the definition of what a user is here?

src/docs/product/releases/health/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+### Crash Free Sessions
+
+Number of sessions in the specified time range that did not end by a crash of the application.
+
+### Crashed Users
+
+Number of users that experienced a crash in the specified time range.
```
jan-auer, line 54: This is actually confusing now. Why is "Crash Free User" a percentage and "Crashed Users" a count? I'll have to check if we really use it like this in the product, but there's some inconsistency.

src/docs/product/releases/health/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+### Release Adoption
+
+Number of users that started at least once during the specified time period of a specific release.
+
+#### Release Version
```
jan-auer, line 60:
```suggestion
### Release Version
```

src/docs/product/releases/health/release-details.mdx
```diff
@@ -0,0 +1,32 @@
...
+description: "Learn more about the details of individual releases."
+---
+
+The Release Details page focuses on an individual release. Elements of the release are presented in visual trends for crashes and sessions, specifics regarding each issue, and commit author breakdowns.
+
+![View of the release details page showing an overview graph, files changed, and other details.](release_details.png)
+
+Hover over specific dates and times in the overview graph for a quick perspective of events that crashed, behaved abnormally, errored, or are healthy.
+
+Crashed
```
jan-auer, line 13: Can we move these definitions into the "Sessions" section on the index page? I feel like they are more canonical information and not specific to the details page. For instance, those states will also show up on the release index page, and they contribute to the definition of what a "crash free user" is.

src/docs/product/releases/health/setup.mdx
jan-auer: + JavaScript https://docs.sentry.io/platforms/javascript/configuration/releases/#release-health

jan-auer (commented):
Some of these can be left for a follow-up. Let's record the ones we defer in https://github.com/getsentry/sentry-docs/issues/2869

src/includes/configuration/auto-session-tracking/android.mdx
PeloWriter: Good point. We are discussing two version of the SDK here, 2.1.-, where you need to modify the `AndroidManifest.xml`file, or version 3.0 where it's captured by default unless disabled. 

I think, to make this more clear, we might simply focus on most recent, so remove the first sentence altogether. What do you think?

PeloWriter: This will work, unless we'd like to not mention 2.1.0, and focus only on 3.0. Let me know.

src/docs/product/releases/health/setup.mdx
PeloWriter: I will add them, and change the Release Health information for these SDKs in the common content to `supported`

src/docs/product/error-monitoring/filtering/index.mdx
```diff
@@ -6,9 +6,11 @@ description: "Learn about the different methods for filtering data in your proje
 
 Sentry provides several methods to filter data in your project. Using sentry.io to filter events is a simple method since you don't have to <PlatformLink to="/configuration/filtering/">configure and deploy your SDK to filter projects</PlatformLink>.
 
-### Inbound Data Filters
+## Inbound Data Filters
 
-Inbound data filters allow you to determine which errors, if any, Sentry should ignore. Explore these by navigating to **[Project] » Project Settings » Inbound Filters.**
+Inbound data filters allow you to determine which errors, if any, Sentry should ignore. Explore these by navigating to **[Project] > Project Settings > Inbound Filters.**
+
+These filters are exclusively applied at ingest time and not later in processing. This, for instance, lets you discard an error by error message when the error is ingested through the JSON API. On the other hand, this filter doesn't apply to ingested minidumps.
```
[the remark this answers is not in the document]
PeloWriter: Lovely. Will do.

src/docs/product/error-monitoring/filtering/index.mdx
```diff
@@ -23,3 +25,35 @@ Inbound filters include:
 Once applied, you can track the filtered events (numbers and cause) using the graph provided at the top of the Inbound Data Filters view.
 
 ![](builtin-inbound-filters.png)
+
+## Filtering for Session Events
```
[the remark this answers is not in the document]
PeloWriter: I got the term from the content I was moving in - I will remove it so it is correct :) .

src/docs/product/releases/health/index.mdx
PeloWriter: 👍

src/docs/product/releases/health/index.mdx
PeloWriter: Yes. Thanks.

src/docs/product/error-monitoring/filtering/index.mdx
PeloWriter: I pulled this content from the Release Health page content; I will fix it so it is corrected.

src/includes/configuration/auto-session-tracking/android.mdx
PeloWriter: I've pulled your suggestion below, but we can modify further, as appropriate.

src/docs/product/releases/health/release-details.mdx
PeloWriter: Yes. Will do.

PeloWriter: I will also combine the definitions that overlap. Do we need more for these definitions, perhaps some examples?

src/docs/product/releases/health/index.mdx
PeloWriter: I'll combine, as makes sense. We will need to review these definitions for accuracy.

src/docs/product/releases/health/index.mdx
PeloWriter: I pulled it from other content, but I agree we should avoid specifically-used terms such as monitor and metrics in this content.

src/docs/product/error-monitoring/filtering/index.mdx
PeloWriter: I've removed it from the content, and stored it in Notion once we determine where to publish it.

src/docs/product/releases/health/index.mdx
PeloWriter: Agreed. I've opened https://github.com/getsentry/sentry-docs/issues/2892 to track.

src/docs/product/releases/health/index.mdx
PeloWriter: Done.

PeloWriter:
Still to do - review the definitions on release/health/index.mdx. The format is not ideal, nor is the flow of this content. Plus, we need to confirm the correctness of the defintions.

jan-auer (approved):
LGTM. We could add a page grid to the root page to make it easier to navigate to the use case docs.

src/docs/product/releases/health/index.mdx
PeloWriter: Added a task to the list in #2869.

PR #2894, opened 2021-01-15
fix: wizard doesn't show selector

bruno-garcia:


src/wizard/dotnet/aspnetcore.md
```diff
@@ -43,6 +46,7 @@ let BuildWebHost args =
 
 ASP.NET Core 3.0:
```
bruno-garcia, line 47:
```suggestion
ASP.NET Core 3 or 5:
```

PeloWriter:
We're adding heads because tabs don't work? I'd suggest @EvanPurkhiser also review, as this may impact his work with onboarding.

PR #2899, opened 2021-01-15
Add Spring boot to Supported Performance SDKs

marandaneto:


dashed (approved):

PeloWriter (approved):

src/docs/product/performance/getting-started.mdx
```diff
@@ -18,6 +18,7 @@ Customers on legacy plans must add transaction events to their subscription in o
 ## Supported SDKs
 
 - [Java/Kotlin/JVM](/platforms/java/performance/)
+  - [Spring Boot](/platforms/java/guides/spring-boot/performance/)
```
bruno-garcia, line 21: We've received feedback that the experience from a Spring dev point of view isn't ideal since we assume SB too much:

```suggestion
  - [Spring](/platforms/java/guides/spring/performance/)
  - [Spring Boot](/platforms/java/guides/spring-boot/performance/)
```

bruno-garcia (approved):

src/docs/product/performance/getting-started.mdx
marandaneto: @bruno-garcia but the `java/guides/spring/performance/` is the same as `/platforms/java/performance/` so why would it be useful to be under `Java/Kotlin/JVM` but linking to the same page? I thought it'd make sense only if there's a custom page/setup

bruno-garcia: @marandaneto When you browse into the platform all content can change. The content must change eventually since java, spring and spring boot have different levels of performance support. So event though today it's all the same content, I would continue to link to the specific platforms though

marandaneto: that makes sense but I did click on both links and they all went to the same place and I was trying to find the gotcha for the specific framework and at the end, I found out it was the same page, no distinction, I feel the other users would get lost too.

bruno-garcia: It's true today is the exact same content but the URL is different so links for example, will be different. The menu options are different (for example SB has `Advanced Usage`, Spring doesn't). But also really as we can (and will) change the content soon for these two to highlight their differences, it's best if we link to both.

Thanks I'll push this suggestion and merge the PR.

bruno-garcia (approved):

PR #2901, opened 2021-01-15
fix: Broken link in grouping-and-fingerprints

untitaker:


matejminar (approved):
Thanks for fixing this!

PR #2910, opened 2021-01-18
Bundle name change

tolgaberk:
Latest react-native cli names the bundles as index.android.bundle when building a release. 
I recently faced with a source map issue about this.
I was following the docs but it was inaccurate.


marandaneto:
good catch, thanks.

I've doubled-check here https://github.com/facebook/react-native/blob/1a67dda668c71d961a4bb3b0cdf6aa22c0e5c138/react.gradle#L45

marandaneto (approved):

PR #2911, opened 2021-01-18
fix: Remove references to auto instrumentation in RN and make clear that it is manual

jennmueng:
Quick fix to make it clear that RN instrumentation is manual only and not automatic yet. This is because the current documentation with references to automatic tracing has confused users such as: https://stackoverflow.com/questions/65750585/sentry-react-native-performance-not-recording-anything


bruno-garcia (approved):

bruno-garcia:
@PeloWriter @lobsterkatie we're putting something out to help with confusion, you might have opinions on the best way forward here.

PR #2920, opened 2021-01-19
Add Ruby (and Rails) to the SDK overview for performance

silent1mezzo:


silent1mezzo:
@PeloWriter Rails doesn't have its own performance section but does have performance configuration in the section I linked. I think it's valuable to call out rails specifically but also fine to take it out if it doesn't make sense.

PeloWriter (approved):

k-fish (approved):

PR #2921, opened 2021-01-19
fix: It is on-premises, not on-premise

### LINKED ISSUE #746: Premises - not Premise
So pretty much everywhere the docs use the word **premise** when it should be **premises**.

https://grammarcops.wordpress.com/tag/premise-vs-premises/

The above article is a perfect example of this error.

More examples: https://www.google.com/search?q=premise+vs+premises

BYK:
Fixes #746


PeloWriter (approved):
Thank you!

PR #2927, opened 2021-01-20
Refactor header to be one piece

cameronmcefee:
This refactors the header to be a single piece, now that we've removed the split layout. This will make another PR much easier to build.


cameronmcefee:
Never mind. This broken test is because jest isn't configured to handle some of the things I don't have time to set all that up presently.

PR #2932, opened 2021-01-21
Update wording and grammar

silent1mezzo:
I also removed the alert on deleting queries. I feel like this is pretty common knowledge and included the text in the paragraph.

### DIFF
--- a/src/docs/product/discover-queries/index.mdx
+++ b/src/docs/product/discover-queries/index.mdx
@@ -72,7 +72,7 @@ To rename a saved query, click on the pencil icon by the header and enter the de
 
 ### Share Queries
 
-Share your queries as often as you want. You can share URLs with other users who also have access to the same organization. As each part of the query is built, the results update and the URL is updated so that in-progress searches can be shared in an email, chat, etc.
+Share your queries as often as you want. You can share URLs with other users who also have access to the same organization. As each part of the query is built, the results update and the URL is updated so that in-progress searches can be shared in an email, chat, and so forth.
 
 ### Delete Queries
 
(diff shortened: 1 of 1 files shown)


src/docs/product/discover-queries/index.mdx
```diff
@@ -72,19 +72,17 @@ To rename a saved query, click on the pencil icon by the header and enter the de
 
 ### Share Queries
 
-Share your queries as often as you want. You can share URLs with other users who also have access to the same organization. As each part of the query is built, the results update, and the URL is updated so that in-progress searches can be shared in email, chat, etc.
+Share your queries as often as you want. You can share URLs with other users who also have access to the same organization. As each part of the query is built, the results update and the URL is updated so that in-progress searches can be shared in an email, chat, etc.
```
PeloWriter, line 75: I see this was in the original - but we don't use Latin abbreviations. Correcting now :) 
```suggestion
Share your queries as often as you want. You can share URLs with other users who also have access to the same organization. As each part of the query is built, the results update and the URL is updated so that in-progress searches can be shared in an email, chat, and so forth.
```

src/docs/product/discover-queries/index.mdx
```diff
@@ -72,19 +72,17 @@ To rename a saved query, click on the pencil icon by the header and enter the de
...
 
 ### Delete Queries
 
-<Alert>Deleting a saved query is irreversible.</Alert>
-
-On the Discover Homepage, each saved query card has an ellipsis that will open a context menu. Delete the query from here. You can also delete the query within Query Results view by clicking the trash can in the upper right.
+On the Discover Homepage, each saved query card has an ellipsis that will open a context menu. From here, you can delete the query. This action is irreversible. You can also delete the query within the Query Results view by clicking the trash can in the upper right.
```
PeloWriter, line 79:
```suggestion
On the Discover homepage, each saved query card has an ellipsis that will open a context menu. From here, you can delete the query. This action is irreversible. You can also delete the query within the Query Results view by clicking the trash can in the upper right.
```

PeloWriter (approved):
Thanks.

PR #2941, opened 2021-01-22
fix(ticket rules): Indicate feature availability by plan

ceorourke:
Automatic issue creation is only available on the new Business plans, so we want to make that very clear in the documentation.

Screen Shot 2021-01-22 at 3 48 44 PM


mgaeta (approved):

PeloWriter (approved):
Thank you.

PR #2943, opened 2021-01-22
add missing dotnet options

bruno-garcia:
During the docs migration it seems like dotnet's options went missing


PeloWriter (approved):
Thank you.

PR #2948, opened 2021-01-25
feat: Remove autoSession from Wizard

HazAT:
Since version `6`, we have it enabled by default and therefore we don't need this anymore.


PeloWriter:
Should we also remove it from the configuration code sample for JavaScript in `includes/configuration/auto-session-tracking`?

PeloWriter (approved):
Thank you.

PR #2951, opened 2021-01-25
fix(wizard): logback dsn todo

bruno-garcia:


src/wizard/java/logback.md
```diff
@@ -24,7 +24,7 @@ Install Sentry's integration with Logback using either Maven or Gradle:
 ### Gradle
 
 ```groovy
-implementation 'io.sentry:sentry-logback:3.2.0'
+implementation 'io.sentry:sentry-logback:3.2.1'
```
marandaneto, line 27: we went away from the placeholder {version} and now the whole Java SDK deps have different versions, little of a mess.
Search for: `implementation 'io.sentry`

marandaneto (approved):

src/wizard/java/logback.md
bruno-garcia: the solution is using `craft` though, not making the user figure out the right version

marandaneto: agreed, but now there's a pinned version in the docs that is not up to date and every page has a different version, not sure which one is worse.

PR #2960, opened 2021-01-28
fix(wizard): install gradle

bruno-garcia:


src/wizard/java/spring-boot.md
```diff
@@ -21,7 +21,7 @@ Install using either Maven or Gradle:
 </dependency>
 ```
 
-### Groovy
+### Gradle
 
 ```groovy {tabTitle:Gradle}
```
marandaneto, line 26: maybe this should change then? to `Groovy`

marandaneto (approved):

src/wizard/java/spring-boot.md
bruno-garcia: the programming language is groovy, so for syntax highlighting it's correct. But as a heading, we want to indicate this line is a "gradle installation" as opposed to Maven above (which is in XML form)

PR #2971, opened 2021-01-29
Fix: missing ' on Android dep

marandaneto:


bruno-garcia:
I reviewed it 3x :)

Thanks

marandaneto:
> I reviewed it 3x :)
> 
> Thanks

but we went GA for v4 so I did the last pass :P

PR #2974, opened 2021-01-29
Polish Java Performance docs.

maciejwalkowiak:
The goal of this PR is to make docs less confusing and easier to follow.

- traces are not captured **automatically** with manual integration
- on the top of Java performance I added links to Spring and Spring Boot guides
- updated sampling function example in Spring & Spring Boot to have already a request taken out from the context.

cc @bruno-garcia

### DIFF
--- a/src/includes/java/performance/getting-started/java.mdx
+++ b/src/includes/java/performance/getting-started/java.mdx
@@ -4,6 +4,17 @@ Performance Monitoring is available for the Sentry Java SDK version ≥ 4.0.
 
 Sentry allows you to monitor the performance of your application, showing you how latency in one service may affect another service, and letting you pinpoint exactly which parts of a given operation may be responsible. To do this, it captures distributed traces consisting of transactions and spans, which measure individual services and individual operations within those services, respectively. You can learn more about this model in our [Distributed Tracing](/product/performance/distributed-tracing/) docs.
 
+<Note>
+
+Sentry Java SDK for Performance comes with dedicated integrations for **Spring** and **Spring Boot** that drastically simplify using the Performance feature.
+
+If you are using any of these, make sure to follow their corresponding guides.
+
+* <PlatformLink to="/spring-boot/performance/">Spring Boot</PlatformLink>
+* <PlatformLink to="/spring/performance/">Spring</PlatformLink>
+
+</Note>
+
 ## Enabling Tracing
 
 <PlatformContent includePath="performance/enable-tracing" />
--- a/src/includes/performance/enable-tracing/java.spring-boot.mdx
+++ b/src/includes/performance/enable-tracing/java.spring-boot.mdx
@@ -19,7 +19,7 @@ sentry.traces-sample-rate=0.3
 
 Or through providing a bean of type `SentryOptions#TracesSamplerCallback`:
 
-```java
+```java {tabTitle:Java}
 import io.sentry.SentryOptions.TracesSamplerCallback;
 import io.sentry.SamplingContext;
 import org.springframework.stereotype.Component;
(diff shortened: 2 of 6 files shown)


maciejwalkowiak:
SDK docs assume that users are already familiar with the Sentry Performance feature - which I believe can be false assumption. I believe we should have either an intro to Sentry performance feature with screenshots (as images say more than text), or a videos presenting it. Right now I don't find it very easy to understand capabilities of the Performance feature in Sentry just by looking at the docs - that's independent from the SDK.

marandaneto:
> SDK docs assume that users are already familiar with the Sentry Performance feature - which I believe can be false assumption. I believe we should have either an intro to Sentry performance feature with screenshots (as images say more than text), or a videos presenting it. Right now I don't find it very easy to understand capabilities of the Performance feature in Sentry just by looking at the docs - that's independent from the SDK.

what's about the performance's product docs?
https://docs.sentry.io/product/performance/

marandaneto:
> The goal of this PR is to make docs less confusing and easier to follow.
> 
> * traces are not captured **automatically** with manual integration
> * on the top of Java performance I added links to Spring and Spring Boot guides
> * updated sampling function example in Spring & Spring Boot to have already a request taken out from the context.
> 
> cc @bruno-garcia

this will probably conflict quite a bit with https://github.com/getsentry/sentry-docs/pull/2970

src/includes/java/performance/getting-started/java.mdx
```diff
@@ -4,6 +4,17 @@ Performance Monitoring is available for the Sentry Java SDK version ≥ 4.0.
 
 Sentry allows you to monitor the performance of your application, showing you how latency in one service may affect another service, and letting you pinpoint exactly which parts of a given operation may be responsible. To do this, it captures distributed traces consisting of transactions and spans, which measure individual services and individual operations within those services, respectively. You can learn more about this model in our [Distributed Tracing](/product/performance/distributed-tracing/) docs.
 
+<Note>
```
marandaneto, line 7: this should be under either a `includes` or `supported` feature flag, we don't want this visible for Android docs.
https://github.com/getsentry/sentry-docs/pull/2970 probably solves this by separating things

maciejwalkowiak: So lets wait till #2970 is merged.

src/includes/performance/enable-tracing/java.spring-boot.mdx
```diff
@@ -28,6 +28,21 @@ import org.springframework.stereotype.Component;
...
+
+```kotlin {tabTitle:Kotlin}
+import io.sentry.SentryOptions.TracesSamplerCallback
+import io.sentry.SamplingContext
+import org.springframework.stereotype.Component
+
+@Component
+class CustomTracesSamplerCallback : TracesSamplerCallback {
+    override fun sample(context: SamplingContext): Double {
+        val request = context.customSamplingContext!!["request"] as HttpServletRequest
```
bruno-garcia, line 45: Is `!!` the way to go here?
Maybe we document with `?` instead to avoid having potential bugs on these snippets?

Stuff like this can get copy pasted in contexts where the request isn't available

src/includes/performance/enable-tracing/java.spring.mdx
```diff
@@ -61,6 +62,7 @@ import org.springframework.stereotype.Component
 @Component
 class CustomTracesSamplerCallback : TracesSamplerCallback {
     override fun sample(context: SamplingContext): Double {
+        val request = context.customSamplingContext!!["request"] as HttpServletRequest
```
marandaneto, line 65: same as https://github.com/getsentry/sentry-docs/pull/2974/files#r567023846

maciejwalkowiak:
Replaced by https://github.com/getsentry/sentry-docs/pull/3045

PR #2983, opened 2021-02-01
feat(apple): Enable sendDefaultPii

philipphofmann:
We added this for Apple recently, see
https://github.com/getsentry/sentry-cocoa/pull/923.

Only merge when sentry-cocoa 6.1.5 is released.

### DIFF
--- a/src/platforms/common/configuration/options.mdx
+++ b/src/platforms/common/configuration/options.mdx
@@ -113,7 +113,7 @@ Grouping in Sentry is different for events with stack traces and without. As a r
 
 </ConfigKey>
 
-<ConfigKey name="send-default-pii" supported={["dotnet", "javascript", "node", "react-native", "python", "java", "php", "rust"]}>
+<ConfigKey name="send-default-pii" supported={["apple", "dotnet", "javascript", "node", "react-native", "python", "java", "php", "rust"]}>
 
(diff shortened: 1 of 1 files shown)


src/platforms/common/configuration/options.mdx
```diff
@@ -113,7 +113,7 @@ Grouping in Sentry is different for events with stack traces and without. As a r
 
 </ConfigKey>
 
-<ConfigKey name="send-default-pii" supported={["dotnet", "javascript", "node", "react-native", "python", "java", "php", "rust"]}>
+<ConfigKey name="send-default-pii" supported={["apple", "dotnet", "javascript", "node", "react-native", "python", "java", "php", "rust"]}>
```
marandaneto, line 116: not related to this PR but:
we'd need to go thru all the options (options.mdx) and check 1 by 1 for each mobile platform, I suspect a few of them are missing, I don't recall adding new options here often but I do recall adding on code :P

marandaneto (approved):

src/platforms/common/configuration/options.mdx
marandaneto: https://github.com/getsentry/sentry-docs/issues/2985 so we don't forget

Issue #2685, opened 2020-11-13
Content partially hidden when navigating to an anchor


rhcarvalho:
This is most likely a CSS issue.

Current behavior when going to https://docs.sentry.io/platforms/go/#configure:

![image](https://user-images.githubusercontent.com/88819/99060523-73a32980-25a0-11eb-82f8-9ec6e6659e29.png)

---

What it should be instead:

![image](https://user-images.githubusercontent.com/88819/99060563-80278200-25a0-11eb-9842-eb0fd576d7cd.png)

PeloWriter:
This issue still exists.

PR #2987, opened 2021-02-01
Fix: push anchor headers below fixed top bar

rhcarvalho:
Based on https://css-tricks.com/hash-tag-links-padding/

Fixes https://github.com/getsentry/sentry-docs/issues/2685

### DIFF
--- a/src/css/_includes/global-header.scss
+++ b/src/css/_includes/global-header.scss
@@ -42,3 +42,31 @@
     }
   }
 }
+
+/*
+
+  Push anchor headers below fixed top bar.
+  Fixes https://github.com/getsentry/sentry-docs/issues/2685.
+
+*/
+h1,
+h2,
+h3,
+h4,
+h5,
+h6,
+.h1,
+.h2,
+.h3,
+.h4,
+.h5,
+.h6 {
+  &::before {
+    display: block;
+    content: " ";
+    margin-top: -($headerHeight + $spacer);
+    height: ($headerHeight + $spacer);
+    visibility: hidden;
+    pointer-events: none;
+  }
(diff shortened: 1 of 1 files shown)


rhcarvalho:
Works for me!

Example https://sentry-docs-git-obscured-anchors.sentry.dev/platforms/go/#configure

image

rhcarvalho:
Tested in current versions of Chrome, Firefox and Safari for macOS.

kamilogorek:
Yesss. Would prefer to have a look from @matejminar @priscilawebdev too though.

src/css/_includes/global-header.scss
```diff
@@ -42,3 +42,31 @@
...
+.h5,
+.h6 {
+  &::before {
+    display: block;
+    content: " ";
+    margin-top: -($headerHeight + $spacer);
+    height: ($headerHeight + $spacer);
+    visibility: hidden;
+    pointer-events: none;
+  }
```
matejminar, line 71: How about this?

```suggestion
  scroll-margin-top: $headerHeight + $spacer;
```

matejminar: Let me check how that works on safari, there's supposed to be a problem:
https://caniuse.com/?search=scroll-margin-top

matejminar: _Sigh_, yep, Safari is not happy with that one - we can go with your solution :)

matejminar (approved):

rhcarvalho:
This had a bug, the anchor link next to the header got shifted:

image

Going to revert and have a second take on it later.

PR #2991, opened 2021-02-01
fix(tls): Bump TLSv1/1.1 sunset to March.

JTCunning:
It is what it is. Giving some users more of a sunset period than initially required.

### DIFF
--- a/src/docs/product/security/ssl.mdx
+++ b/src/docs/product/security/ssl.mdx
@@ -15,7 +15,7 @@ apply to Self-hosted or Single Tenant.
 
 Sentry's API, Dashboard, and Event Submission requires a minimum TLS version of 1.0 to communicate securely.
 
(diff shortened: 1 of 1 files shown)


src/docs/product/security/ssl.mdx
```diff
@@ -15,7 +15,7 @@ apply to Self-hosted or Single Tenant.
 
 Sentry's API, Dashboard, and Event Submission requires a minimum TLS version of 1.0 to communicate securely.
 
-By the end of January, 2021, Sentry's API and Dashboard will require a minimum TLS version of 1.2.
+By the end of February, 2021, Sentry's API and Dashboard will require a minimum TLS version of 1.2.
```
PeloWriter, line 18: Do we want to simply say, "By March 1, 2021"? I'm ok with either :) .

PeloWriter (approved):

PR #2992, opened 2021-02-02
Update Android migration page SDK version

marandaneto:


src/platforms/android/migration.mdx
```diff
@@ -5,7 +5,7 @@ redirect_from:
   - /platforms/android/migrate/
 ---
 
-## Migrating from `sentry-android 2.x` to `sentry-android 3.x`
+## Migrating from `sentry-android 2.x` to `sentry-android 4.x`
```
philipphofmann, line 8: `m`: I would rather keep this and state that there is no migration needed between 3.x to 4.x.

bruno-garcia: Agreed. Could even say that: The performance API was added but API-wise no breaking changes exist.

Also, I find it hard to believe there are 0 breaking changes :D

bruno-garcia: Oh it's Android specific. I believe it now

marandaneto: will just close it then

PR #2993, opened 2021-02-02
Fix: push anchor headers below fixed top bar

### LINKED ISSUE #2685: Content partially hidden when navigating to an anchor

rhcarvalho:
Based on https://css-tricks.com/hash-tag-links-padding/

Change also anchor link to be positioned relative to the associated
header, so that both anchor and header are pushed down equally, visible
right below the top bar.

See also #2987 & #2990.

Fixes #2685.

### DIFF
--- a/src/css/_includes/global-header.scss
+++ b/src/css/_includes/global-header.scss
@@ -42,3 +42,31 @@
     }
   }
 }
+
+/*
+
+  Push anchor headers below fixed top bar.
+  Fixes https://github.com/getsentry/sentry-docs/issues/2685.
+
+*/
+h1,
+h2,
+h3,
+h4,
+h5,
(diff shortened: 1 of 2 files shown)


rhcarvalho (commented):
The current implementation breaks links that are above headers, either in a previous section or even the breadcrumb navigation:

image

image

PeloWriter:
Thanks for this fix, Rodolfo, since the previous anchor caused the title or anchor spot to disappear from view...I've tested this on a few pages and it works great.

PeloWriter (approved):

PR #3003, opened 2021-02-02
Sampling Inheritance: Propagation

bruno-garcia:
I'd like to propose to use the wording:

Trace propagation as opposed to inheritance.


lobsterkatie:
I disagree. Propagation is the "push" - the parent propagates its sampling decision _to_ its children through headers. Here, though, we're coming from the perspective of the child, which inherits its decision _from_ its parent, and therefore need to be on the "pull" end of things.

bruno-garcia:
I continue to see this as propagating the decision. Since this is wording I've seen in things like W3C, I suggested to bring this here but not willing to die on this hill.

PR #3012, opened 2021-02-03
Update dsn-explainer.mdx

nueces:
Update the DSN specification and the javascript example that were incorrect.


HazAT:
Hey, we don’t have the secret part anymore. Also the Host / Path part is there on purpose because most URL split libraries return these vs. URI. To make it more obvious.

Thanks for your effort but I think we should close this PR.

PR #3013, opened 2021-02-03
Remove redundant option onUnmount

akanshgulati:
Remove the same option mentioned twice in order to avoid any confusion.


rhcarvalho (approved):

rhcarvalho:
Thanks!

PR #3015, opened 2021-02-03
Adding organisation slug to relay state URL

rbisol:
Correcting the relay state URL. The screenshot still needs to be updated, unfortunately, I do not have access to an Azure account to get an updated one.


PeloWriter (approved):
This looks correct to me, based on the others in the list. Thank you!

PR #3018, opened 2021-02-03
upgrade(gatsby): Upgrades gatsby & gatsby-cli to latest stable

BYK:
This is mostly to upgrade `node-forge` which has a security vulnerability in earlier versions. That said the newest version of Gatsby brings in many bug fixes and some performance upgrades, especially for development.


jan-auer (approved):
Given that this still builds 👍

PR #3028, opened 2021-02-04
Use safe navigation operator in ruby example config

elucid:
NOTE: if you don't do this, it will bomb when trying to filter sensitive data in contexts where this is not set (e.g. cron scripts running in your app env, active job processing)

### DIFF
--- a/src/includes/configuration/before-send/ruby.mdx
+++ b/src/includes/configuration/before-send/ruby.mdx
@@ -4,7 +4,12 @@ Sentry.init do |config|
   config.before_send = lambda do |event, hint|
     # note1: if you have config.async configured, the event here will be a Hash instead of an Event object
     # note2: the code below is just an example, you should adjust the logic based on your needs
-    event&.request&.data = filter.filter(event.request.data)
+    request = event.request
+
+    if request
+      request.data = filter.filter(request.data)
+    end
+
     event
   end
 end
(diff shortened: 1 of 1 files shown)


src/includes/configuration/before-send/ruby.mdx
```diff
@@ -4,7 +4,7 @@ Sentry.init do |config|
   config.before_send = lambda do |event, hint|
     # note1: if you have config.async configured, the event here will be a Hash instead of an Event object
     # note2: the code below is just an example, you should adjust the logic based on your needs
-    event.request.data = filter.filter(event.request.data)
+    event&.request&.data = filter.filter(event.request.data)
```
st0012, line 7: I don't think the event would ever be `nil` 🤔 

also, I think we can be more explicit here

```ruby
if request = event.request
  request.data = filter.filter(request.data)
end

event
```

elucid: Yeah that's true, event should always be there. Pushed a commit almost identical do this but didn't put the assignment in the conditional.

st0012 (approved):

elucid:
I'm not sure what else is required here. I can see that the PR was approved by @st0012 several weeks ago. If there are further changes required, please let me know.

st0012:
@HazAT can you give it a look as well?

elucid:
Hey there, same comment as 3 weeks ago. Not sure what else, if anything, is required. PR was approved by @st0012 about 6 weeks ago. Please let me know if you want further changes.

PeloWriter:
Thanks for your patience, @elucid. I will follow up with the team.

PeloWriter:
I've checked with the team, and this is good to merge. Thanks again for your patience.

PR #3033, opened 2021-02-05
Update performance getting-started with Go related links

odilonjk:
The goal of this PR is to update this doc with all Go related links.


rhcarvalho (approved):
Thanks!

PR #3039, opened 2021-02-05
Fix a typo in .NET docs

Tyrrrz:
Related: https://github.com/getsentry/sentry-dotnet/issues/799

### DIFF
--- a/src/includes/performance/traces-sampler-as-sampler/dotnet.mdx
+++ b/src/includes/performance/traces-sampler-as-sampler/dotnet.mdx
@@ -13,7 +13,7 @@ SentrySdk.Init(options =>
     {
         var ctx = context.CustomSamplingContext;
 
-        return ctx["url"] switch
+        return ctx.GetValueOrDefault("url") switch
         {
             // These are important - take a big sample
             "/payment" => 0.5,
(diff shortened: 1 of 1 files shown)


src/includes/performance/traces-sampler-as-sampler/dotnet.mdx
```diff
@@ -13,7 +13,7 @@ SentrySdk.Init(options =>
     {
         var ctx = context.CustomSamplingContext;
 
-        return ctx switch
+        return ctx["url"] switch
```
bruno-garcia, line 16: wouldn't this throw though if `url` isn't there? Not sure this is good sample code

Tyrrrz: I just adapted the sample from here: https://github.com/getsentry/sentry-docs/blob/caf5692e6dcd804b33c8832125269b3f9d1e457c/src/includes/performance/traces-sampler-as-sampler/java.mdx

bruno-garcia: Doesn't need to look the same, has to make sense here.
People will 100% copy paste this code into their apps and it'll crash when there's no `url` in the context object so we need to be mindful of our samples

Tyrrrz: In that case...

```suggestion
        return ctx.GetValueOrDefault("url") switch
```

bruno-garcia (approved):
THanks!

PR #3044, opened 2021-02-06
Fix @sentry/cli version inconsistency in hermes.mdx

tgevaert:
Fix inconsistency between incorrect link text version (1.5.1) and link target version (1.51.1)


marandaneto (approved):
thanks @tgevaert

PR #3052, opened 2021-02-09
feat(aws-lambda): adds manual layer instructions

scefali:
This PR adds setup instructions for AWS Lambda with Node using a layer instead of installing Sentry. New functionality has been added with a dropdown that lets the user select their region and we generate the ARN for them to use during setup:

Screen Shot 2021-02-08 at 4 05 01 PM

The information needed to render this dropdown is dynamically loaded from a new endpoint in the release registry which contains info on the latest Lambda layer version.


.github/workflows/test.yml
```diff
@@ -29,6 +29,8 @@ jobs:
             .cache
             public
           key: ${{ runner.os }}-${{ hashfiles('**/yarn.lock') }}-gatsby-build
+      - name: yarn clean
+        run: yarn clean
```
scefali, line 33: I had to add this step in the pipeline because Gatsby caching is crap 🙃

src/gatsby/utils/awsLambdaLayerRegistry.ts
```diff
@@ -0,0 +1,39 @@
...
+  getLayerMap = async () => {
+    if (!this.indexCache) {
+      try {
+        const result = await axios({
+          url: `${BASE_REGISTRY_URL}/aws-lambda-layers`,
+        });
+        this.indexCache = result.data;
+      } catch (err) {
+        console.error(
+          `Unable to fetch index for package registry: ${err.message}`
```
dcramer, line 32: should tweak this msg

dcramer (approved):

src/platforms/node/guides/aws-lambda/layer/index.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+```javascript
+const Sentry = require("@sentry/serverless");
+```
+
+<Alert level="warning">
+
+Note, this approach will not work if you use Webpack.
+
+</Alert>
```
PeloWriter, line 16: Let's clarify this -because it's either a Note or a Warning. Generally, we're moving away from colorful boxes that are more distracting than useful. In addition, I'm not quite clear on what we'd like to say - are we telling people not to use this approach if they're also using Webpack? In that case, perhaps:
```suggestion
<Note>

This installation method will not work if you are using Webpack.

</Note>
```

src/platforms/node/guides/aws-lambda/layer/index.mdx
```diff
@@ -0,0 +1,28 @@
+---
+title: AWS Lambda Layer (Node)
+description: "Learn how to set up Sentry with a Layer"
+---
+
+An alternative way of installing Sentry is using a [Lambda Layer](https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html). This happens instead of adding `@sentry/serverless` with NPM and Yarn. Your can still import Sentry as usual:
```
PeloWriter, line 6:
```suggestion
You can also install Sentry using a [Lambda Layer](https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html) instead of adding `@sentry/serverless` with `npm` or `yarn`. Import Sentry as usual:
```

src/platforms/node/guides/aws-lambda/layer/index.mdx
```diff
@@ -0,0 +1,28 @@
...
+const Sentry = require("@sentry/serverless");
+```
+
+<Alert level="warning">
+
+Note, this approach will not work if you use Webpack.
+
+</Alert>
+
+To add the Sentry Layer, go to your Lambda function, press **Layers**, then **Add a Layer**.
```
PeloWriter, line 18:
```suggestion
Then add the Sentry Layer by navigating to your Lambda function. Select **Layers**, then **Add a Layer**.
```

src/platforms/node/guides/aws-lambda/layer/index.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+   ![](lambda_view.png)
+
+Then go to the **Specify an ARN** tab as shown:
```
PeloWriter, line 22:
```suggestion
 **Specify an ARN** tab as illustrated:
```

src/platforms/node/guides/aws-lambda/layer/index.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+   ![](add_layer.png)
+   
+Then below, set the region and copy the ARN value shown below into the input.
```
PeloWriter, line 26:
```suggestion
Finally, set the region and copy the provided ARN value into the input.
```

PeloWriter (approved):

PR #3059, opened 2021-02-09
fix: Remove performance docs from cordova

### LINKED ISSUE #3058: Cordova does not support the performance feature
url: https://docs.sentry.io/platforms/javascript/guides/cordova/performance/

we should make Cordova `notSupported`

jennmueng:
Fixes #3058 as Cordova does not support performance


marandaneto:
@jennmueng CI is unhappy with bad links, you probably need to add this flag elsewhere too

HazAT (approved):

PR #3076, opened 2021-02-11
error-tracing: use the defined log_format

alxu-sentry:
access_json was defined but not actually used.


dashed:
@alxu-sentry Feel free to merge this as is if you're confident of this change.

alxu-sentry:
> @alxu-sentry Feel free to merge this as is if you're confident of this change.

weeellll I'm not 100% sure it's correct, but it's definitely _better_. it's clearly not intended as an exact copy-paste into your config anyways, just an example.

PR #3077, opened 2021-02-11
Fix tutorial sidebar order

alxu-sentry:
Otherwise they are sorted by title, which is wrong.


alxu-sentry:
I searched for `## Next` in index.mdx files and didn't find any other instances to be corrected.

PeloWriter (approved):
Nice fix (simple, but impactful). Thank you

PR #3085, opened 2021-02-12
RqIntegration documentation: add tip to call sentry_sdk.init once in (Flask) app

ndkv:
Extend docs with tips from https://github.com/getsentry/sentry-python/issues/935 and https://github.com/getsentry/sentry-python/issues/907 to initialise Sentry once per (Flask) app and import it in `rq worker`'s settings file. 

Calling `sentry_sdk.init` in the worker's configuration per the  current docs (https://docs.sentry.io/platforms/python/guides/rq/) results in undelivered messages and exceptions.

### DIFF
--- a/src/platforms/python/guides/rq/index.mdx
+++ b/src/platforms/python/guides/rq/index.mdx
@@ -31,6 +31,24 @@ The integration will automatically report errors from all RQ jobs.
 
 Generally, make sure that the **call to `init` is loaded on worker startup**, and not only in the module where your jobs are defined. Otherwise, the initialization happens too late and events might end up not being reported.
 
+In addition, make sure that **`init` is called once** in your app and imported from there in the different places it is needed. In a Flask app, for example, Sentry and `RqIntegration` can be initalised as 
+
+```python {filename:app.py}
+import sentry_sdk
+from sentry_sdk.integrations.flask import FlaskIntegration
+from sentry_sdk.integrations.rq import RqIntegration
+
+sentry = sentry_sdk.init(
+    dsn=PUBLIC_DSN,
+    integrations=[FlaskIntegration(), RqIntegration()])
+```
+
+`mysettings.py` then becomes
+
+```python {filename:mysettings.py}
+from app import sentry
+```
+
 ## The `--sentry-dsn` CLI option
 
 Passing `--sentry-dsn=""` to RQ forcibly disables [RQ's shortcut for using Sentry](https://python-rq.org/patterns/sentry/). For RQ versions before 1.0 this is necessary to avoid conflicts, because back then RQ would attempt to use the `raven` package instead of this SDK. Since RQ 1.0 it's possible to use this CLI option and the associated RQ settings for initializing the SDK.
--- a/src/platforms/python/guides/rq/index.mdx
+++ b/src/platforms/python/guides/rq/index.mdx
@@ -31,7 +31,7 @@ The integration will automatically report errors from all RQ jobs.
 
 Generally, make sure that the **call to `init` is loaded on worker startup**, and not only in the module where your jobs are defined. Otherwise, the initialization happens too late and events might end up not being reported.
 
-In addition, make sure that **`init` is called once** in your app and imported from there in the different places it is needed. In a Flask app, for example, Sentry and `RqIntegration` can be initalized as 
+In addition, make sure that **`init` is called only once** in your app. For example, if you have a Flask app and a worker that depends on the app, we recommend initializing Sentry with a single configuration that is suitable for Flask and RQ, as in:
 
 ```python {filename:app.py}
 import sentry_sdk
@@ -43,10 +43,10 @@ sentry_sdk.init(
     integrations=[FlaskIntegration(), RqIntegration()])
 ```
 
-`mysettings.py` then becomes
+The worker configuration `mysettings.py` then becomes:
 
 ```python {filename:mysettings.py}
-# import initialised Sentry SDK from app.py
+# This import causes the Sentry SDK to be initialized
 import app
 ```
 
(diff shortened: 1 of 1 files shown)


src/platforms/python/guides/rq/index.mdx
```diff
@@ -31,6 +31,24 @@ The integration will automatically report errors from all RQ jobs.
...
 
+In addition, make sure that **`init` is called once** in your app and imported from there in the different places it is needed. In a Flask app, for example, Sentry and `RqIntegration` can be initalised as 
+
+```python {filename:app.py}
+import sentry_sdk
+from sentry_sdk.integrations.flask import FlaskIntegration
+from sentry_sdk.integrations.rq import RqIntegration
+
+sentry = sentry_sdk.init(
+    dsn=PUBLIC_DSN,
```
rhcarvalho, line 42:
```suggestion
    dsn=___PUBLIC_DSN___,
```

Need the `___` to get it magically replaced with a valid DSN when reading docs.sentry.io.

src/platforms/python/guides/rq/index.mdx
```diff
@@ -31,6 +31,24 @@ The integration will automatically report errors from all RQ jobs.
 
 Generally, make sure that the **call to `init` is loaded on worker startup**, and not only in the module where your jobs are defined. Otherwise, the initialization happens too late and events might end up not being reported.
 
+In addition, make sure that **`init` is called once** in your app and imported from there in the different places it is needed. In a Flask app, for example, Sentry and `RqIntegration` can be initalised as
```
rhcarvalho, line 34:
```suggestion
In addition, make sure that **`init` is called once** in your app and imported from there in the different places it is needed. In a Flask app, for example, Sentry and `RqIntegration` can be initalized as 
```

We use American English spelling in docs.

src/platforms/python/guides/rq/index.mdx
```diff
@@ -31,6 +31,24 @@ The integration will automatically report errors from all RQ jobs.
...
+
+sentry = sentry_sdk.init(
+    dsn=PUBLIC_DSN,
+    integrations=[FlaskIntegration(), RqIntegration()])
+```
+
+`mysettings.py` then becomes
+
+```python {filename:mysettings.py}
+from app import sentry
```
rhcarvalho, line 49: @untitaker could you please have a look? I'm afraid this doesn't match what you had in mind in the referred comments.

I'd expect an import just for the side effect of initializing the SDK, as one shouldn't need to use the return value of `sentry_sdk.init`.

rhcarvalho:
@ndkv thanks for the contribution!

src/platforms/python/guides/rq/index.mdx
untitaker: that's correct. return value of `init()` is not useful, and so this import can read `import app`

`from app import sentry` communicates intent better but that's nothing you can't fix with a code comment

src/platforms/python/guides/rq/index.mdx
```diff
@@ -31,6 +31,24 @@ The integration will automatically report errors from all RQ jobs.
...
+In addition, make sure that **`init` is called once** in your app and imported from there in the different places it is needed. In a Flask app, for example, Sentry and `RqIntegration` can be initalized as 
+
+```python {filename:app.py}
+import sentry_sdk
+from sentry_sdk.integrations.flask import FlaskIntegration
+from sentry_sdk.integrations.rq import RqIntegration
+
+sentry = sentry_sdk.init(
+    dsn=___PUBLIC_DSN___,
+    integrations=[FlaskIntegration(), RqIntegration()])
```
rhcarvalho, line 43: Hmm, in this case, is the `RqIntegration` used in the Flask app too?

If the Flask app and Rq job have different integration needs, it might be a better suggestion to have independent calls to `sentry_sdk.init` -- in https://github.com/getsentry/sentry-python/issues/907 the problem was that the worker never initialized the Sentry SDK.

ndkv: Good point. 

I tried initialising it in my worker but the messages and exceptions were not sent _unless_ I added a breakpoint in the queued function and stepped through it line by line. My hunch was that something was off with the initialisation (order) and https://github.com/getsentry/sentry-python/issues/935 seems to confirm that. Hence my decision to initialise it in Flask.

src/platforms/python/guides/rq/index.mdx
ndkv: Alright, I `import app` now and have added a comment.

src/platforms/python/guides/rq/index.mdx
ndkv: I can also open an issue about this if you prefer.

rhcarvalho: I see, so does your worker import the Flask app?

The advice there makes sense in that case. In this short example I'm afraid it could be misleading for other people.

ndkv: Yes, it does (although I set `sentry = sentry_sdk.init()` and import it in my worker i.e. I use it as I'd initially documented it).

I'll leave it to you to decide whether to include this or not. It took me a while to get it to work and wanted to save a fellow dev's time. :) 

The solution presents itself fairly quickly if one browses `sentry-python` repo's closed issues but I started from `rq`'s docs, searched Sentry's forum and only then looked in the issue tracker.

untitaker: Please do not ever ask people to have two separate init calls. It is extremely confusing if for whatever reason both modules get imported and one init call overrides the other.

rhcarvalho (approved):
@ndkv thanks again for the contribution, I'm getting this merged.

PR #3097, opened 2021-02-15
ref: Rename "GCP Functions" to "Google Cloud Functions"

tonyo:
The official [product name](https://cloud.google.com/functions/) is "Cloud Functions", "GCP Functions" is not really used anywhere.

A small remaining inconsistency is that the Python integration is actually called [`GcpIntegration`](https://github.com/getsentry/sentry-python/blob/a7ef7c05df6669593b168581c9e5d616cb0a1af5/sentry_sdk/integrations/gcp.py#L119), but I'm not changing that here.


rhcarvalho (approved):
Thanks @tonyo 

Optional if you consider: update the URL and add a redirect

dashed (approved):

Issue #1249, opened 2019-10-02
Integrations from CDN, SRI for integrations


jrencz:
I see it's possible to get integrations from the CDN (e.g. https://browser.sentry-cdn.com/5.6.3/captureconsole.min.js) but it's only mentioned in package readme (introduced in https://github.com/getsentry/sentry-javascript/commit/0be23feb306f03f9343128bdb9a4e5e021bb4a11#diff-fb88e65718bc5719a6358b0c18fe2de9R26)

It should be in docks as well, including SRI checksums as there are for bundle (#1134)

jrencz:
Can I ask for this to be triaged?
I'm leaving a comment to prevent issue from being aito-closed

HazAT:
Hey,
Thanks for reporting this, we should def. add this to our docs on this page
https://docs.sentry.io/platforms/javascript/configuration/integrations/plugin/

To switch also always show how to install it from the CDN.

PR #3098, opened 2021-02-15
feat: Add CDN integrations to JS pluggable integrations

iker-barriocanal:
Resolves https://github.com/getsentry/sentry-docs/issues/1249.


HazAT (changes requested):
Looks good, can we also add the code samples to the javascript tab?

HazAT (approved):
Thx

PR #3101, opened 2021-02-16
Update Spring RestTemplate docs.

maciejwalkowiak:
Reflect changes from https://github.com/getsentry/sentry-java/pull/1246 and https://github.com/getsentry/sentry-java/issues/1240


maciejwalkowiak:
cc @marandaneto

marandaneto (approved):
LGTM, let's merge only after released

marandaneto:
@maciejwalkowiak quick question, people that do have a `UriTemplateHandler` already set in the previous version, would need to remove it or it'd work ootb? if so, we'd need to add it to the migration page.

marandaneto:
> @maciejwalkowiak quick question, people that do have a `UriTemplateHandler` already set in the previous version, would need to remove it or it'd work ootb? if so, we'd need to add it to the migration page.

^

maciejwalkowiak:
Yes, they need to remove it. I guess we won't keep anymore migration page as a migration from 1.x to latest, but create sections for each version?

marandaneto:
> Yes, they need to remove it. I guess we won't keep anymore migration page as a migration from 1.x to latest, but create sections for each version?

not sure the best approach either, but we need to figure this out otherwise people won't know what to do, I believe a section for a minor version is a good start.

src/platforms/java/migration.mdx
```diff
@@ -4,6 +4,38 @@ sidebar_order: 1000
 description: "Learn about migrating from io.sentry:sentry 1.x to io.sentry:sentry 3.x."
 ---
 
+## Migrating from `io.sentry:sentry` `4.1.0` to `io.sentry:sentry` `4.1.1`
```
marandaneto, line 7:
```suggestion
## Migrating from `io.sentry:sentry` `4.1.0` to `io.sentry:sentry` `4.2.0`
```

let's bump a minor, a few breaking changes

src/platforms/java/migration.mdx
```diff
@@ -4,6 +4,82 @@ sidebar_order: 1000
 description: "Learn about migrating from io.sentry:sentry 1.x to io.sentry:sentry 3.x."
 ---
 
+## Migrating from `io.sentry:sentry` `4.1.0` to `io.sentry:sentry` `4.2.0`
+
+`operation` is now a required property of the `SentryTransaction` and `SentrySpan`. Whenever a transaction or span is started, the value for `operation` must be provided:
+
+```java
+Sentry.startTransaction("transaction-name", "operation-name");
+```
```
marandaneto, line 13: this should be also under https://github.com/getsentry/sentry-docs/blob/master/src/platforms/android/migration.mdx

marandaneto (approved):

PR #3111, opened 2021-02-17
Fix java configuration typo

robgarden:


marandaneto (approved):
good catch @robgarden thanks :)

PR #3115, opened 2021-02-17
Update performance monitoring verification step

PeloWriter:
The verification step for performance was causing confusion. It made it seem as if transactions would be simply sent, when in fact for some SDKs, you need to first use custom instrumentation to start and stop a transaction to test. This fix modifies the verification step appropriate to the instrumentation supported in each SDK.


bruno-garcia (commented):
LGTM!

PR #3117, opened 2021-02-18
ref(api): Use `fetch` for sentry user info

EvanPurkhiser:


src/components/codeContext.tsx
```diff
@@ -96,56 +97,59 @@ const formatApiUrl = ({ scheme, host }: Dsn) => {
...
-          }),
-        });
-      }
+/**
+ * Fetch project details from sentry
+ */
+export async function fetchCodeKeywords() {
+  let json: { projects: ProjectApiResult[] } | null = null;
+
+  const url = process.env.NODE_ENV
```
markstory, line 106: Would NODE_ENV be truthy during production builds?

EvanPurkhiser: Whoops, that was supposed to be a string comparison :)

EvanPurkhiser: Ok fixed this guy

markstory (approved):

PR #3122, opened 2021-02-18
fix(express): Add express integration to code snippet

lobsterkatie:
Express was relying on the common content, which of course doesn't have anything express-specific. This makes a separate page so that that content can be included.


PeloWriter (approved):
I think this works as an interim solution until we can focus on improving Node (and its children). Thanks.

PR #3123, opened 2021-02-18
Add DSN utilization

PeloWriter:
PR https://github.com/getsentry/sentry-docs/pull/1723 raised some good conversation, which is captured here (I've closed that PR with a pointer to this one as the branch was very stale)


armenzg (commented):
LGTM. I will let others chime in.

PR #3130, opened 2021-02-19
docs(usage): Format capture exception for platform

Zylphrex:
The capture exception function should be formatted appropriately for the
corresponding platform.


PeloWriter (approved):
Nice fix! Thank you

Issue #3057, opened 2021-02-09
 Symfony integrations docs need updates


Jean85:
## Important Details

This an issue with: Platform/SDK documentation

## Description

`sentry/sentry-symfony` got a new major release (4.0) 3 weeks ago, with a couple of breaking changes, up to the fact that it now leverages the latest major from `sentry/sentry` (PHP) package

### What you expected to happen

Have an up to date documentation for the official integration

### Possible Solution

Update the docs following all the informations already available in https://github.com/getsentry/sentry-symfony/blob/master/UPGRADE-4.0.md and in the changelog in https://github.com/getsentry/sentry-symfony/blob/master/CHANGELOG.md#400-2021-01-19

[reopened after the bot shut down #3056]

PR #3137, opened 2021-02-22
Add monolog integration to symfony docs

alexander-schranz:


Jean85:
This partially addresses #3057

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -19,30 +19,46 @@ composer require sentry/sentry-symfony
...
+```yaml {filename:config/packages/sentry.yaml}
 sentry:
   dsn: "___PUBLIC_DSN___"
 ```
 
-If you **are not** using Symfony Flex, you'll also need to enable the bundle in `app/AppKernel.php`:
+If you **are not** using Symfony Flex, you'll also need to enable the bundle in `config/bundles.php`:
 
-```php {filename:app/AppKernel.php}
+```php {filename:config/bundles.php}
```
Jean85, line 31: I'm not sure about this. Isn't the `bundles.php` file a Flex exclusive? IIRC without Flex you need to operate in the `Kernel.php` class, like shown in https://github.com/getsentry/sentry-symfony/pull/452

alexander-schranz: `app/Kernel.php` does not longer exist in symfony 4/5. The new symfony folder structure is `config` and this is normally not depending on if you use flex or not at you would normally still follow the symfony folder structure with `config/bundles.php` when doing symfony 4/5, you just don't have the composer flex plugin which will automatically add it for you. But we could show both examples with `app/Kernel.php` (symfony 2/3)  and the symfony 4/5 structure in this documentation.

Jean85: FTR the bundle now requires 3.4 at the very least: https://github.com/getsentry/sentry-symfony/blob/d6a738f005b57a16f156f3ae1024ec874c21b2a4/composer.json#L26-L32

I would make the Flex approach more prominent, but yeah, showing the old way is ok too.

ste93cry: m2c: considering that the status of [`3.4`](https://symfony.com/releases/3.4) is security fixes only, I would not bother adding any documentation for it because I find unlikely that someone will install the bundle on such an old (and "unmaintained") version. I would rather focus our efforts on `>= 4.x` versions

rhcarvalho:
@ste93cry and @Jean85 own the technical bits here. @PeloWriter FYI if you want to check the structure and fit!

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -19,30 +19,46 @@ composer require sentry/sentry-symfony
 
 ### Configure
```
PeloWriter, line 20:
```suggestion
## Configure
```

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -19,30 +19,46 @@ composer require sentry/sentry-symfony
...
+```
 
-        // ...
-    }
+### Monolog Integration
 
-    // ...
-}
+If monolog is used some additional configuration and services is needed
+to log the errors correctly:
```
PeloWriter, line 43:
```suggestion
If you are using monolog these additional configuration and services are needed to log the errors correctly:
```

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -19,30 +19,46 @@ composer require sentry/sentry-symfony
...
+
+services:
+    Monolog\Processor\PsrLogMessageProcessor:
+        tags: { name: monolog.processor, handler: sentry }
+
+    Sentry\Monolog\Handler:
+        arguments:
+            $hub: '@Sentry\State\HubInterface'
+            $level: !php/const Monolog\Logger::ERROR
 ```
```
PeloWriter, line 64: Can we add a verification step, similar to [this model](https://docs.sentry.io/platforms/javascript/#verify) in JavaScript?

Also, if we add this line to the end of the file, it will automatically list the associated supporting content:

`<PageGrid header="Next Steps" nextSteps />`

PeloWriter (approved):

ste93cry (changes requested):
Speaking generally, would it makes sense to have the Monolog documentation on its own page? Also, it's absolutely worth adding [this](https://github.com/getsentry/sentry-symfony/pull/445) warning because a lot of people unfortunately got hit by it recently

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -19,30 +19,46 @@ composer require sentry/sentry-symfony
...
+If monolog is used some additional configuration and services is needed
+to log the errors correctly:
+
+```php {filename:config/packages/sentry.yaml}
+sentry:
+    register_error_listener: false # Disables the ErrorListener to avoid duplicated log in sentry
+    dsn: "%env(SENTRY_DSN)%"
+
+monolog:
+    handlers:
```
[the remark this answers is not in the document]
alexander-schranz: As it the documentation is about how to integrate it in your symfony application, I would here follow the symfony coding standard where the indention is 4 spaces as all files in the same folder which they are created are the same. See as also the [symfony recipes](https://github.com/symfony/recipes) here all yaml files are 4 spaces. This are the files which are in the same folder this file is created.

src/platforms/php/guides/symfony/index.mdx
alexander-schranz: That is a good idea I think a controller with `$this->logger->error` called would be a good idea here to very if its working.

alexander-schranz: Example added to trigger and error and exception. (Both cases should be tested that the trigger sentry correctly).

alexander-schranz:
I updated the PR free for reviews again :)

alexander-schranz:
> Generally speaking, would it makes sense to have the Monolog documentation on its own page? 

@ste93cry Sorry saw this only yet. I think it doesn't make much sense in case of symfony monolog is so common that basically every project has it in it and the `symfony/website-skeleton` is shipped with it. And if you have monolog installed and don't configure it this way a lot of errors will maybe not send to sentry, which I think we want to avoid.

alexander-schranz:
> Also, it's absolutely worth adding this warning because a lot of people unfortunately got hit by it recently

overlooked this also. did add this also. Let me know if I need to change anything here.

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,104 @@ Install the `sentry/sentry-symfony` package:
...
+
+return [
+    // ...
+    Sentry\SentryBundle\SentryBundle::class => ['all' => true],
+];
+```
+
+### Monolog Integration
+
+If you are using monolog these additional configuration and services are needed to log the errors correctly:
```
apatkinson, line 54:
```suggestion
If you are using monolog, these additional configuration and services are needed to log the errors correctly:
```

ste93cry:
>Let me know if I need to change anything here

Since this warning is pretty important, it looks like by wrapping the text into `<Alert level="warning" title="Note">.../</Alert>` we get a nice colored and visible box that will make it more prominent

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,106 @@ Install the `sentry/sentry-symfony` package:
 composer require sentry/sentry-symfony
 ```
 
-### Configure
+**Warning:** due to a bug in all versions lower than `6.0` of the [`SensioFrameworkExtraBundle`](https://github.com/sensiolabs/SensioFrameworkExtraBundle) bundle,
```
alexander-schranz, line 21: @ste93cry  Should we now remove the `**Warning:**`?

```suggestion
Due to a bug in all versions lower than `6.0` of the [`SensioFrameworkExtraBundle`](https://github.com/sensiolabs/SensioFrameworkExtraBundle) bundle,
```

ste93cry: Yes, the box imho is more than enough to communicate that there is something important to read. I also noticed that the markdown for some reason does not work inside the box, you can see it here https://sentry-docs-git-fork-alexander-schranz-patch-1.sentry.dev/platforms/php/guides/symfony/

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,106 @@ Install the `sentry/sentry-symfony` package:
...
+```
+
+### Monolog Integration
+
+If you are using monolog, these additional configuration and services are needed to log the errors correctly:
+
+```yaml {filename:config/packages/sentry.yaml}
+sentry:
+    register_error_listener: false # Disables the ErrorListener to avoid duplicated log in sentry
+    dsn: "%env(SENTRY_DSN)%"
```
ste93cry, line 61:
```suggestion
```

This is needless for the example to work, so I would remove it. In the first part of the page this is already mentioned and I think that further example should just add things in addition to what is considered to be already present

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,106 @@ Install the `sentry/sentry-symfony` package:
...
 
-        // ...
+        throw new \RuntimeException('Example exception.');
     }
-
-    // ...
 }
 ```
+
+After you visit now `/_sentry-test` this **both** errors should appear on sentry.
```
ste93cry, line 122:
```suggestion
After you visit the `/_sentry-test` page, you should see two errors logged in Sentry.
```

However, this example is a bit tricky because the first error will be logged only if you are using the Monolog handler. Either we should specify this with a little comment in the code, or we should change the example itself to just use the `throw new` as it's enough to work in both cases

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,106 @@ Install the `sentry/sentry-symfony` package:
...
+
+return [
+    // ...
+    Sentry\SentryBundle\SentryBundle::class => ['all' => true],
+];
+```
+
+### Monolog Integration
+
+If you are using monolog, these additional configuration and services are needed to log the errors correctly:
```
ste93cry, line 56:
```suggestion
If you are using [Monolog](https://github.com/Seldaek/monolog), these additional configuration and services are needed to log the errors correctly:
```

ste93cry (changes requested):

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,106 @@ Install the `sentry/sentry-symfony` package:
...
+    }
 
-            new Sentry\SentryBundle\SentryBundle(),
-        );
+    /**
+     * @Route(name="sentry_test", path="/_sentry-test")
+     */
+    public function testLog(Request $request)
+    {
+        $this->logger->error('My custom logged error.');
```
alexander-schranz, line 115:
```suggestion
        // the following code will test if monolog integration works
        $this->logger->error('My custom logged error.');
```

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,106 @@ Install the `sentry/sentry-symfony` package:
...
 
-        // ...
+        throw new \RuntimeException('Example exception.');
```
alexander-schranz, line 117:
```suggestion
        // the following code will test if uncatched exception works with sentry
        throw new \RuntimeException('Example exception.');
```

alexander-schranz:
Updated the suggestions and fixed the markdown rendering: https://sentry-docs-git-fork-alexander-schranz-patch-1.sentry.dev/platforms/php/guides/symfony/

ste93cry (approved):
It looks good to me, lovely work and great patience! LGTM 🚢

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,109 @@ Install the `sentry/sentry-symfony` package:
 composer require sentry/sentry-symfony
 ```
 
-### Configure
 
-Add your DSN to `app/config/config.yml`:
+Due to a bug in all versions lower than `6.0` of the [`SensioFrameworkExtraBundle`](https://github.com/sensiolabs/SensioFrameworkExtraBundle) bundle,
+if you have it installed you will likely get an error during the execution of the command above in regards to the missing `Nyholm\Psr7\Factory\Psr17Factory`
+class. To workaround the issue, if you are not using the PSR-7 bridge, please change the configuration of that bundle as follows:
```
PeloWriter, line 24:
```suggestion
Due to a bug in all versions below `6.0` of the [`SensioFrameworkExtraBundle`](https://github.com/sensiolabs/SensioFrameworkExtraBundle) bundle, you will likely receive an error during the execution of the command above related to the missing `Nyholm\Psr7\Factory\Psr17Factory` class. To workaround the issue, if you are not using the PSR-7 bridge, please change the configuration of that bundle as follows:
```

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,109 @@ Install the `sentry/sentry-symfony` package:
 composer require sentry/sentry-symfony
 ```
 
-### Configure
```
PeloWriter, line 20:
```suggestion
<Note>
```

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,109 @@ Install the `sentry/sentry-symfony` package:
...
-```yaml {filename:app/config/config.yml}
+```yaml
+sensio_framework_extra:
+   psr_message:
+      enabled: false
+```
+
+For more details about the issue see [https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/710](https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/710).
+
+</Alert>
```
PeloWriter, line 34:
```suggestion
</Note>
```

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,109 @@ Install the `sentry/sentry-symfony` package:
...
+
+return [
+    // ...
+    Sentry\SentryBundle\SentryBundle::class => ['all' => true],
+];
+```
+
+### Monolog Integration
+
+If you are using [Monolog](https://github.com/Seldaek/monolog), these additional configuration and services are needed to log the errors correctly:
```
PeloWriter, line 58:
```suggestion
If you are using [Monolog](https://github.com/Seldaek/monolog), you also need these additional configuration and services to log the errors correctly:
```

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,109 @@ Install the `sentry/sentry-symfony` package:
...
+    Monolog\Processor\PsrLogMessageProcessor:
+        tags: { name: monolog.processor, handler: sentry }
 ```
 
-If you **are not** using Symfony Flex, you'll also need to enable the bundle in `app/AppKernel.php`:
+## Test the implementation
 
-```php {filename:app/AppKernel.php}
+To test that both logger error and exception are correctly send to sentry
+we can create the following controller:
```
PeloWriter, line 88:
```suggestion
To test that both logger error and exception are correctly sent to sentry.io, you can create the following controller:
```

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,109 @@ Install the `sentry/sentry-symfony` package:
...
-        // ...
+        // the following code will test if uncatched exception works
+        throw new \RuntimeException('Example exception.');
     }
-
-    // ...
 }
 ```
+
+After you visit the `/_sentry-test` page, you should see two errors logged in Sentry.
```
PeloWriter, line 125:
```suggestion
After you visit the `/_sentry-test` page, you can view and resolve the recorded error by logging into [sentry.io](https://sentry.io) and opening your project. Clicking on the error's title will open a page where you can see detailed information and mark it as resolved.
```

PeloWriter (approved):
Minor edits. Thank you!

alexander-schranz:
@PeloWriter applied your change requests, thx for the corrections :)

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,106 @@ Install the `sentry/sentry-symfony` package:
...
+```yaml {filename:config/packages/sentry.yaml}
 sentry:
-  dsn: "___PUBLIC_DSN___"
+    register_error_listener: false # Disables the ErrorListener to avoid duplicated log in sentry
+
+monolog:
+    handlers:
+        sentry:
+            type: service
+            id: Sentry\Monolog\Handler
```
HypeMC, line 66: Instead of using `type: service` why not use the [built-in `sentry` type](https://github.com/symfony/monolog-bundle/pull/313/files#diff-3ee12206c3b122be4593dbfdb19397c16b3d337f759fad3bde6c785d6f1261c0R671) and pass the `ClientInterface` as the `client_id`. This would remove the need to register the `Sentry\Monolog\Handler` manually:

```yaml
sentry:
    type: sentry
    client_id: Sentry\ClientInterface
```

Jean85: That builtin type refers to the SDK 2.0, we're on 3.0. It may still work, though.

HypeMC: @Jean85 If it's any help, I've been using this configuration with `sentry/sentry` version 3 for a while and haven't experienced any problems.

ste93cry: The built-in type is creating its own instance of the hub on the DI container and does not set it as current hub, so all integrations won't work as expected. It should definitely be possible to improve the Monolog bundle to make things work in the correct way, but as of now our way of instantiating the handler is safer

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,106 @@ Install the `sentry/sentry-symfony` package:
...
-            // ...
+        $this->logger = $logger;
+    }
 
-            new Sentry\SentryBundle\SentryBundle(),
-        );
+    /**
+     * @Route(name="sentry_test", path="/_sentry-test")
+     */
+    public function testLog(Request $request)
```
apatkinson, line 111:
```suggestion
    public function testLog()
```

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,106 @@ Install the `sentry/sentry-symfony` package:
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Annotation\Route;
+
+class SentryTestController
 {
-    public function registerBundles()
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
```
apatkinson, line 101: Do we not want to promote best practise?
```suggestion
    private LoggerInterface  $logger;
```

alexander-schranz: This depends on the PHP Version that it works and so I would avoid this. As the code should also work with older versions I would not use PHP 7.4 specific code if not really necessary to test the sentry implementation.

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -17,32 +17,106 @@ Install the `sentry/sentry-symfony` package:
...
+    /**
+     * @Route(name="sentry_test", path="/_sentry-test")
+     */
+    public function testLog(Request $request)
+    {
+        // the following code will test if monolog integration works
+        $this->logger->error('My custom logged error.');
 
-        // ...
+        // the following code will test if uncatched exception works
```
apatkinson, line 116:
```suggestion
        // The following code will test if an uncaught exception logs to Sentry
```

apatkinson (commented):
Does this PR need to update the `sentry-docs/src/platforms/php/guides/symfony/config.mdx` file which suggests you can have a monolog option, which is not true:

`Unrecognized option "monolog" under "sentry"`

Jean85:
Yes, the `monolog` option was dropped in the 3.0 release of the bundle: https://github.com/getsentry/sentry-symfony/pull/406

rhcarvalho:
> Does this PR need to update the `sentry-docs/src/platforms/php/guides/symfony/config.mdx` file which suggests you can have a monolog option, which is not true

As confirmed by @Jean85 this is a correct observation, but in my understanding orthogonal to this PR -- in other words, even without this PR we would already need to update that file to match the options of the latest SDK release.

I'm going to merge this as is and follow up with an edit to the `config.mdx` file.

Thanks everyone for the diligence!

PR #3138, opened 2021-02-22
Fix: Use originThrowable instead of throwable for Samples

marandaneto:
Merge only when https://github.com/getsentry/sentry-dart/issues/304 fix is deployed

### DIFF
--- a/src/includes/configuration/before-send-fingerprint/android.mdx
+++ b/src/includes/configuration/before-send-fingerprint/android.mdx
@@ -5,7 +5,7 @@ import java.util.Arrays;
 
 SentryAndroid.init(this, options -> {
   options.setBeforeSend((event, hint) -> {
-    if (event.getThrowable() instanceof SQLiteException) {
+    if (event.getOriginThrowable() instanceof SQLiteException) {
       event.setFingerprints(Arrays.asList("database-connection-error"));
     }
     return event;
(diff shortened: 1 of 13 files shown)


src/includes/configuration/before-send-fingerprint/android.mdx
```diff
@@ -5,7 +5,7 @@ import java.util.Arrays;
 
 SentryAndroid.init(this, options -> {
   options.setBeforeSend((event, hint) -> {
-    if (event.getThrowable() instanceof SQLiteException) {
+    if (event.getOriginThrowable() instanceof SQLiteException) {
```
bruno-garcia, line 8: Doesn't the getter of throwable already returns the original throwable?

bruno-garcia (commented):
new API?

src/includes/configuration/before-send-fingerprint/android.mdx
marandaneto: no

marandaneto:
> new API?

wrong docs I'd say

src/includes/configuration/before-send-fingerprint/android.mdx
philipphofmann: I'm just curious. Why did we rename this?

marandaneto: see the linked Dart issue, same thing.

philipphofmann (approved):
LGTM

marandaneto:
in favor of ttps://github.com/getsentry/sentry-docs/pull/3156

PR #3147, opened 2021-02-23
feat(apple): Add fingerprint database connection

philipphofmann:
This was missing for Apple. Sample code for RPC fingerprinting is missing as well. I'm going to add this with another PR.

### DIFF
--- a/src/includes/set-fingerprint/database-connection/apple.mdx
+++ b/src/includes/set-fingerprint/database-connection/apple.mdx
@@ -1,0 +1,29 @@
+```swift
+import Sentry
+import CoreData
+
+SentrySDK.start { options in
+    options.beforeSend = { event in
+        if let error = event.error as NSError? {
+            if (error.domain == NSSQLiteErrorDomain) {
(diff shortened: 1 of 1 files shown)


src/includes/set-fingerprint/database-connection/apple.mdx
```diff
@@ -0,0 +1,29 @@
+```swift
+import Sentry
+import CoreData
+
+SentrySDK.start { options in
+    options.beforeSend = { event in
+        if let error = event.error as NSError? {
```
brustolin, line 7: l: cast is redundant
```suggestion
        if let error = event.error  {
```

brustolin (commented):
Just a little contribution.

src/includes/set-fingerprint/database-connection/apple.mdx
philipphofmann: The problem is that `Error` doesn't have a domain, so I cant remove it.

marandaneto (approved):

PR #3148, opened 2021-02-23
Java: bump dependency versions to 4.2.0

### LINKED ISSUE #3134: Document improvements for Spring and performance annotations
Reflect changes from https://github.com/getsentry/sentry-java/pull/1243

@maciejwalkowiak

maciejwalkowiak:
Fix #3134

### DIFF
--- a/src/includes/getting-started-install/android.mdx
+++ b/src/includes/getting-started-install/android.mdx
@@ -16,6 +16,6 @@ android {
 
 // Add Sentry's SDK as a dependency.
 dependencies {
-    implementation 'io.sentry:sentry-android:{{ packages.version('sentry.java', '4.0.0') }}'
+    implementation 'io.sentry:sentry-android:{{ packages.version('sentry.java', '4.2.0') }}'
 }
 ```
--- a/src/includes/performance/enable-manual-instrumentation/java.spring-boot.mdx
+++ b/src/includes/performance/enable-manual-instrumentation/java.spring-boot.mdx
@@ -24,7 +24,7 @@ import io.sentry.spring.tracing.SentryTransaction;
 class ScheduledJob {
 
   @Scheduled(...)
-  @SentryTransaction("task")
+  @SentryTransaction(operation = "task")
   void execute() {
     ...
   }
@@ -40,7 +40,7 @@ import io.sentry.spring.tracing.SentryTransaction
 class ScheduledJob {
 
   @Scheduled(...)
-  @SentryTransaction("task")
+  @SentryTransaction(operation = "task")
   fun execute() {
     ...
   }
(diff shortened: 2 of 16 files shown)


src/includes/getting-started-install/android.mdx
```diff
@@ -16,6 +16,6 @@ android {
 
 // Add Sentry's SDK as a dependency.
 dependencies {
-    implementation 'io.sentry:sentry-android:{{ packages.version('sentry.java', '4.0.0') }}'
+    implementation 'io.sentry:sentry-android:{{ packages.version('sentry.java', '4.2.0') }}'
```
marandaneto, line 19: missing https://github.com/getsentry/sentry-docs/edit/master/src/platforms/android/integrations/timber.mdx

maciejwalkowiak: Fixed. Seems that Android docs don't follow the new structure much.

marandaneto: yeah it was before the big restructuring, we should introduce the `guides` for Android too but it's only a single integration for now (Timber) so not worth the effort.

src/includes/performance/enable-manual-instrumentation/java.spring-boot.mdx
```diff
@@ -40,14 +40,14 @@ import io.sentry.spring.tracing.SentryTransaction
...
-  @SentryTransaction
+  @SentryTransaction("task")
   fun execute() {
     ...
   }
 }
 ```
 
-`@SentryTransaction` can be configured with custom `name` and `op` properties. If not defined, `name` will be resolved from the class, and the method name.
+`@SentryTransaction` can be configured with custom `name`. If not defined, `name` will be resolved from the class, and the method name.
```
marandaneto, line 50: isn't `operation` mandatory for the annotation too?

maciejwalkowiak: It is mandatory and it's in the sample. Here it's about the optional parameters.

maciejwalkowiak: I still believe it should be somewhere documented what "name", "operation" and "description" are. In some place that is shared between all SDKs. Maybe it's just me, but it's very unclear what goes where in the UI until you try it out.

marandaneto: I understand, but the text says that the `name` will be resolved, what's about the `operation`?
either we infer the `operation` automatically too or our samples should be like `@SentryTransaction(name = "name", operation = "task")`, otherwise, if missing the `operation`, it'd throw I believe?

marandaneto: or `@SentryTransaction("task")` actually means `operation` == `task`?
maybe we should name the annotations?
`@SentryTransaction(operation = "task")` so it'd be easier to read?

maciejwalkowiak: Good point.

marandaneto (approved):
LGTM

PR #3151, opened 2021-02-23
ref: simplify and clarify web vitals docs

alxu-sentry:
I tried to make things clearer from my perspective as a "new user". Let me know if I missed something.

### DIFF
--- a/src/docs/product/performance/web-vitals.mdx
+++ b/src/docs/product/performance/web-vitals.mdx
@@ -20,7 +20,7 @@ The Sentry JavaScript SDK collects web vitals information from the browser (wher
 
 <Note>
 
-Some web vitals such as FP, FCP, LCP, and TTFB are measured relative to the start of the transaction. Values may differ when compared to values generated with other tools such as [Lighthouse](https://github.com/GoogleChrome/lighthouse).
+Some Web Vitals such as FP, FCP, LCP, and TTFB are measured relative to the start of the transaction. Values may differ when compared to values generated with other tools such as [Lighthouse](https://github.com/GoogleChrome/lighthouse).
 
 </Note>
 
--- a/src/docs/product/performance/web-vitals.mdx
+++ b/src/docs/product/performance/web-vitals.mdx
@@ -10,14 +10,14 @@ The in-browser Sentry SDKs collect Web Vitals information (where supported) and 
 
 ![Example of Web Vitals](./perf-web-vitals.png)
 
-| Web Vital                                                       | Availability                  | Failure Threshold |
-| --------------------------------------------------------------- | ----------------------------- | ----------------- |
-| [Largest Contentful Paint](#largest-contentful-paint-lcp) (LCP) | Chromium                      | 4 seconds         |
-| [First Input Delay](#first-input-delay-fid) (FID)               | Chromium, Firefox, and Safari | 300 milliseconds  |
-| [Cumulative Layout Shift](#cumulative-layout-shift-cls) (CLS)   | Chromium                      | 0.25              |
-| [First Paint](#first-paint-fp) (FP)                             | Chromium                      | 3 seconds         |
-| [First Contentful Paint](#first-contentful-paint-fcp) (FCP)     | Chromium                      | 3 seconds         |
-| [Time To First Byte](#time-to-first-byte) (TTFB)                | Chromium, Firefox, and Safari | 0.25              |
+| Web Vital                                                       | Availability                  | Needs improvement/Meh  | Bad/Fail          |
+| --------------------------------------------------------------- | ----------------------------- | ---------------------- | ----------------- |
+| [Largest Contentful Paint](#largest-contentful-paint-lcp) (LCP) | Chromium                      | 2.5+ seconds           | 4+ seconds        |
+| [First Input Delay](#first-input-delay-fid) (FID)               | Chromium, Firefox, and Safari | 100+ milliseconds      | 300+ milliseconds |
+| [Cumulative Layout Shift](#cumulative-layout-shift-cls) (CLS)   | Chromium                      | 0.1+                   | 0.25+             |
+| [First Paint](#first-paint-fp) (FP)                             | Chromium                      | 1+ second              | 3+ seconds        |
+| [First Contentful Paint](#first-contentful-paint-fcp) (FCP)     | Chromium                      | 1+ second              | 3+ seconds        |
+| [Time To First Byte](#time-to-first-byte-ttfb) (TTFB)           | Chromium, Firefox, and Safari | 200+ milliseconds      | 600+ milliseconds |
 
 <Note>
 
@@ -37,7 +37,7 @@ If you wish to see all of the data available, open the dropdown and click "View 
 
 ## Thresholds
 
-Google's "Bad", "Needs improvement", and "Good" levels are used to classify data points into red, yellow, and green for the corresponding Web Vitals.
+Google's "Bad", "Needs improvement", and "Good" levels are used to classify data points into red, yellow, and green for the corresponding Web Vitals. "Needs improvement" is sometimes referred to as "Meh" in Sentry.
 
 The vertical line for each Web Vital is the 75th percentile of the observed data points. In other words, 25% of the recorded values exceed that amount. If the value exceeds the Google Web Vitals "Bad" threshold, the Web Vital status is marked as "Fail".
 
--- a/src/docs/product/performance/web-vitals.mdx
+++ b/src/docs/product/performance/web-vitals.mdx
@@ -4,7 +4,9 @@ sidebar_order: 6
 description: "Learn more about Web Vitals in Sentry and how they give you a better understanding of your application's health."
 ---
 
-Frontend transactions may contain web vitals information if the browser provides them. With this view, you'll be able to examine a summary of the web vitals.
+[Web Vitals](https://web.dev/vitals/) are a set of metrics defined by Google to measure render time, response time, and the sum of shifting rendering processes. Each data point provides insights about the overall performance of your application.
+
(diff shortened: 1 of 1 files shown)


src/docs/product/performance/web-vitals.mdx
```diff
@@ -35,18 +34,22 @@ If you notice a region of interest on any of the histograms, click and drag over
...
 If you wish to see all of the data available, open the dropdown and click "View All". You will likely see extreme outliers when you click "View All". You can click and drag over an area to zoom in for a more detailed view.
 
-## Web Vitals
+## Thresholds
 
-[Web Vitals](https://web.dev/vitals/) are a set of metrics defined by Google to measure render time, response time, and the sum of shifting rendering processes. Each data point provides insights about the overall performance of your application.
+Google's "Bad", "Needs improvement", and "Good" levels are used to classify data points into red, yellow, and green for the corresponding Web Vitals.
 
-![Visualization of Web Vitals](./diagram-transaction-vitals.png)
+The vertical line for each Web Vital is the 75th percentile of the observed data points. In other words, 25% of the recorded values exceed that amount. If the value exceeds the Google Web Vitals "Bad" threshold, the Web Vital status is marked as "Fail".
```
Zylphrex, line 41: Does this need to explicitly call out "Google's Web Vitals Threshold"?

src/docs/product/performance/web-vitals.mdx
```diff
@@ -4,7 +4,9 @@ sidebar_order: 6
 description: "Learn more about Web Vitals in Sentry and how they give you a better understanding of your application's health."
 ---
 
-Frontend transactions may contain web vitals information if the browser provides them. With this view, you'll be able to examine a summary of the web vitals.
+[Web Vitals](https://web.dev/vitals/) are a set of metrics defined by Google to measure render time, response time, and the sum of shifting rendering processes. Each data point provides insights about the overall performance of your application.
```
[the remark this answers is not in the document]
alxu-sentry: didn't you write this?

src/docs/product/performance/web-vitals.mdx
```diff
@@ -16,13 +18,10 @@ Frontend transactions may contain web vitals information if the browser provides
 | [First Input Delay](#first-input-delay-fid)               | Chromium, Firefox, and Safari | 300 milliseconds  |
 | [Cumulative Layout Shift](#cumulative-layout-shift-cls)   | Chromium                      | 0.25              |
 
+<Note>
 
-## Baseline Value
-
-The baseline value for each web vital is the 75th percentile of the observed data points. It is compared against a failure threshold to give a pass or fail for each web vital.
+Some web vitals such as FP, FCP, LCP, and TTFB are measured relative to the start of the transaction. Values may differ when compared to values generated with other tools such as [Lighthouse](https://github.com/GoogleChrome/lighthouse).
```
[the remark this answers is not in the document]
alxu-sentry: i agree it's a little confusing, but 1. i think it was already using it before explaining it, 2. the name is written out in the table above, 3. it's just a note, if the reader doesn't understand i don't think it's too big a deal, and 4. i don't have any better ideas for where to put it. i considered putting all this stuff at the bottom since that's a more logical progression, but i also think that 90% of visitors will probably already be aware of web vitals and i don't want to bury our value-add.

src/docs/product/performance/web-vitals.mdx
```diff
@@ -35,18 +34,22 @@ If you notice a region of interest on any of the histograms, click and drag over
 
 If you wish to see all of the data available, open the dropdown and click "View All". You will likely see extreme outliers when you click "View All". You can click and drag over an area to zoom in for a more detailed view.
 
-## Web Vitals
+## Thresholds
 
-[Web Vitals](https://web.dev/vitals/) are a set of metrics defined by Google to measure render time, response time, and the sum of shifting rendering processes. Each data point provides insights about the overall performance of your application.
+Google's "Bad", "Needs improvement", and "Good" levels are used to classify data points into red, yellow, and green for the corresponding Web Vitals.
```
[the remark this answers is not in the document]
alxu-sentry: actually we kinda don't since https://github.com/getsentry/sentry/pull/23839. there might not be enough room in some places to rename them all to "needs improvement" though, so I guess I can explain that sometimes we call "needs improvement" "meh".

src/docs/product/performance/web-vitals.mdx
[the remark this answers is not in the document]
Zylphrex: Looks like I did, but now that it's been moved so far up on the document, it seems out of place without elaborating further.

src/docs/product/performance/web-vitals.mdx
alxu-sentry: in my first reading i was confused about who defined this threshold. i think maybe wording it as "If the value exceeds Google's "Bad" threshold" would be better?

src/docs/product/performance/web-vitals.mdx
```diff
@@ -4,7 +4,9 @@ sidebar_order: 6
 description: "Learn more about Web Vitals in Sentry and how they give you a better understanding of your application's health."
 ---
 
-Frontend transactions may contain web vitals information if the browser provides them. With this view, you'll be able to examine a summary of the web vitals.
+[Web Vitals](https://web.dev/vitals/) are a set of metrics defined by Google to measure render time, response time, and the sum of shifting rendering processes. Each data point provides insights about the overall performance of your application.
+
+The Sentry JavaScript SDK collects web vitals information from the browser (where supported) and adds that information to frontend transactions. These vitals are then summarized in several graphs for a quick overview of how each frontend transaction is performing for your users.
```
PeloWriter, line 9: Is there a reason we're focusing this solely on the JS SDK? If it's a function of only that SDK, then that's fine and we should say that. Otherwise, we can rephrase to be more comprehensive, such as "The browser-based SDKs" or something similar. Let me know if you need help with phrasing @jeremybourque

src/docs/product/performance/web-vitals.mdx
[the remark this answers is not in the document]
PeloWriter: In general, we do need to introduce the terms prior to using their acronym. @imatwawana has an idea that she'll propose.

src/docs/product/performance/web-vitals.mdx
```diff
@@ -35,18 +34,22 @@ If you notice a region of interest on any of the histograms, click and drag over
...
-[Web Vitals](https://web.dev/vitals/) are a set of metrics defined by Google to measure render time, response time, and the sum of shifting rendering processes. Each data point provides insights about the overall performance of your application.
+Google's "Bad", "Needs improvement", and "Good" levels are used to classify data points into red, yellow, and green for the corresponding Web Vitals.
 
-![Visualization of Web Vitals](./diagram-transaction-vitals.png)
+The vertical line for each Web Vital is the 75th percentile of the observed data points. In other words, 25% of the recorded values exceed that amount. If the value exceeds the Google Web Vitals "Bad" threshold, the Web Vital status is marked as "Fail".
 
 <Note>
 
-Some web vitals such as FP, FCP, LCP, and TTFB are measured relative to the beginning of the transaction's timestamp. Values may differ when compared to values generated with other tools such as [Lighthouse](https://github.com/GoogleChrome/lighthouse).
+For [Performance Metrics](../metrics/), "75th percentile" and "failure threshold" normally apply to transaction duration. Here, they instead refer to the individual Web Vitals.
```
PeloWriter, line 45:
```suggestion
For [Performance Metrics](../metrics/), "75th percentile", and "failure threshold" normally apply to transaction duration. Here, they instead refer to the individual Web Vitals.
```

PeloWriter (approved):
Minor smoothing.

src/docs/product/performance/web-vitals.mdx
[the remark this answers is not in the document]
imatwawana: I'd suggest adding the acronym, in parentheses, next to the spelled out term in the table before this section. For example: First Paint (FP). That should resolve the issue of having to introduce the acronyms in this paragraph.

imatwawana, line 23:
```suggestion
Some Web Vitals such as FP, FCP, LCP, and TTFB are measured relative to the start of the transaction. Values may differ when compared to values generated with other tools such as [Lighthouse](https://github.com/GoogleChrome/lighthouse).
```

src/docs/product/performance/web-vitals.mdx
imatwawana, line 9:
```suggestion
The Sentry JavaScript SDK collects Web Vitals information from the browser (where supported) and adds that information to frontend transactions. These Web Vitals are then summarized in several graphs for a quick overview of how each frontend transaction is performing for your users.
```

imatwawana (approved):

src/docs/product/performance/web-vitals.mdx
alxu-sentry: my understanding is that React etc are classified under the JS SDK, but AIUI it wouldn't be wrong to say "browser SDK" and I agree that it would probably reduce confusion for people thinking "wait does React count or not". then we could probably also drop "from the browser".

src/docs/product/performance/web-vitals.mdx
alxu-sentry: no, this is correct the way that it is. it's supposed to mean "75th percentile and failure threshold are used one way in performance metrics. here, we use it in a different way" but i can move the clause to the end if that sounds better to you

PeloWriter: ah, yes. You're correct -  no serial comma required here.

dashed (approved):
👏 👏 👏

Issue #3139, opened 2019-10-14
Inbound Filters not working?


sherpya:
I've enabled the filter feature, my sentry fails to connect to gitlab server (when the server is restarting), so I would to ignore them.

In custom filters -> error messages I've added:
```
ConnectTimeout
ConnectionError
```

I still get connection errors:
```
ConnectionError: SafeHTTPSConnectionPool(host='my.gitlab.server', port=443): Max retries exceeded with url: /oauth/token (Caused by NewConnectionError('<sentry.net.http.SafeHTTPSConnection object at 0x7f94737f46d0>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution',))
  File "sentry/tasks/check_auth.py", line 80, in check_auth_identity
    provider.refresh_identity(auth_identity)
  File "sentry/auth/providers/oauth2.py", line 196, in refresh_identity
    req = safe_urlopen(self.get_refresh_token_url(), data=data)
  File "sentry/http.py", line 124, in safe_urlopen
    **kwargs
  File "sentry/net/http.py", line 154, in request
    response = _Session.request(self, *args, **kwargs)
  File "requests/sessions.py", line 533, in request
    resp = self.send(prep, **send_kwargs)
  File "requests/sessions.py", line 646, in send
    r = adapter.send(request, **kwargs)
  File "requests/adapters.py", line 516, in send
    raise ConnectionError(e, request=request)
```

PeloWriter:
Thanks for all the input, folks. We're adding to the content for Inbound Data Filters to capture the conversation here.

PR #3154, opened 2021-02-23
Add error message filtering to Inbound Filters content

PeloWriter:
Responding to https://github.com/getsentry/sentry-docs/issues/3139

### DIFF
--- a/src/docs/product/data-management-settings/filtering/index.mdx
+++ b/src/docs/product/data-management-settings/filtering/index.mdx
@@ -24,8 +24,24 @@ Inbound filters include:
 - From specific release versions of your code
 - From certain IP addresses
 
+<Note>
+
+Filters do not apply to [sessions](/product/releases/health/#session). Error messages from minidumps also do not yet apply.
+
+</Note>
+
 Once applied, you can track the filtered events (numbers and cause) using the graph provided at the top of the Inbound Data Filters view.
 
 ![](builtin-inbound-filters.png)
 
-Data filters do not apply to [sessions](/product/releases/health/#session).
+## How Inbound Filters Work
+
+Inbound data filters are not case sensitive.
+
+### Filtering by Error Message
+
+To use inbound data filters for error messages, keep the following in mind:
+
+- The filter matches the entire error description. For example, to match "ConnectionError", use the filter `*ConnectionError*`. Because filters are not case sensitive, `*connectionerror*` is fine, too.
+- Square brackets (`[` and `]`) have a special meaning due to the way our globs are evaluated. To match a square bracket, escape it as in this example: `*please try again \[not found\]*`.
+- Exceptions with an exception title (usually the class name) and description are joined with a colon, which is the the same as it is in the terminal. We don't recommend this type of matching, however, and suggest you match with wildcards instead. For example: Exception title "ConnectionError" and description "Max retries exceeded" will join to a literal "ConnectionError: Max retries exceeded" for filter evaluation.
--- a/src/docs/product/data-management-settings/filtering/index.mdx
+++ b/src/docs/product/data-management-settings/filtering/index.mdx
@@ -24,24 +24,44 @@ Inbound filters include:
 - From specific release versions of your code
 - From certain IP addresses
 
-<Note>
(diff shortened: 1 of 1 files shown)


src/docs/product/data-management-settings/filtering/index.mdx
```diff
@@ -24,8 +24,24 @@ Inbound filters include:
 - From specific release versions of your code
 - From certain IP addresses
 
+<Note>
+
+Filters do not apply to [sessions](/product/releases/health/#session). Error messages from minidumps also do not yet apply.
```
jan-auer, line 29: Can we mark this as a warning? This needs to stick out.

src/docs/product/data-management-settings/filtering/index.mdx
```diff
@@ -24,8 +24,24 @@ Inbound filters include:
...
 Once applied, you can track the filtered events (numbers and cause) using the graph provided at the top of the Inbound Data Filters view.
 
 ![](builtin-inbound-filters.png)
 
-Data filters do not apply to [sessions](/product/releases/health/#session).
+## How Inbound Filters Work
+
+Inbound data filters are not case sensitive.
+
+### Filtering by Error Message
```
jan-auer, line 41: Once we add more sections these will all read a redundant prefix "Filtering by ...". Is that intended?

src/docs/product/data-management-settings/filtering/index.mdx
```diff
@@ -24,8 +24,24 @@ Inbound filters include:
...
+
+To use inbound data filters for error messages, keep the following in mind:
+
+- The filter matches the entire error description. For example, to match "ConnectionError", use the filter `*ConnectionError*`. Because filters are not case sensitive, `*connectionerror*` is fine, too.
+- Square brackets (`[` and `]`) have a special meaning due to the way our globs are evaluated. To match a square bracket, escape it as in this example: `*please try again \[not found\]*`.
+- Exceptions with an exception title (usually the class name) and description are joined with a colon, which is the the same as it is in the terminal. We don't recommend this type of matching, however, and suggest you match with wildcards instead. For example: Exception title "ConnectionError" and description "Max retries exceeded" will join to a literal "ConnectionError: Max retries exceeded" for filter evaluation.
```
jan-auer, line 47: Suggestion:

    - The filter matches the entire error description in the format `{exception.type}: {exception.value}`. We do not recommend matching the full error description including the colon, and suggest you match with wildcards. For example, to match any "ConnectionError", use the filter `*ConnectionError*`.
    - You can provide multiple patterns, one per line. The filter applies if any of the patterns match.
    - Events without errors such as transactions or message events are never matched by this filter.

    ### Filtering by Releases

    To filter releases, keep the following in mind:

    - The filter matches the full release name provided during SDK initialization. If you provide the recommended package prefix, the release is in the format `package@version`, for example: `my-example@1.4.0-beta.1`. 
    - Regular glob rules apply and there is no special casing for SemVer. This allows for matching prefixes, such as `my-example@1.*`.
    - The filter never applies to events without a release.

    ### Glob Matching

    The error messages and releases filters use glob patterns. Globs are case insensitive and allow you to specify wildcards to match variable input. For example, `*panic*` matches any error that contains the words "panic", "PANIC" or "PaNiC". 

    Some symbols, such as the `*` character, receive special meaning as meta characters. To match a literal asterisk, escape it with a backslash: `\*`. Inbound data filters use the following meta characters:

    - `?` matches any single character.
    - `*` matches zero or more characters.
    - `\` escapes the next character making it literal. If it precedes a non-meta character, the backslash is ignored.
    - `[`, `]`, `{`, and `}` are reserved meta characters.
    ```

src/docs/product/data-management-settings/filtering/index.mdx
PeloWriter: I will fix. Thanks.

src/docs/product/data-management-settings/filtering/index.mdx
PeloWriter: Will use this. Thank you.

jan-auer (approved):

PR #3160, opened 2021-02-25
Fix: special case Proguard mapping location

### LINKED ISSUE #3046: Document new ProGuard mapping location in settings.
We're moving the proguard mapping file from the `Debug Info` into its own page under project settings.
Any links or texts pointing user to "debug info" must be updated once the feature flag is dropped.

marandaneto:
Fix #3046

We can't completely remove the support of `debug files` for ProGuard because we still need to support debug files for NDK.
Also, the debug files page tell how to use sentry-cli manually, hence this small change is a quick win instead of moving or duplicating the content around.

### DIFF
--- a/src/docs/product/cli/dif.mdx
+++ b/src/docs/product/cli/dif.mdx
@@ -272,7 +272,7 @@ sentry-cli upload-proguard \
     app/build/outputs/mapping/release/mapping.txt
 ```
 
-After the upload, Sentry deobfuscates future events. If you want to send an obfuscated crash to Sentry to verify correct operation, ensure that the ProGuard mapping files are listed in Project Settings > ProGuard.
+After the upload, Sentry deobfuscates future events. If you want to send an obfuscated crash to Sentry to verify the correct operation, ensure that the ProGuard mapping files are listed in Project Settings > ProGuard.
 
 ### Upload Options
 
--- a/src/docs/product/cli/dif.mdx
+++ b/src/docs/product/cli/dif.mdx
@@ -272,7 +272,7 @@ sentry-cli upload-proguard \
     app/build/outputs/mapping/release/mapping.txt
 ```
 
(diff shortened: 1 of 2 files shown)


src/docs/product/cli/dif.mdx
```diff
@@ -272,6 +272,8 @@ sentry-cli upload-proguard \
     app/build/outputs/mapping/release/mapping.txt
 ```
 
+After the upload, Sentry deobfuscates future events. If you want to send an obfuscated crash to Sentry to verify correct operation, ensure that the ProGuard mapping files are listed in Project Settings > ProGuard.
```
philipphofmann, line 275:
```suggestion
After the upload, Sentry deobfuscates future events. If you want to send an obfuscated crash to Sentry to verify the correct operation, ensure that the ProGuard mapping files are listed in Project Settings > ProGuard.
```

philipphofmann: Should we maybe mention the previous location for enterprise customers?

philipphofmann (commented):
When are we going to see Proguard in the setting?

src/docs/product/cli/dif.mdx
marandaneto: don't think so, the sentence above is about debug files, it'd be natural to just read the sentence below and get it.

src/docs/product/cli/dif.mdx
```diff
@@ -272,6 +272,8 @@ sentry-cli upload-proguard \
     app/build/outputs/mapping/release/mapping.txt
 ```
 
+After the upload, Sentry deobfuscates future events. If you want to send an obfuscated crash to Sentry to verify the correct operation, ensure that the ProGuard mapping files are listed in Project Settings > ProGuard.
```
bruno-garcia, line 275: What if we format the instruction to stand out like:

```suggestion
After the upload, Sentry deobfuscates future events. If you want to send an obfuscated crash to Sentry to verify the correct operation, ensure that the ProGuard mapping files are listed in _Project Settings > ProGuard_.
```

bruno-garcia (approved):

PR #3161, opened 2021-02-25
Excluded exceptions was dropped in v4. Apply changes to Sentry Symfony documentation.

sebo:
Excluded exceptions was dropped in v4.
 Similiar functionality can be created with an integration.

See upgrade guide https://github.com/getsentry/sentry-symfony/blob/master/UPGRADE-4.0.md "Removed the sentry.options.excluded_exceptions configuration option."


dashed:
@ste93cry would you be able to confirm this change?

ste93cry:
Change is valid, however we should add a few notes about how to use the `IgnoreErrorsIntegration` integration in place of the deprecated SDK option

sebo:
Should we add a complete new section to the configuration that is called integrations e.g. https://docs.sentry.io/platforms/php/guides/symfony/configuration/integrations ?  At least for me it was hard to figure out how to do it properly. I spotted it by chance in the upgrade guide.

ste93cry:
I would just document how to use the integration in a way similar to what is done for the [JS SDK](https://docs.sentry.io/platforms/javascript/configuration/integrations/). But I will leave the final word on what to do to the Sentry guys 😃

rhcarvalho:
I opened https://github.com/getsentry/sentry-docs/issues/3362 to track documenting the Symfony integrations. Since the change proposed here is valid we can as well merge it and not block on having the other documentation (rationale is that having outdated documentation that doesn't work doesn't help customers).

rhcarvalho (approved):
Thanks @sebo!

PR #3166, opened 2021-02-26
Update React Native content

PeloWriter:
This PR restructures the React Native content:
- adds a feature overview
- moves content into the common structure
- creates a section for manual setup to house manual configuration, Hermes, Codepush, RAM bundles, and manual source map generation and uploads.


src/includes/getting-started-primer/react-native.mdx
```diff
@@ -0,0 +1,20 @@
+<Note>
+
+_Sentry's React Native SDK enables automatic reporting of errors and exceptions, and identifies perfomance issues in your application. The React Native SDK includes performance monitoring (our `@sentry/tracing` package) by default._
+
+</Note>
+
+Features:
```
bruno-garcia, line 7: I really like the feature list as the first thing people see. It's super helpful to those landing here without understanding well what Sentry can do just yet and should do a good job selling them on moving on with installing.

bruno-garcia (approved):

src/includes/getting-started-primer/react-native.mdx
```diff
@@ -0,0 +1,20 @@
...
+
+- Automatic Native Crash [Error Tracking](/product/error-monitoring/) (using both `sentry-cocoa` & `sentry-android`)
+- Offline storage of events
```
marandaneto, line 10: could be linked to the draining section, noticed RN does not have it.

See https://sentry-docs-git-update-react-native-content.sentry.dev/platforms/android/configuration/draining/

src/includes/getting-started-primer/react-native.mdx
```diff
@@ -0,0 +1,20 @@
...
+- Events enriched with device data
```
marandaneto, line 11: Link to https://sentry-docs-git-update-react-native-content.sentry.dev/platforms/react-native/enriching-events/context/ maybe?

src/includes/capture-error/react-native.mdx
```diff
@@ -0,0 +1,9 @@
+In most situations, you can capture errors automatically with `captureException()`.
+
+```JavaScript
```
jennmueng, line 3:
```suggestion
```javascript
```
L: Lowercase javascript to match the other code blocks

src/includes/configuration/auto-session-tracking/react-native.mdx
```diff
@@ -1,7 +1,28 @@
 To benefit from the health data, you must use at least version 1.4.0 of the React Native SDK and enable the collection of the release health data in the initialization options of the SDK.
 
+We mark the session as crashed if an _unhandled error_ or _unhandled promise rejection_ bubbled up to the global handler.
```
jennmueng, line 3: Off topic of this docs PR, but is this true? unhandled promise rejections don't crash the app in RN and just fail silently, would make more sense to mark the session as errored rather than crashed. I'm not sure where this marking is being done @bruno-garcia

src/includes/configuration/auto-session-tracking/react-native.mdx
```diff
@@ -1,7 +1,28 @@
...
+
+We mark the session as an error if the SDK captures an event that contains an exception (this includes manually captured errors).
+
 ```javascript
 Sentry.init({
   autoSessionTracking: true
 });
 ```
+
+By default, the SDK sends sessions. To disable this toggle the flag `autoSessionTracking` to `false`.
```
jennmueng, line 13:
```suggestion
By default, the SDK sends sessions. To disable this toggle the flag `enableAutoSessionTracking` to `false`.
```

src/includes/getting-started-install/react-native.mdx
```diff
@@ -0,0 +1,66 @@
...
+npm install --save @sentry/react-native
+```
+
+```bash {tabTitle:Yarn}
+yarn add @sentry/react-native
+```
+
+<Note>
+
+If you are using Expo, see [How to Add Sentry to Your Expo Project](https://docs.expo.io/guides/using-sentry/). This SDK works only for either ejected projects or projects that directly use React Native.
```
jennmueng, line 13:
```suggestion
If you are using Expo, see [How to Add Sentry to Your Expo Project](https://docs.expo.io/guides/using-sentry/). This SDK works for both managed and bare projects.
```
Original text is no longer true as Expo updated it so it works on all expo projects now

src/includes/getting-started-install/react-native.mdx
```diff
@@ -0,0 +1,66 @@
...
+```bash {tabTitle:Yarn}
+yarn sentry-wizard -i reactNative -p ios android
+
+cd ios
+pod install
+```
+
+Next, install the dependencies for [auto-linking](https://github.com/react-native-community/cli/blob/master/docs/autolinking.md) in iOS, which relies on CocoaPods.
+
+If you are running a project with `react-native < 0.60`. you still need to call `react-native link`.
```
jennmueng, line 37: `react-native link` still needs to be done for <0.60, and needs to be before the sentry-wizard and pod install steps above

jennmueng (commented):
I just a couple of points, otherwise this is great.

src/includes/getting-started-primer/react-native.mdx
marandaneto: it was missing for Apple btw but I just added it https://github.com/getsentry/sentry-docs/pull/3169

src/includes/configuration/auto-session-tracking/react-native.mdx
marandaneto: @jennmueng what's the `event.mechanism.handled` value for an unhandled promise from RN events?
if it's `false`, on Android, we'd mark it as `crashed`.
on IOS, it's different, see https://github.com/getsentry/sentry-cocoa/pull/906

fwiw, on Flutter, we never mark as crashed unless there's a native crash (C/C++) code otherwise it's always errored, Flutter never crashes on the Dart layer, but RN does, right?

jennmueng: @marandaneto I see, yes on RN an unhandled error on the JS engine would crash the app.

I just checked, `event.exception.mechanism.handled` is true (although `event.mechanism` is not defined), and the level is `error`. So the session wouldn't be marked as crashed then @PeloWriter

marandaneto: its `event.exception.mechanism.handled`, it was a typo, thanks for checking this out.

marandaneto: > So the session wouldn't be marked as crashed then @PeloWriter

for _unhandled error_ or _unhandled promise rejection_, but for native crashes (Android and iOS) bits, they will be marked as Crashed.

jennmueng: Although note as a user, I would expect an unhandled promise rejection to be `handled: false` though. Would give me the wrong idea if an unhandled error is marked as handled.

marandaneto: @jennmueng the idea of `handled: false` for Mobile is, `did my App fully crashed/terminate?` so if it doesn't, it does make sense to be `handled: true`

jennmueng: Hmm wouldn't `level: fatal` make more sense for that use?

marandaneto: that's how the protocol was designed

src/includes/getting-started-primer/react-native.mdx
PeloWriter: RN is listed as `notSupported` for the draining info. I'll enable it. Is the Android snippet correct for RN as well? (if so, I'll modify to specify RN, but use that language.)

src/includes/getting-started-primer/react-native.mdx
PeloWriter: Done. Linked to /platforms/react-native/enriching-events/context/. Thanks.

src/includes/capture-error/react-native.mdx
PeloWriter: done. Thanks.

src/includes/configuration/auto-session-tracking/react-native.mdx
PeloWriter: Updating to:

> We mark the session as an error if the SDK captures an event that contains an exception (this includes manually captured errors) for _unhandled error_ or _unhandled promise rejection_, but for native crashes (Android and iOS), they will be marked as crashed.

src/includes/getting-started-install/react-native.mdx
PeloWriter: Can you clarify? As this reads now we're still telling users to run the `react-native link`, with the code sample, then use Sentry wizard. So, I'm not sure what change to make, other than to move line 37 prior to line 35, perhaps?

src/includes/configuration/auto-session-tracking/react-native.mdx
jennmueng: This would correct:
> We mark the session as errored if the SDK captures an event that contains an exception (this includes manually captured errors) or a unhandled promise rejection, but for unhandled errors and native crashes (Android and iOS), they will be marked as crashed.

PeloWriter: Updated to above. Thanks.

marandaneto (approved):
thanks @PeloWriter way way better :)

PR #3169, opened 2021-03-01
Fix: Add missing Apple draining example

marandaneto:
See Android https://docs.sentry.io/platforms/android/configuration/draining/


philipphofmann (approved):
Thanks, LGTM.

PR #3171, opened 2021-03-01
Adding trace rate comment for prod deploy

rbisol:
Adding the comment telling users to adjust traces sample rate in production. This comment was present only in the performance pages, not on the instrumentation guide in the Sentry web ui.


PeloWriter (approved):
Thank you.

PR #3181, opened 2021-03-03
fix(membership): Remove gratuitous asterisks

JoshFerge:
These asterisks are either a formatting issue or linked to some long lost footnotes that are no longer on the page. This PR removes them


PeloWriter (approved):
Fab. Thanks for finding this and fixing.

PR #3188, opened 2021-03-04
Feat: Android Activity instrumentation docs (Performance)

marandaneto:
merge only after releasing the SDK.


src/platforms/android/common/performance/included-instrumentation.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+<Note>
+
+Capturing transactions requires that you first <PlatformLink to="/performance/">set up performance monitoring</PlatformLink> if you haven't already.
+
+</Note>
+
+<Note>
+
+This is only available since Sentry's Android SDK version `4.3.0`
```
marandaneto, line 15: version TBD

src/platforms/android/common/performance/included-instrumentation.mdx
```diff
@@ -0,0 +1,78 @@
...
+You may finish the transaction manually based on your App's business logic, for example, the transaction should be only finished when your API call triggered on Activity's `onCreate` has been executed and you've shown the the data in the UI to the user, to do that you should do:
+
+```java
+import io.sentry.Sentry;
+import io.sentry.SpanStatus;
+
+ISpan span = Sentry.getSpan();
+if (span != null) {
+  span.finish(SpanStatus.OK);
+}
```
marandaneto, line 52: missing kotlin snippets

src/platforms/android/common/performance/included-instrumentation.mdx
```diff
@@ -0,0 +1,78 @@
...
+  ISpan userDataSpan = span.startChild("displayUserData");
+
+  // do some work
+
+  userDataSpan.finish(SpanStatus.OK);
+
+  // in case you've opt-out of the 'auto-finish' feature, finish the transaction manually.
+  span.finish(SpanStatus.OK);
+}
+```
```
marandaneto, line 76: missing kotlin snippets

HazAT (approved):
I wasn't sure about all the `finish(SpanStatus.OK)` but I think it's fine.
Just makes the code samples more complex.
But no other objections.

marandaneto:
> I wasn't sure about all the `finish(SpanStatus.OK)` but I think it's fine.
> Just makes the code samples more complex.
> But no other objections.

otherwise tranasctions won't have a `status` at all, what we could do is, once `finish` is called ,if there's no status, we set to `ok`.

src/platforms/android/common/performance/included-instrumentation.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+<Note>
+
+This is only available since Sentry's Android SDK version `4.3.0`
+
+</Note>
+
+### Android's Activity Instrumentation
+
+The Activity's instrumentation, once enabled, captures transactions for each Activity's start up, Transaction names follow pattern `ActivityName`, for example for an Activity called `MainActivity`, The Transaction's name is `MainActivity` and its operation is `navigation`.
```
philipphofmann, line 21:
```suggestion
The Activity's instrumentation, once enabled, captures transactions for each launch of an Activity. The SDK sets the Transaction name to the name of the Activity, for example, `MainActivity`, and Transaction operation to `navigation`.
```

src/platforms/android/common/performance/included-instrumentation.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+The transaction starts before each Activity's `onCreate` method is called.
```
philipphofmann, line 23: `m`: Maybe we should mention that we hook into `onActivityPreCreated`.

src/platforms/android/common/performance/included-instrumentation.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+_The Activity's instrumentation is enabled by default, but you may disable it by setting:_
+
+```xml {filename:AndroidManifest.xml}
+<manifest>
+    <meta-data android:name="io.sentry.traces.activity.enable" android:value="false" />
+</manifest>
+```
```
philipphofmann, line 31: `l`: Shouldn't we also mention how to disable this with code?

src/platforms/android/common/performance/included-instrumentation.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+_The transaction finishes automatically, but you may disable it by setting:_
+
+```xml {filename:AndroidManifest.xml}
+<manifest>
+    <meta-data android:name="io.sentry.traces.activity.auto-finish.enable" android:value="false" />
+</manifest>
+```
+
+You may finish the transaction manually based on your App's business logic, for example, the transaction should be only finished when your API call triggered on Activity's `onCreate` has been executed and you've shown the the data in the UI to the user, to do that you should do:
```
philipphofmann, line 43:
```suggestion
We offer the possibility to finish the transaction manually. For example, you might want to finish the transaction after an API call triggered on Activity's `onCreate` has been executed and you've shown the data to the user. To achieve that you can do:
```

src/platforms/android/common/performance/included-instrumentation.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+```java
+import io.sentry.Sentry;
+import io.sentry.SpanStatus;
+
+ISpan span = Sentry.getSpan();
```
philipphofmann, line 49: `m`: Maye we should mention that we put the transaction there with a comment.

src/platforms/android/common/performance/included-instrumentation.mdx
```diff
@@ -0,0 +1,78 @@
...
+if (span != null) {
+  span.finish(SpanStatus.OK);
+}
+```
+
+In case that the user left this Activity before you've finished the transaction manually, the transaction will be finished automatically once the `onDestroy` method is called.
```
philipphofmann, line 55:
```suggestion
In case that the user left this Activity before you've finished the transaction manually, the SDK finishes the transaction automatically when the `onDestroy` method is called.
```

src/platforms/android/common/performance/included-instrumentation.mdx
```diff
@@ -0,0 +1,78 @@
...
+
+Transactions are always finished with the `SpanStatus.OK` status if there's none set.
+
+Transactions are always bound to the Scope automatically if there's none set, because of that, you may create spans using the manual instrumentation and those spans will be automatically associated with the Activity's running transaction, for example.
```
philipphofmann, line 59:
```suggestion
Transactions are always bound to the Scope automatically if there's none set. Because of that, you may create spans using manual instrumentation, and those spans will be automatically associated with the Activity's running transaction.
```

src/platforms/android/common/performance/included-instrumentation.mdx
```diff
@@ -0,0 +1,78 @@
...
+  // do some work
+
+  userDataSpan.finish(SpanStatus.OK);
+
+  // in case you've opt-out of the 'auto-finish' feature, finish the transaction manually.
+  span.finish(SpanStatus.OK);
+}
+```
+
+If you've not finished the transaction yet, but you've started a new Activity, the previous transaction will be finished automatically regardeless the feature 'auto-finish' being enabled or not, due to the limitation of having a single transaction per time bound to the Scope, to work around that, you may create transactions manually using the custom instrumentation and using its instance to start spans instead of the Static API.
```
philipphofmann, line 78:
```suggestion
When you didn't finish the transaction yet, but you start a new Activity, the SDK always automatically finishes the previous Activity transaction. This is due to the limitation of having only one transaction per time-bound to the Scope. To work around that, you may create transactions manually using the custom instrumentation and its instance to start spans instead of the Static API.
```

philipphofmann (commented):
Good docs. Most of my comments are nitpicks on the wording. Feel free to drop them.

src/platforms/android/common/performance/included-instrumentation.mdx
marandaneto: > launch of an Activity

while it does make sense, starting an activity is a `context.startActivity` method and not a `launchActivity` so I guess the wording should be rather `start`, it's more natural

src/platforms/android/common/performance/included-instrumentation.mdx
marandaneto: not sure, that's impl. details, and with an user friendly desc. `before each Activity's `onCreate` method is called` which is what `onActivityPreCreated` does

src/platforms/android/common/performance/included-instrumentation.mdx
marandaneto: I'd keep it rather short as it's enabled by default, we already have the basic options page for both ways of doing it, which I've not added yet

src/platforms/android/common/performance/included-instrumentation.mdx
marandaneto: its already described in this section

> Transactions are always bound to the Scope automatically

src/platforms/android/common/performance/included-instrumentation.mdx
PeloWriter: Let's combine these notes, so we don't have one, then the next. Also suggest we use this wording:
```suggestion
Supported in Sentry's Android SDK version `4.3.0` and above.
```

PeloWriter (approved):
Philip provided most of the copy fixes. LGTM

src/platforms/android/common/performance/included-instrumentation.mdx
bruno-garcia: I second Philipp's suggestion and further suggest we rename:

> This is due to the limitation of having only one transaction per time-bound to the Scope.

To

> This is due to the fact that only one transaction at a time can be bound to the Scope.

bruno-garcia (commented):
No snippet covers setting a Span in a failed state. What about we have them with `try/catch` instead, showing that either way (normal flow or an exception thrown) we need to finish spans.

marandaneto:
> I wasn't sure about all the `finish(SpanStatus.OK)` but I think it's fine.
> Just makes the code samples more complex.
> But no other objections.

I removed it, its cleaner, thanks

src/platforms/android/common/performance/included-instrumentation.mdx
```diff
@@ -0,0 +1,109 @@
...
+
+void displayUserData() {
+  ISpan span = Sentry.getSpan();
+  if (span != null) {
+    ISpan innerSpan = span.startChild("displayUserData");
+    try {
+      // omitted code
+    } catch (Exception e) {
+      innerSpan.setThrowable(e);
+      innerSpan.setStatus(SpanStatus.NOT_FOUND);
```
bruno-garcia, line 76: Curious if we need a status when an exception is set. Is it optional?
In case there's no status and an exception is set, can we infer the status via the exception type?

https://github.com/getsentry/sentry-dotnet/blob/64d4b36002e179464231189f709e98dee087d822/src/Sentry/Span.cs#L95-L110

/cc @maciejwalkowiak

PR #3195, opened 2021-03-04
AWS Lambda integration - serverless framework deployment documentation

owain68:
I have raised this with support.


scefali:
@owain68 Thanks for making this PR! We totally understand that you are having problems using the Serverless framework and Sentry. We definitely want to make this experience better. We will review this PR and get back to you next week.

owain68:
Yes the documentation still needs amending to point out that this integration does not work with the serverless framework. @scefali I am not sure where you are with this.

scefali (approved):

PR #3204, opened 2021-03-06
fix: Split react native install code blocks

jennmueng:
RN >= does not need to run the link step

### DIFF
--- a/src/includes/getting-started-install/react-native.mdx
+++ b/src/includes/getting-started-install/react-native.mdx
@@ -16,13 +16,17 @@ If you are using Expo, see [How to Add Sentry to Your Expo Project](https://docs
 
 ### Linking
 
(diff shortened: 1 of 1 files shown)


src/includes/getting-started-install/react-native.mdx
```diff
@@ -16,13 +16,17 @@ If you are using Expo, see [How to Add Sentry to Your Expo Project](https://docs
...
+Link the SDK to your native projects to enable native crash reporting.
 
 ```bash
 react-native link @sentry/react-native
 ```
 
-```NPX
+<Note>
+If you are running a project with `react-native` < 0.60, you still need to call `react-native link`, otherwise you can skip this step as `react-native` >=0.60 does this automatically.
+</Note>
```
PeloWriter, line 27:
```suggestion
<Note>

If you are running a project with `react-native` prior to 0.60, you still need to call `react-native link`, otherwise you can skip this step as `react-native` version 0.60 and above does this automatically.

</Note>
```

PeloWriter (approved):

PR #3205, opened 2021-03-06
Fix NLog DSN Constructor documentation

TeknoVenus:
Release 3.x of the .NET SDK changes DSN to take a string instead of object - update NLog docs accordingly


bruno-garcia (approved):
Oh good catch! thank you

PR #3206, opened 2021-03-08
fix: Use correct fastlane command

HazAT:
`upload_symbols_to_sentry` doesn't exist, it's called `sentry_upload_dsym`
ref: https://github.com/getsentry/sentry-fastlane-plugin#uploading-symbolication-files


marandaneto (approved):
good catch

philipphofmann (approved):
Thanks 🔥

PR #3208, opened 2021-03-08
fix: Correct a small typo

ruairidhwm:
Hi,

This fixes a small typo to make the use of "its" possessive, instead of a contraction.


dashed (approved):
Thanks for the typo fix! 👍

PR #3215, opened 2021-03-08
fix(tracing): Fix dangling modifiers

lobsterkatie:
They were adjectives in search of a noun. They were lonely. It was sad. Now it's fixed.

### DIFF
--- a/src/includes/performance/enable-automatic-instrumentation/javascript.mdx
+++ b/src/includes/performance/enable-automatic-instrumentation/javascript.mdx
@@ -1,4 +1,4 @@
-After configuration, you will see both `pageload` and `navigation` when viewing transactions in sentry.io.
+After configuration, you will see both `pageload` and `navigation` transactions in sentry.io.
 
 ```javascript {tabTitle: ESM}
 // If you're using one of our integration packages, like `@sentry/react` or `@sentry/angular`,


src/includes/performance/enable-automatic-instrumentation/javascript.mdx
```diff
@@ -1,4 +1,4 @@
-After configuration, you will see both `pageload` and `navigation` when viewing transactions in sentry.io.
+After configuration, you will see both `pageload` and `navigation` transactions in sentry.io.
```
imatwawana, line 1: Small fix to formatting of sentry.io so that it's a link

```suggestion
After configuration, you will see both `pageload` and `navigation` transactions in [sentry.io](https://sentry.io).
```

imatwawana (approved):
Thank you for fixing, Katie!

src/includes/performance/enable-automatic-instrumentation/javascript.mdx
lobsterkatie: Not really meant to be a link - more meant to be talking about it as a general idea. Lemme change to "the Sentry UI."

lobsterkatie, line 1:
```suggestion
After configuration, you will see both `pageload` and `navigation` transactions in the Sentry UI.
```

PR #3223, opened 2021-03-09
fix: push anchor headers below fixed top bar

dashed:
I've resolved the remaining issues that @rhcarvalho commented in https://github.com/getsentry/sentry-docs/pull/2993

Based on:

- https://stackoverflow.com/questions/10732690/offsetting-an-html-anchor-to-adjust-for-fixed-header
- https://css-tricks.com/hash-tag-links-padding/

Closes https://github.com/getsentry/sentry-docs/issues/2685

### DIFF
--- a/src/css/_includes/global-header.scss
+++ b/src/css/_includes/global-header.scss
@@ -43,11 +43,10 @@
   }
 }
 
-// https://stackoverflow.com/a/24298427
-[id]::before {
-  content: '';
-  display: block;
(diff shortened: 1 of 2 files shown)


PeloWriter:
Thanks for digging into this, @dashed. I used the preview to test this, and it looks like it's working.

dashed:
@PeloWriter thanks for checking. I'll merge this in.

PR #3233, opened 2021-03-12
Update doc for Symfony

endermctv:


src/platforms/php/guides/symfony/index.mdx
```diff
@@ -21,28 +21,25 @@ composer require sentry/sentry-symfony
 
 Add your DSN to `app/config/config.yml`:
 
-```yaml {filename:app/config/config.yml}
+```yaml {filename:config/config.yml}
 sentry:
-  dsn: "___PUBLIC_DSN___"
+  dsn: "%env(SENTRY_DSN)%"
```
rhcarvalho, line 26: We use the `___PUBLIC_DSN___ ` "magic keyword" in docs to automatically fill in a DSN for logged in Sentry.io users.

```suggestion
  dsn: "___PUBLIC_DSN___"
```

rhcarvalho (changes requested):
Hi @endermctv! Thanks for your contribution ❤️ .
I'm not a PHP expert, but trying to get this going. Would you mind sharing some context on the changes you're proposing?

In particular, the change of `filename` and order of the paragraphs (was there any content change as well?).

Thanks again!

src/platforms/php/guides/symfony/index.mdx
fridtjof: The `___PUBLIC_DSN___` keyword is still present, this PR moves it to the .env file that's used to hold secrets and other environment specific configuration. `SENTRY_DSN` is defined there.
`%env(SENTRY_DSN)` is the Symfony way of saying "insert the env variable `SENTRY_DSN` here"

fridtjof:
I randomly stumbled upon this PR, so figured I might just explain the changes I see here.

The rich diff provided [here](https://github.com/getsentry/sentry-docs/pull/3233/files?short_path=8743fa5#diff-8743fa5625b9b440ba12745b5b290d21b030486d6e76f32e354ed6dacca7e0ed) (the 'document' button on the top right of the diff in case it does not show that view by default) makes the changes a lot clearer than the file diff.

Two things were changed:

- DSN config: (answered in the review comment)
- Bundles: in modern Symfony projects, the list of Bundles (extensions to Symfony, basically) is no longer specified directly in the `Kernel.php`, but instead in a raw PHP file `config/bundles.php`, which simply returns an array of all bundles. This file is included by Symfony's base kernel by default.

src/platforms/php/guides/symfony/index.mdx
rhcarvalho: Thanks for chiming in @fridtjof. If that's the preferred way in Symfony, then we should align the content here and `src/wizard/php/symfony.md`. In the wizard there's a note:

> If you're using the Symfony Flex plugin, you'll find this file already created for you; it will suggest using an environment variable to inject the DSN value securely.

I wonder if the env var approach also works and the `.env` file is picked up automatically if not using the Flex plugin?

rhcarvalho:
@ste93cry @Jean85 could you please help me with this one? I can probably turn #3274 into an overhauling of the Symfony docs and incorporate the changes from here.

src/platforms/php/guides/symfony/index.mdx
ste93cry: >I wonder if the env var approach also works and the `.env` file is picked up automatically if not using the Flex plugin?

Flex is just a way to automate installation/uninstallation of packages, the functionality of loading the `.env` files is provided by the framework itself

rhcarvalho:
I've incorporated the changes in #3274 => https://github.com/getsentry/sentry-docs/pull/3274/commits/cf47e58a811215f88f84575ca30648868adf0864

PR #3249, opened 2021-03-15
fix: correct bug preventing manual platform selection with usePlatform

cameronmcefee:
The logic here will always evaluate to `valueFromLocation` making it impossible to manually pass a platform. Given the nature of the change, there's a high likelihood something is going to break. I'm not even sure how the site is functioning given that we use this all over the place.


cameronmcefee:
Based on the failed check, it seems like this is a pretty big break.

PR #3251, opened 2021-03-15
Feat: move other guides to sidebar

cameronmcefee:
At @PeloWriter's behest, I've moved "Other guides" from the left sidebar, to the right.

We had discussed also removing the list from the body of the guides, but that is a potentially dogmatic change that is easy enough for the content team to handle themselves if they would like to do it.

This PR is dependent on #3249 where I had to fix a bug to make `usePlatform` work.

JavaScript | Sentry Documentation 2021-03-15 16-27-42

### DIFF
--- a/src/components/basePage.tsx
+++ b/src/components/basePage.tsx
@@ -113,7 +113,9 @@ export default ({
               <Banner isModule={true} />
               <React.Fragment>
                 {prependToc}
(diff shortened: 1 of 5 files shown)


PeloWriter:
Thank you. This makes it easier to dive into the content that is associated with getting started without removing the value of the associated guides. 

Can we rename them "Related Guides" rather than "Other Guides"?

PR #3253, opened 2021-03-16
Update for 4.26 compilation

bulkhead-kev:
Add ".ToSharedRef()" to the first parameter in FJsonSerializer::Serialize


jan-auer (approved):
Thank you!

PR #3274, opened 2021-03-19
Symfony: update config to match latest releases

rhcarvalho:
See https://github.com/getsentry/sentry-symfony/blob/master/UPGRADE-4.0.md

This addresses https://github.com/getsentry/sentry-docs/pull/3137#pullrequestreview-615546347 and generally other changes to bring the Symfony docs into a better state.

Closes #3233

### DIFF
--- a/src/platforms/php/guides/symfony/config.mdx
+++ b/src/platforms/php/guides/symfony/config.mdx
@@ -7,17 +7,12 @@ In addition to the [configuration available in the PHP SDK](../configuration/opt
 
 All configuration for Symfony is done in `app/config/config.yml`.
 
-N.B. The `register_error_listener` and `monolog` configuration options are available since [sentry/sentry-symfony:^3.2.0](https://github.com/getsentry/sentry-symfony/releases/tag/3.2.0)
+N.B. The `register_error_listener` configuration option is available since [sentry/sentry-symfony:^3.2.0](https://github.com/getsentry/sentry-symfony/releases/tag/3.2.0).
 
 ```yaml {filename:app/config/config.yml}
 sentry:
   dsn: "___PUBLIC_DSN___"
   register_error_listener: true
-  monolog:
-    error_handler:
-      enabled: false
-      level: DEBUG
-      bubble: true
   options:
     attach_stacktrace: true
     before_breadcrumb: "@sentry.callback.before_breadcrumb"
@@ -54,6 +49,12 @@ sentry:
     request: 1
     subrequest: 1
     console: 1
+services:
+    Sentry\Monolog\Handler:
+        arguments:
+            $hub: '@Sentry\State\HubInterface'
+            $level: !php/const Monolog\Logger::ERROR
+            $bubble: false
 ```
 
 The DSN option is the only required option: it sets the Sentry DSN, and so reports all events to the related project. If it's
--- a/src/platforms/php/guides/symfony/config.mdx
+++ b/src/platforms/php/guides/symfony/config.mdx
@@ -59,7 +59,7 @@ All the possible configurations under the `options` key map directly to the corr
 you can read more about those in the [general configuration docs](../configuration/),
 and on the [PHP specific SDK docs](../config/).
 
-If you use Monolog, you can configure it by appending a `services` entry to your configuration file:
+If you use Monolog, you can use it to report events instead of the normal error listener approach. To do so, you must turn `register_error_listener` to `false` and configure our `Handler` as a service, like this:
 
 ```yaml {filename:config/packages/sentry.yaml}
 sentry:
--- a/src/platforms/php/guides/symfony/config.mdx
+++ b/src/platforms/php/guides/symfony/config.mdx
@@ -5,11 +5,11 @@ description: "Learn about Symfony-specific configuration options."
 
 In addition to the [configuration available in the PHP SDK](../configuration/options/), there are some Symfony-specific options.
 
-All configuration for Symfony is done in `app/config/config.yml`.
+All configuration for Symfony is done in `config/packages/sentry.yaml`.
 
 N.B. The `register_error_listener` configuration option is available since [sentry/sentry-symfony:^3.2.0](https://github.com/getsentry/sentry-symfony/releases/tag/3.2.0).
 
-```yaml {filename:app/config/config.yml}
+```yaml {filename:config/packages/sentry.yaml}
 sentry:
   dsn: "___PUBLIC_DSN___"
   register_error_listener: true
@@ -49,12 +49,6 @@ sentry:
     request: 1
     subrequest: 1
     console: 1
-services:
-    Sentry\Monolog\Handler:
-        arguments:
-            $hub: '@Sentry\State\HubInterface'
-            $level: !php/const Monolog\Logger::ERROR
-            $bubble: false
 ```
 
 The DSN option is the only required option: it sets the Sentry DSN, and so reports all events to the related project. If it's
@@ -64,6 +58,22 @@ disable it in the `test` and `dev` environments.
 All the possible configurations under the `options` key map directly to the correspondent options from the base SDK;
 you can read more about those in the [general configuration docs](../configuration/),
 and on the [PHP specific SDK docs](../config/).
+
+If you use Monolog, you can configure it by appending a `services` entry to your configuration file:
+
+```yaml {filename:config/packages/sentry.yaml}
+sentry:
+  dsn: "___PUBLIC_DSN___"
+  register_error_listener: false  # avoids double reporting
+  # ...
+
+services:
+  Sentry\Monolog\Handler:
+    arguments:
+      $hub: '@Sentry\State\HubInterface'
+      $level: !php/const Monolog\Logger::ERROR
+      $bubble: false
+```
 
 Below you can find additional documentation that is specific to the bundle usage, or information about the sensible default
 values that you can use in some cases.
--- a/src/platforms/php/guides/symfony/config.mdx
+++ b/src/platforms/php/guides/symfony/config.mdx
@@ -11,44 +11,44 @@ N.B. The `register_error_listener` configuration option is available since [sent
 
 ```yaml {filename:config/packages/sentry.yaml}
 sentry:
-  dsn: "___PUBLIC_DSN___"
-  register_error_listener: true
-  options:
-    attach_stacktrace: true
-    before_breadcrumb: "@sentry.callback.before_breadcrumb"
-    before_send: "@sentry.callback.before_send"
-    capture_silenced_errors: false
-    context_lines: 5
-    default_integrations: true
-    enable_compression: true
-    environment: "%kernel.environment%"
-    error_types: "E_ALL & ~E_NOTICE"
-    http_proxy: "10.0.0.12:3456"
-    in_app_exclude:
-      - "%kernel.cache_dir%"
-      - "%kernel.project_dir%/vendor"
-    integrations:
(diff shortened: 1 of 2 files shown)


rhcarvalho:
@Jean85 could you please have a look?

src/platforms/php/guides/symfony/config.mdx
```diff
@@ -7,17 +7,12 @@ In addition to the [configuration available in the PHP SDK](../configuration/opt
 
 All configuration for Symfony is done in `app/config/config.yml`.
```
Jean85, line 8: This is wrong, it's a remainder from Symfony 2.

If user uses Symfony Flex, it will start from the recipe: https://github.com/symfony/recipes-contrib/blob/master/sentry/sentry-symfony/3.0/config/packages/sentry.yaml
So it's in `config/packages/sentry.yaml

src/platforms/php/guides/symfony/config.mdx
```diff
@@ -54,6 +49,12 @@ sentry:
     request: 1
     subrequest: 1
     console: 1
+services:
```
Jean85, line 52: I would add a small comment here to say that this piece is suggested but not required. It makes duplicate reporting of events if `register_error_listener` is still set to `true`.

Jean85 (changes requested):

src/platforms/php/guides/symfony/config.mdx
apatkinson: Agreed, but it should be underneath where all the other options are being explained.

Also mark this as being in `config/services.yml` file. There are all levels of developers out there and spelling it out is sometimes required and makes for a much less frustrating experience for everyone.

alexander-schranz: would not say this need to live in the `config/services.yaml` its okay to be in the `config/packages/sentry.yaml` because it is part of the sentry configuration. Its not common that bundles the define the service definitions for there package in there config file. So you actually see that this service definition is only here because of sentry and not part of your application service definitions.

https://github.com/symfony/recipes-contrib/search?q=services

src/platforms/php/guides/symfony/config.mdx
rhcarvalho: Thanks, I've updated the references in this file.

src/platforms/php/guides/symfony/config.mdx
rhcarvalho: Okay, thanks for pointing it out. I moved this to a separate paragraph, let me know if it is accurate now.

src/platforms/php/guides/symfony/config.mdx
```diff
@@ -64,6 +59,22 @@ All the possible configurations under the `options` key map directly to the corr
 you can read more about those in the [general configuration docs](../configuration/),
 and on the [PHP specific SDK docs](../config/).
 
+If you use Monolog, you can configure it by appending a `services` entry to your configuration file:
```
Jean85, line 62: I would explain this piece with a bit more detail:
```suggestion
If you use Monolog, you can use it to report events instead of the normal error listener approach. To do so, you must turn `register_error_listener` to `false` and configure our `Handler` as a service, like this:
```

Also, @ste93cry are you sure that this is all it's needed? Shouldn't we hook the handler somehow? With either `autoconfigure: true` or by changing Monolog's config?

Jean85 (changes requested):

src/platforms/php/guides/symfony/config.mdx
ste93cry: >are you sure that this is all it's needed?

Yes, I know there is some work going on in the [`symfony/monolog-bundle`](https://github.com/symfony/monolog-bundle) repository to support out-of-the-box the registration of our handler, but being a WIP work I don't know when it will land

Shouldn't we hook the handler somehow? With either `autoconfigure: true` or by changing Monolog's config?

We could register the handler ourselves in the DI container so that users don't have to, if the service is `private: true` it will be dropped during the compilation phase if it's not used and so users will not have one more unused service in their service container. However, I don't think it's worth doing it because:

- The Monolog configuration is too app-specific and we cannot hook into it because the ordering matters: in fact, even Symfony does not allow merging multiple `monolog` configuration nodes
- The handler has a few options which can be tuned (`$level` and `$bubble`), and if we register the service and the user wants to customize those parameters they will have to anyway override the service definition. Usually these parameters are what you set when registering an handler in the Monolog configuration, so the best DX experience we could provide is to support using our handler directly in the Monolog bundle

Jean85: @ste93cry sorry I posed the wrong question: I was asking if, in the context of this doc, we should instruct the user to do something else or if just by registering the service it will be hooked into Monolog automatically.

ste93cry: Registering the service is not enough, you then need to use it in the configuration of the Monolog bundle. However, how and where you use it is dependant on the user configuration and what he wants to do. Maybe could we link to the Monolog documentation for this part?

src/platforms/php/guides/symfony/config.mdx
```diff
@@ -64,6 +59,22 @@ All the possible configurations under the `options` key map directly to the corr
 you can read more about those in the [general configuration docs](../configuration/),
 and on the [PHP specific SDK docs](../config/).
 
+If you use Monolog, you can use it to report events instead of the normal error listener approach. To do so, you must turn `register_error_listener` to `false` and configure our `Handler` as a service, like this:
+
+```yaml {filename:config/packages/sentry.yaml}
+sentry:
```
alexander-schranz, line 65: think we should here again going on the symfony coding standard and use `4 spaces` in yaml files like we did it in the other doc

alexander-schranz (changes requested):
Small coding style change request

ste93cry:
@rhcarvalho I just noticed that you we have two places where we talk about Monolog, basically saying the same things:

https://sentry-docs-git-rhcarvalho-monolog-config-options.sentry.dev/platforms/php/guides/symfony/config/
https://sentry-docs-git-rhcarvalho-monolog-config-options.sentry.dev/platforms/php/guides/symfony/#monolog-integration

I think this is quite confusing, and it will make harder to update the docs. I would just leave the paragraph in the `Getting Started` section

src/platforms/php/guides/symfony/config.mdx
Jean85: We definitely need to tell the user that adding the handler to the Monolog config is required.

alexander-schranz:
@ste93cry I aggree here, would also remove monolog from the config docs, as there not longer exist a monolog config in the bundle itself and think the getting started is enough.

src/platforms/php/guides/symfony/config.mdx
rhcarvalho: Done

rhcarvalho:
I've applied all suggestions and removed the duplicated Monolog instructions.
Ready for a final pass @ste93cry @Jean85 @alexander-schranz -- thank you all for your input!

alexander-schranz (approved):

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -37,7 +37,15 @@ Add your DSN to `config/packages/sentry.yaml`:
...
 ```yaml {filename:config/packages/sentry.yaml}
 sentry:
-    dsn: "___PUBLIC_DSN___"
+    dsn: "%env(SENTRY_DSN)%"
+```
+
+And in your `.env` file:
+
+```env {filename:.env}
+###> sentry/sentry-symfony ###
```
ste93cry, line 46: If you set the DSN in this file it will be loaded for all environments, which is wrong for obvious reasons. We should instead tell users that they have to set this environment variable in their own shell

ste93cry (changes requested):

src/platforms/php/guides/symfony/index.mdx
alexander-schranz: I would go with the uncommited `.env.local` file here which is used by the devs. The real ENV is only when you use a docker setup or some cloud hosting where you are able to set a global ENV.

ste93cry: That's also a solution of course, but it must be clear that having a DSN set locally will enable the bundle, and setting it in that file won't work for production 😉

alexander-schranz: So I would go witht something like this here then:

~~~
The `SENTRY_DSN` need to be defined as ENVIRONMENT variable or you can be set it in your uncommitet `.env.local` file:


```env {filename:.env.local}
SENTRY_DSN="___PUBLIC_DSN___"
```
~~~

ste93cry: The point is: once you set it in `.env.local`, it will work only for the local environment (at least according to the canonical meaning of that file), so I see little value in suggesting people to set it like that because you usually use Sentry just for the production environment. Instead, if it's set in that file, it will again be enabled for all environments on your own machine, which makes no sense to me

alexander-schranz: Sure thats clear, but think as Symfony dev you should know that as you also need to set `APP_ENV=prod` and your database connection as env variable or in the `.env.local` file. The thing is that most still use some kind of shared servers for hosting there projects and so there the `.env.local` file is more common as you can not set a global env variable. Also devs are more common friendly with the `.env.local` file in symfony then how to set a ENV variable in there webserver, sound strange but my experience was it that most still just use `.env.local` file to configure APP_ENV, DATABASE_URL, ... in symfony.

rhcarvalho: Hmm this is one of the suggestions coming from #3233. Prior to that we were setting the DSN (___PUBLIC_DSN___) straight in the config file.

What ya'll think is best here then?

ste93cry: >Prior to that we were setting the DSN (`___PUBLIC_DSN___`) straight in the config file

Also that was incorrect then. I think that setting things using environment variables is now considered a best practice in the Symfony world since a lot of years, so we should definitely use them

> The thing is that most still use some kind of shared servers for hosting there projects and so there the .env.local file is more common as you can not set a global env variable

I don't have any data to say this is not true, although I strongly believe that Symfony doesn't even run on a shared hosting due to performance issues :sweat_smile: Anyway, `.env.prod.local` should be the right file to use so that it's still a local file not committed to the repo and it works only in production. WDYT?

rhcarvalho: Considering that this `.env` solution is essentially the same as setting a DSN directly in config as we document today (see below), I'll err on the side of iterating quick and merging this in like this. Please send a new PR if you think this is confusing and there's a better way.


---

https://docs.sentry.io/platforms/php/guides/symfony/#configure

image

rhcarvalho: We don't go to this level of detail in any framework docs that I'm aware of, and where the same dev/staging/prod considerations exist.

For example:
- https://docs.sentry.io/platforms/python/guides/django/#configure
- https://docs.sentry.io/platforms/python/guides/flask/#configure
- https://docs.sentry.io/platforms/java/guides/spring-boot/#configure

In the end of the day, developers and teams know better how their per-environment configuration is supposed to be handled, and there is no one right way to do it.

That said, I wouldn't oppose any contributions that improve the Symfony docs to include more specific advice, it is just that I'm not the best person to drive that effort.

PR #3278, opened 2021-03-19
Java: mention that returning null from sampling callback defaults to configured value

maciejwalkowiak:


src/includes/performance/configure-sample-rate/java.mdx
```diff
@@ -16,7 +16,7 @@ Sentry.init(options -> {
   // OR if you prefer, determine traces sample rate based on the sampling context
   options.setTracesSampler(
       context -> {
-        // return a number between 0 and 1
+        // return a number between 0 and 1 or null (to fallback to configured value)
```
marandaneto, line 19: https://github.com/getsentry/sentry-docs/blob/master/src/includes/performance/configure-sample-rate/android.mdx

marandaneto:
@maciejwalkowiak always have a look at the Android docs too when touching the performance API which is also used by the Android package, thanks.

maciejwalkowiak:
@marandaneto have another look please

marandaneto (approved):

PR #3244, opened 2021-03-15
The user in Dart/Flutter was renamed to SentryUser

ueman:
See https://github.com/getsentry/sentry-dart/pull/357 and https://github.com/getsentry/sentry-docs/issues/3196

PR #3291, opened 2021-03-22
Fix the set-user scope example in Dart

timekone:
This is how the set-user scope example looks right now:

```dart
import 'package:sentry/sentry.dart';

Sentry.configureScope(
  (scope) => scope.user = User(id: 1234, email: 'jane.doe@example.com'),
);
```

But `User.id` is a String, so there are missing quotation marks around `1234`.
How it should look:
```dart
import 'package:sentry/sentry.dart';

Sentry.configureScope(
  (scope) => scope.user = User(id: "1234", email: 'jane.doe@example.com'),
);
```

 #3244 may be related.

### DIFF
--- a/src/includes/set-user/dart.mdx
+++ b/src/includes/set-user/dart.mdx
@@ -2,6 +2,6 @@
 import 'package:sentry/sentry.dart';
 
 Sentry.configureScope(
-  (scope) => scope.user = User(id: "1234", email: 'jane.doe@example.com'),
+  (scope) => scope.user = User(id: '1234', email: 'jane.doe@example.com'),
 );
 ```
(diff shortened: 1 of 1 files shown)


src/includes/set-user/dart.mdx
```diff
@@ -2,6 +2,6 @@
 import 'package:sentry/sentry.dart';
 
 Sentry.configureScope(
-  (scope) => scope.user = User(id: 1234, email: 'jane.doe@example.com'),
+  (scope) => scope.user = User(id: "1234", email: 'jane.doe@example.com'),
```
marandaneto, line 5:
```suggestion
  (scope) => scope.user = User(id: '1234', email: 'jane.doe@example.com'),
```

marandaneto:
@timekone good catch, also fixed on https://github.com/getsentry/sentry-docs/pull/3244 thanks for that

marandaneto (approved):

PR #3295, opened 2021-03-23
Fix: Flutter split debug info and obfuscate

marandaneto:
`obfuscate` cannot be used without `split-debug-info` so a `or` could mislead the user


philipphofmann (approved):
🥇 best PR ever. LGTM 😁

Issue #2879, opened 2021-01-13
Document MaxCacheItems for Apple, Java and .NET


philipphofmann:
Add docs for [`MaxCacheItems`](https://github.com/getsentry/sentry-java/blob/81594398ee623b40eb9df1a847dbe21d9e4c25ef/sentry/src/main/java/io/sentry/SentryOptions.java#L108) for Java, Apple, and .NET, when these issues are closed https://github.com/getsentry/sentry-cocoa/issues/899, https://github.com/getsentry/sentry-dotnet/issues/736.

philipphofmann:
Only enable this for Apple once 7.0.0 GA is released. I added this to https://github.com/getsentry/sentry-docs/issues/3307 so we can close this issue.

Issue #3307, opened 2021-03-24
Apple 7.0.0 GA Docs


philipphofmann:
## Issue

This issue is a summary of things to update once Apple 7.0.0 is released.

## Description

- [ ] Enable MaxCacheItems in Basic Options
- [x] Add migration guide
- [x] Remove notes on 7.0.0-alpha.x or 7.0.0-beta.x

### Suggested Solution

Update the docs once Apple 7.0.0 GA is released.

PR #3308, opened 2021-03-24
feat: Add MaxCacheItems to basic options

philipphofmann:
We are going to enable Apple once 7.0.0 GA is released. I added this to https://github.com/getsentry/sentry-docs/issues/3307 to not forget about it.

Fixes GH-2879

### DIFF
--- a/src/platforms/common/configuration/options.mdx
+++ b/src/platforms/common/configuration/options.mdx
@@ -96,6 +96,12 @@ This variable controls the total amount of breadcrumbs that should be captured. 
 
 </ConfigKey>
 
+<ConfigKey name="max-cache-items" supported={[ "dotnet", "android"]} >
+
+The maximum number of [envelopes](https://develop.sentry.dev/sdk/envelopes/) to keep in cache. The SDKs use envelopes to send data, as events, attachments, user feedback, session, etc. to Sentry. An envelope can contain multiple items - for example, an event with a session and two attachments. Depending on the usage of the SDK, the size of an envelope can differ. When the number of envelopes in the local cache exceeds max-cache-items, the SDK deletes the oldest envelope. When deleting an envelope, we migrate sessions to the next envelope to not mess up your release health stats. The default is `30`.
+
+</ConfigKey>
+
 <ConfigKey name="attach-stacktrace">
 
 When enabled, stack traces are automatically attached to all messages logged. Stack traces are always attached to exceptions; however, when this option is set, stack traces are also sent with messages. This option, for instance, means that stack traces appear next to all log messages.
--- a/src/platforms/common/configuration/options.mdx
+++ b/src/platforms/common/configuration/options.mdx
@@ -96,7 +96,7 @@ This variable controls the total amount of breadcrumbs that should be captured. 
 
 </ConfigKey>
 
-<ConfigKey name="max-cache-items" supported={[ "dotnet", "android"]} >
+<ConfigKey name="max-cache-items" supported={[ "dotnet" ]} >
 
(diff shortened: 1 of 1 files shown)


src/platforms/common/configuration/options.mdx
```diff
@@ -96,6 +96,12 @@ This variable controls the total amount of breadcrumbs that should be captured.
 
 </ConfigKey>
 
+<ConfigKey name="max-cache-items" supported={[ "dotnet", "android"]} >
```
marandaneto, line 99: it's called `cacheDirSize` on Android though

src/platforms/common/configuration/options.mdx
```diff
@@ -96,6 +96,12 @@ This variable controls the total amount of breadcrumbs that should be captured.
...
+
+The maximum number of [envelopes](https://develop.sentry.dev/sdk/envelopes/) to keep in cache. The SDKs use envelopes to send data, as events, attachments, user feedback, session, etc. to Sentry. An envelope can contain multiple items - for example, an event with a session and two attachments. Depending on the usage of the SDK, the size of an envelope can differ. When the number of envelopes in the local cache exceeds max-cache-items, the SDK deletes the oldest envelope. When deleting an envelope, we migrate sessions to the next envelope to not mess up your release health stats. The default is `30`.
```
marandaneto, line 101:
```suggestion
The maximum number of [envelopes](https://develop.sentry.dev/sdk/envelopes/) to keep in cache. The SDKs use envelopes to send data, as events, attachments, user feedback, session, etc. to Sentry. An envelope can contain multiple items - for example, an event with a session and two attachments. Depending on the usage of the SDK, the size of an envelope can differ. When the number of envelopes in the local cache exceeds max-cache-items, the SDK deletes the oldest envelope. The default is `30`.
```

I think this is implementation details, if we want to explain it, it'd be better to release health docs

src/platforms/common/configuration/options.mdx
marandaneto: remove `android` for now please, till we get https://github.com/getsentry/sentry-java/issues/1162 done

marandaneto, line 99:
```suggestion
<ConfigKey name="max-cache-items" supported={[ "dotnet"]} >
```

philipphofmann: Ah didn't know, that this is not done yet. Thanks

src/platforms/common/configuration/options.mdx
philipphofmann: I added it because people might think that this messes up their session stats. Just wanted to clarify that this is not affected. @bruno-garcia, what do you think? Should we keep it or remove it?

src/platforms/common/configuration/options.mdx
```diff
@@ -96,6 +96,12 @@ This variable controls the total amount of breadcrumbs that should be captured.
 
 </ConfigKey>
 
+<ConfigKey name="max-cache-items" supported={[ "dotnet" ]} >
+
+The maximum number of [envelopes](https://develop.sentry.dev/sdk/envelopes/) to keep in cache. The SDKs use envelopes to send data, as events, attachments, user feedback, session, etc. to Sentry. An envelope can contain multiple items - for example, an event with a session and two attachments. Depending on the usage of the SDK, the size of an envelope can differ. When the number of envelopes in the local cache exceeds max-cache-items, the SDK deletes the oldest envelope. When deleting an envelope, we migrate sessions to the next envelope to not mess up your release health stats. The default is `30`.
```
PeloWriter, line 101: How about this - because it is an implementation detail, but there's value in being clear about maintaining the integrity of session stats:
```suggestion
The maximum number of [envelopes](https://develop.sentry.dev/sdk/envelopes/) to keep in cache. The SDKs use envelopes to send data, such as events, attachments, user feedback, and sessions to sentry.io. An envelope can contain multiple items, such as an event with a session and two attachments. Depending on the usage of the SDK, the size of an envelope can differ. If the number of envelopes in the local cache exceeds `max-cache-items`, the SDK deletes the oldest envelope and migrates the sessions to the next envelope to maintain the integrity of your release health stats. The default is `30`.
```

PeloWriter (approved):

PR #3318, opened 2021-03-25
Improve Timber docs

marandaneto:
added missing imports
Kotlin snippets first
Verify actually captures the exception using the Timber API

### DIFF
--- a/src/includes/getting-started-config/android.timber.mdx
+++ b/src/includes/getting-started-config/android.timber.mdx
@@ -1,21 +1,9 @@
-```java
-SentryAndroid.init(this, options -> {
-   if (!BuildConfig.DEBUG) {
-
-     // default values:
-     // minEventLevel = ERROR
-     // minBreadcrumbLevel = INFO
-     options.addIntegration(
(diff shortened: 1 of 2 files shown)


src/includes/getting-started-config/android.timber.mdx
```diff
@@ -28,6 +16,32 @@ SentryAndroid.init(this) { options ->
...
 ```
+
+```java
+import io.sentry.android.timber.SentryTimberIntegration;
+import timber.log.Timber
+// import BuildConfig
+
+SentryAndroid.init(this, options -> {
+   if (!BuildConfig.DEBUG) {
```
philipphofmann, line 32: Just for me to understand: Why only when not debug? Wouldn't it be nice to see the events in Sentry?

marandaneto: then you have `Timber.DebugTree`, some people don't want to report to Sentry debug levels, as they can see in the console, otherwise they pay for what they don't need to.

philipphofmann (approved):
LGTM

PR #3323, opened 2021-03-26
Fix sentry-native capture exception with values list

marandaneto:
https://github.com/getsentry/sentry-dart/issues/386

Android NDK requires the exception to be a `values` list, the protocol is flexible enough but we don't parse it like this on Android (relying on Gson).


Swatinem (approved):

jan-auer (approved):
Please do fix your parser on the Android side, though. Since the protocol allows it, users could create custom payloads and expect them to work.

marandaneto:
> Please do fix your parser on the Android side, though. Since the protocol allows it, users could create custom payloads and expect them to work.

we can't unless we get rid of Gson/reflection, so it'd be a trivial task, but it's in our TODO.

PR #3325, opened 2021-03-26
ref(tags): Mention the maximum tag limits

wmak:
- As far as i can tell we don't mention these limits anywhere publicly, so the only way to notice this is to accidentally hit these limits
- Source of limits: https://github.com/getsentry/relay/blob/10f2c93efb8b5a200168a8fb961c3b136c1e5f18/relay-general/src/processor/attrs.rs#L122-L123


PeloWriter (approved):
This is great information to surface here. Thank you.

PR #3328, opened 2021-03-27
feat: add "other" wizard

alxu-sentry:
Improve the "other" onboarding process in Sentry:

 

Requires https://github.com/getsentry/sentry/pull/24772 to be useful.

This could of course be special-cased in sentry, but I think it makes more sense to put it here in part because it's the easiest way to get the DSN filled in properly.


alxu-sentry:
I'm not sure how much there is to say here, but I think showing the DSN is definitely useful for any community client.

PR #3332, opened 2021-03-29
fix broken redirects in Issues pages

imatwawana:
Fixing broken redirects on Issues index page and Reprocessing page


imatwawana:
Tested redirects in Preview

PR #3335, opened 2021-03-30
delayed_job is supported now

pwim:


imatwawana:
Thanks for catching this!

PR #3355, opened 2021-04-06
fix: misspelling of SentrySDK in some code snippets.

brustolin:
SentrySDK is misspelled as SentrySdk in some apple snippets.


bruno-garcia (approved):
Thanks for fixing this!

PR #3356, opened 2021-04-06
Java: add Java guide for Apache HttpClient based transport.

### LINKED ISSUE #3333: Java Server Performance docs: Point users to Apache transport
There's no reference to our Apache HTTP transport from the Java performance docs:

https://docs.sentry.io/platforms/java/configuration/transport/

We need to make sure to point users to this if they expect to instrument high throughput server apps with Java.

/cc @maciejwalkowiak

maciejwalkowiak:
Fixes #3333

### DIFF
--- a/src/includes/performance/configure-sample-rate/java.mdx
+++ b/src/includes/performance/configure-sample-rate/java.mdx
@@ -35,3 +35,49 @@ Sentry.init { options ->
   }
 }
 ```
+
+## High-Throughput Transport
+
+For high traffic applications, we recommend using high-throughput HTTP transport based on [Apache HTTP Client 5](https://hc.apache.org/httpcomponents-client-5.0.x/index.html).
+
+```xml {tabTitle:Maven}
+<dependency>
+    <groupId>io.sentry</groupId>
+    <artifactId>sentry-apache-http-client-5</artifactId>
+    <version>{{ packages.version('sentry.java.apache-http-client-5', '4.2.0') }}</version>
+</dependency>
+```
+
+```groovy {tabTitle:Gradle}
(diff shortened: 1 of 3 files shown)


maciejwalkowiak:
cc @marandaneto @bruno-garcia

bruno-garcia (approved):
LGTM but we probably should merge once the dep bump goes out to central

src/includes/performance/configure-sample-rate/java.mdx
```diff
@@ -35,3 +35,49 @@ Sentry.init { options ->
   }
 }
 ```
+
+## High-Throughput Transport
```
marandaneto, line 39: why don't we extract this to an includes file? the same content is within 3 files.

Also, I'm not entirely sure this lib is compatible with Android, so it should not be visible for Android, and right now it is

marandaneto:
> LGTM but we probably should merge once the dep bump goes out to central

also, let's address my comment about Android at least.

marandaneto:
moving to Draft till cutting a release

src/includes/performance/configure-sample-rate/java.mdx
maciejwalkowiak: In each of the guides this section is slightly different as there is a different way to setup transport in plain Java, Spring and Spring Boot.

maciejwalkowiak: Android has a separate file for configuring sample rate, so this section does not appear there. Or you meant something else?

marandaneto: I (almost) swear I've seen it on https://sentry-docs-git-fork-maciejwalkowiak-gh-java-3333.sentry.dev/platforms/android/performance/
but you're right, my bad, it's Friday!

marandaneto (approved):
LGTM

PR #3363, opened 2021-04-07
.NET: document distributed tracing

### LINKED ISSUE #3338: Expand Document distributed tracing for .NET
## Issue

Expand Document distributed tracing for .NET

## Description

Related to https://github.com/getsentry/sentry-docs/pull/3324

### Suggested Solution

Expand the content with .NET code snippets etc

Tyrrrz:
Closes #3338

cc @marandaneto @bruno-garcia

### DIFF
--- a/src/includes/performance/distributed-tracing/dotnet.mdx
+++ b/src/includes/performance/distributed-tracing/dotnet.mdx
@@ -16,7 +16,7 @@ request.Headers.Add("sentry-trace", sentryTraceHeader.ToString());
 
 <Note>
 
-Sentry also provides `SentryHttpMessageHandler`, which can be used to [automatically attach trace headers to outgoing HTTP requests](platforms/dotnet/performance/included-instrumentation/).
+Sentry also provides `SentryHttpMessageHandler`, which can be used to [automatically attach trace headers to outgoing HTTP requests](/platforms/dotnet/performance/included-instrumentation/).
 
 </Note>
 
@@ -46,6 +46,6 @@ if (httpContext.Request.Headers.TryGetValue("sentry-trace", out var sentryTraceH
 
 <Note>
 
-In Sentry's ASP.NET Core integration, trace headers are [handled automatically by the tracing middleware](platforms/dotnet/performance/included-instrumentation/).
+In Sentry's ASP.NET Core integration, trace headers are [handled automatically by the tracing middleware](/platforms/dotnet/performance/included-instrumentation/).
 
 </Note>
--- a/src/includes/performance/distributed-tracing/dotnet.mdx
+++ b/src/includes/performance/distributed-tracing/dotnet.mdx
@@ -5,6 +5,10 @@ Traces can bridge across multiple software services. Each span in a trace can be
 To obtain a trace header from the span, use `ISpan.GetTraceHeader()` method, then pass it to the downstream service. If the communication happens over HTTP, it is recommended that you set the value to the `sentry-trace` HTTP header.
 
 ```csharp
+using System;
+using System.Net.Http;
+using Sentry;
+
 var transaction = SentrySdk.StartTransaction("name", "op");
 var sentryTraceHeader = transaction.GetTraceHeader();
 
--- a/src/includes/performance/distributed-tracing/dotnet.mdx
+++ b/src/includes/performance/distributed-tracing/dotnet.mdx
@@ -2,7 +2,7 @@ To obtain a trace header from the span, use `ISpan.GetTraceHeader()` method, the
 
 <Note>
 
-For HTTP communications, Sentry's SDK providers `SentryHttpMessageHandler`, which can be used to [automatically attach trace headers to outgoing HTTP requests](/platforms/dotnet/performance/included-instrumentation/).
+For HTTP communications, Sentry's SDK provides `SentryHttpMessageHandler`, which can be used to [automatically attach trace headers to outgoing HTTP requests](/platforms/dotnet/performance/included-instrumentation/).
 
 </Note>
 
--- a/src/includes/performance/distributed-tracing/dotnet.mdx
+++ b/src/includes/performance/distributed-tracing/dotnet.mdx
@@ -38,6 +38,8 @@ if (httpContext.Request.Headers.TryGetValue("sentry-trace", out var sentryTraceH
     catch (FormatException e)
     {
         // Handle invalid trace header
+        e.Data["bad-sentry-header"] = sentryTraceHeaderValue;
+        SentrySdk.CaptureException(e);
     }
 }
 ```
--- a/src/includes/performance/distributed-tracing/dotnet.mdx
+++ b/src/includes/performance/distributed-tracing/dotnet.mdx
@@ -1,4 +1,12 @@
-To obtain a trace header from the span, use `ISpan.GetTraceHeader()` method, then pass it to the downstream service. If the communication happens over HTTP, it is recommended that you set the value to the `sentry-trace` HTTP header.
+To obtain a trace header from the span, use `ISpan.GetTraceHeader()` method, then pass it to the downstream service.
+
+<Note>
+
+For HTTP communications, Sentry's SDK providers `SentryHttpMessageHandler`, which can be used to [automatically attach trace headers to outgoing HTTP requests](/platforms/dotnet/performance/included-instrumentation/).
+
+</Note>
+
+The code snippet below demonstrates propagating the trace programatically. Even though `SentryHttpMessageHandler` does that automatically for your, this illustrates the mechanism so it can be replicated in other protocols and technologies such a messaging systems.
 
 ```csharp
 using System;
@@ -12,11 +20,6 @@ using var request = new HttpRequestMessage(HttpMethod.Get, "https://example.com"
 request.Headers.Add("sentry-trace", sentryTraceHeader.ToString());
 ```
 
-<Note>
-
-Sentry also provides `SentryHttpMessageHandler`, which can be used to [automatically attach trace headers to outgoing HTTP requests](/platforms/dotnet/performance/included-instrumentation/).
-
-</Note>
 
 To create a span as a continuation of the trace retrieved from the upstream service, pass the `sentry-trace` header value to the `StartTransaction(...)` method:
 
--- a/src/includes/performance/distributed-tracing/dotnet.mdx
+++ b/src/includes/performance/distributed-tracing/dotnet.mdx
@@ -2,7 +2,7 @@
 
 Traces can bridge across multiple software services. Each span in a trace can be represented as a `sentry-trace` header, containing the trace id, span id, and sampling details. This `sentry-trace` header can be passed to downstream services so that they can create spans that are a continuation of the trace started in the originating service.
 
-To obtain a trace header from the span, use `ISpan.GetTraceHeader()` method. Then pass it to the downstream service. If the communication happens over HTTP, it is recommended that you set the value to the `sentry-trace` HTTP header.
+To obtain a trace header from the span, use `ISpan.GetTraceHeader()` method, then pass it to the downstream service. If the communication happens over HTTP, it is recommended that you set the value to the `sentry-trace` HTTP header.
 
 ```csharp
 var transaction = SentrySdk.StartTransaction("name", "op");
@@ -18,7 +18,7 @@ Sentry also provides `SentryHttpMessageHandler`, which can be used to automatica
 
 </Note>
 
-To create a span as a continuation of the trace retrieved from the upstream service, pass the `sentry-trace` header value to the transaction context:
+To create a span as a continuation of the trace retrieved from the upstream service, pass the `sentry-trace` header value to the `StartTransaction(...)` method:
 
 ```csharp
 using System;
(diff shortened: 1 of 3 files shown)


src/includes/performance/distributed-tracing/dotnet.mdx
```diff
@@ -0,0 +1,45 @@
+## Distributed Tracing
+
+Traces can bridge across multiple software services. Each span in a trace can be represented as a `sentry-trace` header, containing the trace id, span id, and sampling details. This `sentry-trace` header can be passed to downstream services so that they can create spans that are a continuation of the trace started in the originating service.
+
+To obtain a trace header from the span, use `ISpan.GetTraceHeader()` method, then pass it to the downstream service. If the communication happens over HTTP, it is recommended that you set the value to the `sentry-trace` HTTP header.
```
marandaneto, line 5: turn out that this text is pretty similar to https://docs.sentry.io/platforms/java/performance/custom-instrumentation/#distributed-tracing (java.mdx), maybe we could actually move the generic text to the common file and only extract the important bits which is the method name.
wdyt @PeloWriter ?

src/includes/performance/distributed-tracing/dotnet.mdx
```diff
@@ -0,0 +1,45 @@
...
+var transaction = SentrySdk.StartTransaction("name", "op");
+var sentryTraceHeader = transaction.GetTraceHeader();
+
+using var request = new HttpRequestMessage(HttpMethod.Get, "https://example.com");
+request.Headers.Add("sentry-trace", sentryTraceHeader.ToString());
+```
+
+<Note>
+
+Sentry also provides `SentryHttpMessageHandler`, which can be used to automatically attach trace headers to outgoing HTTP requests. Learn more about it in <PlatformLink to="/performance/included-instrumentation/">this article</PlatformLink>.
```
marandaneto, line 17: iirc `PlatformLink` is used for something else, not really just linking the user to a different page, correct me if I am wrong @PeloWriter

src/includes/performance/distributed-tracing/dotnet.mdx
```diff
@@ -0,0 +1,45 @@
+## Distributed Tracing
+
+Traces can bridge across multiple software services. Each span in a trace can be represented as a `sentry-trace` header, containing the trace id, span id, and sampling details. This `sentry-trace` header can be passed to downstream services so that they can create spans that are a continuation of the trace started in the originating service.
+
+To obtain a trace header from the span, use `ISpan.GetTraceHeader()` method, then pass it to the downstream service. If the communication happens over HTTP, it is recommended that you set the value to the `sentry-trace` HTTP header.
+
+```csharp
+var transaction = SentrySdk.StartTransaction("name", "op");
```
marandaneto, line 8: on other SDKs, we always add the import of the classes, so users can copy/paste the whole code snippet.

https://develop.sentry.dev/sdk/philosophy/#assume-novice-developers

src/includes/performance/distributed-tracing/dotnet.mdx
```diff
@@ -0,0 +1,45 @@
+## Distributed Tracing
```
marandaneto, line 1: `distributed-tracing` is under a `PlatformSection` section with only `java` as supported, either we move it outside of Java or add .NET as a supported platform, you cannot see this page right now.

https://raw.githubusercontent.com/getsentry/sentry-docs/master/src/platforms/common/performance/custom-instrumentation.mdx

Tyrrrz: Thanks, I added a separate section with `supported={["dotnet"]}` because there's only one page we currently support.

src/includes/performance/distributed-tracing/dotnet.mdx
PeloWriter: I think that would work.
- Add the content from line 3 "Traces can bridge..." to line 45 of `/common/performance/custom-instrumentation.mdx`
- Add .NET to the `<PlatformSection supported>` listed on line 42 of that same file

src/includes/performance/distributed-tracing/dotnet.mdx
```diff
@@ -0,0 +1,49 @@
...
+}
+catch (FormatException e)
+{
+    // Handle invalid trace header
+}
+```
+
+<Note>
+
+In Sentry's ASP.NET Core integration, trace headers are handled automatically by the tracing middleware. Learn more about it in <PlatformLink to="/performance/included-instrumentation/">this article</PlatformLink>.
```
PeloWriter, line 47: see previous note on linking to a different page

src/platforms/common/performance/custom-instrumentation.mdx
```diff
@@ -48,3 +48,9 @@ To instrument certain regions of your code, you can create transactions to captu
 <PlatformContent includePath="performance/distributed-tracing" />
 
 </PlatformSection>
+
+<PlatformSection supported={["dotnet"]}>
```
PeloWriter, line 52: ah, I see you added this already.

PeloWriter (approved):

src/includes/performance/distributed-tracing/dotnet.mdx
bruno-garcia: I thought the approach from @Tyrrrz was correct. Wouldn't `platforms/dotnet/performance/included-instrumentation/` navigate the user out of one of the guides?

If the user is browsing this content through `/platforms/dotnet/guides/aspnetcore/performance/...` wouldn't we take them out of it?

src/includes/performance/distributed-tracing/dotnet.mdx
```diff
@@ -0,0 +1,49 @@
...
+</Note>
+
+To create a span as a continuation of the trace retrieved from the upstream service, pass the `sentry-trace` header value to the `StartTransaction(...)` method:
+
+```csharp
+using System;
+using Sentry;
+
+var httpContext = ...;
+var sentryTraceHeaderValue = httpContext.Request.Headers["sentry-trace"].ToString();
```
bruno-garcia, line 32: if a request coming in without Sentry's header, this would throw.
I rather we avoid such code snippet. Can we do something else like?:

```suggestion
if (!httpContext.Request.Headers.TryGetValue("sentry-trace", out var sentryTraceHeaderValue))
{
    return; // No Sentry header in this request
}
```
To be clear users can't assume such header will always be there?

src/includes/performance/distributed-tracing/dotnet.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+try
+{
+    var sentryTraceHeader = SentryTraceHeader.Parse(sentryTraceHeaderValue);
+    var transaction = SentrySdk.StartTransaction("name", "op", sentryTraceHeader);
+}
+catch (FormatException e)
+{
+    // Handle invalid trace header
```
bruno-garcia, line 41: Good opportunity to show them how to use Sentry:

```suggestion
    // Handle invalid trace header
    e.Data["bad-sentry-header"] = sentryTraceHeaderValue;
    SentrySdk.CaptureException(e);
```

src/platforms/common/performance/custom-instrumentation.mdx
marandaneto: looks duplicated though, could be as simple as

`<PlatformSection supported={["dotnet", "java"]}>`

above

`<PlatformContent includePath="performance/distributed-tracing" />`

so we don't need to duplicate this line

src/includes/performance/distributed-tracing/dotnet.mdx
marandaneto: mm idk about it, if its the case, I'm pretty sure that every doc using Guides should be using `PlatformLink` instead and it's not the case.

src/includes/performance/distributed-tracing/dotnet.mdx
Tyrrrz: I'm not sure if I understood it correctly, but I updated it. Can you please verify that I got it right?

src/platforms/common/performance/custom-instrumentation.mdx
Tyrrrz: I changed it, please take a look

src/includes/performance/distributed-tracing/dotnet.mdx
Tyrrrz: Done. Inverted it to avoid `return`, I think it's clearer that way.

src/includes/performance/distributed-tracing/dotnet.mdx
PeloWriter: Couple of things:
- the link is missing the preceding `/` (my fault, as I didn't c/p properly) which is why the linkchecker is failing
- we use the PlatformLink when the specific platform isn't already selected

Tyrrrz: fixed the links

src/includes/performance/distributed-tracing/dotnet.mdx
```diff
@@ -0,0 +1,51 @@
+## Distributed Tracing
```
marandaneto, line 1: this header should be under `custom-instrumentation.mdx ` otherwise it appears after the common content

src/includes/performance/distributed-tracing/java.mdx
```diff
@@ -1,7 +1,5 @@
 ## Distributed Tracing
 
-Traces can bridge across multiple software services. Each span in a trace can be represented as a `sentry-trace` header, containing the trace id, span id, and sampling details. This `sentry-trace` header can be passed to downstream services so that they can create spans that are a continuation of the trace started in the originating service.
```
marandaneto, line 3: same as https://github.com/getsentry/sentry-docs/pull/3363/files#r610391270

use the preview link https://sentry-docs-git-fork-tyrrrz-dotnet-distributed-tracing.sentry.dev/ so you can see how it looks like and spot bugs like that

src/includes/performance/distributed-tracing/dotnet.mdx
marandaneto: @PeloWriter what @bruno-garcia said makes sense, if you're in a Guide but using normal links instead of PlatformLink, once you click it, you move out of the Guide, is that the desired behavior? I'd think so

marandaneto (approved):
LGTM
although I'd wait for @PeloWriter review once more as there's a few open points

src/includes/performance/distributed-tracing/dotnet.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+using var request = new HttpRequestMessage(HttpMethod.Get, "https://example.com");
+request.Headers.Add("sentry-trace", sentryTraceHeader.ToString());
+```
+
+<Note>
+
+Sentry also provides `SentryHttpMessageHandler`, which can be used to [automatically attach trace headers to outgoing HTTP requests](/platforms/dotnet/performance/included-instrumentation/).
+
+</Note>
```
bruno-garcia, line 19: What if we point out it's out of the box before we show the snippet?

We don't just recommend doing this for HTTP, we recommend for any system communication which the user would like visibility end-to-end. Specifically for HTTP we provide `SentryHttpMessageHandler` so lets point them to it, and after that demonstrate how to programatically do it:


````suggestion
To obtain a trace header from the span, use `ISpan.GetTraceHeader()` method, then pass it to the downstream service.

<Note>

For HTTP communications, Sentry's SDK providers `SentryHttpMessageHandler`, which can be used to [automatically attach trace headers to outgoing HTTP requests](/platforms/dotnet/performance/included-instrumentation/).

</Note>

The code snippet below demonstrates propagating the trace programatically. Even though `SentryHttpMessageHandler` does that automatically for your, this illustrates the mechanism so it can be replicated in other protocols and technologies such a messaging systems.

```csharp
using System;
using System.Net.Http;
using Sentry;

var transaction = SentrySdk.StartTransaction("name", "op");
var sentryTraceHeader = transaction.GetTraceHeader();

using var request = new HttpRequestMessage(HttpMethod.Get, "https://example.com");
request.Headers.Add("sentry-trace", sentryTraceHeader.ToString());
```

````

src/includes/performance/distributed-tracing/dotnet.mdx
PeloWriter: Verified that we now have that common content properly displaying for .NET and Java. Thank you.

src/includes/performance/distributed-tracing/dotnet.mdx
PeloWriter: We have been using this format - [click this](page to to which you want users to go). When we need the platform context to be derived, we've been using the [click this](/platform-redirect/?next=/directory/content/).

PlatformLink will use the context of your most recently viewed platform; for example, if you've been reading platform content, then been directed to a product page and then click back into platform content.

PeloWriter: So this link - [automatically attach trace headers to outgoing HTTP requests](/platforms/dotnet/performance/included-instrumentation/) - when viewed on the  ASP.NET page, sends you to the primary .NET page, which looks correct to me (that page is common to the .NET family).

If we want users to stay in the guide content, we'd want the link to use [automatically attach trace headers to outgoing HTTP requests](/platforms/dotnet/guides/aspnet/performance/included-instrumentation/).

Either way, the page content is the same.

PeloWriter:
LGTM (and I tested the links to be sure we're being navigated to what we'd like to see). I think the only remaining question, @bruno-garcia, is whether to direct users to the content common to the .NET family or keep them within a guide.

PeloWriter (approved):

PR #3369, opened 2021-04-08
Align notes about using standalone crash reporting backends

torarnv:
Use same formatting and similar language.

The references to the Apple and Electron SDKs in the minidump docs
have been removed, as the link points to the root of the native
platform docs, which in turn links to all the guides, and alternate
SDKs. This gives a more consistent flow than duplicating the content
at multiple levels (resulting in slight differences over time due to
bitrot).

### DIFF
--- a/src/platforms/native/guides/breakpad/index.mdx
+++ b/src/platforms/native/guides/breakpad/index.mdx
@@ -8,11 +8,9 @@ description: "Learn about Sentry's integration with Google Breakpad."
 
 [Breakpad](https://chromium.googlesource.com/breakpad/breakpad/) is an open-source multiplatform crash reporting system written in C++ by Google and the predecessor of Crashpad. It supports macOS, Windows and Linux, and features an uploader to submit minidumps to a configured URL right when the process crashes.
 
-As opposed to Crashpad, Breakpad uses in-process crash reporting. This is less robust and has several disadvantages over out-of-process crash reporting. Unless you have integrated Breakpad already, we strongly recommend you to consider using Crashpad instead.
-
 <Note>
 
-We strongly encourage you to use our higher-level [Native SDK](/platforms/native/) if possible, as native Breakpad events are very limited in functionality.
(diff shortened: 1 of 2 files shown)


torarnv:
cc @PeloWriter  @bruno-garcia

Swatinem (approved):

imatwawana:
The wording looks good to me, so I'll go ahead and merge this. Thank you!!

torarnv:
Thanks! 😊

torarnv:
@PeloWriter  After having merged this I realized the original rationale for using `Alert` with warning level, instead of just a `Note`, was that the existing crashpad guide used a warning: https://docs.sentry.io/platforms/native/guides/crashpad/

So now the breakpad (https://docs.sentry.io/platforms/native/guides/breakpad/) and minidump (https://docs.sentry.io/platforms/native/guides/minidumps/) are using `Note`.

I personally think a note in this case is too low a level from a purely visual point of view, and the text talking about "strongly encouraging", which is why I went with a warning initially, but I'm fine either way 😊 Is there perhaps a middle ground that can be used? Is a plain Alert e.g. a thing?

torarnv:
Let me know and I can push a new fix to align the 3 pages )(crashpad, breakpad, minidumps)

PeloWriter:
so we have two primary options - Note as gray text (this our new standard, and looks like this)
Screen Shot 2021-04-08 at 3 46 51 PM

But that is sometimes not enough, so we also have an alert as a box with no background color, like this:
Screen Shot 2021-04-08 at 3 47 32 PM

And we have two other options - Note with an orange background (complementary to our purple highlight, which looks like this)
Screen Shot 2021-04-08 at 3 48 17 PM

or Note with a green background.

Here, I agree that the low level Note isn't sufficient. I'd change it to an Alert with no background. You can do that like so:

"
<Alert>

important text

</Alert>
"

And thanks for your keen eye for detail.

torarnv:
Thanks, done! 😊 https://github.com/getsentry/sentry-docs/pull/3373

PR #3374, opened 2021-04-08
Fix redirect from platform docs to specific guides

torarnv:
The note would end up in in guides as well due to the missing
noGuides attribute.

The PlaformSection was also excluding more than the guide redirection,
causing sub-guides to lose content. If we want to do this the change
should probably happen along with a followup to add it back to the
sub-guides unless they explicitly don't want that content.

The redirect is also now a green note, instead of a warning. Warnings
should be reserved for very extraordinary cases.

### DIFF
--- a/src/platforms/common/index.mdx
+++ b/src/platforms/common/index.mdx
@@ -1,18 +1,18 @@
 <PlatformContent includePath="getting-started-primer" />
 
-<PlatformSection notSupported={["android", "dart", "elixir", "flutter", "perl", "react-native", "unity"]}>
+On this page, we get you up and running with Sentry's SDK, so that it will automatically report errors and exceptions in your application.
 
-<Alert level="warning" title="Using a framework?">
+<PlatformSection noGuides notSupported={["android", "dart", "elixir", "flutter", "perl", "react-native", "unity"]}>
+
+<Alert level="info" title="Using a framework?">
 
 Get started using a guide listed in the right sidebar.
 
 </Alert>
 
-On this page, we get you up and running with Sentry's SDK, so that it will automatically report errors and exceptions in your application.
+</PlatformSection>
 
 Don't already have an account and Sentry project established? Head over to [sentry.io](https://sentry.io/signup/), then return to this page.
-
-</PlatformSection>
 
 ## Install
 


torarnv:
@PeloWriter

PeloWriter:
> The note would end up in in guides as well due to the missing
> noGuides attribute.

Thank you for adding this. I hadn't thought it an issue to surface the sibling guides, but I can see that this would be distracting if you're in Angular, for example, you've made your choice.

> The PlaformSection was also excluding more than the guide redirection,
> causing sub-guides to lose content. If we want to do this the change
> should probably happen along with a followup to add it back to the
> sub-guides unless they explicitly don't want that content.

Can you help me understand what was missing? I checked on my local and saw only the alert box added, but not any loss of content. 

> The redirect is also now a green note, instead of a warning. Warnings
> should be reserved for very extraordinary cases.

No strong opinion. I mildly preferred the visual impact of orange with our color palette, but that's why I'm a writer not a designer :)

torarnv:
> > The PlaformSection was also excluding more than the guide redirection,
> > causing sub-guides to lose content. If we want to do this the change
> > should probably happen along with a followup to add it back to the
> > sub-guides unless they explicitly don't want that content.
> 
> Can you help me understand what was missing? I checked on my local and saw only the alert box added, but not any loss of content.

Sure! Here's what I think is going on:

Before a0bf46c55 the introduction section of e.g. the Android and React Native pages had the following bit:

"On this page, we get you up and running with Sentry's SDK, so that it will automatically report errors and exceptions in your application.

Don't already have an account and Sentry project established? Head over to [sentry.io](https://sentry.io/signup/), then return to this page."

This is because they do not have an `index.md` in `/src/platforms/foo` that override the whole page, but rely on injecting sub-pages.

Since a0bf46c55 added a `notSupported` section around that text, these pages lost the snippet.

With this change, any sub-guide without their own `index.md` would also lose those bits.

> > The redirect is also now a green note, instead of a warning. Warnings
> > should be reserved for very extraordinary cases.
> 
> No strong opinion. I mildly preferred the visual impact of orange with our color palette, but that's why I'm a writer not a designer :)

Hehe, I'm neither, I was just thinking of the advice you gave me earlier about reserving warnings for special dangerous occasions, so a more middle ground green box felt like a good compromise where it stands out visually, but doesn't scream "warning, do this or your computer will burn" 😊

PeloWriter:
You are completely correct - I was focused on the box placement in my testing. Thanks for seeing that!

And green is fine :) Good middle ground - suggests you look and read, but not necessarily stop. We do generally try to avoid "do this or your computer will burn" instructions - ha!

torarnv:
Thanks for quick review! 😊

PR #3375, opened 2021-04-09
docs(js): Add note regarding browser devtools not triggering errors

kamilogorek:
Fixes https://github.com/getsentry/sentry-javascript/issues/3378

### DIFF
--- a/src/platforms/common/index.mdx
+++ b/src/platforms/common/index.mdx
@@ -34,6 +34,7 @@ This snippet includes an intentional error, so you can test that everything is w
 
 <Note>
 
+Keep in mind that errors triggered within Browser DevTools are sandboxed and they will not trigger an error handler. Place the snippet directly in your code instead.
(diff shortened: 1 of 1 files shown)


src/platforms/common/index.mdx
```diff
@@ -34,6 +34,7 @@ This snippet includes an intentional error, so you can test that everything is w
 
 <Note>
 
+Keep in mind that errors triggered within Browser DevTools are sandboxed and they will not trigger an error handler. Place the snippet directly in your code instead.
```
PeloWriter, line 37: This will appear on the Getting Stated page for every SDK - which I suspect we don't want as this is a note specific to JS. 

I'd recommend instead putting this in the `includes/getting-started-verify/javascript.mdx` file.

```suggestion
Errors triggered from within Browser DevTools are sandboxed, so will not trigger an error handler. Place the snippet directly in your code instead.
```

PR #3395, opened 2021-04-12
Remove aspnet advise for 4.8

bruno-garcia:
Has confused ppl that this is a requirement.


PeloWriter (commented):
Good to pull this out (also it's written in the first person...!). The current version should display already in the package/version/repo box.

PR #3408, opened 2021-04-15
Remove Apache Transport docs.

maciejwalkowiak:
Apache HttpClient that's used in Apache HttpClient Transport has bugs causing high CPU usage. Until bugs are fixed we should not include it in the docs.


maciejwalkowiak:
cc @bruno-garcia

bruno-garcia (approved):

Issue #3407, opened 2021-04-15
Unity docs next steps


bruno-garcia:
We just merged #3330 that describing setting up Unity with [our new SDK](https://github.com/getsentry/sentry-unity).
This was just the first iteration of Unity docs. 

A few things:

- [x] * Performance Monitoring snippets were added but the section isn't displayed yet~ Addressed by #3411
- [x]  Wizard for Unity to replace Unity Lite
- [x] Some code snippets assume .NET SDK init (`SentrySdk.Init`) i.e: [BeforeBreadcrumb](https://docs.sentry.io/platforms/unity/enriching-events/breadcrumbs/#customize-breadcrumbs)~ Addressed on #3409
- [x] Use common content installation with the new SDK [instead of sentry-unity-lite](https://github.com/getsentry/sentry-docs/blob/master/src/includes/getting-started-install/unity.mdx) - addressed by https://github.com/getsentry/sentry-docs/pull/3414

To address the code snippets, we decided to avoid showing the whole init call, and move towards simply the option.
I.e:

Before:
```
using Sentry;

SentrySdk.Init(options =>
{
    options.BeforeBreadcrumb = breadcrumb
        // Ignore breadcrumbs from Spammy logger
        => breadcrumb.Category == "Spammy.Logger"
            ? null
            : breadcrumb;
});
```

After
```
// Add this to the SDK initialization callback
options.BeforeBreadcrumb = breadcrumb
    // Ignore breadcrumbs from Spammy logger
    => breadcrumb.Category == "Spammy.Logger"
        ? null
        : breadcrumb;
```

This snippet is now valid for ASP.NET Core, Unity, Xamarin and the rest of .NET. Where the original one was not valid for the first 3.

PR #3409, opened 2021-04-15
.NET: simplify initialization snippets

Tyrrrz:
@bruno-garcia 
Realted to #3407


src/includes/configuration/sample-rate/dotnet.mdx
```diff
@@ -1,6 +1,7 @@
 ```csharp
 using Sentry;
 
+// Add this to the SDK initialization callback
 using (SentrySdk.Init(o =>
```
bruno-garcia, line 5: Still doing init

Tyrrrz: oops

Tyrrrz: here we have `// app code here` so i just left it as it is.
removed my comment.

bruno-garcia (commented):
Everything here makes sense but browsing the code I continue to see SentrySdk.Init such as:

```
using (SentrySdk.Init(o =>
{
    o.SampleRate = 0.25; // Capture 25%
})
{
    // app code here
}
```

bruno-garcia:
under `configuration/sampling/`

bruno-garcia:
I went through all docs and only spotted that one.

bruno-garcia (approved):

PR #3412, opened 2021-04-15
fix(integrations): Separate out metric alerts v issue alerts

manuzope:
To make it clearer on what we support v not support on each integration.

Screen Shot 2021-04-15 at 4 26 50 PM


jonesphillip (approved):
lgtm

PeloWriter (approved):
Nice update. Thanks, @manuzope

PR #3415, opened 2021-04-16
Fixing filter by issue

rbisol:


src/docs/product/accounts/quotas/index.mdx
```diff
@@ -198,7 +198,9 @@ Sentry supports filtering out a specific or certain kind of error as well. Navig
   Delete and Discard is available on the Business plan.
 </Alert>
 
-When you are unable to take immediate action on an issue, but it continues to occur, Sentry supports deleting and discarding that issue from the issue details page. Navigate to **[Project] » Project Settings » Inbound Filters**, then click “delete and discard future events.” This setting deletes most data associated with the issue and filters out matching events before they count against your quota.
+When you are unable to take immediate action on an issue, but it continues to occur, Sentry supports deleting and discarding that issue from the issue details page. Navigate to the issue you would like to filter, then click on the drop-down menu next to the bin icon, and select the “delete and discard future events” option. This setting deletes most data associated with the issue and filters out matching events before they count against your quota.
```
imatwawana, line 201: Original phrasing "...deleting from the issue details page" could be be understood to mean the effect is only taking place in the UI and not in the system generally. Reworded slightly.

```suggestion
When you are unable to take immediate action on an issue, but it continues to occur, Sentry supports deleting and discarding that issue, which you can do from the **Issue Details** page. Navigate to the issue you would like to filter, click on the drop-down menu next to the bin icon, and select “Delete and discard future events”. This setting deletes most data associated with the issue and filters out new matching events before they count against your quota.
```

imatwawana (approved):
Minor wording edits for clarity and concision. Otherwise LGTM.

PR #3420, opened 2021-04-16
apple: list types of errors captured

bruno-garcia:


src/includes/getting-started-primer/apple.mdx
```diff
@@ -8,6 +8,12 @@
 
 **Features:**
 
+- Multiple types of errors are captured, including:
+  - C++ Exception
+  - Mach Exceptions
+  - Unhandled Exceptions
+  - UNIX Signal
+  - Out of memory
```
PeloWriter, line 16: out of memory is still in Alpha - are we comfortable listing it here?
```suggestion
- Multiple types of errors are captured, including:
  - C++ exceptions
  - Mach exceptions
  - Unhandled exceptions
  - UNIX signal
  - Out of memory
```

PeloWriter (approved):

src/includes/getting-started-primer/apple.mdx
bruno-garcia: It's in beta now. We plan to GA next week.

bruno-garcia:
Thanks for the review!
Lets see if @philipphofmann and @HazAT have ideas before we merge this

src/includes/getting-started-primer/apple.mdx
```diff
@@ -8,6 +8,12 @@
 
 **Features:**
 
+- Multiple types of errors are captured, including:
+  - C++ exceptions
+  - Mach exceptions
+  - Unhandled exceptions
+  - UNIX signal
+  - Out of memory
```
HazAT, line 16: Strictly speaking, I think it's not correct, we only capture breadcrumbs and optionally an event

HazAT: But since I learned that everyone else is guessing when an OOM crash happened, I think we are fine merging it.

bruno-garcia: We added a feature in Q1 to create an event for OOM (it's on by default): https://github.com/getsentry/sentry-cocoa/pull/1015

It's based on heuristics, but as you said, so is anyone else doing this since it's the only way

PeloWriter:
Ready to merge, @bruno-garcia ?

bruno-garcia:
I was hoping @philipphofmann would engage. Let me try Slack

bruno-garcia:
Better, lets merge and if there's more, can come in follow up PRs.

PR #3421, opened 2021-04-17
#1 mojombo#1 mojombo/github-flavored-markdown#1

kennethbaker66:


PeloWriter:
I'm uncertain what the intended change is here. And it's not clearly improving the docs, so will close.

PR #3427, opened 2021-04-20
Fix migration guides to make the behaviors of two codes same

tomoasleep:
This PR fixes an example code in the migration guide for Ruby to make the behaviors of the before code and the after code same.


PeloWriter:
@st0012 - can you please review?

st0012 (approved):

PR #3436, opened 2021-04-21
ref(integration-platform): Update permissions language

MeredithAnya:
Docs PR to reflect the changes made in https://github.com/getsentry/sentry/pull/25458. The two changes are:
* Make it explicit that permissions are scoped to the organization, you can't limit the integration to only have access to certain projects/teams within an integration (at least for now). 
* Mention that Managers and Owners are the ones that can create the internal or public integrations. 

Screen Shot 2021-04-21 at 12 26 42 PM
Screen Shot 2021-04-21 at 12 26 31 PM

### DIFF
--- a/src/docs/product/integrations/integration-platform/index.mdx
+++ b/src/docs/product/integrations/integration-platform/index.mdx
@@ -20,7 +20,7 @@ Permissions specify what level of access your service requires of Sentry resourc
 
 ![Form that allows developer to set what permissions they'll need from their user.](permissions.png)
 
-You cannot create an integration or change an existing integration to have permissions greater than your own user settings.
+You cannot create an integration or change an existing integration to have permissions greater than your own user settings. Permissions are scoped to the organization, so adding scopes for issues and events, projects, or teams will enable access to issues and events, projects, or teams across the organization. 
 
 ### Using Auth Tokens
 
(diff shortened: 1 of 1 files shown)


ceorourke (approved):

src/docs/product/integrations/integration-platform/index.mdx
```diff
@@ -20,7 +20,7 @@ Permissions specify what level of access your service requires of Sentry resourc
 
 ![Form that allows developer to set what permissions they'll need from their user.](permissions.png)
 
-You cannot create an integration or change an existing integration to have permissions greater than your own user settings.
+You cannot create an integration or change an existing integration to have permissions greater than your own user settings. Permissions are scoped to the organization, so adding scopes for issues and events, projects, or teams will enable access to issues and events, projects, or teams across the organization.
```
mgaeta, line 23: I wish we could use parens in normal human language. Maybe something like this would make the sentence less complex?
> Permissions are scoped to the organization, so adding scopes for issues and events, projects, or teams will enable access to **the corresponding features** across the organization.

manuzope (approved):

src/docs/product/integrations/integration-platform/index.mdx
MeredithAnya: I think this will make more sense when looking at the image (which is not shown in the pr description) but is [here](https://docs.sentry.io/product/integrations/integration-platform/#permissions).

PR #3444, opened 2021-04-21
ref(trace-view): Add mention of errors and how quota limits

wmak:
- Mention errors as well to help troubleshoot better

### DIFF
--- a/src/docs/product/performance/trace-view.mdx
+++ b/src/docs/product/performance/trace-view.mdx
@@ -36,7 +36,7 @@ The transaction view provides a split view similar to the [span view](/product/p
 
 ### Orphan Traces and Broken Subtraces
 In a fully instrumented system, a span in one service will connect to a transaction in a subsequent service. For a variety of reasons, a transaction in a service may not arrive in Sentry. When Sentry encounters these types of transactions within a trace, we link the transactions with a dashed line since they can no longer be directly connected to the root.
-In addition, broken subtraces can occur when Sentry receives an error for a trace, but the corresponding transaction does not exist. In this case, the error will not appear on the trace view and the quick trace won't appear on the Issue Details page.
+In addition, broken subtraces can occur when Sentry receives an error for a trace, but the corresponding transaction does not exist. In this case, the error will not appear on the trace view and the quick trace won't appear on the Issue Details page. In these cases you can click "Open In Discover" to see all the events in the trace.
 
 Broken subtraces may be caused by:
 - [Ad blockers](/platforms/javascript/troubleshooting/#dealing-with-ad-blockers) may prevent transactions in browsers being sent, but HTTP requests to backend projects will still create child transactions
--- a/src/docs/product/performance/trace-view.mdx
+++ b/src/docs/product/performance/trace-view.mdx
@@ -36,11 +36,13 @@ The transaction view provides a split view similar to the [span view](/product/p
 
 ### Orphan Traces and Broken Subtraces
 In a fully instrumented system, a span in one service will connect to a transaction in a subsequent service. For a variety of reasons, a transaction in a service may not arrive in Sentry. When Sentry encounters these types of transactions within a trace, we link the transactions with a dashed line since they can no longer be directly connected to the root.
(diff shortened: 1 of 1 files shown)


src/docs/product/performance/trace-view.mdx
```diff
@@ -36,11 +36,13 @@ The transaction view provides a split view similar to the [span view](/product/p
 
 ### Orphan Traces and Broken Subtraces
 In a fully instrumented system, a span in one service will connect to a transaction in a subsequent service. For a variety of reasons, a transaction in a service may not arrive in Sentry. When Sentry encounters these types of transactions within a trace, we link the transactions with a dashed line since they can no longer be directly connected to the root.
+As well broken subtraces can occur when sentry received an error for a trace, but the corresponding transaction does not exist. This means that the error will not appear on the trace view, and the quick trace will be empty.
```
PeloWriter, line 39:
```suggestion
In addition, broken subtraces can occur when Sentry receives an error for a trace, but the corresponding transaction does not exist. In this case, the error will not appear on the trace view and the quick trace will be empty.
```

src/docs/product/performance/trace-view.mdx
```diff
@@ -36,11 +36,13 @@ The transaction view provides a split view similar to the [span view](/product/p
...
 
 Broken subtraces may be caused by:
 - [Ad blockers](/platforms/javascript/troubleshooting/#dealing-with-ad-blockers) may prevent transactions in browsers being sent, but HTTP requests to backend projects will still create child transactions
 - [Rate-limiting](/product/accounts/quotas/#limiting-events) on a project may cause only some events to be sent to Sentry
 - [Project permissions](/product/accounts/membership/#restricting-access) may mean you do not have access to transactions in another project
+- [Quota Limits](/product/accounts/quotas/) differ between transactions and errors which means that an error was received, but not the corresponding transaction.
```
PeloWriter, line 45:
```suggestion
- [Differences in quota limits](/product/accounts/quotas/) between transactions and errors. When a quota limit is reached - for example, for transactions - an error is received, but corresponding transaction is not.
```

PeloWriter (approved):

src/docs/product/performance/trace-view.mdx
imatwawana, line 39: I'm not sure if "the quick trace will be empty" is clear. Do you mean the Quick Trace element won't show up at all or some part of it will be empty? If it's the first, that needs to be explicit and if it's the latter, a screenshot might be needed to illustrate that.

wmak: It won't show up at all would this be better?
```suggestion
As well broken subtraces can occur when sentry received an error for a trace, but the corresponding transaction does not exist. This means that the error will not appear on the trace view, and the quick trace will not show up.
```

imatwawana: Suggest the following for clarity:

...and the quick trace won't appear on the **Event Details** or **Issue Details** pages.

wmak: :+1: its only for issue details, but updated in 61b92d6

src/docs/product/performance/trace-view.mdx
```diff
@@ -36,11 +36,13 @@ The transaction view provides a split view similar to the [span view](/product/p
 
 ### Orphan Traces and Broken Subtraces
 In a fully instrumented system, a span in one service will connect to a transaction in a subsequent service. For a variety of reasons, a transaction in a service may not arrive in Sentry. When Sentry encounters these types of transactions within a trace, we link the transactions with a dashed line since they can no longer be directly connected to the root.
+In addition, broken subtraces can occur when Sentry receives an error for a trace, but the corresponding transaction does not exist. In this case, the error will not appear on the trace view and the quick trace won't appear on the Issue Details page.
```
Zylphrex, line 39: Do we want a shoutout to the "Open in Discover" button here?

wmak: Hmm, I'm not sure I understand why?

Zylphrex: Because errors that are part of a missing transaction will not appear in trace view but can be found in discover.

PR #3467, opened 2021-04-26
Update user feedback to common

PeloWriter:
User Feedback content was a bit confusing to manage. This PR pulls the content into one common file, with either the embeddable widget or the API content (and sometimes both) based on the SDK.


src/includes/user-feedback/http-request/android.mdx
```diff
@@ -0,0 +1,26 @@
+```java
+import io.sentry.Sentry;
```
bruno-garcia, line 2: The name `user-feedback/http-request` is a bit odd since these examples are showing the SDK API in that programming language. As opposed to the Web API added to the bottom of the common user feedback page:

> If you'd prefer an alternative to the widget or do not have a JavaScript frontend, you can use the [User Feedback API](/api/projects/submit-user-feedback/).

PeloWriter: Yes, I agree. I can change it to `user-feedback/sdk-api-example`

src/includes/user-feedback/example-widget/dotnet.mdx
```diff
@@ -0,0 +1,24 @@
...
+  crossorigin="anonymous"
+></script>
+```
+
+You'll then need to call `showReportDialog` and pass in the generated event ID.
+This event ID is returned from all calls to <PlatformIdentifier
+name="capture-event" /> and <PlatformIdentifier name="capture-exception" />.
+There is also a function called <PlatformIdentifier name="last-event-id" />
+that returns the ID of the most recently sent event.
+
```
bruno-garcia, line 16: Can we add a `note` that for users using ASP.NET or ASP.NET Core, there's a specific content for this?
We have overrides for this on those that include a proper snippet:

https://docs.sentry.io/platforms/dotnet/guides/aspnet/enriching-events/user-feedback/#integration

https://docs.sentry.io/platforms/dotnet/guides/aspnetcore/enriching-events/user-feedback/#integration

I'm just afraid folks land on .NET and click on User Feedback on the left without understanding that there's a "better" content if they are in fact using a web technology.

bruno-garcia (approved):
Left one suggestion for improving, but LGTM

src/includes/user-feedback/example-widget/dotnet.mdx
PeloWriter: yep. The design does that now. 

ASP.NET - https://sentry-docs-kl81mjezf.sentry.dev/platforms/dotnet/guides/aspnet/enriching-events/user-feedback/#integration


ASP.NET Core - https://sentry-docs-kl81mjezf.sentry.dev/platforms/dotnet/guides/aspnetcore/enriching-events/user-feedback/#integration

bruno-garcia: I meant that in here: https://sentry-docs-kl81mjezf.sentry.dev/platforms/dotnet/enriching-events/user-feedback/

There's no indication that this snippet:
image

Is not what you need if you're on ASP.NET or ASP.NET Core (90% of all Web Apps in .NET fall into one of these two)

PeloWriter: ah, got it. Yes, we can do that.

philipphofmann:
Thanks, @PeloWriter for improving this 👏🏼

Issue #3463, opened 2021-04-25
Add documentation about tracing


VincentLanglet:
## Issue

This an issue with:
https://docs.sentry.io/platforms/php/guides/symfony/

## Description

Tracing support is new in sentry-symfony
https://github.com/getsentry/sentry-symfony/pull/477
https://github.com/getsentry/sentry-symfony/pull/426
https://github.com/getsentry/sentry-symfony/pull/460
https://github.com/getsentry/sentry-symfony/pull/430
https://github.com/getsentry/sentry-symfony/pull/423

The doc should be updated to explain how to use it.

rhcarvalho:
Noting there's already https://docs.sentry.io/platforms/php/guides/symfony/performance/.

PR #3470, opened 2021-04-27
Initial content for Symfony performance integrations

PeloWriter:
Adding initial content to cover the Symfony tracing after a brief convo with @HazAT. Related to https://github.com/getsentry/sentry-docs/issues/3463,

Fixes https://github.com/getsentry/sentry-docs/issues/3463

### DIFF
--- a/src/platforms/php/common/integrations.mdx
+++ b/src/platforms/php/common/integrations.mdx
@@ -8,49 +8,42 @@ redirect_from:
 
 ## Default Integrations
 
-Default integrations are integrations enabled by default that integrate into the
-standard library or the interpreter itself. They are documented so you can see
-what they do and that they can be disabled if they cause issues. To disable
-system integrations set `default_integrations => false` when calling `init()`.
+These integrations are enabled by default and integrate into the
+standard library or the interpreter itself. They are documented so you can be both aware of what they do and disable them if they cause issues.
+
+To disable system integrations, set `default_integrations => false` when calling `init()`.
 
 ### ExceptionListenerIntegration
 
 This integration catches all global uncaught exceptions and emits events when an error occurs.
 
-To do that, it ensures that Sentry's `ErrorHandler` is registered, and adds a callback to it
-as an exception listener.
+To do so, it ensures that Sentry's `ErrorHandler` is registered, and adds a callback to it as an exception listener.
 
 ### ErrorListenerIntegration
 
 This integration hooks into the global PHP `error_handler` and emits events when an error occurs.
 
-To do that, it ensures that Sentry's `ErrorHandler` is registered, and adds a callback to it
-as an error listener. By default, the `ErrorHandler` reserves 10 kilobytes of memory to handle
-fatal errors.
+To do so, it ensures that Sentry's `ErrorHandler` is registered, and adds a callback to it as an error listener. By default, the `ErrorHandler` reserves 10 kilobytes of memory to handle fatal errors.
 
-For some frameworks or projects, there are specific integrations provided both
-officially and by third-party users that automatically register the error
-handlers. In that case, please refer to their documentation.
+For some frameworks or projects, specific integrations are provided both officially and by third-party users that automatically register the error handlers. Please refer to their documentation.
 
-Also, by default `E_ALL` will be handled, you can change it by setting `error_types`
-a different constant in `init()`.
+Also, by default `E_ALL` will be handled. You can change it by setting `error_types` to a different constant in `init()`.
 
 ### FatalErrorListenerIntegration
 
 This integration catches all fatal errors and emits the corresponding events.
 
-To do that, it ensures that Sentry's `ErrorHandler` is registered, and adds a callback to it
-as a fatal error listener.
+To do so, it ensures that Sentry's `ErrorHandler` is registered, and adds a callback to it as a fatal error listener.
 
 ### RequestIntegration
 
-This integration adds to the event request data like:
+This integration adds to the event request data:
 
 - HTTP method
 - URL (including the query string)
 - Body (by default only if the body is 10Kb or less)
 
-If the [`send_default_pii` option](../configuration/options/#send-default-pii) is enabled, it will also send PII information like:
+If the [`send_default_pii` option](../configuration/options/#send-default-pii) is enabled, it will also send the following PII:
 
 - IP address
 - Cookies
@@ -68,12 +61,11 @@ This integration reads excerpts of code around the line that originated an error
 
 This integration fills the event data with PHP runtime and server OS information.
 
-
 ## Optional Integrations
 
-You can enable option integration by passing them to the `integrations` option when initializing the Sentry SDK.
+You can enable these integrations by passing them to the `integrations` option when initializing the Sentry SDK.
 
-In the example below we add the optional `ModulesIntegration`.
+In this example, the optional `ModulesIntegration` is added:
 
 ```php
 \Sentry\init([
@@ -90,14 +82,13 @@ This integration decides whether an event should not be captured according to a 
 
 ### ModulesIntegration
 
-This integration logs with the event details all the versions of the packages installed with Composer; the root project is included too.
+This integration logs all the versions of the packages installed with Composer with the event details; the root project is also included.
 
+## Custom Integrations
 
-## Customize Integrations
+You can customize the list of integration without disabling the default integration using the `integrations` option and passing a callable.
 
-It is also possible to customize the list of integration without disabling the default integration using the `integrations` option and passing a callable.
-
-In the example below we leave all integrations enabled exception for the `ExceptionListenerIntegration` and add the `ModulesIntegration`.
+In the example below, all integrations are enabled except the `ExceptionListenerIntegration`. In addition, the `ModulesIntegration` is added.
 
 ```php
 \Sentry\init([
--- a/src/platforms/php/common/integrations.mdx
+++ b/src/platforms/php/common/integrations.mdx
@@ -27,7 +27,8 @@ To do so, it ensures that Sentry's `ErrorHandler` is registered, and adds a call
 
 For some frameworks or projects, specific integrations are provided both officially and by third-party users that automatically register the error handlers. Please refer to their documentation.
 
-Also, `E_ALL` defaults to the value returned by the `error_reporting()` function, which can change at runtime. Alternately, you can set the option to a fixed value by setting `error_types` to a different constant in `\Sentry\init()`. In this case, Sentry will log errors of that specific type regardless of the current reporting level.
+By default, the error_types option defaults to the value returned by the `error_reporting()` function, which can change at runtime. Alternately, you can set the option to a fixed value by setting its value to a bitmask of the PHP `E_*` constants in `\Sentry\init()`. In this case, Sentry will log only errors of those specific types regardless of the current reporting level.
+
 
 ### FatalErrorListenerIntegration
 
--- a/src/platforms/php/common/integrations.mdx
+++ b/src/platforms/php/common/integrations.mdx
@@ -87,7 +87,7 @@ This integration logs all the versions of the packages installed with Composer w
 
 ## Custom Integrations
 
-You can customize the list of integration without disabling the default integration using the `integrations` option. You can pass a callable in advanced needs, though a fixed list of integrations will work in use cases.
+You can customize the list of integration without disabling the default integration using the `integrations` option. You can pass a callable for advanced needs, though a fixed list of integrations will work in most use cases.
 
 In the example below, all integrations are enabled except the `ExceptionListenerIntegration`. In addition, the `ModulesIntegration` is added.
 
--- a/src/platforms/php/common/integrations.mdx
+++ b/src/platforms/php/common/integrations.mdx
@@ -27,7 +27,7 @@ To do so, it ensures that Sentry's `ErrorHandler` is registered, and adds a call
 
 For some frameworks or projects, specific integrations are provided both officially and by third-party users that automatically register the error handlers. Please refer to their documentation.
 
-Also, `E_ALL` defaults to the current PHP reporting level. You can explicitly set the option by setting `error_types` to a different constant in `\Sentry\init()`.
+Also, `E_ALL` defaults to the value returned by the `error_reporting()` function, which can change at runtime. Alternately, you can set the option to a fixed value by setting `error_types` to a different constant in `\Sentry\init()`. In this case, Sentry will log errors of that specific type regardless of the current reporting level.
 
 ### FatalErrorListenerIntegration
 
--- a/src/platforms/php/common/integrations.mdx
+++ b/src/platforms/php/common/integrations.mdx
@@ -11,7 +11,7 @@ redirect_from:
 These integrations are enabled by default and integrate into the
 standard library or the interpreter itself. They are documented so you can be both aware of what they do and disable them if they cause issues.
 
-To disable system integrations, set `default_integrations => false` when calling `init()`.
+To disable system integrations, set `default_integrations => false` when calling `\Sentry\init()`.
 
 ### ExceptionListenerIntegration
 
@@ -27,7 +27,7 @@ To do so, it ensures that Sentry's `ErrorHandler` is registered, and adds a call
 
 For some frameworks or projects, specific integrations are provided both officially and by third-party users that automatically register the error handlers. Please refer to their documentation.
 
-Also, by default `E_ALL` will be handled. You can change it by setting `error_types` to a different constant in `init()`.
+Also, `E_ALL` defaults to the current PHP reporting level. You can explicitly set the option by setting `error_types` to a different constant in `\Sentry\init()`.
 
 ### FatalErrorListenerIntegration
 
@@ -86,7 +86,7 @@ This integration logs all the versions of the packages installed with Composer w
 
 ## Custom Integrations
 
-You can customize the list of integration without disabling the default integration using the `integrations` option and passing a callable.
+You can customize the list of integration without disabling the default integration using the `integrations` option. You can pass a callable in advanced needs, though a fixed list of integrations will work in use cases.
 
 In the example below, all integrations are enabled except the `ExceptionListenerIntegration`. In addition, the `ModulesIntegration` is added.
 
--- a/src/platforms/php/guides/symfony/index.mdx
+++ b/src/platforms/php/guides/symfony/index.mdx
@@ -48,11 +48,13 @@ SENTRY_DSN="___PUBLIC_DSN___"
 ###< sentry/sentry-symfony ###
 ```
 
-<Note>
+<Alert level= "warning" title="Performance">
 
-Performance monitoring integrations related to database, caching, and template rendering are enabled by default. To disable them, please see the [Integrations documentation](/platforms/php/guides/symfony/performance/pm-integrations/).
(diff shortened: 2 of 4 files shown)


src/platforms/php/guides/symfony/index.mdx
```diff
@@ -48,6 +48,12 @@ SENTRY_DSN="___PUBLIC_DSN___"
 ###< sentry/sentry-symfony ###
 ```
 
+<Note>
+
+Performance monitoring integrations related to database, caching, and template rendering are disabled by default. To enable them, please see the [Integrations documentation](/platforms/php/guides/symfony/performance/pm-integrations/).
```
Jean85, line 53: This is not true: we've enabled it by default.

@HazAT if that is not desired let us know.

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
```diff
@@ -0,0 +1,63 @@
+---
+title: Performance Integrations
+description: "Learn about the the integrations Sentry provides for performance monitoring and how to configure them."
+sidebar_order: 500
+---
+
+The Symfony SDK offers integrations that support distributed tracing, including database queries, Twig template rendering, and cache pools. These integrations are disabled by default to avoid adding unexpected overhead to your services.
```
Jean85, line 7: Again, false, it's enabled by default.

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+To use them, you need to use the latest version of the SDK and add the specific integration as illustrated below:
```
Jean85, line 9: These below are the default settings, if each of those components are present.

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
```diff
@@ -0,0 +1,63 @@
...
+      enabled: true
+    twig: # templating engine
+      enabled: true
+```
+
+### Dbal
+
+This integration supports tracking database queries as Spans. Enabling this option will create Spans for you queries.
+
+It is disabled by default. To enable it, set `enabled` to `true`.
```
Jean85, line 27: Again.

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
```diff
@@ -0,0 +1,63 @@
...
+    enabled: true
+    dbal: # DB queries
+      enabled: true
+```
+
+### twig
+
+This integration adds distributed tracing for Twig template rendering using Sentry.
+
+It is disabled by default. To enable it, set `enabled` to `true`.
```
Jean85, line 41: Again..

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
```diff
@@ -0,0 +1,63 @@
...
+    enabled: true
+    twig: # DB queries
+      enabled: true
+```
+
+### cache
+
+This integration adds support for distributed tracing of cache pools.
+
+It is disabled by default. To enable it, set `enabled` to `true`.
```
Jean85, line 55: Again.

Jean85 (changes requested):
There's one common misconception: tracing is enabled by default.

Apart from that, I find the current structure of the guide for Symfony confusing: there's `Config` and `Configuration` sections, there's constant cross-reference to content related to the generic PHP SDK, there's no line between what's responsibility of the Symfony bundle and what belongs to the base SDK...

PeloWriter:
I'm checking [this file](https://github.com/getsentry/sentry-symfony/blob/develop/tests/DependencyInjection/Fixtures/yml/full.yml#L40-L48), and see that `enabled` is set to `false`, which is why the content says it is disabled by default. 

The structure issue requires a larger investigation, thanks for raising it.

ste93cry:
>I'm checking this file, and see that enabled is set to false, which is why the content says it is disabled by default.

Sorry, that's just a fixture file used for the tests. As soon as the conditions for each integration to be enabled are satisfied the integration is automatically enabled

PeloWriter:
Got it. Thanks for clarifying. I'll modify appropriately.

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
PeloWriter: Fixed.

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
PeloWriter: Fixed.

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
PeloWriter: Fixed.

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
PeloWriter: Fixed.

src/platforms/php/guides/symfony/index.mdx
PeloWriter: Fixed.

PeloWriter:
I've also added an [issue](https://github.com/getsentry/sentry-docs/issues/3479) to track the need to unknot the confusion between Symfony and PHP.

Jean85 (approved):

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
```diff
@@ -0,0 +1,63 @@
...
+      enabled: true
+    twig: # templating engine
+      enabled: true
+```
+
+### Dbal
+
+This integration supports tracking database queries as Spans. Enabling this option will create Spans for your queries.
+
+It is enabled by default. To disable it, set `enabled` to `false`.
```
ste93cry, line 27: Since there are two `enabled` keys, I suggest to either remove `sentry.tracing.enabled` (which is the global toggle and has no meaning in the context of this specific part of the documentation) or to refer to `sentry.tracing.dbal.enabled` in this paragraph

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
```diff
@@ -0,0 +1,63 @@
...
+    enabled: true
+    dbal: # DB queries
+      enabled: true
+```
+
+### twig
+
+This integration adds distributed tracing for Twig template rendering using Sentry.
+
+It is enabled by default. To disable it, set `enabled` to `false`.
```
ste93cry, line 41: Since there are two `enabled` keys, I suggest to either remove `sentry.tracing.enabled` (which is the global toggle and has no meaning in the context of this specific part of the documentation) or to refer to `sentry.tracing.dbal.enabled` in this paragraph

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
```diff
@@ -0,0 +1,63 @@
...
+    enabled: true
+    twig: # DB queries
+      enabled: true
+```
+
+### cache
+
+This integration adds support for distributed tracing of cache pools.
+
+It is enabled by default. To disable it, set `enabled` to `false`.
```
ste93cry, line 55: Since there are two `enabled` keys, I suggest to either remove `sentry.tracing.enabled` (which is the global toggle and has no meaning in the context of this specific part of the documentation) or to refer to `sentry.tracing.dbal.enabled` in this paragraph

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
ste93cry: As long as the vendors (DBAL, Cache and Twig) are installed the `enabled` key will be set to `true`, so we may rephrase a bit this paragraph to tell users that it's likely that this config is the same of their app, but they should adjust it to their needs if they want to disable specific integrations.

ste93cry (changes requested):

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
PeloWriter: I've added this to the intro.

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
PeloWriter: Good call. I edited the code sample to remove everything but:

```
dbal: # DB queries
      enabled: true
```

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
PeloWriter: edited as above.

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
PeloWriter: edited as above.

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
ste93cry: To avoid any mistake, you should use the full config path in the snippets, otherwise users may misinterpret them and think that they have to configur the `dbal` key at the root node of the file which is not what we mean. TL;DR: I would expect the code snippet to be

```yaml
sentry:
    tracing:
        dbal:
            enabled: true
```

PeloWriter:
@ste93cry - can you please review with the updated samples? Many thanks.

src/platforms/php/guides/symfony/index.mdx
```diff
@@ -48,6 +48,12 @@ SENTRY_DSN="___PUBLIC_DSN___"
 ###< sentry/sentry-symfony ###
 ```
 
+<Note>
+
+Performance monitoring integrations related to database, caching, and template rendering are enabled by default. To disable them, please see the [Integrations documentation](/platforms/php/guides/symfony/performance/pm-integrations/).
```
ste93cry, line 53: A few suggestions:

- Can we be more generic on this 🙏? Since I expect new integrations to come up in the future, this way we won't have to remember that this note needs to be updated to cite all of them
- Wouldn't be better to make this a warning rather than a note? In fact, integrations are hooking into critical paths of the framework and of the vendors and therefore, as I already said, there is a performance penalty. Given that people updating from a previous version or installing the SDK will suddendly see their app performance lower, even if maybe it's not noticeable for most of them, this reason should be enough to make prominent this advise of disabling the tracing feature when unneeded

src/platforms/php/common/integrations.mdx
```diff
@@ -8,49 +8,42 @@ redirect_from:
...
 ## Default Integrations
 
-Default integrations are integrations enabled by default that integrate into the
-standard library or the interpreter itself. They are documented so you can see
-what they do and that they can be disabled if they cause issues. To disable
-system integrations set `default_integrations => false` when calling `init()`.
+These integrations are enabled by default and integrate into the
+standard library or the interpreter itself. They are documented so you can be both aware of what they do and disable them if they cause issues.
+
+To disable system integrations, set `default_integrations => false` when calling `init()`.
```
ste93cry, line 14: Actually, since in PHP we have namespaces the `init()` function resides under the `Sentry` namespace. Not sure it's worth mentioning it here, but I see that our examples are always explicit to avoid people copy-pasting to have errors so I will let you decide what to do. Should you decide to add the namespace to the snippet, remember to use the backslash as first character so that the resolving of the namespace start from the root

src/platforms/php/common/integrations.mdx
```diff
@@ -8,49 +8,42 @@ redirect_from:
...
+To do so, it ensures that Sentry's `ErrorHandler` is registered, and adds a callback to it as an error listener. By default, the `ErrorHandler` reserves 10 kilobytes of memory to handle fatal errors.
 
-For some frameworks or projects, there are specific integrations provided both
-officially and by third-party users that automatically register the error
-handlers. In that case, please refer to their documentation.
+For some frameworks or projects, specific integrations are provided both officially and by third-party users that automatically register the error handlers. Please refer to their documentation.
 
-Also, by default `E_ALL` will be handled, you can change it by setting `error_types`
-a different constant in `init()`.
+Also, by default `E_ALL` will be handled. You can change it by setting `error_types` to a different constant in `init()`.
```
ste93cry, line 30: This does not hold true anymore. We changed a while back the default of this option so that whatever is the current PHP reporting level we use it. You can still explicitly set the option to a fixed value of course

src/platforms/php/common/integrations.mdx
```diff
@@ -90,14 +82,13 @@ This integration decides whether an event should not be captured according to a
...
 
-This integration logs with the event details all the versions of the packages installed with Composer; the root project is included too.
-
+This integration logs all the versions of the packages installed with Composer with the event details; the root project is also included.
 
-## Customize Integrations
+## Custom Integrations
 
-It is also possible to customize the list of integration without disabling the default integration using the `integrations` option and passing a callable.
+You can customize the list of integration without disabling the default integration using the `integrations` option and passing a callable.
```
ste93cry, line 89: While it's true that you can pass a `callable` if you have advanced needs, you can also pass a fixed list of integrations that you want to use which should be enough for most of the use cases

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
```diff
@@ -0,0 +1,62 @@
+---
+title: Performance Integrations
+description: "Learn about the the integrations Sentry provides for performance monitoring and how to configure them."
+sidebar_order: 500
+---
+
+The Symfony SDK offers integrations that support distributed tracing, including database queries, Twig template rendering, and cache pools. These integrations are enabled by default.
```
ste93cry, line 7: As already asked above, can we be more generic on this so that we won't have to remember that this note needs to be updated all the times we add a new integration?

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
```diff
@@ -0,0 +1,62 @@
...
+It is enabled by default. To disable tracking database queries, set `enabled` to `false`.
+
+```yaml {filename:config/packages/sentry.yaml}
+sentry:
+  tracing:
+    dbal: # DB queries
+      enabled: true
+```
+
+### twig
```
ste93cry, line 38:
```suggestion
### Twig
```

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
```diff
@@ -0,0 +1,62 @@
...
+It is enabled by default. To disable it, set `enabled` to `false`.
+
+```yaml {filename:config/packages/sentry.yaml}
+sentry:
+  tracing:
+    twig: # templating engine
+      enabled: true
+```
+
+### cache
```
ste93cry, line 51:
```suggestion
### Cache
```

ste93cry (changes requested):

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
PeloWriter: Yes, I agree, it makes the content more scalable.

src/platforms/php/guides/symfony/index.mdx
PeloWriter: Yes, let's make this more generic.

I'll make the Note a Warning, and add the detail you've provided (succinctly in the warning and more robustly on the Integrations page), thanks.

src/platforms/php/common/integrations.mdx
PeloWriter: I think it is worth mentioning, thanks for look at this so carefully. I've updated so that it reads:

`\Sentry\init()` 

LMK if that's correct.

ste93cry: Yes, it is. Also, I'm pretty sure I saw some other code snippets that should be updated as well 😉

src/platforms/php/common/integrations.mdx
PeloWriter: updated to reflect that `E_ALL` defaults to the current PHP reporting level.

src/platforms/php/common/integrations.mdx
PeloWriter: I found a couple of other instances and updated those.

src/platforms/php/common/integrations.mdx
ste93cry: I'm sorry, I probably expressed myself badly: what I meant was that while earlier the `error_types()` option defaulted to that constant, since a while it defaults to the value returned by the `error_reporting()` function which can change at runtime. So, if the user adjust the level of errors to report during the application execution by default now Sentry will react accordingly. If the user sets the option to a fixed value instead, Sentry will always log the errors of that specific type regardless of the current error reporting level

ste93cry:
It's still wrong. Here the corrected version:

>By default, the `error_types` option defaults to the value returned by the `error_reporting()` function, which can change at runtime. Alternately, you can set the option to a fixed value by setting its value to a bitmask of the PHP `E_*` constants in `\Sentry\init()`. In this case, Sentry will log only errors of those specific types regardless of the current reporting level.

src/platforms/php/common/integrations.mdx
ste93cry: See https://github.com/getsentry/sentry-docs/pull/3470#issuecomment-830426263 for the continuation to the discussion

src/platforms/php/common/integrations.mdx
```diff
@@ -90,14 +83,13 @@ This integration decides whether an event should not be captured according to a
...
 
-This integration logs with the event details all the versions of the packages installed with Composer; the root project is included too.
-
+This integration logs all the versions of the packages installed with Composer with the event details; the root project is also included.
 
-## Customize Integrations
+## Custom Integrations
 
-It is also possible to customize the list of integration without disabling the default integration using the `integrations` option and passing a callable.
+You can customize the list of integration without disabling the default integration using the `integrations` option. You can pass a callable in advanced needs, though a fixed list of integrations will work in use cases.
```
Jean85, line 90: Am I wrong or this is what was intended?
```suggestion
You can customize the list of integration without disabling the default integration using the `integrations` option. You can pass a callable for advanced needs, though a fixed list of integrations will work in most use cases.
```

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
```diff
@@ -0,0 +1,64 @@
...
+    enabled: true
+    dbal: # DB queries
+      enabled: true
+    cache:
+      enabled: true
+    twig: # templating engine
+      enabled: true
+```
+
+You can adjust the integrations to your needs if you want to disable a specific integrations.
```
Jean85, line 25:
```suggestion
You can adjust the integrations to your needs if you want to disable a specific integration.
```

src/platforms/php/guides/symfony/performance/pm-integrations.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+### Dbal
```
Jean85, line 27: It's an acronym
```suggestion
### DBAL
```

Jean85 (changes requested):

HazAT (approved):
I think this is good enough to merge for now.

ste93cry (approved):

PR #3490, opened 2021-04-29
[dotnet] Add migration about serialization

bruno-garcia:
Relates to: https://github.com/getsentry/sentry-dotnet/issues/962

### DIFF
--- a/src/platforms/dotnet/migration.mdx
+++ b/src/platforms/dotnet/migration.mdx
@@ -25,7 +25,7 @@ options.StackTraceMode = StackTraceMode.Original;
 ### Serialization
 
 In version 3.0.0, the Sentry SDK for .NET changed from `Newtonsoft.Json` to `System.Text.Json`.
-Using the new library has many advantages. One of which is the lack of an additional dependency (when targeting .NET Core 3, .NET 5 and later versions) and much faster performance.
+Using the new library has many advantages. These include the lack of an additional dependency (when targeting .NET Core 3, .NET 5 and later versions) and much faster performance.
 
 One breaking change that affects objects set as [Context](../enriching-events/context/) and `Extra` is that `Newtonsoft.Json` was previously able to serialize some types, which is not possible with `System.Text.Json`.
 
@@ -44,7 +44,7 @@ Setting an instance of `Exception` that was caught in a `catch` block could fail
    at System.Text.Json.Serialization.JsonConverter`1.WriteCore(Utf8JsonWriter writer, T& value, JsonSerializerOptions options, WriteStack& state)
 ```
 
-To avoid such error, make sure to pass a serializable object into Sentry. For example, map the instance of `Exception` to an anonymous object:
+To avoid such an error, pass a serializable object into Sentry. For example, map the instance of `Exception` to an anonymous object:
 
 ```
 try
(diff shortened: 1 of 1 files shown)


src/platforms/dotnet/migration.mdx
```diff
@@ -22,6 +22,51 @@ One big caveat is that this will **affect grouping!**. If you prefer to stay wit
 options.StackTraceMode = StackTraceMode.Original;
 ```
 
+### Serialization
+
+On version 3.0.0 the Sentry SDK for .NET changed from `Newtonsoft.Json` to `System.Text.Json`.
+This introduces some advantages, such as no dependency is needed (when targeting .NET Core 3.0, .NET 5 and later version) and much faster performance.
+
+One breaking change in behavior that came up was that `Newtonsoft.Json` was able to serialize some types that are not possible with `System.Text.Json`.
+This affects objects set as [Context](../enriching-events/context/) and `Extra`.
```
imatwawana, line 31:
```suggestion
One breaking change that affects objects set as [Context](../enriching-events/context/) and `Extra` is that `Newtonsoft.Json` was previously able to serialize some types, which is not possible with `System.Text.Json`.
```

imatwawana (approved):
Minor wording edits for clarity/style. One area where there's a question. Otherwise LGTM.

src/platforms/dotnet/migration.mdx
```diff
@@ -22,6 +22,51 @@ One big caveat is that this will **affect grouping!**. If you prefer to stay wit
 options.StackTraceMode = StackTraceMode.Original;
 ```
 
+### Serialization
+
+On version 3.0.0 the Sentry SDK for .NET changed from `Newtonsoft.Json` to `System.Text.Json`.
+This introduces some advantages, such as no dependency is needed (when targeting .NET Core 3.0, .NET 5 and later version) and much faster performance.
```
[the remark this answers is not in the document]
bruno-garcia: I was hoping to describe that the SDK has "no dependency". Let me see if I can reword it:

bruno-garcia, line 28:
```suggestion
Using the new library has many advantages. One of which is the lack of an additional dependency (when targeting .NET Core 3, .NET 5 and later versions) and much faster performance.
```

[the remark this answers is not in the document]
bruno-garcia: Is the new version a bit better?

src/platforms/dotnet/migration.mdx
```diff
@@ -24,13 +24,12 @@ options.StackTraceMode = StackTraceMode.Original;
 
 ### Serialization
 
-On version 3.0.0 the Sentry SDK for .NET changed from `Newtonsoft.Json` to `System.Text.Json`.
-This introduces some advantages, such as no dependency is needed (when targeting .NET Core 3.0, .NET 5 and later version) and much faster performance.
+In version 3.0.0, the Sentry SDK for .NET changed from `Newtonsoft.Json` to `System.Text.Json`.
+Using the new library has many advantages. One of which is the lack of an additional dependency (when targeting .NET Core 3, .NET 5 and later versions) and much faster performance.
```
imatwawana, line 28:
```suggestion
Using the new library has many advantages. These include the lack of an additional dependency (when targeting .NET Core 3, .NET 5 and later versions) and much faster performance.
```

PR #3491, opened 2021-04-29
fix(ui): Fix content reflow on hover

leedongwei:
Visible when the browser/page width is 1280px.

image


imatwawana (approved):
Looks great! Thanks for the fix!

PR #3492, opened 2021-04-29
fix: Method name toTraceparent => toSentryTrace

rhcarvalho:
The method name has changed in the developer documentation / Performance spec late last year, and some SDKs have only the newer name, some only the older name.

Therefore now we list both possible names here with the intention to help people find the right method in their SDK of choice.

See https://github.com/getsentry/develop/pull/193 (where the name change happened)
See also https://github.com/getsentry/sentry-ruby/issues/1411 (an example of confusion caused by current docs)


src/platforms/common/performance/connect-services.mdx
```diff
@@ -28,13 +28,19 @@ For traces that begin in your backend, you can connect the automatically-generat
...
   <head>
-    <meta name="sentry-trace" content="{{ span.toTraceparent() }}" />
+    <meta name="sentry-trace" content="{{ span.toSentryTrace() }}" />
     <!-- ... -->
   </head>
 </html>
 ```
 
-The `name` attribute must be the string `"sentry-trace"` and the `content` attribute must be generated by your backend's Sentry SDK using `span.toTraceparent()` (or equivalent, depending on the backend platform). This guarantees that a new and unique value will be generated for each request.
+The `name` attribute must be the string `"sentry-trace"` and the `content` attribute must be generated by your backend's Sentry SDK using `span.toSentryTrace()`  (or equivalent, depending on the backend platform). This guarantees that a new and unique value will be generated for each request.
```
imatwawana, line 37:
```suggestion
The `name` attribute must be the string `"sentry-trace"` and the `content` attribute must be generated by your backend's Sentry SDK using `span.toSentryTrace()` (or equivalent, depending on the backend platform). This ensures that a new and unique value will be generated for each request.
```

imatwawana (approved):
Removed extra space

rhcarvalho:
> Removed extra space

Good catch! Thanks!

PR #3495, opened 2021-04-29
Document update for sentry-ruby 4.4.0

st0012:


st0012:
This one is ready for review 🙂  @rhcarvalho @HazAT @PeloWriter

src/platforms/ruby/common/configuration/options.mdx
```diff
@@ -82,18 +82,22 @@ end
 
 `breadcrumbs_logger`
 
-: Sentry supports two breadcrumbs loggers in the Ruby SDK:
-- `SentryLogger` - A general breadcrumbs logger for all Ruby applications.
-- `ActiveSupportLogger` - Built on top of Rails ActiveSupport's instrumentation and provides many Rails-specific information.
+: Sentry supports different breadcrumbs loggers in the Ruby SDK:
+- `:sentry_logger` - A general breadcrumbs logger for all Ruby applications.
+- `:active_support_logger` - Built on top of Rails ActiveSupport's instrumentation and provides many Rails-specific information.
+- `:http_logger` - It captures requests made with the standard `net/http` library.
```
PeloWriter, line 88: the `:` is displaying for these. Do we mean it to? 
Screen Shot 2021-05-03 at 1 59 49 PM

PeloWriter (approved):
some minor format issues. But otherwise, great job, @st0012 - thank you.

src/platforms/ruby/common/configuration/options.mdx
st0012: Yes, that means Ruby's `symbol` type.

src/platforms/ruby/common/configuration/integration_options.mdx
```diff
@@ -23,3 +23,31 @@ config.rails.report_rescued_exceptions
...
 config.rails.skippable_job_adapters = ["ActiveJob::QueueAdapters::MyAdapter"]
 ```
+
+
+`rails.tracing_subscribers`
+
+: Tracing subscribers are responsible for capturing tracing spans from Rails instrumentation.
+
+
+The default subscribers are:
```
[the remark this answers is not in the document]
st0012: It's not possible to indent this line so I just merged it with the previous line.

st0012:
@PeloWriter I've updated the PR. I'll ask @HazAT to merge this when we rollout the new version later today. Thanks for the quick review 👍

src/includes/performance/traces-sampler-as-filter/ruby.mdx
```diff
@@ -0,0 +1,38 @@
+```ruby
+Sentry.init do |config|
+  config.traces_sampler = lambda do |sampling_context|
+    # if this is the continuation of a trace, just use that decision (rate controlled by the caller)
+    unless sampling_context[:parent_sampled].nil?
+      next sampling_context[:parent_sampled]
```
rhcarvalho, line 6: 👍  just learned `next` here works like `return`.

src/includes/performance/traces-sampler-as-filter/ruby.mdx
```diff
@@ -0,0 +1,38 @@
...
+        0.1
+      end
+    when /sidekiq/
+      0.01 # you may want to set a lower rate for background jobs if the number is large
+    else
+      0.0 # ignore all other transactions
+    end
+  end
+end
+```
```
rhcarvalho, line 38: Why are `src/includes/performance/traces-sampler-as-filter/ruby.mdx` and `src/includes/performance/traces-sampler-as-sampler/ruby.mdx` identical?

I'm hesitant to add two files that are exactly the same.

Looking at the examples for other platforms, they are different. The `traces-sampler-as-filter` example focuses on dropping certain transactions when a condition is met. The `traces-sampler-as-sampler` focuses on returning different rates depending on transaction.

src/platforms/ruby/common/configuration/integration_options.mdx
```diff
@@ -23,3 +23,29 @@ config.rails.report_rescued_exceptions
...
+config.rails.tracing_subscribers << MySubscriber
+# or replace the set completely
+config.rails.tracing_subscribers = [MySubscriber]
+```
+
+## sentry-delayed_job
+
+`delayed_job.report_after_job_retries`
+
+: `sentry-delayed_job` captures exceptions whenever they occur in a job execution. You can make it wait utill the retry is exhausted. Default is `false`.
```
rhcarvalho, line 47:
```suggestion
: `sentry-delayed_job` captures exceptions whenever they occur in a job execution. You can make it wait until the retry is exhausted. Default is `false`.
```

---

Is it really `sentry-delayed_job` with `-` and `_` in the same word?

src/platforms/ruby/common/configuration/options.mdx
```diff
@@ -216,7 +220,7 @@ config.trusted_proxies = ["2.2.2.2"]
 
 : You can gain more control on tracing event (transaction)'s sampling decision by providing a callable object (`Proc` or `Lambda`) as a `traces_sampler`:
 
-<PlatformContent includePath="performance/dynamic-sampling-function-intro" />
+<PlatformContent includePath="performance/traces-sampler-as-sampler" />
```
rhcarvalho, line 223: 👍  Good fix, the page was renamed to `src/includes/performance/sampling-function-intro/ruby.mdx` in 3e3980e907fc62d707d408c8c193c2e6cd183fb2 and so this got broken. Surprised the Gatsby build and CI has not caught this bad reference.

rhcarvalho (changes requested):
Looks good. Most important change is to fix the traces sampler examples to match other platforms and the intention of the examples.

src/includes/getting-started-config/ruby.delayed_job.mdx
```diff
@@ -3,6 +3,6 @@ Initialize the SDK:
 ```ruby
 Sentry.init do |config|
   config.dsn = '___PUBLIC_DSN___'
-  config.breadcrumbs_logger = [:sentry_logger]
+  config.breadcrumbs_logger = [:sentry_logger, :http_logger]
```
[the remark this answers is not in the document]
st0012: @rhcarvalho If we're talking about `sentry_logger`, it really depends on what the libraries the user has. 
If the application already uses Rails and has `active_support_logger`, having `sentry_logger` would generally create noise. For example, my company's project doesn't use it.

But if the application does something like import/export operation with logging, having it could be useful. That's why I added it to the [rubyapi](https://github.com/rubyapi/rubyapi/pull/739) project.

As for `http_logger`, because it's a newly added logger and didn't exist in the old SDK. So I want to see how our users think about it before making it a default.

src/platforms/ruby/common/configuration/integration_options.mdx
st0012: yes. `delayed_job` is the gem's name and I don't want to change it.

src/includes/performance/traces-sampler-as-filter/ruby.mdx
st0012: Because I think the example happen to cover both of the cases.

st0012: I've simplified this example to make it focus more on filtering.

rhcarvalho (approved):

st0012:
@rhcarvalho thanks for the review, can you help me merge this too 😄 ?

PR #3521, opened 2021-05-05
Update JS family import statements

PeloWriter:
This PR introduces a minor change to the Enriching Events pages because the JS family of content did not have `import` statements. To avoid copy and pasting, the content now allows for a unique import statement for these SDKs, but re-uses the JS example as appropriate.

This PR also restructures the `include` statements that support Enriching Events to nest them in that directory as the `include` directory was becoming unwieldy.


src/includes/enriching-events/import/javascript.cordova.mdx
```diff
@@ -0,0 +1,4 @@
+```javascript
+onDeviceReady: function() {
+  var Sentry = cordova.require("sentry-cordova.Sentry");
```
HazAT, line 3:
```suggestion
var Sentry = cordova.require("sentry-cordova.Sentry");
```

src/includes/enriching-events/import/javascript.ember.mdx
```diff
@@ -0,0 +1,3 @@
+```javascript
+import { InitSentryForEmber } from "@sentry/ember";
```
HazAT, line 2: @k-fish I think this has to change, is there `import * as Sentry from "@sentry/ember";`

HazAT (changes requested):
I couldn't find node includes, it's also true for them.

PeloWriter:
you're right, Node is falling back to JS, but I will switch the import statement to `import * as Sentry from "@sentry/node";`

src/includes/enriching-events/import/javascript.ember.mdx
PeloWriter: If so, we should fix it both here and in the Config sample, since that's where I copied it from.

k-fish: Yep `import * as Sentry from "@sentry/ember";` should work, you'll have to change it in the config sample, or I can if you like.

k-fish: Actually I'll just make a commit and you can revert it / edit it or whatever suits.

PeloWriter: thanks, @k-fish

Issue #3507, opened 2021-05-03
Document missing code snippets for Apple SDK


marandaneto:
## Issue

Document missing code snippets for Apple SDK

eg https://docs.sentry.io/platforms/apple/guides/ios/data-management/event-grouping/sdk-fingerprinting/#group-errors-with-greater-granularity

## Description

The platform or SDK you've selected either does not support this functionality, or it is missing from documentation.

Apple SDK does support fingerprints though.

### Suggested Solution

Similar to Android and other SDKs
https://docs.sentry.io/platforms/android/data-management/event-grouping/sdk-fingerprinting/#group-errors-with-greater-granularity

PR #3524, opened 2021-05-05
feat(apple): Add RPC fingerprint sample

philipphofmann:
This sample was missing for Apple.

Fixes GH-3507

### DIFF
--- a/src/includes/set-fingerprint/rpc/apple.mdx
+++ b/src/includes/set-fingerprint/rpc/apple.mdx
@@ -3,9 +3,8 @@ import Sentry
 
 SentrySDK.start { options in
     options.beforeSend = { event in
-        if let error = event.error as NSError? {
-            if let failingUrl = error.userInfo[NSURLErrorFailingURLErrorKey] as? String,
-               error.domain == NSURLErrorDomain {
+        if let error = event.error as? NSError, error.domain == NSURLErrorDomain {
+            if let failingUrl = error.userInfo[NSURLErrorFailingURLErrorKey] as? String {
                 event.fingerprint = [
                     "{{ default }}",
                     failingUrl,
--- a/src/includes/set-fingerprint/rpc/apple.mdx
+++ b/src/includes/set-fingerprint/rpc/apple.mdx
@@ -3,7 +3,7 @@ import Sentry
 
 SentrySDK.start { options in
     options.beforeSend = { event in
-        if let error = event.error as? NSError, error.domain == NSURLErrorDomain {
+        if let error = event.error as NSError?, error.domain == NSURLErrorDomain {
             if let failingUrl = error.userInfo[NSURLErrorFailingURLErrorKey] as? String {
                 event.fingerprint = [
                     "{{ default }}",
@@ -28,7 +28,7 @@ SentrySDK.start { options in
                 event.fingerprint = @[
                     @"{{ default }}",
                     failingUrl,
-                    [NSString stringWithFormat:@"%i",event.error.code]
+                    @(event.error.code).stringValue
                 ];
             }
         }
--- a/src/includes/set-fingerprint/rpc/apple.mdx
+++ b/src/includes/set-fingerprint/rpc/apple.mdx
@@ -1,0 +1,39 @@
+```swift
+import Sentry
+
+SentrySDK.start { options in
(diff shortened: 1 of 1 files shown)


src/includes/set-fingerprint/rpc/apple.mdx
```diff
@@ -0,0 +1,39 @@
+```swift
+import Sentry
```
marandaneto, line 2: on Android/Java SDKs we're used to add all the necessary imports, because of https://develop.sentry.dev/sdk/philosophy/#assume-novice-developers
so people can copy/paste the code snippet and just works.

would be the case here for  `NSError`, `NSURLErrorFailingURLErrorKey` or are they inferred automatically in Swift/ObjC?

brustolin: In swift they are inferred automatically, not the case for objc.

But I don't think this is necessary, the snippet cannot be copied as it is to be used in an empty file, a little understanding of development is required to use the sample.

src/includes/set-fingerprint/rpc/apple.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+SentrySDK.start { options in
+    options.beforeSend = { event in
+        if let error = event.error as NSError? {
+            if let failingUrl = error.userInfo[NSURLErrorFailingURLErrorKey] as? String,
+               error.domain == NSURLErrorDomain {
```
brustolin, line 8:
```suggestion
        if let error = event.error as? NSError, error.domain == NSURLErrorDomain {
                    if let failingUrl = error.userInfo[NSURLErrorFailingURLErrorKey] as? String {
```

I would group the error domain check with the error casting.
Move "?" from the type (NSError?) to the cast (as?), is much safer.

src/includes/set-fingerprint/rpc/apple.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+[SentrySDK startWithConfigureOptions:^(SentryOptions *options) {
+    options.beforeSend = ^SentryEvent * _Nullable(SentryEvent * _Nonnull event) {
+        if ([event.error.domain isEqualToString:NSURLErrorDomain]) {
+            NSString *failingUrl = event.error.userInfo[NSURLErrorFailingURLErrorKey];
+            if (failingUrl != nil) {
+                event.fingerprint = @[
+                    @"{{ default }}",
+                    failingUrl,
+                    @(event.error.code).stringValue
```
brustolin, line 32: L: 
```suggestion
                    [NSString stringWithFormat:@"%i",event.error.code];
```

This is me nit picking, but you are boxing an integer in a NSNumber, and then converting it to string, when you could do it directly.

brustolin (approved):
I added a couple of nit picking but its look good to me.

src/includes/set-fingerprint/rpc/apple.mdx
bruno-garcia: > a little understanding of development is required to use the sample.

What this means can vary by person. We try to include imports so that the only understanding customers need is that the import stuff goes on the top, and the snippet elsewhere. And hence remove the chance of using copying code then trying to build or run and not being able to. Which happened in the past.

philipphofmann: Thanks for pointing that out, @marandaneto. `import Sentry` is all you need for Sentry classes. `NSError` and `NSURLErrorFailingURLErrorKey` are part of the Foundation framework, so you don't need to add imports for them. The code sample compiles if you just have `import Sentry. So this is fine.

philipphofmann:
Thanks for pointing out some improvements @brustolin. I went back to `event.error as NSError?`, because Xcode gives you a compiler warning for this `Conditional downcast from 'Error?' to 'NSError' is a bridging conversion; did you mean to use 'as'?`.

I also changed `[NSString stringWithFormat:@"%i",event.error.code]` back to `@(event.error.code).stringValue`, because again Xcope complains and changes it to `[NSString stringWithFormat:@"%li",(long)event.error.code]`, which is not as nice as `@(event.error.code).stringValue`. I rather use an extra boxing so the code sample looks a bit nicer.

marandaneto (approved):

PR #3527, opened 2021-05-05
fix(performance): Remove sample rate advice

### LINKED ISSUE #1678: Review sampling rate advice
While updating the docs in https://github.com/getsentry/sentry-docs/pull/1667#discussion_r422064998, a conversation started about what advice we give for setting the sampling rate of different components, specially in the light of different load patterns.

The updated docs with the advice reads:

> If you enable tracing in services with multiple entry points, we recommend choosing similar sampling rates, to avoid biasing your data. [...]

And the discussion is that even setting all sampling rates to the same value may not help with sample bias.

<details>

This is a screenshot of the advice prior to the update (red underline to emphasize the advice):

https://docs.sentry.io/performance/distributed-tracing/#sampling-transactions

image

</details>

rhcarvalho:
Different sample rates in different interconnected Sentry projects do
not affect sample bias any more than other factors such as uneven load
patterns.

The current advice suggests that "similar sampling rates [...] avoid
biasing [...] data", but that's not a fact.

We could either change the paragraph to discuss sources of data bias and
the effects on summaries like latency percentiles in the product, or
remove the paragraph altogether.

Since our docs are not the best place to discuss statistics, opting for
the latter.

The sample rate can be anything users want and keep them within their
quota. We even document how to have custom sample rates for different
endpoints as users see fit (tracesSampler callback).

Closes #1678.

### DIFF
--- a/src/docs/product/performance/distributed-tracing.mdx
+++ b/src/docs/product/performance/distributed-tracing.mdx
@@ -248,10 +248,6 @@ On the other hand, when your backend processes the requests from B's browser, it
 
 Put simply: as a result of this head-based approach, where the decision is made once in the originating service and passed to all subsequent services, either all of the transactions for a given trace are collected, or none are, so there shouldn't be any incomplete traces.
 
-### Consistency Between Traces
-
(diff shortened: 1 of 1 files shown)


silent1mezzo (approved):
Works for me.

PeloWriter (approved):
I agree. Thanks for following up on this, @rhcarvalho

Issue #3254, opened 2021-03-16
RH sidebar needs to scroll


PeloWriter:
## Issue

This an issue with:
Page scrolling with all areas.

## Description

The right column "on this page" / "common options" https://docs.sentry.io/platforms/java/configuration/options/ isn't scrollable. SDKs that have a bunch of options make the RH sidebar unusable on smaller screens.

The RH sidebar is fully rendering but doesn't respect screen size, so you can only see it if you zoom out. For example, on the page listed above, on 22" monitor at full screen, you can only see the full RH sidebar menu if you zoom out to 67% .

![F835CF89-5E0E-4C02-BF4D-B18F44BA8DBF_1_105_c](https://user-images.githubusercontent.com/76437239/111522427-a9b01000-8730-11eb-87cf-a293b8436d53.jpeg)

### Suggested Solution

Enable scrolling in the RH sidebar.

rohitkataria-sentryio:
On this page, for instance: https://docs.sentry.io/platforms/python/data-management/event-grouping/#grouping-by-stack-trace ,one has to Zoom out to 50% to view all the clickable options on the right side bar.  
Several other pages have this issue. 
The problem accentuates if the "upcoming workshop banner" is also present.

rohitkataria-sentryio:
Leaving a comment to prevent auto-close

dashed:
For posterity, this might be the fix. Leaving it for anyone willing to apply them:

<img width="1903" alt="Screen Shot 2021-04-13 at 4 40 44 PM" src="https://user-images.githubusercontent.com/139499/114619051-1c50d480-9c78-11eb-939a-03f3ff174a51.png">

PR #3539, opened 2021-05-06
fix: Enable right side to scroll vertically

dashed:
Fixes https://github.com/getsentry/sentry-docs/issues/3254

https://user-images.githubusercontent.com/139499/117362191-f5597d00-ae88-11eb-9dc8-96839a4e01bf.mp4

### DIFF
--- a/src/css/_includes/grid.scss
+++ b/src/css/_includes/grid.scss
@@ -60,6 +60,9 @@ body {
 .page-nav {
   position: fixed;
   width: 27vw;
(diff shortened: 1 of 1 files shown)


PeloWriter:
Thank you, @dashed!

PeloWriter (approved):
Tested with both Python and .NET (two of our longest scrolling RH sidebars) as well as Basic Options. Many thanks for the fix!

PR #3541, opened 2021-05-07
feat(node): Add docs for server side release health

ahmedetefy:
This PR:
- Adds docs on how Release health in server-side (node) operates

### DIFF
--- a/src/includes/configuration/auto-session-tracking/javascript.mdx
+++ b/src/includes/configuration/auto-session-tracking/javascript.mdx
@@ -1,5 +1,3 @@
-### Client-Side
-
 We create a session for every page load. For single-page applications, we will create a new session for every navigation change (History API).
 
 We mark the session as:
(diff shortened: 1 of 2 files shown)


HazAT:
We should be able to use `node.jsx` include as an import and have the separation there. So we don't need to write server-side, browser-side.

src/includes/configuration/auto-session-tracking/javascript.mdx
```diff
@@ -1,8 +1,39 @@
...
+```
+
+Detection of which mode Release Health will operate in is done automatically by the SDK. By default, the SDK runs in Single Sessions mode, but if the `requestHandler` express middleware is enabled, Release Health mode is automatically toggled to Session Aggregates mode.
+
+It is important to note that if a web server application process is being run for example, but the `requestHandler` express middleware is not enabled, Session Aggregates mode will not be enabled and sessions will represent the health of the web server application process.
 
 We mark the session as crashed if an _unhandled error_ or _unhandled promise rejection_ bubbled up to the global handler.
 
-We mark the session as an error if the SDK captures an event that contains an exception (this includes manually captured errors). 
+We mark the session as an error if the SDK captures an event that contains an exception (this includes manually captured errors).
```
iker-barriocanal, line 36:
```suggestion
We mark the session as an error if the SDK captures an event that contains an exception (this includes manually captured exceptions).
```
?

rhcarvalho: We use "errors" all over the place in docs. Don't remember if we have a particular guidance here, couldn't find one in https://docs.sentry.io/contributing/ for now, but maybe @getsentry/docs can own the choice of words here.

PeloWriter (changes requested):
I think we may have duplicated information on the page (lines 5 and 36, lines 9 and 38). If so, we should have that content as a separate section that introduces what we capture, then cover client v. server side.

PR #3544, opened 2021-05-07
fix: Scrollable right sidebar should be responsive friendly

dashed:
Enable scrollable right sidebar introduced in https://github.com/getsentry/sentry-docs/pull/3539 on certain screen widths


imatwawana (approved):
Thanks for the fix!

PR #3549, opened 2021-05-07
ref(javascript): Update manual instrumentation docs

wmak:
- For automatic spans configureScope must be called, calling that out
  specifically so that users are more aware of this
- Also show how to get hub too

### DIFF
--- a/src/includes/performance/add-spans-example/javascript.mdx
+++ b/src/includes/performance/add-spans-example/javascript.mdx
@@ -5,7 +5,8 @@ For example, if you want to create a transaction for a user interaction on your 
 shopCheckout() {
   // This will create a new Transaction for you
   const transaction = Sentry.startTransaction({ name: "shopCheckout" });
-  // Required to get automatic spans, but this will end any existing transaction
+  // Set transaction on scope to associate with errors and get included span instrumentation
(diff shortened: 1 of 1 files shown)


src/includes/performance/add-spans-example/javascript.mdx
```diff
@@ -5,8 +5,8 @@ For example, if you want to create a transaction for a user interaction on your
 shopCheckout() {
   // This will create a new Transaction for you
   const transaction = Sentry.startTransaction({ name: "shopCheckout" });
-  // set the transaction on the scope so it picks up any errors
-  hub.configureScope(scope => scope.setSpan(transaction));
+  // Required to get automatic spans, but this will end any existing transaction
```
rhcarvalho, line 8: It won't really "end" an existing transaction, not the same way as calling `transaction.finish()`.

Unfortunately, getting this right is a lot more nuanced, requiring more code to handle what to do with an open transaction already on the scope.

This is going to change 🤞

wmak: Hmm would `drop` be a better word here? The existing transaction is lost right? Actually should we include code that checks for one that exists + finishes any existing transaction here?

PR #3572, opened 2021-05-12
Use withScope docs from Java for Android

marandaneto:


philipphofmann (commented):
Is an empty file really the way to go @marandaneto?

marandaneto:
> Is an empty file really the way to go @marandaneto?

its not empty, but deleted, so it will inherit the java.mdx

PeloWriter:
Yes, Java is the fallback for Android, so if there's no Android snippet, we use the one from Java. You can also visually inspect that it's displaying properly to confirm - https://sentry-docs-dz1l56aiz.sentry.dev/platforms/android/enriching-events/scopes/

bruno-garcia:
I confirmed the docs did inherit as we wanted but I just realized that's not going to work on Android since push/pop (which withScope relies is a no-op.
Closing this one then. Lets keep that note

marandaneto:
@bruno-garcia the idea was to fix `withScope` for Android, before merging this, https://github.com/getsentry/sentry-java/pull/1472
I still think we should do it.

PR #3574, opened 2021-05-12
Remove deprecated `IncludeRequestPayload` option

MartinZid:
Remove deprecated `IncludeRequestPayload` option from example and documentation, use `MaxRequestBodySize` instead.

### DIFF
--- a/src/platforms/dotnet/guides/aspnetcore/index.mdx
+++ b/src/platforms/dotnet/guides/aspnetcore/index.mdx
@@ -71,7 +71,7 @@ For example:
 ```json {filename:appsettings.json}
   "Sentry": {
     "Dsn": "___PUBLIC_DSN___",
-    "IncludeRequestPayload": true,
+    "MaxRequestBodySize": "Always",
     "SendDefaultPii": true,
     "MinimumBreadcrumbLevel": "Debug",
     "MinimumEventLevel": "Warning",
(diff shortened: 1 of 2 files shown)


src/platforms/dotnet/guides/aspnetcore/index.mdx
```diff
@@ -232,9 +232,16 @@ Methods that take precedence over `IHostingEnvironment` are:
...
 
-Opt-in to send the request body to Sentry. By enabling this flag, all requests will have the `EnableRewind` method invoked. This is done so that the request data can be read at a later point in case an error happens while processing the request.
+This parameter controls if integrations should capture HTTP request bodies. It can be set to one of the following values:
+
+- `never`: request bodies are never sent
+- `small`: only small request bodies will be captured where the cutoff for small depends on the SDK (typically 4KB)
+- `medium`: medium and small requests will be captured (typically 10KB)
+- `always`: the SDK will always capture the request body for as long as Sentry can make sense of it
+
+If the request bodies should be captured, all requests will have the `EnableRewind` method invoked. This is done so that the request data can be read at a later point in case an error happens while processing the request.
```
imatwawana, line 244:
```suggestion
This parameter controls whether integrations should capture HTTP request bodies. It can be set to one of the following values:

- `never`: Request bodies are never sent.
- `small`: Only small request bodies will be captured. The cutoff for small depends on the SDK (typically 4KB).
- `medium`: Medium and small requests will be captured (typically 10KB).
- `always`: The SDK will always capture the request body as long as Sentry can make sense of it.

If the request bodies should be captured, all requests will have the `EnableRewind` method invoked. This is done so that the request data can be read later, in case an error happens while processing the request.
```

imatwawana (approved):
Made some minor wording suggestions

src/platforms/dotnet/guides/aspnetcore/index.mdx
MartinZid: I do like these suggestions, however I copied this part of documentation for [here](https://docs.sentry.io/platforms/dotnet/guides/aspnetcore/configuration/options/#max-request-body-size) (so it is consistent).

So if we commit your improvements, I should also make changes to the part of documentation linked above.

imatwawana: Assuming @bruno-garcia has no objections, you're right, it should be updated in both spots. Let's wait for review from @bruno-garcia first. Thanks for pointing that out!

bruno-garcia (approved):
Great catch! Thanks a lot for this PR

src/platforms/dotnet/guides/aspnetcore/index.mdx
bruno-garcia: Makes sense to align them both. Thanks for raising this

imatwawana (approved):
Looks great! Thanks so much!

PR #3581, opened 2021-05-13
Improve Ruby SDK's documents

st0012:
I hope this makes examples easier to understand for users with different setups.

Related issues:
- https://github.com/getsentry/sentry-ruby/issues/1443
- https://github.com/getsentry/sentry-ruby/issues/1429


src/includes/configuration/before-send/ruby.rails.mdx
```diff
@@ -0,0 +1,12 @@
+```ruby
+Sentry.init do |config|
+  # this example uses Rails' parameter filter to sanitize the event payload
```
st0012, line 3: I chose this as Rails' example because it's the most requested feature that we don't support (client side data scrubbing).

PR #3585, opened 2021-05-14
hide javascript docs from mobile

bruno-garcia:


src/platforms/common/performance/index.mdx
```diff
@@ -85,7 +85,7 @@ Leaving the sample rate at `1.0` means that included instrumentation will send a
 
 <!-- notSupported here should be the opposite of whats in common/connect-services.mdx -->
 
-<PlatformSection notSupported={["javascript", "javascript.cordova"]}>
+<PlatformSection notSupported={["javascript", "javascript.cordova", "dotnet.xamarin", "dotnet.xamarin", "android", "react-native", "apple", "flutter", "unity", "capacitor" ]}>
```
bruno-garcia, line 88:
```suggestion
<PlatformSection notSupported={["javascript", "javascript.cordova", "dotnet.xamarin", "android", "react-native", "apple", "flutter", "unity", "capacitor" ]}>
```

src/platforms/common/performance/index.mdx
```diff
@@ -85,7 +85,7 @@ Leaving the sample rate at `1.0` means that included instrumentation will send a
 
 <!-- notSupported here should be the opposite of whats in common/connect-services.mdx -->
 
-<PlatformSection notSupported={["javascript", "javascript.cordova"]}>
+<PlatformSection notSupported={["javascript", "javascript.cordova", "dotnet.xamarin", "android", "react-native", "apple", "flutter", "unity", "capacitor" ]}>
```
bruno-garcia, line 88:
```suggestion
<PlatformSection notSupported={["javascript", "javascript.cordova", "dotnet.xamarin", "dotnet.wpf", "dotnet.winforms", "android", "react-native", "apple", "flutter", "unity", "capacitor" ]}>
```

wmak (approved):

wmak (commented):
Sentry.io will link here whenever we see a trace that is missing a connected service, could we add a mobile specific section here?

bruno-garcia:
With that in mind I believe we should always have the heading, and only change the content.
These already exist but are located in Included Instrumentation at the moment.

Some examples:
* [Xamarin](https://docs.sentry.io/platforms/dotnet/guides/xamarin/performance/included-instrumentation/#propagating-a-trace)
* [Android](https://docs.sentry.io/platforms/android/performance/included-instrumentation/#okhttp-library-instrumentation)

bruno-garcia:
@PeloWriter is this something you can help us with?

Currently the product links to this page/anchor when a transaction is found not to be connected to other transactions (did I get that right @wmak ?).

If it's a JavaScript project, it links to the `https://docs.sentry.io/platforms/javascript/performance/connect-services/` for everything else (assumed to be a backend) links to `https://docs.sentry.io/platforms/{PLATFORM}/performance/#connecting-services`.

That means that platforms such as: `Android, macOS, iOS, Flutter, Cordova, Capacitor, Unity, WPF, WinForms, Qt, Xamarin, React Native` will show this content (designed for backend (Django, Spring, ASP.NET) platforms).

I wonder if there's somethign we can do low maintenance here. Like a generic text. Or we need to have overrides and `if else` on those.

PeloWriter:
If I understand correctly, we want to keep the anchor link from sentry.io, but modify the text per platform because other platforms keep this content on their Included Instrumentation page.

In that case, 

- keep the unSupported list to only javascript and javascript.cordova as the content works for those SDK as is
- Insert, prior to line 96, "If you are using one of our mobile SDKs, we cover connecting services on the <PlatformLink to="/performance/custom-instrumentation/">Custom Instrumentation content</PlatformLink>.

A bit more work, but more elegant, would be to pull the content beneath ##Connecting Services into an include statement, which could then be modified for each SDK/type of SDK pretty easily. I can help with this, if that's the route you'd prefer to go.

Issue #3566, opened 2021-05-11
Document local scopes for Dart/Flutter


marandaneto:
## Issue

https://github.com/getsentry/sentry-dart/releases/tag/5.1.0-beta.1 via https://github.com/getsentry/sentry-dart/pull/463 can now use withScope but its not documented.

https://docs.sentry.io/platforms/dart/enriching-events/scopes/#local-scopes

## Description

Document behavior like Apple does https://docs.sentry.io/platforms/apple/enriching-events/scopes/#local-scopes

### Suggested Solution

Document behavior like Apple does https://docs.sentry.io/platforms/apple/enriching-events/scopes/#local-scopes

marandaneto:
cc @ueman let's add a note that this is beta still, how to add a note you can learn from this page, https://docs.sentry.io/platforms/java/guides/spring/advanced-usage/ there's a note about the feature being beta too

PeloWriter:
We took a look at the language for Beta and Early Adopter notes, and updated to:

This feature is available as a Beta. Features available in Beta are still in-progress and may have bugs. We recognize the irony. 

We wrap this in the usual Note /Note markdown tags (which I can't fully copy here as GH then makes them Notes :)).

I will update our Contributor Guidelines (thanks for alerting me to this being missing)

PR #3593, opened 2021-05-17
feat(Dart): With Scope Docs

ueman:
Fixes https://github.com/getsentry/sentry-docs/issues/3566

cc @marandaneto

### DIFF
--- a/src/includes/enriching-events/scopes/with-scope/dart.mdx
+++ b/src/includes/enriching-events/scopes/with-scope/dart.mdx
@@ -1,0 +1,22 @@
+<Note>
+
+This feature is available as a Beta. Features available in Beta are still in-progress and may have bugs. We recognize the irony.
+
+This is available in 5.1.0.beta.1 version of the Sentry Dart/Flutter SDK.
+
+</Note>
+
+```dart
+import 'package:sentry/sentry.dart';
+
+Sentry.captureMessage(
+    'This event has an extra tag',
+    withScope: (scope) {
+        // will be tagged with my-tag="my value"
+        scope.setTag('my-tag', 'my value');
+    },
+);
+
+// will not be tagged with my-tag
+Sentry.captureMessage('This event has no extra tag');
+```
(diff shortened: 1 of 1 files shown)


ueman:
See https://sentry-docs-git-fork-ueman-flutter-local-scopes.sentry.dev/platforms/dart/enriching-events/scopes/#local-scopes

src/includes/enriching-events/scopes/with-scope/dart.mdx
```diff
@@ -0,0 +1,22 @@
+<Note>
+
+This feature is available as a Beta. Features available in Beta are still in-progress and may have bugs. We recognize the irony.
+
+This is available in 5.1.0.beta.1 version of the Sentry Dart/Flutter SDK.
+
```
marandaneto, line 6:
```suggestion
```

as its the the beta.1, we don't need to mention that.\, it'd make sense if its available under beta.2, 3 etc

src/includes/enriching-events/scopes/with-scope/dart.mdx
```diff
@@ -0,0 +1,22 @@
...
+Sentry.captureMessage(
+    'This event has an extra tag',
+    withScope: (scope) {
+        // will be tagged with my-tag="my value"
+        scope.setTag('my-tag', 'my value');
+    },
+);
+
+// will not be tagged with my-tag
+Sentry.captureMessage('This event has no extra tag');
```
marandaneto, line 21: the description in the docs is:

> We also have support for pushing and configuring a scope in one go. This is typically called with-scope or push-scope which is also very helpful if you only want to send data with one specific event. In the following example we are using that function to attach a level and a tag to only one specific error:

so I suggest changing `captureMessage` to `captureException`, as its an error.
also, setting the scope.level.

this might be wrong in the other SDKs and would have to be aligned with the description too, if you copied this example from other SDK, lets raise an issue for that

ueman: Raised an issue: https://github.com/getsentry/sentry-docs/issues/3601

marandaneto (approved):

PR #3596, opened 2021-05-17
Install of sentry cli in pipeline must return true

lauer:
Ensure that the install of Sentry CLI always returns true, as the example on Freestyle project does

### DIFF
--- a/src/docs/product/releases/setup/release-automation/jenkins/index.mdx
+++ b/src/docs/product/releases/setup/release-automation/jenkins/index.mdx
@@ -63,7 +63,7 @@ pipeline {
             }
             steps {
                 // Install Sentry CLI
-                sh 'curl -sL https://sentry.io/get-cli/ | bash'
+                sh 'curl -sL https://sentry.io/get-cli/ | bash || true'
 
                 sh '''
                     export SENTRY_RELEASE=$(sentry-cli releases propose-version)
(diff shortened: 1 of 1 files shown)


imatwawana:
@jan-auer can you review?

manuzope:
Hey @lauer, thanks for opening this PR! I'm not very familiar with Jenkins but why would you always want to return `true`?

lauer:
It's only the install cmd. Because that will fail if the cli tool is already installed. 

If the install cmd fails the next cmd would fail anyway.

src/docs/product/releases/setup/release-automation/jenkins/index.mdx
```diff
@@ -63,7 +63,7 @@ pipeline {
             }
             steps {
                 // Install Sentry CLI
-                sh 'curl -sL https://sentry.io/get-cli/ | bash'
+                sh 'curl -sL https://sentry.io/get-cli/ | bash || true'
```
joshuarli, line 66:
```suggestion
                command -v sentry-cli || sh 'curl -sL https://sentry.io/get-cli/ | bash'
```

joshuarli, line 66: Is this more what you're trying to say? If sentry-cli isn't installed, then we should install it. This won't fail if sentry-cli isn't installed, assuming the installation succeeds.

```suggestion
                sh 'command -v sentry-cli || curl -sL https://sentry.io/get-cli/ | bash'
```

joshuarli (approved):

joshuarli:
Thanks @lauer!

Issue #3567, opened 2021-05-11
Migration docs for Sentry Java and Android related to null-safety


marandaneto:
## Issue

Migration docs for Sentry Java and Android related to null-safety
https://docs.sentry.io/platforms/android/migration/
and https://docs.sentry.io/platforms/java/migration/

## Description

Migration docs for Sentry Java and Android related to null-safety

### Suggested Solution

Add migration entry related to null-safety compat.

PR #3597, opened 2021-05-17
Java: polish Kotlin sample code.

maciejwalkowiak:
- Update Kotlin samples to match API nullability updated in Java SDK 5.0 (#3567)
- Fix minor mistakes in Kotlin samples.

### DIFF
--- a/src/includes/performance/set-transaction-current-scope/java.mdx
+++ b/src/includes/performance/set-transaction-current-scope/java.mdx
@@ -35,7 +35,7 @@ import io.sentry.Sentry
 
 fun executeTask() {
   val transaction = Sentry.startTransaction("transaction-name", "task")
-  Sentry.configureScope(scope -> scope.setTransaction(transaction))
+  Sentry.configureScope { it.transaction = transaction }
 
   executeSubTask()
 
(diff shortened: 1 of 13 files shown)


src/includes/performance/set-transaction-current-scope/java.mdx
```diff
@@ -35,7 +35,7 @@ import io.sentry.Sentry
 
 fun executeTask() {
   val transaction = Sentry.startTransaction("transaction-name", "task")
-  Sentry.configureScope(scope -> scope.setTransaction(transaction))
+  Sentry.configureScope { it.transaction = transaction }
```
marandaneto, line 38: enriching-events/scopes/with-scope/java.mdx changed to using a `setTransaction` instead of `transaction = transaction`, is that correct? or is it different because of the use in `configureScope` and `withScope`?

maciejwalkowiak: In Kotlin, `setTransaction` takes only `String`, `transaction = ` takes `ITransaction`. Perhaps we can make it more explicit, change `setTransaction(String)` into `setTransactionName` since it's a breaking change.

marandaneto: oh yeah true, they are not the same property

marandaneto (approved):
LGTM

marandaneto:
@maciejwalkowiak do you think we should already merge it as it fixes a few minor issues in the docs, thanks for that.
or should we wait for v5 GA?

bruno-garcia:
do we merge?

marandaneto:
> do we merge?

let's wait 5 GA, the code snippets wont work in current stable version

PR #3609, opened 2021-05-20
fix(nextjs): Clarify env variable usage on SDK init

iker-barriocanal:
Clarify the usage of environment variables to initialize the Next.js SDK. Environment variables have never been an alternative to not providing a DSN, but an alternative to not hardcode it.

Closes https://github.com/getsentry/sentry-javascript/issues/3565.

### DIFF
--- a/src/platforms/javascript/guides/nextjs/manual-setup.mdx
+++ b/src/platforms/javascript/guides/nextjs/manual-setup.mdx
@@ -52,7 +52,6 @@ export default withSentry(handler);
 
 You can hardcode your DSN and other configuration values directly inside your `Sentry.init` calls, or use use `process.env` placeholders, in these two files; or Next.js' built-in [environment variables](https://nextjs.org/docs/basic-features/environment-variables) support instead.
 
-
 ## Extend Default Webpack Usage for Source Maps
 
 Extend the default Next.js usage of Webpack to generate source maps and upload them to Sentry. Include the following in `next.config.js`:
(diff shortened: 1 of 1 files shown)


src/platforms/javascript/guides/nextjs/manual-setup.mdx
```diff
@@ -50,7 +50,7 @@ const handler = async (req, res) => {
 export default withSentry(handler);
 ```
 
-You can include your DSN directly in these two files, or provide it in either of two environment variables, `SENTRY_DSN` or `NEXT_PUBLIC_SENTRY_DSN`.
+You can provide your DSN to `Sentry.init` by hardcoding it directly or using environment variables (such as `SENTRY_DSN` or `NEXT_PUBLIC_SENTRY_DSN`), in these two files.
```
kamilogorek, line 53:
```suggestion
You can hardcode your DSN and other configuration values directly inside `Sentry.init` call, or use `process.env` placeholders, and Next.js built-in [environment variables](https://nextjs.org/docs/basic-features/environment-variables) support instead.
```

lobsterkatie: I thought the idea was we were going to look for the env vars ourselves, in the SDK. No?

If that's the case, then I think we should tell people, basically, either hardcode it in those files or put it in the following env variables and we'll pick it up. Telling them to pick it up makes their lives no easier.

kamilogorek: Yes, I think we should do that. Just as we do here https://github.com/getsentry/sentry-javascript/blob/master/packages/gatsby/gatsby-node.js#L1-L15

iker-barriocanal:
Closing, since we're leaving it in the config files.

PR #3084, opened 2021-02-12
new strapi listing

rahulchhabria:
@PeloWriter here's a PR for the new strapi integration. it's still WIP. please don't merge yet. i'd appreciate any feedback on what's up there tho. 

thank you.

PR #3614, opened 2021-05-20
Add a Strapi integration page

remidej:
This is a new approach to add Strapi to the list of integrations, after @rahulchhabria's #3084 was closed.

This time, we link to the plugin's documentation instead of copying it.


PeloWriter:
Hi - thanks for adding this info. Looking at this doc for [Linear](https://docs.sentry.io/product/integrations/linear/) as a model raises a couple of questions:
1. Do you know if this needs to be set up once per org, or set up per project?
2. Does this require owner or manager permission to install?
3. Will it work with self-hosted Sentry (we explicit state when an integration won't)?

PeloWriter:
Also, looking at [3084](https://github.com/getsentry/sentry-docs/pull/3084/files), the index page should be updated to add Strapi to the list of APIs.

PeloWriter:
@remidej - thanks for this. Any update?

remidej:
Hello @PeloWriter, and sorry for getting back to you so late.

Thanks for pointing out the index page change on #3084. It made me realize that Strapi felt out of place in that list. Unlike all the other items in the "integrations" section, Strapi is not a CI/CD or deployment solution.

Instead, the strapi-plugin-sentry is more of an SDK, it's a wrapper around your `@sentry/node` plugin. And since Strapi is a Node framework, I thought it made more sense to move it to the platforms section, next to other guides like Koa or Express.

About your questions:

>Do you know if this needs to be set up once per org, or set up per project?
>Does this require owner or manager permission to install?
>Will it work with self-hosted Sentry (we explicitly state when an integration won't)?

I think there is also a misunderstanding here. Our Sentry plugin is not an integration that you install in your Sentry dashboard. It's an NPM plugin that you only have to install in your Strapi application. So to install it, you need to be able to push code changes to your Strapi app, but it's not related to Sentry's permission system. And you have to install it once for every Strapi app.

I assume that it should be compatible with self-hosted Sentry. As far as I know, the plugin only requires a valid DSN, just like `@sentry/node`.

Please let me know if some things about this plugin are unclear, or if you think the docs I added are not in the right place

PeloWriter (commented):
Thanks for this, @remidej, and for clarifying that Strapi is a community-supported SDK. In that case, you've correctly added it to the list on the platforms/index page. 

We don't need a guide page for this, though (and note that the page in this PR would still be entitled "Koa"), because we aren't maintaining it. 

Delete the guides/strapi/index file, and we're good to go with the link for our users to the Strapi content.

remidej:
Would there be a way to show that the plugin is somewhat official?

Even though it's not owned by the Sentry team, it's maintained by our team at Strapi. I'm asking because all the other items in the "community supported" section seem to be made by individual contributors. I feel like the situation is somewhat different here.

Otherwise, I'd be happy to remove the guides/strapi/index file as you suggested.

PeloWriter:
Thanks for asking, we don't have a way of differentiating among those types of community support currently. But if isn't already in the readme in GH, it's worth adding there.

src/docs/product/integrations/strapi/index.mdx
```diff
@@ -0,0 +1,16 @@
+---
```
PeloWriter, line 1: We should also remove this file :)

remidej:
Alright, I removed the Strapi files then

PeloWriter:
Thank you, @remidej

Issue #3605, opened 2021-05-18
Node Docs Updates


PeloWriter:
Node docs need a few updates:

- [ ] Add a list of features (/includes/getting-started-primer)
- [x]  Verify the list of Options - @onurtemizkan
- [ ] Review redundancy between common Filtering content and Node-specific Hints content - @PeloWriter 
- [x] Review integrations content for completeness and currency - @onurtemizkan
- [ ] Update integrations to current structure - @PeloWriter 
- [ ] Review Event Processors for redundancy with common content
- [ ] Review source maps content structure and accuracy (review current JS content to see if it can be used) - @onurtemizkan
- [ ] Review wizard - @PeloWriter 
- [ ] Update performance to use the common content - @PeloWriter

PeloWriter:
This is already triaged, just adding the label for tracking.

PeloWriter:
More feedback for the source maps work, from @framerate on Discord:

Mentioned this a few times other places but I feel like this documentation drastically needs to be improved, as it was critical to my integration: 

https://docs.sentry.io/platforms/node/configuration/integrations/pluggable-integrations/#rewriteframes

Also, doesn't RewriteFrames have 3 options, not two? (but all 3 need to be better fleshed out and have examples)

https://github.com/getsentry/sentry-javascript/blob/master/packages/integrations/src/rewriteframes.ts#L31

PeloWriter:
And more detail from our helpful friend on Discord:

If anyone comes here and reads my errors, here are the two fixes (thanks to Sentry guys on twitter)

1) the webpack:// prefix is only on dev and likely a different problem to solve. Test your electron app with a prod build and you'll likely see a file path with bundle.min.js at the end

2) Its likely that the stripCommonPrefix option in the webpack app (and API as a whole?) is broken. See github open issues for more info. Also possible it works but Sentry's UI just doesn't show you any difference.

3) You have to use Integrations.RewriteFrames() from @sentry/integrations to strip out the file path on every frame (gross).

If the Artifacts on the Sentry UI says ~/bundle.min.js (note the ~/ shows up here regardless if you tell it to stripCommonPrefix. Bug?) you have to strip the path down to the file name but make sure to send it WITH the ~/ from RewriteFrames

I've reached out to support to try and fix documentation and debug whether stripeCommonPrefix is actually broken or not and hopefully I can help the next people with this problem :thumbsup:

framerate:
[...]

PR #3621, opened 2021-05-23
Node: Review Options List

onurtemizkan:
Part of: #3605

- [x] Removed `sendDefaultPii` from options list as it's not supported on JavaScript SDKs 
- [x] Removed `allowUrls` and `denyUrls` from Node options list.
(Might need to also update [inbound filters](https://docs.sentry.io/platforms/node/configuration/integrations/default-integrations/#inboundfilters) on Node, if this gets approved.) 
- [x] Added `frameContextLines` to `node`
- [x] Added `initialScope` to `node` and `javascript`
- [x] Added `maxValueLength` to `node` and `javascript`

### DIFF
--- a/src/platforms/common/configuration/options.mdx
+++ b/src/platforms/common/configuration/options.mdx
@@ -120,7 +120,7 @@ Grouping in Sentry is different for events with stack traces and without. As a r
 
 </ConfigKey>
 
-<ConfigKey name="send-default-pii" supported={["apple", "dotnet", "javascript", "node", "react-native", "python", "java", "php", "rust", "dart"]}>
+<ConfigKey name="send-default-pii" supported={["apple", "dotnet", "react-native", "python", "java", "php", "rust", "dart"]}>
 
 If this flag is enabled, certain personally identifiable information (PII) is added by active integrations. By default, no such data is sent.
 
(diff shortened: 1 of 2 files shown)


src/platforms/common/configuration/options.mdx
```diff
@@ -227,6 +227,24 @@ A path to an alternative CA bundle file in PEM-format.
...
+
+<ConfigKey name="initial-scope" supported={["javascript","node"]}>
+
+Data to be set to the initial scope.
+
+</ConfigKey>
+
+<ConfigKey name="max-value-length" supported={["node"]}>
+
+Maxium number of characters a single value can have before it will be truncated.
```
PeloWriter, line 244: Is this a duplicate of line 212? If so, suggest we simply add node to the list of `supported`? If not, there's a tiny typo :) 
```suggestion
Maximum number of characters a single value can have before it will be truncated.
```

PeloWriter (approved):
One q for `max-value-length`. Other than that, thanks for these fixes.

src/platforms/common/configuration/options.mdx
onurtemizkan: I think it's not the same with 212, at least the default values are different. Made minor updates. :)

PeloWriter: Thanks for following up.

PeloWriter:
@onurtemizkan - ready to merge?

onurtemizkan:
Yes 🚀

PR #3623, opened 2021-05-24
Enable some of enriching events for minidumps

PeloWriter:
We're missing some of the content for Minidumps.


jan-auer (approved):
Thanks, this is good to merge once we remove the transaction name page.

PR #3629, opened 2021-05-26
Update Android Native Bundle group

marandaneto:
Plugin https://github.com/howardpang/androidNativeBundle migrated from JCenter to MavenCentral, but the group and artifact id needed to change, hence updating our docs


philipphofmann (approved):
LGTM

PR #3634, opened 2021-05-26
Document Kotlin coroutines support

### LINKED ISSUE #3628: Document Java coroutines support
## Issue

Document new Sentry Context https://github.com/getsentry/sentry-java/pull/1479

## Description

Docs about adding `SentryContext`

### Suggested Solution

I'm not entirely sure where this should go but I believe somewhere in the Java docs
https://docs.sentry.io/platforms/java/
Android docs would not need it for now as its a single Hub/Scope anyway

Also add a`Note` this is a Beta feature
https://docs.sentry.io/contributing/pages/components/#note

maciejwalkowiak:
Fixes #3628

### DIFF
--- a/src/platforms/common/enriching-events/scopes.mdx
+++ b/src/platforms/common/enriching-events/scopes.mdx
@@ -106,6 +106,12 @@ and every error that happens there will be silently ignored and *not* reported.
 
 ## Kotlin Coroutines
 
+<Note>
+
+Support for Kotlin Coroutines is available as a Beta. Features available in Beta are still in-progress and may have bugs. We recognize the irony.
+
+</Note>
+
 Sentry's SDK for Java stores the scope and the context in a thread-local variable. To make sure that a coroutine has access to the correct Sentry context, an instance of `SentryContext` must be provided when launching a coroutine.
 
 ### Installation
--- a/src/platforms/common/enriching-events/scopes.mdx
+++ b/src/platforms/common/enriching-events/scopes.mdx
@@ -101,3 +101,44 @@ attach any context information at all by calling `clear` on the scope, while the
 
 Keep in mind that `with-scope` will not capture any exceptions that happen inside its callback function,
 and every error that happens there will be silently ignored and *not* reported.
+
+<PlatformSection supported={["java", "android"]}>
+
+## Kotlin Coroutines
+
+Sentry SDK for Java stores the scope and the context in a thread-local variable. To make sure that a coroutine has access to correct Sentry context, an instance of `SentryContext` must be provided when launching a coroutine.
+
+### Installation
+
+```xml {tabTitle:Maven}
+<dependency>
+    <groupId>io.sentry</groupId>
+    <artifactId>sentry-kotlin-extensions</artifactId>
+    <version>{{ packages.version('sentry.java.kotlin-extensions', '5.0.0') }}</version>
+</dependency>
+```
+
+```groovy {tabTitle:Gradle}
+implementation 'io.sentry:sentry-kotlin-extensions:{{ packages.version('sentry.java.kotlin-extensions', '5.0.0') }}'
+```
+
+```scala {tabTitle: SBT}
+libraryDependencies += "io.sentry" % "sentry-kotlin-extensions" % "{{ packages.version('sentry.java.kotlin-extensions', '5.0.0') }}"
+```
+
+### Usage
+
+```kotlin
+import io.sentry.kotlin.SentryContext
+import io.sentry.Sentry
+
+launch(SentryContext()) {
+  // tag set in parent coroutine is visible to child coroutine
+  Sentry.setTag("parent-tag", "value")
+  launch() {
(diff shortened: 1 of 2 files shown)


maciejwalkowiak:
I've added it to Kotlin page, I think it's the place where users should look for coroutines support.

cc @marandaneto @bruno-garcia

src/platforms/kotlin/index.mdx
```diff
@@ -5,4 +5,22 @@ Select your platform to get started:
 - [Java](/platforms/java)
 - [Android](/platforms/android)
 
+## Using Sentry SDK with Coroutines
+
+Sentry SDK for Java stores the scope and the context in a thread-local variable. To make sure that a coroutine has access to correct Sentry context, an instance of `SentryContext` must be provided when launching a coroutine:
```
imatwawana, line 10:
```suggestion
Sentry's SDK for Java stores the scope and the context in a thread-local variable. To make sure that a coroutine has access to the correct Sentry context, an instance of `SentryContext` must be provided when launching a coroutine:
```

imatwawana (approved):
Minor wording edits. Tagging you @PeloWriter to review placement of content.

PeloWriter:
Please don't put this on the index page. It's designed to send users to the Java or Android platforms, which is where the Kotlin samples live.

Instead, I'd recommend putting this content on the [Scopes & Breadcrumbs](https://docs.sentry.io/platforms/java/scope/) page. To enhance searchability, I'd recommend also adding the co-routines keyword to the [frontmatter](https://docs.sentry.io/contributing/pages/frontmatter/) of that file.

bruno-garcia:
Now things got tricky..
@PeloWriter the problem is that this is Kotlin specific. And applies for both Android and JVM uses of Kotlin.

Could we extend the Kotlin redirect page to point users to some Kotlin specific content besides linking them to Android and Java?

PeloWriter:
Yes, we can direct them to headings on specific pages. That would work. But I don't want this page to become a bundle of "oh, and this for Kotlin, too" :) We want users to know we support Kotlin, with the content residing in Java or Android (or, in this case, both)

marandaneto:
> Yes, we can direct them to headings on specific pages. That would work. But I don't want this page to become a bundle of "oh, and this for Kotlin, too" :) We want users to know we support Kotlin, with the content residing in Java or Android (or, in this case, both)

makes sense, but this content is Kotlin specific, it could be used by Android and Java, yes, but it's rarely the case I guess, it's a Kotlin library only.

maybe we move this to https://docs.sentry.io/platforms/java/enriching-events/scopes/ with a new section, and link this new section header on the main Kotlin page, how does that sound?

PeloWriter:
Yes, this is a great approach @marandaneto 

> maybe we move this to https://docs.sentry.io/platforms/java/enriching-events/scopes/ with a new section, and link this new section header on the main Kotlin page, how does that sound?

src/platforms/common/enriching-events/scopes.mdx
```diff
@@ -101,3 +101,44 @@ attach any context information at all by calling `clear` on the scope, while the
 
 Keep in mind that `with-scope` will not capture any exceptions that happen inside its callback function,
 and every error that happens there will be silently ignored and *not* reported.
+
+<PlatformSection supported={["java", "android"]}>
```
marandaneto, line 105:
```suggestion
<PlatformSection supported={["java"]}>
```

Android is single hub already, actually maybe android inherits from `java` already, maybe we need to add non support for it

src/platforms/common/enriching-events/scopes.mdx
```diff
@@ -101,3 +101,44 @@ attach any context information at all by calling `clear` on the scope, while the
...
+
+## Kotlin Coroutines
```
marandaneto, line 107: let's add a note that this is a beta version and may contain bugs, so people are aware of it, like Flutter

https://github.com/getsentry/sentry-docs/blame/master/src/includes/getting-started-primer/flutter.mdx#L3

maciejwalkowiak: 👍

src/platforms/common/enriching-events/scopes.mdx
maciejwalkowiak: 👍

src/platforms/kotlin/index.mdx
```diff
@@ -5,4 +5,14 @@ Select your platform to get started:
 - [Java](/platforms/java)
 - [Android](/platforms/android)
 
+## Kotlin Extensions
+
+<Note>
+
+Support for Kotlin Coroutines is available as a Beta. Features available in Beta are still in-progress and may have bugs. We recognize the irony.
+
+</Note>
```
marandaneto, line 14: this should be under the `## Kotlin Coroutines` section in the `src/platforms/common/enriching-events/scopes.mdx` file probably otherwise anyone accessing from there would not see it

PR #3636, opened 2021-05-26
Fix incorrect sampling configuration method.

imajes:
This cost us a few hours of stress as we thought it was a legit thing to be configured -- however, this mistake caused us to turn on performance sampling, quickly exceed the quota and cause all kinds of alarms to go off. :/


PeloWriter:
@st0012 - could you please take a look?

st0012 (approved):
@imajes I'm sorry for the trouble 😞  
thanks for your fix 👍

imajes:
No worries. All fixed. Just make sure no one else makes the same mistake :)

PR #3644, opened 2021-05-27
Feat/docs notes theme update

stevenplewis:


src/components/note.tsx
```diff
@@ -5,10 +5,7 @@ type Props = {
   children?: any;
 };
 
-const Note = styled.div`
-  color: var(--light-text);
-  line-height: 1.65;
-`;
+const Note = styled.div``;
```
cameronmcefee, line 8: Is this dom element even necessary at this point?

PeloWriter:
The notes look great, but some of the text is also now changed.

Our homepage looks a bit peculiar (some of the text that should remain black is now purple):
Screen Shot 2021-06-01 at 2 55 30 PM
I'm less concerned about these two changes(I think I just need to retrain my eye):
Screen Shot 2021-06-01 at 2 55 59 PM
Screen Shot 2021-06-01 at 2 57 39 PM

stevenplewis:
> The notes look great, but some of the text is also now changed.
> 
> Our homepage looks a bit peculiar (some of the text that should remain black is now purple):
> Screen Shot 2021-06-01 at 2 55 30 PM
> I'm less concerned about these two changes(I think I just need to retrain my eye):
> Screen Shot 2021-06-01 at 2 55 59 PM
> Screen Shot 2021-06-01 at 2 57 39 PM

Colors have been updated.

imatwawana:
Maybe this is just me, but the link colour looks like the colour of clicked links to me. I assume the colour is meant to look like Sentry Purple. Wondering if we could try Sentry Violet or Light Violet (I think Sentry Dark Purple would lean too close to black).

PR #3653, opened 2021-05-28
Add Next.js SDK's performance monitoring docs

iker-barriocanal:
Docs for the new performance monitoring of the Next.js SDK.

### DIFF
--- a/src/includes/getting-started-primer/javascript.nextjs.mdx
+++ b/src/includes/getting-started-primer/javascript.nextjs.mdx
@@ -12,7 +12,7 @@ Features:
 - Events [enriched](/platforms/javascript/enriching-events/context/) with device data
 - [Breadcrumbs](/platforms/javascript/enriching-events/breadcrumbs/) created for outgoing HTTP request with XHR and Fetch, and console logs
 - [Release Health](/product/releases/health/) for tracking crash-free users and sessions
-- Automatic [Performance Monitoring](/product/performance/) for both the client and server
+- Automatic [Performance Monitoring](/product/performance/) for both the client and server, from version `6.5.0`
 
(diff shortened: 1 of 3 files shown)


iker-barriocanal:
@rahulchhabria

AbhiPrasad (commented):
What do you think about switching from `frontend` -> `client-side` and `backend` -> `server-side`? I think that better matches to how nextjs refers to those concepts, but I'm not sure it is as clear as just saying frontend, backend.

iker-barriocanal:
> What do you think about switching from `frontend` -> `client-side` and `backend` -> `server-side`? I think that better matches to how nextjs refers to those concepts, but I'm not sure it is as clear as just saying frontend, backend.

Yes @AbhiPrasad, in fact, that's the language we've been using in the SDK. Updated.

src/includes/getting-started-primer/javascript.nextjs.mdx
```diff
@@ -12,6 +12,7 @@ Features:
 - Events [enriched](/platforms/javascript/enriching-events/context/) with device data
 - [Breadcrumbs](/platforms/javascript/enriching-events/breadcrumbs/) created for outgoing HTTP request with XHR and Fetch, and console logs
 - [Release Health](/product/releases/health/) for tracking crash-free users and sessions
+- Automatic [Performance Monitoring](/product/performance/) for both the client and server
```
AbhiPrasad, line 15: Can we put something like `**new in 6.5.0**`?

AbhiPrasad (changes requested):
We need some indication that performance monitoring is only available for the SDK v6.5.0, otherwise people will def get confused. I left a comment on for one place where we can add it, but there might be more places.

HazAT (changes requested):
We should also add it to the wizard `tracesSampleRate`

HazAT (approved):
Sorry, it's already there 😓

PR #3655, opened 2021-05-28
Fix: Unity specific docs for 'release' and 'environment'

bitsandfoxes:
Docs for how Unity handles release and environment by default.


src/platforms/common/configuration/options.mdx
```diff
@@ -89,7 +89,9 @@ By default the SDK will try to read this value from the `SENTRY_ENVIRONMENT` env
 
 <PlatformSection supported={["unity"]}>
 
-Sets the environment. This string is freeform and set by default. A release can be associated with more than one environment to separate them in the UI (think `staging` vs `prod` or similar). Depending on `Application.isEditor` the environment will be set to `editor` or `production` respectively.
+Sets the environment. This string is freeform and set by default. A release can be associated with more than one environment to separate them in the UI (think `staging` vs `prod` or similar).
```
bitsandfoxes, line 92: I'm not sure if I'm doing this correctly here. Just one big copy-pasta. It's just the difference between "set by default" and "NOT set by default". Is there any better way to handle this?

src/includes/set-environment/unity.mdx
```diff
@@ -0,0 +1,15 @@
...
+```
+
+```fsharp
+open Sentry
+
+// Add this to the SDK initialization callback
+options.Environment <- "production"
+```
+
+The SDK will set the environment by default depending on `Application.isEditor` to `editor` or `production` respectively.
```
bruno-garcia, line 15: I think it's best if we don't add this file (meaning copy paste the whole thing from .NET) and simply add this extra line conditionally to the common content (as you did for the release and env pages).

If you still prefer adding this here, might as well remove F# then. And even write `SentryUnity(o => ` since we are specifically doing the unity part here.

src/includes/set-release/unity.mdx
```diff
@@ -0,0 +1,18 @@
...
+```
+
+```fsharp {tabTitle:F#}
+open Sentry
+
+// Add this to the SDK initialization callback
+options.Release <- "my-project-name@2.3.12"
+```
+
+The SDK will set the release by default from `Application.productName` and `Application.version`.
```
bruno-garcia, line 15: Same as above, we could simply add to the common content, right above the `include="set-release"`:

```
<PlatformSection supported={["unity"]}>

The SDK will set the release by default from `Application.productName` and `Application.version`.

</PlatformSection>
````

This way we tell the user: The C# snippet you'll see likely isn't relevant because the SDK takes care of this for you But if you want to override it anyway:

File: src/platforms/common/configuration/options.mdx [hunk 4]
```diff
@@ -63,16 +63,38 @@ Enabling `debug` mode makes the SDK generate as much diagnostic data as possible
...
 
+<PlatformSection notSupported={["unity"]}>
+
 By default the SDK will try to read this value from the `SENTRY_RELEASE` environment variable (in the browser SDK, this will be read off of the `window.SENTRY_RELEASE` if available).
 
+</PlatformSection>
+
+<PlatformSection supported={["unity"]}>
+
+By default the SDK will read from `Application.productName` and `Application.version` to create the release.
```
bruno-garcia, line 74:
```suggestion
By default the SDK will read from `Application.productName` and `Application.version` to create the release in the format `$"{Application.productName}@{Application.version}".`
```

src/platforms/common/configuration/options.mdx
bruno-garcia: I think we could just add this at the end, and keep the other content available to all platforms (not exclude for Unity).

src/platforms/common/configuration/options.mdx
```diff
@@ -63,16 +63,38 @@ Enabling `debug` mode makes the SDK generate as much diagnostic data as possible
...
 
 By default the SDK will try to read this value from the `SENTRY_ENVIRONMENT` environment variable (except for the browser SDK where this is not applicable).
 
+</PlatformSection>
+
+<PlatformSection supported={["unity"]}>
+
+Sets the environment. This string is freeform and set by default. A release can be associated with more than one environment to separate them in the UI (think `staging` vs `prod` or similar).
+
+By default the SDK will set the environment to `editor` or `production`, depending on `Application.isEditor`.
```
bruno-garcia, line 94: What about something like this:

```suggestion
By default, the SDK reports `editor` when running inside the Unity. Otherwise, the default environment is `production`.
```

bruno-garcia (approved):

src/includes/set-release/unity.mdx
bitsandfoxes: I'm adding its own file so we override the Dotnet documentation that fetches the version from the assembly.

bruno-garcia (approved):

PR #3659, opened 2021-05-31
Feat: Unity offline caching

bitsandfoxes:
Added offline caching to primer.
Added configuration/offline caching

Because I was already touching the primer I emulated the way other mobile SDKs display their breadcrumb feature + link.


src/platforms/unity/configuration/offline-caching.mdx
```diff
@@ -0,0 +1,14 @@
+---
+title: Offline Caching
```
marandaneto, line 2: should this be its own page? usually we only make it available in the basic options
https://docs.sentry.io/platforms/android/configuration/options/#cache-directory-path

and also in the shutdown & draining page https://docs.sentry.io/platforms/android/configuration/draining/

src/platforms/unity/configuration/offline-caching.mdx
```diff
@@ -0,0 +1,14 @@
...
+description: "Learn how the SDK handles offline caching."
+sidebar_order: 200
+---
+
+Offline caching is enabled by default. When a device is unable to send an event it is cached and we send a report once we receive another event. The SKD uses `Application.persistentDataPath` as base directory for caching. If you change this path take note that it should persist between sessions to capture events that would be lost otherwise.
```
philipphofmann, line 7: `l`: some wording suggestions.
```suggestion
Offline caching is enabled by default. When a device cannot send an event, it is cached, and the SDK sends a report once we receive another event. The SKD uses `Application.persistentDataPath` as the base directory for caching. If you change this path, take note that it should persist between sessions to capture events that would be lost otherwise.
```

philipphofmann (commented):
I just have some wording suggestions and what @marandaneto said.

src/platforms/unity/configuration/offline-caching.mdx
bitsandfoxes: Thanks for pointing that out. I completely missed it.

src/platforms/common/configuration/options.mdx
```diff
@@ -312,14 +312,22 @@ Configures a separate proxy for outgoing HTTPS requests. This value might not be
 
 </ConfigKey>
 
-<ConfigKey name="cache-directory-path" supported={["dotnet", "android"]}>
+<ConfigKey name="cache-directory-path" supported={["dotnet", "android"]} notSupported={["unity"]}>
```
bitsandfoxes, line 315: This seems very wrong. But all of `Transport Options` is inside a PlatformSection. So I can't just switch for individual lines. So how else would I go about it?

src/includes/getting-started-primer/unity.mdx
```diff
@@ -2,6 +2,8 @@ Our Unity SDK builds on top of the [.NET SDK](/platforms/dotnet/) and extends it
 
 **Additional Features:**
 
-- Unity's `Debug.Log` and `Debug.LogWarning` get captured with [breadcrumbs](/product/issues/issue-details/breadcrumbs/)
-- Scene loading and unloading get captured with [breadcrumbs](/product/issues/issue-details/breadcrumbs/)
+- [Offline caching](/platforms/unity/configuration/options/#cache-directory-path) when a device is unable to connect; we send a report once we receive another event
```
bruno-garcia, line 5:
```suggestion
- [Offline caching](/platforms/unity/configuration/options/#cache-directory-path) stores data to disk in case the device is not online
```

bruno-garcia (approved):

src/platforms/common/configuration/options.mdx
PeloWriter: I agree, duplication means the content becomes wrong in at least one place. Rather than duplicate, I'd suggest updating the `Default` content (I'll edit that below)

src/platforms/common/configuration/options.mdx
```diff
@@ -312,14 +312,22 @@ Configures a separate proxy for outgoing HTTPS requests. This value might not be
...
 
 Default: not set (caching is disabled).
 
 </ConfigKey>
 
+<ConfigKey name="cache-directory-path" supported={["unity"]}>
+
+Specifies a local directory used for cacheing payloads. When this option is enabled (that is, when the directory is set), the Sentry SDK will persist envelopes locally before sending to Sentry. This configuration option is particularly useful if you expect your application to run in environments where internet connectivity is limited.
+
+Default: Set to `Application.persistentDataPath` (caching is enabled).
```
PeloWriter, line 327:
```suggestion
Default: Varies per SDK, for .NET and Android, the default is not set (caching is disabled). For Unity, the default is set to `Application.persistentDataPath` (caching is enabled).
```

src/platforms/common/configuration/options.mdx
PeloWriter: If that doesn't make sense, lmk, and I'll play with it a bit more.

PeloWriter (changes requested):

src/platforms/common/configuration/options.mdx
bruno-garcia: All Mobile SDKs have offline caching enabled by default.
Java and .NET have it disabled by default.

src/platforms/common/configuration/options.mdx
```diff
@@ -318,6 +318,8 @@ Specifies a local directory used for cacheing payloads. When this option is enab
 
 Default: not set (caching is disabled).
 
+Unity builds on top of the .NET framework but enables caching by default to be in line with other Mobile SDKs. Default set to `Application.persistentDataPath`, set to `null` to disable offline caching.
```
bruno-garcia, line 321:
```suggestion
Client SDKs built on top of the Sentry SDK for .NET such Xamarin and Unity have this feature enabled by default. For example, in the case of Unity, the default value is `Application.persistentDataPath`. You can disable offline caching by setting it to `null`.
```

bruno-garcia (approved):

PeloWriter (approved):
Thank you.

PR #3679, opened 2021-06-02
connect span to error aspnetcore

bruno-garcia:
@Tyrrrz 

Should we change this everywhere?


bruno-garcia:
Checked and it compiles

PR #3688, opened 2021-06-04
feat(nextjs): Add nextjs performance docs

lobsterkatie:
This follows up on https://github.com/getsentry/sentry-docs/pull/3653, which added mentions of performance to the nextjs docs, but didn't actually add the performance docs themselves. This does so, and includes the following changes:

- add nextjs-specific code samples where required
- turn spots where nextjs needs to override the main JS content into includes
- simplify example code for `BrowserTracing` options
- remove note about skipping setup page, as it contains useful information about sampling and testing
- do general wordsmithing

### DIFF
--- a/src/includes/performance/beforeNavigate-example/javascript.mdx
+++ b/src/includes/performance/beforeNavigate-example/javascript.mdx
@@ -1,9 +1,8 @@
-```javascript
-import * as Sentry from "@sentry/browser";
-import { Integrations } from "@sentry/tracing";
+One common use case is parameterizing transaction names. For both `pageload` and `navigation` transactions, the `BrowserTracing` integration uses the browser's `window.location` value to generate a transaction name. Using `beforeNavigate` you can modify the transaction name to make it more generic, so that, for example, transactions named `GET /users/12312012` and `GET /users/11212012` can both be renamed `GET /users/:userid`, so that they'll group together.
 
+```javascript
 Sentry.init({
-  dsn: "___PUBLIC_DSN___",
+  // ...
   integrations: [
     new Integrations.BrowserTracing({
       beforeNavigate: context => {
@@ -19,9 +18,5 @@ Sentry.init({
       },
     }),
   ],
-
-  // We recommend adjusting this value in production, or using tracesSampler
-  // for finer control
-  tracesSampleRate: 1.0,
 });
 ```
--- a/src/includes/performance/configure-sample-rate/javascript.nextjs.mdx
+++ b/src/includes/performance/configure-sample-rate/javascript.nextjs.mdx
@@ -3,12 +3,11 @@ In both `sentry.server.config.js` and `sentry.client.config.js`:
 ```javascript
 import * as Sentry from "@sentry/nextjs";
 
-
 Sentry.init({
   dsn: "___PUBLIC_DSN___",
 
   // To set a uniform sample rate
-  tracesSampleRate: 0.2
+  tracesSampleRate: 0.2,
 
   // Alternatively, to control sampling dynamically
   tracesSampler: samplingContext => { ... }
--- a/src/includes/performance/configure-sample-rate/javascript.nextjs.mdx
+++ b/src/includes/performance/configure-sample-rate/javascript.nextjs.mdx
@@ -1,0 +1,16 @@
+In both `sentry.server.config.js` and `sentry.client.config.js`:
+
+```javascript
+import * as Sentry from "@sentry/nextjs";
+
+
+Sentry.init({
+  dsn: "___PUBLIC_DSN___",
+
+  // To set a uniform sample rate
+  tracesSampleRate: 0.2
+
+  // Alternatively, to control sampling dynamically
+  tracesSampler: samplingContext => { ... }
+});
+```
--- a/src/includes/performance/group-transaction-example/javascript.mdx
+++ b/src/includes/performance/group-transaction-example/javascript.mdx
@@ -1,12 +1,11 @@
 An example of doing this in a node.js application:
 
 ```javascript
-// all JavaScript-based SDKs include this function, so it's safe to replace `@sentry/node`
+// All JavaScript-based SDKs include this function, so it's safe to replace `@sentry/node`
 // with your particular SDK
 import { addGlobalEventProcessor } from "@sentry/node";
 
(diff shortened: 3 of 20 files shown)


src/includes/performance/beforeNavigate-example/javascript.mdx
```diff
@@ -1,9 +1,8 @@
-```javascript
-import * as Sentry from "@sentry/browser";
-import { Integrations } from "@sentry/tracing";
+One common use case is parameterizing transaction names. For both `pageload` and `navigation` transactions, the `BrowserTracing` integration uses the browser's `window.location` value to generate a transaction name. Using `beforeNavigate` you can modify the transaction name to make it more generic, so that, for example, transactions named `GET /users/12312012` and `GET /users/11212012` can both be renamed `GET /users/:userid`, so that they'll group together.
```
iker-barriocanal, line 1: This paragraph should also be included for platform-specific content (e.g. also Next.js).

src/includes/performance/configure-sample-rate/javascript.nextjs.mdx
```diff
@@ -0,0 +1,16 @@
...
+
+```javascript
+import * as Sentry from "@sentry/nextjs";
+
+
+Sentry.init({
+  dsn: "___PUBLIC_DSN___",
+
+  // To set a uniform sample rate
+  tracesSampleRate: 0.2
```
iker-barriocanal, line 11:
```suggestion
  tracesSampleRate: 0.2,
```

src/includes/performance/configure-sample-rate/javascript.nextjs.mdx
```diff
@@ -0,0 +1,16 @@
+In both `sentry.server.config.js` and `sentry.client.config.js`:
+
+```javascript
+import * as Sentry from "@sentry/nextjs";
+
+
```
iker-barriocanal, line 6:
```suggestion

```

src/includes/performance/filter-span-example/javascript.mdx
```diff
@@ -1,8 +1,6 @@
 ```javascript
-import * as Sentry from "@sentry/browser";
-import { Integrations } from "@sentry/tracing";
```
iker-barriocanal, line 3: This is personal preference, but I'd leave the imports.

src/includes/performance/group-transaction-example/javascript.mdx
```diff
@@ -1,6 +1,8 @@
 An example of doing this in a node.js application:
 
 ```javascript
+// all JavaScript-based SDKs include this function, so it's safe to replace `@sentry/node`
```
iker-barriocanal, line 4:
```suggestion
// All JavaScript-based SDKs include this function, so it's safe to replace `@sentry/node`
```

src/includes/performance/tracingOrigins-example/javascript.nextjs.mdx
```diff
@@ -0,0 +1,18 @@
+For example:
+
+- A frontend application is served from `example.com`
+- A backend service is served from `api.example.com`
+- The frontend application makes API calls to the backend
+- Set the `tracingOrigins` option to `['api.example.com']`
+- Now outgoing XHR/fetch requests to `api.example.com` will get the `sentry-trace` header attached
```
iker-barriocanal, line 7:
```suggestion
- A frontend application is served from `example.com`.
- A backend service is served from `api.example.com`.
- The frontend application makes API calls to the backend.
- Set the `tracingOrigins` option to `['api.example.com']`.
- Now outgoing XHR/fetch requests to `api.example.com` will get the `sentry-trace` header attached.
```
?

src/platforms/common/performance/connect-services.mdx
```diff
@@ -12,14 +12,15 @@ To connect backend and frontend transactions into a single coherent trace, Sentr
 
 ### Navigation and Other XHR Requests
 
-For traces that begin in JavaScript, any requests it makes (and any requests your backend makes as a result) are linked through a request header.
+For traces that begin in the front end, any requests made (and any requests your backend makes as a result) are linked through a request header.
 
-All of Sentry's tracing-related integrations (`BrowserTracing`, `Http`, and `Express`) either generate or pick up and propagate the trace header automatically as appropriate for all transactions and spans that they generate.
+All of Sentry's tracing-related integrations (`BrowserTracing`, `Http`, and `Express`), as well as the Next.JS SDK either generate or pick up and propagate the trace header automatically as appropriate, for all transactions and spans that they generate.
```
iker-barriocanal, line 17:
```suggestion
All of Sentry's tracing-related integrations (`BrowserTracing`, `Http`, and `Express`), as well as the Next.JS SDK, either generate or pick up and propagate the trace header automatically as appropriate, for all transactions and spans that they generate.
```

iker-barriocanal (changes requested):

src/includes/performance/beforeNavigate-example/javascript.mdx
lobsterkatie: I specifically didn't include it for nextjs because we automatically parameterize transaction names, for any transaction `beforeNavigate` would handle, since manual transactions don't pass through it, and we handle both pageload and navigation transaction-name parameterization. 

For all other frameworks, them not having a file of their own means they'll use this one.

src/includes/performance/filter-span-example/javascript.mdx
lobsterkatie: I felt like, since they're (now it's, as of https://github.com/getsentry/sentry-javascript/pull/3647) in the snippet at the top of the section, it was enough. 

image

I also took out the DSN there, because the point of these lower snippets is really just to highlight the use of the option in question not to be a fully-formed implementation. We do the same thing [here](https://docs.sentry.io/platforms/javascript/guides/nextjs/configuration/releases/), for example.

src/includes/performance/tracingOrigins-example/javascript.nextjs.mdx
lobsterkatie: Yeah, fair. I don't love sentences as bullet points in the first place, but you're right, they are full sentences for the moment.

iker-barriocanal (approved):

PR #3691, opened 2021-06-04
clarify auto ip address

bruno-garcia:
"the connection" for users can read as: The connection between my users and my server. But we mean here exclusively: The connection between Sentry's SDK and Sentry's server.

### DIFF
--- a/src/platforms/common/enriching-events/identify-user.mdx
+++ b/src/platforms/common/enriching-events/identify-user.mdx
@@ -26,7 +26,8 @@ Users consist of a few critical pieces of information that construct a unique id
 
 `ip_address`
 
-: The user's IP address. If the user is unauthenticated, Sentry uses the IP address as a unique identifier for the user. Sentry will attempt to pull this from the HTTP request data, if available. Set to `"{{auto}}"` to let Sentry infer the IP address from the connection.
+: The user's IP address. If the user is unauthenticated, Sentry uses the IP address as a unique identifier for the user. Sentry will attempt to pull this from the HTTP request data, if available. That might required `sendDefaultPii` set to `true` in the SDK options.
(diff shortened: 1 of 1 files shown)


src/platforms/common/enriching-events/identify-user.mdx
```diff
@@ -26,7 +26,8 @@ Users consist of a few critical pieces of information that construct a unique id
 
 `ip_address`
 
-: The user's IP address. If the user is unauthenticated, Sentry uses the IP address as a unique identifier for the user. Sentry will attempt to pull this from the HTTP request data, if available. Set to `"{{auto}}"` to let Sentry infer the IP address from the connection.
+: The user's IP address. If the user is unauthenticated, Sentry uses the IP address as a unique identifier for the user. Sentry will attempt to pull this from the HTTP request data, if available. That might required `sendDefaultPii` set to `true` in the SDK options.
+You can set to `"{{auto}}"` to let Sentry infer the IP address from the connection between your app and Sentry's server. This is useful for client application such as a website, desktop and mobile.
```
PeloWriter, line 30:
```suggestion
: The user's IP address. If the user is unauthenticated, Sentry uses the IP address as a unique identifier for the user. Sentry will attempt to pull this from the HTTP request data, if available. That might required `sendDefaultPii` set to `true` in the SDK options. If set to `"{{auto}}"`, Sentry will infer the IP address from the connection between your app and Sentry's server, which is useful for client applications such as a website, desktop, or mobile.
```

PeloWriter (approved):
This is true for all platforms, correct? Because this update affects the common content that displays across platforms.

PeloWriter:
Thanks, @bruno-garcia

PR #1080, opened 2019-06-26
Add missing `error_types` parameter

rcknr:
Add missing `error_types` configuration parameter to PHP SDK docs.

PR #3702, opened 2021-06-07
Remove SpoolTransport mention in php transports doc

JeckLabs:
Hello, as I can see, SpoolTranspot was removed, see [#1080](https://github.com/getsentry/sentry-php/pull/1080). So the reference here may be misguiding.


PeloWriter:
Thanks, @JeckLabs.

@stayallive - can you please confirm that this change is technically correct? Thank you.

stayallive:
@PeloWriter, yes, this should indeed be removed! 👌

Thanks!

PeloWriter:
Terrific. Thank you both!

PR #3707, opened 2021-06-08
Update identify-user.mdx

Tiagojdferreira:
Removed the email from the examples, since due to GDPR this information is deemed sensitive. Did so based on https://docs.sentry.io/platforms/javascript/data-management/sensitive-data/

Is username also considered sensitive, since it was generated by the user and can be seen as:
> Confidential IP (Intellectual Property), such as your favorite color, or your upcoming plans for world domination.
?

I think uid and ip should be fine.


untitaker:
We still accept this data and if you get your ToS right it *may* be ok for you to send it. This is technical documentation, not legal advice. We can't tell you what you are legally allowed to send.

PR #3709, opened 2021-06-08
open-source → self-hosted in SaaS migration guide

chadwhitacre:
I gather that sentry-basics is coming up for an overhaul. Short of that, here's a small revision to the SaaS migration guide. The primary motivation here is to stop referring to self-hosted Sentry as "open-source" Sentry ... _Sentry_ is open-source Sentry! 💃


PeloWriter (approved):
Thanks for these changes, @chadwhitacre

BYK (approved):

Issue #3697, opened 2021-06-07
Celery guide: Clarify init on worker startup 


danmichaelo:
## Issue

This an issue with: [Celery guide](https://docs.sentry.io/platforms/python/guides/celery/)

## Description

The guide emphasizes that you should make sure *"the call to init is loaded on worker startup, and not only in the module where your tasks are defined"*, but doesn't show an example of how to do that in practice.

### Suggested Solution

I'm a Celery newbie, but from reading the Celery docs, my guess is that something like this might be a good approach?

```python
import sentry_sdk
from celery.signals import celeryd_init

@celeryd_init.connect()
def configure_sentry(**kwargs):
    sentry_sdk.init(...)
```

since [`celeryd_init`](https://docs.celeryproject.org/en/stable/userguide/signals.html#celeryd-init) is *"the first signal sent when celery worker starts up"*.

If I understand it correctly, it's not necessary to initialize Sentry per worker process ([`worker_process_init`](https://docs.celeryproject.org/en/stable/userguide/signals.html#worker-process-init)), just per worker?

untitaker:
There are multiple ways. Connecting a signal will work. It's also sufficient to initialize the SDK in the configuration file that celery reads with `--config`, or in the module that celery imports when importing your app.

getsentry-release:
Routing to @getsentry/team-webplatform for [triage](https://develop.sentry.dev/processing-tickets/#3-triage). ⏲️

getsentry-release:
Routing to @getsentry/team-webplatform for [triage](https://develop.sentry.dev/processing-tickets/#3-triage). ⏲️

PeloWriter:
@untitaker - should we update the Docs to capture this information? Perhaps add to https://docs.sentry.io/platforms/python/guides/celery/ in the paragraph:

"Generally, make sure that the call to init is loaded on worker startup, and not only in the module where your tasks are defined. Otherwise, the initialization happens too late and events might end up not being reported."

Adding: "You don't need to initialize Sentry per worker process. Connecting a signal will work. It's also sufficient to initialize the SDK in the configuration file that celery reads with `--config`, or in the module that celery imports when importing your app."

untitaker:
I opened a PR

PR #3711, opened 2021-06-09
fix: Clarify when Celery integration should be initialized

untitaker:
Fix https://github.com/getsentry/sentry-docs/issues/3697

### DIFF
--- a/src/platforms/python/guides/celery/index.mdx
+++ b/src/platforms/python/guides/celery/index.mdx
@@ -26,6 +26,10 @@ The integration will automatically report errors from all celery jobs.
 
 Generally, make sure that the **call to `init` is loaded on worker startup**, and not only in the module where your tasks are defined. Otherwise, the initialization happens too late and events might end up not being reported.
 
+There are many valid ways to do that, for example by using a signal like [`worker_process_init`](https://docs.celeryproject.org/en/stable/userguide/signals.html#worker-process-init), or by initializing the SDK in the configuration file loaded with Celery's `--config` parameter.
+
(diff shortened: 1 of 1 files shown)


src/platforms/python/guides/celery/index.mdx
```diff
@@ -26,6 +26,10 @@ The integration will automatically report errors from all celery jobs.
 
 Generally, make sure that the **call to `init` is loaded on worker startup**, and not only in the module where your tasks are defined. Otherwise, the initialization happens too late and events might end up not being reported.
 
+There are many valid ways to do that, for example by using a signal like [`worker_process_init`](https://docs.celeryproject.org/en/stable/userguide/signals.html#worker-process-init), or by initializing the SDK in the configuration file loaded with Celery's `--config` parameter.
+
+To verify if your SDK is initialized on worker start, you can pass `debug=True` to see extra output when the SDK is initialized. If the output appears during worker startup and not only after a task has started, you should be fine.
+
```
imatwawana, line 32:
```suggestion
There are many valid ways to do this like, for example, using a signal like [`worker_process_init`](https://docs.celeryproject.org/en/stable/userguide/signals.html#worker-process-init), or by initializing the SDK in the configuration file loaded with Celery's `--config` parameter.

To verify if your SDK is initialized on worker start, you can pass `debug=True` to see extra output when the SDK is initialized. If the output appears during worker startup and not only after a task has started, then it's working properly.

```

imatwawana (approved):
Minor smoothing

PR #3713, opened 2021-06-09
Update options.mdx

georgekallilas:
"legacy" is a leftover word from then it was called whitelistUrl instead of allowUrl which is now called. So the legacy word can be omitted.


PeloWriter (approved):
Thank you for fixing this error, @georgekallilas

PR #3731, opened 2021-06-14
Document "ignored-exceptions-for-type" configuration option.

### LINKED ISSUE #3675: Document IgnoredExceptionsForType for Java/Android
## Issue

Related to https://github.com/getsentry/sentry-java/pull/1352/files
Java/Android supports IgnoredExceptionsForType but its not documented under https://docs.sentry.io/platforms/java/configuration/options/ and Android docs

## Description

option to be documented

### Suggested Solution

Add to default options

maciejwalkowiak:
Fixes #3675

### DIFF
--- a/src/platforms/common/configuration/options.mdx
+++ b/src/platforms/common/configuration/options.mdx
@@ -209,6 +209,12 @@ This option can be overridden using <PlatformIdentifier name="in-app-include" />
 
 </ConfigKey>
 
+<ConfigKey name="ignored-exceptions-for-type" supported={["java", "android"]}>
+
+A list of exception types that will be filtered out before sending to Sentry.
+
+</ConfigKey>
+
 <ConfigKey name="request-bodies" supported={["python"]}>
 
 This parameter controls if integrations should capture HTTP request bodies. It can be set to one of the following values:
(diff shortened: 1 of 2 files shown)


src/platforms/common/configuration/options.mdx
```diff
@@ -209,6 +209,12 @@ This option can be overridden using <PlatformIdentifier name="in-app-include" />
 
 </ConfigKey>
 
+<ConfigKey name="ignored-exceptions-for-type" supported={["java", "android"]}>
```
marandaneto, line 212: iirc C# already has this feature, isn't documented for C# yet, or is it another name?

marandaneto (approved):
LGTM

src/platforms/common/configuration/options.mdx
PeloWriter: I don't see `ExceptionFilters` in the current content, can we add that as well?

bruno-garcia: @PeloWriter recently refactored the .NET Docs, this is now here: https://docs.sentry.io/platforms/dotnet/usage/#ignoring-errors

bruno-garcia: Though in C# it's only through the API not via String names of types

bruno-garcia (approved):
LGTM

src/platforms/common/configuration/options.mdx
marandaneto: we should have aligned the naming, its called `addIgnoredExceptionForType` on Java, should we rename it in the next major?

PR #3736, opened 2021-06-15
fix Xamarin getting started

lucas-zimerman:
A simple fix for the getting started config example that was using the wrong variable name for SentryOptions.


bruno-garcia (approved):
Thanks!

PR #3747, opened 2021-06-16
docs(js): Enhance tunnel documentation

kamilogorek:
Added more details to explain what/why/how.


src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -66,9 +66,19 @@ If `instrument.js` displays in your console while debugging, add Sentry to your
 
 ## Dealing with Ad-Blockers
 
+When you are using our CDN, ad-blocking or script-blocking extensions may prevent our SDK from being fetched and initialized properly. Because of this, any call to SDKs API will fail and may cause your application to misbehave.
```
imatwawana, line 69:
```suggestion
When you are using our CDN, ad-blocking or script-blocking extensions may prevent our SDK from being fetched and initialized properly. Because of this, any call to the SDKs API will fail and may cause your application to behave unexpectedly.
```

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -66,9 +66,19 @@ If `instrument.js` displays in your console while debugging, add Sentry to your
...
+
+Additionally, even when the SDK is downloaded and initialized correctly, Sentry endpoints that need to receive captured data may be blocked as well. This will prevent any error reports, sessions health or performance data to be delivered, and effectively not available in your Sentry dashboard.
```
imatwawana, line 71: I think you're trying to say this data won't be available to you in the Sentry app UI. If so then, how I've updated it below is the way we reference that.

```suggestion
Additionally, even when the SDK is downloaded and initialized correctly, Sentry endpoints that need to receive captured data may be blocked as well. This prevents any error reports, sessions health, or performance data from being delivered, making it effectively unavailable in [sentry.io](https://sentry.io).
```

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -66,9 +66,19 @@ If `instrument.js` displays in your console while debugging, add Sentry to your
...
+
+The first issue can be worked around in multiple ways explained above. However, the endpoint blockage can be only resolved using the tunnel.
```
imatwawana, line 73:
```suggestion
You can work around the first issue in multiple ways explained above. However, the endpoint blockage can be only resolved using the tunnel.
```

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -66,9 +66,19 @@ If `instrument.js` displays in your console while debugging, add Sentry to your
...
+
 ### Using the `tunnel` Option
 
-Starting with version `6.7.0` of the JavaScript SDK, you can use the `tunnel` option to tell the SDK to deliver events to the configured URL, instead of using the DSN. This allows the SDK to remove `sentry_key` from the query parameters, which is one of the main reasons ad-blockers prevent sending events. This option also stops the SDK from sending preflight requests, which was one of the requirements that necessitated sending the `sentry_key` in the query parameters.
+A tunnel is an HTTP endpoint that will act as a proxy between Sentry and your application. Due to the fact that you control this server, there is no risk of any requests sent to that server being blocked. When the endpoint lives under the same origin (although it does not have to in order for the tunnel to work), the browser will not treat any requests to it as a third party requests, and they will have different security measures applied, which ad-blockers does not capture by default. A quick summary of the flow can be found below.
```
imatwawana, line 77: The second-to last sentence was quite long so I tried to break it up, but I think it's also a bit unclear. I **think** you mean that requests with a certain type of security measure aren't captured by ad-blockers. But I think it's also a bit confusing to use the word "captured" because we also say that the ad-blocker prevents data from being captured by the SDK. I've attempted an edit, but update as it makes sense.

```suggestion
A tunnel is an HTTP endpoint that acts as a proxy between Sentry and your application. Because you control this server, there is no risk of any requests sent to it being blocked. When the endpoint lives under the same origin (although it does not have to in order for the tunnel to work), the browser will not treat any requests to the endpoint as a third-party request. As a result, these requests will have different security measures applied which, by default, don't trigger ad-blockers. A quick summary of the flow can be found below.
```

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -105,9 +115,9 @@ if (isset($header["dsn"])) {
 
 With this set up, events should not be blocked by ad-blockers, as they are never sent to a third-party URL.
 
-### Using the JavaScript Proxy API
+If your use-case is related to the SDK package itself being blocked, any of the following solutions will help you mitigate this issue.
```
imatwawana, line 118:
```suggestion

If your use-case is related to the SDK package itself being blocked, any of the following solutions will help you resolve this issue.
```

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -105,9 +115,9 @@ if (isset($header["dsn"])) {
...
 
-When you are using our CDN, ad-blocking or script-blocking extensions may prevent our SDK from being fetched and initialized properly. Because of this, any call to SDKs API will fail and may cause your application to misbehave. If this applies to your use-case, any of the following solutions will mitigate this issue.
+### Using Package Directly
```
imatwawana, line 120: Just adding a space before the heading
```suggestion

### Using Package Directly
```

imatwawana (approved):
Mostly minor wording changes

kamilogorek:
Thanks!

PeloWriter:
Thanks for updating this, Kamil. I can't comment directly on the file (since it isn't new), but I suggest we add this to the frontmatter (after line 3):

keywords: ["adblocker", "blocked", "tunnel"]

PR #3749, opened 2021-06-16
ref: macho -> mach-o

bruno-garcia:
Possibly a nitpick on the same boat as `Javascript` -> `JavaScript` and `.Net` -> `.NET` but the correct format for it is `Mach-O`: https://en.wikipedia.org/wiki/Mach-O

[👁️ ](https://www.youtube.com/watch?v=yxxj73JCV4Q)


PeloWriter (approved):
Thanks, Bruno

jan-auer (approved):

PR #3754, opened 2021-06-17
feat(performance): Per-Project Thresholds

shruthilayaj:
Users can now set Apdex/User Misery thresholds per Project
as opposed to a single org level threshold if they have
the feature flag enabled.

### DIFF
--- a/src/docs/product/performance/metrics.mdx
+++ b/src/docs/product/performance/metrics.mdx
@@ -8,6 +8,17 @@ description: "Learn more about Sentry's Performance metrics such as, Apdex, fail
 
 Metrics provide insight about how users are experiencing your application. In [Performance](/product/performance/), we'll set you up with a few of the basic metrics to get you started. For further customizations on target thresholds, feel free to build out a query using the Discover [Query Builder](/product/discover-queries/query-builder/). By identifying useful thresholds to measure your application, you have a quantifiable measurement of your application's health. This means you can more easily identify when errors occur or if performance issues are emerging.
 
+## Set Thresholds per Project
+<Note>
+
+This feature is available only if you're in the Early Adopter program. Features available to Early Adopters are still in-progress and may have bugs. We recognize the irony. If you’re interested in being an Early Adopter, you can turn your organization’s Early Adopter status on/off in General Settings. This will affect all users in your organization and can be turned back off just as easily.
+
+</Note>
+
+You can configure response time thresholds per Project for the [Apdex](#apdex) and [User Misery](#user-misery) scores. Additionally, you can also set which response time metric (e.g. Transaction Duration, [Largest Contentful Paint](/product/performance/web-vitals/#largest-contentful-paint-lcp)) is used in the calculations. This allows you to use different standards for gauging performance across multiple projects.
+
+Configure Project-level thresholds in **[Project] > Settings > Performance**.
+
 ## Apdex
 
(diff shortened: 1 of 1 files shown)


src/docs/product/performance/metrics.mdx
```diff
@@ -8,6 +8,17 @@ description: "Learn more about Sentry's Performance metrics such as, Apdex, fail
...
 
+## Set Thresholds per Project
+<Note>
+
+This feature is available only if you're in the Early Adopter program. Features available to Early Adopters are still in-progress and may have bugs. We recognize the irony. If you’re interested in being an Early Adopter, you can turn your organization’s Early Adopter status on/off in General Settings. This will affect all users in your organization and can be turned back off just as easily.
+
+</Note>
+
+You can configure response time thresholds per Project for the [Apdex](#apdex) and [User Misery](#user-misery) scores. Additionally, you can also set which response time metric (e.g. Transaction Duration, [Largest Contentful Paint](/product/performance/web-vitals/#largest-contentful-paint-lcp)) is used in the calculations. This allows you to use different standards for gauging performance across multiple projects.
```
PeloWriter, line 18: Thanks for this update. One style tip: we avoid Latin abbreviations
```suggestion
You can configure response time thresholds per Project for the [Apdex](#apdex) and [User Misery](#user-misery) scores. Additionally, you can set which response time metric (for example, Transaction Duration, [Largest Contentful Paint](/product/performance/web-vitals/#largest-contentful-paint-lcp)) is used in the calculation, which allows you to use different standards for gauging performance across multiple projects.
```

PeloWriter (approved):
thank you!

shruthilayaj:
Will merge this on June 21 when the feature is ready to EA, reverting to draft until then!

src/docs/product/performance/metrics.mdx
```diff
@@ -8,6 +8,17 @@ description: "Learn more about Sentry's Performance metrics such as, Apdex, fail
...
 
+## Set Thresholds per Project
+<Note>
+
+This feature is available only if you're in the Early Adopter program. Features available to Early Adopters are still in-progress and may have bugs. We recognize the irony. If you’re interested in being an Early Adopter, you can turn your organization’s Early Adopter status on/off in General Settings. This will affect all users in your organization and can be turned back off just as easily.
+
+</Note>
+
+You can configure response time thresholds per Project for the [Apdex](#apdex) and [User Misery](#user-misery) scores. Additionally, you can set which response time metric (for example, Transaction Duration, [Largest Contentful Paint](/product/performance/web-vitals/#largest-contentful-paint-lcp)) is used in the calculation, which allows you to use different standards for gauging performance across multiple projects.
```
wmak, line 18: minor: 🤔 would it be too specific to mention some example thresholds here?

wmak (approved):
LGTM

src/docs/product/performance/metrics.mdx
shruthilayaj: hmm just saw this, since the settings page is pretty straightforward and doesn't have options, we may not need to. I'll merge this as is but let @PeloWriter weigh in and make a follow up PR if we think examples can be useful.

wmak: Yeah, we've had folks ask what good values are for things "what's a good sample rate" "what's a good apdex" so i wonder if some example values in the docs can help them out

PR #3757, opened 2021-06-18
Clarify disabling auto session tracking on React Native

apuyou:
This improves instructions on how to disable auto session tracking.

Most importantly, the correct attribute is used in the code block, which is most likely to be copy/pasted.


PeloWriter:
@jennmueng - can you take a look, please?

jennmueng (approved):
LGTM, Thanks!

PR #3761, opened 2021-06-19
tunnel example dotnet

bruno-garcia:
Links to a location pending this PR: https://github.com/getsentry/examples/pull/93


PeloWriter (approved):
On line 82, can we please update to say
"Here's an example for your server component:" 

because we're expanding beyond PHP.

jan-auer (approved):

PR #3775, opened 2021-06-22
fix(nextjs): Indicate that the serverless target is not supported

AbhiPrasad:
@sentry/nextjs does not support the serverless target. This patch
adds a note to the manual configuration section of the nextjs docs
that surfaces this. It also recommends user's use the
experimental-serverless-trace option as an alternative.

The note was put in an alert under manual configuration as only a few
users will hit this situation.

### DIFF
--- a/src/platforms/javascript/guides/nextjs/manual-setup.mdx
+++ b/src/platforms/javascript/guides/nextjs/manual-setup.mdx
@@ -62,7 +62,7 @@ export default withSentry(handler);
 
 You can include your DSN directly in these two files, or provide it in either of two environment variables, `SENTRY_DSN` or `NEXT_PUBLIC_SENTRY_DSN`.
 
-<Alert level="warning" title="Note">
+<Alert level="warning" title="Serverless Environments">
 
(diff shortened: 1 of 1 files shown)


src/platforms/javascript/guides/nextjs/manual-setup.mdx
```diff
@@ -62,6 +62,12 @@ export default withSentry(handler);
 
 You can include your DSN directly in these two files, or provide it in either of two environment variables, `SENTRY_DSN` or `NEXT_PUBLIC_SENTRY_DSN`.
 
+<Alert level="warning" title="Note">
```
PeloWriter, line 65: If it's just a note, we can use the Note/Note format, which will display with a purple background.

If we want to ensure it is truly displayed as a warning, with a red background, then we shouldn't use "Note" as the title. Instead, use "Serverless Environments" or similar.

PeloWriter (commented):
Suggestion for UX, other than that, the language LGTM.

src/platforms/javascript/guides/nextjs/manual-setup.mdx
AbhiPrasad: I wanted to emphasize this as a warning, so I changed the title to `Serverless Environments`.

PR #3776, opened 2021-06-22
Update Grouping Content

PeloWriter:
In preparation for future functionality, we're moving the grouping content into SDK-specific content and product content. 

The SDK fingerprinting content now lives in Usage > SDK Fingerprinting. The data management content is now in product > Data Management > Issue Grouping. This also had the knock-on effect of removing a series of `include` folders that weren't really needed (as the product side doesn't currently support the use of `include` statements).


src/docs/product/data-management-settings/event-grouping/index.mdx
```diff
@@ -4,14 +4,15 @@ sidebar_order: 0
 redirect_from:
   - /data-management/rollups/
   - /learn/rollups/
+  - /data-management/event-grouping/
```
jan-auer, line 7: Shouldn't this redirect from the platform pages, or is that what it does?

jan-auer (approved):
The document is called "Issue Grouping", so since we're moving it already, I'd also use `issue-grouping` in the URL rather than `event-grouping`.

src/includes/grouping/example/_default.mdx
```diff
@@ -1,10 +0,0 @@
-```discover {tabTitle: Enhancement Rules}
-# mark all functions in the std namespace to be outside the app
-family:native stack.function:std::*       -app
```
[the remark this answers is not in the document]
PeloWriter: Currently, on the product side of the house, the include statements don't work as they do for platforms. For now, we can add more examples and hard code them. In future, with more tools support, we can enable platform-specific support/include statements on the product side of our content.

src/docs/product/data-management-settings/event-grouping/index.mdx
PeloWriter: This is the location in which it used to reside.

PR #3777, opened 2021-06-23
update proguard.mdx

mathrocco:
- make it clear that the sentry.properties file should be in Android project root (it can be misleading for flutter projects setup)
- make it clear that properties should not be declared around quotes

### DIFF
--- a/src/platforms/android/common/proguard.mdx
+++ b/src/platforms/android/common/proguard.mdx
@@ -40,7 +40,7 @@ buildscript {
 }
 ```
 
-The plugin will automatically generate appropriate ProGuard mapping files and upload them when you run `gradle assemble{BuildVariant}`. For example, `assembleRelease` — Release is the default, but the plugin works for others if you have enabled ProGuard/R8. The credentials for the upload step are loaded via environment variables or from a `sentry.properties` file in your project root.
+The plugin will automatically generate appropriate ProGuard mapping files and upload them when you run `gradle assemble{BuildVariant}`. For example, `assembleRelease` — Release is the default, but the plugin works for others if you have enabled ProGuard/R8. The credentials for the upload step are loaded via environment variables or from a `sentry.properties` file in your Android project root.
 
 For more information, see the [full sentry-cli documentation](/product/cli/configuration/#configuration-values).
 
@@ -51,6 +51,13 @@ defaults.project=your-project
 defaults.org=your-org
 auth.token=YOUR_AUTH_TOKEN
 ```
+
+<Note>
+
+Values inside `sentry.properties` file should not be wrapped around quotes, you should only replace the placeholders. Otherwise, it may lead to authentication or resource errors.
+
+</Note>
+
 <Note>
 
 You can find your authentication token [on the Sentry API page](https://sentry.io/api/). For more information about the available configuration options, see the [full sentry-cli documentation](/product/cli/configuration/#configuration-values).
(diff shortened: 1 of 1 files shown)


src/platforms/android/common/proguard.mdx
```diff
@@ -51,6 +51,13 @@ defaults.project=your-project
 defaults.org=your-org
 auth.token=YOUR_AUTH_TOKEN
 ```
+
+<Note>
+
+Values inside `sentry.properties` file should not be wrapped around quotes, you should only replace the placeholders. Otherwise, it may lead to authentication or resource errors.
```
imatwawana, line 57:
```suggestion
Values inside `sentry.properties` file should not be wrapped around quotes; you should only replace the placeholders. Otherwise, it may lead to authentication or resource errors.
```

imatwawana (approved):
Minor punctuation fix

src/platforms/android/common/proguard.mdx
```diff
@@ -51,6 +51,13 @@ defaults.project=your-project
 defaults.org=your-org
 auth.token=YOUR_AUTH_TOKEN
 ```
+
+<Note>
+
+Values inside `sentry.properties` file should not be wrapped around quotes; you should only replace the placeholders. Otherwise, it may lead to authentication or resource errors.
+
+</Note>
```
marandaneto, line 59: @mathrocco what made you believe it should have been wrapped in quotes? the example above isnt.

src/platforms/android/common/proguard.mdx
```diff
@@ -40,7 +40,7 @@ buildscript {
 }
 ```
 
-The plugin will automatically generate appropriate ProGuard mapping files and upload them when you run `gradle assemble{BuildVariant}`. For example, `assembleRelease` — Release is the default, but the plugin works for others if you have enabled ProGuard/R8. The credentials for the upload step are loaded via environment variables or from a `sentry.properties` file in your project root.
+The plugin will automatically generate appropriate ProGuard mapping files and upload them when you run `gradle assemble{BuildVariant}`. For example, `assembleRelease` — Release is the default, but the plugin works for others if you have enabled ProGuard/R8. The credentials for the upload step are loaded via environment variables or from a `sentry.properties` file in your Android project root.
```
marandaneto, line 43: this is wrong, the project root isn't necessarily an Android project.
A Gradle project with many projects Java/Android/Groovy etc isn't an Android project root and still requires to be in the project root folder.
eg https://github.com/getsentry/sentry-java/blob/main/sentry.properties

src/platforms/android/common/proguard.mdx
mathrocco: I'm not used to this `.properties` file format, so in troubleshooting I accidentally left the quotes, which led me to unauthorized errors. Since it took me some time to realize it, I thought it could help save others' time.

src/platforms/android/common/proguard.mdx
mathrocco: Isn't it the Android doc?

mathrocco:
@marandaneto I replied the comments you left. If you don't think this change is needed, I can just close this PR. Let me know, thanks 👍🏼

src/platforms/android/common/proguard.mdx
marandaneto: yes, but you may have an android module inside of a bigger Gradle solution that may have different projects, the current description fits both use cases.

PR #3800, opened 2021-06-25
Release Health for Unity

bruno-garcia:
Worth noting:

1. I didn't link to `https://docs.sentry.io/platforms/unity/configuration/releases/` as we did on other pages (like [.NET](https://docs.sentry.io/platforms/dotnet/), [Apple](https://docs.sentry.io/platforms/apple/guides/ios/), ..) but instead linked to the ["Release and Health" part of the SDK itself](https://docs.sentry.io/platforms/unity/configuration/releases/). The reason is that we have more specific information like for example we detect release automatically from Unity.

Would it make sense to edit the common content for Release and Health and link to `https://docs.sentry.io/platforms/unity/configuration/releases/` from there instead?

2. I rephrased it. On the other docs we say:

> Release Health tracks crash free users and sessions

So to be clear we're not "tracking users"([see this answer](https://stackoverflow.com/questions/68067822/is-it-possible-to-avoid-collecting-specific-data-in-sentry/68108641#68108641)) I reworded as:

> Release Health to keep track of crash free users and sessions

Downside is that Unity would differ from the rest so like best we agree on one approach and apply everywhere.

### DIFF
--- a/src/includes/configuration/auto-session-tracking/dotnet.mdx
+++ b/src/includes/configuration/auto-session-tracking/dotnet.mdx
@@ -1,0 +1,21 @@
+We mark the session as:
+- Health
+- Abnormal
+- Errored: if the SDK captures an event that contains an exception (this includes manually captured errors).
+- Crashed
+
+By default, the .NET SDK is creating a session on application startup and end it on shut down. To disable sending sessions, set the `AutoSessionTracking` flag to `false`.
+
+```csharp {tabTitle:C#}
+using Sentry;
+
+// Add this to the SDK initialization callback
+options.AutoSessionTracking = false; // default: true
+```
+
+```fsharp {tabTitle:F#}
+open Sentry
+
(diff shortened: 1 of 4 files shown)


PeloWriter:
> I didn't link to https://docs.sentry.io/platforms/unity/configuration/releases/ as we did on other pages (like .NET, Apple, ..) but instead linked to the "Release and Health" part of the SDK itself. The reason is that we have more specific information like for example we detect release automatically from Unity.
We typically point the bullet on Release Health to the index page in product, so users can understand why its of value. You can shift it to point to the configuration content in the SDK itself if you think that's more useful to Unity readers.

> So to be clear we're not "tracking users"(see this answer) 
Release Health to keep track of crash free users and sessions

Yes, this edit is both more technically correct and better. Let's make that change throughout.

PeloWriter:
sorry, saw one typo :)

PR #3815, opened 2021-06-29
ref(membership): Update add/remove repo role

ceorourke:
Update the table on the [membership](https://docs.sentry.io/product/accounts/membership/#membership
) page to reflect current member roles needed to remove a repository. I consolidated the rows since the same permissions are needed for adding and removing.

See https://github.com/getsentry/sentry/pull/18987 where we updated this in the product.

Screen Shot 2021-06-29 at 3 55 16 PM


scefali (approved):

leedongwei (approved):
Thanks @ceorourke

PR #3822, opened 2021-06-30
[Ruby migration guide] List another removed config method: silence_ready

ragesoss:


imatwawana:
@st0012 - Can you review?

st0012 (approved):
good catch. thanks 👍

PR #3826, opened 2021-06-30
feat(nextjs): Document options for disabling webpack plugin

lobsterkatie:
This PR adds docs for the options introduced in https://github.com/getsentry/sentry-javascript/pull/3771, which allow the user to disable the addition of `SentryWebpackPlugin` to server and client builds.

### DIFF
--- a/src/platforms/javascript/guides/nextjs/manual-setup.mdx
+++ b/src/platforms/javascript/guides/nextjs/manual-setup.mdx
@@ -6,7 +6,7 @@ description: "Learn how to set up the SDK manually."
 
 If you can’t (or prefer not to) run the [configuration step](/platforms/javascript/guides/nextjs/#configure), you can follow the instructions below to configure your application.
 
-## Set Initialization Config Files
+## Create Initialization Config Files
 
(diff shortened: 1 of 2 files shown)


src/platforms/javascript/guides/nextjs/manual-setup.mdx
```diff
@@ -68,7 +68,7 @@ You can include your DSN directly in these two files, or provide it in either of
 
 </Alert>
 
-## Extend Default Webpack Usage
+## Use `withSentryConfig`
```
iker-barriocanal, line 71: I rather titles focus on the goal over what you need to do. Before the change, the title was telling you what the section is about; after the change, you have to guess it. Wdyt?

lobsterkatie: Fair point. I changed it to `Extend Next.js Configuration`, since that makes it immediately clear where they're supposed to make the changes.

HazAT (approved):

iker-barriocanal (approved):

PR #3827, opened 2021-07-01
feat(apidocs): high level api descriptions markdown rendering

JoshFerge:
for the new SCIM api we'd like to render a markdown description describing the API at a high level and giving context about using it. This PR creates a new node for the rendered markdown if the field `x-display-description` on the tag is set to true, and displays it on the page if it exists. Screenshot below of what that will look like. (text is not finalized).

Screen Shot 2021-06-30 at 4 08 51 PM

requires https://github.com/getsentry/sentry/pull/27054


src/gatsby/plugins/gatsby-plugin-openapi/gatsby-node.ts
```diff
@@ -32,6 +32,36 @@ export const sourceNodes = async (
 
     const parsedContent = parseContent();
 
+    // temporarily create a mock api description so the apiDoc query doesnt fail
```
JoshFerge, line 35: seems weird but this is what gatsby recommends doing

JoshFerge:
note: also seems like prettier isn't being ran in pre-commit, so spacing might not be perfect

src/gatsby/plugins/gatsby-plugin-openapi/gatsby-node.ts
NisanthanNanthakumar: @JoshFerge hmm why is there no nodes present. I see `'x-display-description'` exists in the derefed openapi. https://github.com/getsentry/sentry-api-schema/blob/main/openapi-derefed.json#L25

JoshFerge: kk went ahead and removed this -- will wait to merge until other scim PRs are in.

NisanthanNanthakumar (approved):

PR #3832, opened 2021-07-05
Java: Add BOM docs.

### LINKED ISSUE #3652: Document sentry-bom for Java and Android
## Issue

Document the new sentry-bom that control transitive dependencies

https://github.com/getsentry/sentry-java/pull/1486

## Description

we're offering a new package that control transitive dependencies

### Suggested Solution

in case they want to support for example, sentry-android and sentry-android-okhttp, etc they don't need to set the version to all dependencies but rather the sentry-bom, eg

```
    implementation platform("io.sentry:sentry-bom:{version}")
    implementation "io.sentry:sentry-android"
    implementation "io.sentry:sentry-android-okhttp"
```

maciejwalkowiak:
Closes #3652


src/platforms/common/index.mdx
```diff
@@ -20,6 +20,8 @@ Sentry captures data by using an SDK within your application’s runtime.
 
 <PlatformContent includePath="getting-started-install" />
 
+<PlatformContent includePath="getting-started-install-bom" supported={["java"]} />
```
maciejwalkowiak, line 23: Perhaps there is a better way to include this piece into every Java guide?

maciejwalkowiak:
cc @marandaneto @bruno-garcia

maciejwalkowiak:
I am not 100% convinced that this is the right place to put BOM docs. AFAIR we aimed "Getting Started" for having only essential information - BOM docs are useful only if more than one Sentry dependency is used (Android + Coroutines, Spring + Logback etc).

marandaneto:
> I am not 100% convinced that this is the right place to put BOM docs. AFAIR we aimed "Getting Started" for having only essential information - BOM docs are useful only if more than one Sentry dependency is used (Android + Coroutines, Spring + Logback etc).

based on my comment here https://github.com/getsentry/sentry-docs/issues/3652#issuecomment-872967071
I'd say its best to be within the install step, so here most of the cases -> https://github.com/getsentry/sentry-docs/tree/master/src/includes/getting-started-install

if the install steps don't include 2 dependencies or more, we don't add the bom, but rather to the integrations or guides that do, eg https://github.com/getsentry/sentry-docs/blob/master/src/platforms/android/configuration/integrations/fragment.mdx
we'd just need to find all the files that add 2 or more dependencies, and update them to use the bom

maciejwalkowiak:
I see a scenario where users uses Spring and Logback - there is no dedicated page for this combination, and there wouldn't be a place to put BOM docs. 

What about having a separate section "Using BOM" or "Managing Dependencies" somewhere in the main menu:

image

marandaneto:
> I see a scenario where users uses Spring and Logback - there is no dedicated page for this combination, and there wouldn't be a place to put BOM docs.
> 
> What about having a separate section "Using BOM" or "Managing Dependencies" somewhere in the main menu:
> 
> image

that makes sense, not sure how the `BOM` word is known, but something related to Dependencies makes sense to me.

maciejwalkowiak:
@marandaneto updated.

src/platforms/android/common/dependencies.mdx
```diff
@@ -0,0 +1,45 @@
...
+sidebar_order: 1000
+---
+
+## Managing versions for multiple Sentry dependencies
+
+When multiple Sentry dependencies are used, to avoid specifying the version of each dependency, you can use a *BOM* or *Bill Of Materials*.
+
+Using Maven you can add the following to `dependencyManagement` section in your `pom.xml`:
+
+```xml {tabTitle:Maven}{filename:pom.xml}
```
bruno-garcia, line 12: Does it make sense to add the maven information on Android? Wouldn't Android be 90% (or more?) gradle?

bruno-garcia (approved):
I think this is something @PeloWriter and @marandaneto might have thoughts but if they are happy with it, it also LGTM

One question/suggestion about Android. 
Another detail is to possibly not have it as a top level item. Maybe add under configuration?

src/platforms/android/common/dependencies.mdx
```diff
@@ -0,0 +1,45 @@
+---
+title: Managing Dependencies
```
PeloWriter, line 2: Since this page is devoted to using a BOM, the page should reflect that
```suggestion
title: Using a BOM
description: "Learn more about using a bill of materials with multiple Sentry dependencies."
```

PeloWriter: I also agree that this isn't quite the right spot. I'd do the following:
1. Move the page to a subpage of Configuration
2. Add a link to the page where we currently tell users about NDK and the dependency on Gson (`src/includes/getting-started-install/android`) adding "If you are using multiple Sentry dependencies, you can add a  [bill of materials](link to new page) to avoid specifying the version of each dependency."

src/platforms/android/common/dependencies.mdx
```diff
@@ -0,0 +1,45 @@
...
+sidebar_order: 1000
+---
+
+## Managing versions for multiple Sentry dependencies
```
PeloWriter, line 6: We don't need this title as the content is pretty descriptive.

src/platforms/android/common/dependencies.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+When multiple Sentry dependencies are used, to avoid specifying the version of each dependency, you can use a *BOM* or *Bill Of Materials*.
```
PeloWriter, line 8:
```suggestion
When you are using multiple Sentry dependencies, you can avoid specifying the version of each dependency with a *BOM* or *Bill Of Materials*.
```

src/platforms/android/common/dependencies.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+Using Maven you can add the following to `dependencyManagement` section in your `pom.xml`:
```
PeloWriter, line 10:
```suggestion
Using Maven, add the following to the `dependencyManagement` section in your `pom.xml`:
```

src/platforms/android/common/dependencies.mdx
```diff
@@ -0,0 +1,45 @@
...
+            <artifactId>sentry-bom</artifactId>
+            <version>{{ packages.version('sentry.java', '5.1.0') }}</version>
+            <type>pom</type>
+            <scope>import</scope>
+        </dependency>
+    </dependencies>
+</dependencyManagement>
+```
+
+and then use dependencies without specifying a version, for example:
```
PeloWriter, line 26:
```suggestion
Then use dependencies without specifying a version, for example:
```

src/platforms/android/common/dependencies.mdx
```diff
@@ -0,0 +1,45 @@
...
+    <groupId>io.sentry</groupId>
+    <artifactId>sentry-android</artifactId>
+</dependency>
+<dependency>
+    <groupId>io.sentry</groupId>
+    <artifactId>sentry-android-fragment</artifactId>
+</dependency>
+```
+
+Using Gradle 5.0 or higher, you can add the following to the `dependencies` section in your `build.gradle`:
```
PeloWriter, line 39: Given @bruno-garcia's comment above, do you want this content to move up on the page? (prior to the Maven content?)

src/platforms/java/common/dependencies.mdx
```diff
@@ -0,0 +1,45 @@
+---
+title: Managing Dependencies
```
PeloWriter, line 2: Is this entirely duplicated? I can see how that's an issue without sibling inheritance (Java and Android live separately). In that case, we can do one of two things:
1. Edit this content as noted in the Android comments, and recognize we've got duplicate content
2. Put the content into a giant `include`, then set up the page, calling the `include` from both locations (a kludge, but it would ensure the content is consistent between the two platforms)

src/platforms/android/common/dependencies.mdx
marandaneto: if its an Android page only, only Gradle would be fine

src/platforms/java/common/dependencies.mdx
maciejwalkowiak: Package names are different, and since we will remove Maven samples from Android page I think they can stay separate.

src/platforms/android/common/dependencies.mdx
maciejwalkowiak: Removed maven part completely from android.

src/platforms/android/common/dependencies.mdx
maciejwalkowiak: 👍

maciejwalkowiak:
@PeloWriter @marandaneto @bruno-garcia please take another look.

bruno-garcia (approved):
👍

bruno-garcia:
@maciejwalkowiak seems we didn't ship `bom` yet though: https://github.com/getsentry/sentry-java/pull/1486/files

I jumped the gun here by merging this, should I revert?

maciejwalkowiak:
@bruno-garcia I believe it should be reverted, and merged only once BOM PR is merged and 5.1.0 is released.

bruno-garcia:
Raised #3939

bruno-garcia:
Draft to add this back once 5.1.0 is out: https://github.com/getsentry/sentry-docs/pull/3940

PR #3833, opened 2021-07-05
ref(apple): Rename option for enabling auto performance

philipphofmann:
enableAutoUIPerformanceTracking was renamed to enableAutoPerformanceTracking.

Only merge after https://github.com/getsentry/sentry-cocoa/pull/1209 is released.

### DIFF
--- a/src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
+++ b/src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
@@ -44,7 +44,7 @@ import Sentry
 
 SentrySDK.start { options in
     options.dsn = "___PUBLIC_DSN___"
-    options.enableAutoUIPerformanceTracking = false
+    options.enableAutoPerformanceTracking = false
 }
 ```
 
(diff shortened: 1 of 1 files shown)


jennmueng (approved):

src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -44,7 +44,7 @@ import Sentry
 
 SentrySDK.start { options in
     options.dsn = "___PUBLIC_DSN___"
-    options.enableAutoUIPerformanceTracking = false
+    options.enableAutoPerformanceTracking = false
```
PeloWriter, line 47: When this goes GA, could we please also add this option to the Basic Options file, noting that it's an opt-out and that a sample rate ("adding performance") must be done so this takes effect? File is options.mdx. And we update it using these [guidelines](https://docs.sentry.io/contributing/approach/sdk-docs/write-configuration/#basic-options).

PeloWriter (approved):
We should also add this option to our Basic Options content.

PR #3841, opened 2021-07-07
feat: Clearer RN documentation on Source Maps and release and dist values

jennmueng:
- Mention in the "Releases and Health" page that a `dist` value is required.
- Completely rewrite the React Native sourcemaps page to be clear and to account for all the common pitfalls I've seen.
- Move the sourcemaps page to a top level page instead of being under "Manual Setup" so it's noticable.
- Add a warning for setting the release and dist values that sourcemaps will now need to be manually uploaded.
- Add instructions on how to disable the automatic source map upload script.

Fixes #3820
Fixes #3821

PR #3843, opened 2021-07-07
feat: Initial troubleshooting page for RN

jennmueng:
Initial troubleshooting page for React Native with mentions to common issues and their fixes.

1. Source maps
2. Xcode CLI command
3. Release Health/Codepush

### DIFF
--- a/src/platforms/react-native/troubleshooting.mdx
+++ b/src/platforms/react-native/troubleshooting.mdx
@@ -12,16 +12,16 @@ When there is an issue with source maps on your event, there will usually be an 
 
 Common fixes to this issue include:
 
-- Checking that both the `release` and `dist` and are set on the event by [setting it in the call to `init`](/platforms/react-native/configuration/releases/#bind-the-version). 
+- Checking that both the `release` and `dist` and are set on the event by [setting it in the call to `init`](/platforms/react-native/configuration/releases/#bind-the-version).
 - Checking that source maps are correctly uploaded with the exact same `release` and `dist` values as the event. They **must match** for source maps to upload correctly.
 - If you set a custom `release` and `dist`, confirming that you manually uploaded the sourcemaps since our automatic source map upload script will not detect the custom values.
-- Checking that you are not disabling the `RewriteFrames` default integration by filtering it out or setting `defaultIntegrations: false`.
+- Checking that you are not disabling the [`RewriteFrames`](/platforms/javascript/configuration/integrations/plugin/#rewriteframes) default integration by overwriting it, filtering it out, or setting `defaultIntegrations: false`.
 
 [For more details, please read the guide on React Native source maps.](/platforms/react-native/sourcemaps)
 
 ## iOS Build Script Failed
 
-Try passing `--force_foreground` to the Sentry CLI command in the build script. This is possibly a bug with our CLI that we are investigating.
+Try passing `--force_foreground` to the Sentry CLI command in the build script. This is possibly a bug with our CLI that [we are investigating](https://github.com/getsentry/sentry-react-native/issues/1424).
 
 ```bash
 # ...
--- a/src/platforms/react-native/troubleshooting.mdx
+++ b/src/platforms/react-native/troubleshooting.mdx
@@ -1,0 +1,35 @@
+---
+title: Troubleshooting
+description: "Troubleshoot and resolve common issues with the React Native SDK."
+sidebar_order: 1000
+---
+
+## Source Maps
+
+When there is an issue with source maps on your event, there will usually be an error dialog at the top of the event page on your dashboard:
+
+> There was 1 problem processing this event: Source code was not found
+
+Common fixes to this are:
+
+- Make sure that both the `release` and `dist` and are set on the event by [setting it in the call to `init`](/platforms/react-native/configuration/releases/#bind-the-version).
+- Make sure that the source maps are correctly uploaded with the exact same `release` and `dist` values as the event.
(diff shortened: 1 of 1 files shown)


src/platforms/react-native/troubleshooting.mdx
```diff
@@ -0,0 +1,35 @@
+---
+title: Troubleshooting
+description: "Troubleshoot and resolve common issues with the React Native SDK."
+sidebar_order: 1000
+---
+
+## Source Maps
+
+When there is an issue with source maps on your event, there will usually be an error dialog at the top of the event page on your dashboard:
```
PeloWriter, line 9:
```suggestion
When there is an issue with source maps on your event, there will usually be an error dialog at the top of the Issue Details page in [sentry.io](https://sentry.io):
```

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+> There was 1 problem processing this event: Source code was not found
+
+Common fixes to this are:
```
PeloWriter, line 13:
```suggestion
Common fixes to this issue include:
```

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+- Make sure that both the `release` and `dist` and are set on the event by [setting it in the call to `init`](/platforms/react-native/configuration/releases/#bind-the-version).
```
PeloWriter, line 15:
```suggestion
- Checking that both the `release` and `dist` and are set on the event by [setting it in the call to `init`](/platforms/react-native/configuration/releases/#bind-the-version). 
```

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -0,0 +1,35 @@
...
+- Make sure that the source maps are correctly uploaded with the exact same `release` and `dist` values as the event.
```
PeloWriter, line 16:
```suggestion
- Checking that source maps are correctly uploaded with the exact same `release` and `dist` values as the event. They **must match** for source maps to upload correctly.
```

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -0,0 +1,35 @@
...
+- If you set a custom `release` and `dist` make sure that you manually upload the sourcemaps as our automatic sourcemap upload script will not detect the custom values.
```
PeloWriter, line 17:
```suggestion
- If you set a custom `release` and `dist`, confirming that you manually uploaded the sourcemaps since our automatic source map upload script will not detect the custom values.
```

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -0,0 +1,35 @@
...
+- Make sure that you are not disabling the `RewriteFrames` default integration by filtering it out or setting `defaultIntegrations: false`.
```
PeloWriter, line 18:
```suggestion
- Checking that you are not disabling the `RewriteFrames` default integration by filtering it out or setting `defaultIntegrations: false`.
```

PeloWriter: We don't seem to have `RewriteFrames` (or any integrations) documented for RN. Do we want to link to the content in JS?

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+[For more details, please read the guide on React Native source maps.](/platforms/react-native/sourcemaps)
+
+## iOS Build Script Failed
+
+Try passing `--force_foreground` to the Sentry CLI command in the build script. This is possibly a bug with our CLI that we are investigating.
```
PeloWriter, line 24: Can we add a link to the the issue that tracks this potential bug? That way, as soon as it's resolved, if we don't immediately update this content, users will know of the resolution.

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+```bash
+# ...
+../node_modules/@sentry/cli/bin/sentry-cli react-native xcode \
+  ../node_modules/react-native/scripts/react-native-xcode.sh --force_foreground
+```
+
+## Release Health
+
+- If your release health statistics are being attributed to the wrong release, make sure you [pass the `release` and `dist` to the `init` call](/platforms/react-native/configuration/releases/#bind-the-version).
```
PeloWriter, line 34:
```suggestion
- If your release health statistics are being attributed to the wrong release, confirm that you [pass the `release` and `dist` to the `init` call](/platforms/react-native/configuration/releases/#bind-the-version).
```

PeloWriter (approved):

src/platforms/react-native/troubleshooting.mdx
jennmueng: Yes we should!

jennmueng: @PeloWriter Actually, additional thought is that we should include the integrations content in RN as well. All the pluggable integrations and most (but not all) of the default integrations are in RN as well.

jennmueng:
Let's hold off on merging this after #3841 because this links to that page.

src/platforms/react-native/troubleshooting.mdx
PeloWriter: That makes sense. The sad part is that we don't have the ability (yet) to inherit from siblings (that is, since RN is at the same level as the JS content, we can't just inherit the JS content). The best way (it's a kludge for sure) to resolve this is to build those pages out, then `include` them in both RN and JS. Alternately, just linking works for now until we resolve this issue.

PR #3845, opened 2021-07-07
docs(alerts): Add session percentage alert condition

scttcper:
related https://github.com/getsentry/develop/pull/373

https://getsentry.atlassian.net/browse/WOR-1049

Some docs on the new % of session based alerting
image

### DIFF
--- a/src/docs/product/alerts/create-alerts/issue-alert-config.mdx
+++ b/src/docs/product/alerts/create-alerts/issue-alert-config.mdx
@@ -33,6 +33,7 @@ Give your alert a descriptive name, such as the team affected and the topic of t
 - Changes state from `ignored` to `unresolved`
 - Seen more than a certain number of times in an interval
 - Seen by more than a certain number of unique users in an interval
+- *Beta* An issue affects more than {X} percent of [sessions](/product/releases/health/#session) in {time}. This is an approximation and will trigger when the ratio of the issue’s frequency to the number of sessions exceeds the threshold.
 
 Triggers are optional. If you don’t select a trigger, the "When" conditions are considered met by default. That is, _all_ issues will meet this condition.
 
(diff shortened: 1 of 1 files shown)


src/docs/product/alerts/create-alerts/issue-alert-config.mdx
```diff
@@ -33,6 +33,7 @@ Give your alert a descriptive name, such as the team affected and the topic of t
 - Changes state from `ignored` to `unresolved`
 - Seen more than a certain number of times in an interval
 - Seen by more than a certain number of unique users in an interval
+- *Beta* An issue affects more than {X} percent of [sessions](/product/releases/health/#session) in {time}. This is an approximation and will trigger when the ratio of the issue’s frequency to the number of sessions exceeds the threshold.
```
imatwawana, line 36: Does "beta" mean EA or something else? If it means EA, I've added an EA note in my suggested edit. If it means something else, then we might need to format it differently because it doesn't stand out enough in italics. Also wondering if you need everything after "approximation" as the initial definition seems pretty clear. 

```suggestion
- An issue affects more than {X} percent of [sessions](/product/releases/health/#session) in {time}. This is an approximation and will trigger when the ratio of the issue’s frequency to the number of sessions exceeds the threshold.

<Note>

The "percent of sessions" alert trigger is available only if you're in the Early Adopter program. Features available to Early Adopters are still in-progress and may have bugs. We recognize the irony.

If you’re interested in being an Early Adopter, you can turn your organization’s Early Adopter status on/off in General Settings. This will affect all users in your organization and can be turned back off just as easily.

</Note>
```

imatwawana (approved):
Had a question about beta vs EA and how much we need to explain, but otherwise looks good. Thanks!!

scttcper:
@imatwawana exactly what i meant by beta, thanks for putting it in the right language!

imatwawana:
Good stuff! LGTM!

PR #3852, opened 2021-07-09
feat(docs): update Lumen installation guide to include tracing

pataar:
I wasn't sure why the tracing didn't work. Turned out that an extra service provider was needed.


imatwawana:
@stayallive - Can you review this?

PeloWriter:
Thank you for this update. I agree that either @stayallive or @HazAT are the right folks for a technical review. The changes LGTM from the content side.

stayallive:
I'll make sure to take a look at it a bit later. But my first thoughts:

I feel like this is a bit too general since it now looks like the tracing service provider is required for normal operation (it is not) so we should add it to an appropriate section on how to enable tracing specifically for Lumen. 

Next to that I think we should also not replicate the full config file but link the the file on GitHub so it's always up-to-date and only duplicate relevant sections in the docs.

pataar:
I agree with you on the second point. Regarding the first point, if you're using Laravel auto discovery, both service providers will be loaded eventually. So you're having the same situation there. That's why I thought it would make more sense to load both providers.

stayallive:
I can agree on that, but since this is Lumen, (the "performance" framework) it makes sense to separate it in my eyes even though it would probably be a negligible difference.

pataar:
I'll try to update my PR to reflect that. Good feedback!

pataar:
@stayallive I've updated the documentation and made it in line with the other documents.

stayallive:
@pataar looks good! Thanks for taking the time to make these changes!

@PeloWriter this looks good to me 👍

src/platforms/php/guides/laravel/other-versions/lumen.mdx
```diff
@@ -37,16 +45,41 @@ public function report(Exception $exception)
...
+[Sentry] Transaction sent: b22f130764c940c69d6860c212849d48
+[Sentry] Event sent with ID: 4cde238791a9433c8876f4a545d9a95b
+```
+
+## Verify with code
 
-    // capture release as git sha
-    // 'release' => trim(exec('git log --pretty="%h" -n1 HEAD')),
-);
+You can easily verify that Sentry is capturing errors in your Laravel application by creating a debug route that will throw an exception:
```
pataar, line 77:
```suggestion
You can easily verify that Sentry is capturing errors in your Lumen application by creating a debug route that will throw an exception:
```

pataar:
Sorry, @stayallive, saw a reference to 'Laravel application' which didn't seem right. Corrected it to 'Lumen application'.

imatwawana:
Thank you, @pataar!

PR #3859, opened 2021-07-10
feat: Add Capacitor wizard

jennmueng:
Adds an initial Capacitor wizard

### DIFF
--- a/src/wizard/capacitor/index.md
+++ b/src/wizard/capacitor/index.md
@@ -1,5 +1,5 @@
 ---
-name: Cordova
+name: Capacitor
 doc_link: https://docs.sentry.io/platforms/javascript/guides/capacitor/
 support_level: production
 type: framework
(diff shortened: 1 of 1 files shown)


src/wizard/capacitor/index.md
```diff
@@ -0,0 +1,146 @@
...
+
+```javascript
+// App.js
+import * as Sentry from "@sentry/capacitor";
+
+Sentry.init({
+  dsn: "https://examplePublicKey@o0.ingest.sentry.io/0",
+
+  // Set your release version, such as "getsentry@1.0.0"
+  release: "my-project-name@<release-name>",
```
bruno-garcia, line 117: Where does the user learn about matching the release with uploaded sourcemaps?

bruno-garcia (approved):
Nice!

I wanted to propose adding performance monitoring but there's already quite a lot going on here. Lets see how this works first.

src/wizard/capacitor/index.md
jennmueng: @bruno-garcia Should we just link to [this section](https://docs.sentry.io/platforms/javascript/guides/capacitor/#source-maps) on the docs or should we also have a section in this wizard?

PR #3860, opened 2021-07-12
feat(python): Updates Sanic versions supported

ahmedetefy:
This PR updates the versions the SDK supports in the Python SDK Sanic integration

### DIFF
--- a/src/wizard/python/sanic.md
+++ b/src/wizard/python/sanic.md
@@ -11,9 +11,10 @@ Framework](https://github.com/huge-success/sanic). We support the following vers
 - `0.8`
 - `18.12`
 - `19.12`
+- `20.12`
 - Any version of the form `x.12` (LTS versions).
 
-**We do not support the latest version of Sanic.** Versions between LTS releases [have introduced breaking changes in the past](https://github.com/huge-success/sanic/issues/1532) without prior notice, so we cannot support them as they are too fast of a moving target.
+**We do support all versions of Sanic.** However, SDK Sanic support for versions between LTS releases should be considered Early Adopters. We might not be supporting all the features in those non-LTS versions, since non-LTS versions [have introduced breaking changes in the past](https://github.com/huge-success/sanic/issues/1532) without prior notice, so we cannot support them as they are too fast of a moving target.
 
 A Python version of 3.6 or greater is also required.
 
--- a/src/wizard/python/sanic.md
+++ b/src/wizard/python/sanic.md
@@ -14,7 +14,7 @@ Framework](https://github.com/huge-success/sanic). We support the following vers
 - `20.12`
 - Any version of the form `x.12` (LTS versions).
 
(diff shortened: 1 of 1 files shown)


src/wizard/python/sanic.md
```diff
@@ -11,9 +11,10 @@ Framework](https://github.com/huge-success/sanic). We support the following vers
 - `0.8`
 - `18.12`
 - `19.12`
+- `20.12`
 - Any version of the form `x.12` (LTS versions).
 
-**We do not support the latest version of Sanic.** Versions between LTS releases [have introduced breaking changes in the past](https://github.com/huge-success/sanic/issues/1532) without prior notice, so we cannot support them as they are too fast of a moving target.
+**We do support all versions of Sanic.** However, SDK Sanic support for versions between LTS releases should be considered while knowing that we might not be supporting all the features in those non-LTS versions, since non-LTS versions [have introduced breaking changes in the past](https://github.com/huge-success/sanic/issues/1532) without prior notice, so we cannot support them as they are too fast of a moving target.
```
HazAT, line 17:
```suggestion
**We do support all versions of Sanic.** However, SDK Sanic support for versions between LTS releases should be considered Early Adopters. Knowing that we might not be supporting all the features in those non-LTS versions, since non-LTS versions [have introduced breaking changes in the past](https://github.com/huge-success/sanic/issues/1532) without prior notice, so we cannot support them as they are too fast of a moving target.
```

HazAT (changes requested):

src/wizard/python/sanic.md
```diff
@@ -11,9 +11,10 @@ Framework](https://github.com/huge-success/sanic). We support the following vers
 - `0.8`
 - `18.12`
 - `19.12`
+- `20.12`
 - Any version of the form `x.12` (LTS versions).
 
-**We do not support the latest version of Sanic.** Versions between LTS releases [have introduced breaking changes in the past](https://github.com/huge-success/sanic/issues/1532) without prior notice, so we cannot support them as they are too fast of a moving target.
+**We do support all versions of Sanic.** However, SDK Sanic support for versions between LTS releases should be considered Early Adopters. We might not be supporting all the features in those non-LTS versions, since non-LTS versions [have introduced breaking changes in the past](https://github.com/huge-success/sanic/issues/1532) without prior notice, so we cannot support them as they are too fast of a moving target.
```
imatwawana, line 17: Tried to reword for improved clarity. Does this work?

```suggestion
**We do support all versions of Sanic.** However, Sanic SDK versions between LTS releases should be considered Early Adopter. We may not support all the features in these non-LTS versions, since non-LTS versions change quickly and [have introduced breaking changes in the past](https://github.com/huge-success/sanic/issues/1532), without prior notice. 
```

imatwawana (approved):
Reworded for improved clarity

src/wizard/python/sanic.md
ahmedetefy: Yeah this is better, just removed the `SDK` after Sanic cause it probably does not make sense

ahmedetefy: Thanks!

PR #3882, opened 2021-07-14
fix(traversing): update terminology and screenshot

doralchan:
Linked to https://github.com/getsentry/sentry/pull/27431

Screen Shot 2021-07-14 at 3 40 32 PM


src/docs/product/sentry-basics/tracing/event-detail.mdx
```diff
@@ -68,19 +68,19 @@ The span view may be limited to one project at a time if you're on the [Team pla
 
 </Alert>
 
-#### Traversing and Nesting Transactions
```
dashed, line 71: we might want to be explicit about "Nesting"

imatwawana: What about "Traversing and Adding Transactions" as the heading? Does that address your concern @dashed or do you think we need to stick with the "nesting" terminology?

dashed: I think we should be explicit that "Traversing" isn't the same as embedding/nesting/inlining/adding a transaction. I don't have a strong opinion of what the new product feature should be called.

src/docs/product/sentry-basics/tracing/event-detail.mdx
```diff
@@ -68,19 +68,19 @@ The span view may be limited to one project at a time if you're on the [Team pla
...
+#### Traversing Transactions
```
imatwawana, line 71:
```suggestion
#### Traversing & Adding Transactions
```

src/docs/product/sentry-basics/tracing/event-detail.mdx
imatwawana: I've added my heading suggestion on line 71

src/docs/product/sentry-basics/tracing/event-detail.mdx
imatwawana, line 71:
```suggestion
#### Traversing & Adding Transactions
```

dashed (approved):

PR #3884, opened 2021-07-15
feat: Add alert for Identifier field

leedongwei:
Users will encounter `Error AADSTS650056 - Misconfigured application` if they didn't include a trailing slash in the Identifier field when setting up Sentry on Azure AD. It causes some back-and-forth between them and support, which isn't ideal.

This adds an explicit alert for it so users can be warned of this subtle error.

Screen Shot 2021-07-15 at 11 51 24 AM

### DIFF
--- a/src/docs/product/accounts/sso/azure-sso.mdx
+++ b/src/docs/product/accounts/sso/azure-sso.mdx
@@ -28,7 +28,7 @@ description: "Set up Azure Active Directory Single Sign-On (SSO) on Sentry."
 
 1. For Section (1), labeled "Basic SAML Configuration", enter the following data in each line and save your changes.
 
-   - Identifier (Entity ID): `https://sentry.io/saml/metadata/YOUR_ORG_SLUG/` <Alert level="warning">Ensure that the URL includes a trailing slash to avoid causing a misconfiguration error (AADSTS650056).</Alert>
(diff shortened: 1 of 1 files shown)


src/docs/product/accounts/sso/azure-sso.mdx
```diff
@@ -28,7 +28,7 @@ description: "Set up Azure Active Directory Single Sign-On (SSO) on Sentry."
 
 1. For Section (1), labeled "Basic SAML Configuration", enter the following data in each line and save your changes.
 
-   - Identifier (Entity ID): `https://sentry.io/saml/metadata/YOUR_ORG_SLUG/`
+   - Identifier (Entity ID): `https://sentry.io/saml/metadata/YOUR_ORG_SLUG/` <Alert level="warning">Ensure that the URL includes a trailing slash to prevent Azure from throwing a misconfiguration error (AADSTS650056).</Alert>
```
markstory, line 31: Are we able to update our side of the integration to always add the slash?

imatwawana (approved):

src/docs/product/accounts/sso/azure-sso.mdx
leedongwei: Unfortunately not. This config lives on Azure as a ID and our app has to send a string that matches against it. 

We could try the validation twice (once with the slash, once without) but IMO this is an easier/faster fix to take some load off the support team.

PR #3894, opened 2021-07-16
Unity: Config window updated

bitsandfoxes:
The config window got a facelift and we no longer store the config as json.


imatwawana (approved):
LGTM!

PR #3909, opened 2021-07-20
fix: Use correct clubhouse image links

AbhiPrasad:
See: https://docs.sentry.io/product/integrations/integration-platform/ui-components/#issue-link for image not loading.

image

Can see that this is fixed in https://sentry-docs-git-abhi-fix-clubhouse-links.sentry.dev/product/integrations/integration-platform/ui-components/#issue-link


imatwawana (approved):
Thanks for the catch!!

PR #3922, opened 2021-07-22
fix(csp) Improve CSP docs around policy requirements

markstory:
Include a note that sentry needs to be in the allowed host list or submitting a CSP report will also fail.

Fixes getsentry/sentry#27524

### DIFF
--- a/src/docs/product/security-policy-reporting.mdx
+++ b/src/docs/product/security-policy-reporting.mdx
@@ -28,6 +28,8 @@ Alternatively you can setup CSP reports to simply send reports rather than actua
 Content-Security-Policy-Report-Only: ...; report-uri https://___ORG_INGEST_DOMAIN___/api/___PROJECT_ID___/security/?sentry_key=___PUBLIC_KEY___
 ```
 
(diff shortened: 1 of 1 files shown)


src/docs/product/security-policy-reporting.mdx
```diff
@@ -28,6 +28,8 @@ Alternatively you can setup CSP reports to simply send reports rather than actua
 Content-Security-Policy-Report-Only: ...; report-uri https://___ORG_INGEST_DOMAIN___/api/___PROJECT_ID___/security/?sentry_key=___PUBLIC_KEY___
 ```
 
+When defining your policy it is important to ensure that __ORG_INGEST_DOMAIN__ is in your `default-src` or `connect-src` policy, or browsers will block requests that submit policy violations.
```
markstory, line 31: Is `__ORG_INGEST_DOMAIN__` replaced everywhere or only in code blocks?

imatwawana: Is this a question for the Docs team or just a reminder for yourself?

markstory: For the docs team. I'm not well versed in how the docs are compiled and was unsure if this would turn into something useful, or if I should be more explicit in the text and say something like `... ensure that sentry.io (or your self hosted sentry domain) is in ...`

PR #3923, opened 2021-07-22
add clarity to spike protection content 

imatwawana:
Adds information about spike protection being org-level and how it works.


PeloWriter (approved):
Thank you.

PR #3924, opened 2021-07-22
fix(scim): update troubleshooting blurb about import users

JoshFerge:
- Another round of feedback from Okta, they'd like us specifically to call out not supporting the "Import Users" feature due to our missing firstName/lastName.

### DIFF
--- a/src/docs/product/accounts/sso/okta-sso/okta-scim.mdx
+++ b/src/docs/product/accounts/sso/okta-sso/okta-scim.mdx
@@ -63,8 +63,10 @@ As a result of these changes, users who are assigned will be sent an invitation 
 You can use "Push Groups" to sync and assign groups in Okta; they will be reflected in Sentry teams.
 
 ### Known Issues / Troubleshooting
-- Sentry does not currently support setting any User attributes other than `userName` and `active`
-- The User `name` attribute will contain fields firstName and lastName with the values of `N/A`. Sentry's SCIM API does not currently support these fields but returns them for compatibility purposes.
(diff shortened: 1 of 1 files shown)


src/docs/product/accounts/sso/okta-sso/okta-scim.mdx
```diff
@@ -63,8 +63,10 @@ As a result of these changes, users who are assigned will be sent an invitation
 You can use "Push Groups" to sync and assign groups in Okta; they will be reflected in Sentry teams.
 
 ### Known Issues / Troubleshooting
-- Sentry does not currently support setting any User attributes other than `userName` and `active`
-- The User `name` attribute will contain fields firstName and lastName with the values of `N/A`. Sentry's SCIM API does not currently support these fields but returns them for compatibility purposes.
+- Sentry does not currently support setting any User attributes other than `userName` and `active`.
+- Sentry does not support the Import Users feature fully.
+   - This is because the User `name` attribute will contain fields firstName and lastName with the values of `N/A`.
+   - Sentry's SCIM API does not currently support these fields but returns them for compatibility purposes.
```
PeloWriter, line 69:
```suggestion
- Sentry does not currently support setting any User attributes other than `userName` and `active`.
- The Import Users feature is not currently supported. Sentry's SCIM API does not at this time support the User `name` attribute fields `firstName` and `lastName`. Instead, we return these with values of `N/A` for compatibility purposes.
```

PeloWriter (approved):

PR #3927, opened 2021-07-23
ref: Move Release Details to the root of Releases

matejminar:
## Before
Screenshot 2021-07-23 at 12 52 36

## After
Screenshot 2021-07-23 at 12 53 03

I also removed one unused legacy image.


PeloWriter (approved):

adhiraj:
This is fine for now but we need to update the structure of release docs

PR #3933, opened 2021-07-26
fix: Fix the Release and Dist content in the capacitor init for angular

jennmueng:
We fixed the incorrect release/dist example in the "Other Platforms" tab but not the Angular tab.


imatwawana (approved):
Apart from technical review, LGTM!

bruno-garcia (approved):

PR #3950, opened 2021-07-28
fix(elixir): add required hackney dependency to installation steps

arusahni:
As of version 8, [the Elixir plugin has an additional dependency](https://github.com/getsentry/sentry-elixir#installation).


lobsterkatie (approved):
LGTM - thanks!

PR #3959, opened 2021-07-30
feat(apple): Update Automatic Instrumentation

philipphofmann:
Updated Screenshots and added HTTP instrumentation and
slow and frozen frames.


src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -37,7 +37,24 @@ The SDK uses the process start time as the beginning of the app start and the [`
 
 ![App Start Transaction](app-start-transaction.png)
 
-You can opt out of UIViewController and App Start Instrumentation using options:
+
+## Slow and Frozen Frames
+
+This feature is available for iOS, tvOS, and Mac Catalyst.
+
+Unresponsive UI, animation hitches, or just jank annoy users and degrade the user experience. Two measurements to track this unwanted experience are slow and frozen frames. A phone or tablet typically renders with 60 frames per second (fps). The frame rate can also be higher, especially as 120 fps displays are becoming more popular. With 60 fps, every frame has 16.67 ms to render. If your app needs more than 16.67 ms for a frame, it is a slow frame. Frozen frames are UI frames that take longer than 700 ms. A smoothly running app should try to avoid both. That's why the SDK adds these two measurements for the transactions you capture. The detail view of the transaction displays the slow, frozen, and total frames to the right.
```
PeloWriter, line 45: "just jank" may not translate well, suggest either removing it or choosing another phrase. Otherwise, a few edits to better introduce what we're adding:
```suggestion
Unresponsive UI, animation hitches, or just jank annoy users and degrade the user experience. Two measurements to track this unwanted experience are slow frames and frozen frames. A smoothly running app should try to avoid both. That's why the SDK adds these two measurements for the transactions you capture.

- **Slow frames** are captured by the SDK when the app needs more than 16.67 ms for a frame. Typically, a phone or tablet renders with 60 frames per second (fps), though the frame rate can be higher, especially as 120 fps displays become popular. With 60 fps, every frame has 16.67 ms to render; slower than that, and the app has slow frames.
- **Frozen frames** are UI frames that take longer than 700 ms.  

The detail view of the transaction displays the slow, frozen, and total frames:
```

src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -37,7 +37,24 @@ The SDK uses the process start time as the beginning of the app start and the [`
...
+
+![Slow and Frozen Frames](slow-frozen-frames.png)
+
+## HTTP Instrumentation
+
+The SentrySDK adds spans for outgoing HTTP requests to ongoing transactions that are bound to the Scope.
+The SDK attaches auto-generated transactions automatically to the Scope if there is no other transaction bound.
+When starting a transaction, you can use bindToScope to indicate whether the SDK should bind the new transaction to the Scope or not.
+
+## How to opt out
```
PeloWriter, line 55:
```suggestion
## Opt Out
```

PeloWriter (approved):

philipphofmann:
Thanks, @PeloWriter, for the improvements 👏

PR #3963, opened 2021-08-01
Show ReactElement as fallback in ErrorBoundary example

jcomo:
Docs fix as a follow-up from https://github.com/getsentry/sentry-javascript/pull/3857


AbhiPrasad (approved):
Thanks!

PR #3964, opened 2021-08-02
tunnel: Update proxy examples

### LINKED ISSUE #3945: SSI vulnerability at example in JS troubleshooting doc for tunnel
### Core or SDK?

Platform/SDK

### Which part? Which one?

javascript/troubleshooting

### Description

Example on the page https://docs.sentry.io/platforms/javascript/guides/react/troubleshooting/#using-the-tunnel-option (screenshot) contains ssi vulnerability. We can manipulate with dsn header and pass any strings to it. So, attacker can scan domains at internal application network and send requests to it.

image

### Suggested Solution

As fix, for example, you may specify sentry domain at internal network and replace string:
```
"https://{$dsn['host']}/api/$project_id/envelope/",
```
to
```
"https://sentry-host.company/api/123456/envelope/",
```

rhcarvalho:
Update examples to stop reading the target host from request body.
Instead, use explicit host configuration.

Similarly, limit project ID to a set of known IDs.

Fixes #3945.

### DIFF
--- a/src/platforms/javascript/common/troubleshooting/index.mdx
+++ b/src/platforms/javascript/common/troubleshooting/index.mdx
@@ -95,23 +95,31 @@ Once configured, all events will be sent to the `/tunnel` endpoint. This solutio
 
 ```php
 <?php
+// Change $host appropriately if you run your own Sentry instance.
+$host = "sentry.io";
+// Set $known_project_ids to an array with your Sentry project IDs which you
+// want to accept through this proxy.
+$known_project_ids = array(  );
+
 $envelope = stream_get_contents(STDIN);
 $pieces = explode("\n", $envelope, 2);
 $header = json_decode($pieces[0], true);
 if (isset($header["dsn"])) {
     $dsn = parse_url($header["dsn"]);
     $project_id = intval(trim($dsn["path"], "/"));
-    $options = array(
+    if (in_array($project_id, $known_project_ids)) {
+      $options = array(
         'http' => array(
-            'header'  => "Content-type: application/octet-stream\r\n",
+            'header'  => "Content-type: application/x-sentry-envelope\r\n",
             'method'  => 'POST',
             'content' => $envelope
         )
-    );
-    echo file_get_contents(
-        "https://{$dsn['host']}/api/$project_id/envelope/",
-        false,
-        stream_context_create($options));
+      );
+      echo file_get_contents(
+          "https://$host/api/$project_id/envelope/",
+          false,
+          stream_context_create($options));
+    }
 }
 ```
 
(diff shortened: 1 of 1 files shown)


rhcarvalho:
I wanted to add a note like 

> Note that the example above is merely a quick way to get you started.

But not sure how to continue for now ;)

Running this in production could benefit from more measures, e.g. some form of auditing, error monitoring, etc.

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -95,23 +95,31 @@ Once configured, all events will be sent to the `/tunnel` endpoint. This solutio
...
 $header = json_decode($pieces[0], true);
 if (isset($header["dsn"])) {
     $dsn = parse_url($header["dsn"]);
     $project_id = intval(trim($dsn["path"], "/"));
-    $options = array(
+    if (in_array($project_id, $known_project_ids)) {
+      $options = array(
         'http' => array(
-            'header'  => "Content-type: application/octet-stream\r\n",
+            'header'  => "Content-type: application/x-sentry-envelope\r\n",
```
rhcarvalho, line 113: Updated based on https://develop.sentry.dev/sdk/envelopes/#http-headers

I don't know why we had `application/octet-stream` here before. We could either ignore this completely for simplicity (as Relay doesn't enforce it), or use `application/x-sentry-envelope`.

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -127,6 +135,12 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 
+// Change host appropriately if you run your own Sentry instance.
+const host = "sentry.io";
```
rhcarvalho, line 139: @bruno-garcia I have no idea if my C# changes actually work, did my best effort looking up docs. Could you please check it?

Thanks :)

C# rocks!

rhcarvalho: Hmm we might as well align with https://github.com/getsentry/examples/blob/master/tunneling/dotnet/Program.cs

If the example here and in the `getsentry/examples` repo is supposed to be exactly the same, could we simply link to the example instead of copying the code here?

HazAT (approved):

src/platforms/javascript/common/troubleshooting/index.mdx
AbhiPrasad: +1 on linking to the `examples` repo.

rhcarvalho:
Merging as is for now to address the original issue, we can iterate on improvements later.

src/platforms/javascript/common/troubleshooting/index.mdx
bruno-garcia: I thought this was just a quick copy paste for users but in the examples repo we can go more in-depth. In .NET we actually ship a package that includes this and all it takes to add os `UseSentryTunnel` to you request pipeline.

Using it on nuget trends:

https://github.com/dotnet/nuget-trends/blob/9a8fdc836b6e0f01dc44ced87495531aa89d066d/src/NuGetTrends.Web/Startup.cs#L100-L101

I'll fix the sample here.

Should we remove all samples here and only link to GitHub where we maintain this code?

PR #3967, opened 2021-08-02
Remove unneccessary `{ }` from the fallback prop

jonahallibone:
This is going to throw an error in most people's linters, including the default rules in create-react-app and create-next-app


AbhiPrasad:
Thanks for your contribution! We updated this file in https://github.com/getsentry/sentry-docs/pull/3963, so these changes should no longer be necessary.

PR #3988, opened 2021-08-05
doc(trace-view): Missing trace can be caused by sample rate

Zylphrex:


src/docs/product/sentry-basics/tracing/trace-view.mdx
```diff
@@ -33,6 +33,7 @@ In a fully instrumented system, a span in one service will connect to a transact
 In addition, broken subtraces can occur when Sentry receives an error for a trace, but the corresponding transaction does not exist. In this case, the error will not appear on the trace view and the trace navigator won't appear on the **Issue Details** page. In these cases you can click "Open In Discover" to see all the events in the trace.
 
 Broken subtraces may be caused by:
+- Sample rates may cause a transaction to be dropped by the SDK, so not all events will be sent to Sentry.
```
wmak, line 36:
```suggestion
- Sample rates may cause a transaction to not be sent by the SDK, so not all errors will have an associated transaction sent to Sentry.
```

imatwawana: Tightening up a bit. If "in Sentry" changes the meaning, then stick with what Will suggested.

```suggestion
- Sample rates may prevent the SDK from sending a transaction, so not all errors will have an associated transaction in Sentry.
```

imatwawana (approved):

PR #4004, opened 2021-08-11
Update default log level typo

marandaneto:
not sure if `warning` is an alias for `warn` but seeing sentry-clis code, its wrong: https://github.com/getsentry/sentry-cli/blob/e51e11302567dba927abc7f8383cf2d5b22c517b/src/commands/mod.rs#L231


marandaneto:
@jan-auer @kamilogorek please close it if its an alias and it'll work anyway

kamilogorek:
It's not, thanks!

PR #4005, opened 2021-08-12
feat (hackweek): automated screenshots

cameronmcefee:
This PR adds the infrastructure needed to save scenario files which will automatically generate screenshot based on the sequence of commands it's given. The end result is that we can trigger an automated update to all screenshots on the site as needed, be it manually, via a cron, or whenever the core app updates.

This does not contain any of the screenshot scenarios itself.

```
# Record a screenshot scenario and save a .screenshot.js file
yarn configure-screenshot
```

```
# Produce screenshots from all the .screenshot.js files
yarn capture-screenshots
```

### DIFF
--- a/src/lib/capture-screenshots.js
+++ b/src/lib/capture-screenshots.js
@@ -10,9 +10,9 @@ const captureScreenshots = async () => {
   q.concurrency = 5;
 
   screenshotConfigs.forEach(filepath => {
-    q.push(async () => {
+    q.push(() => {
       console.log(`Capturing ${filepath}`);
-      await require(path.join(process.cwd(), filepath));
+      return require(path.join(process.cwd(), filepath));
     });
   });
 
--- a/src/lib/capture-screenshots.js
+++ b/src/lib/capture-screenshots.js
@@ -1,0 +1,27 @@
+const glob = require("glob-promise");
+const queue = require("queue");
+const path = require("path");
+
+const captureScreenshots = async () => {
+  console.log("Finding screenshot configs…");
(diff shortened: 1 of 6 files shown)


src/lib/capture-screenshots.js
```diff
@@ -0,0 +1,27 @@
...
+const path = require("path");
+
+const captureScreenshots = async () => {
+  console.log("Finding screenshot configs…");
+  const screenshotConfigs = await glob("**/*.screenshot.js");
+  console.log(`Found ${screenshotConfigs.length} screenshot configs.`);
+  const q = queue({ results: [] });
+  q.concurrency = 5;
+
+  screenshotConfigs.forEach(filepath => {
```
BYK, line 12: I wonder if we can run all these in parallel with something like https://yarnpkg.com/package/jest-worker

BYK: Oh wait, `queue` does this already nvm.

src/lib/capture-screenshots.js
```diff
@@ -0,0 +1,27 @@
...
+    q.push(async () => {
+      console.log(`Capturing ${filepath}`);
+      await require(path.join(process.cwd(), filepath));
+    });
```
BYK, line 16:
```suggestion
    q.push(() => {
      console.log(`Capturing ${filepath}`);
      return require(path.join(process.cwd(), filepath));
    });
```

(saves a cycle)

dayanmasiha (approved):

PR #4015, opened 2021-08-17
feat: Update examples to cover `pg-native` integration usage.

onurtemizkan:
Ref: https://github.com/getsentry/sentry-javascript/pull/3894


AbhiPrasad (changes requested):
We need to make sure we update the `Supported packages and their integration name:` section as well. Maybe add a note that it is new with version XXX (not sure what will be until we cut the sdk release)

onurtemizkan:
@AbhiPrasad, I added a note to mention the released version of this feature.

AbhiPrasad (approved):

PR #4029, opened 2021-08-18
fix: Display body parameter information in API docs

leeandher:
See: [API-1671](https://getsentry.atlassian.net/browse/API-1671)

Request body parameters for the in the API provide descriptions for how to use them that should render in markdown but were not appearing. To remedy this, I added a type that recognized body parameters as nodes to 'markdownify' and modified the static query on API Pages. I referred to the node `openApiPathParameter` when creating `openApiBodyParameter`, which is why there may be some redundant fields on the node.

_Note to reviewers:_ It also looks like these files hadn't been touched in a while, and my linter auto-corrected some spaces, semi-colons, etc. I double checked, and it was linting according to the tsconfig file, if that changes anything 👍 

**Before**
image
https://docs.sentry.io/api/integration/create-an-external-issue/

image
https://docs.sentry.io/api/events/update-an-issue/

**After**
image

image
_Note the rendered markdown in the above example 👀_

### DIFF
--- a/src/gatsby/plugins/gatsby-plugin-openapi/gatsby-node.ts
+++ b/src/gatsby/plugins/gatsby-plugin-openapi/gatsby-node.ts
@@ -4,9 +4,8 @@ import {
   DeRefedOpenAPI,
   OpenApiPath,
   RequestBody,
+  RequestBodySchema,
 } from "./types";
-
-import { RequestBodySchema } from "~src/gatsby/plugins/gatsby-plugin-openapi/types";
 
 export const sourceNodes = async (
   { actions, createNodeId, createContentDigest },
@@ -188,10 +187,12 @@ export const onCreateNode = async ({
       });
     });
 
+    // Optional chaining seems to affect the Vercel build process
     const bodyParameterSchemaString =
       node.path.requestBody !== null &&
       node.path.requestBody.content !== null &&
       node.path.requestBody.content.schema;
+
     if (bodyParameterSchemaString) {
       const bodyParameterSchema: RequestBodySchema = JSON.parse(
         bodyParameterSchemaString
--- a/src/gatsby/plugins/gatsby-plugin-openapi/gatsby-node.ts
+++ b/src/gatsby/plugins/gatsby-plugin-openapi/gatsby-node.ts
@@ -188,13 +188,18 @@ export const onCreateNode = async ({
       });
     });
 
-    const bodyParameterSchemaString = node.path?.requestBody?.content?.schema;
+    const bodyParameterSchemaString =
+      node.path.requestBody !== null &&
+      node.path.requestBody.content !== null &&
+      node.path.requestBody.content.schema;
     if (bodyParameterSchemaString) {
       const bodyParameterSchema: RequestBodySchema = JSON.parse(
         bodyParameterSchemaString
       );
-      const bodyParameterRequired = { ...bodyParameterSchema?.required };
-      Object.entries(bodyParameterSchema?.properties).map(
+      const bodyParameterRequired = {
+        ...(bodyParameterSchema.required || []),
+      };
+      Object.entries(bodyParameterSchema.properties || []).map(
         ([name, { type, description }], index) => {
           if (description) {
             const bodyParamNode = {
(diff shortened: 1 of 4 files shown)


src/gatsby/plugins/gatsby-plugin-openapi/gatsby-node.ts
```diff
@@ -6,6 +6,8 @@ import {
   RequestBody,
 } from "./types";
 
+import { RequestBodySchema } from "~src/gatsby/plugins/gatsby-plugin-openapi/types";
```
NisanthanNanthakumar, line 9: @leeandher we should merge imports of types.

src/gatsby/plugins/gatsby-plugin-openapi/gatsby-node.ts
```diff
@@ -184,6 +186,46 @@ export const onCreateNode = async ({
         parent: node,
         child: paramNode,
       });
-    })
+    });
+
+    const bodyParameterSchemaString =
+      node.path.requestBody !== null &&
+      node.path.requestBody.content !== null &&
+      node.path.requestBody.content.schema;
```
NisanthanNanthakumar, line 194: I think we can clean this up with optional chaining.

src/templates/apiPage.tsx
```diff
@@ -102,7 +97,7 @@ export default props => {
   useEffect(() => {
     Prism.highlightAll();
   }, []);
-
+  console.log(bodyParameters);
```
NisanthanNanthakumar, line 100: @leeandher we should remove this console.log

src/templates/apiPage.tsx
```diff
@@ -77,7 +72,7 @@ export default props => {
     apiExample.push(` -H 'Content-Type: ${contentType}'`);
   }
 
-  if (bodyParameters) {
+  if (bodyParameters.length > 0) {
```
NisanthanNanthakumar, line 75: Nit: we can remove the `>0` since `0` is falsy in javascript

src/gatsby/plugins/gatsby-plugin-openapi/gatsby-node.ts
leeandher: I agree, but I actually reverted this from optional chaining specifically because it breaks something it the Gatsby Build process 👀 that only applies to this file.

src/templates/apiPage.tsx
leeandher: oops, thanks 👍

src/templates/apiPage.tsx
leeandher: Good catch!

src/gatsby/plugins/gatsby-plugin-openapi/gatsby-node.ts
leeandher: whoops, will fix 👍  #justautoimportthings

src/gatsby/plugins/gatsby-plugin-openapi/gatsby-node.ts
leeandher: Here's what happened to the Vercel preview when I used optional chaining 
image

NisanthanNanthakumar: @leeandher I see it used here: https://github.com/getsentry/sentry-docs/blob/master/src/components/codeContext.tsx#L130

leeandher: @NisanthanNanthakumar Yeah, it seems to only break in the `src/gatsby/plugins/gatsby-plugin-openapi/gatsby-node.ts` file, because Gatsby processes those as is, without webpack. Outside of that it seems to work just fine.

See: https://github.com/gatsbyjs/gatsby/issues/23802

leeandher:
Had some trouble with rebasing, sorry for the pings on those reviewers 😞

src/gatsby/plugins/gatsby-plugin-openapi/gatsby-node.ts
NisanthanNanthakumar: hmm the values could also be `undefined`.

```suggestion
      node.path.requestBody &&
      node.path.requestBody.content &&
      node.path.requestBody.content.schema;
```

leeandher: Added a falsy check instead, thanks :)

PR #4033, opened 2021-08-19
fix release sourcemap param typo

thakkaryash94:
As per the CLI --help command, it should be `upload-sourcemaps`. There is an extra space between 'source' and 'maps'. Based on https://docs.sentry.io/product/cli/releases/ page.


bruno-garcia (approved):
Thanks!

PR #4034, opened 2021-08-19
Document propagating scope in Spring async methods.

maciejwalkowiak:


maciejwalkowiak:
cc @marandaneto @bruno-garcia

src/platforms/java/guides/spring-boot/async.mdx
```diff
@@ -0,0 +1,138 @@
+---
+title: Async Methods
```
marandaneto, line 2: is it exactly the same as src/platforms/java/guides/spring/async.mdx ?
What's about using the `includes` folder or linking to the other page? duplicating everything is always a bit hard to maintain I guess, not sure how much we'd have to maintain this

marandaneto:
adding @imatwawana for wording review, the technical side looks good.

src/platforms/java/guides/spring-boot/async.mdx
imatwawana: Given that these are so similar, yes, I think it makes sense to make this a piece of common content for Java (so it would live in the folder **src > platforms > Java > common**. Then add includes for the parts that differ between the two. And you can add a "not supported" component in the front matter so that it doesn't show up for the other Java frameworks that don't use this.

imatwawana (approved):
I've made some minor wording updates, but agree that this be restructured into common content for Java + includes per @marandaneto's suggestion

maciejwalkowiak:
Thanks @marandaneto @imatwawana for quick review and suggestions. I did not know pages can be shared like that between different guides.

Please take another look.

bruno-garcia (approved):
:heart:

PR #4044, opened 2021-08-22
Deprecate sentry-ruby's async config option

st0012:
This option is planned to be removed in the future and I want to discourage users from using it.

### DIFF
--- a/src/platforms/ruby/common/configuration/options.mdx
+++ b/src/platforms/ruby/common/configuration/options.mdx
@@ -20,35 +20,11 @@ end
 
 `async`
 
-: When an error or message occurs, the notification is immediately sent to Sentry, synchronously. This means that returning a response to a user may be delayed.
+**This option has been deprecated and could be removed in the future.**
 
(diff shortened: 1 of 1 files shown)


src/platforms/ruby/common/configuration/options.mdx
```diff
@@ -20,35 +20,11 @@ end
...
-Or if you also use `sentry-rails`, you can directly use the job we defined for you:
-
-```ruby
-config.async = lambda do |event, hint|
-  Sentry::SendEventJob.perform_later(event, hint)
-end
-```
-
-If the async callback raises an exception, Sentry will attempt to send synchronously.
+: By default, `sentry-ruby` sends events asynchronouly with its own background worker. This option only exists for compatibility reason and it's not recommended to new Sentry users.
```
imatwawana, line 27:
```suggestion
: By default, `sentry-ruby` sends events asynchronouly with its own background worker. This option only exists for compatibility reasons, and it's not recommended to new Sentry users.
```

PR #4052, opened 2021-08-24
fix(grouping): Remove useless byline about being consistent with discover

untitaker:
It does not help the user and implies more than it needs to (as not all
attributes are available). It's like saying "this button has style
consistent with this other button someplace else in the product". It
would belong in internal design docs.


jjbayer (approved):
No objections, but it is a bad thing to link to related concepts?

untitaker:
@jjbayer we have one customer who was confused that it didn't literally work like discover

PR #4080, opened 2021-08-30
feat(api-schema): Always build from master of api-schema repo

BYK:
We are removing some security tokent from sentry-api-schema, preventing it from making commits directly to sentry-docs repo. For this reason, we are removing this variable and defaulting to `main` branch. Deploy triggers on new versions of sentry-api-schema will be done through [Vercel deploy hooks](https://vercel.com/docs/more/deploy-hooks) tied to sentry-docs repo.

### DIFF
--- a/src/gatsby/utils/resolveOpenAPI.ts
+++ b/src/gatsby/utils/resolveOpenAPI.ts
@@ -1,10 +1,6 @@
 import axios from "axios";
 import { promises as fs } from "fs";
 
-// SENTRY_API_SCHEMA_SHA is used in the sentry-docs GHA workflow in getsentry/sentry-api-schema.
-// DO NOT change variable name unless you change it in the sentry-docs GHA workflow in getsentry/sentry-api-schema.
-const SENTRY_API_SCHEMA_SHA = "4cf8276a19f412a316f9bc723a6081a1acc98ba1"
-
 const activeEnv =
   process.env.GATSBY_ENV || process.env.NODE_ENV || "development";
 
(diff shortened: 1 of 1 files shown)


untitaker:
I think it was done this way to allow for reproducible builds. Is there any chance we can retain that?

BYK:
>I think it was done this way to allow for reproducible builds. Is there any chance we can retain that?

I don't think we can as we cannot update the files without bypassing the branch protections easily, which requires admin access on the repo.

untitaker (approved):
That's very unfortunate, I don't think this solution will work for setnry-data-schemas though. So if we manage to find a soln there that works here I'd prefer that

NisanthanNanthakumar (approved):

manuzope:
Do we need to update any instructions here https://develop.sentry.dev/api/public/#how-to-make-an-endpoint-public?

BYK:
@untitaker how do you feel about actually triggering a new "release" of docs with every update to `sentry-api-schema` repo? That way we can pull the latest `main` branch SHA into sentry-docs as part of the release process?

BYK:
@manuzope from a quick glance, it doesn't seem necessary but the process seems to be a bit involved for me to be entirely sure.

untitaker:
>how do you feel about actually triggering a new "release" of docs with every update to sentry-api-schema repo? That way we can pull the latest main branch SHA into sentry-docs as part of the release process?

I think that's alright but wouldn't that release need manual approval?

untitaker:
Also I suppose it would become tricky if you have multiple datasources feeding into docs (which we have with data-schemas + api-schema), becuase it would mean two release naming conventions. Which is maybe ok but the releases don't make any sense to me

PR #4105, opened 2021-09-06
Update Next.js `sentry-cli` configuration

iker-barriocanal:
All the properties in a Next.js project used to be in `sentry.properties`, used by Sentry CLI. The auth token is the only property that shouldn't be committed to the VCS, so it has been moved to `.sentryclirc`. This allows committing `sentry.properties` and ignoring `.sentryclirc` to safely commit everything but the auth token.

### DIFF
--- a/src/platforms/javascript/guides/nextjs/manual-setup.mdx
+++ b/src/platforms/javascript/guides/nextjs/manual-setup.mdx
@@ -153,7 +153,7 @@ The URL, organization, and project properties identify your organization and pro
 
 ### Use Configuration Files
 
-You should commit all the properties to your VCS, but the auth token. You can accomplish this by using two files: `sentry.properties` including the properties of your organization and project, and `.sentryclirc` including your auth token. This is the approach taken by the wizard and allows you to commit the former while ignoring the latter in your VCS.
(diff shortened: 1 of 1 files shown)


src/platforms/javascript/guides/nextjs/manual-setup.mdx
```diff
@@ -145,22 +145,33 @@ In that case you can also skip the `sentry-cli` configuration step below.
...
+If you choose to combine the two approaches, the environment variables will take precedence over values set in the configuration files. One common approach is to set sensitive data (like tokens) in the environment and include everything else in the configuration files added to your VCS.
 
-### Use a Properties File
+The URL, organization, and project properties identify your organization and project, and the auth token authenticates your user account.
 
-You can create a `sentry.properties` file in the base directory of your project. (This is the approach taken by the wizard.) Here is an example:
+### Use Configuration Files
+
+You should commit all the properties to your VCS, but the auth token. You can accomplish this by using two files: `sentry.properties` including the properties of your organization and project, and `.sentryclirc` including your auth token. This is the approach taken by the wizard and allows you to commit the former while ignoring the latter in your VCS.
```
imatwawana, line 156:
```suggestion
You should commit all the properties to your VCS, except the auth token. You can accomplish this by using two files: `sentry.properties` including the properties of your organization and project, and `.sentryclirc` including your auth token. This is the approach taken by the wizard and it allows you to commit the former while ignoring the latter in your VCS.
```

imatwawana (approved):

PR #4124, opened 2021-09-09
Add a not (typo) to lazy loading page

hdimitrov1:
Unless I'm misreading the sentence badly, i think this needs to have a not. I could well be wrong


imatwawana:
Thanks for the catch!

PR #4131, opened 2021-09-09
ref(codeowners): Update screenshots, remove EA mention

ceorourke:
Remove the EA note for codeowners and update the screenshots on this page https://docs.sentry.io/product/issues/issue-owners/ to reflect the current UI. I also updated some of the non-codeowners screenshots as they were strangely stretched out and have slightly changed as well. Finally, I noticed a line about only sending issue owners notifications via email and updated that as the notification platform has launched and we can potentially notify issue owners via Slack if that is what their setting is.

Peep the changes here: https://sentry-docs-git-ceorourke-api-2101.sentry.dev/product/issues/issue-owners/


ceorourke:
Is it safe to have the screenshot of our codeowners rules? I figured yes since it's open source anyway but I wanna double check.

manuzope:
Yea should be fine. I think we need to remove the mention of EA under `External Team/User Mappings`

manuzope (approved):

PR #4132, opened 2021-09-09
dotnet(nlog): Update NLog specific documentation

### LINKED ISSUE #4114: Inaccuracy in documentation for NLog "Identify Users".ip_address
### Core or SDK?

Platform/SDK

### Which part? Which one?

https://docs.sentry.io/platforms/dotnet/guides/nlog/enriching-events/identify-user/

### Description

https://docs.sentry.io/platforms/dotnet/guides/nlog/enriching-events/identify-user/

On this page it's described that to pass IP address of the client it's needed to specify XML attribute ip_address on user XML node. However noticed that after applying such application starts throwing the following exception:
"Exception when parsing [PATH_TO_APPLICATION]\NLog.config."
This basically means that XML scheme is broken and config couldn't be parsed. Then I experimentally discovered that the correct attribute is ipaddress -- everything works fine if pass IP address in this XML attribute from user XML node. Could you please consolidate this situation.

Also when I skipped this XML attribute on user XML node NLog set "the IP address from the connection between your app and Sentry's server" however I expected that "Sentry will attempt to pull this from the HTTP request data, if available" as "sendDefaultPii set to true". Could you please advice?

Please check the same items for other logging frameworks (and other technologies than .NET) -- probably some other logging frameworks also suffer of such stuff.

### Suggested Solution

https://docs.sentry.io/platforms/dotnet/guides/nlog/enriching-events/identify-user/

Please consolidate documentation as this creates inconveniences in development process.
Advice on inconsistency with sending IP address for User and if there any bug please fix it.
Please check the same items for other logging frameworks (and other technologies than .NET).

josh-degraw:
Updates documentation related to NLog.

cc: @bruno-garcia 

Closes #4114

### DIFF
--- a/src/includes/enriching-events/set-user/dotnet.nlog.mdx
+++ b/src/includes/enriching-events/set-user/dotnet.nlog.mdx
@@ -1,0 +1,50 @@
+<Alert level="" title="NLog Configuration">
+
+- If setting it via `NLog.config`, the camelCase key `ipAddress` should be used instead of `ip_address`.
+- For more information on how to dynamically set event data via `NLog.config`, see NLog's [layout renderer documentation](https://nlog-project.org/config/?tab=layout-renderers).
+
+</Alert>
+
+```xml {tabTitle:NLog.config}
+<?xml version="1.0" encoding="utf-8" ?>
+<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+>
+  <extensions>
+    <add assembly="Sentry.NLog" />
+  </extensions>
+  <targets>
+    <target xsi:type="Sentry" name="sentry" dsn="___PUBLIC_DSN___">
+      <user email="john.doe@example.com" />
+    </target>
+  </targets>
+
+  <rules>
+    <logger name="*" writeTo="sentry" />
+  </rules>
+</nlog>
+```
+
+```csharp
+using Sentry;
+
+SentrySdk.ConfigureScope(scope =>
+{
+    scope.User = new User
+    {
+        Email = "john.doe@example.com"
+    };
+});
+```
+
+```fsharp
+open Sentry
+
+SentrySdk.ConfigureScope(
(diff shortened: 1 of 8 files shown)


src/includes/enriching-events/set-user/dotnet.nlog.mdx
```diff
@@ -0,0 +1,50 @@
+<Alert level="" title="NLog Configuration">
```
bruno-garcia, line 1: I noticed we're currently getting: 
image

It's because we're missing snippets here: https://github.com/getsentry/sentry-docs/tree/master/src/includes/enriching-events/unset-user

For .NET it would be just setting it to `null`. Would be nice as a follow up PR 🙏

src/includes/set-release/dotnet.nlog.mdx
```diff
@@ -0,0 +1,50 @@
...
+options.Release <- "my-project-name@2.3.12"
+```
+
+The SDK attempts to locate the release to add to events sent to Sentry.
+
+The SDK will first look at the [entry assembly's](<https://msdn.microsoft.com/en-us/library/system.reflection.assembly.getentryassembly(v=vs.110).aspx>) `AssemblyInformationalVersionAttribute`, which accepts a string as
+value and is often used to set a GIT commit hash.
+
+If that returns null, it'll look at the default `AssemblyVersionAttribute` which accepts the numeric version number. When creating a project with Visual Studio, the value is set to _1.0.0.0_.
+Since that usually means that the version is either not being set, or is set via a different method. The **automatic version detection will disregard** this value and no _Release_ will be reported automatically.
```
bruno-garcia, line 50: We dropped this thing so now it'll report 1.0.0 too so at least something shows up in Sentry (better than the alternative that nothing showed up)

```suggestion
If that returns null, it'll look at the default `AssemblyVersionAttribute` which accepts the numeric version number.
```

Could you please fix it in the main .NET page too?

bruno-garcia (approved):
This is really great! Thanks @josh-degraw 

I left two follow up items that could be done on another PR (not blocking this)

PR #4141, opened 2021-09-13
Added note about required roles to view SOC2 report.

alek-sentry:


alek-sentry:
Wrong link type used.

PR #4143, opened 2021-09-13
sdk(dotnet): diagnostic logger pointer

bruno-garcia:
Users seem confused when they turn debug on and "don't see anything" since they are building WinForms, WPF and other types of apps.

Also adds DiagnosticLogger doc to guides and Unity.
Fixes Attachments not being visible to Unity


src/platforms/common/configuration/options.mdx
```diff
@@ -59,6 +59,12 @@ Enabling `debug` mode makes the SDK generate as much diagnostic data as possible
 
 </ConfigKey>
 
+<ConfigKey name="diagnostic-level" supported={["dotnet"]}>
+
+For app models that don't have a console to print to, such as WinForms and ASP.NET, you can <PlatformLink to="/configuration/diagnostic-logger/">customize the SDK diagnostic logger</PlatformLink> to write to a file or to Visual Studio's debug window.
+
+</ConfigKey>
```
bruno-garcia, line 66:
```suggestion
<PlatformSection supported={["dotnet"]}>

For app models that don't have a console to print to, such as WinForms and ASP.NET, you can <PlatformLink to="/configuration/diagnostic-logger/">customize the SDK's diagnostic logger</PlatformLink> to write to a file or to Visual Studio's debug window.

</ConfigKey>
```

src/platforms/common/configuration/options.mdx
```diff
@@ -59,6 +59,12 @@ Enabling `debug` mode makes the SDK generate as much diagnostic data as possible
 
 </ConfigKey>
 
+<PlatformSection supported={["dotnet"]}>
+
+For app models that don't have a console to print to, such as WinForms and ASP.NET, you can <PlatformLink to="/configuration/diagnostic-logger/">customize the SDK's diagnostic logger</PlatformLink> to write to a file or to Visual Studio's debug window.
```
imatwawana, line 64: Since the content isn't there for these two frameworks, I'm not sure if we should mention them. 

```suggestion
For app models that don't have a console to print to, you can <PlatformLink to="/configuration/diagnostic-logger/">customize the SDK's diagnostic logger</PlatformLink> to write to a file or to Visual Studio's debug window.
```

imatwawana (approved):
It looks like none of the the .NET guides or Unity have debug logger content, so I think you should probably add a `notSupported` property in the platform section tag to include these guides.  It seems like there's a gap here that's outside the scope of this PR.

lucas-zimerman:
Might be a good idea to mention TraceDiagnosticLogger on the docs?

bruno-garcia:
> It looks like none of the the .NET guides or Unity have debug logger content, so I think you should probably add a `notSupported` property in the platform section tag to include these guides. It seems like there's a gap here that's outside the scope of this PR.

Is it possible to add to those? I wish this content would be 'inherited' by everything under .NET: guides and Unity.

That's true for other content such as Attachments: https://docs.sentry.io/platforms/dotnet/enriching-events/attachments/
Came up today that this isn't available for Unity, not sure how to make it so

imatwawana:
I think it can be, but it'll take some time for me to figure out as I've not refactored SDK docs before. It's up to you if you want to close this PR for now or put in the placeholder `notSupported` property until this can be changed.

PR #4157, opened 2021-09-15
update plan notes for sso features

imatwawana:
Add and remove notes pertaining to plan-based availability for SSO features

- Github and Google - Team+ and Trial
- SAML2 - Biz+, no Trial

### DIFF
--- a/src/docs/product/accounts/sso/index.mdx
+++ b/src/docs/product/accounts/sso/index.mdx
@@ -13,10 +13,6 @@ Single Sign-On (or SSO) allows you to manage your organization’s entire member
 
 Before you get around to actually turning on SSO, you’ll want to keep in mind that once it’s activated, all existing users will need to link their account before they are able to continue using Sentry. Because of that we recommend coordinating with your team during off-peak hours. That said, it’s super quick to link accounts, so we don’t consider it a true hurdle.
 
-<Alert title="Note" level="warning">
-  SSO is not available on free, trial or certain other plans.
-</Alert>
-
 ## Getting Started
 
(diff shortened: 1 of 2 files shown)


imatwawana:
@scefali - Decided to go with your suggestion of just talking about the lowest level plan + trial for Google and Github to keep the wording from getting clunky. But was more explicit for SAML2, as we discussed.

src/docs/product/accounts/sso/index.mdx
```diff
@@ -37,17 +33,37 @@ Sessions last for [Django's default session length](https://docs.djangoproject.c
 
 ### Google Business App
 
+<Note>
+
+This feature is available only if your organization is on a Team or Trial plan.
```
scefali, line 38: @imatwawana It's also available on a Business plan (Business Plan has everything a Team Plan does and more)

imatwawana:
> @imatwawana I also actually need to make a PR to enable SSO on trials for everyone, currently only 50% of trial orgs are getting SSO.

Do you want me to hold off on merging this until you've created that PR?

src/docs/product/accounts/sso/index.mdx
imatwawana: Right, Team, Biz, Ent, and Trial plans. But I thought that wording would sound clunky, so I was gonna stick with just saying Team, with the implication that we mean Team and higher. I don't think it's a good idea to just say Team + Biz and not mention Enterprise because we have an Enterprise plan on our pricing page, but then that gets too long. What about this?

```suggestion
This feature is available only if your organization is on a Team plan or higher, or a Trial plan.
```

scefali: @imatwawana Ok, I'll leave it up to you. Not sure what you do for other parts of the docs but we should be consistent with other featurees.

PR #4165, opened 2021-09-20
fix(scim): correct okta api token instructions

JoshFerge:
After okta pushed our integration to production, how the token was supplied slightly changed and you no longer have to provide "Bearer " in the token. We remove that part of the instruction and also remove screenshots, as they can become outdated and didn't provide much added benefit here in these instructions.

### DIFF
--- a/src/docs/product/accounts/sso/okta-sso/okta-scim.mdx
+++ b/src/docs/product/accounts/sso/okta-sso/okta-scim.mdx
@@ -36,14 +36,12 @@ Okta SCIM provisioning requires:
 1. Select the "Provisioning" tab, then "Configure API integration".
    ![Okta configure SCIM API](scim-okta-configure.png)
 
-1. Select "Enable API Integration", provide the SCIM URL from the auth settings page as the Base URL field. For the API Token, copy the Auth Token value from the auth settings page, including `Bearer` in the value you copy.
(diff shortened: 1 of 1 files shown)


src/docs/product/accounts/sso/okta-sso/okta-scim.mdx
```diff
@@ -36,14 +36,11 @@ Okta SCIM provisioning requires:
 1. Select the "Provisioning" tab, then "Configure API integration".
    ![Okta configure SCIM API](scim-okta-configure.png)
 
-1. Select "Enable API Integration", provide the SCIM URL from the auth settings page as the Base URL field. For the API Token, copy the Auth Token value from the auth settings page, including `Bearer` in the value you copy.
+1. Select "Enable API Integration", provide the SCIM URL from the auth settings page as the Base URL field. For the API Token, copy the Auth Token value from the auth settings page.
```
imatwawana, line 39:
```suggestion
1. Select "Enable API Integration", enter the SCIM URL from the auth settings page as the Base URL field. For the API Token, copy the Auth Token value from the auth settings page.
```

imatwawana, line 39:
```suggestion
1. Select "Enable API Integration", enter the SCIM URL from the auth settings page as the Base URL field. For the API Token, copy the Auth Token value from the auth settings page.
```

imatwawana (approved):
One minor edit. Otherwise LGTM!

PR #4171, opened 2021-09-21
[WIP] add google cloud functions docs and wizard for .NET.

lucas-zimerman:
Still in WIP.

This PR adds the appropriate documentation and wizard for the Google Cloud Functions SDK.

### DIFF
--- a/src/platforms/dotnet/guides/google-cloud-functions/config.yml
+++ b/src/platforms/dotnet/guides/google-cloud-functions/config.yml
@@ -1,0 +1,2 @@
+title: Google Cloud Functions
+sdk: sentry.google.cloud.functions
--- a/src/platforms/dotnet/guides/google-cloud-functions/index.mdx
+++ b/src/platforms/dotnet/guides/google-cloud-functions/index.mdx
@@ -1,0 +1,233 @@
+---
+title: Google Cloud Functions
+description: "Learn about Sentry's .NET integration with Google Cloud Functions."
+redirect_from:
+  - /platforms/dotnet/google.cloud.functions/
+---
+
+                           ___
+                            {-)   |\
+                       [m,].-"-.   /
+      [][__][__]         \(/\__/\)/
+      [__][__][__][__]~~~~  |  |
+      [][__][__][__][__][] /   |
+      [__][__][__][__][__]| /| |
+      [][__][__][__][__][]| || |  ~~~~
+  WIP [__][__][__][__][__]__,__,  \__/
+
+
+Sentry provides an integration with `Google Cloud Functions` through the [Sentry.Google.Cloud.Functions NuGet package](https://www.nuget.org/packages/sentry.google.cloud.functions).
+
+## Overview of the features
+
+- Easy Google Cloud Functions integration, single line: `UseSentry`
+- Captures unhandled exceptions in the middleware pipeline
+- Captures exceptions handled by the framework `UseExceptionHandler` and Error page display
+- Captures process-wide unhandled exceptions (AppDomain)
+- Captures `LogError` or `LogCritical`
+- Any event sent will include relevant application log messages
+- RequestId as tag
+- URL as tag
+- Environment is automatically set (IHostingEnvironment)
+- Release automatically set (`AssemblyInformationalVersionAttribute`, `AssemblyVersion` or environment variable)
+- Request payload can be captured if opt-in
+- Support for EventProcessors registered with DI
+- Support for ExceptionProcessors registered with DI
+- Supports configuration system (bind to `SentryAspNetCoreOptions`)
+- Server OS info sent
+- Server Runtime info sent
+- Request headers sent
+- HTTP Proxy configuration
+- Request body compressed
+- Event flooding protection (429 retry-after and internal bound queue)
+- Assembly Strong named
+
+## Installation
+
+Add the Sentry dependency:
+
+```powershell {tabTitle:Package Manager}
+Install-Package Sentry.Google.Cloud.Functions -Version {{ packages.version('sentry.google.cloud.functions') }}
+```
+
+```shell {tabTitle:.NET Core CLI}
+dotnet add package Sentry.Google.Cloud.Functions -v {{ packages.version('sentry.google.cloud.functions') }}
+```
+
+This package extends `Sentry` main SDK. That means that besides the logging related features, through this package you'll also get access to all API and features available in the main `Sentry` SDK.
+
+<Alert level="warning" title="Note">
+
+Messages logged from assemblies with the name starting with `Sentry` will not generate events.
+
+</Alert>
+
+## Features
+
+- Store log messages as breadcrumbs
+- Send events to sentry
+
+Two separate settings define the minimum log level to keep the log entry as a `Breadcrumb` and to send an `Event` to Sentry. The events include any stored breadcrumb on that [scope](enriching-events/scopes/).
+
+By default, Sentry will keep any message with log level `Info` or higher as a `Breadcrumb`.
+
+The default value to report a log entry as an event to Sentry is `Error`.
+
+This means that out of the box, any `Error` call will create an `Event` which will include all log messages of level `Info`, `Warn` and also `Error` and `Critical`.
+
+## Configuration
+
+The simplest way to configure the `Sentry.Google.Cloud.Functions` package is by using the assembly method `FunctionsStartup` to include the type `SentryStartup`, for example:
+```csharp
+//your function.cs
+[assembly: FunctionsStartup(typeof(SentryStartup))]
+
+public class Function : IHttpFunction
+{
+  
+}
+```
+
+When using `[assembly: FunctionsStartup(typeof(SentryStartup))]`, the framework automatically loads `appsettings.json` and environment variables, binding to `SentryAspNetCoreOptions`.
+
+For example:
+
+```json {filename:appsettings.json}
+  "Sentry": {
+    "Dsn": "___PUBLIC_DSN___",
+    "MaxRequestBodySize": "Always",
+    "SendDefaultPii": true,
+    "MinimumBreadcrumbLevel": "Debug",
+    "MinimumEventLevel": "Warning",
+    "AttachStackTrace": true,
+    "Debug": true,
+    "DiagnosticsLevel": "Error"
+  },
+```
+
+> An example of some of the options that can be configured via `appsettings.json`.
+
+It's also possible to bind properties to the SDK via environment variables, like:
+
+Windows
+
+```shell
+set Sentry__Debug=true
+```
+
+Linux or macOS
+
+```shell
+export Sentry__Debug=true
+```
+
+ASP.NET Core will automatically read this environment variable and bind it to the SDK configuration object.
+
+When configuring the SDK via the frameworks configuration system, it's possible to add the SDK by simply calling `UseSentry` without providing any further information:
+
+```csharp
+public static IHostBuilder CreateHostBuilder(string[] args) =>
+    Host.CreateDefaultBuilder(args)
+        .ConfigureWebHostDefaults(webBuilder =>
+        {
+            // Add the following line:
+            webBuilder.UseSentry();
+        });
+```
+
+```fsharp
+let CreateHostBuilder args =
+    Host.CreateDefaultBuilder(args)
+        .ConfigureWebHostDefaults(fun webBuilder ->
+            // Add the following line:
+            webBuilder.UseSentry() |> ignore
+        )
+```
+
+Your DSN would be location if it was defined in `appsettings.json` or any other form of configuration.
+
+Some of the settings require actual code. For those, like the `BeforeSend` callback you will need to create a new FunctionsStartup class to set ip up:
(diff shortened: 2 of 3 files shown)


src/platforms/dotnet/guides/google-cloud-functions/index.mdx
```diff
@@ -0,0 +1,190 @@
+---
+title: Google Cloud Functions
+description: "Learn about Sentry's .NET integration with Google Cloud Functions."
+redirect_from:
+  - /platforms/dotnet/google.cloud.functions/
+---
+
+Sentry provides an integration with `Google Cloud Functions` through the [Sentry.Google.Cloud.Functions NuGet package](https://www.nuget.org/packages/sentry.google.cloud.functions).
+
+## Overview of the features
```
imatwawana, line 10:
```suggestion
## Overview of the Features
```

src/platforms/dotnet/guides/google-cloud-functions/index.mdx
```diff
@@ -0,0 +1,190 @@
...
+}
+```
+
+> Example modifying all events before they are sent to avoid server names being reported.
+
+## Dependency Injection
+
+Much of the behavior of the Google cloud functions integration with Sentry can be customized by using the frameworks dependency injection system. That is done by registering your own implementation of some of the exposed abstraction inside of `FunctionsStartup`.
+
+### Capturing the affected user
```
imatwawana, line 126:
```suggestion
### Capturing the Affected User
```

src/platforms/dotnet/guides/google-cloud-functions/index.mdx
```diff
@@ -0,0 +1,190 @@
...
+{
+    public override void ConfigureServices(WebHostBuilderContext context, IServiceCollection services)
+    {
+        base.ConfigureServices(context, services);
+        services.AddSingleton<IUserFactory, MyUserFactory>();
+    }
+}
+```
+
+## Options and Initialization
```
imatwawana, line 150:
```suggestion
## Options & Initialization
```

src/platforms/dotnet/guides/google-cloud-functions/index.mdx
```diff
@@ -0,0 +1,190 @@
...
+This parameter controls whether integrations should capture HTTP request bodies. It can be set to one of the following values:
+
+- `never`: Request bodies are never sent.
+- `small`: Only small request bodies will be captured. The cutoff for small depends on the SDK (typically 4KB).
+- `medium`: Medium and small requests will be captured (typically 10KB).
+- `always`: The SDK will always capture the request body as long as Sentry can make sense of it.
+
+If the request bodies should be captured, all requests will have the `EnableRewind` method invoked. This is done so that the request data can be read later, in case an error happens while processing the request.
+
+### Other options
```
imatwawana, line 183:
```suggestion
### Other Options
```

src/wizard/dotnet/google-cloud-functions.md
```diff
@@ -0,0 +1,101 @@
...
+To verify your setup, you can capture a message with the SDK:
+
+```csharp
+public Task HandleAsync(HttpContext context)
+{
+    SentrySdk.CaptureMessage("Hello Sentry");
+}
+```
+
+### Performance monitoring
```
imatwawana, line 66:
```suggestion
### Performance Monitoring
```

imatwawana (commented):
Just fixing some headings

src/platforms/dotnet/guides/google-cloud-functions/index.mdx
```diff
@@ -0,0 +1,190 @@
...
+description: "Learn about Sentry's .NET integration with Google Cloud Functions."
+redirect_from:
+  - /platforms/dotnet/google.cloud.functions/
+---
+
+Sentry provides an integration with `Google Cloud Functions` through the [Sentry.Google.Cloud.Functions NuGet package](https://www.nuget.org/packages/sentry.google.cloud.functions).
+
+## Overview of the Features
+
+- Easy Google Cloud Functions integration, single line: `UseSentry`
```
bruno-garcia, line 12: There's no UseSentry on GCP integration

src/platforms/dotnet/guides/google-cloud-functions/index.mdx
```diff
@@ -0,0 +1,190 @@
...
+- Captures unhandled exceptions in the middleware pipeline
+- Captures exceptions handled by the framework `UseExceptionHandler` and Error page display
+- Captures process-wide unhandled exceptions (AppDomain)
+- Captures `LogError` or `LogCritical`
+- Any event sent will include relevant application log messages
+- RequestId as tag
+- URL as tag
+- Environment is automatically set (IHostingEnvironment)
+- Release automatically set (`AssemblyInformationalVersionAttribute`, `AssemblyVersion` or environment variable)
```
bruno-garcia, line 21: Is this applicable on GCP?

src/includes/configuration/config-intro/dotnet.google-cloud-functions.mdx
```diff
@@ -0,0 +1,36 @@
+Options can be set on `appsettings.json` or on a class inheriting from `FunctionsStartup`.
```
bruno-garcia, line 1: It's not clear here that these are two alternatives.
Either provide values through `appsettings.json` OR provide the `CustomOptions` class.

Also, I don't see a usage of `[assembly: FunctionsStartup(typeof(SentryStartup))]`
Is this going to init the SDK?


See the sample: https://github.com/getsentry/sentry-dotnet/blob/main/samples/Sentry.Samples.Google.Cloud.Functions/Function.cs

src/platforms/dotnet/guides/google-cloud-functions/config.yml
```diff
@@ -0,0 +1,2 @@
+title: Google Cloud Functions
+sdk: sentry.google.cloud.functions
```
bruno-garcia, line 2: Below it's called `sentry.google.cloud.function` (note no **s**` at the end)
Which one is the correct?

```suggestion
sdk: sentry.google.cloud.functions
```

src/platforms/dotnet/guides/google-cloud-functions/index.mdx
```diff
@@ -0,0 +1,190 @@
...
+- Request body compressed
+- Event flooding protection (429 retry-after and internal bound queue)
+- Assembly Strong named
+
+## Installation
+
+Add the Sentry dependency:
+
+```powershell {tabTitle:Package Manager}
+Install-Package Sentry.Google.Cloud.Functions -Version {{ packages.version('sentry.dotnet.google-cloud-function') }}
```
bruno-garcia, line 39: I noticed the sdk name is `sentry.dotnet.google-cloud-function`
I believe I mixed up at first between `function` and functions` but would be worth double checking if it's aligned everywhere (at least consistent even though the package is `Sentry.Google.Functions` and sdk.name is `sentry.dotnet.google-cloud-function`

src/platforms/dotnet/guides/google-cloud-functions/index.mdx
```diff
@@ -0,0 +1,190 @@
...
+```
+
+```shell {tabTitle:.NET Core CLI}
+dotnet add package Sentry.Google.Cloud.Functions -v {{ packages.version('sentry.dotnet.google-cloud-function') }}
+```
+
+This package extends `Sentry` main SDK. That means that besides the logging-related features, through this package you'll also get access to all API and features available in the main `Sentry` SDK.
```
bruno-garcia, line 46:
```suggestion
This package extends `Sentry.AspNetCore` which in turn extends `Sentry`'s main SDK. That means that besides the logging-related features, through this package you'll also get access to all API and features available in those packages.
```

src/platforms/dotnet/guides/google-cloud-functions/index.mdx
```diff
@@ -0,0 +1,190 @@
...
+Messages logged from assemblies with the name starting with `Sentry` will not generate events.
+
+</Alert>
+
+## Configuration
+
+The simplest way to configure the `Sentry.Google.Cloud.Functions` package is by using the assembly method `FunctionsStartup` to include the type `SentryStartup`, for example:
+```csharp
+//your function.cs
+
```
bruno-garcia, line 59:
````suggestion
```csharp {tabTitle:Function.cs}
````

File: src/platforms/dotnet/guides/google-cloud-functions/index.mdx [hunk 13]
```diff
@@ -0,0 +1,190 @@
...
+Some of the settings require actual code. For those, like the `BeforeSend` callback you will need to create a new class implementing the abstract class `FunctionsStartup` to set it up:
+
+```csharp
+using Google.Cloud.Functions.Hosting;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using Sentry.AspNetCore;
+
+[assembly: FunctionsStartup(typeof(CustomOptions))]
```
bruno-garcia, line 101: Is there a chance this is loaded after the SDK initializes?
The reference to the options object is the same so in the case of BeforeSend it might not matter (only if events are captured before this runs) but what about if you set a value that's read during `init` and then not anymore? Like HTTP proxy?

src/platforms/dotnet/guides/google-cloud-functions/index.mdx
```diff
@@ -0,0 +1,190 @@
...
+        services.AddSingleton<IUserFactory, MyUserFactory>();
+    }
+}
+```
+
+## Options & Initialization
+
+As previously mentioned, this package is a wrapper around [Sentry.Extensions.Logging](/platforms/dotnet/guides/extensions-logging/), [Sentry.AspNetCore](/platforms/dotnet/guides/aspnetcore/) and [Sentry](/platforms/dotnet/). Please refer to the documentation of these packages to get the options that are defined at those levels.
+
+Below, the options that are specific to `Sentry.Google.Cloud.Functions` will be described.
```
bruno-garcia, line 154: It says that below are specifics to this package but the values below seems copied from other packages we tell the user to refer to the docs already above

src/wizard/dotnet/google-cloud-functions.md
```diff
@@ -0,0 +1,107 @@
...
+Additionally, you'll need to set up your `Sentry` settings on `appsettings.json`:
+
+```json
+{
+  "Sentry": {
+    "Dsn": "___PUBLIC_DSN___",
+    // Sends Cookies, User Id when one is logged on and user IP address to sentry. It's turned off by default.
+    "SendDefaultPii": true,
+    // Enable tracing.
+    "TracesSampleRate": 1.0,
```
bruno-garcia, line 47: This doesn't get any transactions in automatically since we're not adding the middleware yet

src/wizard/dotnet/google-cloud-functions.md
```diff
@@ -0,0 +1,107 @@
...
+    SentrySdk.ConfigureScope(scope => scope.Transaction = transaction);
+
+    // Transactions can have child spans (and those spans can have child spans as well)
+    var span = transaction.StartChild("test-child-operation");
+
+    // ...
+    // (Perform the operation represented by the span/transaction)
+    // ...
+
+    span.Finish(); // Mark the span as finished
```
bruno-garcia, line 94: In the event of an exception this won't be called an no transaction will be captured.

Ideally we'd make the integration start the transaction for the HTTP request and avoid this instruction altogether. We can then just show the StartChild here

src/platforms/dotnet/guides/google-cloud-functions/config.yml
lucas-zimerman: I'm using sentry.google.cloud.functions since it's the NuGet package name and sentry.dotnet.google-cloud-function only for the packages.version since it was defined with that name on sentry-release-registry.

src/platforms/dotnet/guides/google-cloud-functions/index.mdx
lucas-zimerman: `Configure<SentryAspNetCoreOptions>` is only invoked once the SDK is initialized.
If you set a parameter Like an HTTP Proxy, it'll stay set.

src/platforms/dotnet/guides/google-cloud-functions/config.yml
bruno-garcia: I believe the name here should match the `sdk.name` so should be `sentry.google.cloud.function`
Or we can bite the bullet and fix the release registry and SDK to report `sentry.google.cloud.functions`

src/platforms/dotnet/guides/google-cloud-functions/index.mdx
bruno-garcia: True so we'll have all these startup registrations get picked up first, all `ConfigureServices` will run and only once `Configure` runs we'll get the configuration callback invoked.
Should work (at least for the second init, since the first happens while setting up the logger only)

imatwawana:
@lucas-zimerman - what's the status of this PR?

PR #4175, opened 2021-09-22
fix(apple): Features for macOS and watchOS

philipphofmann:
The feature set of MacOS and WatchOS included too many features
and confuses users.
This is fixed now by removing the nonsupported features.

@imatwawana, I had to copy the pages. Is there maybe a way to reuse some common text of the pages?

### DIFF
--- a/src/includes/getting-started-primer/apple.macos.mdx
+++ b/src/includes/getting-started-primer/apple.macos.mdx
@@ -1,6 +1,6 @@
 <Note>
 
-Sentry's SDKs report an error automatically whenever a thrown error or exception goes uncaught in your application causing the application to crash. Sentry's iOS, tvOS, and watchOS SDK hook into all signal and exception handlers, except for MacOS.
+Sentry's SDKs report an error automatically whenever a thrown error or exception goes uncaught in your application causing the application to crash. Sentry's iOS, tvOS, and watchOS SDK hook into all signal and exception handlers, except for macOS.
 
 The SDK builds a crash report that persists to disk and tries to send the report right after the crash. Since the environment may be unstable at the crash time, the report is guaranteed to send once the application is started again.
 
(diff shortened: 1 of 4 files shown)


imatwawana:
> @imatwawana, I had to copy the pages. Is there maybe a way to reuse some common text of the pages?

You could try adding platform supported/not supported tags in the apple primer include rather than breaking it out into three includes. I'm not sure certain that will work, but might be worth a try to avoid the duplication.

imatwawana (approved):

bruno-garcia:
Renamed the PR to watchOS and macOS as that's the casing of the brand name:

image

src/includes/getting-started-primer/apple.macos.mdx
```diff
@@ -0,0 +1,27 @@
+<Note>
+
+Sentry's SDKs report an error automatically whenever a thrown error or exception goes uncaught in your application causing the application to crash. Sentry's iOS, tvOS, and watchOS SDK hook into all signal and exception handlers, except for MacOS.
```
bruno-garcia, line 3:
```suggestion
Sentry's SDKs report an error automatically whenever a thrown error or exception goes uncaught in your application causing the application to crash. Sentry's iOS, tvOS, and watchOS SDK hook into all signal and exception handlers, except for macOS.
```

src/includes/getting-started-primer/apple.macos.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+The SDK builds a crash report that persists to disk and tries to send the report right after the crash. Since the environment may be unstable at the crash time, the report is guaranteed to send once the application is started again.
+
+</Note>
+
+**Features:**
+
+- Multiple types of errors are captured, including:
```
bruno-garcia, line 11: The common entries could be moved [into an `include`](https://github.com/getsentry/sentry-docs/tree/master/src/includes) and then used anywhere it's needed.

Here's an example: https://github.com/getsentry/sentry-docs/tree/master/src/includes/configuration/diagnostic-logger

That can then be used in different platforms even:

https://github.com/getsentry/sentry-docs/blob/31b2ea2a13fd01b1929248199c82cf8f463d72e3/src/platforms/unity/common/configuration/diagnostic-logger.mdx#L7

https://github.com/getsentry/sentry-docs/blob/31b2ea2a13fd01b1929248199c82cf8f463d72e3/src/platforms/dotnet/common/configuration/diagnostic-logger.mdx#L7

PR #4183, opened 2021-09-23
Added worked example of using an authentication token

julianharty:
I struggled for a bit trying to make sense of how to format the command line with my authentication token. Here's a suggested change to the documentation to make the details of what to do clearer.

### DIFF
--- a/src/api/auth.mdx
+++ b/src/api/auth.mdx
@@ -11,7 +11,10 @@ Authentication tokens are passed using an auth header, and are used to authentic
 curl -H 'Authorization: Bearer {TOKEN}' https://sentry.io/api/0/projects/
 ```
 
-You can [find or create authentication tokens within Sentry](https://sentry.io/api/).
+You can [find or create authentication tokens within Sentry](https://sentry.io/api/). Replace `{TOKEN}` with your authentiation token, for instance if your authentication token is `1a2b3c` then the command would need to be:
+```bash
+curl -H 'Authorization: Bearer 1a2b3c' https://sentry.io/api/0/projects/
+```
 
 For self-hosted, you can find or create authentication tokens by visiting `{instance_url_prefix}/settings/account/api/auth-tokens/`
 


manuzope:
Hi @julianharty, thanks for the PR. This applies to the other methods too. We could replicate this sentence for other authentication types but that seems a little repetitive.

Would it have helped if the docs said `EXAMPLE_TOKEN` instead of a generic `TOKEN`?

julianharty:
@manuzope EXAMPLE_TOKEN might help a bit, however what confused me were the curly braces either side of the word and I'd still have been confused even if it said EXAMPLE_TOKEN. It probably took me about 15 minutes to realise what I needed to do, the error message in the API response was vague. Again if that was able to detect the caller had either left the placeholder unchanged or left the {} around the actual authentication token and provide a suitable clue then it may help improve the usability of the API and streamline the onboarding for devs using the APIs. There may well be ways to improve the documentation to keep it DRY yet informative. My approach was to simply explain what needed to happen once at the top of the relevant page. As a suggestion, perhaps you could accept a version of my PR (you're welcome to tweak it) and see if it helps future users (devs) of the API...

manuzope:
Gotcha, how about we just remove the curly braces from all the examples and make it "EXAMPLE_API_TOKEN", similar to here: https://docs.sentry.io/platforms/javascript/guides/gatsby/sourcemaps/troubleshooting_js/#using-the-api

Would that have helped?

julianharty:
> https://docs.sentry.io/platforms/javascript/guides/gatsby/sourcemaps/troubleshooting_js/#using-the-api

Yes it helps a bit (at least that removes the confusion of the opening and closing curly braces). However even that doesn't explain that AUTH_TOKEN needs to be replaced in order to actually use the API call successfully. 

Perhaps it'd be worth spending some time working out ways to help onboard people into using the API successfully  - some sort of getting started worked example/tutorial for the various ways devs can interact with your APIs? I expect some devs are so used to working with similar APIs that they'll have a unerring ability to subconsciously replace AUTH_TOKEN, etc. with the correct value. For others it'll become a speedbump they need to surmount, and a few may abandon hope at that point (or appeal on StackOverflow for someone to help them).  

I'd also suggest there's be links to the tutorial/worked example from various documentation pages. Possibly you could show this conditionally e.g. if they're deemed to be newcomers to the site and your ecosystem vs. those who can dismiss the prompt to look at the newbie oriented materials, etc. 

PS Thank you for considering this request seriously, I appreciate your time and help to improve the onboarding.

imatwawana:
@julianharty - Thanks for your thoughtful suggestions. I think a happy medium (that I must admit @manuzope initially suggested to me) is that we add a sentence at the top of the Authorization page indicating how all content in curly braces or chevrons should be treated throughout the API docs, along with an example like yours. I think this works for the following reasons: 
- Avoids the repetition of adding examples or explanation in several places
- Maintains the standard way of indicating text that should be replaced in API docs (see examples from [Microsoft](https://docs.microsoft.com/en-us/rest/api/apimanagement/apimanagementrest/azure-api-management-rest-api-authentication#ProgrammaticallyCreateToken) and [Google](https://developers.google.com/docs/api/quickstart/js#step_1_set_up_the_sample))
- Adds this information to a page that is likely the starting point for most folks new to API docs, so they'll see this before they kick off 

In addition, we'll change some of the content in the curly braces to be more clear (e.g., YOUR_AUTH_TOKEN).

I can't pull your branch into my local to access the appropriate lines of the file to make this change. Would you like to make the change on this PR or would you like me to create a new PR to make these changes?

julianharty:
@imatwawana - I'll have a go at making the excellent changes you suggest tomorrow. If I fail for any reason or if what I submit doesn't suit then you're welcome to create a new PR and make the changes yourself. I'd like to have a go though - it does me good and helps me learn a bit more about your project.

imatwawana:
@julianharty - Excellent! FYI, I'll be OOO after today until October 12th, but I'll review your updates as soon as I return.

julianharty:
@imatwawana thanks for your suggested changes generally they make sense to me with one exception, in https://github.com/getsentry/sentry-docs/pull/4183#discussion_r727177085 where I used the conditional 'if'. For me your suggested changes don't quite work grammatically... In my view there needs to be an 'if' or a 'when' in the sentence after the comma for the sentence to make sense.
e.g. For example, "**when** the documentation says:" or  "**if** the documentation says:" 
The choice somewhat depends on whether the guidance is for a single consistent term that needs to be replaced throughout e.g. if all the documentation uses `{TOKEN}` and no other term, then **when** probably is more correct than if. However, if there are several possible terms that need replacing then I'd suggest **if** would be more correct.

You've requested a review for accuracy, etc. from @manuzope , is there anything you'd like me to do in the interim?

imatwawana:
hey @julianharty, IMO the "for example" implies the "if" or "when". While I don't see this being a barrier to comprehension in most cases, I'm happy to add in a "when". I can either commit the changes that don't require @manuzope's review or you can do so. I just wanted you to see them before I committed anything.

manuzope:
Made some suggestions to the suggestions! Once those are in, I think we should be good to go.

PR #4188, opened 2021-09-24
Add performance to Android Wizard

marandaneto:


philipphofmann (approved):
LGTM 🚀

bruno-garcia (approved):
Thank you

src/wizard/android/index.md
```diff
@@ -69,6 +79,39 @@ class MyActivity : AppCompatActivity() {
...
+You can measure the performance of your code by capturing transactions and spans.
+
+```kotlin
+import io.sentry.Sentry
+
+// Transaction can be started by providing, at minimum, the name and the operation
+val transaction = Sentry.startTransaction("updateRepos", "task")
+
+// Transactions can have child spans (and those spans can have child spans as well)
+val span = transaction.startChild("test-child-operation")
```
philipphofmann, line 102: `l`: Maybe we should choose better operation names, as their are used for filtering.

src/wizard/android/index.md
```diff
@@ -69,6 +79,40 @@ class MyActivity : AppCompatActivity() {
...
```
philipphofmann, line 103: l: Maybe we should choose better operation names, as they are used for filtering.
```suggestion
val transaction = Sentry.startTransaction("Update Repos", "db")

// Transactions can have child spans (and those spans can have child spans as well)
val span = transaction.startChild("db", "Update first repo")
```

PR #4191, opened 2021-09-24
Align transaction name and op - Android wizard

marandaneto:
similar to C#
https://github.com/getsentry/sentry-docs/blob/master/src/wizard/dotnet/index.md


philipphofmann:
I chose different names, see https://github.com/getsentry/sentry-docs/pull/4192/files, because operations are used for filtering and stats.

marandaneto:
> I chose different names, see [#4192 (files)](https://github.com/getsentry/sentry-docs/pull/4192/files), because operations are used for filtering and stats.

makes sense, but it's just an example, nobody would use these `name` and `op` anyway

PR #4201, opened 2021-09-28
fix: Add fallback userID to mobile SDKs

philipphofmann:
Align or add info on fallback for userID to installationID for
Android, Apple, Dart, and ReactNative.

### DIFF
--- a/src/includes/enriching-events/set-user/android.mdx
+++ b/src/includes/enriching-events/set-user/android.mdx
@@ -19,6 +19,6 @@ Sentry.setUser(user)
 
 <Note>
 
-If you do not provide a user identity, the SDK falls back to `installationId`, which the SDK once randomly generates during an app's installation.
+If you do not provide a user identity, the SDK falls back to `installationId`, which the SDK randomly generates once during an app's installation.
 
 </Note>
--- a/src/includes/enriching-events/set-user/android.mdx
+++ b/src/includes/enriching-events/set-user/android.mdx
@@ -19,6 +19,6 @@ Sentry.setUser(user)
 
 <Note>
 
-Sentry will by fallback to a random generated `installationId` if you do not provide a user identity.
+If you do not provide a user identity, the SDK falls back to `installationId`, which the SDK once randomly generates during an app's installation.
 
 </Note>
--- a/src/includes/enriching-events/set-user/apple.mdx
+++ b/src/includes/enriching-events/set-user/apple.mdx
@@ -16,6 +16,6 @@ user.email = @"john.doe@example.com";
 
 <Note>
 
-Since 6.0.1: If you do not provide a user identity, the SDK falls back to `installationId`, which the SDK once randomly generates during an app's installation.
+Since 6.0.1: If you do not provide a user identity, the SDK falls back to `installationId`, which the SDK randomly generates once during an app's installation.
 
 </Note>
--- a/src/includes/enriching-events/set-user/apple.mdx
+++ b/src/includes/enriching-events/set-user/apple.mdx
@@ -14,4 +14,8 @@ user.email = @"john.doe@example.com";
 [SentrySDK setUser:user];
 ```
 
-Since 6.0.1: Sentry will by fallback to `installationId` if you do not provide a user identity.
+<Note>
+
+Since 6.0.1: If you do not provide a user identity, the SDK falls back to `installationId`, which the SDK once randomly generates during an app's installation.
+
+</Note>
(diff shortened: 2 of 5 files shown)


src/includes/enriching-events/set-user/dart.mdx
```diff
@@ -5,3 +5,9 @@ Sentry.configureScope(
   (scope) => scope.user = SentryUser(id: '1234', email: 'jane.doe@example.com'),
 );
 ```
+
+<Note>
+
+If you do not provide a user identity, the SDK falls back to `installationId`, which the SDK once randomly generates during an app's installation.
+
+</Note>
```
marandaneto, line 13: we should create a new file (copy from this one) called `flutter.mdx` and add this information only for Flutter, the Dart SDK does not have such fallback since there's no offline caching feature.

src/includes/enriching-events/set-user/android.mdx
```diff
@@ -19,6 +19,6 @@ Sentry.setUser(user)
 
 <Note>
 
-Sentry will by fallback to a random generated `installationId` if you do not provide a user identity.
+If you do not provide a user identity, the SDK falls back to `installationId`, which the SDK once randomly generates during an app's installation.
```
marandaneto, line 22: any reason to change the wording, previous text has been reviewed by the docs team already and seems clear to me

marandaneto:
missing capacitor, cordova, xamarin, and unity if they apply as well.
@jennmueng and @bitsandfoxes would know

src/includes/enriching-events/set-user/dart.mdx
philipphofmann: Good catch, thanks.

src/includes/enriching-events/set-user/android.mdx
philipphofmann: I thought that maybe some users won't understand that the `installationId` stays the same. I can also revert it back to the old version.

marandaneto: no strong feeling either, maybe @imatwawana find it even better, just wanna understand the motivation, but I recall being more specific about that, and Fiona reworded this way

imatwawana: Given that @philipphofmann wants to convey more information, I think the second is fine with a few tweaks. And the previous sentence didn't entirely make sense anyway. The word "by" shouldn't be there.

```suggestion
If you do not provide a user identity, the SDK falls back to `installationId`, which the SDK randomly generates once during an app's installation.
```

src/includes/enriching-events/set-user/apple.mdx
```diff
@@ -14,4 +14,8 @@ user.email = @"john.doe@example.com";
 [SentrySDK setUser:user];
 ```
 
-Since 6.0.1: Sentry will by fallback to `installationId` if you do not provide a user identity.
+<Note>
+
+Since 6.0.1: If you do not provide a user identity, the SDK falls back to `installationId`, which the SDK once randomly generates during an app's installation.
```
imatwawana, line 19:
```suggestion
Since 6.0.1: If you do not provide a user identity, the SDK falls back to `installationId`, which the SDK randomly generates once during an app's installation.
```

src/includes/enriching-events/set-user/flutter.mdx
```diff
@@ -0,0 +1,13 @@
...
+import 'package:sentry/sentry.dart';
+
+Sentry.configureScope(
+  (scope) => scope.user = SentryUser(id: '1234', email: 'jane.doe@example.com'),
+);
+```
+
+<Note>
+
+If you do not provide a user identity, the SDK falls back to `installationId`, which the SDK once randomly generates during an app's installation.
```
imatwawana, line 11:
```suggestion
If you do not provide a user identity, the SDK falls back to `installationId`, which the SDK randomly generates once during an app's installation.
```

src/includes/enriching-events/set-user/react-native.mdx
```diff
@@ -0,0 +1,9 @@
+```javascript
+Sentry.setUser({ email: "john.doe@example.com" });
+```
+
+<Note>
+
+If you do not provide a user identity, the SDK falls back to `installationId`, which the SDK once randomly generates during an app's installation.
```
imatwawana, line 7:
```suggestion
If you do not provide a user identity, the SDK falls back to `installationId`, which the SDK randomly generates once during an app's installation.
```

imatwawana (approved):

philipphofmann:
> missing capacitor, cordova, xamarin, and unity if they apply as well. @jennmueng and @bitsandfoxes would know

Thanks for pointing that out. I created an issue for this @marandaneto https://github.com/getsentry/sentry-docs/issues/4206

PR #4203, opened 2021-09-28
consolidate sentry basics issue grouping content into issue grouping guide

imatwawana:
- Moves content of Sentry Basics > Best Practices > Grouping Issues and consolidates it into Data Management > Issue Grouping
- Several edits to existing pages under Issue Grouping
- Adds new page called Merging Issues
- Updates screenshots


src/docs/product/data-management-settings/event-grouping/index.mdx
```diff
@@ -57,4 +62,22 @@ If the stack trace is not available, but exception information is, then the grou
...
+
+There are four different ways in in [sentry.io](https://sentry.io) to update how events group into issues:
+
+- [Fingerprint Rules](/product/data-management-settings/event-grouping/fingerprint-rules/): Sets a new fingerprint for incoming events based on matchers. This doesn't affect already existing issues.
+- [Stack Trace Rules](/product/data-management-settings/event-grouping/stack-trace-rules/): Sets rules for how incoming events should be treated based on matchers. This doesn't affect already existing issues.
+- [Merging Issues](/product/data-management-settings/event-grouping/merging-issues/): Merges together similar issues that have already been created. No settings or configuration changes are required for this.
+- [Grouping Breakdown](/product/data-management-settings/event-grouping/grouping-breakdown/): Creates coarser issue groups by considering the most relevant part of the stack trace.
```
untitaker, line 83: this section seems a bit repetitive since we have the same 4 points at the top of the page too. maybe it's a better intro for the page than "all events have a fingerprint"

untitaker (approved):

src/docs/product/data-management-settings/event-grouping/index.mdx
imatwawana: Yeah I was worried about the repetition, but wasn't sure of exactly what to do since the set of links at the top is slightly different from the set of links at the bottom. But without the links, this section wouldn't offer a solution for how to fix the problem. And with regard to the intro, do you mean just the phrase "all events have a fingerprint" or more of the intro? Maybe could you use the suggest feature to show me what you would propose instead?

src/docs/product/data-management-settings/event-grouping/fingerprint-rules.mdx
```diff
@@ -27,6 +32,8 @@ error.type:ConnectionError -> system-down
 
 error.value:"connection error: *" -> connection-error, {{ transaction }}
 ```
+Now, all the events with the error type `DatabaseUnavailable` or `ConnectionError` will be grouped into an issue with the type `system-down`. All events with the error value `connection error` will be grouped into an issue with a type of the same name and the name of the transaction appended.
+<!-- is what i'm saying about transaction correct here? not sure -->
```
imatwawana, line 36: @untitaker @jjbayer - is what i've said here about the error.value:connection error (last) example true?

src/docs/product/data-management-settings/event-grouping/merging-issues/index.mdx
```diff
@@ -0,0 +1,25 @@
+---
+title: "Merging Issues"
+sidebar_order: 40
+description: "Learn about merging similar issues that aren't automatically grouped."
+---
+
+If you have similar-looking issues that have not been grouped automatically and you want to reduce noise, you can do so by merging them.
+
+This is for merging similar issues and will not auto-merge the next occurrence of this issue coming in.
+<!-- what exactly what does the statement above mean? Do we mean, this won't happen if the next event doesn't have the same stack trace? See note at end.-->
```
imatwawana, line 10: @untitaker @jjbayer I'm not sure what we mean here as it pertains to what's on line 23. Can you clarify?

src/docs/product/data-management-settings/event-grouping/fingerprint-rules.mdx
untitaker: first sentence is correct, but the behavior for the last rule is that all matching events will be grouped by transaction name.

So if your transaction /api/users/foo/ crashes with a value "connection error", and your transaction /api/events/foo/ crashes the same way, that'll be exactly two issues regardless of stacktrace or whatever else we would normally group by

src/docs/product/data-management-settings/event-grouping/merging-issues/index.mdx
untitaker: I don't know where this content comes from but essentially it means that this is a one-off operation, and we don't try to infer any "rules" from a user merging two issues together. I would omit the sentence entirely because I can't think of a better way to explain and I don't think it's relevant at this stage. Perhaps we can point to fingerprinting rules at the bottom of the page, as a way to tweak grouping more aggressively if issue merging doesn't cut it. Replace alert at the bottom with:

>Future events will be added to the merged issue by the same criteria as they would've been added to the duplicate issues. We don't try to infer any general rules from how you merge issues. If you see that Sentry creates more issues than you can merge, or that Sentry creates too few issues, check out Fingeprinting Rules and Stack Trace Rules.

src/docs/product/data-management-settings/event-grouping/index.mdx
untitaker: I would update the list on the top to include:

* merging issues
* fingerprint rules
* stack trace rules
* sdk fingerprints
* grouping breakdown

(that order goes from easy to hard/advanced feature)

..then delete this entire section

I believe the links are the same though, i.e. Fingerprint Rules at the top and the bottom point to the same thing.

src/docs/product/data-management-settings/event-grouping/index.mdx
```diff
@@ -5,23 +5,28 @@ redirect_from:
...
   - /learn/rollups/
   - /data-management/event-grouping/
-description: "Learn about fingerprints, the grouping algorithm, grouping by stack trace, grouping by exceptions, and fallback grouping."
+  - /guides/grouping-and-fingerprints/
+  - /product/sentry-basics/guides/grouping-and-fingerprints/
+description: "Learn about fingerprints, the grouping algorithm, grouping by stack trace, grouping by exceptions, and fallback grouping. Learn more about how Sentry groups issues together as well as different approaches for updating how events group into issues"
 ---
 
-All events have a fingerprint. Events with the same fingerprint are grouped together into an issue.
+A fingerprint is a way to uniquely identify an event, and all events have one. Events with the same fingerprint are grouped together into an issue. So if two events have the same stack trace, then they will group into one issue.
```
jjbayer, line 13: As multiple events can have the same fingerprint, I would not write "uniquely identify an event" here. Maybe "uniquely identify an error"?

jjbayer: Also maybe we can remove the second sentence, because the same information is repeated in the next paragraph?

src/docs/product/data-management-settings/event-grouping/index.mdx
```diff
@@ -5,23 +5,28 @@ redirect_from:
...
 
-By default, Sentry will run one of our built-in grouping algorithms to generate a fingerprint based on information available within the event such as `stacktrace`, `exception`, and `message`. To extend the default grouping behavior or change it completely, you can use a combination of the following options:
+By default, Sentry will run one of our built-in grouping algorithms to generate an event fingerprint based on information available within the event, such as `stacktrace`, `exception`, and `message`. If the way that Sentry is grouping issues for your organization is already ideal, you don't need to make any changes. However, you can extend the default grouping behavior or change it completely, using a combination of the following options:
 
 1. In your SDK, using [SDK Fingerprinting](/platform-redirect/?next=/usage/sdk-fingerprinting/). Note that is not supported in WebAssembly.
 2. In your project, using [Fingerprint Rules](./fingerprint-rules/)
 3. In your project, using [Stack Trace Rules](./stack-trace-rules/)
+4. In your project, using [Grouping Breakdown](./grouping-breakdown). Early Adopters only.
```
jjbayer, line 20: The current Grouping Breakdown only provides information, it does not allow the user to extend or change the grouping behavior. It probably will in the future. @untitaker what do you think, should we list it here anyway?

src/docs/product/data-management-settings/event-grouping/merging-issues/index.mdx
```diff
@@ -0,0 +1,25 @@
+---
+title: "Merging Issues"
+sidebar_order: 40
+description: "Learn about merging similar issues that aren't automatically grouped."
+---
+
+If you have similar-looking issues that have not been grouped automatically and you want to reduce noise, you can do so by merging them.
```
jjbayer, line 7:
```suggestion
If you have similar-looking issues that have not been grouped together automatically and you want to reduce noise, you can do so by merging them.
```

src/docs/product/data-management-settings/event-grouping/merging-issues/index.mdx
```diff
@@ -0,0 +1,25 @@
...
+
+This is for merging similar issues and will not auto-merge the next occurrence of this issue coming in.
+<!-- what exactly what does the statement above mean? Do we mean, this won't happen if the next event doesn't have the same stack trace? See note at end.-->
+
+In this example, before merge, there are two issues that are very similiar, but have not been grouped together. One issue has 39 events in the past 24 hours, and the other has 76.
+![Before merge.](before-merge.png)
+
+After the two issues are are merged into one, the event count of 115 on that one issue reflects that they've been grouped.
```
jjbayer, line 15:
```suggestion
After the two issues are are merged into one, the event count of 115 on that one issue reflects that they've been grouped together.
```

Not sure if this makes a difference to out target audience, or only to me.

PR #4212, opened 2021-09-30
Fix typo for ANR manifest meta-data name

lfielke:
The current docs say the name of the meta-data element is
`io.sentry.anr.timeout-interval-mills`
however I think this might be a typo for (note the "i" near the end):
`io.sentry.anr.timeout-interval-millis`

Assuming this is the constant for the name:
https://github.com/getsentry/sentry-java/blob/dd6f9dbe69575869b94c85d03411285ea932a829/sentry-android-core/src/main/java/io/sentry/android/core/ManifestMetadataReader.java#L27


bruno-garcia:
Thanks @lfielke

PR #4214, opened 2021-09-30
Added OpenReplay (ex Asayer)

estradino:
Asayer got rebranded into OpenReplay. Also added the URL to our website.


manuzope (changes requested):
Thanks for the PR @estradino. Before we can make this change, can we make the change in the product (https://github.com/getsentry/sentry/search?q=asayer). I'd also change the name of [this](https://github.com/getsentry/sentry/blob/master/src/sentry/static/sentry/images/logos/logo-asayer.svg) file.

Otherwise, it might be confusing to folks reading the docs and being unable to find OpenReplay in our integrations list.

Screen Shot 2021-10-04 at 11 03 08 AM

PR #4226, opened 2021-10-03
Fix unescaped backslash

ZSwaff:
Unescaped, the backslash does not appear in the docs either on GitHub or live at https://docs.sentry.io/product/cli/releases/#creating-releases. This change escapes it so that it appears.


kamilogorek:
Thanks!

PR #4227, opened 2021-10-04
Document Python HTTPX integration

### LINKED ISSUE #4036: Add documentation for Python Httpx Integration
### Core or SDK?

Platform/SDK

### Which part? Which one?

sentry-python

### Description

https://github.com/getsentry/sentry-python/pull/1119 Merged in adding an integration for Httpx for Python which is not documented.

### Suggested Solution

We should add docs for this so users can find out about it.

rhcarvalho:
Fixes #4036.

### DIFF
--- a/src/platforms/python/common/configuration/integrations/httpx.mdx
+++ b/src/platforms/python/common/configuration/integrations/httpx.mdx
@@ -1,0 +1,20 @@
+---
+title: HTTPX
+description: "Learn about the HTTPX integration and how it adds support for the HTTPX HTTP client."
+---
+
+_(New in version 1.2.0)_
+
+The HTTPX integration instruments outgoing HTTP requests using either the sync or the async [HTTPX](https://www.python-httpx.org/) clients.
+
+Use this integration to create spans for outgoing requests and ensure traces are properly propagated to downstream services.
+
+```python
+import sentry_sdk
+from sentry_sdk.integrations.httpx import HttpxIntegration
+
+sentry_sdk.init(
+    dsn="___PUBLIC_DSN___",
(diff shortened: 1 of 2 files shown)


src/platforms/python/common/configuration/integrations/httpx.mdx
```diff
@@ -0,0 +1,20 @@
+---
+title: HTTPX
+description: "Learn about the HTTPX integration and how it adds support for the HTTPX HTTP client."
+---
+
+_(New in version 1.2.0)_
+
+The HTTPX integration instruments outgoing HTTP requests using either the sync or the async [HTTPX](https://www.python-httpx.org/) clients.
+
+Use this integration to create spans for outgoing requests and ensure traces are properly propagated to downstream services.
```
AbhiPrasad, line 10: Shall we include a note like:

```suggestion
Use this integration to create spans for outgoing requests and ensure traces are properly propagated to downstream services. For more information on setting up performance monitoring with Sentry, please see [Getting Started with Python Performance](/platforms/python/performance/).
```

AbhiPrasad (approved):

src/platforms/python/common/configuration/integrations/httpx.mdx
rhcarvalho: I considered something like that but gave up for a few reasons:

- no precedent in other integration pages (e.g. the standard library integration doesn't even mention spans for `httplib`).
- The performance page doesn't link back (unless we make more changes), for example to suggest you may need to add more integrations.

The above would possibly be valid future improvements. Let's at least get it started and go step-by-step ;)

PR #4229, opened 2021-10-04
ref(apple): Remove duplicate wizard pages

philipphofmann:
Apple has three wizard files, but we only need one.


marandaneto:
@philipphofmann were they identical?

philipphofmann:
Initially, yes, then I added performance to the wizard and didn't realize that there were 3 of them. They should be identical.

PR #4233, opened 2021-10-05
ref(slack): Add troubleshooting section

ceorourke:
Add a troubleshooting section to the Slack setup docs telling users what to do if they are trying to save an alert rule and are getting rate limited. This will be linked to in the product as well.

### DIFF
--- a/src/docs/product/integrations/notification-incidents/slack/index.mdx
+++ b/src/docs/product/integrations/notification-incidents/slack/index.mdx
@@ -47,6 +47,12 @@ Your Slack integration is now available to all projects in your Sentry organizat
 
 In the next section, we'll walk you through configuring your notification settings.
 
+### Troubleshooting
+
+#### Rate Limiting Error
+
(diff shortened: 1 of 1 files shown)


src/docs/product/integrations/notification-incidents/slack/index.mdx
```diff
@@ -47,6 +47,12 @@ Your Slack integration is now available to all projects in your Sentry organizat
 
 In the next section, we'll walk you through configuring your notification settings.
 
+### Troubleshooting
```
mgaeta, line 50: I'd either put this at the bottom of the page or just above the transitional sentence: `In the next section, we'll walk you through configuring your notification settings.`.

ceorourke: @mgaeta at the _end_ end or before "Deleting the legacy project-based Slack integration"?

ceorourke: @mgaeta also I was following what we do on this page: https://docs.sentry.io/product/integrations/cloud-monitoring/aws-lambda/#troubleshooting

Issue #4187, opened 2021-09-24
React Native migration page from 2.x to 3.x


marandaneto:
### Core or SDK?

Platform/SDK

### Which part? Which one?

react-native

### Description

Migration page from 2.x to 3.x is missing
https://docs.sentry.io/platforms/react-native/migration/
https://github.com/getsentry/sentry-react-native/releases

### Suggested Solution

Document if any breaking changes, if none, close the issue

marandaneto:
cc @jennmueng

jennmueng:
@marandaneto Putting this as next priority after new react navigation instrumentation

PR #4234, opened 2021-10-06
feat: React Native 3.x migration guide and document event.origin/event.environment tags

jennmueng:
Adds a RN 3.x migration page that says no change is needed.

Add a new `Event Information` page under `Usage` (unless this fits better somewhere else?) that documents `event.origin` and `event.environment` tags.

Resolves https://github.com/getsentry/sentry-docs/issues/4187

### DIFF
--- a/src/platforms/react-native/migration.mdx
+++ b/src/platforms/react-native/migration.mdx
@@ -6,7 +6,7 @@ description: "Learn about migrating from 1.x to 2.x to enable release health tra
 
 ## From 2.x to 3.x
 
-There are no changes needed when migrating from versions `2.x` to `3.x`, although you will need to make sure that you run `pod install` on iOS and update your native code on both platforms.
+There are no changes needed when migrating from versions `2.x` to `3.x`, although you will need to make sure that you run `pod install` on iOS and to rebuild your app on both platforms.
 
 ### Event origin and environment tags
 


src/platforms/react-native/migration.mdx
```diff
@@ -4,6 +4,14 @@ sidebar_order: 100
 description: "Learn about migrating from 1.x to 2.x to enable release health tracking and native stack traces by default."
 ---
 
+## From 2.x to 3.x
+
+There are no changes needed when migrating from versions `2.x` to `3.x`, although you will need to make sure that you run `pod install` on iOS and update your native code on both platforms.
```
marandaneto, line 9: if its iOS only, why do you say both platforms?

src/platforms/react-native/usage/event-information.mdx
```diff
@@ -0,0 +1,26 @@
...
+description: "Learn about the information we attach onto events."
+---
+
+The Sentry React Native SDK will automatically add information onto your event to make it easier to debug and understand your app.
+
+## Event Origin and Environment Tags
+
+The event origin tag tells you which layer of your app the event originated from.
+
+The possible values for `event.origin` are:
```
AbhiPrasad, line 12: Can we put `event.origin` and `event.environment` and it’s possible values in the develop docs?

src/platforms/react-native/migration.mdx
jennmueng: Hmm guess this is bad wording, this is more of that you will need to rebuild your app after this update. Even though you're supposed to for every update, RN devs often use codepush which just updates the js code ota so this is just to make it clear that this will definitely need a rebuild.

src/platforms/react-native/usage/event-information.mdx
jennmueng: Agreed, so we can align the values across the SDKs

PR #4253, opened 2021-10-08
fix(android): can't assign val ListProperty in Kotlin

Whathecode:
I just tried setting ignored build types in Kotlin/Gradle following the existing documentation and got an error "Val cannot be reassigned".

Instead, `set` in Kotlin works.


marandaneto:
@Whathecode thanks for doing this, indeed, our unit tests only covered with groovy code.

PR #4266, opened 2021-10-13
Clarify not supported Next.js features

iker-barriocanal:
We should add to the docs the Next.js features we don't support. I'm not sure what the best place is, so I've added it in the paragraph after the features in the getting started page (see picture below). If the set of unsupported features grows, we can make an actual list, make another paragraph, add a section... If you can think of any other place where this section may suit better, please suggest.

image


iker-barriocanal:
At the moment, not planned. There's [this new issue](https://github.com/getsentry/sentry-javascript/issues/4062) related to it too.

PR #4278, opened 2021-10-16
Steps for Sentry integration with SST

jayair:
Adding a section on how to automatically add Sentry to all the Lambda functions in your SST stacks.

Based on: https://docs.serverless-stack.com/monitoring-your-app-in-prod#sentry


jayair:
Hey @AbhiPrasad, yeah that could work. Maybe instead of the full steps and code snippets, we can just mention the 2 steps: import the layer and add the environment variables. Then link to the docs on our side? So at least somebody reading it has a general idea of what's going on from the Sentry side?

Issue #4290, opened 2021-10-20
Release health is not enabled by default in .NET sdk


rodolfoBee:
### Core or SDK?

Platform/SDK

### Which part? Which one?

.NET SDK

### Description

Page: https://docs.sentry.io/platforms/dotnet/configuration/releases/#release-health

As commented in the source code, AutoSessionTracking defaults to FALSE instead of TRUE as mentioned in the docs.

https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/SentryOptions.cs#L552-L562

### Suggested Solution

Edit the text/code snippets to reflect actual SDK behaviour.

philipphofmann:
Thanks for reporting this @rbisol.

PR #4291, opened 2021-10-20
fix(dotnet): Default value of AutoSessionTracking

philipphofmann:
The default value of AutoSessionTracking is false instead of true.
This is fixed now.

Fixes GH-4290

### DIFF
--- a/src/includes/configuration/auto-session-tracking/dotnet.mdx
+++ b/src/includes/configuration/auto-session-tracking/dotnet.mdx
@@ -4,7 +4,7 @@ We mark the session as:
 - Errored: if the SDK captures an event that contains an exception (this includes manually captured errors).
 - Crashed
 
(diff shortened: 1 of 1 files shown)


philipphofmann:
> before merging, we'd need to know if the docs are actually the one to be fixed, IMO, the SDK should be fixed ^^

Fixing the SDK will take longer. At least the docs are now in sync with the SDK. We can update the docs later if we decide to fix the SDK. Is that fine for you @marandaneto?

marandaneto:
> > before merging, we'd need to know if the docs are actually the one to be fixed, IMO, the SDK should be fixed ^^
> 
> Fixing the SDK will take longer. At least the docs are now in sync with the SDK. We can update the docs later if we decide to fix the SDK. Is that fine for you @marandaneto?

I am fine with that, but its just a boolean in the SDK side too, so I will leave that to @bruno-garcia (if we merge docs or not)

Issue #4308, opened 2021-10-25
Missing information in vercel integration docs


jotacv:
### Core or SDK?

Core Sentry product

### Which part? Which one?

https://docs.sentry.io/product/integrations/deployment/vercel/#project-linking

### Description

Step 1 states _Select a Sentry project and a Vercel project to link together_, but where should I select that? The steps should include a place to start from.

### Suggested Solution

The docs should something like: 

1. Go to https://whatever-url-that-I-need-to-visit-before-step-2
2. Select a Sentry project and a Vercel project to link together
3. ...the rest of the docs

PR #4314, opened 2021-10-26
fix(integrations): Clarify Vercel/Sentry project-linking step

lobsterkatie:
It's currently not obvious that this step isn't independent of installation but part of it.

Fixes https://github.com/getsentry/sentry-docs/issues/4308.


imatwawana (approved):
Sorry! Didn't realized I hadn't actually approved this.

PR #4335, opened 2021-10-29
feat(discover): World Map and Bar Chart display options and Multi Y-Axis Selection

edwardgou-sentry:
Added new World Map and Bar Chart display options to list and added section on Multi Y-Axis Selection
image


imatwawana (approved):
One comment about the heading for y-axis update and a few wording changes, but otherwise LGTM!

PR #4336, opened 2021-10-29
docs(laravel): added alternative Sentry setup with logging channel

sykezz:
- Added alternative Sentry setup via logging channel to the install guide
- Added Laravel logging level doc link
- Changed logging channel default level to LOG_LEVEL environment variable
- Fixed Laravel doc link to latest (8.x)

Logging through a channel is pretty neat if you want your info and debug logs on Sentry as well (I use it myself), and I think this information should be available on the installation page as well as an alternative setup.


src/platforms/php/guides/laravel/index.mdx
```diff
@@ -20,6 +20,8 @@ Install the `sentry/sentry-laravel` package:
 composer require sentry/sentry-laravel
 ```
 
+### Sentry reporting via Exception Handler
```
imatwawana, line 23:
```suggestion
### Sentry Reporting Using Exception Handler
```

src/platforms/php/guides/laravel/index.mdx
```diff
@@ -33,6 +35,32 @@ public function register()
 }
 ```
 
+### Alternative: Log Channel
+
+Add Sentry as a log channel in `config/logging.php`, and configure your app to both log to a log file and to Sentry by modifying the log stack:
```
imatwawana, line 40:
```suggestion
Add Sentry as a log channel in `config/logging.php`. Then configure your app to both log to a log file and to Sentry by modifying the log stack:
```

src/platforms/php/guides/laravel/index.mdx
```diff
@@ -33,6 +35,32 @@ public function register()
...
+        'driver' => 'sentry',
+    ],
+],
+```
+
+You can further [configure the log channel](usage/#log-channels) to set the logging level, so you can log `info` and `debug` as well.
+
+<Alert level="info" title="Note">
+
+If you're using log channels to log your exceptions and are also logging exceptions to Sentry in your [exception handler](#sentry-reporting-via-exception-handler), exceptions might end up twice in Sentry.
```
imatwawana, line 60:
```suggestion
If you're using log channels to log your exceptions and are also logging exceptions to Sentry in your [exception handler](#sentry-reporting-via-exception-handler), exceptions might show up in Sentry twice.
```

src/platforms/php/guides/laravel/usage/index.mdx
```diff
@@ -90,7 +90,7 @@ class SentryContext
 
 <Alert level="info" title="Note">
 
-If you're using log channels to log your exceptions and are also logging exceptions to Sentry in your exception handler (as you would have configured above) exceptions might end up twice in Sentry.
+If you're using log channels to log your exceptions and are also logging exceptions to Sentry in your exception handler (as you would have configured above), exceptions might end up twice in Sentry.
```
imatwawana, line 93:
```suggestion
If you're using log channels to log your exceptions and are also logging exceptions to Sentry in your exception handler (as you would have configured above), exceptions might show up in Sentry twice.
```

src/platforms/php/guides/laravel/usage/index.mdx
```diff
@@ -120,23 +120,24 @@ After you configured the Sentry log channel, you can configure your app to both
 ],
 ```
 
-Optionally, you can set the logging level and if events should bubble on the driver:
+Optionally, you can set the ([logging level](https://laravel.com/docs/8.x/logging#log-levels)) and if events should bubble on the driver:
```
imatwawana, line 123:
```suggestion
Optionally, you can set the [logging level](https://laravel.com/docs/8.x/logging#log-levels) and if events should bubble on the driver:
```

imatwawana (approved):
Thanks for logging this! I've made some style and wording edits. This looks good so far, but will just require technical and structural review.

imatwawana:
@stayallive, can you take a look at this PR for technical review?

src/platforms/php/guides/laravel/usage/index.mdx
stayallive: We should probably also change `as you would have configured above` to `as you would have configured in the getting started section)` since the `getting started` section is nog longer above this alert.

stayallive:
It all checks out, just a few minor changes noted too for consistency.

imatwawana:
Thanks @stayallive! 

@sykezz, I want to review the structure of this with one of our developers tomorrow and then I'll circle back.

src/platforms/php/guides/laravel/index.mdx
```diff
@@ -33,6 +35,32 @@ public function register()
 }
 ```
 
+### Alternative: Log Channel
```
bruno-garcia, line 38: What are users missing out if they pick this versus the other? I'm not familiar with PHP but wondering why there are these two alternatives

bruno-garcia: Just saw @stayallive already went through. LGTM, just curious here?

bruno-garcia (approved):

src/platforms/php/guides/laravel/index.mdx
stayallive: The log channel has a lot more noise considering it also contains, well logs. And not everything dumped in logs should end up in Sentry so the "normal" installation only captures uncaptured exceptions.

bruno-garcia: Would you say it's best we add the log installation on a separate page? Doesn't seem to be clear to users landing here what path to take and if we prefer they go with unhandled exceptions only, perhaps that's all we should keep here

stayallive: It is currently on a [separate page](https://docs.sentry.io/platforms/php/guides/laravel/usage/#log-channels) but @sykezz found that too hidden and suggested this. I'm not a 100% sure what is the best solution here. 

Having 2 instructions might be a little adding more confusion that not having it. For the index page especially since doing both means you get every exception duplicated which of course not something we want to have users do by accident.

Maybe it would be better to add a new section under [customization](https://docs.sentry.io/platforms/php/guides/laravel/usage/#customization) linking to the [log channels page](https://docs.sentry.io/platforms/php/guides/laravel/usage/#log-channels) with a little excerpt that that is also a option?

bruno-garcia: One of the principles of this page is to keep it as short and simple as possible. The downside is that we're hiding some possibilities and complexities from the user, but ideally this can be introduced at the very end with "next steps". What we want to avoid is having someone load this page and 'decide to take this later because no time to read a wall of text. Not to say this PR gets this in that place but this is the principal we try to follow.

That said, I'll defer to @imatwawana on what's best to do here. Thank you for clarifying everything🙏

sykezz: Thanks everyone for all the inputs. The reason I see log channel as an alternative setup is because if you choose log channel, you'd have to remove the code from the Exception Handler. With log channel approach, you've the option to push info and warnings into Sentry as well. It's like a little hidden feature :smiley:
Currently log channel setup is under [Usage->Customization->Log Channels](https://docs.sentry.io/platforms/php/guides/laravel/usage/#log-channels) and I find it a bit hidden. Also, doesn't it sound more related to [configuration](https://docs.sentry.io/platforms/php/guides/laravel/configuration/) rather than [usage customization](https://docs.sentry.io/platforms/php/guides/laravel/usage/#customization)? :sweat_smile:

I believe someone familiar with Laravel would be able to understand what the two approaches means, and would be able to choose. But, I agree that installation step should be short and less complicated. So like @stayallive suggested, how about a little excerpt on the installation page with the link to log channel section?

imatwawana: Thanks, everyone! I'm going to add another suggestion, what about adding the page here:

stayallive: That is not a bad idea I think 👍

imatwawana: @sykezz, is this a change you feel comfortable/have time to make? If not, I'll spin up a new PR to make the changes (unfortunately I can't pull your branch in to do it on the same PR).

sykezz: @imatwawana yeah that sounds good. I'll push the changes soon.

sykezz: @imatwawana Hi. I've had some trouble adding that page under there. I don't think the current structure allows me to do that. Any advice?

imatwawana: As it turns out, I don't know how to make create that structure either. Sorry about that! For the sake of not dragging out this process any longer,  let's just stick with adding a little section in the Install part of the Getting Started page that points to the content where it currently exists. Does that make sense?

sykezz: @imatwawana Yup. How about a little note?

image

imatwawana: yep! looks good

sykezz: Pushed the changes and resolved conflicts :smile:

src/platforms/php/guides/laravel/index.mdx
```diff
@@ -34,6 +34,11 @@ public function register()
 }
 ```
 
+<Note>
+
+Alternatively, you can configure Sentry in your [Laravel Log Channel](usage/#log-channels), allowing to log `info` and `debug` as well.
```
imatwawana, line 39:
```suggestion
Alternatively, you can configure Sentry in your [Laravel Log Channel](usage/#log-channels), allowing you to log `info` and `debug` as well.
```

imatwawana:
This looks great @sykezz! Thank you so much for all your work!

sykezz:
Awesome! Thanks for all the feedback guys :smiley:

PR #4337, opened 2021-10-29
unity: note on missing sentry.framework

bruno-garcia:


jeffcampbellmakesgames:
This looks accurate 👍  There were also errors from the Sentry XCode post-processor not being able to copy the sentry frameworks to the XCode project. The exact error messages can be reproduced by importing Sentry directly as a local asset in the project and attempt to build to iOS.

bruno-garcia:
Thank you for reporting this @jeffcampbellmakesgames

PR #4340, opened 2021-11-01
GCP and ASP.NET SDKs are not strong named

bruno-garcia:
@lucas-zimerman


bruno-garcia:
Actually I'm wrong, there's no CLS Compliance but it's strong named

PR #4345, opened 2021-11-02
fix(nextjs): Add note about lack of Next 12 support

lobsterkatie:
There seem to be a number of changes in Next 12 which can break the SDK (depending on how you use it), so for the moment, let's make it clear that we're not expecting that combo to work out of the box.

### DIFF
--- a/src/includes/getting-started-primer/javascript.nextjs.mdx
+++ b/src/includes/getting-started-primer/javascript.nextjs.mdx
@@ -1,10 +1,10 @@
 <Note>
 
-Sentry's Next.js SDK enables automatic reporting of errors and exceptions.
+Currently, the Sentry Next.js SDK supports Next.js versions `10.0.8` - `11.1.3`. The SDK may work with some projects using Next.js 12, but it is not yet officially supported.
 
 </Note>
 
-Currently, the minimum Next.js supported version is `10.0.8`.
+The Sentry Next.js SDK automatically catches errors and collects performance data from your app.
 
 Features:
 
--- a/src/platforms/javascript/guides/nextjs/manual-setup.mdx
+++ b/src/platforms/javascript/guides/nextjs/manual-setup.mdx
@@ -182,7 +182,7 @@ module.exports = withSentryConfig(moduleExports, SentryWebpackPluginOptions);
 
 <Alert level="warning" title="Serverless Environments">
 
-`@sentry/nextjs` does not support configurations with the `serverless` target. To use the SDK in serverless environments, switch to using the `experimental-serverless-trace` target, which is [recommended by the Next.js maintainers](https://github.com/vercel/next.js/issues/20487#issuecomment-753884085).
+`@sentry/nextjs` does not support configurations with the `serverless` target. To use the SDK in serverless environments, switch to using the `experimental-serverless-trace` target, which is [recommended by the Next.js maintainers](https://github.com/vercel/next.js/issues/20487#issuecomment-753884085) for apps using Next.js 10 or Next.js 11. (The Sentry SDK does not yet officially support Next.js 12.)
 
 </Alert>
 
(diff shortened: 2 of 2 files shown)


src/includes/getting-started-primer/javascript.nextjs.mdx
```diff
@@ -1,10 +1,10 @@
 <Note>
 
-Sentry's Next.js SDK enables automatic reporting of errors and exceptions.
+Currently, the Sentry Next.js SDK supports Next.js versions `10.0.8` - `11.1.3`. The SDK may work with some projects using Next.js 12, but it is not yet officially supported.
 
 </Note>
 
-Currently, the minimum Next.js supported version is `10.0.8`.
+The Sentry Next.js SDK automatically catches errors and collects performance data from your app.
```
iker-barriocanal, line 7: I agree having what the SDK is able to do in a note doesn't look like the best option, but I don't think we should change it unless we also make the change for all the other platforms and products.

src/platforms/javascript/guides/nextjs/manual-setup.mdx
```diff
@@ -182,7 +182,7 @@ module.exports = withSentryConfig(moduleExports, SentryWebpackPluginOptions);
 
 <Alert level="warning" title="Serverless Environments">
 
-`@sentry/nextjs` does not support configurations with the `serverless` target. To use the SDK in serverless environments, switch to using the `experimental-serverless-trace` target, which is [recommended by the Next.js maintainers](https://github.com/vercel/next.js/issues/20487#issuecomment-753884085).
+`@sentry/nextjs` does not support configurations with the `serverless` target. To use the SDK in serverless environments, switch to using the `experimental-serverless-trace` target, which is [recommended by the Next.js maintainers](https://github.com/vercel/next.js/issues/20487#issuecomment-753884085) for apps using Next.js 10 or Next.js 11. (The Sentry SDK does not yet officially support Next.js 12.)
```
iker-barriocanal, line 185: This is redundant if we are adding the requirement on the getting started page.
```suggestion
`@sentry/nextjs` does not support configurations with the `serverless` target. To use the SDK in serverless environments, switch to using the `experimental-serverless-trace` target, which is [recommended by the Next.js maintainers](https://github.com/vercel/next.js/issues/20487#issuecomment-753884085) for apps using Next.js 10 or Next.js 11.
```

iker-barriocanal (changes requested):
Pushing back because of the format changes on `getting-started-primer`.

src/includes/getting-started-primer/javascript.nextjs.mdx
lobsterkatie: It's actually inconsistent - some frameworks have it (Angular, for example) and some don't (Gatsby, Electron, etc).

src/platforms/javascript/guides/nextjs/manual-setup.mdx
lobsterkatie: Yes, but I thought it worth mentioning here as well because it's directly relevant to this issue (since Next 12 does away with `target` all together), and because "do this for these versions" begs the question of what the case is for the remaining version.

(Also, Lee specifically asked me to fix this for the Next 12 launch, so I figured being extra clear here wouldn't be a bad thing.)

src/includes/getting-started-primer/javascript.nextjs.mdx
iker-barriocanal: Mmm... Then not blocking.

src/platforms/javascript/guides/nextjs/manual-setup.mdx
iker-barriocanal: I still rather having each piece of info in one place, it makes it easier to maintain. But not strongly opinionated in this context, so your call.

iker-barriocanal (approved):

src/platforms/javascript/guides/nextjs/manual-setup.mdx
lobsterkatie: I also don't feel strongly, and as a general rule, you're right - the fewer spots to update the better. I can take it out.

PR #4349, opened 2021-11-02
Sandbox btn

Neo-Zhixing:
This PR adds a Demo button to the navbar. When you click on it, it takes you to the Sentry sandbox.
image
image

Meanwhile, this PR also adds the `SandboxOnly` component so that it's available to use in mdx files also.

### DIFF
--- a/src/components/navbar.tsx
+++ b/src/components/navbar.tsx
@@ -6,7 +6,7 @@ import Search from "./search";
 import SmartLink from "./smartLink";
 
 import NavbarPlatformDropdown from "./navbarPlatformDropdown";
-import { getSandboxURL } from "./sandboxLink";
+import { getSandboxURL, SandboxOnly } from "./sandboxLink";
 
 type Props = {
   platforms?: string[];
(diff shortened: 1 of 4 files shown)


src/pages/index.tsx
```diff
@@ -74,6 +75,11 @@ const IndexPage = () => {
                   API
                 </SmartLink>
               </Nav.Item>
+              <Nav.Item>
+                <Nav.Link className="text-primary" onClick={() => {
+                  startSandbox();
```
scefali, line 80: @Neo-Zhixing we should replicate what we do on static sites to be able to turn off links to the sandbox while we upgrade

src/utils.ts
```diff
@@ -58,3 +58,60 @@ export const sortPages = (
...
+  }
+
+  if (trackingValue) url.searchParams.append('acceptedTracking', trackingValue);
+
+  if (scenario) url.searchParams.append('scenario', scenario);
+
+  if (projectSlug) url.searchParams.append('projectSlug', projectSlug);
+
+  if (errorType) url.searchParams.append('errorType', errorType);
+   url.searchParams.append('source', 'docs');
```
scefali, line 95: @Neo-Zhixing why does the tabbing look so weird here?

Neo-Zhixing: oh i thought there would be auto format but it doesn't exist... will fix.

src/components/navbar.tsx
```diff
@@ -25,6 +26,9 @@ export default ({ platforms }: Props): JSX.Element => {
               API
             </SmartLink>
           </Nav.Item>
+          <Nav.Item>
+            <Nav.Link className="text-primary" href={getSandboxURL().toString()}>Demo</Nav.Link>
```
scefali, line 30: @Neo-Zhixing why can't we use the same component as in the other PR?

Neo-Zhixing: Because that one always returns `<a>` so it's easier to use in mdx files.

scefali (approved):

PR #4353, opened 2021-11-03
ref(app-store-connect): Update docs

priscilawebdev:
we no longer need iTunes credentials to fetch dSYM files and all related code will be removed  in this PR https://github.com/getsentry/sentry/pull/29675

We are updating the docs to comply with the new changes

**Before:**

image

image

**After:**

image


src/platforms/apple/common/dsym.mdx
```diff
@@ -47,41 +47,23 @@ Free subscription plans are limited to one App Store Connect source per project,
 
 </Alert>
 
-Setting up App Store Connect currently requires two separate credentials.
-
-1. **App Store Connect API Key**:
+Setting up an App Store Connect custom repository requires only the App Store Connect API Key.
```
flub, line 50:
```suggestion
Setting up an App Store Connect custom repository requires an App Store Connect API Key.
```

src/platforms/apple/common/dsym.mdx
```diff
@@ -47,41 +47,23 @@ Free subscription plans are limited to one App Store Connect source per project,
...
-
-</Alert>
-
-These credentials will need to be verified by a two-factor authentication token that you can choose to have sent by SMS.
-
-**Re-authentication**
+**Invalid Credentials**
 
-The iTunes credentials and two-factor authentication are short-lived and need to be refreshed on a regular basis. Sentry will prompt users to re-authenticate in the **Issue Details** pages of a project, in the settings menu of a project, as well as in the "Custom Repositories" section of a project's "Debug Files" settings once their iTunes credentials expire.
+If your App Store Connect API key is revoked on the Apple side, the Sentry custom repository using this key's credentials will no longer be able to discover and fetch dSYM files and the following warnings will be displayed in the Senty's User Interface:
```
flub, line 56: I wonder how important it still is to show the screenshots given that this isn't a normal or often occurrence, I think just mentioning that it will stop working and you will get warnings in the UI might be enough.  But it doesn't do any harm though.

Independently I wonder if we should say something like "To update the credentials with new valid credentials click the 'configure' button"?

flub (approved):

src/platforms/apple/common/dsym.mdx
priscilawebdev: I also think that it does not do any harm...screenshots are always good to have :)

About the "config button", I think an extra sentence informing the user about it will be good.. I'll add it

PR #4354, opened 2021-11-03
Add SandboxLinks component

Neo-Zhixing:
This allows you to add links to the sandbox in docs.
Usage:

```
<SandboxLink scenario="releases">sandbox</SandboxLink>
```

This will automatically take you to the correct project based on the current project selection so there's usually no need to specify `projectSlug`.

### DIFF
--- a/src/components/sandboxLink.tsx
+++ b/src/components/sandboxLink.tsx
@@ -1,0 +1,70 @@
+import React from "react";
+import usePlatform from "./hooks/usePlatform";
+
+type Props = {
+    children: React.ReactNode;
+    scenario: string;
+    projectSlug: string;
+    errorType: string;
+    platform?: string;
+}
+
+/**
+ * 
+ * @param param0.scenario: One of the scenarios. Available scenarios include:
+ * performance, releases, alerts, discover, dashboards, projects,
+ * oneDiscoverQuery, oneIssue, oneBreadcrumb, oneStackTrace, oneTransaction,
+ * oneWebVitals, oneTransactionSummary, oneRelease
+ * @param param0.projectSlug:
+ * One of react, python, react-native, android, iOS
+ * @param param0.errorType: A string matching the title of the error.
+ * @returns URL to the sandbox start endpoint
+ */
+export function getSandboxURL({ scenario, projectSlug, errorType }: {
+  scenario?: string,
+  projectSlug?: string,
+  errorType?: string,
+} = {}) {
+  const url = new URL('https://try.sentry-demo.com/demo/start/');
+
+  if (scenario) url.searchParams.append('scenario', scenario);
+
+  if (projectSlug) url.searchParams.append('projectSlug', projectSlug);
+
+  if (errorType) url.searchParams.append('errorType', errorType);
+  url.searchParams.append('source', 'docs');
+
+  return url;
+};
+
+export const isSandboxHidden = () => process.env.GATSBY_HIDE_SANDBOX === '1';
+
+const SANDBOX_PLATFORMS: string[] = [
+    'react',
+    'python',
+    'react-native',
+    'android',
+    'ios',
+];
+
+const SANDBOX_PLATFORM_MAP: { [key: string]: string } = {
+    apple: 'ios',
+    javascript: 'react',
+    node: 'react',
+};
+
+export default function SandboxLink({ children, platform, ...params }: Props) {
+    if (isSandboxHidden()) {
+        return children;
+    }
+    const [currentPlatform] = usePlatform(platform);
+    if (!params.projectSlug) {
+        if (SANDBOX_PLATFORMS.includes(currentPlatform.key)) {
+            params.projectSlug = currentPlatform.key;
+        } else if (SANDBOX_PLATFORM_MAP[currentPlatform.key]) {
+            params.projectSlug = SANDBOX_PLATFORM_MAP[currentPlatform.key];
+        }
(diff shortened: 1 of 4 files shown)


src/components/sandboxLink.tsx
```diff
@@ -0,0 +1,70 @@
+import React from "react";
+import usePlatform from "./hooks/usePlatform";
+
+type Props = {
+    children: React.ReactNode;
+    scenario: string;
+    projectSlug: string;
+    errorType: string;
```
scefali, line 8: @Neo-Zhixing Most of these props should be optional, yea?

src/components/sandboxLink.tsx
```diff
@@ -0,0 +1,70 @@
...
+    platform?: string;
+}
+
+/**
+ * 
+ * @param param0.scenario: One of the scenarios. Available scenarios include:
+ * performance, releases, alerts, discover, dashboards, projects,
+ * oneDiscoverQuery, oneIssue, oneBreadcrumb, oneStackTrace, oneTransaction,
+ * oneWebVitals, oneTransactionSummary, oneRelease
```
scefali, line 17: @Neo-Zhixing maybe we should just enumerate the scenarios with types?

src/components/sandboxLink.tsx
Neo-Zhixing: right

src/components/sandboxLink.tsx
```diff
@@ -0,0 +1,86 @@
...
+    OneStackTrace = 'oneStackTrace',
+    OneTransaction = 'oneTransaction',
+    OneWebVitals = 'oneWebVitals',
+    OneTransactionSummary = 'oneTransactionSummary',
+    OneRelease = 'oneRelease',
+}
+
+type Props = {
+    children: React.ReactNode;
+    scenario?: Scenario;
```
scefali, line 24: @Neo-Zhixing Nit: you could have done something like:
```
const scenarios = ['performance', 'releases', ...] as const
...
scenario?: typeof scenarios[number]
```

src/components/sandboxLink.tsx
```diff
@@ -0,0 +1,86 @@
+import React from "react";
+import usePlatform from "./hooks/usePlatform";
+
+
+export enum Scenario {
```
scefali, line 5: @Neo-Zhixing why export this?

Neo-Zhixing: why not?

scefali: @Neo-Zhixing I don't see why we'd need to use this anywhere else

src/components/sandboxLink.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  if (errorType) url.searchParams.append('errorType', errorType);
+  url.searchParams.append('source', 'docs');
+
+  return url;
+};
+
+export const isSandboxHidden = () => process.env.GATSBY_HIDE_SANDBOX === '1';
+
+const SANDBOX_PLATFORMS: string[] = [
```
scefali, line 58: @Neo-Zhixing I would probably type this as:
```
const SANDBOX_PLATFORMS = [
    'react',
    'python',
    'react-native',
    'android',
    'ios',
] as const;
```
Not sure why you want to explicitly type it as `string[]`

scefali (approved):

src/components/sandboxLink.tsx
Neo-Zhixing: If you're calling `getSandboxURL` you'd need to have `Scenario`

scefali: @Neo-Zhixing I see, I think this might be because of how you are typing it. If you typed it like in my other [comment](https://github.com/getsentry/sentry-docs/pull/4354/files#r742279479), then you can just pass in `scenario` as literal string like `discover` and then you wouldn't need to import this type.

src/components/sandboxLink.tsx
Neo-Zhixing: Because then I have to cast it on line 77 but I guess it's fine

src/components/sandboxLink.tsx
```diff
@@ -0,0 +1,85 @@
...
+    'oneTransaction',
+    'oneWebVitals',
+    'oneTransactionSummary',
+    'oneRelease',
+  ] as const;
+
+type Props = {
+    children: React.ReactNode;
+    scenario?: typeof scenarios[number];
+    projectSlug?: string;
```
scefali, line 24: @Neo-Zhixing wondering if we should be more explicit with our types here too to prevent someone from using an invalid project

src/components/sandboxLink.tsx
```diff
@@ -0,0 +1,85 @@
...
+    errorType?: string;
+    platform?: string;
```
scefali, line 26: @Neo-Zhixing are we even using `platform`?

src/components/sandboxLink.tsx
Neo-Zhixing: In the end it doesn't matter, I don't think you have any type checking when using the component from a `mdx` file

src/components/sandboxLink.tsx
Neo-Zhixing: Yes, it's used as in the `usePlatform` calls, which allows you to specify the platform manually. This seems to be the standard practice in the docs repo.

PR #4356, opened 2021-11-03
Docs: Fix `MaxRequestBodySize` values

tomi-bigpi:
Update the values listed on this page for the `MaxRequestBodySize` option for `appsettings.json` to match those available in the `Sentry.Extensibility.RequestSize` enumeration (https://github.com/getsentry/sentry-dotnet/blob/main/src/Sentry/Extensibility/RequestSize.cs). The current value `never` is not a valid value and the values should start with an uppercase letter.


bruno-garcia (approved):
Thanks for fixing it

PR #4359, opened 2021-11-04
Add setSpan to custom instrumentation

rbisol:
In PHP the transaction is not automatically set to the Hub when it is created. I added the manual call to setSpan() that allow users to retrieve it later with getTransaction()

Tagging @stayallive for additional review as SDK mainteiner.

### DIFF
--- a/src/platforms/php/common/performance.mdx
+++ b/src/platforms/php/common/performance.mdx
@@ -67,6 +67,9 @@ $transactionContext->setName('External Call');
 $transactionContext->setOp('http.caller');
 $transaction = \Sentry\startTransaction($transactionContext);
 
+// Setting the transaction to the current Hub, this allows you to retrieve the current transaction
+SentrySdk::getCurrentHub()->setSpan($transaction);
+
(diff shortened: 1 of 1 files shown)


src/platforms/php/common/performance.mdx
```diff
@@ -67,6 +67,9 @@ $transactionContext->setName('External Call');
 $transactionContext->setOp('http.caller');
 $transaction = \Sentry\startTransaction($transactionContext);
 
+// Setting the transaction to the current Hub, this allows you to retrieve the current transaction
+SentrySdk::getCurrentHub()->setSpan($transaction);
```
stayallive, line 71: We namespace everything so you can copy-pasta this in working without imports.

```suggestion
// Setting the transaction to the current Hub allows you to retrieve it later
\Sentry\SentrySdk::getCurrentHub()->setSpan($transaction);
```

stayallive (approved):
Good change, this was missing indeed!

Just needs the namespace added but otherwise 👍

imatwawana (approved):

PR #4367, opened 2021-11-05
ref/apple primer

bruno-garcia:
I don't think the most important information about our product is the fact it uses Swizzling so I suggest we remove from the primer.
Instead, I added more information about features we added recently.

We could add a note to the foot of the Getting started page saying:

> * To learn more about what features use Swizzling and how to control that: click here.

### DIFF
--- a/src/includes/getting-started-primer/apple.mdx
+++ b/src/includes/getting-started-primer/apple.mdx
@@ -18,31 +18,9 @@
   - System events (battery level or state changed, memory warnings, device orientation changed, keyboard did show and did hide, screenshot taken)
   - Outgoing HTTP requests
 - [Release Health](/product/releases/health/) tracks crash free users and sessions
+- <PlatformLink to="/performance/instrumentation/">Measure the performance of your app</PlatformLink>
+  - Automatically measures rendering for UIViewControllers
+  - <PlatformLink to="/performance/instrumentation/??">Performance of HTTP requests</PlatformLink>
+  - <PlatformLink to="/performance/connect-services/">Distributed tracing</PlatformLink>
 - [Attachments](/platforms/apple/enriching-events/attachments/) enrich your event by storing additional files, such as config or log files
 - [User Feedback](/platforms/apple/enriching-events/user-feedback/) provides the ability to collect user information when an event occurs
-
-**Features using Swizzling:**
-
-The Cocoa SDK uses [swizzling](https://nshipster.com/method-swizzling/) to provide some features out of the box without boilerplate code. The following features use swizzling:
-
-
-
-Since Cocoa 7.5.0, you can opt out of swizzling using options. When you disable swizzling, the SDK disables the features above:
-
-```swift {tabTitle:Swift}
-import Sentry
-
-SentrySDK.start { options in
-    options.dsn = "___PUBLIC_DSN___"
-    options.enableSwizzling = false
-}
-```
-
-```objc {tabTitle:Objective-C}
-@import Sentry;
-
-[SentrySDK startWithConfigureOptions:^(SentryOptions *options) {
-    options.dsn = @"___PUBLIC_DSN___";
-    options.enableSwizzling = NO;
-}];
-```
--- a/src/includes/getting-started-primer/apple.mdx
+++ b/src/includes/getting-started-primer/apple.mdx
@@ -22,7 +22,7 @@
   - Rendering of UIViewControllers
   - Performance of HTTP requests
   - Distributed tracing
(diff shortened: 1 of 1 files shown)


imatwawana:
These links aren't rendering properly and I think it's the HTML-style tagging and markdown not playing well together in a bullet list:


You could try using HTML to build the entire bullet list rather than markdown and that might fix it.

src/includes/getting-started-primer/apple.mdx
```diff
@@ -18,31 +18,12 @@
...
+  - <PlatformLink to="/performance/instrumentation/automatic-instrumentation/#uiviewcontroller-instrumentation">Rendering of UIViewControllers</PlatformLink>
+  - <PlatformLink to="/performance/instrumentation/automatic-instrumentation/#http-instrumentation">Performance of HTTP requests</PlatformLink>
+  - <PlatformLink to="/performance/connect-services/">Distributed tracing</PlatformLink>
+  - Mobile Vitals
+    - <PlatformLink to="/performance/connect-services/#app-start-instrumentation">Cold and warm start</PlatformLink>
+    - <PlatformLink to="/performance/connect-services/#slow-and-frozen-frames">Slow and frozen frames</PlatformLink>
 - [Attachments](/platforms/apple/enriching-events/attachments/) enrich your event by storing additional files, such as config or log files
 - [User Feedback](/platforms/apple/enriching-events/user-feedback/) provides the ability to collect user information when an event occurs
-
-**Features using Swizzling:**
```
imatwawana, line 24: is the scope of this PR simply to remove the swizzling from this page and to have it be reinstated elsewhere on a separate PR?

imatwawana (approved):
I have one question about the scope of this PR and a note about how bullet points are rendering. But otherwise, looks good.

src/includes/getting-started-primer/apple.mdx
bruno-garcia: It's already live under configuration: https://docs.sentry.io/platforms/apple/configuration/swizzling/

In fact, that's where the "list of features above" is.

imatwawana: oh okay. so then it just seems like a note pointing to that is missing then? it seemed like you were suggesting we add one on this page.

bruno-garcia: I'm personally OK with not having any "Disclaimer". That's because we don't see the practice per-se as "evil" else we wouldn't be using it. But there are some customer who unfortunately have had a bad experience with features built using this features, mainly because there were bugs in the implementation, which caused these bugs.

Given the auto instrumentation is only possible with this, it shouldn't be a surprise that there's some 'magic' going on. And hence for those (who we expect to be a slim % of ppl) who don't want any automatic instrumentation, this feature list will sound more like a red flag than a good product feature list. For that slice of customers, we have advanced configuration documentation clarifying that indeed Swizzling is used, and what features rely on it, and finally how to disable it and go completely manual.

That said, it's safe to say the grand majority of customers just want to add the SDK and get the features. We target them here, with a simple page, quick to read that lists what the SDK can do and how to get started with the minimum setting. Further configuration requires further "next steps" anyway and IMHO that includes any Swizzling information and configuration.

To close, I believe it's noisy to write about Swizzling in the primer of the SDK. For those interested it's easily found when searching the docs and we can link to that page from parts of the doc, but I would rather we don't waste space and distract users of the main goal of this document with 'warnings' about swizzling.

bruno-garcia:
image


No idea how to get the bullet list with a proper platform link.

src/includes/getting-started-primer/apple.mdx
```diff
@@ -18,31 +18,12 @@
   - System events (battery level or state changed, memory warnings, device orientation changed, keyboard did show and did hide, screenshot taken)
   - Outgoing HTTP requests
 - [Release Health](/product/releases/health/) tracks crash free users and sessions
+- Automatic <PlatformLink to="/performance/instrumentation/">Performance Tracking</PlatformLink>
+  - Rendering of UIViewControllers
+  - Performance of HTTP requests
+  - Distributed tracing
+  - [Mobile Vitals](https://blog.sentry.io/2021/08/23/mobile-vitals-four-metrics-every-mobile-developer-should-care-about)
```
imatwawana, line 25: Only suggestion would be linking to our docs on mobile vitals vs a blog post.
https://docs.sentry.io/product/performance/mobile-vitals/

bruno-garcia:
```suggestion
  - [Mobile Vitals](https://docs.sentry.io/product/performance/mobile-vitals/)
```

PR #4377, opened 2021-11-10
fix(apple): Reverse parameters for scope.setTag

vincentisambart:
Hi,

Looking at https://docs.sentry.io/platforms/apple/guides/ios/enriching-events/tags/, it looked like the parameters key and value for `setTag` where reversed (probably due to them be in the reverse order in other languages).

### DIFF
--- a/src/includes/enriching-events/set-tag/apple.mdx
+++ b/src/includes/enriching-events/set-tag/apple.mdx
@@ -2,7 +2,7 @@
 import Sentry
 
 SentrySDK.configureScope { scope in
-    scope.setTag(value: "page_locale", key: "de-at")
+    scope.setTag(value: "de-at", key: "page_locale")
 }
 ```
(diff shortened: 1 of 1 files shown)


imatwawana:
> good catch @vincentisambart thanks @philipphofmann maybe have a look in all docs (iOS) if there are more cases like this? eg extra, data, etc, all are hashmaps

@marandaneto should we merge this in the meantime or wait for a review of other instances?

philipphofmann:
> good catch @vincentisambart thanks @philipphofmann maybe have a look in all docs (iOS) if there are more cases like this? eg extra, data, etc, all are hashmaps

I opened up another PR https://github.com/getsentry/sentry-docs/pull/4401

philipphofmann (approved):
Thanks @vincentisambart 👏

PR #4383, opened 2021-11-11
add note about Spark 3 support

AbhiPrasad:
We've got some users asking about this. Let's make sure we are clear that the worker integration is not support for Spark 3.

### DIFF
--- a/src/platforms/python/guides/pyspark/index.mdx
+++ b/src/platforms/python/guides/pyspark/index.mdx
@@ -33,7 +33,7 @@ if __name__ == "__main__":
 
 ### Worker
 
-The spark worker integration is supported for Spark 2.4.0 and above.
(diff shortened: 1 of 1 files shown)


src/platforms/python/guides/pyspark/index.mdx
```diff
@@ -33,7 +33,7 @@ if __name__ == "__main__":
 
 ### Worker
 
-The spark worker integration is supported for Spark 2.4.0 and above.
+The spark worker integration is supported for Spark 2.4.x. It is **not supported for Spark 3**.
```
imatwawana, line 36: I recommend either bolding the whole sentence "It is not supported..." or just bolding "not supported".

imatwawana (approved):
Just a recommendation for styling but otherwise looks good!

PR #4385, opened 2021-11-12
Add `sentry-graphql` docs.

maciejwalkowiak:
Related to https://github.com/getsentry/sentry-java/issues/1802

### DIFF
--- a/src/platforms/java/common/graphql.mdx
+++ b/src/platforms/java/common/graphql.mdx
@@ -19,16 +19,16 @@ Sentry GraphQL integration provides an integration with [GraphQL Java](https://g
 <dependency>
     <groupId>io.sentry</groupId>
     <artifactId>sentry-graphql</artifactId>
-    <version>{{ packages.version('sentry.java.openfeign', '5.4.0') }}</version>
+    <version>{{ packages.version('sentry.java.graphql', '5.4.0') }}</version>
 </dependency>
 ```
 
 ```groovy {tabTitle:Gradle}
-implementation 'io.sentry:sentry-graphql:{{ packages.version('sentry.java.jdbc', '5.4.0') }}'
+implementation 'io.sentry:sentry-graphql:{{ packages.version('sentry.java.graphql', '5.4.0') }}'
 ```
 
 ```scala {tabTitle: SBT}
-libraryDependencies += "io.sentry" % "sentry-graphql" % "{{ packages.version('sentry.java.jdbc', '5.4.0') }}"
+libraryDependencies += "io.sentry" % "sentry-graphql" % "{{ packages.version('sentry.java.graphql', '5.4.0') }}"
 ```
 
 For other dependency managers, check out the [central Maven repository](https://search.maven.org/artifact/io.sentry/sentry-grapql).
--- a/src/platforms/java/common/graphql.mdx
+++ b/src/platforms/java/common/graphql.mdx
@@ -31,7 +31,7 @@ implementation 'io.sentry:sentry-graphql:{{ packages.version('sentry.java.graphq
 libraryDependencies += "io.sentry" % "sentry-graphql" % "{{ packages.version('sentry.java.graphql', '5.4.0') }}"
 ```
 
-For other dependency managers, check out the [central Maven repository](https://search.maven.org/artifact/io.sentry/sentry-grapql).
+For other dependency managers, check out the [central Maven repository](https://search.maven.org/artifact/io.sentry/sentry-graphql).
 
 ## Capture Exceptions
 
@@ -40,9 +40,9 @@ For other dependency managers, check out the [central Maven repository](https://
 Create a new instance of `SentryDataFetcherExceptionHandler`, pass the delegate that handles the exception and configure `defaultDataFetcherExceptionHandler` when building a `GraphQL` instance:
 
 ```java
+import graphql.GraphQL;
 import graphql.execution.SimpleDataFetcherExceptionHandler;
 import io.sentry.graphql.SentryDataFetcherExceptionHandler;
-import graphql.GraphQL;
 
 SimpleDataFetcherExceptionHandler defaultExceptionHandler = new SimpleDataFetcherExceptionHandler();
 SentryDataFetcherExceptionHandler sentryExceptionHandler = new SentryDataFetcherExceptionHandler(defaultExceptionHandler);
@@ -51,6 +51,19 @@ GraphQL graphQL = GraphQL.newGraphQL(...)
     // ...
     .defaultDataFetcherExceptionHandler(sentryExceptionHandler)
     .build();
+```
+
+```kotlin
+import graphql.GraphQL
+import graphql.execution.SimpleDataFetcherExceptionHandler
+import io.sentry.graphql.SentryDataFetcherExceptionHandler
+
+val defaultExceptionHandler = SimpleDataFetcherExceptionHandler()
+val sentryExceptionHandler = SentryDataFetcherExceptionHandler(defaultExceptionHandler)
+
+val graphql = GraphQL.newGraphQL()
+  .defaultDataFetcherExceptionHandler(sentryExceptionHandler)
+  .build()
 ```
 
 ## Capture Performance
--- a/src/platforms/java/common/graphql.mdx
+++ b/src/platforms/java/common/graphql.mdx
@@ -1,0 +1,164 @@
+---
+title: GraphQL Java Integration
+sidebar_order: 7
+description: "Learn how to capture exceptions and the performance of queries executed with GraphQL Java ."
+notSupported:
+  - java.logback
+  - java.log4j2
+  - java.jul
+---
+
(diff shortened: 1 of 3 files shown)


src/platforms/java/common/graphql.mdx
```diff
@@ -0,0 +1,164 @@
...
+    <version>{{ packages.version('sentry.java.openfeign', '5.4.0') }}</version>
+</dependency>
+```
+
+```groovy {tabTitle:Gradle}
+implementation 'io.sentry:sentry-graphql:{{ packages.version('sentry.java.jdbc', '5.4.0') }}'
+```
+
+```scala {tabTitle: SBT}
+libraryDependencies += "io.sentry" % "sentry-graphql" % "{{ packages.version('sentry.java.jdbc', '5.4.0') }}"
```
marandaneto, line 31: ids of `packages.version` are wrong

src/platforms/java/common/graphql.mdx
```diff
@@ -0,0 +1,164 @@
...
+```
+
+For other dependency managers, check out the [central Maven repository](https://search.maven.org/artifact/io.sentry/sentry-grapql).
```
marandaneto, line 34:
```suggestion
For other dependency managers, check out the [central Maven repository](https://search.maven.org/artifact/io.sentry/sentry-graphql).
```

src/platforms/java/common/graphql.mdx
```diff
@@ -0,0 +1,164 @@
...
+
+## Capture Exceptions
+
+`SentryDataFetcherExceptionHandler` captures exception, sends it to Sentry and calls the delegate responsible for handling the exception.
+
+Create a new instance of `SentryDataFetcherExceptionHandler`, pass the delegate that handles the exception and configure `defaultDataFetcherExceptionHandler` when building a `GraphQL` instance:
+
+```java
```
marandaneto, line 42: what's about adding kotlin too

src/platforms/java/common/graphql.mdx
maciejwalkowiak: Good catch

marandaneto (commented):
left comments

src/platforms/java/common/graphql.mdx
maciejwalkowiak: Added

marandaneto (approved):

src/platforms/java/common/graphql.mdx
```diff
@@ -0,0 +1,177 @@
+---
+title: GraphQL Java Integration
+sidebar_order: 7
+description: "Learn how to capture exceptions and the performance of queries executed with GraphQL Java ."
```
imatwawana, line 4:
```suggestion
description: "Learn how to capture exceptions and about the performance of queries executed with GraphQL Java ."
```

src/platforms/java/common/graphql.mdx
```diff
@@ -0,0 +1,177 @@
...
+notSupported:
+  - java.logback
+  - java.log4j2
+  - java.jul
+---
+
+Sentry GraphQL integration provides an integration with [GraphQL Java](https://github.com/graphql-java/graphql-java/) through:
+
+- `SentryDataFetcherExceptionHandler` which captures exceptions thrown during data fetcher executions
```
imatwawana, line 13:
```suggestion
- `SentryDataFetcherExceptionHandler` which captures exceptions thrown during data fetcher executions.
```

src/platforms/java/common/graphql.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+```scala {tabTitle: SBT}
+libraryDependencies += "io.sentry" % "sentry-graphql" % "{{ packages.version('sentry.java.graphql', '5.4.0') }}"
+```
+
+For other dependency managers, check out the [central Maven repository](https://search.maven.org/artifact/io.sentry/sentry-graphql).
+
+## Capture Exceptions
+
+`SentryDataFetcherExceptionHandler` captures exception, sends it to Sentry and calls the delegate responsible for handling the exception.
```
imatwawana, line 38:
```suggestion
`SentryDataFetcherExceptionHandler` captures the exception, sends it to Sentry, and calls the delegate responsible for handling the exception.
```

src/platforms/java/common/graphql.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+Create a new instance of `SentryDataFetcherExceptionHandler`, pass the delegate that handles the exception and configure `defaultDataFetcherExceptionHandler` when building a `GraphQL` instance:
```
imatwawana, line 40:
```suggestion
Create a new instance of `SentryDataFetcherExceptionHandler`, pass the delegate that handles the exception, and configure `defaultDataFetcherExceptionHandler` when building a `GraphQL` instance:
```

imatwawana (commented):
A few minor wording changes. I'm wondering about the location of this. Should it be under Configuration like this: 
https://docs.sentry.io/platforms/android/configuration/integrations/fragment/
or these:
https://docs.sentry.io/platforms/javascript/configuration/integrations/

maciejwalkowiak:
Regarding the location - if we decide to move things under "Configuration -> Integrations", then I guess we should move other pages there too? Like this one: https://docs.sentry.io/platforms/java/guides/spring-boot/logging-frameworks/

I have no opinions here and will trust your decision.

marandaneto:
> Regarding the location - if we decide to move things under "Configuration -> Integrations", then I guess we should move other pages there too? Like this one: [docs.sentry.io/platforms/java/guides/spring-boot/logging-frameworks](https://docs.sentry.io/platforms/java/guides/spring-boot/logging-frameworks/)
> 
> I have no opinions here and will trust your decision.

I guess integrations that have the ability to init the SDK as well can be under "Guides".
When the integration is additional to another SDK/integration setup, could go under `integrations`.

sentry-graphql should go under integrations IMO, logback and log4j2 too.
logback and log4j2 should also be a guide because they also have the ability to be inited via Appender.

what do you think?

imatwawana:
> > Regarding the location - if we decide to move things under "Configuration -> Integrations", then I guess we should move other pages there too? Like this one: [docs.sentry.io/platforms/java/guides/spring-boot/logging-frameworks](https://docs.sentry.io/platforms/java/guides/spring-boot/logging-frameworks/)
> > I have no opinions here and will trust your decision.
> 
> I guess integrations that have the ability to init the SDK as well can be under "Guides". When the integration is additional to another SDK/integration setup, could go under `integrations`.
> 
> sentry-graphql should go under integrations IMO, logback and log4j2 too. logback and log4j2 should also be a guide because they also have the ability to be inited via Appender.
> 
> what do you think?

For logback and log4j2, do you mean that the content should show up as both a guide and as a configuration? That seems like a confusing navigation experience to me. 

As for the placement of this content, I just wanted to raise the question of where it should go. If we use this as our guide, does that clarify things at all:
https://docs.sentry.io/contributing/approach/sdk-docs/write-configuration/#integrations

marandaneto:
@imatwawana it's because the content of https://docs.sentry.io/platforms/java/guides/logback/ and https://docs.sentry.io/platforms/java/guides/spring-boot/logging-frameworks/#logback are slightly different, the same for log4j2.
or we merge them in a single page, but there's a difference when setting them up standalone or alongside Spring boot.

imatwawana:
> @imatwawana it's because the content of https://docs.sentry.io/platforms/java/guides/logback/ and https://docs.sentry.io/platforms/java/guides/spring-boot/logging-frameworks/#logback are slightly different, the same for log4j2. or we merge them in a single page, but there's a difference when setting them up standalone or alongside Spring boot. so what do you think? I believe it should go under integrations as suggested before. we can move logback and log4j2 if agreed on another PR, so we don't block this one

Ah okay. Apologies, I misunderstood. That makes sense to me. So to confirm we're all on the same page:
Move Sentry-graphql page to here: https://docs.sentry.io/platforms/java/configuration/, adding a folder called Integrations to put it in?

And agreed, we can deal with the logging frameworks page on a different PR.

marandaneto:
> > @imatwawana it's because the content of [docs.sentry.io/platforms/java/guides/logback](https://docs.sentry.io/platforms/java/guides/logback/) and [docs.sentry.io/platforms/java/guides/spring-boot/logging-frameworks/#logback](https://docs.sentry.io/platforms/java/guides/spring-boot/logging-frameworks/#logback) are slightly different, the same for log4j2. or we merge them in a single page, but there's a difference when setting them up standalone or alongside Spring boot. so what do you think? I believe it should go under integrations as suggested before. we can move logback and log4j2 if agreed on another PR, so we don't block this one
> 
> Ah okay. Apologies, I misunderstood. That makes sense to me. So to confirm we're all on the same page: Move Sentry-graphql page to here: [docs.sentry.io/platforms/java/configuration](https://docs.sentry.io/platforms/java/configuration/), adding a folder called Integrations to put it in?
> 
> And agreed, we can deal with the logging frameworks page on a different PR.

exactly.

maciejwalkowiak:
Ok, GrapQL moved under "Integrations". "Integrations" is hidden from logging frameworks guide, as it would display empty page.

PR #4400, opened 2021-11-16
Set custom releand and dist via environment variables

marandaneto:
resolves a lot of issues like https://github.com/getsentry/sentry-react-native/issues/914#issuecomment-963244589

### DIFF
--- a/src/platforms/react-native/sourcemaps.mdx
+++ b/src/platforms/react-native/sourcemaps.mdx
@@ -16,6 +16,19 @@ However, if you use custom values for your release other than the version includ
 
 For events sent from Sentry to correctly be attributed to a release and subsequently its source maps, **both the `release` and `dist` values will need to be set.** You can set these values in the call to `init`, as discussed in our [Releases & Health content](/platforms/react-native/configuration/releases/).
 
+<Note>
+
+Before going ahead and disabling the automatic source maps upload, you may set the `release` and `dist` as environment variables, `SENTRY_RELEASE` and `SENTRY_DIST` respectively. The script that does the automatic source maps upload will use those values instead the default ones and they will match the given `release` and `dist`.
+
+</Note>
+
+```bash
+export SENTRY_RELEASE="my-project-name@2.3.12"
+export SENTRY_DIST="52"
+```
+
+In case you've exported the environment variables and they match the given `release` and `dist` during the `Sentry.init` call, you can skip the next steps.
+
 ### 2. Disable the automatic source maps upload script
 
 When you set a custom release or manually upload source maps, you will need to disable the automatic source map upload script:
--- a/src/platforms/react-native/sourcemaps.mdx
+++ b/src/platforms/react-native/sourcemaps.mdx
@@ -18,7 +18,7 @@ For events sent from Sentry to correctly be attributed to a release and subseque
 
 <Note>
 
-Before disabling the automatic source maps upload, you can set the `release` and `dist` as environment variables, `SENTRY_RELEASE` and `SENTRY_DIST` respectively. The script that performs the automatic source maps upload will use those values instead the default ones, and they will match the given `release` and `dist`. This ensures that auto-generated release names are accepted.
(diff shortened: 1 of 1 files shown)


marandaneto:
cc @jennmueng

src/platforms/react-native/sourcemaps.mdx
```diff
@@ -16,6 +16,19 @@ However, if you use custom values for your release other than the version includ
 
 For events sent from Sentry to correctly be attributed to a release and subsequently its source maps, **both the `release` and `dist` values will need to be set.** You can set these values in the call to `init`, as discussed in our [Releases & Health content](/platforms/react-native/configuration/releases/).
 
+<Note>
+
+Before going ahead and disabling the automatic source maps upload, you may set the `release` and `dist` as environment variables, `SENTRY_RELEASE` and `SENTRY_DIST` respectively. The script that does the automatic source maps upload will use those values instead the default ones and they will match the given `release` and `dist`.
```
imatwawana, line 21: Should we add here the problem users are trying to avoid by doing this. i've suggested a final sentence, but please update for accuracy. 

```suggestion
Before disabling the automatic source maps upload, you can set the `release` and `dist` as environment variables, `SENTRY_RELEASE` and `SENTRY_DIST` respectively. The script that performs the automatic source maps upload will use those values instead the default ones, and they will match the given `release` and `dist`. This ensures that auto-generated release names are accepted.
```

src/platforms/react-native/sourcemaps.mdx
```diff
@@ -16,6 +16,19 @@ However, if you use custom values for your release other than the version includ
...
+
+</Note>
+
+```bash
+export SENTRY_RELEASE="my-project-name@2.3.12"
+export SENTRY_DIST="52"
+```
+
+In case you've exported the environment variables and they match the given `release` and `dist` during the `Sentry.init` call, you can skip the next steps.
```
imatwawana, line 30:
```suggestion
If you've exported the environment variables and they match the given `release` and `dist` during the `Sentry.init` call, you can skip the next steps.
```

imatwawana (approved):
Made a suggestion about adding some more information in the note, but otherwise LGTM!

src/platforms/react-native/sourcemaps.mdx
marandaneto: @imatwawana since it's not an auto-generated release anymore, I'd rather change to:

This ensures that the custom release and dist names are used.

wdyt?

imatwawana: I like it!

imatwawana: Oh maybe just add the word "values" or "variables" — whichever makes sense. So "This ensures that the given custom `release` and `dist` values are used.

marandaneto: values is a good option, thanks

src/platforms/react-native/sourcemaps.mdx
```diff
@@ -16,6 +16,19 @@ However, if you use custom values for your release other than the version includ
 
 For events sent from Sentry to correctly be attributed to a release and subsequently its source maps, **both the `release` and `dist` values will need to be set.** You can set these values in the call to `init`, as discussed in our [Releases & Health content](/platforms/react-native/configuration/releases/).
 
+<Note>
+
+Before disabling the automatic source maps upload, you can set the `release` and `dist` as environment variables, `SENTRY_RELEASE` and `SENTRY_DIST` respectively. The script that performs the automatic source maps upload will use those values instead the default ones, and they will match the given `release` and `dist`. This ensures that auto-generated release names are accepted.
```
marandaneto, line 21:
```suggestion
Before disabling the automatic source maps upload, you can set the `release` and `dist` as environment variables, `SENTRY_RELEASE` and `SENTRY_DIST` respectively. The script that performs the automatic source maps upload will use those values instead the default ones, and they will match the given `release` and `dist`. This ensures that the given custom `release` and `dist` values are used.
```

PR #4403, opened 2021-11-16
Note that SetDefaultLoggerName() is not available.

shane-kerr:


src/platforms/go/common/migration.mdx
```diff
@@ -79,6 +78,8 @@ sentry.Init(sentry.ClientOptions{
 })
 ```
 
+The `SetDefaultLoggerName()` method is not available in `sentry-go`. You can emulate it using a tag.
```
rhcarvalho, line 81: Good catch, indeed there's a confusion in the guide, `DebugWriter` != `SetDefaultLoggerName`.

This is the equivalent of `SetDefaultLoggerName`:

```go
sentry.Init(sentry.ClientOptions{
	BeforeSend: func(event *Event, hint *EventHint) *Event {
		event.Logger = "DefaultLoggerName"
		return event
	},
})
```

In `sentry-go` any event field can be updated with the same mechanism, removing the need for individual setters for every event field.

Read on for details.

---

## Logger

`Packet.Logger` in `raven-go` is `Event.Logger` in `sentry-go`. The network protocol has not changed. New fields have been added over time, but all the old fields still remain. Source:

- https://github.com/getsentry/raven-go/blob/5c24d5110e0e198d9ae16f1f3465366085001d92/client.go#L173
- https://github.com/getsentry/sentry-go/blob/cf6576f67123e401d130b900fd93aab7575e505e/interfaces.go#L184

## Event Processors

Alternative to `BeforeSend` above, a `Client` in `sentry-go` can also have an arbitrary number of "event processors" (`BeforeSend` is just a special case of an event processor good enough for most needs). Thus, `SetDefaultLoggerName` in `raven-go` could also become something like this in `sentry-go`:

```go
client.AddEventProcessor(func(event *Event, hint *EventHint) *Event {
	event.Logger = "DefaultLoggerName"
	return event
})
```

## DebugWriter

Note that `raven-go` has an [internal `debugLogger`](https://github.com/getsentry/raven-go/blob/5c24d5110e0e198d9ae16f1f3465366085001d92/client.go#L563), but it always writes to `os.Stdout` and cannot be configured. In `sentry-go`, the internal logger can write to any `io.Writer` as configured by `ClientOptions.DebugWriter`.

rhcarvalho (changes requested):

PR #4413, opened 2021-11-18
Add note for slow and frozen frames Android

marandaneto:


src/platforms/android/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -146,6 +146,12 @@ Cold and warm start are Mobile Vitals, which you can learn about in the [full do
 
 ### Slow and Frozen Frames
 
+<Note>
+
+Supported in Android OS version `7.0` and above.
```
philipphofmann, line 151: Shouldn't it be supported on?
```suggestion
Supported on Android OS version `7.0` and above.
```

philipphofmann (approved):
LGTM 🚀

PR #4423, opened 2021-11-22
Document enableNdkScopeSync option for RN

marandaneto:
similar to Android https://docs.sentry.io/platforms/android/enriching-events/scopes/#configuring-the-scope

### DIFF
--- a/src/includes/enriching-events/scopes/configure-scope/react-native.mdx
+++ b/src/includes/enriching-events/scopes/configure-scope/react-native.mdx
@@ -10,7 +10,7 @@ Sentry.configureScope(function(scope) {
 
 ### Scope Synchronization
 
-If you want to set context data in the JS layer, then receive that context data when a native (C/C++) crash happens, you need to synchronize the JS `Scope` with the native `Scope`. This feature was introduced in version `2.0.0` of the React-Native SDK and requires an opt-in:
+If you want to set context data in the JS layer and then receive that context data when a native (C/C++) crash happens, you need to synchronize the JS `Scope` with the native `Scope`. This feature was introduced in version `2.0.0` of the React-Native SDK and requires an opt-in:
 
 ```javascript {filename:App.js}
 import * as Sentry from "@sentry/react-native";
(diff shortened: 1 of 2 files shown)


src/includes/enriching-events/scopes/configure-scope/react-native.mdx
```diff
@@ -0,0 +1,22 @@
...
+  scope.setUser({
+    id: 42,
+    email: "john.doe@example.com",
+  });
+});
+```
+
+### Scope Synchronization
+
+If you want to set context data in the JS layer, then receive that context data when a native (C/C++) crash happens, you need to synchronize the JS `Scope` with the native `Scope`. This feature was introduced in version `2.0.0` of the React-Native SDK and requires an opt-in:
```
imatwawana, line 13:
```suggestion
If you want to set context data in the JS layer and then receive that context data when a native (C/C++) crash happens, you need to synchronize the JS `Scope` with the native `Scope`. This feature was introduced in version `2.0.0` of the React-Native SDK and requires an opt-in:
```

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -106,3 +106,7 @@ module.exports = {
 As this applies to **all** class and function names — not just React components — this could increase your JS bundle size.
 
 </Note>
+
+## Missing context on Android NDK events
```
imatwawana, line 110:
```suggestion
## Missing Context on Android NDK Events
```

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -106,3 +106,7 @@ module.exports = {
...
+
+In case it is missing context on Android NDK events as breadcrumbs, user, etc..., You've to enable the [Scope Synchronization](/platforms/react-native/enriching-events/scopes/#scope-synchronization) feature.
```
imatwawana, line 112:
```suggestion
If the context is missing on Android NDK events such as breadcrumbs, user, and so on, you need to enable the [Scope Synchronization](/platforms/react-native/enriching-events/scopes/#scope-synchronization) feature.
```

imatwawana (approved):
A few style and wording edits

marandaneto:
thanks @imatwawana

Issue #4421, opened 2021-11-21
Purpose and objective of Proguard documentation is unclear


alvindizon:
### Core or SDK?

Platform/SDK

### Which part? Which one?

Android SDK

### Description

I was looking at Sentry and noticed that our logs were obfuscated (as expected, since we're using Proguard/R8), and noticed a little helper saying that "some frames are minified". I was then led to the Proguard docs [here](https://docs.sentry.io/platforms/android/proguard/). Reading the whole section, I never got a full understanding of what the additional Proguard configuration is for--is it so that you'd see de-obfuscated traces at Sentry?

I also checked the "Next Steps" section of the Getting Started docs section and the description for Proguard wasn't clear as to what it's purpose is and doesn't sound correct--"`Learn about using ProGuard to upload the ProGuard/R8 mappings, source bundle, and native symbols Sentry.`", sounds like it should be saying "`and native symbols to Sentry`".

### Suggested Solution

Maybe I'm just dense but it would be clear for the Proguard section to start with something like "If you want to see de-obfuscated logs configure the following". Also, since most users would have their apps to shrunk/minified/obfuscated the info for Proguard configuration should be more prominent--at the very least the basic Proguard configuration should be described in Getting Started.

getsentry-release:
Routing to @getsentry/team-mobile for [triage](https://develop.sentry.dev/processing-tickets/#3-triage). ⏲️

marandaneto:
I agree, thanks for reporting @alvindizon , we don't really state the problem that the plugin/sentry is aiming to solve
@imatwawana can you help us with that?

https://github.com/getsentry/sentry-docs/pull/4425 this PR is actually addressing something similar to that, for RN Apps, that also require the plugin setup

romtsn:
Closing as we made the suggested improvements to the docs

PR #4434, opened 2021-11-23
added intro text to proguard docs

imatwawana:
Adds more intro text to ProGuard page
Fixes typo in page description
Addresses (in part) [issue 4421](https://github.com/getsentry/sentry-docs/issues/4421)

### DIFF
--- a/src/platforms/android/common/proguard.mdx
+++ b/src/platforms/android/common/proguard.mdx
@@ -1,7 +1,7 @@
 ---
 title: ProGuard
 sidebar_order: 100
-description: "Learn about using ProGuard to upload the ProGuard/R8 mappings, source bundle, and native symbols to Sentry."
+description: "Learn about using the Sentry Gradle Plugin or sentry-cli to upload ProGuard/R8 mappings, source bundle, and native symbols to Sentry."
 ---
 
 If you want to see de-obfuscated stack traces, you'll need to use ProGuard with Sentry. To do so, upload the ProGuard mapping files by either the recommended method of using our Gradle integration or manually by using [sentry-cli](/product/cli/dif/#proguard-mapping-upload).
--- a/src/platforms/android/common/proguard.mdx
+++ b/src/platforms/android/common/proguard.mdx
@@ -4,7 +4,7 @@ sidebar_order: 100
 description: "Learn about using ProGuard to upload the ProGuard/R8 mappings, source bundle, and native symbols to Sentry."
 ---
 
-If you want to see de-obfuscated logs, you'll need to to use ProGuard with Sentry. To do so, upload the ProGuard mapping files by either the recommended method of using our Gradle integration or manually by using [sentry-cli](/product/cli/dif/#proguard-mapping-upload).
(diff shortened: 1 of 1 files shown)


src/platforms/android/common/proguard.mdx
```diff
@@ -1,10 +1,10 @@
 ---
 title: ProGuard
 sidebar_order: 100
-description: "Learn about using ProGuard to upload the ProGuard/R8 mappings, source bundle, and native symbols Sentry."
+description: "Learn about using ProGuard to upload the ProGuard/R8 mappings, source bundle, and native symbols to Sentry."
```
marandaneto, line 4:
```suggestion
description: "Learn about using the Sentry Gradle Plugin or sentry-cli for providing debug symbols to Sentry."
```

what's about this? similar to https://github.com/getsentry/sentry-docs/edit/master/src/platforms/react-native/upload-debug.mdx

src/platforms/android/common/proguard.mdx
```diff
@@ -1,10 +1,10 @@
...
 ---
 
-To use ProGuard with Sentry, upload the ProGuard mapping files by either the recommended method of using our Gradle integration or manually by using [sentry-cli](/product/cli/dif/#proguard-mapping-upload).
+If you want to see de-obfuscated logs, you'll need to to use ProGuard with Sentry. To do so, upload the ProGuard mapping files by either the recommended method of using our Gradle integration or manually by using [sentry-cli](/product/cli/dif/#proguard-mapping-upload).
```
marandaneto, line 7:
```suggestion
If you want to see de-obfuscated stack traces, you'll need to use ProGuard with Sentry. To do so, upload the ProGuard mapping files by either the recommended method of using our Gradle integration or manually by using [sentry-cli](/product/cli/dif/#proguard-mapping-upload).
```

src/platforms/android/common/proguard.mdx
marandaneto: is the description shown somewhere? I don't see in the docs page actually.

marandaneto:
@imatwawana the very same intro could fit better the Android wizard, See https://github.com/getsentry/sentry-docs/blob/master/src/wizard/android/index.md?plain=1#L127

src/platforms/android/common/proguard.mdx
imatwawana: It seems like two pages with different content shouldn't have the exact same description, but if you think it's reasonably accurate, then let's go with it. The complaint about the description from the user was just about the typo. What about:
```suggestion
description: "Learn about using the Sentry Gradle Plugin or sentry-cli to upload the ProGuard/R8 mappings, source bundle, and native symbols to Sentry."
```

imatwawana: > 

No, it's only in the metadata of the page.

imatwawana: > is the description shown somewhere? I don't see in the docs page actually.

No, it's only in the metadata of the page.

PR #4445, opened 2021-11-29
Add backtrace to options for `capture_message`

sidk:
It looks like `backtrace` is supported as per https://github.com/getsentry/sentry-ruby/pull/1550


imatwawana:
@st0012 - Can you take a look at this for accuracy, please?

st0012 (approved):
Yeah this options is now available. Thanks for the update!

PR #4447, opened 2021-11-30
Add note for App start Android

marandaneto:


romtsn (approved):

src/platforms/android/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -134,6 +134,12 @@ The SDK sets the Span operation to `app.start.cold` for *Cold start_* and `app.s
 
 The SDK uses the `SentryPerformanceProvider` (ContentProvider) creation time as the beginning of the app start and the first `Activity#onResume` call as the end.
 
+<Note>
+
+The App Start Instrumentation is only measured if the Process is of the importance [RunningAppProcessInfo.IMPORTANCE_FOREGROUND](https://developer.android.com/reference/android/app/ActivityManager.RunningAppProcessInfo#IMPORTANCE_FOREGROUND), that means the process is running the foreground UI.
```
marandaneto, line 139: maybe worth mentioning that when the Process is opened via background work, it's not measured, since it's not running the foreground UI? or is it obvious?

romtsn: Yeah, good idea. Maybe also give an example of the notification case?

philipphofmann, line 139:
```suggestion
The App Start Instrumentation is only measured if the process is of the importance [RunningAppProcessInfo.IMPORTANCE_FOREGROUND](https://developer.android.com/reference/android/app/ActivityManager.RunningAppProcessInfo#IMPORTANCE_FOREGROUND), which means the process is running the foreground UI.
```

src/platforms/android/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -134,6 +134,12 @@ The SDK sets the Span operation to `app.start.cold` for *Cold start_* and `app.s
 
 The SDK uses the `SentryPerformanceProvider` (ContentProvider) creation time as the beginning of the app start and the first `Activity#onResume` call as the end.
 
+<Note>
+
+The App Start Instrumentation is only measured if the process is of the importance [RunningAppProcessInfo.IMPORTANCE_FOREGROUND](https://developer.android.com/reference/android/app/ActivityManager.RunningAppProcessInfo#IMPORTANCE_FOREGROUND), which means the process is running the foreground UI.
```
imatwawana, line 139: Is the app start instrumentation the actual thing being measured here? If so, then this is fine. If not, then you might need to clarify what is being measured. Or maybe "measured" isn't the right word.

imatwawana (approved):
One question, but otherwise, LGTM!

src/platforms/android/common/performance/instrumentation/automatic-instrumentation.mdx
marandaneto: the App start is a "measurement" and it's measured since it's calculated from a start and end timestamp.
eg App start = 0.5s
I'm open for suggestions tho :)

imatwawana: Ah okay. So then is this more accurate:
```suggestion
The app start is only measured if the process is of the importance [RunningAppProcessInfo.IMPORTANCE_FOREGROUND](https://developer.android.com/reference/android/app/ActivityManager.RunningAppProcessInfo#IMPORTANCE_FOREGROUND), which means the process is running the foreground UI.
```

marandaneto: thanks as always @imatwawana

PR #4457, opened 2021-12-01
[React Native] Fix integration example for shouldCreateSpanForRequest

Abbondanzo:
The documentation for the [`shouldCreateSpanForRequest`](https://docs.sentry.io/platforms/react-native/performance/instrumentation/automatic-instrumentation/#shouldcreatespanforrequest) section shows the `BrowserTracing` integration, which may lead to confusion (as seen in the screenshot below). This PR adds a snippet for the `react-native` platform so the correct integration shows

image

### DIFF
--- a/src/includes/performance/filter-span-example/react-native.mdx
+++ b/src/includes/performance/filter-span-example/react-native.mdx
@@ -1,0 +1,13 @@
+```javascript
+Sentry.init({
+  // ...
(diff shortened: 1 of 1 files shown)


marandaneto:
good catch @Abbondanzo thanks for doing this

marandaneto (approved):

PR #4464, opened 2021-12-03
Fix typo in example code

set0gut1:


untitaker (approved):

untitaker (changes requested):
actually can you change the comment above to match the rule

Ignoring all sourcefiles isn't a very realistic example. I think `.gen.c` is supposed to be an example where generated C source code is excluded from grouping.

set0gut1:
@untitaker Thanks for the review. I have fixed it.

set0gut1:
@untitaker Thank you for review. I have fixed it.

untitaker (approved):

PR #4467, opened 2021-12-06
Add note for Hermes and RAM Bundles

marandaneto:
Adding note for not supported Hermes and RAM Bundles https://reactnative.dev/docs/ram-bundles-inline-requires

### DIFF
--- a/src/platforms/react-native/manual-setup/hermes.mdx
+++ b/src/platforms/react-native/manual-setup/hermes.mdx
@@ -16,6 +16,12 @@ For Sentry open source, self-hosted users, the minimum version required is [f073
 
 Once you have the minimum version of the SDK, Sentry provides the standard integration as described in the [React Native Sentry documentation](/platforms/react-native/).
 
+<Note>
+
+If you are using [Hermes](/platforms/react-native/manual-setup/hermes/), you should not have [RAM bundles](/platforms/react-native/manual-setup/ram-bundles/) feature enabled. In Hermes, when loading the bytecode, `mmap` ensures that the entire file is not loaded. Using Hermes with RAM bundles might lead to issues, because those mechanisms are not compatible with each other.
+
+</Note>
+
 ## Source Maps
 
 If you do not need custom sourcemaps, the `sentry.gradle` build step fully supports Hermes source maps.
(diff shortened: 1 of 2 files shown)


src/platforms/react-native/manual-setup/hermes.mdx
```diff
@@ -16,6 +16,12 @@ For Sentry open source, self-hosted users, the minimum version required is [f073
 
 Once you have the minimum version of the SDK, Sentry provides the standard integration as described in the [React Native Sentry documentation](/platforms/react-native/).
 
+<Note>
+
+If you are using Hermes, you should not have [RAM bundles](/platforms/react-native/manual-setup/ram-bundles/) feature enabled. In Hermes, when loading the bytecode, `mmap` ensures that the entire file is not loaded. Using Hermes with RAM bundles might lead to issues, because those mechanisms are not compatible with each other.
```
imatwawana, line 21:
```suggestion
If you are using Hermes, you should not have the [RAM bundles](/platforms/react-native/manual-setup/ram-bundles/) feature enabled. In Hermes, when loading the bytecode, `mmap` ensures that the entire file is not loaded. Using Hermes with RAM bundles might lead to issues, because those mechanisms are not compatible with each other.
```

src/platforms/react-native/manual-setup/ram-bundles.mdx
```diff
@@ -12,6 +12,12 @@ All the existing RAM bundle formats are explained in detail in the [Metro Bundle
 
 ## Enabling RAM Bundles
 
+<Note>
+
+If you are using [Hermes](/platforms/react-native/manual-setup/hermes/), you should not have RAM bundles feature enabled. In Hermes, when loading the bytecode, `mmap` ensures that the entire file is not loaded. Using Hermes with RAM bundles might lead to issues, because those mechanisms are not compatible with each other.
```
imatwawana, line 17:
```suggestion
If you are using [Hermes](/platforms/react-native/manual-setup/hermes/), you should not have the RAM bundles feature enabled. In Hermes, when loading the bytecode, `mmap` ensures that the entire file is not loaded. Using Hermes with RAM bundles might lead to issues, because those mechanisms are not compatible with each other.
```

PR #4470, opened 2021-12-06
fix(unity): Attachment upload sample

bitsandfoxes:
Came up here: https://github.com/getsentry/sentry-unity/issues/448
The Unity SDK has the Global Hub Mode set by default which prevents `WithScope` from working as it relies on `PushScope`.


bruno-garcia (approved):

bruno-garcia:
Thank you Stefan

PR #4474, opened 2021-12-08
Update BrowserTracing import for tree shaking

AbhiPrasad:
In https://github.com/getsentry/sentry-javascript/pull/4204 we updated the `BrowserTracing` import in the JS SDK to be exported individually. We want users to use this individual import so that the other integrations are treeshaken. This patch updates our documentation to import `BrowserTracing` directly from `@sentry/tracing` instead of through the `Integrations` object, which hopefully should lead to a bundle size reduction for people.


lobsterkatie (commented):
LGTM.

I think you missed one, though: https://sentry-docs-git-abhi-browser-tracing-docs.sentry.dev/platforms/javascript/performance/instrumentation/automatic-instrumentation/#tracingorigins

PR #4476, opened 2021-12-08
fix: Fix typo for cancelled span status

dashed:


src/docs/product/performance/metrics.mdx
```diff
@@ -24,7 +24,7 @@ Configure what a satisfactory response time threshold (ms) is for Apdex in **Set
 
 ## Failure Rate
 
-`failure_rate()` indicates the percentage of unsuccessful transactions. Sentry treats transactions with a status other than “ok,” “canceled,” and “unknown” as failures. For more details, see a [list of possible status values](https://develop.sentry.dev/sdk/event-payloads/span/).
+`failure_rate()` indicates the percentage of unsuccessful transactions. Sentry treats transactions with a status other than “ok,” “cancelled,” and “unknown” as failures. For more details, see a [list of possible status values](https://develop.sentry.dev/sdk/event-payloads/span/).
```
imatwawana, line 27: Are we quoting this from somewhere that's using this spelling? If not, then leave it as is since we use American spellings. If so, then it's fine to merge this.

imatwawana (commented):
Are we quoting this from somewhere that's using this spelling? If not, then leave it as is since we use American spellings. If so, then it's fine to merge this.

PR #4485, opened 2021-12-10
Update triage link to point to triage section

silent1mezzo:
When starting to look into Issues the triage links us to the Issue Details section which doesn't mention anything about triaging an issue.


imatwawana (approved):
good catch!

PR #4497, opened 2021-12-15
Update troubleshooting.mdx

georgekallilas:
Add a link which explains how to bundle the source code.

### DIFF
--- a/src/platforms/go/common/troubleshooting.mdx
+++ b/src/platforms/go/common/troubleshooting.mdx
@@ -20,4 +20,4 @@ Since the Go SDK tries to read the source files from your local disk, it's possi
 
 This is common in Go because you can compile the binary, and then have no more need for the source code. Without access to the source code, however, we can’t map anything back, and we don’t support uploading Go source code with a release to stitch this data on.
 
-You can see [in the Serverless section](serverless/#source-context) how to bundle the source code with the binary, which applies to other forms of deployment as well.
+You can see [in the Serverless section](/platforms/go/serverless/#source-context) how to bundle the source code with the binary, which applies to other forms of deployment as well.
--- a/src/platforms/go/common/troubleshooting.mdx
+++ b/src/platforms/go/common/troubleshooting.mdx
@@ -20,4 +20,4 @@ Since the Go SDK tries to read the source files from your local disk, it's possi
 
 This is common in Go because you can compile the binary, and then have no more need for the source code. Without access to the source code, however, we can’t map anything back, and we don’t support uploading Go source code with a release to stitch this data on.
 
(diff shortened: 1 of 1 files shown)


src/platforms/go/common/troubleshooting.mdx
```diff
@@ -19,3 +19,5 @@ supported:
 Since the Go SDK tries to read the source files from your local disk, it's possible that you’re compiling your binary and deploying it in a manner that makes the source code inaccessible. 
 
 This is common in Go because you can compile the binary, and then have no more need for the source code. Without access to the source code, however, we can’t map anything back, and we don’t support uploading Go source code with a release to stitch this data on.
+
+You can see [here](https://docs.sentry.io/platforms/go/serverless/#source-context) how to bunde the source code with the binary.
```
rhcarvalho, line 23:
```suggestion
You can see [in the Serverless section](serverless/#source-context) how to bundle the source code with the binary, which applies to other forms of deployment as well.
```

Just some ideas / suggestions:

- relative link will work during deployment preview and in production
- add note about the idea being applicable outside of Serverless (eventually we need to better organize the docs to avoid this potential for confusion)

rhcarvalho (approved):
thanks @georgekallilas 👍 

cc @imatwawana

src/platforms/go/common/troubleshooting.mdx
imatwawana: I agree with the suggestions from @rhcarvalho. I'd say apply those and you're good to go!

imatwawana (approved):

src/platforms/go/common/troubleshooting.mdx
```diff
@@ -20,4 +20,4 @@ Since the Go SDK tries to read the source files from your local disk, it's possi
 
 This is common in Go because you can compile the binary, and then have no more need for the source code. Without access to the source code, however, we can’t map anything back, and we don’t support uploading Go source code with a release to stitch this data on.
 
-You can see [here](https://docs.sentry.io/platforms/go/serverless/#source-context) how to bunde the source code with the binary.
+You can see [in the Serverless section](serverless/#source-context) how to bundle the source code with the binary, which applies to other forms of deployment as well.
```
rhcarvalho, line 23: Oh, the relative link here doesn't work for the guides. Probably for now we go with 

```suggestion
You can see [in the Serverless section](/platforms/go/serverless/#source-context) how to bundle the source code with the binary, which applies to other forms of deployment as well.
```

which will work for previews and prod, and from a guide will link out of the guide and into the Serverless section.

This really needs some re-organizing :)

PR #4511, opened 2021-12-17
react-native: perf auto instr

bruno-garcia:


src/platforms/react-native/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -22,6 +22,12 @@ export default Sentry.wrap(App);
 
 ### Enable Routing Instrumentation
 
+<Note>
+
+When no routing instrumentation is used, a transaction for `App Start` is automatically captured. That stops being sent when one of the routing integrations below is added. That information lives as a Span in a transaction captured by the routing instrumentation instead.
```
bruno-garcia, line 27: @imatwawana It's likely I didn't explain this really well. Do you think this is OK?

imatwawana: I think it's pretty good. What about this?

```suggestion
When no routing instrumentation is used, a transaction for `App Start` is automatically captured. However, that transaction stops being sent when one of the routing integrations below is added. Instead, the `App Start` information is included as a span in a transaction captured by the routing instrumentation.
```

imatwawana (approved):

carantes:
@bruno-garcia @imatwawana You should also mention here something about the cold/warm start metrics. In my case, once I integrated the auto instrumentation with the react-native navigation, the `App Start` transaction stopped being sent, and these columns/information is now missing.

bruno-garcia:
> @bruno-garcia @imatwawana You should also mention here something about the cold/warm start metrics. In my case, once I integrated the auto instrumentation with the react-native navigation, the `App Start` transaction stopped being sent, and these columns/information is now missing.

That sounds like a bug. We'll look into it

PR #4512, opened 2021-12-17
fix role permissions info for integrations

imatwawana:
Fixes role permission notes for integrations
Adds notes where they are missing
Makes styling of role permission notes consistent


leeandher (approved):
much clearer, thanks so much 🙏🔥

PR #4516, opened 2021-12-20
fix: Add --force-foreground to Xcode Run Script

philipphofmann:
Add --force-foreground for calling sentry-cli in the Xcode run script
phase. This flag lets Xcode wait for the upload process to finish.

I had issues with uploading dSYMS with Xcode. Adding this flag fixed it for me and it also makes sense to wait until the dSYMS are uploaded.

### DIFF
--- a/src/platforms/apple/common/dsym.mdx
+++ b/src/platforms/apple/common/dsym.mdx
@@ -197,7 +197,7 @@ fi
 ${DWARF_DSYM_FOLDER_PATH}/${DWARF_DSYM_FILE_NAME}/Contents/Resources/DWARF/${TARGET_NAME}
 ```
 
-If the upload never completes you can try to pass `--force-foreground` to the Sentry CLI command in the build script. This is possibly due to a bug [we are investigating](https://github.com/getsentry/sentry-react-native/issues/1424).
+If the upload never completes you can try to pass `--force-foreground` to the Sentry CLI command in the build script. This is possibly due to a bug [we are investigating](https://github.com/getsentry/sentry-cli/issues/932).
 
 <Alert level="" title="On Prem">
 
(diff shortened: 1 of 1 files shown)


src/platforms/apple/common/dsym.mdx
```diff
@@ -179,7 +179,7 @@ if which sentry-cli >/dev/null; then
 export SENTRY_ORG=___ORG_SLUG___
 export SENTRY_PROJECT=___PROJECT_SLUG___
 export SENTRY_AUTH_TOKEN=YOUR_AUTH_TOKEN
-ERROR=$(sentry-cli upload-dif "$DWARF_DSYM_FOLDER_PATH" 2>&1 >/dev/null)
+ERROR=$(sentry-cli upload-dif --force-foreground "$DWARF_DSYM_FOLDER_PATH" 2>&1 >/dev/null)
```
marandaneto, line 182: We currently have an issue on RN as well, See https://docs.sentry.io/platforms/react-native/troubleshooting/#ios-build-script-failed
@jan-auer was supposed to investigate that on sentry-cli, @jan-auer do you have any updates on that?

philipphofmann: Thanks for the info. I think it even makes sense to add this flag if we fix the bug as Xcode and then wait until the upload is finished.

marandaneto: That has the side effect of slowing down every build release, that's why it's not the default behavior.
Think about slow connection and large apps that spit out hundreds of MBs of symbols.

philipphofmann: Better to mention below then?

src/platforms/apple/common/dsym.mdx
```diff
@@ -197,6 +197,8 @@ fi
 ${DWARF_DSYM_FOLDER_PATH}/${DWARF_DSYM_FILE_NAME}/Contents/Resources/DWARF/${TARGET_NAME}
 ```
 
+If the upload never completes you can try to pass `--force-foreground` to the Sentry CLI command in the build script. This is possibly due to a bug [we are investigating](https://github.com/getsentry/sentry-react-native/issues/1424).
```
marandaneto, line 200:
```suggestion
If the upload never completes you can try to pass `--force-foreground` to the Sentry CLI command in the build script. This is possibly due to a bug [we are investigating](https://github.com/getsentry/sentry-cli/issues/932).
```

I'd rather link to the Sentry CLI tho.

marandaneto (approved):
Left a suggestion, LGTM.

PR #4517, opened 2021-12-20
AndroidX support for RN apps

marandaneto:
Similar to https://docs.sentry.io/platforms/android/performance/instrumentation/automatic-instrumentation/#androidx-support

### DIFF
--- a/src/platforms/react-native/performance/instrumentation/automatic-instrumentation.mdx
+++ b/src/platforms/react-native/performance/instrumentation/automatic-instrumentation.mdx
@@ -308,6 +308,20 @@ Unresponsive UI and animation hitches annoy users and degrade the user experienc
 
 Slow and frozen frames are Mobile Vitals, which you can learn about in the [full documentation](/product/performance/mobile-vitals).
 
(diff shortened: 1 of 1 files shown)


src/platforms/react-native/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -308,6 +308,20 @@ Unresponsive UI and animation hitches annoy users and degrade the user experienc
...
+
+We check for availability at runtime, so if you're not using `androidx.core`, you can remove it from Sentry's transitive dependencies.
+
+```Groovy
+api ('io.sentry:sentry-android:{{ packages.version('sentry.java.android', '4.2.0') }}') {
+    exclude group: 'androidx.core', module: 'core'
+}
+```
+
+Note that by removing this transitive dependency, slow and frozen frames won't be reported.
```
imatwawana, line 323:
```suggestion
Note that if you remove this transitive dependency, slow and frozen frames won't be reported.
```

PR #4518, opened 2021-12-20
fix: Split consecutive code blocks in Go migration guide

rhcarvalho:
Similar to #2040.

Affected page https://docs.sentry.io/platforms/go/migration/#integrations

# Before

image

# After

image

### DIFF
--- a/src/platforms/go/common/migration.mdx
+++ b/src/platforms/go/common/migration.mdx
@@ -386,9 +386,13 @@ sentry-go
 go get github.com/getsentry/sentry-go/http
 ```
 
+<Break />
+
 ```go
 import sentryhttp "github.com/getsentry/sentry-go/http"
 ```
+
+<Break />
 
 ```go
 sentryHandler := sentryhttp.New(sentryhttp.Options{


src/platforms/go/common/migration.mdx
```diff
@@ -386,10 +386,14 @@ sentry-go
 go get github.com/getsentry/sentry-go/http
 ```
 
+<Break />
+
 ```go
 import sentryhttp "github.com/getsentry/sentry-go/http"
 ```
 
+<Break />
```
iker-barriocanal, line 395: An alternative that might be more clear is to set the full examples, by duplicating the `import` statement and having one of the options in a different tab. I don't know which option is more clear and easier to understand, so your call.

iker-barriocanal (approved):

src/platforms/go/common/migration.mdx
rhcarvalho: I still consider the migration guide needs a rewrite to be more approachable.

Instead of a big bang change, I decided to do the easy part(s) first.

In this particular part, I don't think multiple tabs help, because they'd make it a bit harder to see old-new side by side.

And the tab grouping we're fixing here was a case that was never intended to be grouped.

PR #4520, opened 2021-12-20
change order of JS source maps troubleshooting steps

imatwawana:
Per @souredoutlook 
Currently we list the "Verify artifact names match sourceMappingURL value" step in our source maps troubleshooting guide above the "Verify Artifact Names Match Stack Trace Frames". Given that we've tightened source map fetch timeouts and recommend to always upload, switching the order of these two steps, as "Verify Artifact Names Match Stack Trace Frames" is the advice that's more relevant to users that upload.


lobsterkatie (approved):
👍🏻

PR #4523, opened 2021-12-21
use arrow function in sample code for consistency

ali-farooq0:
Using `=>` to create a function in the sample code for consistency.


ali-farooq0:
@kamilogorek does the build process need to be kicked again to get the expected checks required to get this merged in?

PR #4524, opened 2021-12-21
sessions(dotnet) note client mode only

bruno-garcia:


src/platforms/common/configuration/releases.mdx
```diff
@@ -65,6 +65,17 @@ After configuring your SDK, you can install a repository integration or manually
 <PlatformSection supported={["apple", "android", "javascript", "flutter", "native", "rust", "python", "react-native", "dotnet", "unity"]}>
 
 ## Release Health
+  
+<PlatformSection supported={["dotnet"]}>
+
+<Note>
+
+Release health support for .NET is currently limited to single-instance _Client_ applications only. This means it can be used with WPF, Xamarin, WinForms and similar app models.
```
imatwawana, line 73: For emphasis, we've settled on bold, no italics. For new terms, we go with italics. Also, I don't think "client" needs to be capitalized here. Otherwise, looks good!

```suggestion
Release health support for .NET is currently limited to single-instance **client** applications only. This means it can be used with WPF, Xamarin, WinForms and similar app models.
```

imatwawana (approved):

rohitkataria-sentryio (approved):

PR #4531, opened 2021-12-23
fix status unknown in example

Maxim-Mazurok:
transaction.finish() overrides status set for the last span resulting in the last span always sending unknown status

related to https://github.com/getsentry/sentry-dotnet/issues/919


Maxim-Mazurok:
Actually, this seems to introduce another problem: transaction keeps running until it's cancelled....

Maxim-Mazurok:
never mind, I think I just got confused, it seems to work fine now

PR #4544, opened 2021-12-30
fix: Document {{auto}} properly [ISSUE-737]

untitaker:
We special-case some platform identifiers in Relay when it comes to the
default value of `user.ip_address`. This causes a bunch of confusion for
customers and SDK authors.

This problem came up in an old customer issue involving the native SDK, and it appears that the native SDK does not send `ip_address={{auto}}` when running as part of a desktop application (which it could maybe try to detect)

### DIFF
--- a/src/platforms/common/enriching-events/identify-user.mdx
+++ b/src/platforms/common/enriching-events/identify-user.mdx
@@ -37,7 +37,7 @@ Users consist of a few critical pieces of information that construct a unique id
 
 : The user's IP address. If the user is unauthenticated, Sentry uses the IP address as a unique identifier for the user. Sentry will attempt to pull this from the HTTP request data (`request.env.REMOTE_ADDR` field in JSON), if available. That might require <PlatformIdentifier name="send-default-pii" /> set to `true` in the SDK options.
 
-If the user's `ip_address` is set to `"{{auto}}"`, Sentry will infer the IP address from the connection between your app and Sentry's server. For certain SDKs, particularly iOS and frontend JavaScript, this is the default behavior and requires no further configuration. For most other SDKs the default is not to collect any IP address. If you find that you're missing the user's IP address, setting this value to `"{{auto}}"` explicitly might help.
+If the user's `ip_address` is set to `"{{auto}}"`, Sentry will infer the IP address from the connection between your app and Sentry's server. For certain SDKs, particularly iOS and frontend JavaScript, this is the default behavior and requires no further configuration. For most other SDKs the default is to not collect any IP address. If you find that you're missing the user's IP address, setting this value to `"{{auto}}"` explicitly might help.
 
 <!--
 SDK devs who read this and wonder why their SDK does not behave like the JS
@@ -49,7 +49,7 @@ default. We don't really want to special-case more platforms.
 See https://github.com/getsentry/relay/blob/54fc479259c5c8a70c48ad1994d5e8eb8629ba32/relay-general/src/store/normalize.rs#L380-L383
 -->
 
-To opt-out from collecting user's IP addresses, you can use Sentry's [server-side data](/product/data-management-settings/scrubbing/) scrubbing to remove `$user.ip_address`. Adding such a rule ultimately overrules any other logic.
+To opt-out of collecting users' IP addresses, you can use Sentry's [server-side data](/product/data-management-settings/scrubbing/) scrubbing to remove `$user.ip_address`. Adding such a rule ultimately overrules any other logic.
 
 Additionally, you can provide arbitrary key/value pairs beyond the reserved names, and the Sentry SDK will store those with the user.
 
--- a/src/platforms/common/enriching-events/identify-user.mdx
+++ b/src/platforms/common/enriching-events/identify-user.mdx
@@ -11,45 +11,41 @@ description: "Learn how to configure the SDK to capture the user and gain critic
 
 Users consist of a few critical pieces of information that construct a unique identity in Sentry. Each of these is optional, but one **must** be present for the Sentry SDK to capture the user:
 
+<DefinitionList>
+
 <PlatformSection notSupported={["apple"]}>
 
-`id`
+### `id`
 
-: Your internal identifier for the user.
+Your internal identifier for the user.
 </PlatformSection>
 
 <PlatformSection supported={["apple"]}>
 
-`userId`
+### `userId`
 
-: Your internal identifier for the user.
+Your internal identifier for the user.
 </PlatformSection>
 
-`username`
+### `username`
 
-: The username. Typically used as a better label than the internal id.
+The username. Typically used as a better label than the internal id.
 
-`email`
+### `email`
 
-: An alternative, or addition, to the username. Sentry is aware of email addresses and can display things such as Gravatars and unlock messaging capabilities.
+An alternative, or addition, to the username. Sentry is aware of email addresses and can display things such as Gravatars and unlock messaging capabilities.
 
-`ip_address`
+### `ip_address`
 
-: The user's IP address. If the user is unauthenticated, Sentry uses the IP address as a unique identifier for the user. Sentry will attempt to pull this from the HTTP request data (`request.env.REMOTE_ADDR` field in JSON), if available. That might require <PlatformIdentifier name="send-default-pii" /> set to `true` in the SDK options.
+The user's IP address. If the user is unauthenticated, Sentry uses the IP address as a unique identifier for the user. Sentry will attempt to pull this from the HTTP request data (`request.env.REMOTE_ADDR` field in JSON), if available. That might require <PlatformIdentifier name="send-default-pii" /> set to `true` in the SDK options.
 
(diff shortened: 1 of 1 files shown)


priscilawebdev (approved):

src/platforms/common/enriching-events/identify-user.mdx
```diff
@@ -35,13 +35,21 @@ Users consist of a few critical pieces of information that construct a unique id
...
-<PlatformSection supported={["apple"]}>
+If the user's `ip_address` is set to `"{{auto}}"`, Sentry will infer the IP address from the connection between your app and Sentry's server. For certain SDKs, particularly iOS and frontend JavaScript, this is the default behavior and requires no further configuration. For most other SDKs the default is not to collect any IP address. If you find that you're missing the user's IP address, setting this value to `"{{auto}}"` explicitly might help.
 
-Events from the SDK for Apple already consider the incoming connection as the user's IP address. To opt-out from this, you can use Sentry's [server-side data](/product/data-management-settings/scrubbing/) scrubbing to remove `$user.ip_address`.
+<!--
+SDK devs who read this and wonder why their SDK does not behave like the JS
+SDK: Relay changes the default value from `null` to `{{auto}}` for the
+platforms javascript, cocoa, objc. If you want to collect IP addresses by
+default because you're building a frontend/desktop SDK, just send {{auto}} by
+default. We don't really want to special-case more platforms.
```
relaxolotl, line 47: nit: are developers usually referencing sentry docs in a way that would make comments like these visible to them? using the native documents as an example (https://sentry-docs-git-doc-ip-auto.sentry.dev/platforms/native/enriching-events/identify-user/) it's not immediately obvious to me how i would find this bit of information without directly referencing the source document. this looks like useful information, but it's squirreled away in a place that doesn't appear to be too discoverable for those that it would help.

src/platforms/common/enriching-events/identify-user.mdx
```diff
@@ -35,13 +35,21 @@ Users consist of a few critical pieces of information that construct a unique id
...
+: The user's IP address. If the user is unauthenticated, Sentry uses the IP address as a unique identifier for the user. Sentry will attempt to pull this from the HTTP request data (`request.env.REMOTE_ADDR` field in JSON), if available. That might require <PlatformIdentifier name="send-default-pii" /> set to `true` in the SDK options.
 
-<PlatformSection supported={["apple"]}>
+If the user's `ip_address` is set to `"{{auto}}"`, Sentry will infer the IP address from the connection between your app and Sentry's server. For certain SDKs, particularly iOS and frontend JavaScript, this is the default behavior and requires no further configuration. For most other SDKs the default is not to collect any IP address. If you find that you're missing the user's IP address, setting this value to `"{{auto}}"` explicitly might help.
```
relaxolotl, line 40: this paragraph should belong under `ip_address`, but it looks like "breaks out" of that section and is de-indented:

Screenshot 2022-01-03 at 11 55 45 AM

i think this could cause a bit of confusion - is there a way to fix this?

relaxolotl (approved):
i don't really have much insight to add on the native side of things wrt IP addresses, but this looks good minus the nitpicky points i raised 👍

src/platforms/common/enriching-events/identify-user.mdx
imatwawana, line 40:
```suggestion
If the user's `ip_address` is set to `"{{auto}}"`, Sentry will infer the IP address from the connection between your app and Sentry's server. For certain SDKs, particularly iOS and frontend JavaScript, this is the default behavior and requires no further configuration. For most other SDKs the default is to not collect any IP address. If you find that you're missing the user's IP address, setting this value to `"{{auto}}"` explicitly might help.
```

src/platforms/common/enriching-events/identify-user.mdx
```diff
@@ -35,13 +35,21 @@ Users consist of a few critical pieces of information that construct a unique id
...
+SDK: Relay changes the default value from `null` to `{{auto}}` for the
+platforms javascript, cocoa, objc. If you want to collect IP addresses by
+default because you're building a frontend/desktop SDK, just send {{auto}} by
+default. We don't really want to special-case more platforms.
 
-</PlatformSection>
+See https://github.com/getsentry/relay/blob/54fc479259c5c8a70c48ad1994d5e8eb8629ba32/relay-general/src/store/normalize.rs#L380-L383
+-->
+
+To opt-out from collecting user's IP addresses, you can use Sentry's [server-side data](/product/data-management-settings/scrubbing/) scrubbing to remove `$user.ip_address`. Adding such a rule ultimately overrules any other logic.
```
imatwawana, line 52:
```suggestion
To opt-out of collecting users' IP addresses, you can use Sentry's [server-side data](/product/data-management-settings/scrubbing/) scrubbing to remove `$user.ip_address`. Adding such a rule ultimately overrules any other logic.
```

imatwawana (approved):
A couple small wording fixes

src/platforms/common/enriching-events/identify-user.mdx
imatwawana: Indenting in markdown without showing anything isn't always very straightforward in Markdown, but I wonder if the `: ` would work on the following paragraphs.

src/platforms/common/enriching-events/identify-user.mdx
untitaker: i now went with less technical phrasing such that it can be shown in user docs

src/platforms/common/enriching-events/identify-user.mdx
untitaker: I found a component in our docs page that works like the `: `, but allows for paragraphs.

untitaker:
asking for another review since I significantly changed the contents to get rid of `<!-- -->`

src/platforms/common/enriching-events/identify-user.mdx
```diff
@@ -11,37 +11,41 @@ description: "Learn how to configure the SDK to capture the user and gain critic
...
 
-<PlatformSection supported={["apple"]}>
+If the user's `ip_address` is set to `"{{auto}}"`, Sentry will infer the IP address from the connection between your app and Sentry's server.
 
-Events from the SDK for Apple already consider the incoming connection as the user's IP address. To opt-out from this, you can use Sentry's [server-side data](/product/data-management-settings/scrubbing/) scrubbing to remove `$user.ip_address`.
+If the field is omitted, the default value is `null`. However, due to backwards compatibility concerns, certain platforms (in particular JavaScript) have a different default value of `"{{auto}}"`. SDKs and other clients are strongly discouraged from relying on this behavior and should set IP-addresses or `"{{auto}}"` explicitly.
 
-</PlatformSection>
+To opt-out of collecting users' IP addresses, you can use Sentry's [server-side data](/product/data-management-settings/scrubbing/) scrubbing to remove `$user.ip_address`. Adding such a rule ultimately overrules any other logic.
```
imatwawana, line 46:
```suggestion
To opt out of collecting users' IP addresses, you can use Sentry's [server-side data](/product/data-management-settings/scrubbing/) scrubbing to remove `$user.ip_address`. Adding such a rule ultimately overrules any other logic.
```

src/platforms/common/enriching-events/identify-user.mdx
```diff
@@ -11,37 +11,41 @@ description: "Learn how to configure the SDK to capture the user and gain critic
...
+The user's IP address. If the user is unauthenticated, Sentry uses the IP address as a unique identifier for the user. Sentry will attempt to pull this from the HTTP request data (`request.env.REMOTE_ADDR` field in JSON), if available. That might require <PlatformIdentifier name="send-default-pii" /> set to `true` in the SDK options.
 
-<PlatformSection supported={["apple"]}>
+If the user's `ip_address` is set to `"{{auto}}"`, Sentry will infer the IP address from the connection between your app and Sentry's server.
 
-Events from the SDK for Apple already consider the incoming connection as the user's IP address. To opt-out from this, you can use Sentry's [server-side data](/product/data-management-settings/scrubbing/) scrubbing to remove `$user.ip_address`.
+If the field is omitted, the default value is `null`. However, due to backwards compatibility concerns, certain platforms (in particular JavaScript) have a different default value of `"{{auto}}"`. SDKs and other clients are strongly discouraged from relying on this behavior and should set IP-addresses or `"{{auto}}"` explicitly.
```
imatwawana, line 44: We can't really say that SDKs and other clients are strongly discouraged because it's giving them human qualities. We can either say the makers of those SDKs and other clients are strongly discouraged or perhaps the suggestion I've made below. Whichever you prefer.

```suggestion
If the field is omitted, the default value is `null`. However, due to backwards compatibility concerns, certain platforms (in particular JavaScript) have a different default value for `"{{auto}}"`. SDKs and other clients should not rely on this behavior and should set IP addresses or `"{{auto}}"` explicitly.
```

imatwawana (approved):
A couple small fixes and one thing that needs to be clarified, but otherwise looks good!

src/platforms/common/enriching-events/identify-user.mdx
untitaker: yup thanks, your suggestion makes sense. i think programmers tend to anthropomorphize their creations so it's normal to me.

PR #4549, opened 2022-01-04
Remove redundant check from Laravel code example

devfrey:
Laravel's exception handler does not call the report(able) callbacks for exceptions that should not be reported. See https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Exceptions/Handler.php#L223-L225


imatwawana:
@stayallive - Can you check this?

stayallive:
Woops, that is probably a copy pasta error and a relic from the past, yes looks good! 👍 

Thanks @devfrey, although it's not harmful there is also no point of course 😄

PR #4555, opened 2022-01-05
fix: apple scope level

### LINKED ISSUE #4543: Usage > Set Level refers to an old property name
### Core or SDK?

Platform/SDK

### Which part? Which one?

iOS

### Description

In the documentation page "Home > Platforms > Apple > iOS > Usage > Set the Level" ([here](https://docs.sentry.io/platforms/apple/guides/ios/usage/set-level/)), the referenced property `options.logLevel` and the possible values have been renamed by this pull request: https://github.com/getsentry/sentry-cocoa/pull/979.

### Suggested Solution

Use the correct property: `diagnosticLevel` and `SentryLevel` enum.

bruno-garcia:
Resolves #4543 

This Usage item refers to using the Scope to set a level for a capture. Doesn't make much sense with Global scope since it would affect all `captureMessage` calls so just removing it from the docs for Apple


bruno-garcia:
Because this is not about diagnostic level as I described in the PR description

bruno-garcia:
Check other platforms for clarification.

philipphofmann:
We have diagnosticLevel and Debug see https://github.com/getsentry/sentry-cocoa/blob/7cf0f7441aed40e4c344404d10172547fb3edd4a/Sources/Sentry/Public/SentryOptions.h#L33-L39, but we don't have the possibility for a custom implementation.

marandaneto:
> @philipphofmann Diagnostic level is documented in the Options: [docs.sentry.io/platforms/apple/configuration/options/#diagnostic-level](https://docs.sentry.io/platforms/apple/configuration/options/#diagnostic-level)
> 
> Diagnostic Logger in the .NET SDK: [docs.sentry.io/platforms/dotnet/configuration/diagnostic-logger](https://docs.sentry.io/platforms/dotnet/configuration/diagnostic-logger/) but I don't believe we have this in Cocoa, do we? It could be done for Java/Android though. /cc @marandaneto

Android/Java has the `logger` interface and `diagnosticLevel` field, so custom implementation is already possible.

marandaneto:
> > Android/Java has the logger interface and diagnosticLevel field, so custom implementation is already possible.
> 
> But I don't see docs like for .NET [docs.sentry.io/platforms/dotnet/configuration/diagnostic-logger](https://docs.sentry.io/platforms/dotnet/configuration/diagnostic-logger/) or am I missing something @marandaneto?

docs are indeed missing, https://docs.sentry.io/platforms/java/configuration/options/#diagnostic-level
but the diagnostic page we don't have at all.

PR #4562, opened 2022-01-10
ref(app-store-connect): Update plan alert note

priscilawebdev:
Users on a non-Bussiness Plan or above were getting backend errors when trying to add a custom repository. This bug (regression) was fixed in the PR https://github.com/getsentry/sentry/pull/30820

This PR updates the App Store Connect documentation informing the users that they needed to be on the business plan or above to have access to the integration.

### DIFF
--- a/src/platforms/apple/common/dsym.mdx
+++ b/src/platforms/apple/common/dsym.mdx
@@ -43,7 +43,7 @@ Once the integration is set up it will ensure that new builds are detected and f
 
 <Alert level="Note">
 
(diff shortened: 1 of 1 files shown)


flub (changes requested):
Afaik the change in https://github.com/getsentry/sentry/pull/30820 is a regression and the way this is documented here is the intention of how it works.

imatwawana:
Not sure what the outcome of this PR will be, but if this note gets added, I'd suggest moving it up to just after the first paragraph in the section, like so:

priscilawebdev:
Thanks, @imatwawana but @flub is right. We will have to update the upsell logic again, so I am closing this PR

PR #4565, opened 2022-01-10
Unity: Native support limited to config window

bitsandfoxes:
To at least document our current limitation to have to use the config window to get native support.


bruno-garcia (approved):
Great! THanks

PR #4586, opened 2022-01-14
feat(ratelimits): Added page for rate limit in api reference

AniketDas-Tekky:
Document focuses on giving relevant information to API Users regarding the rate limit and how to monitor it.

### DIFF
--- a/src/api/index.mdx
+++ b/src/api/index.mdx
@@ -14,3 +14,4 @@ The current version of the web API is known as **v0** and is considered to be in
 - [Authentication](/api/auth/)
 - [Permissions](/api/permissions/)
 - [Pagination](/api/pagination/)
+- [Rate Limits](/api/ratelimits)
--- a/src/api/index.mdx
+++ b/src/api/index.mdx
@@ -10,8 +10,8 @@ The current version of the web API is known as **v0** and is considered to be in
 
 ## Getting Started
 
-- [Requests](/api/requests/)
 - [Authentication](/api/auth/)
-- [Permissions](/api/permissions/)
 - [Pagination](/api/pagination/)
+- [Permissions](/api/permissions/)
 - [Rate Limits](/api/ratelimits)
+- [Requests](/api/requests/)
--- a/src/api/ratelimits.mdx
+++ b/src/api/ratelimits.mdx
@@ -1,0 +1,48 @@
+---
+title: Rate Limits
+sidebar_order: 5
+---
+
+Every API request is
+
+```bash
+curl -i https://sentry.io/api/0/projects/1/groups/
+```
+
+```http
+HTTP/1.0 200 OK
+Date: Sat, 14 Feb 2015 18:47:20 GMT
+Content-Type: application/json
+Content-Language: en
+Allow: GET, HEAD, OPTIONS
+Link: <https://sentry.io/api/0/projects/1/groups/?&cursor=1420837590:0:1>;
+  rel="previous"; results="false",
+  <https://sentry.io/api/0/projects/1/groups/?&cursor=1420837533:0:0>;
+  rel="next"; results="true"
+```
+
+When supported, cursors will **always** be returned for both a previous and a next page, even if there are no results on these pages. This allows you to make a query against the API for yet-undiscovered results. An example where this would be used is when you’re implementing polling behavior and you want to see if there is any new data. We return a `results="[true|false]"` indicator to determine if you actually need to paginate.
+
+## Pagination Example
+
+Here is a pagination example using this API endpoint:
+
+https://docs.sentry.io/api/events/list-an-issues-events/
+
+The HTTP request in this example returns 100 events for the issue and has the following link header in the response:
+```discover {tabTitle:Http}
+<https://sentry.io/api/0/issues/123456/events/?&cursor=0:0:1>; rel="previous"; results="false"; cursor="0:0:1", <https://sentry.io/api/0/issues/123456/events/?&cursor=0:100:0>; rel="next"; results="true"; cursor="0:100:0"
+```
+
+One of the URLs in the link response has `rel=next`, which indicates the next results page. It also has `results=true`, which means that there are more results.
+
+Based on this, the next request is `GET <https://sentry.io/api/0/issues/123456/events/?&cursor=0:100:0>`.
+
+This request will return the next 100 events for that issue, again, with the following link header:
+```discover {tabTitle:Http}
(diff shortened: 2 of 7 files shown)


src/api/ratelimits.mdx
```diff
@@ -0,0 +1,28 @@
+---
+title: Rate Limits
+sidebar_order: 5
+---
+
+Sentry rate limits every API request made to prevent abuse and resource overuse. The limit is applied to each caller per endpoint.
+
+The rate limit follows a fixed window approach. Requests are counted into time buckets, determined by the window size. If the number of requests from a caller exceeds the number of allowed requests within a window, then that request will be rejected.
```
vladkluev, line 8: Should we also say that there will be a concurrent request rate limiter in the future? Might as well get people ready for it now. Also is there a killswitch for the global rate limiter? It may be good to provide instructions to self-hosted customers on how to disable it because they may not need it.

AniketDas-Tekky: I think it's too early to add stuff on the concurrent rate limiter. Once we have a better idea on what it will look like for users, then we can talk about adding it to the reference.

Is this the right place for self-hosted specific points? The kill switch right now is the `enforce_rate_limits` flag in `Endpoint`.

src/api/ratelimits.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+While there is a default value, each endpoint can have their own maximum number of requests and window size. You can track your rate limit usage by looking at special headers in the response
```
brianthi, line 10: If defaults can be overwritten, should there be some mention on how to do it?

AniketDas-Tekky: Right now, rate limit overrides can only be done in the code. Is that something we want to include in this reference?

brianthi: No, let's not include it then.

src/api/ratelimits.mdx
vladkluev: > Is this the right place for self-hosted specific points? 

That's not a kill switch. It requires a deploy and a code change. Does sentry not have a runtime flags system in redis or flagr or something?

src/api/ratelimits.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+## Headers
+
+Every API request response will include the following headers:
+
+- `X-Sentry-Rate-Limit-Limit`
```
NisanthanNanthakumar, line 16: @AniketDas-Tekky why 2 `Limit`? Im assuming its a typo

src/api/ratelimits.mdx
AniketDas-Tekky: We didn't want to use flagr because it would add too much latency to every request. We have to create a new system for this.

AniketDas-Tekky: We didn't want to use flagr because it would add too much latency to every request. We have to create a new system for this.

src/api/ratelimits.mdx
AniketDas-Tekky: not a type, "Rate Limit" is the feature, while the 2nd Limit is the maximum number of request.
Feels weird to me too, but it's apparently pretty standard.

src/api/index.mdx
```diff
@@ -14,3 +14,4 @@ The current version of the web API is known as **v0** and is considered to be in
 - [Authentication](/api/auth/)
 - [Permissions](/api/permissions/)
 - [Pagination](/api/pagination/)
+- [Rate Limits](/api/ratelimits)
```
imatwawana, line 17: This is a bit outside the scope of this PR, but does anyone know why this list here doesn't follow the same order as these left hand navigation? Thinking these should match.

src/api/ratelimits.mdx
imatwawana, line 10:
```suggestion
While there is a default value, each endpoint can have their own maximum number of requests and window size. You can track your rate limit usage by looking at special headers in the response.
```

src/api/ratelimits.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+Sentry rate limits every API request made to prevent abuse and resource overuse. The limit is applied to each caller per endpoint.
+
+The rate limit follows a fixed window approach. Requests are counted into time buckets, determined by the window size. If the number of requests from a caller exceeds the number of allowed requests within a window, then that request will be rejected.
+
+While there is a default value, each endpoint can have their own maximum number of requests and window size. You can track your rate limit usage by looking at special headers in the response
+
+## Headers
+
+Every API request response will include the following headers:
```
imatwawana, line 14:
```suggestion
Every API request response includes the following headers:
```

src/api/ratelimits.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+- `X-Sentry-Rate-Limit-Limit`
+  - The maximum number of requests allowed within the window
+- `X-Sentry-Rate-Limit-Remaining`
+  - The number of requests this caller has left on this endpoint within the current window
+- `X-Sentry-Rate-Limit-Reset`
+  - The time when the next rate limit window begins and the count resets
+  - Measured in UTC seconds from epoch
```
imatwawana, line 22: To make it fit the set pattern a bit better. If this section grows in the future, then it's fine to break it back out. 
```suggestion
  - The time when the next rate limit window begins and the count resets, measured in UTC seconds from epoch
```

src/api/ratelimits.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+## Additional Information
+
+The rate limiter looks at the caller's identity instead of the bearer token or cookie. As such, the rate limit cannot be bypassed by generating multiple tokens.
+
+Rate limits for internal integrations may differ from other rate limits, and tend to be higher. We reccomend using internal integration bearer tokens if you anticipate needing to make a large volume of calls.
```
imatwawana, line 28:
```suggestion
Rate limits for internal integrations may differ from other rate limits, and tend to be higher. We recommend using internal integration bearer tokens if you anticipate needing to make a large volume of calls.
```

src/api/ratelimits.mdx
```diff
@@ -0,0 +1,28 @@
+---
+title: Rate Limits
+sidebar_order: 5
+---
+
+Sentry rate limits every API request made to prevent abuse and resource overuse. The limit is applied to each caller per endpoint.
```
imatwawana, line 6: I don't understand the second sentence. If it makes sense, you could change it to "on a per endpoint basis" to make it grammatically correct, but I'm not sure. Feel free to ping me if you want to discuss to clarify this.
```suggestion
Sentry rate limits every API request made to prevent abuse and resource overuse. The limit is applied to each caller per endpoint.
```

imatwawana (approved):
A couple fixes and a question, but otherwise LGTM!

PR #4591, opened 2022-01-14
Update: .NET replacing WithScope with Capture overload

bitsandfoxes:
WithScope will be replaced by the individual Capture methods that have a callback for a local scope.


philipphofmann (commented):
Is this a breaking change in a major bump? If so, I think we should mention this in the migration guide. If you marked `WithScope` as deprecated, I think we should still mention it here and point out it was deprecated.

bruno-garcia (approved):

PR #4597, opened 2022-01-14
feat(apidocs): support for examples OpenAPIField

JoshFerge:
OpenAPI 3 can have a response containing a plural `examples` field, and with the new API doc tooling introduced in https://github.com/getsentry/sentry/pull/30040, it will be populated, but we currently don't support showing it. This PR modifies the apiPage to check for both fields and use the one that will be populated. Multiple examples for one response aren't trivial with our current layout so for now we only show the first one.
- https://swagger.io/specification/#example-object


JoshFerge:
build will fail until we have an endpoint response with `examples` defined in our OpenAPI json.

JoshFerge:
don't close pr please bot. will be ready for review once we have an example fully merged in.

RyanSkonnord (approved):
LGTM

PR #4603, opened 2022-01-18
fixe file i/o capitalization

imatwawana:
Changes "file" in "file I/O" to lower case in most places. 
Since I reviewed previous PRs using this term, I've learned this is a generic term that's generally not capitalized in the places where it's spelled out, so we should avoid setting a precedent that doesn't align with other documentation.


philipphofmann (approved):
LGTM 🚀

philipphofmann:
cc @romtsn.

romtsn:
Thanks @imatwawana

PR #4607, opened 2022-01-18
ref(apple): Mention other install methods sooner

philipphofmann:
The wizard and getting started page mentions that we support
alternate installation methods after explaining how to install the
SDK with CocoaPods. I hypothesize that most people don't read
the hint because they see CocoaPods and then skip the whole
section. This happened to me, and I wondered where we point
out SPM and Carthage. Mentioning the other installation
methods earlier could fix this problem.


marandaneto (approved):
makes sense to me, thanks @philipphofmann

PR #4611, opened 2022-01-19
Update cocoapods.mdx

gupta-ji6:
Updated the "Integration with existing apps" hyperlink which was redirecting to a 404 page.


marandaneto (approved):
good catch @gupta-ji6 thanks

PR #4614, opened 2022-01-19
fix(RN): Rework RN installation linking section to include wizard

jennmueng:
When working with older RN versions that require the linking step, an error is thrown when calling `react-native link` when it tries to auto-run the sentry wizard because we no longer include it as a dependency of the RN SDK. Currently in newer versions we have users run `npx` which will also fetch the sentry wizard dependency but it does not exist for older versions. Instead, for RN <0.60 we include the installation of sentry wizard as a dev dependency in the linking step to avoid this error. This PR also updates the RN wizard.

### DIFF
--- a/src/includes/getting-started-install/react-native.mdx
+++ b/src/includes/getting-started-install/react-native.mdx
@@ -48,13 +48,6 @@ For React Native versions 0.60 and after, you will need to execute the Sentry Wi
 npx @sentry/wizard -i reactNative -p ios android
 ```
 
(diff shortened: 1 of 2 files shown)


jennmueng:
As discussed with @bruno-garcia on call, we should rather just re-include @sentry/wizard as a dependency on the RN SDK as this is way more convenient and the dependency shouldn't be included in the js bundle itself when bundled with metro. cc @marandaneto

PR #4629, opened 2022-01-24
Fix: how to configure BeforeSend callback with Spring

maciejwalkowiak:


maciejwalkowiak:
These files are copies from Spring Boot docs. If there is a nice way to reuse them instead of duplicating, please let me know. 

cc @bruno-garcia

src/includes/configuration/before-send/java.spring.mdx
```diff
@@ -0,0 +1,32 @@
+`beforeSend` callback can be registered simply by creating a Spring bean implementing the `BeforeSendCallback`.
```
imatwawana, line 1: We want to avoid suggesting how simple or not simple something is going to be in case the user doesn't have the same experience. 

```suggestion
`beforeSend` callback can be registered by creating a Spring bean implementing the `BeforeSendCallback`.
```

maciejwalkowiak: Thanks!

bruno-garcia (approved):

Issue #4509, opened 2021-12-16
Link styling in main navigation doesn't always make links obvious


imatwawana:
### Core or SDK?

Core Sentry product

### Which part? Which one?

This applies to all docs

### Description

Per @armcknight 

1. The section headers in the left side navigation are not obviously links. In addition, the breadcrumb navigation relies entirely on colour/weight changes to indicate that they are links (vs underlining), which can be confusing since folks might think it's simply emphasized text and not a link. 
![2021-12-17_16-54-34](https://user-images.githubusercontent.com/76437239/146612673-683d7bfd-9747-439e-ac75-f8b7e76a8606.png)

2. When you're on the page represented by a top level heading in the left side nav, there's no indication in the navigation of where you are as with all the other pages in a section. For example:

This clearly shows where you are in the left nav:
![2021-12-17_16-52-54](https://user-images.githubusercontent.com/76437239/146612378-0cbb157e-fd31-4b2e-acd3-9b68abbc1585.png)

vs being on this top level page doesn't give you any indication of it in the left nav:
![2021-12-17_16-54-34](https://user-images.githubusercontent.com/76437239/146612425-4175bab5-3d56-4234-8719-7503d09fa5c1.png)

### Suggested Solution

Change link styling for left side nav and breadcrumbs to be unambiguous, perhaps with underlines like the links in the body content of each page.

Change styling of top level headings in side nav so that it's obvious that you're on that page.

imatwawana:
@stevenplewis what do you think?

stevenplewis:
@imatwawana Yeah, it's definitely not obvious the headers in the sidebar are links. I think the things that are a different color from the normal body text indicate to some degree those are different and could be links, as well as being in a sidebar which is a pretty standard place/way for links to be shown. It's true an underline is a very universal cue for links, I worry in the sidebar adding underlines to _all_ the sidebar links could be a bit distracting, but it's easy enough to try and find out. It looks like links in the body are our pink color, so they stand out a bit more, I imagine we landed on this light purple for side links to not be too distracting though. 

I'll play around and see if I can come up with something that makes it more obvious those are links, especially the header bits, and also try to come up with a solution for when the header has a page and there currently isn't a state like the other pages.

imatwawana:
[...]

PR #4634, opened 2022-01-24
giving titles active style and colors more similar to other links

stevenplewis:
This is in response to https://github.com/getsentry/sentry-docs/issues/4509


armcknight:
Awesome, thanks! ❤️

Issue #4350, opened 2021-11-03
OOM Detection should address the `exit(code)` issue


marandaneto:
### Core or SDK?

Platform/SDK

### Which part? Which one?

RN SDK and Cocoa SDK

### Description

OOM Detection should address the `exit(code)` issue, that if you use it, you should either turn off the feature or live with false positive

Related to https://github.com/getsentry/sentry-react-native/issues/1830

### Suggested Solution

Add to the heuristics page that you cant use it or live with it, also link to the Apple docs that this is not allowed.

PR #4635, opened 2022-01-25
feat(apple): How to sample OOMs

philipphofmann:
Explain how to sample OOMs, so they don't eat up the quota.


marandaneto:
I approve the technical part of it but I'm a bit skeptical about the change, I'd rather investigate and improve the OOM detection rather than tell how to sample them.
Addressing https://github.com/getsentry/sentry-docs/issues/4350 already helps people that are interested in the heuristics logic.
@bruno-garcia ?

PR #4647, opened 2022-01-27
docs: clarify message filtering

meichstedt:
Add details about which value to set and where to get it from for inbound filtering errors by message.

Unfortunately, there's no clear indication in the docs as to what is the correct value to set for filtering errors by message. It took me 30 minutes to find out why message filtering with the help of this [help center article](https://help.sentry.io/product-features/configuration/why-isn-t-filter-by-error-message-working/). The obvious place for that information would be here IMHO.

### DIFF
--- a/src/docs/product/accounts/quotas/index.mdx
+++ b/src/docs/product/accounts/quotas/index.mdx
@@ -230,6 +230,10 @@ If you discover a problematic release causing excessive noise, Sentry supports i
 
 Sentry supports filtering out a specific or certain kind of error as well. Navigate to **[Project] » Project Settings » Inbound Filters**, then add the error message to **Filter errors by error message**.
 
+To ensure you’re adding the correct message to the inbound filter setting, check the JSON for an event in the issue. The filter by error message setting matches the data found in the "title" field towards the bottom.
+
+The Error Message field in a project's inbound filter settings recognizes "*" as a wildcard matcher at the beginning or end of the string.
+
 ### Filter by Issue
 
 <Alert level="info">
--- a/src/docs/product/accounts/quotas/index.mdx
+++ b/src/docs/product/accounts/quotas/index.mdx
@@ -272,7 +272,7 @@ The 24-hour window ends at the beginning of the current hour, not at the current
 
 This means if you experience a spike, we'll temporarily protect you, but if the increase in volume is sustained, the spike protection limit will gradually **increase until Sentry accepts all events**.
 
-For example, in the last 24 hours, your organization has been receiving, on average, 10 events per minute (after any inbound filters have been applied). That means your current per-minute limit is 6 \* 10 or 60 events. There have been no spikes in that time, so spike control is "inactive". Something breaks, and in the next minute, your organization sends Sentry 100 events. When we see the 61st event, three things happen:
+For example, in the last 24 hours, your organization has been receiving, on average, 10 events per minute (after any inbound filters have been applied). That means your current per-minute limit is 6 * 10 or 60 events. There have been no spikes in that time, so spike control is "inactive". Something breaks, and in the next minute, your organization sends Sentry 100 events. When we see the 61st event, three things happen:
 
 1. Spike protection becomes "active".
 
--- a/src/docs/product/accounts/quotas/index.mdx
+++ b/src/docs/product/accounts/quotas/index.mdx
@@ -230,9 +230,7 @@ If you discover a problematic release causing excessive noise, Sentry supports i
 
 Sentry supports filtering out a specific or certain kind of error as well. Navigate to **[Project] » Project Settings » Inbound Filters**, then add the error message to **Filter errors by error message**.
 
-To ensure you’re adding the correct message to the inbound filter setting, check the JSON for an event in the issue. The filter by error message setting matches the data found in the "title" field towards the bottom.
-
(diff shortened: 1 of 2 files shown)


imatwawana:
Hi @meichstedt, thanks for the suggestion! This area of the content is about introducing the idea that you can use filtering by error message to control your quota usage, but not about the specifics of setting those filters. If we're to add this to the docs, I think the appropriate spot is here:
https://docs.sentry.io/product/data-management-settings/filtering/#error-message

Also, we generally keep troubleshooting type content in the Help, but this might be a good candidate for moving over. @georgekallilas what do you think?

meichstedt:
Hi @imatwawana, yes this makes total sense to me. Maybe just link the troubleshooting content from the docs you linked?

I _think_ I had seen that docs page. However, it talks about `{exception.type}` and `{exception.value}`, and as a user, it's rather opaque as to what those variables _are_. The note about checking the JSON linked from an event was the "light bulb" in my case.

Feel free to discard this PR or let me know if/how I can help further. In any case thanks for reaching out and processing this 😄

georgekallilas:
> Hi @meichstedt, thanks for the suggestion! This area of the content is about introducing the idea that you can use filtering by error message to control your quota usage, but not about the specifics of setting those filters. If we're to add this to the docs, I think the appropriate spot is here: https://docs.sentry.io/product/data-management-settings/filtering/#error-message
> 
> Also, we generally keep troubleshooting type content in the Help, but this might be a good candidate for moving over. @georgekallilas what do you think?

@imatwawana We should definitely move this from the help center to the docs. Also, I do agree with you, it fits better [here](https://docs.sentry.io/product/data-management-settings/filtering/#error-message), however I do get the feeling that customers are going to stumble across the page from this PR first...can we somehow mention it here too?

imatwawana:
> > Hi @meichstedt, thanks for the suggestion! This area of the content is about introducing the idea that you can use filtering by error message to control your quota usage, but not about the specifics of setting those filters. If we're to add this to the docs, I think the appropriate spot is here: https://docs.sentry.io/product/data-management-settings/filtering/#error-message
> > Also, we generally keep troubleshooting type content in the Help, but this might be a good candidate for moving over. @georgekallilas what do you think?
> 
> @imatwawana We should definitely move this from the help center to the docs. Also, I do agree with you, it fits better [here](https://docs.sentry.io/product/data-management-settings/filtering/#error-message) however I do get the feeling that customers are going to stumble across the page from this PR first...can we somehow mention it here too?

That makes sense to me. @meichstedt, did you still want to work on this or leave it with us?

imatwawana:
Hi @meichstedt, just checking in, did you want to continue working on this PR or leave it with us? I'm happy to complete this, but I didn't want to overstep.

meichstedt:
Hey @imatwawana thanks for checking – I'd like to leave it up to you; I'm super busy at the moment and would need to better understand what goes where. I think what you folks discussed makes sense, so I'm happy to review once it's done!

src/docs/product/data-management-settings/filtering/index.mdx
```diff
@@ -53,6 +53,10 @@ To use inbound data filters for error messages, keep the following in mind:
 - On message events, the filter matches the fully formatted message.
 - Transactions are never matched by this filter.
 
+To ensure you’re adding the correct message to the inbound filter setting, check the JSON for an event in the issue. The filter by error message setting matches the data found in the "title" field near the end of the file.
+
+The "Error Message" field in a project's inbound filter settings recognizes "\*" as a wildcard matcher at the beginning or end of the string.
+
```
imatwawana, line 59:
```suggestion
```

imatwawana:
@meichstedt - would you like to review?

src/docs/product/accounts/quotas/index.mdx
```diff
@@ -270,7 +272,7 @@ The 24-hour window ends at the beginning of the current hour, not at the current
 
 This means if you experience a spike, we'll temporarily protect you, but if the increase in volume is sustained, the spike protection limit will gradually **increase until Sentry accepts all events**.
 
-For example, in the last 24 hours, your organization has been receiving, on average, 10 events per minute (after any inbound filters have been applied). That means your current per-minute limit is 6 * 10 or 60 events. There have been no spikes in that time, so spike control is "inactive". Something breaks, and in the next minute, your organization sends Sentry 100 events. When we see the 61st event, three things happen:
+For example, in the last 24 hours, your organization has been receiving, on average, 10 events per minute (after any inbound filters have been applied). That means your current per-minute limit is 6 \* 10 or 60 events. There have been no spikes in that time, so spike control is "inactive". Something breaks, and in the next minute, your organization sends Sentry 100 events. When we see the 61st event, three things happen:
```
georgekallilas, line 275: @imatwawana Did you make this change by mistake? It seems irrelevant to this PR

imatwawana: Not intentionally, lol.

imatwawana, line 275:
```suggestion
For example, in the last 24 hours, your organization has been receiving, on average, 10 events per minute (after any inbound filters have been applied). That means your current per-minute limit is 6 * 10 or 60 events. There have been no spikes in that time, so spike control is "inactive". Something breaks, and in the next minute, your organization sends Sentry 100 events. When we see the 61st event, three things happen:
```

georgekallilas:
Looks good to me

PR #4657, opened 2022-01-29
fix(react native): Add `maxTransactionDuration` to tracing config options

lobsterkatie:
We [support the `maxTransactionDuration` option](https://github.com/getsentry/sentry-react-native/blob/4fd13137fe71984d1166c33198076da9db28f158/src/js/tracing/reactnativetracing.ts#L53) in the React Native SDK, but it's missing from the docs.


marandaneto (approved):
thanks @lobsterkatie

PR #4658, opened 2022-01-31
Fix Symfony callback syntax

### LINKED ISSUE #4653: Distributed Tracing configuration does not work
### Core or SDK?

Platform/SDK

### Which part? Which one?

Symfony

### Description

After following the steps and using the code snippets for configuring Distributed Tracing with a sampler function, I get the following error:
```shell
Please add the class to service &quot;sentry_tracer_sampler&quot; even if it is constructed by a factory since we might need to add method calls based on compile-time checks. (500 Internal Server Error)
```

My assumption is that the example for the `sentry.yaml` configuration is incorrect, however I am not sure what the correct settings should be.

### Suggested Solution

The following code snippet from the documentation needs to be updated with a functional configuration:

```yaml
sentry:
    options:
        traces_sampler: 'sentry_traces_sampler'

services:
    sentry_traces_sampler:
        factory: ['@\App\Service\Sentry', 'getTracesSampler']
```

stayallive:
Fixes #4653

### DIFF
--- a/src/platforms/php/guides/symfony/configuration/symfony-options.mdx
+++ b/src/platforms/php/guides/symfony/configuration/symfony-options.mdx
@@ -73,8 +73,8 @@ In this bundle it has three default values:
 
 ## Callables
 
-The `before_breadcrumb`, `before_send` and `traces_sampler` options accept a `callable`; thus, you cannot provide it directly through
-a YAML file; the bundle accepts a service reference, which you can build in your DIC container like this:
+The `before_breadcrumb`, `before_send` and `traces_sampler` options accept a `callable`; so, you cannot provide it directly through
+a YAML file. The bundle accepts a service reference, which you can build in your DIC container like this:
 
 ```yaml {filename:config/packages/sentry.yaml}
 sentry:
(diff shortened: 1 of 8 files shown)


src/platforms/php/guides/symfony/configuration/symfony-options.mdx
```diff
@@ -75,3 +70,37 @@ In this bundle it has three default values:
 - `%kernel.build_dir%`, to exclude Symfony's build dir
 - `%kernel.cache_dir%`, to exclude Symfony's cache dir
 - `%kernel.project_dir%/vendor`, to exclude Composer's dependencies
+
+## Callables
+
+The `before_breadcrumb`, `before_send` and `traces_sampler` options accept a `callable`; thus, you cannot provide it directly through
+a YAML file; the bundle accepts a service reference, which you can build in your DIC container like this:
```
imatwawana, line 77:
```suggestion
The `before_breadcrumb`, `before_send` and `traces_sampler` options accept a `callable`; so, you cannot provide it directly through
a YAML file. The bundle accepts a service reference, which you can build in your DIC container like this:
```

imatwawana (approved):
Just a small wording change and a punctuation fix

PR #4676, opened 2022-02-02
Fix docs for CodePush and Appcenter missing params

marandaneto:


src/platforms/react-native/manual-setup/codepush.mdx
```diff
@@ -69,4 +69,16 @@ If you use custom deployment names in CodePush, you will need to use `--deployme
 sentry-cli react-native appcenter {APP} ios ./build/CodePush --deployment {DEPLOYMENT} --dist {DIST}
 ```
 
+For non-macOS devices building for iOS, you will need to pass extra parameters that can't be computed automatically (`--bundle-id` and `--version-name`).
+
+```bash
+sentry-cli react-native appcenter {APP} ios ./build/CodePush --deployment {DEPLOYMENT} --dist {DIST} --bundle-id {BUNDLE_ID} --version-name {VERSION_NAME}
+```
+
+In case you don't want to use the automatically computed release format, you will need to pass the `--release-name` parameter.
```
imatwawana, line 78:
```suggestion
If you don't want to use the automatically computed release format, you will need to pass the `--release-name` parameter.
```

imatwawana (approved):

PR #4678, opened 2022-02-02
update discover filter by table columns section

imatwawana:
Updates "Filter by Table Columns" section in Discover docs to more accurate "Group and Stack With Table Columns". The results aren't being further filtered/narrowed, they're being grouped and/or stacked. 
Other minor wording and structural changes to that section.

### DIFF
--- a/src/docs/product/discover-queries/query-builder.mdx
+++ b/src/docs/product/discover-queries/query-builder.mdx
@@ -85,11 +85,13 @@ When the "Display" selection is either "Top Period" or "Top Daily", an optional 
 
 When your "Display" selection is either "Total Period", "Previous Period", "Total Daily" or "Bar Chart", you can set up to three y-axes by selecting multiple checkboxes in the "Y-Axis" dropdown.
 
-## Filter by Table Columns
+## Group and Stack With Table Columns {#filter-by-table-columns}
 
-Above the table, click "Columns" to open a modal. This will show you a list of all the columns in the results table. You can add, delete and move basic key field columns or custom tags columns. With the same view, you can also stack events with any of the following functions:
+Above the table, click "Columns" to open the modal that shows you a list of all the columns in the results table. You can add, delete, and move basic key field columns or custom tag columns by which results are grouped.
 
 ### Stacking Functions
+
+You can also add any of the following functions as columns to stack events:
 
 - `avg(...)`
 - `count(...)`
@@ -98,7 +100,7 @@ Above the table, click "Columns" to open a modal. This will show you a list of a
 - `min(...)`
 - `sum(...)`
 
-Each function will ask you to assign a parameter. Some are required while others are optional. Functions will stack events based on the same values. If no functions are applied, the events in your Query Results will remain individually listed. Once you are done editing the columns, click "Apply" and results will be reflected in the Query Results. Keep in mind, the table may horizontall scroll if too many columns are added.
+Each function will ask you to assign a parameter. Some are required while others are optional. Functions will stack events based on the same values. If no functions are applied, the events in your "Query Results" table will remain individually listed. Once you are done editing the columns, click "Apply" and results will be reflected in the query results. Keep in mind, the table may horizontally scroll if many columns are added.
 
 ### Cell Filters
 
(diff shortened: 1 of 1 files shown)


src/docs/product/discover-queries/query-builder.mdx
```diff
@@ -85,20 +85,22 @@ When the "Display" selection is either "Top Period" or "Top Daily", an optional
 
 When your "Display" selection is either "Total Period", "Previous Period", "Total Daily" or "Bar Chart", you can set up to three y-axes by selecting multiple checkboxes in the "Y-Axis" dropdown.
 
-## Filter by Table Columns
+## Group and Stack With Table Columns {#filter-by-table-columns}
```
imatwawana, line 88:
```suggestion
## Group & Stack With Table Columns {#filter-by-table-columns}
```

src/docs/product/discover-queries/query-builder.mdx
```diff
@@ -85,20 +85,22 @@ When the "Display" selection is either "Top Period" or "Top Daily", an optional
...
 
-Above the table, click "Columns" to open a modal. This will show you a list of all the columns in the results table. You can add, delete and move basic key field columns or custom tags columns. With the same view, you can also stack events with any of the following functions:
+Above the table, click "Columns" to open the modal that shows you a list of all the columns in the results table. You can add, delete, and move basic key field columns or custom tag columns by which results are grouped.
 
 ### Stacking Functions
 
+You can also add any of the following functions as columns to stack events:
+
 - `avg(...)`
 - `count(...)`
```
imatwawana, line 97: This doesn't seem to be what the UI reflects anymore. Now there's a function that looks like this: `count()` and it doesn't take a parameter. So I'm not sure if this just needs to be removed or if the paragraph below is no longer accurate.

dashed: @imatwawana `count()` would be one of the few functions that do not take in a parameter. In circa 2019, it did take in a parameter; but this was changed. 

`count()` would need to be here; some paragraphs might need some re-wording.

src/docs/product/discover-queries/query-builder.mdx
```diff
@@ -85,20 +85,22 @@ When the "Display" selection is either "Top Period" or "Top Daily", an optional
...
 - `count_unique(...)`
 - `max(...)`
 - `min(...)`
 - `sum(...)`
 
-Each function will ask you to assign a parameter. Some are required while others are optional. Functions will stack events based on the same values. If no functions are applied, the events in your Query Results will remain individually listed. Once you are done editing the columns, click "Apply" and results will be reflected in the Query Results. Keep in mind, the table may horizontall scroll if too many columns are added.
+Each function will ask you to assign a parameter. Some are required while others are optional. Functions will stack events based on the same values. If no functions are applied, the events in your "Query Results" table will remain individually listed. Once you are done editing the columns, click "Apply" and results will be reflected in the query results. Keep in mind, the table may horizontally scroll if many columns are added.
```
imatwawana, line 103:
```suggestion
Some, but not all, functions require a parameter. Functions will stack events based on the same values. If no functions are applied, the events in your "Query Results" table will remain individually listed. Once you are done editing the columns, click "Apply" and results will be reflected in the query results. Keep in mind, the table may horizontally scroll if many columns are added.
```

src/docs/product/discover-queries/query-builder.mdx
imatwawana, line 97:
```suggestion
- `count()`
```

src/docs/product/discover-queries/query-builder.mdx
```diff
@@ -85,20 +85,22 @@ When the "Display" selection is either "Top Period" or "Top Daily", an optional
 
 When your "Display" selection is either "Total Period", "Previous Period", "Total Daily" or "Bar Chart", you can set up to three y-axes by selecting multiple checkboxes in the "Y-Axis" dropdown.
 
-## Filter by Table Columns
+## Group & Stack With Table Columns {#filter-by-table-columns}
```
imatwawana, line 88: @doralchan - I think this aligns with our conversation and the existing description, however, I'm a bit concerned because in the columns modal text, it uses the word "group" in the way we're using "stack" here. I think we should probably decide on exactly which terminology we want to use and then be consistent between docs and the UI.

src/docs/product/discover-queries/query-builder.mdx
dashed: Looks good to me! 👍

src/docs/product/discover-queries/query-builder.mdx
imatwawana: Ah, thanks for the cc on this!

imatwawana:
Hey @doralchan and @getsentry/discover-n-dashboards, can I get a quick review of this?

wmak (approved):
Content looks accurate to me :+1:

PR #4679, opened 2022-02-02
Update tips.mdx to use updated ignoreErrors and denyOptions

cvoege:
Update tips.mdx to use updated `ignoreErrors` and `denyUrls`, also update gist link to a version that is compatible. If you try to use the code given here, it will error in typescript, or simply do nothing as far as URLs go in javascript, since the `ignoreUrls` option was renamed to `denyUrls`


AbhiPrasad (approved):
Thank you for your contribution!

PR #4684, opened 2022-02-03
escape asterisk on integrations page

asztalosdani:
Asterisk escaped so it is not rendered as a bullet list.


billyvg (approved):

imatwawana:
Thanks very much @asztalosdani!

PR #4690, opened 2022-02-04
fix: Mention the minimum symbol server download speed

flub:
For custom symbols servers we have a minimum download speed in order
to protect our own infrastructure.  Document this publicly.


flub:
wait, this just merged without approval?

PR #4705, opened 2022-02-08
Correct link to spike.sh page

flynniec6:
Corrected the destination for the link for ```spike.sh```.


imatwawana:
Thanks for the catch @flynniec6!

PR #4709, opened 2022-02-08
ref(perf): Cleanup EA notes on suspect spans

k-fish:
### Summary
Suspect spans GA means the EA notes no longer apply, cleaning these up.


Zylphrex (approved):
I'm going to bump to GA rollout to 100% soon.

PR #4711, opened 2022-02-08
Update Gin docs to use correct middleware order

lewislbr:
This relates to https://github.com/getsentry/sentry-go/issues/370

### DIFF
--- a/src/platforms/go/guides/gin/index.mdx
+++ b/src/platforms/go/guides/gin/index.mdx
@@ -41,10 +41,13 @@ if err := sentry.Init(sentry.ClientOptions{
 }
 
 // Then create your app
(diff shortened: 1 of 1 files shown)


imatwawana:
@onurtemizkan, can you review this?

sl0thentr0py (changes requested):
@lewislbr thx for your contribution!
Since this is a more specific use case about middleware ordering, I'd prefer it if you would add a new section to the page instead of changing the default sample.

lewislbr:
Actually, I further tested it and this doesn't seem to be an issue anymore. Nor middleware order nor the way of initializing the server matter, so closing 👌

PR #4714, opened 2022-02-09
Fix integrations example

carestad:
The example used (call to `new Sentry.BrowserTracing()`) gave errors in the browser. The correct code seems to be `new Sentry.Integrations.BrowserTracing()`.


lobsterkatie:
Hi, @carestad! Thanks for your contribution!

We [changed how `BrowserTracing` is accessed](https://github.com/getsentry/sentry-javascript/pull/4204), so the docs are right, or at least, describe what should be right. Unfortunately, the CDN bundle was missed in that PR. I've pushed a new PR to fix this, linked above.

PR #4725, opened 2022-02-14
chore(php): add leading slash to avoid import

caugner:


imatwawana:
@stayallive can you take a look, please?

stayallive (approved):
Good catch 👍

Issue #4762, opened 2022-02-24
Adjust Wording In Description of ignoreErrors option


souredoutlook:
### Core or SDK?

Core Sentry product

### Which part? Which one?

Manage Event Stream Guide

### Description

In reference to [Issue#4605 from getsentry/sentry-javascript](https://github.com/getsentry/sentry-javascript/issues/4605):

The description of ignoreErrors located [here](https://docs.sentry.io/product/accounts/quotas/manage-event-stream-guide/#inboundfilters) states: 

>it matches any of the listed error messages. If no message is available, the SDK will try to compare against an underlying exception type and value.

This isn't exactly true since the string that it matches against is inclusive of both the exception type and value like: `ExceptionType: some value`.

This lead some customers to try and use RegExp assertions to strictly match against the error type. Which can fail depending on the RegExp.

### Suggested Solution

Change this text:

>it matches any of the listed error messages. If no message is available, the SDK will try to compare against an underlying exception type and value.

To this:

>it matches any of the listed error messages or regular expressions. The SDK will try to match against the message, or exception type and value in the format `ExceptionType: value`.

lobsterkatie:
@souredoutlook - feel free to open a PR!

PR #4763, opened 2022-02-24
Adjust description of ignoreErrors

souredoutlook:
This PR clarifies the fact that `ignoreErrors` matches on `"ExceptionType: exceptionValue"` as one string.

Fixes https://github.com/getsentry/sentry-docs/issues/4762


lobsterkatie (approved):
👍🏻

PR #4767, opened 2022-02-26
ref(spans): Rename span exclusive time to span self time

udameli:
Rename `exclusive time` to `self time`. Self time conveys the meaning of the term better and, thus, should be used everywhere instead of old `exclusive time`, as was discussed by the performance team.

Mirrors changes in https://github.com/getsentry/sentry/pull/32049


priscilawebdev (approved):
lgtm

PR #4772, opened 2022-02-28
feat: Add very simple glossary [INGEST-901]

### LINKED ISSUE #3417: add terminology pop-ups
## Issue

This an issue with:

Platform/SDK documentation (all SDKs)

Sentry Product (all areas)

## Description

Adding key term definitions into the body of text can be clunky and longer to read.
A glossary can often be overwhelming and requires the reader to leave the page they're reading, which can interrupt their comprehension.
With many contributors, there is often no source of truth on the definition of a given term.

### Suggested Solution

Add pop-ups for terms that allow the user to stay on the same page. Also, these pop-ups would all be sourced from one file which would act as the source of truth for definitions.

To be completed: February 28, 2022

untitaker:
Fix #3417

Screenshot 2022-02-28 at 16 31 07

### DIFF
--- a/src/gatsby/plugins/gatsby-remark-terms/index.js
+++ b/src/gatsby/plugins/gatsby-remark-terms/index.js
@@ -23,8 +23,8 @@ const PATTERN = Object.keys(TERMS)
 const REGEX = new RegExp(`(\\b|\\W)(${PATTERN})(\\b|\\W)`);
 
 function replace(node) {
-  if (Object.keys(TERMS).length === 0) return;
   if (!node.children) return;
+  if (node.type == "heading") return;
 
   // If a text node is present in child nodes, check if an abbreviation is present
   for (let c = 0; c < node.children.length; c++) {
--- a/src/gatsby/plugins/gatsby-remark-terms/index.js
+++ b/src/gatsby/plugins/gatsby-remark-terms/index.js
@@ -1,0 +1,63 @@
+// WIP Plugin to define global glossary of terminology. Parts of remark-abbr
+// were vendored, particularly around auto-linking definitions and walking the
+// markdown AST. Currently the global glossary needs to be defined in-source.
+// It's unlikely that there is a way to define glossary in a separate markdown
+// file without blocking all pages processing on processing the glossary file.
+//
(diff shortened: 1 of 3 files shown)


imatwawana:
@untitaker is there a way for me to test this?

untitaker:
I forgot to post, if you check e.g. the term "DSN" here, you'll see the effects of having the DSN definition in the JS file: https://sentry-docs-git-feat-glossary.sentry.dev/product/cli/configuration/#configuration-values

imatwawana:
Ah okay. Looks good!

imatwawana (approved):

imatwawana:
I'm thinking to complete this PR, we remove the DNS entry and just push the structure. Then I can circle back and do a separate PR with all the actual terminology added to the JS file. What do you think?

imatwawana:
> @imatwawana I would start with at least a couple definitions instead of rolling everything out all at once, to gather feedback on UI
> 
> like I wouldn't spend effort on finishing _all_ the definitions without getting some preliminary user feedback on the general concept before/meanwhile

Ah okay, gotcha. I'll circle back on this PR and add in a couple more.

src/gatsby/plugins/gatsby-remark-terms/index.js
```diff
@@ -0,0 +1,66 @@
...
+// It's unlikely that there is a way to define glossary in a separate markdown
+// file without blocking all pages processing on processing the glossary file.
+//
+// remark-abbr is Copyright (c) Zeste de Savoir (https://zestedesavoir.com)
+// https://github.com/zestedesavoir/zmarkdown/tree/master/packages/remark-abbr
+
+const visit = require("unist-util-visit");
+
+const TERMS = {
+  DSN: "Data Source Name: Credentials used in SDK to send data to Sentry.",
```
imatwawana, line 13:
```suggestion
  DSN: "Data Source Name: Credentials used in SDK to send data to Sentry.",
  transaction: "A single instance of an activity you want to measure or track, like a page load, page navigation, or asynchronous task.",
  quota: "The monthly number of events — errors, attachments, and transactions — that you pay Sentry to track.",

```

Issue #4722, opened 2022-02-13
Undocumented Limitation of Lazy-Loading SDK


ArmoredCavalry:
### Core or SDK?

Platform/SDK

### Which part? Which one?

Javascript

### Description

I recently added version for release health / sessions to my JS SDK, but was only seeing sessions with errors. I ended up tracing this back to use of the lazy-loading script. Since it will only initialize the SDK if there is an error / exception, 100% of reported sessions were 'crashes'.

### Suggested Solution

I believe this should be added as a note under the Limitations section of the Lazy Loading docs -

https://docs.sentry.io/platforms/javascript/install/lazy-load-sentry/#limitations

This was when searching the docs for "sessions" there is a better chance of discovering why this behavior is happening.

getsentry-release:
Routing to @getsentry/{"message":"Not Found","documentation_url":"https://docs.github.com/rest/reference/teams#get-a-team-by-name"} for [triage](https://develop.sentry.dev/processing-tickets/#3-triage). ⏲️

lobsterkatie:
Sorry 'bout that! We [changed how `BrowserTracing` is accessed](https://github.com/getsentry/sentry-javascript/pull/4204), so the docs are right, or at least, describe what should be right. Unfortunately, the CDN bundle was missed in that PR. I've pushed a new PR to fix this.

lobsterkatie:
I'm pushing a fix now - thanks for pointing this out!

PR #4775, opened 2022-02-28
fix(browser): Mention release health in loader limitations section

lobsterkatie:
In our current implementation of the feature, a session and a pageload transaction are essentially one and the same, which means that release health doesn't work any better than tracing does when you're using the loader. This clarifies that in the loader's Limitations section.

Fixes https://github.com/getsentry/sentry-docs/issues/4722.

### DIFF
--- a/src/platforms/javascript/common/install/lazy-load-sentry.mdx
+++ b/src/platforms/javascript/common/install/lazy-load-sentry.mdx
@@ -99,6 +99,7 @@ The loader script includes a call to `Sentry.init` with one configuration option
 
 Because the loader injects the SDK asynchronously, only *unhandled errors* and *unhandled promise rejections* will be caught and buffered before the SDK is fully loaded. Specifically, the capturing of any [breadcrumb data](../../enriching-events/breadcrumbs/) will not be available until the SDK is fully loaded and initialized. To reduce the amount of time these features are unavailable, set `data-lazy="no"` or call `forceLoad()` as described above.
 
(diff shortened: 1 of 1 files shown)


src/platforms/javascript/common/install/lazy-load-sentry.mdx
```diff
@@ -99,6 +99,7 @@ The loader script includes a call to `Sentry.init` with one configuration option
...
 
-For similar reasons, the loader is not available in a form which includes performance monitoring. If you want to monitor the performance of your app, including pageload times, please [bundle the SDK with your app](../npm) or use our CDN, specifically the [bundle which includes tracing features](../cdn/#performance-bundle).
+For similar reasons, the loader is not available in a form which includes either performance monitoring or release health (sessions). If you want performance monitoriing, including pageload times, you can [bundle both `@sentry/browser` and `@sentry/tracing` with your app](../npm) or use our CDN, specifically the [bundle which includes tracing features](../cdn/#performance-bundle). For release health, bundling is the only option, though no additional package is needed.
```
imatwawana, line 102: a couple minor fixes

```suggestion
For similar reasons, the loader is not available in a form which includes either performance monitoring or release health (sessions). If you want performance monitoring, including pageload times, you can [bundle both `@sentry/browser` and `@sentry/tracing` with your app](../npm) or use our CDN, specifically the [bundle that includes tracing features](../cdn/#performance-bundle). For release health, bundling is the only option, though no additional package is needed.
```

imatwawana (approved):

AbhiPrasad (approved):

PR #4787, opened 2022-03-02
fix(typo): Angular custom span creation example

Lms24:
This PR fixes two typos in the code example to add a custom span to the active transaction. It changes a variable name in the declaration from `boostrapSpan` to `bootstrapSpan` and repositions the semicolon to the end of the promise chain.


Lms24:
Thanks @imatwawana for the review and for letting me know about this. With this small fix I'm very confident that it is correct, so if that's ok, I'm gonna merge it. I'll be sure to keep that in mind and tag a teammate for the next PR though.

PR #4792, opened 2022-03-03
fix(metrics): Remove Apdex ambiguity in tolerable cases

iker-barriocanal:
"between T and 4T" leads to ambiguity in inclusive/exclusive ranges,
that can only be resolved by reading the other alternatives. This PR
removes that ambiguity.


jjbayer (approved):
Not sure if this increases readability, but it is more precise, so 👍 .

PR #4804, opened 2022-03-07
feat(RN): Add artifact file names to RN troubleshooting docs

jennmueng:
Fixes https://github.com/getsentry/sentry-react-native/issues/2103


src/platforms/react-native/troubleshooting.mdx
```diff
@@ -86,6 +86,7 @@ Common fixes to this issue include:
 - Checking that both the `release` and `dist` and are set on the event by [setting it in the call to `init`](/platforms/react-native/configuration/releases/#bind-the-version).
 - Checking that source maps are correctly uploaded with the exact same `release` and `dist` values as the event. They **must match** for source maps to upload correctly.
 - If you set a custom `release` and `dist`, confirming that you manually uploaded the sourcemaps since our automatic source map upload script will not detect the custom values.
+- Make sure that the name of the bundle in the stack trace exactly matches with the name on the sourcemaps; i.e. if your stack frames say `app:///index.bundle`, the artifacts must be named `index.bundle` and `index.bundle.map` respectively.
```
imatwawana, line 89: We just try to avoid semi-colons in general, and we don't use Latin words or abbreviations.

```suggestion
- Make sure that the name of the bundle in the stack trace matches exactly with the name in the source map. For example, if your stack frames say `app:///index.bundle`, the artifacts must be named `index.bundle` and `index.bundle.map` respectively.
```

imatwawana (approved):
Just a couple style fixes

PR #4812, opened 2022-03-08
move event processors to common content

imatwawana:
Moved the Event Processors page that was under Node.js to the common content for all platforms so it can be shared by JavaScript and React Native.


AbhiPrasad (approved):

src/platforms/common/enriching-events/event-processors.mdx
```diff
@@ -4,7 +4,24 @@ sidebar_order: 300
...
+notSupported:
+  - android
+  - apple
+  - dart
+  - dotnet
+  - elixir
+  - flutter
+  - go
+  - java
+  - kotlin
```
imatwawana, line 18:
```suggestion
  - kotlin
  - native
```

PR #4827, opened 2022-03-11
feat(nextjs): Add `widenClientFileUpload` option

lobsterkatie:
This documents the option added in https://github.com/getsentry/sentry-javascript/pull/4705, which allows users to choose to cast a wider net when uploading sourcemaps for client-side code.

### DIFF
--- a/src/platforms/javascript/guides/nextjs/manual-setup.mdx
+++ b/src/platforms/javascript/guides/nextjs/manual-setup.mdx
@@ -217,7 +217,6 @@ In that case you can also skip the `sentry-cli` configuration step below.
 
 ### Use `hidden-source-map`
 
-
 _(New in version 6.17.1)_
 
 If you would like to use `hidden-source-map` rather than `source-map` as your webpack `devtool`, so that your built files do not contain a `sourceMappingURL` comment, add a `sentry` object to `moduleExports` above, and set the `hideSourceMaps` option to `true`:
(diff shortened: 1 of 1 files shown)


lforst (approved):
Hi Katie, the new section reads excellent and also explains our intention behind the added option really well.

I also peeked over at https://github.com/getsentry/sentry-javascript/pull/4705 and the changes there also LGTM. (love the tests)

I have one tiny nit but consider this PR approved: `widenClientFileUpload` doesn't explain too well what it does. The main question I would have while seeing this option for the first time would be: To what scale am I widening the file upload? What files are now included?

**However,** nobody is gonna use this option unless they have read the documentation behind it, so I don't think it really matters.

src/platforms/javascript/guides/nextjs/manual-setup.mdx
```diff
@@ -232,6 +231,22 @@ const moduleExports = {
 
 Note that this only applies to client-side builds, and requires the `SentryWebpackPlugin` to be enabled.
 
+### Widen the Upload Scope
+
+_(New in version 6.18.3)_
+
+If you find that there are in-app frames in your client-side stacktraces which aren't getting sourcemapped even when most others are, it is likely because they are from files in `static/chunks/` rather than `static/chunks/pages/`. By default, such files aren't uploaded because the majority of the files in `static/chunks/` only contain nextjs or third-party code, and are named in such a way that it's hard to tell relevant files (ones containing your code) apart from irrelevant ones.
```
imatwawana, line 238: I'm worried about the words "tell" and "apart" being separated in a somewhat idiomatic phrase, so I've tried to simplify the sentence a bit. Also, should "nextjs" be "Next.js"? Not sure if there's a convention I'm unaware of here.

```suggestion
If you find that there are in-app frames in your client-side stack traces that aren't getting source-mapped even when most others are, it's likely because they are from files in `static/chunks/` rather than `static/chunks/pages/`. By default, such files aren't uploaded because the majority of the files in `static/chunks/` only contain nextjs or third-party code, and are named in such a way that it's hard to recognize relevant files (ones containing your code) among irrelevant ones.
```

imatwawana (approved):
Some small wording stuff, but otherwise fine. If you don't like the suggestion I made for replacing "tell...apart", let me know and we can figure out something better.

src/platforms/javascript/guides/nextjs/manual-setup.mdx
lobsterkatie: Went with "distinguish between," but otherwise used your edits.

PR #4832, opened 2022-03-14
fix(unity): Removed obsolete iOS native docs

bitsandfoxes:
The SDK is resolving this issue now internally and does no longer require special case documentation.


imatwawana:
I'm wondering if you should add redirects for those removed pages to relevant existing pages so folks who might have had them bookmarked don't end up getting a 404?

PR #4843, opened 2022-03-16
unity: remove not needed doc

bruno-garcia:
From version 0.12.0 we don't need this anymore. And since doing this makes updating the SDK, we should get rid of these instructions now


bitsandfoxes:
Dup of: #4832

PR #4846, opened 2022-03-17
Correct method name

carlobeltrame:
According to https://docs.sentry.io/platforms/javascript/usage/ it should be `captureException`.


AbhiPrasad (approved):
Yup, good catch! Thank you for your contribution!

PR #4847, opened 2022-03-17
chore(android): Update SAGP version to stable 3.0.0

romtsn:


marandaneto:
@romtsn let's add a note in the docs and wizard that if you are using AGP < 7.x, you have to stick to SAGP 2.x

romtsn:
@marandaneto where exactly would you add it? There's nothing on the wizard about AGP version. I can only think of the gradle page, but we're saying there that 3.0.0 requires AGP 7.0.0. Or do you want to extend the note?

marandaneto:
> @marandaneto where exactly would you add it? There's nothing on the wizard about SAGP version. I can only think of the gradle page, but we're saying there that 3.0.0 requires AGP 7.0.0. Or do you want to extend the note?

https://github.com/getsentry/sentry-docs/blob/master/src/wizard/android/index.md

> This feature is available in the Beta release...

I'd rather remove the note above or adapt to not be Beta anymore.

I'd remove the comment below from every Gradle Plugin option since it's stable now:

> Only available v3.0.0 and above.  

The note below is fine, but I'd rather comment that you should stick to the latest 2.x if you are not using AGP 7.x as well, otherwise it may pass the message that we don't support older versions.

> The io.sentry.android.gradle >= 3.0.0 requires [Android Gradle Plugin >= 7.0.0](https://developer.android.com/studio/releases/gradle-plugin#7-0-0).

romtsn:
> I'd rather remove the note above or adapt to not be Beta anymore.

I've filed https://github.com/getsentry/sentry-docs/pull/4850/ to address this (and also prompt to try out auto-installation).

The rest sounds good, will address shortly 👍

marandaneto (approved):

src/platforms/android/common/gradle.mdx
```diff
@@ -34,12 +34,12 @@ buildscript {
...
 
 plugins {
-    id("io.sentry.android.gradle") version "{{ packages.version('sentry.java.android.gradle-plugin', '2.0.0') }}"
+    id("io.sentry.android.gradle") version "{{ packages.version('sentry.java.android.gradle-plugin', '3.0.0') }}"
 }
 ```
 <Note>
 
-The `io.sentry.android.gradle` >= `3.0.0` requires [Android Gradle Plugin >= 7.0.0](https://developer.android.com/studio/releases/gradle-plugin#7-0-0).
+The `io.sentry.android.gradle` >= `3.0.0` requires [Android Gradle Plugin >= 7.0.0](https://developer.android.com/studio/releases/gradle-plugin#7-0-0). For an older version of the Android Gradle Plugin, consider using `io.sentry.android.gradle:2.+`.
```
philipphofmann, line 42: `l`: Do we know what the older version is? Can we be more specific here?

philipphofmann (approved):
LGTM 🚀

src/platforms/android/common/gradle.mdx
imatwawana, line 42: Wondering if this can be generalized and if we can be more direct.

```suggestion
The `io.sentry.android.gradle` >= `3.0.0` requires [Android Gradle Plugin >= 7.0.0](https://developer.android.com/studio/releases/gradle-plugin#7-0-0). For older versions of the Android Gradle Plugin, use `io.sentry.android.gradle:2.+`.
```

src/platforms/android/migration.mdx
```diff
@@ -12,7 +12,7 @@ Starting from version `5.6.2`, the `Timber.tag` usage is no longer supported by
 
 ## Migrating from `io.sentry:sentry-android-gradle-plugin 2.x` to `io.sentry:sentry-android-gradle-plugin 3.0.0`
 
-The `io.sentry.android.gradle` >= `3.0.0` requires [Android Gradle Plugin >= 7.0.0](https://developer.android.com/studio/releases/gradle-plugin#7-0-0).
+The `io.sentry.android.gradle` >= `3.0.0` requires [Android Gradle Plugin >= 7.0.0](https://developer.android.com/studio/releases/gradle-plugin#7-0-0). For an older version of the Android Gradle Plugin, consider using `io.sentry.android.gradle:2.+`.
```
imatwawana, line 15: Same comment as in gradle doc

imatwawana (approved):
made a couple suggestions

src/platforms/android/migration.mdx
romtsn: good spot, thank you 🙏

src/platforms/android/common/gradle.mdx
romtsn: Yes, that looks good, thank you!

src/platforms/android/migration.mdx
romtsn, line 15:
```suggestion
The `io.sentry.android.gradle` >= `3.0.0` requires [Android Gradle Plugin >= 7.0.0](https://developer.android.com/studio/releases/gradle-plugin#7-0-0). For older versions of the Android Gradle Plugin, use `io.sentry.android.gradle:2.+`.
```

src/platforms/android/common/gradle.mdx
romtsn: it's basically many, so Isabel's suggestion made it clearer I think

PR #4850, opened 2022-03-18
feat(android): Remove beta note about Sentry Android Gradle plugin from the wizard

romtsn:


src/wizard/android/index.md
```diff
@@ -116,9 +116,9 @@ transaction.finish() // Mark the transaction as finished and send it to Sentry
 
 Check out [the documentation](https://docs.sentry.io/platforms/android/performance/instrumentation/) to learn more about the API and automatic instrumentations.
 
-> Want to play with some new features? Try out our beta [Room](https://docs.sentry.io/platforms/android/configuration/integrations/room-and-sqlite/) and [file I/O](https://docs.sentry.io/platforms/android/configuration/integrations/file-io/) performance integrations.
```
marandaneto, line 119: Should this block then be at the very beginning of the Wizard? since it's about the installation steps, if somebody came till here, they already have installed it manually following the above steps.
@bruno-garcia ? it's a risky move tho since it might have side effects.

PR #4856, opened 2022-03-19
ref(unreal): Move Unreal to top-level platform

GloryOfNight:
Adding unreal platform and providing some documentation to it.


imatwawana (commented):
I know you haven't marked this ready for review yet, but since the PR is pretty big, I wanted to take an early look so I can review for wording and style more quickly later.  I've left a few comments. 😃

src/docs/contributing/platforms/index.mdx
```diff
@@ -110,8 +110,6 @@ If you wanted to make hide the content for a platform by default, but enable it
 
 ```markdown
 ---
-supported:
-  - native.ue4
```
imatwawana, line 114: For this text to make sense, there still has to be a guide in the `supported` part of the example. It just can't be Unreal since Unreal is now a top-level platform. It has to be one of the frameworks under Native. It doesn't matter which one, but there has to be one there. I had chosen `wasm`, but you can use any of them.

GloryOfNight: ok, thanks! For now I'll add wasm. I probably should discuss with Bruno what will fit here the best.

PR #4864, opened 2022-03-22
fix(py): Clean up webhook snippets

mgaeta:


src/docs/product/integrations/integration-platform/webhooks.mdx
```diff
@@ -51,15 +51,20 @@ function verifySignature(request, secret = "") {
...
+
+expected = request.headers['sentry-hook-signature']
+body = json.dumps(request.body)
+digest = hmac.new(
+    key=client_secret.encode('utf-8'),
+    msg=body,
+    digestmod=hashlib.sha256,
+).hexdigest()
+
+if digest != expected:
```
leeandher, line 66: IMO think it'd be more effective to change the variable name to `expected_digest` and declare it after L60.

leeandher: > after L60
I meant after `digest`
```python
digest = ...
expected_digest = ...
```

mgaeta: I prefer to collect all params from `request` together at the top of the function.

leeandher (approved):

PR #4867, opened 2022-03-22
docs(discover): Adds some clarification around the message key search

edwardgou-sentry:
The `message:` token appears in multiple areas of Sentry but searching on `message:` behaves differently in different areas. This doc update is to provide clarification on the behaviour of `message:` for event based searches.

Areas that have searching on `message:` but with different behaviour include: Issue Alert Rules and Fingerprint Rules
This PR doesn't address those areas.
image


jas-kas:
If we want to provide an example, I think it should be a separate line item, similar to `event.timestamp` on this page, for clarity. Not after a colon in the same sentence imo.

Everything else looks good! :)

jas-kas (approved):

src/docs/product/sentry-basics/search/searchable-properties.mdx
```diff
@@ -46,7 +46,7 @@ Below is a list of event-level keys and tokens reserved and known to Sentry:
 | Key/Token                     | Description |
 |---------------------------| ----------- |
 | location | Restrict results to the events with a matching location. |
-| message | Restrict results to events with a matching message. |
+| message | Restrict results to events with a matching message. This filter will also match on any message containing the supplied value.<br /><br /> Searching `message:undefined` will match an event with a message of `undefined is not an object`.<br /><br />  Raw token searches are searched against `message`. |
```
imatwawana, line 49: I'm not sure what you mean by "raw token searches". I'm thinking maybe searches that have no key and only a value, but I'm not sure. Is there another way we can say that?

imatwawana (commented):
Just one point of clarification

src/docs/product/sentry-basics/search/searchable-properties.mdx
edwardgou-sentry: Hey Isabel! Yup that's what we're trying to mean with "raw token searches".
I took a look at some of the existing docs for this kind of search and it seems they're referred to as "optional raw searches". Should we also use this terminology here?

```suggestion
| message | Restrict results to events with a matching message. This filter will also match on any message containing the supplied value.<br /><br /> Searching `message:undefined` will match an event with a message of `undefined is not an object`.<br /><br />  Optional raw searches are searched against `message`. |
```

imatwawana: Since we only use that terminology once in the docs, I'd say let's be really explicit here, so folks don't have to figure this out. 

```suggestion
| message | Restrict results to events with a matching message. This filter will also match on any message containing the supplied value.<br /><br /> Searching `message:undefined` will match an event with a message of `undefined is not an object`.<br /><br />  Raw text searches (searches without the `message` key) are also checked against this field. |
```

edwardgou-sentry: 👍  Thanks Isabel! That update looks good to me, I've committed it!

PR #4869, opened 2022-03-23
Document Gradle experimentalGuardsquareSupport

marandaneto:


romtsn:
LGTM! we also have a section for `ProGuard/R8` below, should we change it to `ProGuard/R8/DexGuard` or `ProGuard/R8 & DexGuard` and update the wording there a bit?

romtsn:
LGTM! we also have a section for `ProGuard/R8` below, should we change it to `ProGuard/R8/DexGuard` or `ProGuard/R8 & DexGuard` and update the wording there a bit?

romtsn (approved):

PR #4893, opened 2022-04-01
Document tracing-origins

maciejwalkowiak:


maciejwalkowiak:
I am not sure if this is the right moment to merge it. Java documentation has duplicated content for configuration options:

- https://docs.sentry.io/platforms/java/configuration/
- https://docs.sentry.io/platforms/java/configuration/options/

Perhaps this should be cleaned up first.

cc @bruno-garcia @adinauer

marandaneto (approved):

imatwawana (approved):
Added a wording suggestion

adinauer (approved):
lgtm; not sure about our plans regarding the two pages

marandaneto:
> > lgtm; not sure about our plans regarding the two pages
> 
> The Java docs have been mostly refactored, so I'm not sure why the Configuration page here isn't using the common content. Perhaps Bruno knows. IMO, if there's any time pressure about this PR, I would just merge it and log an issue for this duplication so we don't lose track of it.

I think the big difference is that the common options tell you that property X exists, while the other page tells you how to do it, which file to change, the format, the nuances of the flag, etc...
I also don't like duplicated stuff, but if we find a better way how to do it, I'd do in a different PR., so merging it.

PR #4896, opened 2022-04-01
Use secret fossa api key when available

chadwhitacre:


chadwhitacre:
Looks good!

https://github.com/getsentry/sentry-docs/runs/5795979314?check_suite_focus=true#step:2:853

PR #4898, opened 2022-04-03
add Serverless Cloud base guide

mwarkentin:
Adds a Sentry Guide for [Serverless Cloud](https://www.serverless.com/cloud) ([docs](https://www.serverless.com/cloud/docs)).

Serverless Cloud's API functionality is based on ExpressJS, and allows us to use many of the same concepts (with a few small tweaks).

I haven't tested out any of the advanced performance monitoring stuff yet, so I haven't touched that part of the guide yet.

### DIFF
--- a/src/platforms/node/guides/serverless-cloud/index.mdx
+++ b/src/platforms/node/guides/serverless-cloud/index.mdx
@@ -1,0 +1,164 @@
+---
+title: Serverless Cloud
+redirect_from:
+  - /clients/node/integrations/serverless-cloud/
+  - /platforms/node/serverless-cloud/
+description: "Learn about using Sentry with Serverless Cloud."
+---
+
+Add `@sentry/node` as a dependency:
+
+```bash {tabTitle:npm}
+cloud install @sentry/node
+```
+
+Sentry should be initialized as early in your app as possible:
+
+```javascript
+import api from "@serverless/cloud";
+import * as Sentry from "@sentry/node";
+
+// or using CommonJS
+// const api = require("@serverless/cloud");
+// const Sentry = require('@sentry/node');
+
+Sentry.init({
+  dsn: "___PUBLIC_DSN___",
+  // or pull from params
+  // dsn: params.SENTRY_DSN,
+  environment: params.INSTANCE_NAME,
+});
+
+// The request handler must be the first middleware on the app
+api.use(Sentry.Handlers.requestHandler());
+
+// All controllers should live here
+api.get("/", function rootHandler(req, res) {
+  res.end("Hello world!");
+});
+
+// The error handler must be before any other error middleware and after all controllers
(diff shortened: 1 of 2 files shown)


src/platforms/node/guides/serverless-cloud/index.mdx
```diff
@@ -0,0 +1,164 @@
+---
+title: Serverless Cloud
+redirect_from:
+  - /clients/node/integrations/serverless-cloud/
+  - /platforms/node/serverless-cloud/
```
imatwawana, line 5: Since this is a new page, no redirect is required. They're used to ensure users can still get to a page when it's been moved. 
```suggestion
```

AbhiPrasad:
Content looks good 👍. 

I'm not sure what our philosophy is to add new platforms to docs (like do we prefer this exists in serverless's docs instead?), I'll defer to @smeubank about that.

imatwawana:
I've added the SDK team as reviewers, but if you weren't looking for this to be reviewed yet, then you can [put it into Draft](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft) and keep working on it. What you've added covers the installation part, but the SDK folks will have to determine if the other pages that automatically show up for this guide are accurate/sufficient or if this entails a heavier lift and/or if this belongs here.

sl0thentr0py:
thx for the PR @mwarkentin, the content looks good to me as well.
I do want to raise one question here though (cc: @AbhiPrasad @antonpirker @smeubank).

We have multiple ways in which people can use sentry in an AWS/serverless setting.
* Manual import/init as done here
* Adding the sentry layer via [serverless.yml](https://docs.sentry.io/product/integrations/cloud-monitoring/aws-lambda/#serverless-framework)
* Adding the layer via [UI in Sentry](https://docs.sentry.io/product/integrations/cloud-monitoring/aws-lambda/#install)

While all of these work in principle, I think it also creates confusion for a new user if the docs contain all of these various methods. We should really push one method imo, which one I'm more or less neutral about. Especially once we have the extension running, I think we should push the layer one consistently for both AWS and serverless.yml.

Thoughts and opinions?

mwarkentin:
@sl0thentr0py Good question, and I probably could have clarified this more. While [Serverless Cloud](https://www.serverless.com/cloud) is from the same folks who make the [Serverless Framework](https://www.serverless.com/framework), it's really it's own product which abstracts away all of the AWS stuff (there is no `serverless.yml`, the resources are deployed into Serverless' AWS accounts, not your own, and so on).

So the Sentry layer would not be possible to use within Serverless Cloud (unless they expose it somehow down the line or build a native integration for us).

sl0thentr0py (approved):

PR #4905, opened 2022-04-05
feat(dashboards): Doc improvements for Dashboard Widget Dataset Selector

edwardgou-sentry:
Adjusts some of the text in the Dataset Selector docs to be more descriptive

image

### DIFF
--- a/src/docs/product/dashboards/custom-dashboards/index.mdx
+++ b/src/docs/product/dashboards/custom-dashboards/index.mdx
@@ -89,4 +89,15 @@ transactions'.
 
 ## Data Set Selection
 
-Select between events and [issues](/product/issues/) data in a widget with the "Data Set" selector. Choosing "Issues" allows you to query using issues filters, such as `is:unresolved` for unresolved issues, and to display and sort by issues fields. This feature only applies to [table](#table-results) visualizations.
+Select between events and [issues](/product/issues/) data in a widget with the "Data Set" selector.
+
+Choosing "All Events" allows you to query and aggregate individual errors and transactions in the same way you would for a [Discover query](/product/discover-queries/). This is data set fits most common use cases. Some widget examples include:
+- Tracking performance on an endpoint.
+- Amount of activity by country.
+- Users most affected by errors.
+
+Choosing "Issues" allows you to query using issues filters, such as `is:unresolved` for unresolved issues, and to display and sort by issues fields. Query filters are only applied to the issue group itself and not on the individual error events that make up an issue. Choose this data set if you want to customize a filtered list of issues on your dashboard. Some widget examples include:
+- Issues assigned to your team.
+- Most frequently occuring issues.
+
+The "Issues" data set is only available [table](#table-results) visualization widgets.


src/docs/product/dashboards/custom-dashboards/index.mdx
```diff
@@ -89,4 +89,15 @@ transactions'.
 
 ## Data Set Selection
 
-Select between events and [issues](/product/issues/) data in a widget with the "Data Set" selector. Choosing "Issues" allows you to query using issues filters, such as `is:unresolved` for unresolved issues, and to display and sort by issues fields. This feature only applies to [table](#table-results) visualizations.
+Select between events and [issues](/product/issues/) data in a widget with the "Data Set" selector.
```
narsaynorath, line 92: Should we link to some events docs like: https://docs.sentry.io/product/accounts/pricing/#events ?
Kind of weird to link to the pricing page. Isabel has an entry for `event` under key terms, maybe we just sync up later and fix the link

src/docs/product/dashboards/custom-dashboards/index.mdx
```diff
@@ -89,4 +89,15 @@ transactions'.
...
+
+Choosing "All Events" allows you to query and aggregate individual errors and transactions in the same way you would for a [Discover query](/product/discover-queries/). This is data set fits most common use cases. Some widget examples include:
+- Tracking performance on an endpoint.
```
narsaynorath, line 95:
```suggestion
- Tracking performance of an endpoint.
```

src/docs/product/dashboards/custom-dashboards/index.mdx
```diff
@@ -89,4 +89,15 @@ transactions'.
 
 ## Data Set Selection
 
-Select between events and [issues](/product/issues/) data in a widget with the "Data Set" selector. Choosing "Issues" allows you to query using issues filters, such as `is:unresolved` for unresolved issues, and to display and sort by issues fields. This feature only applies to [table](#table-results) visualizations.
+Select between events and [issues](/product/issues/) data in a widget with the "Data Set" selector.
+
+Choosing "All Events" allows you to query and aggregate individual errors and transactions in the same way you would for a [Discover query](/product/discover-queries/). This is data set fits most common use cases. Some widget examples include:
```
priscilawebdev, line 94:
```suggestion
Choosing "All Events" allows you to query and aggregate individual errors and transactions in the same way you would for a [Discover query](/product/discover-queries/). This data set fits the most common use cases. Some widget examples include:
```

src/docs/product/dashboards/custom-dashboards/index.mdx
```diff
@@ -89,4 +89,15 @@ transactions'.
...
+- Tracking performance on an endpoint.
+- Amount of activity by country.
+- Users most affected by errors.
+
+Choosing "Issues" allows you to query using issues filters, such as `is:unresolved` for unresolved issues, and to display and sort by issues fields. Query filters are only applied to the issue group itself and not on the individual error events that make up an issue. Choose this data set if you want to customize a filtered list of issues on your dashboard. Some widget examples include:
```
priscilawebdev, line 99:
```suggestion
Choosing "Issues" allows you to query using issues filters, such as `is:unresolved` for unresolved issues, and to display and sort by issues fields. Query filters are only applied to the issue group itself and not to the individual error events that make up an issue. Choose this data set if you want to customize a filtered list of issues on your dashboard. Some widget examples include:
```

src/docs/product/dashboards/custom-dashboards/index.mdx
edwardgou-sentry: Hmm, took a look around at what existing events doc we have. I think the closest thing is this page? https://docs.sentry.io/product/sentry-basics/enrich-data/
But it contains quite a lot of info outside of just events so I'm not sure if it would be confusing to users.

The Key Terms pr got merged here https://docs.sentry.io/product/accounts/quotas/#key-terms which we could link to as well. 

I think the first link might be more suitable since the Key Terms page is more around quota management or we could leave it unlinked for now. I don't have a strong opinion on this so maybe @jas-kas can make the call?

jas-kas: I don't think it's necessary to link an Events doc, since we already link to Discover documentation and that would be closest representation of what we're referring to. Additionally, we're also changing this name from "Events" to "Errors and Transactions" in the near future as well.

jas-kas (commented):
Commit suggestions were a result of...

- Removing redundant phrases
- "Query filters are only applied to the issue group itself and not to the individual error events that make up an issue" - This is somewhat correct, however it doesn't acknowledge that you can search for specific error properties and we'll pick up issues that contain such errors
- Calling out issue properties v. event properties
- Adding more examples

src/docs/product/dashboards/custom-dashboards/index.mdx
imatwawana: There is a definition on this page:
https://docs.sentry.io/product/sentry-basics/key-terms/

However, you can't link directly to the definition, so I'd agree with Jasmin to not worry about linking out at this point.

src/docs/product/dashboards/custom-dashboards/index.mdx
```diff
@@ -89,4 +89,15 @@ transactions'.
 
 ## Data Set Selection
 
-Select between events and [issues](/product/issues/) data in a widget with the "Data Set" selector. Choosing "Issues" allows you to query using issues filters, such as `is:unresolved` for unresolved issues, and to display and sort by issues fields. This feature only applies to [table](#table-results) visualizations.
+Select between events and [issues](/product/issues/) data in a widget with the "Data Set" selector.
+
+Choosing "All Events" allows you to query and aggregate error and transaction events in the same way you would for a [Discover query](/product/discover-queries/). Some widget examples include:
+- Tracking performance of an endpoint.
+- Throughput by country.
+- Users most affected by errors.
```
imatwawana, line 97: Since none of these are actually sentences, removing the period.

```suggestion
- Tracking performance of an endpoint
- Throughput by country
- Users most affected by errors
```

src/docs/product/dashboards/custom-dashboards/index.mdx
```diff
@@ -89,4 +89,15 @@ transactions'.
...
+
+Choosing "Issues" allows you to query issue properties, such as `is:unresolved`, event properties such as `error.handled:0`, and to sort by issue fields such as `First Seen`. When searching on event properties, the search will return any issue that has one or more events matching the supplied event filters. Choose this data set if you want to customize a list of issues on your dashboard. Some widget examples include:
```
imatwawana, line 99:
```suggestion
Choosing "Issues" allows you to query issue properties, such as `is:unresolved` or event properties such as `error.handled:0`, and to sort by issue fields such as `First Seen`. When searching on event properties, the search will return any issue that has one or more events matching the supplied event filters. Choose this data set if you want to customize a list of issues on your dashboard. Some widget examples include:
```

src/docs/product/dashboards/custom-dashboards/index.mdx
```diff
@@ -89,4 +89,15 @@ transactions'.
...
+- Issues assigned to your team.
+- Most frequently occuring issues.
```
imatwawana, line 101:
```suggestion
- Issues assigned to your team
- Most frequently occurring issues
```

imatwawana (approved):
A couple small grammar and wording things, but otherwise looks good!

PR #4911, opened 2022-04-07
Update expo.mdx

tompap:
Missing information about override of url in config settings


marandaneto:
@tompap Thanks for doing this, but this SDK/page has already been deprecated, See the note on the top of the page -> https://docs.sentry.io/clients/react-native/
Maybe make a PR to the expo docs -> https://docs.expo.dev/guides/using-sentry/#step-3-app-config=

PR #4917, opened 2022-04-07
Update options.mdx

timlawrenz:
I think that feature doesn't exist anymore. I tried using this in an example app, and checked https://github.com/getsentry/sentry-ruby/blob/master/sentry-ruby/lib/sentry/configuration.rb.

### DIFF
--- a/src/docs/clients/ruby/configuration/options.mdx
+++ b/src/docs/clients/ruby/configuration/options.mdx
@@ -246,14 +246,6 @@ config.silence_ready = true
 config.ssl_verification = false
 ```
 
-`tags`
-
-: Default tags to send with each event.
-
-```ruby
-config.tags = { foo: :bar }
-```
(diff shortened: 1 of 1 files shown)


sl0thentr0py:
Hi @timlawrenz, thx for the PR but those docs are visible in the [Legacy Client section for the old raven-ruby gem](https://docs.sentry.io/clients/ruby/configuration/options/) and are hence correct. Though that's a good point that we should clarify on top of the page that this is for the old gem. Will make another PR for that.

sl0thentr0py:
added https://github.com/getsentry/sentry-docs/pull/4919

timlawrenz:
Thank you! And sorry for the confusion.

sl0thentr0py:
no worries and thank you! you still helped us clean up our docs :)

Issue #4915, opened 2022-04-07
Wrong left header behaviour in "Contributing to docs" section


stefanosiano:
### Core or SDK?

Core Sentry product

### Which part? Which one?

Contributing to docs

### Description

I was having a look at the “Contributing to docs” page https://docs.sentry.io/contributing/
I found an issue in the headers on the left, which occurs only when i’m visiting one of the “root” pages of that header. 

Basically if I’m in one of:
https://docs.sentry.io/contributing/environment/
https://docs.sentry.io/contributing/approach/
https://docs.sentry.io/contributing/platforms/
https://docs.sentry.io/contributing/linking/
https://docs.sentry.io/contributing/pages/
https://docs.sentry.io/contributing/onboarding-wizard/

If i reload the page the header is not highlighted anymore, and the link shown from the browser on mouse hover (i use chrome and opera) is different from what it should be:
should be https://docs.sentry.io/contributing/ but it’s https://docs.sentry.io/product/
should be https://docs.sentry.io/contributing/environment/ but it’s https://docs.sentry.io/product/sentry-basics/
should be https://docs.sentry.io/contributing/approach/ but it’s https://docs.sentry.io/product/issues/
should be https://docs.sentry.io/contributing/platforms/ but it’s https://docs.sentry.io/product/projects/
should be https://docs.sentry.io/contributing/linking/ but it’s https://docs.sentry.io/product/alerts/
should be https://docs.sentry.io/contributing/pages/ but it’s https://docs.sentry.io/product/performance/
should be https://docs.sentry.io/contributing/onboarding-wizard/ but it’s https://docs.sentry.io/product/releases/

When i click on the links the correct page is shown, but if i cmd+click to open the link in a new page (i’m used to do it) i am redirected to the wrong page.

when i try to disable javascript the left header is the general docs header, not the contributing one.

### Suggested Solution

I guess the contributing header is loaded through javascript after page load, and while it doesn't change the links, it catches the clicks to open the correct page.
This is just a guess, obviously.

imatwawana:
See issue #4580

stephanie-anderson:
This behaviour seems to be fixed due to the migration from Gatsby to Next.js

PR #4918, opened 2022-04-07
Fix sidebar reconciliation issue

fitiskin:
#4915

My guess is that the `<Sidebar />` subtree is incorrectly reused during reconciliation on client-side (perhaps because JSX composition is same). I don't know exactly why.

[src/components/layout.tsx](https://github.com/getsentry/sentry-docs/blob/ad0cd9a3d12cbcee1faffe83c730bfc030abe9ae/src/components/layout.tsx#L46):
```jsx
<div className="text-white p-3">
  {sidebar ? sidebar : <Sidebar />}
</div>
```

Most likely my fix is not optimal. As far as I know it will be lazily evaluated when used as default function parameter value.

### DIFF
--- a/src/components/layout.tsx
+++ b/src/components/layout.tsx
@@ -24,7 +24,7 @@ type Props = {
 
 export default ({
   children,
-  sidebar,
+  sidebar = <Sidebar />,
   pageContext = {},
 }: Props): JSX.Element => {
   const searchPlatforms = [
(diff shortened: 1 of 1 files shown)


untitaker (approved):
this appears to work but I have no idea why.

untitaker:
Thanks!

untitaker:
we had to revert this fix in #4935 because it made a sidebar somewhere else disappear

PR #4931, opened 2022-04-12
ref(rate limits): Add concurrent headers

ceorourke:
Update the rate limit documentation to include the concurrent rate limiter headers.

### DIFF
--- a/src/api/ratelimits.mdx
+++ b/src/api/ratelimits.mdx
@@ -5,9 +5,15 @@ sidebar_order: 3
 
 Sentry rate limits every API request made to prevent abuse and resource overuse. The limit is applied to each unique combination of caller and endpoint.
 
-We restrict both how frequently a request is made (requests per second rate limit) and how many concurrent requests a caller can make (concurrent rate limit). The requests per second rate limit follows a fixed window approach. Requests are counted into time buckets, determined by the window size. If the number of requests from a caller exceeds the number of allowed requests within a window, then that request will be rejected.  Meanwhile, the concurrent rate limiter will reject requests if the caller has too many requests in progress at the same time.
 
-While there is a default rate limit of **40 requests per second** for the requests per second rate limit, each endpoint can have their own maximum number of requests and window size. You can track your rate limit usage by looking at special headers in the response.
+We restrict both how frequently a request is made (requests per second rate limit) and how many concurrent requests a caller can make (concurrent rate limit).
+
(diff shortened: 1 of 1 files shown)


src/api/ratelimits.mdx
```diff
@@ -5,7 +5,8 @@ sidebar_order: 3
 
 Sentry rate limits every API request made to prevent abuse and resource overuse. The limit is applied to each unique combination of caller and endpoint.
 
-The rate limit follows a fixed window approach. Requests are counted into time buckets, determined by the window size. If the number of requests from a caller exceeds the number of allowed requests within a window, then that request will be rejected.
+The rate limit follows a fixed window approach. Requests are counted into time buckets, determined by the window size. The concurrent rate limit limits the number of requests happening at the same time to a given endpoint grouping. If the number of requests from a caller exceeds the number of allowed requests or concurrent requests within a window, then that request will be rejected.
+
```
ceorourke, line 9: Ugh words are hard, especially when I barely have a grasp on the concept. Sorry if this sucks.

src/api/ratelimits.mdx
```diff
@@ -5,7 +5,7 @@ sidebar_order: 3
 
 Sentry rate limits every API request made to prevent abuse and resource overuse. The limit is applied to each unique combination of caller and endpoint.
 
-The rate limit follows a fixed window approach. Requests are counted into time buckets, determined by the window size. If the number of requests from a caller exceeds the number of allowed requests within a window, then that request will be rejected.
+The rate limit follows a fixed window approach. Requests are counted into time buckets, determined by the window size. If the number of requests from a caller exceeds the number of allowed requests within a window, then that request will be rejected. We restrict both how frequently a request is made and how many requests are occurring within same time period.
```
AniketDas-Tekky, line 8:
```suggestion
We restrict both how frequently a request is made and how many concurrent requests a caller can make.

The former rate limit follows a fixed window approach. Requests are counted into time buckets, determined by the window size. If the number of requests from a caller exceeds the number of allowed requests within a window, then that request will be rejected.  Meanwhile, the concurrency limit will reject requests if the user has too many requests in progress at the same time.
```

I don't know if this is the best way to phrase it. It might be worth getting Angela involved to get the correct wording.

src/api/ratelimits.mdx
```diff
@@ -5,7 +5,7 @@ sidebar_order: 3
...
 
 While there is a default rate limit of **40 requests per second**, each endpoint can have their own maximum number of requests and window size. You can track your rate limit usage by looking at special headers in the response.
```
AniketDas-Tekky, line 10: This is going to be confusing now.  We might just need to have separate sections on this page for RPS and CRL.

AniketDas-Tekky: Specifically the `40 requests per second`

AniketDas-Tekky (approved):

PR #4940, opened 2022-04-14
[JS v7] **do not merge** ref(javascript): Replace deprecated severity enums with string literals

lforst:
In https://github.com/getsentry/sentry-javascript/pull/4926, we deprecated the `Severity`. This PR replaces all occurrences of that enum with their now recommended string value counterpart.

Ref: https://getsentry.atlassian.net/browse/WEB-849

**DO NOT MERGE: Only merge this branch after the Sentry JavaScript SDK version 7 has been published.**

### DIFF
--- a/src/platforms/javascript/common/troubleshooting/index.mdx
+++ b/src/platforms/javascript/common/troubleshooting/index.mdx
@@ -395,12 +395,12 @@ document.body.addEventListener(
     if (event.target.tagName === "IMG") {
       Sentry.captureMessage(
         `Failed to load image: ${event.target.src}`,
-        Sentry.Severity.Warning
+        "warning" // message severiity level
       );
     } else if (event.target.tagName === "LINK") {
       Sentry.captureMessage(
         `Failed to load css: ${event.target.href}`,
-        Sentry.Severity.Warning
+        "warning" // message severiity level
       );
     }
   },
--- a/src/platforms/javascript/common/troubleshooting/index.mdx
+++ b/src/platforms/javascript/common/troubleshooting/index.mdx
@@ -395,12 +395,12 @@ document.body.addEventListener(
     if (event.target.tagName === "IMG") {
       Sentry.captureMessage(
         `Failed to load image: ${event.target.src}`,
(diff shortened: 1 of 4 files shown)


src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -389,12 +389,12 @@ document.body.addEventListener(
     if (event.target.tagName === "IMG") {
       Sentry.captureMessage(
         `Failed to load image: ${event.target.src}`,
-        Sentry.Severity.Warning
+        "warning" // message severiity level
```
Lms24, line 392:
```suggestion
        "warning" // message severity level
```

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -389,12 +389,12 @@ document.body.addEventListener(
...
       );
     } else if (event.target.tagName === "LINK") {
       Sentry.captureMessage(
         `Failed to load css: ${event.target.href}`,
-        Sentry.Severity.Warning
+        "warning" // message severiity level
```
Lms24, line 397:
```suggestion
        "warning" // message severity level
```

Lms24 (commented):
Looking good. I wonder if we should include a list of the possible `SeverityLevel`s? Or is this self explanatory when using it - WDYT?

lforst:
> Looking good. I wonder if we should include a list of the possible `SeverityLevel`s? Or is this self explanatory when using it - WDYT?

I was a bit on the fence about this one because other sdks don't seem to document the levels either but I think we should because otherwise it's documented nowhere except in typescript...

It was obvious with the enums but now it's not. Updated in: 2cdeaafe35c8d1b20769e2d15d9b08d73b6eaf69

Lms24 (approved):

Lms24:
> because otherwise it's documented nowhere except in typescript...

Agreed. For people using just JS that might be a problem.

lobsterkatie (approved):
LGTM!

PR #4946, opened 2022-04-19
Migration docs for Android and Java v6

marandaneto:
https://github.com/getsentry/sentry-java/releases

TODO:
* Remove notes about min. version Gson and deprecated features from docs once this is GA.

### DIFF
--- a/src/platforms/android/migration.mdx
+++ b/src/platforms/android/migration.mdx
@@ -29,6 +29,8 @@ description: "Migrating between versions of Sentry's SDK for Android."
   * `IHandler` is now `MainLooperHandler` only.
 * Spans now have higher precision using the `System#nanoTime` instead of milliseconds.
 
+There are more changes and refactors, but they are not user breaking changes.
+
 ## Migrating to `io.sentry:sentry-android-timber 5.6.2`
 
 Starting from version `5.6.2`, the `Timber.tag` usage is no longer supported by our SDK (the tag will not be captured and reflected on Sentry). Please, vote on this [GitHub issue](https://github.com/getsentry/sentry-java/issues/1900) to let us know if we should provide support for that.
--- a/src/platforms/android/migration.mdx
+++ b/src/platforms/android/migration.mdx
@@ -23,7 +23,7 @@ description: "Migrating between versions of Sentry's SDK for Android."
   * `SentryOptions#isEnableSessionTracking`
   * `SentryOptions#setEnableSessionTracking`
 * Hints changed its type from `Object` to `Map<String, Object>`
-* `enableScopeSync` is now enabled by default.
+* `SentryOptions#enableScopeSync` is now enabled by default.
 * Removed unnecessary abstractions
   * `IBuildInfoProvider` is now `BuildInfoProvider` only.
   * `IHandler` is now `MainLooperHandler` only.
--- a/src/platforms/android/migration.mdx
+++ b/src/platforms/android/migration.mdx
@@ -27,7 +27,7 @@ description: "Migrating between versions of Sentry's SDK for Android."
 * Removed unnecessary abstractions
   * `IBuildInfoProvider` is now `BuildInfoProvider` only.
   * `IHandler` is now `MainLooperHandler` only.
-* Spans now have higher precision using the `System#nanoTime` instead of milliseconds.
+* `ISpan` now has higher precision using the `System#nanoTime` instead of milliseconds.
 
 There are more changes and refactors, but they are not user breaking changes.
 
--- a/src/platforms/android/migration.mdx
+++ b/src/platforms/android/migration.mdx
@@ -22,12 +22,31 @@ description: "Migrating between versions of Sentry's SDK for Android."
   * `SentryOptions#setCacheDirSize`
   * `SentryOptions#isEnableSessionTracking`
   * `SentryOptions#setEnableSessionTracking`
-* Hints changed its type from `Object` to `Map<String, Object>`
-* `SentryOptions#enableScopeSync` is now enabled by default.
 * Removed unnecessary abstractions
   * `IBuildInfoProvider` is now `BuildInfoProvider` only.
   * `IHandler` is now `MainLooperHandler` only.
 * `ISpan` now has higher precision using the `System#nanoTime` instead of milliseconds.
+
+* Hints changed its type from `Object` to `Map<String, Object>`
+
+_Old_:
+
+```kotlin
+Sentry.captureException(RuntimeException("exception"), "myStringHint")
+```
+
+_New_:
+
+```kotlin
+val hints = mutableMapOf<String, Any>("myHint" to "myStringHint")
+Sentry.captureException(RuntimeException("exception"), hints)
+```
+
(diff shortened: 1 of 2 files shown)


src/platforms/java/migration.mdx
```diff
@@ -4,6 +4,28 @@ sidebar_order: 1000
 description: "Learn about migrating from io.sentry:sentry 1.x to io.sentry:sentry 3.x."
 ---
 
+## Migrating from `io.sentry:sentry` `5.x` to `io.sentry:sentry` `6.0.0`
```
marandaneto, line 7: @maciejwalkowiak would you be able to help us add the migration items for the Java backend integrations? iirc:

> `SentryOptions` cannot be created anymore using `PropertiesProvider` with `SentryOptions#from` method. Use `ExternalOptions#from` instead and merge created object with `SentryOptions#merge`
> `SentryOptions` can merge properties from `ExternalOptions` instead of another instance of `SentryOptions`
> Feat: Relax TransactionNameProvider
> Ref: Remove deprecated and scheduled fields

src/platforms/android/migration.mdx
```diff
@@ -24,7 +49,7 @@ The `io.sentry.android.gradle` >= `3.0.0` requires [Android Gradle Plugin >= 7.0
 
 The `autoUpload` flag has been deprecated in favor of two new options:
 
-* `includeProguardMapping` - Disables/enables Proguard mapping functionality. When enabled, generates a UUID and attempts to upload a Proguard mapping associated with that UUID to Sentry.  When `autoUploadProguardMapping` is disabled, the plugin will only generate a `sentry-debug-meta.properties` file containg UUID, which later can be used to manually upload the mapping, for example, using [sentry-cli](/product/cli/dif/#proguard-mapping-upload).  
+* `includeProguardMapping` - Disables/enables Proguard mapping functionality. When enabled, generates a UUID and attempts to upload a Proguard mapping associated with that UUID to Sentry.  When `autoUploadProguardMapping` is disabled, the plugin will only generate a `sentry-debug-meta.properties` file containg UUID, which later can be used to manually upload the mapping, for example, using [sentry-cli](/product/cli/dif/#proguard-mapping-upload).
```
imatwawana, line 52:
```suggestion
* `includeProguardMapping` - Disables/enables Proguard mapping functionality. When enabled, generates a UUID and attempts to upload a Proguard mapping associated with that UUID to Sentry.  When `autoUploadProguardMapping` is disabled, the plugin will only generate a `sentry-debug-meta.properties` file containg UUID, which can later be used to manually upload the mapping using, for example [sentry-cli](/product/cli/dif/#proguard-mapping-upload).
```

imatwawana (commented):
Just made a couple small wording edits

src/platforms/java/migration.mdx
maciejwalkowiak: Change in `SentryOptions` does not affect directly backend integrations.

As far as I can see, only these notes should be added:

### Spring Boot

* `sentry.enable-tracing` property has been removed. To enable tracing, set `sentry.traces-sample-rate` property
* `SentryTracingFilter` can take a custom implementation of `TransactionNameProvider` when used with other web frameworks than Spring MVC

marandaneto: Thanks, added those and a few more that I found out.

src/platforms/android/migration.mdx
```diff
@@ -6,6 +6,31 @@ redirect_from:
 description: "Migrating between versions of Sentry's SDK for Android."
 ---
 
+## Migrating from `io.sentry:sentry-android 5.x` to `io.sentry:sentry-android 6.0.0`
+
+* Kotlin plugin is upgraded to `1.5`.
+* Kotlin `languageVersion` is upgraded to `1.4`.
```
bruno-garcia, line 12: The wording is a bit confusing. Does the user need to take action? Are we telling users that we updated the minimum required version? If not, maybe: "the sdks kotlin version minimum requirements is now 1.4" or something

src/platforms/android/migration.mdx
```diff
@@ -6,6 +6,31 @@ redirect_from:
...
+* `Gson` is removed as a transitive dependency and vendored in the SDK.
+  * Protocol classes now implement the `JsonSerializable` and `JsonDeserializer` interfaces.
```
bruno-garcia, line 14: Because this is under migration, I feel the assumption is that users must follow steps. So for items like this it could be good to add a heading "Note on major changes".

If there are no steps the user must follow let's point that out:

"no major api change in this version but a few bevahior and other relevant changes worth mentioning:"

src/platforms/android/migration.mdx
```diff
@@ -6,6 +6,31 @@ redirect_from:
...
+* Removed `@Deprecated` and `@ApiStatus.ScheduledForRemoval` methods
+  * `ITransaction#setRequest`
+  * `ITransaction#getRequest`
+  * `ITransaction#getContexts`
+  * `SentryBaseEvent#getOriginThrowable`
+  * `SentryOptions#getCacheDirSize`
+  * `SentryOptions#setCacheDirSize`
+  * `SentryOptions#isEnableSessionTracking`
+  * `SentryOptions#setEnableSessionTracking`
+* Hints changed its type from `Object` to `Map<String, Object>`
```
bruno-garcia, line 25: Here we could give an example how users can change their code if things break? Before it was a type cast now add it to the map first and read it later when using it

bruno-garcia (approved):

src/platforms/android/migration.mdx
marandaneto: They don't need to do anything, but it's incompatible with Kotlin 1.3 now, 1.3 is deprecated anyway with Kotlin 1.5 which is upgraded in the item above.

src/platforms/android/migration.mdx
marandaneto: Users don't need to do anything, we don't have any other entry with this breakdown of "API changes" or "behavior changes", it's not a bad idea tho.

If you look down this page, you see that we have code snippets "Old way / New way" when they need to follow steps.

src/platforms/android/migration.mdx
marandaneto: That makes sense, will do.

marandaneto:
> Should we mention those as well?
> 
> * [getsentry/sentry-java#1995 (files)](https://github.com/getsentry/sentry-java/pull/1995/files)
> * [getsentry/sentry-java@`6.0.0-alpha.1` (release)](https://github.com/getsentry/sentry-java/releases/tag/6.0.0-alpha.1) - here `ExternalOptions` are new and also some options became non-nullable.

No for [getsentry/sentry-java@`6.0.0-alpha.1` (release)](https://github.com/getsentry/sentry-java/releases/tag/6.0.0-alpha.1), See comment https://github.com/getsentry/sentry-docs/pull/4946#discussion_r853171606
For [getsentry/sentry-java#1995 (files)](https://github.com/getsentry/sentry-java/pull/1995/files), No as well, since `getPrintUncaughtStackTrace` was only used by the SDK itself, `isPrintUncaughtStackTrace` is still there.

romtsn:
Ah, didn't notice the comment thread, sorry!

PR #4949, opened 2022-04-19
ref(page-filters): Update a number of issue / alert screenshots

evanpurkhiser:


davidenwang (approved):
🙏 
thanks for doing these!

imatwawana (approved):
In the images where the left side navigation is showing, you've got your name in there as the user (vs the ubiquitous Jane Schmidt). I think we should mostly stick with Jane, but I'm fine with that for this batch if you don't care. Otherwise, these look great to me!

evanpurkhiser:
I don't care, it's tricky to get everything perfectly right haha

PR #4960, opened 2022-04-20
feat(perf): Add span details histogram note for EA

udameli:
I've added a note for EAs describing changes to the span details page. I've mentioned a new histogram and new data filters (min and max self time) that will enable users to narrow down the population.

Fixes PERF-1419

### DIFF
--- a/src/docs/product/performance/transaction-summary.mdx
+++ b/src/docs/product/performance/transaction-summary.mdx
@@ -103,7 +103,13 @@ The "Spans" tab displays a list of suspect spans that correspond to where most o
 
 Choose from one of several [metrics](#metrics) to sort spans and identify different types of problems. Filter spans to only see the span operation type you're interested in.
 
-Clicking on a span will give you more details about the specific span [group](#grouping). You can see the performance of the span over time and see a list of transaction events that contain the span.
+Clicking on a span will take you to its summary page that should give you more details about this specific span [group](#grouping). On the page you can see the performance of the span over time and see a list of transaction events that contain the span.
+
+<Note>
+
+If you're in the Early Adopter program, you can also see the distribution of the span self time. The self time histogram can help you identify systemic performance issues accross the entire dataset by illustrating how self time for every span instance is distributed. You can also select a subset of the dataset by zooming into the distribution histogram which will effectively set min and max self time bounds. The stats and transaction events table will reflect that change and will only include spans with self time that are within the specified bounds. Adding self time filters allows you to narrow down your focus on the best/worse performing spans so you can troubleshoot performance issues easier.
+
+</Note>
 
 ### Self Time
 
--- a/src/docs/product/performance/transaction-summary.mdx
+++ b/src/docs/product/performance/transaction-summary.mdx
@@ -107,7 +107,7 @@ Clicking on a span will take you to its summary page that should give you more d
 
 <Note>
 
-If you're in the Early Adopter program, you can also see the distribution of the span self time. The self time histogram can help you identify systemic performance issues accross the entire dataset by illustrating how self time for every span instance is distributed. You can also select a subset of the dataset by zooming into the distribution histogram which will effectively set min and max self time bounds. The stats and transaction events table will reflect that change and will only include spans with self time that are within the specified bounds. Adding self time filters allows you to narrow down your focus on the best/worse performing spans so you can troubleshoot performance issues easier.
+If you're in the Early Adopter program, you can also see the distribution of the span self time. The histogram can help you identify systemic performance issues across the entire dataset by illustrating how self time for every span instance is distributed. You can also select a subset of the dataset by zooming into the distribution histogram which will effectively set min and max bounds. The stats and transaction events table will reflect that change and will only include spans with self time that are within the specified bounds. Adding filters allows you to narrow down your focus on the best/worse performing spans so you can troubleshoot performance issues easier.
 
 </Note>
 
--- a/src/docs/product/performance/transaction-summary.mdx
+++ b/src/docs/product/performance/transaction-summary.mdx
@@ -118,7 +118,6 @@ If you're in the Early Adopter program, you can also see the distribution of the
 You can also select a subset of the dataset by zooming into the distribution histogram. The stats and transaction events table will reflect that change and will only include spans with self time that are within the bounds of your subset. This allows you to focus on the best or worse performing spans so you can troubleshoot performance issues more easily.
 
 
-</Note>
 
 ### Self Time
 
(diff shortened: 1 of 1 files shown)


src/docs/product/performance/transaction-summary.mdx
```diff
@@ -103,7 +103,13 @@ The "Spans" tab displays a list of suspect spans that correspond to where most o
 
 Choose from one of several [metrics](#metrics) to sort spans and identify different types of problems. Filter spans to only see the span operation type you're interested in.
 
-Clicking on a span will give you more details about the specific span [group](#grouping). You can see the performance of the span over time and see a list of transaction events that contain the span.
+Clicking on a span will take you to its summary page that should give you more details about this specific span [group](#grouping). On the page you can see the performance of the span over time and see a list of transaction events that contain the span.
```
silent1mezzo, line 106:
```suggestion
Clicking on a span will take you to its summary page that will give you more details about this specific span [group](#grouping). On the page you can see the performance of the span over time and see a list of transaction events that contain the span.
```

I feel like it's better to assume it will give more details than leave the user wondering whether it will or won't.

src/docs/product/performance/transaction-summary.mdx
```diff
@@ -103,7 +103,13 @@ The "Spans" tab displays a list of suspect spans that correspond to where most o
...
+
+<Note>
+
+If you're in the Early Adopter program, you can also see the distribution of the span self time. The self time histogram can help you identify systemic performance issues accross the entire dataset by illustrating how self time for every span instance is distributed. You can also select a subset of the dataset by zooming into the distribution histogram which will effectively set min and max self time bounds. The stats and transaction events table will reflect that change and will only include spans with self time that are within the specified bounds. Adding self time filters allows you to narrow down your focus on the best/worse performing spans so you can troubleshoot performance issues easier.
```
silent1mezzo, line 110: I feel like self time needs to be hyphenated (so does Grammarly)

silent1mezzo, line 110:
```suggestion
If you're in the Early Adopter program, you can also see the distribution of the span self time. The histogram can help you identify systemic performance issues across the entire dataset by illustrating how self time for every span instance is distributed. You can also select a subset of the dataset by zooming into the distribution histogram which will effectively set min and max bounds. The stats and transaction events table will reflect that change and will only include spans with self time that are within the specified bounds. Adding filters allows you to narrow down your focus on the best/worse performing spans so you can troubleshoot performance issues easier.
```

I feel like we've said self time (self-time?) a lot here. I removed a couple of instances but may need to play around with where we include it.

silent1mezzo (changes requested):

src/docs/product/performance/transaction-summary.mdx
udameli: This reads better for sure. Yeah I've never written self time so much in a single paragraph 😅

udameli: It looks like in the product, we're referring to it as [`self time`](url). I wonder if we can make an exception since it's a technical term? I don't hold a strong opinion though, just flagging that we'll need to update all other self time mentions in the [docs](http://localhost:3000/product/performance/transaction-summary/#self-time) and in the product

silent1mezzo: Sounds good. I'm fine with that, just wanted to make sure.

imatwawana: Also prefer we're consistent with the product. If we change it, we change it everywhere, but generally, we should quote the UI.

imatwawana (commented):
Generally speaking, we don't need screenshots for EA docs, but this is pretty hard to visualize, especially when you get into how the table changes to reflect the zooming. While I think I understand why you've formatted this the way you have, I think we should find a way to fit this into the body text and just use a regular EA note. I'm not sure what this feature looks like so I'm not sure of the best placement for the content though. Ping me when you have some time to work on this.

dashed (approved):

src/docs/product/performance/transaction-summary.mdx
```diff
@@ -103,7 +103,13 @@ The "Spans" tab displays a list of suspect spans that correspond to where most o
...
 
-Clicking on a span will give you more details about the specific span [group](#grouping). You can see the performance of the span over time and see a list of transaction events that contain the span.
+Clicking on a span will take you to its summary page that will give you more details about this specific span [group](#grouping). On the page you can see the performance of the span over time and see a list of transaction events that contain the span.
+
+<Note>
+
+If you're in the Early Adopter program, you can also see the distribution of the span self time. This histogram can help you identify systemic performance issues across the entire dataset by showing how self time for every span instance is distributed. You can also select a subset of the dataset by zooming into the distribution histogram which will effectively set min and max bounds. The stats and transaction events table will reflect that change and will only include spans with self time that are within the set bounds. Adding filters allows you to focus on the best or worse performing spans so you can troubleshoot performance issues more easily.
+
```
imatwawana, line 111:
```suggestion

### Span Summary

Clicking on a span will take you to its summary page which will give you more details about this specific span [group](#grouping). On the page, you can see the performance of the span over time and see a list of transaction events that contain the span.

<Note>

This span distribution feature is available only if you're in the Early Adopter program. Features available to Early Adopters are still in-progress and may have bugs. We recognize the irony. If you’re interested in being an Early Adopter, you can turn your organization’s Early Adopter status on/off in General Settings. This will affect all users in your organization and can be turned back off just as easily.

</Note>

If you're in the Early Adopter program, you can also see the distribution of the span self time using the (added) "Display" dropdown. This histogram can help you identify systemic performance issues across the entire dataset by showing how self time for every span instance is distributed. 

You can also select a subset of the dataset by zooming into the distribution histogram. The stats and transaction events table will reflect that change and will only include spans with self time that are within the bounds of your subset. This allows you to focus on the best or worse performing spans so you can troubleshoot performance issues more easily.


```

src/docs/product/performance/transaction-summary.mdx
```diff
@@ -103,7 +103,13 @@ The "Spans" tab displays a list of suspect spans that correspond to where most o
...
+</Note>
```
imatwawana, line 112:
```suggestion
```

udameli:
> Generally speaking, we don't need screenshots for EA docs, but this is pretty hard to visualize, especially when you get into how the table changes to reflect the zooming. While I think I understand why you've formatted this the way you have, I think we should find a way to fit this into the body text and just use a regular EA note. I'm not sure what this feature looks like so I'm not sure of the best placement for the content though. Ping me when you have some time to work on this.

as @imatwawana and I discussed earlier, there's going to be more work on adding a new page for the span summary. We will write a more explicit and complete description of that page. as a short-term solution, @imatwawana has proposed adding a new section "span summary" that will encapsulate text related to the histogram.

PR #4963, opened 2022-04-21
[JS v7] ** do not merge ** feat(angular): Update Angular docs for v7 release

Lms24:
This PR updates our Angular SDK docs for the v7 release. It adds information about compatibility of our SDK with different Angular versions and updates information on `TraceDirective` compatibility.

**Important:** Do not merge this PR before v7 of the JS SDKs is released!

### DIFF
--- a/src/includes/getting-started-install/javascript.angular.mdx
+++ b/src/includes/getting-started-install/javascript.angular.mdx
@@ -1,7 +1,21 @@
+<Alert level="info" title="Angular Version Compatibility" >
+
+  Starting with v7 of the Angular SDK, only Angular versions 10 and newer are supported. For older Angular versions, please keep using v6 as shown below until you upgrade your Angular version.
+
+</Alert>
+
 ```bash {tabTitle:npm}
+# for Angular 10 and newer
 npm install --save @sentry/angular @sentry/tracing
+
+# for Angular 9 and older
+npm install --save @sentry/angular^6.x @sentry/tracing^6.x
 ```
 
 ```bash {tabTitle:Yarn}
+# for Angular 10 and newer
+yarn add @sentry/angular @sentry/tracing
+
+# for Angular 9 and older
 yarn add @sentry/angular @sentry/tracing
 ```
--- a/src/platforms/javascript/guides/angular/components/tracehelpers.mdx
+++ b/src/platforms/javascript/guides/angular/components/tracehelpers.mdx
@@ -4,80 +4,16 @@ title: Track Angular Components
 
 To track Angular components as part of your transactions, use any of these three options:
 
-1. `TraceClassDecorator` tracks a duration between `OnInit` and `AfterViewInit` lifecycle hooks in components:
+1. `TraceDirective` tracks a duration between `OnInit` and `AfterViewInit` lifecycle hooks in template:
 
-```javascript
-import { Component } from "@angular/core";
-import * as Sentry from "@sentry/angular";
+<Alert level="info" title="Compatibility">
 
-@Component({
-  selector: "layout-header",
-  templateUrl: "./header.component.html",
-})
-@Sentry.TraceClassDecorator()
-export class HeaderComponent {
-  // ...
-}
-```
-
-2. `TraceMethodDecorator` tracks a specific lifecycle hooks as point-in-time spans in components:
-
-```javascript
-import { Component, OnInit } from "@angular/core";
-import * as Sentry from "@sentry/angular";
-
-@Component({
-  selector: "app-footer",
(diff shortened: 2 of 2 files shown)


src/includes/getting-started-install/javascript.angular.mdx
```diff
@@ -1,3 +1,11 @@
+<Alert level="info" title="Angular Version Compatibility" >
+
+The latest version of the Sentry Angular SDK officially supports Angular 10 and newer.
+If you are using an older Angular version and you experience problems with the SDK, you can
+try using version 6 (at your own risk) until you upgrade your Angular version.
```
lforst, line 5: This isn't what we want to say right? If you are using an old angular version and you have problems, you **should** use v6 (which is not risky I believe). Or does v6 also not support angular version less than 10? If so we should clarify that.

src/platforms/javascript/guides/angular/components/tracehelpers.mdx
```diff
@@ -4,7 +4,42 @@ title: Track Angular Components
 
 To track Angular components as part of your transactions, use any of these three options:
 
-1. `TraceClassDecorator` tracks a duration between `OnInit` and `AfterViewInit` lifecycle hooks in components:
+1. `TraceDirective` tracks a duration between `OnInit` and `AfterViewInit` lifecycle hooks in template:
+
+<Alert level="info" title="Compatibility">
```
lforst, line 9: I wonder if this wouldn't fit better in the troubleshooting page. If it is really important that everybody sees this I would leave it here, if it's only relevant to a limited amount of people I'd move it to the troubleshooting page. I leave this up to you. :D

src/includes/getting-started-install/javascript.angular.mdx
Lms24: So here's the thing:

Like v7, v6 also has NG10-13 in its peer deps. So "officially" (but unfortunately nowhere specified, except for `package.json`) it also only supported these versions. If you wanted to use it with NG <10, you have to specify the `--force` flag when `npm install`ing. In that sense, it's "risky" but apart from that it seems to work for people (except for `TraceDirective`).

I want to put in writing that v7 officially supports 10-13. It _seems_ to support NG9 as well but I don't think we should officially add an older NG version to v7  when v6 didn't include it (also, NG9 is old).  Which is why I don't want to explicitly highlight the fact that that one can use v7 with NG9. At the same time I don't want to say "if you're using NG<10, use v6" because NG9 users could still upgrade... I know that's kinda a stupid situation. 

Therefore, I wanted say something like "if you're using v7 with older NG versions and experience problems, downgrade to v6". Does that kinda make sense? 
I can remove the "(at your own risk)". My intent here was to state that we're not gonna support v6 further. I can generally try to rewrite this to better reflect my intention.

src/platforms/javascript/guides/angular/components/tracehelpers.mdx
Lms24: That's a good point. I'll just move it there and in case people really complain (and don't find the info in Troubleshooting), we can put it back to this page again.

Lms24: Hmm it seems like I can't add anything to Troubleshooting without copying the entire document from `javascript/common` to `javscript/guides/angular`. Any ideas?
Tbh it doesn't seem worth the trouble to duplicate the entire document just for this paragraph. Debating whether to simply leave it in `tracehelpers.mdx`

src/includes/getting-started-install/javascript.angular.mdx
lforst: Hmm ok I get it now. I would change the wording of the first sentence to something along the lines of:

> The Sentry Angular SDK officially only supports Angular version 10 and newer.

Implying that even Sentry SDK v6 doesn't support Angular v9.

Then we could add additional info:

> If you need to use Angular version 9 and the latest version of the Sentry SDK gives you trouble, try downgrading the Sentry SDK to version 6. Version 6 of the Sentry SDK was compiled differently and might work with older versions of Angular. Please do this at your own risk, as this combination of packages is not being maintained and tested.

Wdyt? We can ofc still adapt the wording.

Lms24: Thanks for the suggestion. I added it (with some small adjustments) and repositioned the info box as well. Wdyt, does it sound better now?

lforst: Reads well now! Thanks for considering my input :D

lforst (approved):

PR #4970, opened 2022-04-25
chore(unity): Updated config screenshot

bitsandfoxes:


imatwawana:
Just curious if there's a specific reason you're using a jpg rather than a png file?

imatwawana:
> 

You're right! Ignore me! I read the diff incorrectly and thought you were removing the png file. All good!

marandaneto (approved):

PR #4974, opened 2022-04-25
feat(android): Troubleshooting page

### LINKED ISSUE #4397: Move troubleshooting section to proper troubleshooting page
### Core or SDK?

Platform/SDK

### Which part? Which one?

sentry-java

### Description

troubleshooting page is a section under some other pages https://docs.sentry.io/platforms/android/using-ndk/#troubleshooting

### Suggested Solution

having its own troubleshooting page like RN -> https://docs.sentry.io/platforms/react-native/troubleshooting/

romtsn:
I looked through the docs, but couldn't find anything else that could be added here, lmk if I missed something.

Closes #4397

### DIFF
--- a/src/platforms/android/using-ndk.mdx
+++ b/src/platforms/android/using-ndk.mdx
@@ -110,37 +110,3 @@ To support the devices on the API level lower than 16, update your `AndroidManif
 ```
 
 With these changes the SDK will be enabled on all devices with API level ≥ 14 and the SDK with NDK will be enabled on all devices with API level ≥ 16.
-
-## Troubleshooting
-
-If you are using `sentry-android` version earlier than 5.0, due to current limitations in how `sentry-native` works, it cannot identify native libraries loaded directly from `.apk` files. Instead, it needs to have access to the extracted libraries on the file system.
-This functionality is [disabled by default](https://developer.android.com/studio/releases/gradle-plugin#extractNativeLibs)
-when using an Android Gradle Plugin >= v3.6.0, and it needs to be enabled with the
-[`extractNativeLibs`](https://developer.android.com/guide/topics/manifest/application-element#extractNativeLibs) configuration option.
-
-If you're using [Android App Bundle](https://developer.android.com/guide/app-bundle), set `android.bundle.enableUncompressedNativeLibs=false` along with the `extractNativeLibs` configuration option.
-
-Apart from this, `sentry-native` also uses [tagged pointers](https://source.android.com/devices/tech/debug/tagged-pointers)
-internally. As a result, the Android SDK will not work out of the box on newer Android devices which have their own conflicting pointer tagging.
-
-<Note>
-
-If you are using `sentry-android` version earlier than 3.1, you must manually disable the [`allowNativeHeapPointerTagging`](https://developer.android.com/guide/topics/manifest/application-element#allowNativeHeapPointerTagging) configuration option because `sentry-native` uses [tagged pointers](https://source.android.com/devices/tech/debug/tagged-pointers)
-internally. As a result, this option will not work out of the box on newer Android devices that have their own conflicting pointer tagging. This is fixed in `sentry-android` versions 3.1 and above.
-
-</Note>
-
-**Example AndroidManifest.xml**:
-
-```xml {filename:AndroidManifest.xml}
-<application
-    android:allowNativeHeapPointerTagging="false"
-    android:extractNativeLibs="true">
-</application>
-```
-
-**Example gradle.properties**:
-
-```properties {filename:gradle.properties}
-android.bundle.enableUncompressedNativeLibs=false
-```


marandaneto (approved):

src/platforms/android/troubleshooting.mdx
```diff
@@ -0,0 +1,88 @@
...
+## NDK
+
+If you are using `sentry-android` version earlier than 5.0, due to current limitations in how `sentry-native` works, it cannot identify native libraries loaded directly from `.apk` files. Instead, it needs to have access to the extracted libraries on the file system.
+This functionality is [disabled by default](https://developer.android.com/studio/releases/gradle-plugin#extractNativeLibs)
+when using an Android Gradle Plugin >= v3.6.0, and it needs to be enabled with the
+[`extractNativeLibs`](https://developer.android.com/guide/topics/manifest/application-element#extractNativeLibs) configuration option.
+
+If you're using [Android App Bundle](https://developer.android.com/guide/app-bundle), set `android.bundle.enableUncompressedNativeLibs=false` along with the `extractNativeLibs` configuration option.
+
+Apart from this, `sentry-native` also uses [tagged pointers](https://source.android.com/devices/tech/debug/tagged-pointers)
```
imatwawana, line 16: Can we replace "apart from this" with "In addition"?
```suggestion
Apart from this, `sentry-native` also uses [tagged pointers](https://source.android.com/devices/tech/debug/tagged-pointers)
```

src/platforms/android/troubleshooting.mdx
```diff
@@ -0,0 +1,88 @@
...
+internally. As a result, the Android SDK will not work out of the box on newer Android devices which have their own conflicting pointer tagging.
```
imatwawana, line 17: Do we mean the SDK won't work on: 
1. specifically newer Android devices that have this conflicting pointer tagging
OR
2. all newer Android devices because they all have this conflicting pointer tagging

Let me know which, and I can add either add a comma or change a word to clarify this.

```suggestion
internally. As a result, the Android SDK will not work out of the box on newer Android devices which have their own conflicting pointer tagging.
```

src/platforms/android/troubleshooting.mdx
```diff
@@ -0,0 +1,88 @@
...
+</application>
+```
+
+**Example gradle.properties**:
+
+```properties {filename:gradle.properties}
+android.bundle.enableUncompressedNativeLibs=false
+```
+
+## Sentry Android Gradle Plugin build issues
```
imatwawana, line 41:
```suggestion
## Sentry Android Gradle Plugin Build Issues
```

src/platforms/android/troubleshooting.mdx
```diff
@@ -0,0 +1,88 @@
...
+
+The [Sentry Android Gradle plugin](/platforms/android/gradle/) uses bytecode manipulation to automatically measure performance of your application. This process requires looking at the dependencies of the application, which can potentially break the build process, if there are libraries which have been compiled/minified with a non-default java compiler, like R8/D8.
```
imatwawana, line 43:
```suggestion
The [Sentry Android Gradle plugin](/platforms/android/gradle/) uses bytecode manipulation to automatically measure the performance of your application. This process requires looking at the dependencies of the application, which can potentially break the build process if there are libraries which have been compiled/minified with a non-default java compiler, like R8/D8.
```

src/platforms/android/troubleshooting.mdx
marandaneto: Option 2, and it's not about newer Android devices, but rather Android devices with newer Android Operating system

imatwawana: Gotcha. How about this:
```suggestion
internally. As a result, the Android SDK will not work out of the box on devices with a newer Android OS because they have their own conflicting pointer tagging.
```

src/platforms/android/troubleshooting.mdx
```diff
@@ -0,0 +1,88 @@
...
+
+If you're using [Android App Bundle](https://developer.android.com/guide/app-bundle), set `android.bundle.enableUncompressedNativeLibs=false` along with the `extractNativeLibs` configuration option.
+
+In addition, `sentry-native` uses [tagged pointers](https://source.android.com/devices/tech/debug/tagged-pointers)
+internally. As a result, the Android SDK will not work out of the box on newer Android devices which have their own conflicting pointer tagging.
+
+<Note>
+
+If you're using a version of `sentry-android` before 3.1, you must manually disable the [`allowNativeHeapPointerTagging`](https://developer.android.com/guide/topics/manifest/application-element#allowNativeHeapPointerTagging) configuration option because `sentry-native` uses [tagged pointers](https://source.android.com/devices/tech/debug/tagged-pointers)
+internally. As a result, this option will not work out of the box on newer Android devices that have their own conflicting pointer tagging. This is fixed in `sentry-android` versions 3.1 and above.
```
imatwawana, line 22:
```suggestion
internally. As a result, this option won't work out of the box on devices with a newer Android OS because they have their own conflicting pointer tagging. This is fixed in `sentry-android` versions 3.1 and above.
```

PR #4988, opened 2022-04-28
Update manage-event-stream-guide.mdx

fimmicon:
----

By submitting this pull request, I confirm that Sentry can use, modify, copy, and redistribute this contribution, under Sentry's choice of terms.


imatwawana:
Thank you for the catch, @fimmicon!

PR #5005, opened 2022-05-04
feat(javascript): Link to Github `MIGRATION.md` file from docs

Lms24:
This PR adds two links to the JS SDK repo's `MIGRATION.md` file which contains information on breaking changes and instructions how to update to the latest version of Sentry. 
Currently, the URLs link to the `7.x` version of the migration guide to include v7 migration instructions for early adopters of v7 beta versions. This means we'll have to adjust the URLs as soon as v7 is released.

Currently, there are two links, one in the "Getting Started" (which arguably many people won't look up when updating but IMO it doesn't hurt to have it there either) and one in the "Troubleshooting" section.

ref: https://getsentry.atlassian.net/browse/WEB-807

### DIFF
--- a/src/includes/getting-started-install/javascript.mdx
+++ b/src/includes/getting-started-install/javascript.mdx
@@ -11,8 +11,5 @@ We also support alternate [installation methods](/platforms/javascript/install/)
 
 </Note>
 
-<Note>
+ If you're updating your Sentry SDK to the latest version, check out our [migration guide](https://github.com/getsentry/sentry-javascript/blob/7.x/MIGRATION.md) to learn more about breaking changes.
 
-Are you updating your Sentry SDK to the latest version? Check out our [migration guide](https://github.com/getsentry/sentry-javascript/blob/7.x/MIGRATION.md) to learn more about breaking changes!
-
-</Note>
--- a/src/includes/getting-started-install/javascript.mdx
+++ b/src/includes/getting-started-install/javascript.mdx
@@ -10,3 +10,9 @@ npm install --save @sentry/browser @sentry/tracing
 We also support alternate [installation methods](/platforms/javascript/install/).
 
 </Note>
+
+<Note>
+
+Are you updating your Sentry SDK to the latest version? Check out our [migration guide](https://github.com/getsentry/sentry-javascript/blob/7.x/MIGRATION.md) to learn more about breaking changes!
+
+</Note>
--- a/src/platforms/javascript/common/troubleshooting/index.mdx
+++ b/src/platforms/javascript/common/troubleshooting/index.mdx
@@ -9,6 +9,12 @@ sidebar_order: 1000
 
 If you need help solving issues with your Sentry JavaScript SDK integration, you can read the edge cases documented here. If you need additional help, you can view our forums, and customers on a paid plan may also contact support.
 
+## Updating to a new Sentry SDK Version
+
+If you update your Sentry SDK to a new major version, you might encounter breaking changes that need some adaption on your end.
+Check out our [migration guide](https://github.com/getsentry/sentry-javascript/blob/7.x/MIGRATION.md) to learn everything you need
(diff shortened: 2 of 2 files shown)


src/includes/getting-started-install/javascript.mdx
```diff
@@ -10,3 +10,9 @@ npm install --save @sentry/browser @sentry/tracing
 We also support alternate [installation methods](/platforms/javascript/install/).
 
 </Note>
+
+<Note>
+
+Are you updating your Sentry SDK to the latest version? Check out our [migration guide](https://github.com/getsentry/sentry-javascript/blob/7.x/MIGRATION.md) to learn more about breaking changes!
```
lforst, line 16: I like this!

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -9,6 +9,12 @@ sidebar_order: 1000
 
 If you need help solving issues with your Sentry JavaScript SDK integration, you can read the edge cases documented here. If you need additional help, you can view our forums, and customers on a paid plan may also contact support.
 
+## Updating to a new Sentry SDK Version
+
+If you update your Sentry SDK to a new major version, you might encounter breaking changes that need some adaption on your end.
+Check out our [migration guide](https://github.com/getsentry/sentry-javascript/blob/7.x/MIGRATION.md) to learn everything you need
```
lforst, line 15: What do you think about pointing that link to the current file (ie the one that's on master)? I think this section won't really matter until we merged v7 anyways. And as far as the beta goes I think it's enough if we put a link in the corresponding github release.

Lms24: I think people might check out troubleshooting if they get stuck with the breaking changes. I think that either we should link in both places to the `master` or to the `7.x` version. I think we can use for now the `7.x` version and then adjust the links once we have a stable v7 release.  If we use the 7.x version once, we have to update docs anyway. But I have no strong feelings about it, so I'm happy to change it either way. 

As you said, having the 7.x link in the GH changelog (which I think we do?) is probably most important for now.

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -9,6 +9,12 @@ sidebar_order: 1000
 
 If you need help solving issues with your Sentry JavaScript SDK integration, you can read the edge cases documented here. If you need additional help, you can view our forums, and customers on a paid plan may also contact support.
 
+## Updating to a new Sentry SDK Version
```
imatwawana, line 12: AP style for titles/headings

```suggestion
## Updating to a New Sentry SDK Version
```

src/includes/getting-started-install/javascript.mdx
```diff
@@ -10,3 +10,9 @@ npm install --save @sentry/browser @sentry/tracing
...
+
+</Note>
```
imatwawana, line 18: Suggest changing the language here so we can take this out of a note and make it body text. Right now, it comes right after another note that's already on the page, and having two notes in a row with no body text between them is something we should avoid. Also, we don't use exclamation points in docs.

```suggestion
 If you're updating your Sentry SDK to the latest version, check out our [migration guide](https://github.com/getsentry/sentry-javascript/blob/7.x/MIGRATION.md) to learn more about breaking changes.

```

src/platforms/javascript/common/troubleshooting/index.mdx
lforst: I'll let you make the call :)

lforst (approved):

src/platforms/javascript/common/troubleshooting/index.mdx
Lms24: Gonna leave it as is for now and I'll add a remark in our v7 docs release subtask to change the URL

PR #5008, opened 2022-05-04
ref(integrations): Remove indirect reference to plugin behavior

ceorourke:
Remove the sentence "This integration needs to set up only once per organization, then it is available for _all_ projects." from our first party integration docs. This is an artifact of differentiating between plugins and first party integrations, but since first party integrations (and partner integrations) are the majority and the path forward, this sentence doesn't need to be here.


imatwawana (approved):
Thanks for cleaning this up!

leeandher (approved):
What a great catch 🚢 🚀

PR #5010, opened 2022-05-05
Apple: Remove note on ViewController name matching

philipphofmann:
Only merge after releasing https://github.com/getsentry/sentry-cocoa/pull/1802

### DIFF
--- a/src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
+++ b/src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
@@ -14,7 +14,7 @@ Capturing transactions requires that you first <PlatformLink to="/performance/">
 
 This feature is available for iOS, tvOS, and Mac Catalyst, and only works for UIViewControllers.
 
-The UIViewController Instrumentation, once enabled, captures transactions when your app loads an [UIViewController][UIViewController], which is in-app and a subclass of UIViewController`. Nevertheless, the SDK doesn't capture transactions for UIViewControllers of third-party libraries or SwiftUI.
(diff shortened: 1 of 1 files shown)


src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -14,7 +14,7 @@ Capturing transactions requires that you first <PlatformLink to="/performance/">
 
 This feature is available for iOS, tvOS, and Mac Catalyst, and only works for UIViewControllers.
 
-The UIViewController Instrumentation, once enabled, captures transactions when your app loads an [UIViewController][UIViewController], which is in-app, a subclass of UIViewController, and the class name contains `ViewController`. Nevertheless, the SDK doesn't capture transactions for UIViewControllers of third-party libraries or SwiftUI.
+The UIViewController Instrumentation, once enabled, captures transactions when your app loads an [UIViewController][UIViewController], which is in-app and a subclass of UIViewController`. Nevertheless, the SDK doesn't capture transactions for UIViewControllers of third-party libraries or SwiftUI.
```
imatwawana, line 17: A back tick got added that wasn't closed so it changed all the plain text in the rest of the paragraph to code font. 

Also though, I'm not really sure the second part of that first sentence makes sense. ...when your app loads a UIViewController, which is in-app and a subclass of UIViewController." If you can explain what we mean to say here, I can suggest a fix. 

```suggestion
The UIViewController Instrumentation, once enabled, captures transactions when your app loads a [UIViewController][UIViewController], which is in-app and a subclass of UIViewController. Nevertheless, the SDK doesn't capture transactions for UIViewControllers of third-party libraries or SwiftUI.
```

brustolin (approved):

PR #5012, opened 2022-05-05
fix(search): Mention wildcards are not allowed with multi value search

Zylphrex:
The docs currently do not mention that wildcards are not allowed when you use
the multi value search. As is, it implies that wildcards do work, which is not
the case.


wmak (approved):
ty Tony

imatwawana (approved):

PR #5023, opened 2022-05-09
build: Use Sentry JS SDK version 7 beta

lforst:
The JavaScript SDK team intends to release major version 7 of their SDK soon. We see the docs as a great opportunity to test the new version.

This PR bumps the SDK version to use the beta. I've looked through the app and there shouldn't be any necessary adjustments.

If you're reviewing this PR, please take a look at the [migration docs](https://github.com/getsentry/sentry-javascript/blob/7.x/MIGRATION.md#upgrading-from-6x-to-7x) and let me know if you think there are any further changes that need to be made.

Ref: https://getsentry.atlassian.net/browse/WEB-907


AbhiPrasad (approved):
good upgrade process

let's see if this affects performance stats for pageloads!

imatwawana (approved):

PR #5033, opened 2022-05-12
clarify team plan behaviour for app-store-connect

souredoutlook:
- Change Alert level from Note to info to make the content more apparent. Customers may miss this content when being redirected from the app like so:

example of app store connect up-sell modal

- Clarify App Store Connect behaviour for Team subscription plans

### DIFF
--- a/src/includes/debug-symbols-apple/_default.mdx
+++ b/src/includes/debug-symbols-apple/_default.mdx
@@ -34,7 +34,7 @@ Once the integration is set up it will ensure that new builds are detected and f
 
 ![A fully set-up App Store Connect Integration](asc-repository.png)
 
-<Alert level="info">
+<Alert level="Note">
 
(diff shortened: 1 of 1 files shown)


src/includes/debug-symbols-apple/_default.mdx
```diff
@@ -34,9 +34,9 @@ Once the integration is set up it will ensure that new builds are detected and f
 
 ![A fully set-up App Store Connect Integration](asc-repository.png)
 
-<Alert level="Note">
+<Alert level="info">
```
imatwawana, line 37: We try to reserve notes at this level for information that's critical to know; things at the same level of versions that introduce breaking changes. And we try to use them infrequently. If you think this meets that standard, then leave as is. If not, then go with level "Note".
```suggestion
<Alert level="Note">
```

imatwawana (approved):
Left a comment about the note level. Otherwise LGTM!

PR #5051, opened 2022-05-19
fix source maps references and time range change example

imatwawana:
Fixes references to source maps to remove JavaScript-specific text and links
Fixes time range change example to improve clarity


scttcper (approved):
not just javascript 👍

PR #5052, opened 2022-05-19
add more user & session release health details

imatwawana:
Adds definitions for a user 
Adds example for crash-free user
Adds more information about sessions and session status

### DIFF
--- a/src/docs/product/releases/health/index.mdx
+++ b/src/docs/product/releases/health/index.mdx
@@ -150,4 +150,4 @@ The application shut down normally, but the session experienced handled errors. 
 
 The session ends normally and no errors occurred during its lifetime.
 
-Session status can be defined slightly differently by platform, and not all platforms use all of the statuses above. Check out the docs for your [specific platform](/platform-redirect/?next=/configuration/releases/#sessions) for more information.
+Session status can be defined slightly differently by platform, and not all platforms use all of the statuses above. Check out the docs for your [specific platform](/platform-redirect/?next=/configuration/releases/%23sessions) for more information.
--- a/src/docs/product/releases/health/index.mdx
+++ b/src/docs/product/releases/health/index.mdx
@@ -25,7 +25,7 @@ Sentry distinguishes between two kinds of sessions:
 
 <ul>
 
-<li style="list-style-type:none;">A user- or application-mode session begins with the start of the application. Or, it begins with bringing the already started application back from background to the foreground. In this context, a user is defined by one or more of the following: a unique ID or UUID, username, email, or IP address.</li>
+<li style="list-style-type:none;">A user- or application-mode session begins with the start of the application. Or, it begins with bringing the already started application back from background to the foreground. In this context, a user is defined by one or more of the following: a unique ID or UUID, username, or email.</li>
 
 <br></br>
 
--- a/src/docs/product/releases/health/index.mdx
+++ b/src/docs/product/releases/health/index.mdx
@@ -150,4 +150,4 @@ The application shut down normally, but the session experienced handled errors. 
 
 The session ends normally and no errors occurred during its lifetime.
 
-Session status can be defined slightly differently by platform, and not all platforms use all of the statuses above. Check out the docs for your [specific platform](/platform-redirect/?next=/configuration/releases/) for more information.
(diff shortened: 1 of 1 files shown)


src/docs/product/releases/health/index.mdx
```diff
@@ -147,3 +149,5 @@ The application shut down normally, but the session experienced handled errors.
 ### Healthy
 
 The session ends normally and no errors occurred during its lifetime.
+
+Session status can be defined slightly differently by platform, and not all platforms use all of the statuses above. Check out the docs for your [specific platform](/platform-redirect/?next=/configuration/releases/#sessions) for more information.
```
imatwawana, line 153:
```suggestion
Session status can be defined slightly differently by platform, and not all platforms use all of the statuses above. Check out the docs for your [specific platform](/platform-redirect/?next=/configuration/releases/%23sessions) for more information.
```

matejminar (approved):
Looks good to me.

src/docs/product/releases/health/index.mdx
```diff
@@ -25,7 +25,7 @@ Sentry distinguishes between two kinds of sessions:
 
 <ul>
 
-<li style="list-style-type:none;">A user- or application-mode session begins with the start of the application. Or, it begins with bringing the already started application back from background to the foreground.</li>
+<li style="list-style-type:none;">A user- or application-mode session begins with the start of the application. Or, it begins with bringing the already started application back from background to the foreground. In this context, a user is defined by one or more of the following: a unique ID or UUID, username, email, or IP address.</li>
```
sl0thentr0py, line 28: just as a sanity check, I'm not sure if any of the SDKs set the ip address here, because the ip address is a separate field in the session. And I'm not sure either how the UI/product side consolidates `did` and `ip_address` while counting sessions.

Here's what javascript sets as `did` (which is the unique session user identifier).
https://github.com/getsentry/sentry-javascript/blob/c518955b774e5bbc36e7f1a22410703469911a2a/packages/hub/src/session.ts#L40-L42

sl0thentr0py (approved):
just a short comment, but lgtm!

src/docs/product/releases/health/index.mdx
imatwawana: Do you think I should just take out IP address here? I've looked at the Python SDK and it seems like it has a similar treatment:
https://github.com/getsentry/sentry-python/blob/23e02952e41321e19303fcc0b1c29b4ed15a9ecc/sentry_sdk/session.py#L103-L104

sl0thentr0py: yea let's remove ip address

imatwawana, line 28:
```suggestion
<li style="list-style-type:none;">A user- or application-mode session begins with the start of the application. Or, it begins with bringing the already started application back from background to the foreground. In this context, a user is defined by one or more of the following: a unique ID or UUID, username, or email.</li>
```

src/docs/product/releases/health/index.mdx
```diff
@@ -47,7 +47,9 @@ Sentry distinguishes between two kinds of sessions:
...
 
-In terms of which mode a project runs in, Sentry is able to automatically detect if it is in a server setting and so should run in server-mode/request-mode or if it is in an application setting and so should run in user-mode/application-mode
+In terms of which mode a project runs in, Sentry is able to automatically detect if it is in a server setting and so should run in server-mode/request-mode or if it is in an application setting and so should run in user-mode/application-mode.
+
+While the descriptions of a session above cover many cases, the definition of a session can vary by platform. For example, with JavaScript applications, we create a session for every page load and on every navigation change for single-page applications. By contrast, backend applications don't have a well-defined user entity like web and mobile apps, so for those each new API request represents a new session. Check out the docs for your [specific platform](/platform-redirect/?next=/configuration/releases/) for more information.
```
imatwawana, line 52:
```suggestion
While the descriptions of a session above cover many cases, the definition of a session can vary by platform. For example, with JavaScript applications, we create a session for every page load and on every navigation change for single-page applications. By contrast, backend applications don't have a well-defined user entity like web and mobile apps, so for those, each new API request represents a new session. Check out the docs for your [specific platform](/platform-redirect/?next=/configuration/releases/) for more information.
```

PR #5074, opened 2022-05-24
feat(discover): extend list of stacking functions

wmak:
- This list is a bit outdated, and there are many more functions that
  discover supports. Updating this to include all of them, but longterm
  probably worth revisiting this section of docs
- This is so the API docs can link here


k-fish (approved):
Is there descs of these anywhere? People might not be sure what `epm` or `eps` stand for, likewise count_miserable etc. It might be nice to have a 1-liner explaining the function if we don't elsewhere?

imatwawana (approved):

priscilawebdev (approved):

wmak:
@k-fish agreed, will be doing that in a follow up. Just extending this list for now so that we can link to it from API docs

PR #5085, opened 2022-05-29
ref(native): capture-error using high-level API.

supervacuus:
@Swatinem and @jan-auer: this is my suggested change to the documentation after our conversation on Tuesday. It replaces the low-level calls with all currently available high-level APIs. It also shows how to add a stack trace to the exception instead of only referring to it.

Instead of having two, we can provide only one example with the stack trace (and comment on its optionality).

### DIFF
--- a/src/includes/capture-error/native.mdx
+++ b/src/includes/capture-error/native.mdx
@@ -1,23 +1,30 @@
 To capture an error or exception condition, create events containing an
-exception object. It needs to contain at least a value and type:
+exception object:
 
 ```cpp
 #include <sentry.h>
 
-sentry_value_t exc = sentry_value_new_object();
-sentry_value_set_by_key(exc, "type", sentry_value_new_string("Exception"));
-sentry_value_set_by_key(exc, "value", sentry_value_new_string("Error message."));
-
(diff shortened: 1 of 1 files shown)


src/includes/capture-error/native.mdx
```diff
@@ -1,23 +1,30 @@
...
+```cpp
+#include <sentry.h>
 
 sentry_value_t event = sentry_value_new_event();
-sentry_value_set_by_key(event, "exception", exceptions);
+
+sentry_value_t exc = sentry_value_new_exception("Exception", "Error message.");
+
+sentry_value_t stacktrace = sentry_value_new_stacktrace(NULL, 0);
+sentry_value_set_by_key(exc, "stacktrace", stacktrace);
```
supervacuus, line 25: I think it was these two which we could replace with another function:

```void sentry_value_add_stacktrace(sentry_value_t, void**, size_t)```

to get rid of key/entry management on the client-side, correct?

Swatinem (approved):
I think this is spot on!

Swatinem:
I will merge this as is now, we can then add the API and update the docs again once it is in a published release.

PR #5091, opened 2022-05-30
Added traces_sample_rate parameter to all code snippets

antonpirker:
Added the parameter `traces_sample_rate` to all code snippets in the Python documentation. 

With this everyone copy-pasting the code snippets will have performance enabled by default which should help with increasing the adoption.

### DIFF
--- a/src/includes/configuration/before-breadcrumb-hint/python.mdx
+++ b/src/includes/configuration/before-breadcrumb-hint/python.mdx
@@ -7,12 +7,7 @@ def before_breadcrumb(crumb, hint):
     return crumb
 
 sentry_sdk.init(
-    dsn="___PUBLIC_DSN___",
-
-    # Set traces_sample_rate to 1.0 to capture 100%
-    # of transactions for performance monitoring.
-    # We recommend adjusting this value in production,
-    traces_sample_rate=1.0,
+    # ...
 
     before_breadcrumb=before_breadcrumb,
 )
--- a/src/includes/configuration/before-breadcrumb-hint/python.mdx
+++ b/src/includes/configuration/before-breadcrumb-hint/python.mdx
@@ -6,7 +6,16 @@ def before_breadcrumb(crumb, hint):
         crumb['data']['thread'] = hint['log_record'].threadName
     return crumb
 
-sentry_sdk.init(before_breadcrumb=before_breadcrumb)
+sentry_sdk.init(
+    dsn="___PUBLIC_DSN___",
+
+    # Set traces_sample_rate to 1.0 to capture 100%
+    # of transactions for performance monitoring.
+    # We recommend adjusting this value in production,
+    traces_sample_rate=1.0,
+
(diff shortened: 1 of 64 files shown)


src/includes/configuration/before-breadcrumb-hint/python.mdx
```diff
@@ -6,7 +6,16 @@ def before_breadcrumb(crumb, hint):
...
 
-sentry_sdk.init(before_breadcrumb=before_breadcrumb)
+sentry_sdk.init(
+    dsn="___PUBLIC_DSN___",
+
+    # Set traces_sample_rate to 1.0 to capture 100%
+    # of transactions for performance monitoring.
+    # We recommend adjusting this value in production,
+    traces_sample_rate=1.0,
+
```
untitaker, line 16: i personally would avoid changing codesamples which don't intend to showcase the setup of sentry. this codesample shows off the `before_breadcrumb` option, and nothing else

imatwawana (changes requested):
Seconding what @untitaker has said, this is a docs-ecosystem level change you're making here. When we're telling a user what to copy/paste to set the environment or whatever else, we shouldn't just drop other code into the sample. And definitely not on one platform alone. Also, in every case, we'd need to update the prose to say "this sets x...and enables performance." Maybe some of these code snippets might make sense to make this kind of change, but I'll have to go through this more carefully to see where that might make sense. I'll circle back.

src/includes/configuration/before-breadcrumb-hint/python.mdx
antonpirker: I thought also about that. And from my own experience when I did those kind of thing (I mean going into a documentation or googling, to find a code snipped that I need from time to time, but always forget) 

I always remember the way how I get to the code sample that I need. For example, I know: I go to docs.sentry.io, go to "configuration", "filtering" and then copy the snippet there, because it has everything I need. (because in this example I use before_send in all my projects)

So that was the rational behind changing it everywhere. But maybe that is overkill...

antonpirker:
And a side note for @imatwawana : I will do this later also for the Ruby, Go, and Serverless documentation.

imatwawana:
> And a side note for @imatwawana : I will do this later also for the Ruby, Go, and Serverless documentation.

@untitaker - if we're doing this for these four platforms, should we not be doing this for all of them?

untitaker:
in principle yes but I presume those are most important to have good docs right now

antonpirker:
I have now kept the `traces_sample_rate` parameter only on setup snippets. Please re-review @untitaker and/or @imatwawana .

(One thing: the paths containing `/wizard/` I could not find in the documentation. But they look like guides for different frameworks, so I guess they are setup snippets and should include the parameter.

untitaker (approved):

antonpirker:
Ok, removed the params from integration configuration, migration guide and the one place you pointed out @imatwawana

PR #5097, opened 2022-05-31
ref(android): Simplify UI transaction code sample

philipphofmann:
Reduce the complexity of the code sample for manually adding spans
to user interaction transactions.


romtsn:
I'm curious, why do you think it's better to simplify? The previous example showed how to link a span with an exception, now we kinda lost this info

philipphofmann:
> I'm curious, why do you think it's better to simplify? The previous example showed how to link a span with an exception, now we kinda lost this info

I think there should be a different place in the docs to highlight this feature. Here we should only point out how manually add spans IMO.

mattjohnsonpint (approved):

PR #5098, opened 2022-05-31
feat(apple): User interaction instrumentation

philipphofmann:
Add docs similar to user interaction instrumentation on Android.
As the implementations don't match exactly we can't create reusable docs.

### DIFF
--- a/src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
+++ b/src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
@@ -199,7 +199,7 @@ If the UI transaction has idled, but didn't have any child spans added, it will 
 
 </Note>
 
-The SDK binds user interaction transactions bound to the `Scope` automatically if there's no other transaction set. Because of that, you can create spans using manual instrumentation, and those spans will be automatically associated with the running UI transaction.
+The SDK binds user interaction transactions to the `Scope` automatically if there's no other transaction set. Because of that, you can create spans using manual instrumentation, and those spans will be automatically associated with the running UI transaction.
 
 ```Swift
 import Sentry
--- a/src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
+++ b/src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
@@ -223,7 +223,7 @@ func loadUserDataOnClick() {
 }
 ```
 
-When the user interaction transaction is not finished yet, but the user makes a new interaction, the SDK automatically finishes the previous user interaction transaction. This is because only one transaction can be bound to the scope at a time. However, if the same view has been interacted with (for example, a `UIButton` was clicked again within the `idleTimeout` window), the idle timer will be reset and the transaction duration will be extended with the `idleTimeout` value.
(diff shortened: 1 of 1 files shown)


src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -164,6 +164,67 @@ SentrySDK.start { options in
...
+
+The SDK composes the transaction name out of the host `UIViewController` and the method that the `UIView` is calling; for example, `YourApp_LoginUIViewController.loginButton`. The transaction operation is set to `ui.action` plus the interaction type `click`. The transaction finishes automatically after it reaches the specified [idleTimeout](/platforms/android/configuration/options/#idle-timeout) and all of its child spans are finished. The `idleTimeoout` defaults to `3000` milliseconds (three seconds).
+
+<Note>
+
+If the UI transaction has idled, but didn't have any child spans added, it will be dropped.
+
+</Note>
+
+The SDK binds user interaction transactions bound to the `Scope` automatically if there's no other transaction set. Because of that, you can create spans using manual instrumentation, and those spans will be automatically associated with the running UI transaction.
```
romtsn, line 202:
```suggestion
The SDK binds user interaction transactions to the `Scope` automatically if there's no other transaction set. Because of that, you can create spans using manual instrumentation, and those spans will be automatically associated with the running UI transaction.
```

src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -164,6 +164,67 @@ SentrySDK.start { options in
...
+```objc {tabTitle:Objective-C}
+@import Sentry;
+
+[SentrySDK startWithConfigureOptions:^(SentryOptions *options) {
+    options.dsn = @"___PUBLIC_DSN___";
+    options.enableUserInteractionTracing = YES;
+}];
+```
+
+The SDK composes the transaction name out of the host `UIViewController` and the method that the `UIView` is calling; for example, `YourApp_LoginUIViewController.loginButton`. The transaction operation is set to `ui.action` plus the interaction type `click`. The transaction finishes automatically after it reaches the specified [idleTimeout](/platforms/android/configuration/options/#idle-timeout) and all of its child spans are finished. The `idleTimeoout` defaults to `3000` milliseconds (three seconds).
```
romtsn, line 194: This means the UI transaction will always be captured right? So there's no situation similar to Android, where we can have no `view.id` assigned?

romtsn (approved):

src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
philipphofmann: Yes, exactly. We know which method the click is calling on the linked UIViewController.

src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -164,6 +164,67 @@ SentrySDK.start { options in
...
+
+<Note>
+
+If the UI transaction has idled, but didn't have any child spans added, it will be dropped.
+
+</Note>
+
```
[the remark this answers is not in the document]
philipphofmann: Yep, we should mention that.

PR #5102, opened 2022-05-31
fix(quotas): Clarify filters application scope

iker-barriocanal:
Inbound data filters are applied to all events, including transactions.
See [envelope
filtering](https://github.com/getsentry/relay/blob/7c6b308bbf04e63e8e0b456489da1ddc320e2e6a/relay-server/src/actors/envelopes.rs#L1603)
and the [should_filter
check](https://github.com/getsentry/relay/blob/7c6b308bbf04e63e8e0b456489da1ddc320e2e6a/relay-filter/src/lib.rs#L38-L60) in relay.

### DIFF
--- a/src/docs/product/accounts/quotas/index.mdx
+++ b/src/docs/product/accounts/quotas/index.mdx
@@ -208,7 +208,7 @@ If you have enabled the storage of crash reports, you may set limits for the max
 
 ## Inbound Filters {#inbound-data-filters}
 
-In some cases, the data you’re receiving in Sentry is hard to filter, or you don’t have the ability to update the SDK’s configuration to apply the filters. Sentry provides several methods to filter error events server-side, which apply to all events before checking for potential rate limits.
+In some cases, the data you’re receiving in Sentry is hard to filter, or you don’t have the ability to update the SDK’s configuration to apply the filters. Sentry provides several methods to filter all events and attachments server-side, which are applied before checking for potential rate limits.
 
 Inbound filters include:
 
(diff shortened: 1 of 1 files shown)


untitaker (approved):

src/docs/product/accounts/quotas/index.mdx
```diff
@@ -208,7 +208,7 @@ If you have enabled the storage of crash reports, you may set limits for the max
 
 ## Inbound Filters {#inbound-data-filters}
 
-In some cases, the data you’re receiving in Sentry is hard to filter, or you don’t have the ability to update the SDK’s configuration to apply the filters. Sentry provides several methods to filter error events server-side, which apply before checking for potential rate limits. This does not apply to transaction events or attachments.
+In some cases, the data you’re receiving in Sentry is hard to filter, or you don’t have the ability to update the SDK’s configuration to apply the filters. Sentry provides several methods to filter error events server-side, which apply to all events before checking for potential rate limits.
```
imatwawana, line 211: The updated text still specifically mentions error events, so I've made an update to fix that.

Also, based on what you brought up in the docs channel, thinking we should clarify that attachments aren't events wherever it makes sense.

```suggestion
In some cases, the data you’re receiving in Sentry is hard to filter, or you don’t have the ability to update the SDK’s configuration to apply the filters. Sentry provides several methods to filter all events and attachments server-side, which are applied before checking for potential rate limits.
```

imatwawana (commented):
Hopefully my fix works

PR #5112, opened 2022-06-02
adding new styles and making it somewhat more accessible

stevenplewis:
### Legal Boilerplate

### DIFF
--- a/src/css/_includes/terms.scss
+++ b/src/css/_includes/terms.scss
@@ -5,8 +5,12 @@
 }
 
 span.term {
-  text-decoration: underline dotted $darkPink 2px;
+  text-decoration: underline;
+  text-decoration-line: underline;
+  text-decoration-color: $brandDecoration;
+  text-decoration-style: dotted;
   text-decoration-skip-ink: auto;
+  text-decoration-thickness: 2px;
   cursor: help;
 }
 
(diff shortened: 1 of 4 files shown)


imatwawana (commented):
I haven't tested the screen reading yet, but a few things:

1. The styling is overriding a change we made before where if a word had a hyperlink in it in already, then we didn't want to the word to have the glossary underline or the tooltip:


2. When the tooltip is displayed, it covers up regular text, but not other terms that have tooltips:


3. The sidebar causes the tooltip not to display:


4. If the tooltip bumps into a note, the tooltip disappears:


Though this doesn't seem to be consistent between browsers and (on Firefox the note disappears in one case, but a weird note that's not rendering properly on Safari and Firefox causes the tooltip to disappear).

5. On Safari, the underlines don't show up.

All of these examples are on this page: https://sentry-docs-git-docs-term-tooltip-design-update.sentry.dev/product/accounts/quotas/

6. The pink is intense. Can we make it either less red or less bright without losing too much contrast?

imatwawana:
I'm only testing VoiceOver on Safari since that's the only browser it's really meant to work well with. It's pretty good, but it doesn't tell you there's a definition or tooltip happening so it just sort of launches into it. 

Btw I haven't tested on IE. I can circle back with that tomorrow.

stevenplewis:
@imatwawana All the things in your first comment should be addressed now, and I added an aria-label to the tooltip so it will announce itself more clearly to address your second comment. I think adding the label requires more button presses for the reader to advance to the tooltip, then the content of the tooltip, then out of the tooltip, then back to the rest of the text, but it's probably clearer this way. Good call.

imatwawana (approved):
All the visual stuff looks good now and I like the new underline and tooltip colour a lot more. Also, the screen reading sounds good too. Looks great!

untitaker (changes requested):
nice!

untitaker (approved):

imatwawana:
Because this change adds in screen reading of tooltips, let's not merge this until we've reduced the instances of a glossary pop-up to once per page. cc @untitaker

PR #5115, opened 2022-06-03
feat(api): Update the bump script since CI doesn't run it

wmak:
- this is so the script can be run on MacOS where the date function is
  slightly different


AniketDas-Tekky (approved):

JoshFerge:
CI still runs, this, only on releases though. we still probably want that in there.

wmak:
Changed it so that locally we'll do 
`./scripts/bump-verson.sh local` which will skip the two sed commands

wmak:
also my bad should've waited for this to merge before updating develop: https://github.com/getsentry/develop/pull/597

JoshFerge (approved):

PR #5126, opened 2022-06-08
adding version info

imatwawana:
Removed this version information from the Releases Set Up page in product and moved it here since this is when you need to know about version naming behavior (during the actual SDK configuration)

### DIFF
--- a/src/platforms/common/configuration/releases.mdx
+++ b/src/platforms/common/configuration/releases.mdx
@@ -53,6 +53,17 @@ Releases are global per organization; prefix them with something project-specifi
 
 </Note>
 
+The behavior of a few features depends on whether a project is using semantic or time-based versioning.
+
+- Regression detection
+- `release:latest`
+
+We automatically detect whether a project is using semantic or time-based versioning based on:
+
+- If ≤ 2 releases total: we look at most recent release.
+- If 3-10 releases total: if any of the most recent 3 releases is semver, project is semver.
+- If 10+ releases: if any of the most recent 3 releases is semver, and 3 out of the most recent 10 releases is semver, then the project is semver.
+
 ## Setting a Release
 
 <PlatformContent includePath="set-release" notateUnsupported />
--- a/src/platforms/common/configuration/releases.mdx
+++ b/src/platforms/common/configuration/releases.mdx
@@ -62,7 +62,7 @@ We automatically detect whether a project is using semantic or time-based versio
 
 - If ≤ 2 releases total: we look at most recent release.
 - If 3-9 releases (inclusive): if any of the most recent 3 releases is semver, project is semver.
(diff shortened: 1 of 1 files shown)


imatwawana:
@matejminar & @sl0thentr0py - review from either one of you is fine!

src/platforms/common/configuration/releases.mdx
```diff
@@ -53,6 +53,17 @@ Releases are global per organization; prefix them with something project-specifi
...
 
+The behavior of a few features depends on whether a project is using semantic or time-based versioning.
+
+- Regression detection
+- `release:latest`
+
+We automatically detect whether a project is using semantic or time-based versioning based on:
+
+- If ≤ 2 releases total: we look at most recent release.
+- If 3-10 releases total: if any of the most recent 3 releases is semver, project is semver.
```
ahmedetefy, line 64: I feel like `3-10` is a bit confusing because it is not right inclusive as in with up to 9 releases this rule applies

ahmedetefy (changes requested):

src/platforms/common/configuration/releases.mdx
imatwawana: Does this work? If not, please feel free to suggest an edit.

```suggestion
- If 3-9 releases (inclusive): if any of the most recent 3 releases is semver, project is semver.
```

ahmedetefy: This makes sense to me .. looks good

src/platforms/common/configuration/releases.mdx
```diff
@@ -53,6 +53,17 @@ Releases are global per organization; prefix them with something project-specifi
...
+- If 10+ releases: if any of the most recent 3 releases is semver, and 3 out of the most recent 10 releases is semver, then the project is semver.
```
ahmedetefy, line 65: Does `10+` mean 10 and above ? or just above 10?

imatwawana: > Does `10+` mean 10 and above ? or just above 10?

I don't know the answer to that. This was just a copy/paste from another part of the docs. @wedamija, @sl0thentr0py - do either of you know what the answer to that?

wedamija: The rules are
 - <= 2
 - > 2 < 10
 - >= 10

Not sure what the best way to word it is

imatwawana: Okay so: 
```suggestion
- If 10 or more releases: if any of the most recent 3 releases is semver, and 3 out of the most recent 10 releases is semver, then the project is semver.
```

ahmedetefy (approved):

PR #5134, opened 2022-06-09
Unify Ruby code snippets.

antonpirker:
Unify the use of Ruby code snippets. Basically just adding the ellipsis to show that other parameters need to be added.


imatwawana (approved):
Looks good!

PR #5151, opened 2022-06-16
Move API SHA bump to its own script

chadwhitacre:
Overloading the `bump-version.sh` Craft [hook](https://github.com/getsentry/craft#pre-release-command) to also be used from localhost during [more-frequent API doc updates](https://develop.sentry.dev/api/public/#build-process) added some [inter-environment friction](https://github.com/getsentry/sentry-docs/pull/5115) that resulted in [a CalVer release failure](https://github.com/getsentry/self-hosted/issues/1515#issuecomment-1156787127).

Seems great to update the API docs more frequently than once a month, and that means it would be redundant to also do this during monthly CalVer release, so let's move that to a separate script.

### DIFF
--- a/scripts/bump-api-schema-sha.sh
+++ b/scripts/bump-api-schema-sha.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# This script is part of the API docs build process:
+#
+#  https://develop.sentry.dev/api/public/#build-process
+
+set -eu
+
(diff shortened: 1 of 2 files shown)


chadwhitacre:
Oh geez automerge in repo w/ no branch protection. 🤦‍♂️

chadwhitacre:
Ftr here's where [API SHA bump was added](https://github.com/getsentry/sentry-docs/pull/4085/files) to `bump-version.sh`.

PR #5153, opened 2022-06-16
Fix bug in bump workflow

chadwhitacre:
Follow-up to https://github.com/getsentry/sentry-docs/pull/5152

https://github.com/getsentry/sentry-docs/runs/6921967697?check_suite_focus=true#step:3:29


chadwhitacre:
Tested in https://github.com/getsentry/sentry-docs/runs/6922264046?check_suite_focus=true, commit shows up as expected in c9b994d806e3f02f854e824370aec275c04e42fc.

PR #5165, opened 2022-06-16
feat(sentry apps): Add page for Threads

ceorourke:
Add a docs page for the new partner integration from Threads


src/docs/product/integrations/notification-incidents/threads/index.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+### Alerts
+
+The Threads Sentry integration supports [issue alerts](/product/alerts/alert-types/#issue-alerts) and [metric alerts](/product/alerts/alert-types/#metric-alerts). When selecting the [action](/product/alerts/create-alerts/issue-alert-config/#then-conditions-actions), choose "Notify Integration" and then "Threads" for issue alerts, or select "Threads" from the action dropdown for metric alerts. Finally, choose the channel to be notified in:
+
+![](threads-rule.png)
+
+A Threads notification will be created automatically when the alert is triggered.
+
+
```
ceorourke, line 33: I _kinda_ feel like there should be a heading here but I'm not sure, and if there is, I'm unsure what to call it.

src/docs/product/integrations/notification-incidents/threads/index.mdx
```diff
@@ -0,0 +1,40 @@
...
+Threads **won't** work with self-hosted Sentry.
+
+</Note>
+
+To add the Threads integration, navigate to **Settings > Integrations > Threads** and click "Install".
+
+![](threads-install.png)
+
+### Alerts
```
imatwawana, line 25:
```suggestion
### Alert Rules
```

src/docs/product/integrations/notification-incidents/threads/index.mdx
imatwawana: This section seems to be talking about issues rather than issue alerts. But then you go back to talking about metric alerts. Maybe the metric alert content should come first as part of the Alert Rules section, and the part about issues should be under its own H2 heading of some sort. I'm thinking "Issue Statuses" or something.

src/docs/product/integrations/notification-incidents/threads/index.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+![](threads-rule.png)
+
+A Threads notification will be created automatically when the alert is triggered.
+
+Threads will notify you when a metric alert has been triggered, and again when it was resolved:
+
+![](threads-metric-alert.png)
+
+### Issue Statuses
```
imatwawana, line 37: This looks good. My only question is whether this should be an H3 or H2. Like Alert Rules makes sense to me under Install and Configure because you have to configure the alert rules. And then the outcome of that is in that second screenshot. But it seems like we don't really talk about any configuration for issues. If my question isn't making sense, please feel free to ping me.

imatwawana (commented):
There's one thing I'm still unsure about, but otherwise this looks good.

src/docs/product/integrations/notification-incidents/threads/index.mdx
ceorourke: Nah that makes sense, I updated it to an h2.

imatwawana (approved):

PR #5175, opened 2022-06-19
docs: clarifying Hermes artifacts

kbrandwijk:
This PR hopes to clarify the sentence 'You will upload the **original** bundle (output from Step 1) and the **composed** source map (output from Step 3)' from the docs. It took me a while to figure out what was meant here, because Step 1 has _two_ outputs, the bundle and the packager source map. 

So I clarified in step one that there's two outputs, and clarified which of those outputs need to be uploaded.

### Legal Boilerplate


marandaneto:
Thanks @kbrandwijk

marandaneto (approved):

PR #5179, opened 2022-06-20
Remove explicit rewrite flags from sentry cli RN

marandaneto:
### Legal Boilerplate

### DIFF
--- a/src/docs/clients/react-native/sourcemaps.mdx
+++ b/src/docs/clients/react-native/sourcemaps.mdx
@@ -15,7 +15,7 @@ This SDK has been superseded by the new React Native SDK. Sentry preserves this 
 
 Currently automatic source map handling is only implemented for iOS with Xcode and Android with gradle. If you manually invoke the [react-native packager](https://github.com/facebook/metro) you can however get source maps anyways by passing _–sourcemap-output_ to it.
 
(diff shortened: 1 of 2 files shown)


src/docs/clients/react-native/sourcemaps.mdx
```diff
@@ -40,6 +40,12 @@ node_modules/@sentry/cli/bin/sentry-cli releases \
     /path/to/your/files
 ```
 
+<Note>
+
+If you're using `sentry-cli` prior to version 1.59.0, pass `--rewrite` to the `upload-sourcemaps` command to fix up the source maps before the upload (inlines sources and so forth). Version 1.59.0 does this automatically.
```
bitsandfoxes, line 45:
```suggestion
If you're using `sentry-cli` prior to version 1.59.0, pass `--rewrite` to the `upload-sourcemaps` command to fix up the source maps before the upload (inlines sources and so forth). Version 1.59.0 and newer does this automatically.
```
Would that work to make it "future proof"?

bitsandfoxes (approved):

src/docs/clients/react-native/sourcemaps.mdx
marandaneto: The warning is a copy/paste from another docs page that has been reviewed before, but the change still makes sense, thanks.

PR #5182, opened 2022-06-20
feat(discover): Include count_web_vitals in query builder and query equations

0Calories:
Includes the `count_web_vitals` function within the Query Builder and Query Equations pages in the docs

image

image


imatwawana:
Wondering if this also needs to be added to the [Search properties table](https://docs.sentry.io/product/sentry-basics/search/searchable-properties/#properties-table).

0Calories:
Thanks @imatwawana, I'll add it to the table as well

wmak (commented):
lgtm

PR #5185, opened 2022-06-21
Prevent confusion between environment and tag. 

antonpirker:
This text makes users confuse environments with tags. In this context it is the environment parameter that is given to the init() function.


bruno-garcia (approved):
Makes sense. Thanks Anton

PR #5219, opened 2022-06-27
chore(android): Migration to 6.2.0

romtsn:


src/platforms/android/migration.mdx
```diff
@@ -6,6 +6,10 @@ redirect_from:
 description: "Migrating between versions of Sentry's SDK for Android."
 ---
 
+## Migrating from `io.sentry:sentry-android 6.x` to `io.sentry:sentry-android 6.2.0`
+
+* Kotlin plugin is upgraded to `1.6.10`.
```
philipphofmann, line 11: I always thought you don't need Kotlin for using `sentry-android`. So why do we point it out here?

marandaneto: I guess this is more for Timber, OkHttp, Fragment.
We should mention that we are still compatible with `1.4` compatibility.

philipphofmann (approved):
LGTM 🚀

PR #5194, opened 2022-06-22
Remove references to forum.sentry.io

chadwhitacre:
Part of https://github.com/getsentry/team-ospo/issues/27.

PR #5223, opened 2022-06-27
Standardize links to GH Issues

chadwhitacre:
Addresses @drguthals comments at https://github.com/getsentry/sentry-docs/pull/5194#pullrequestreview-1018073588 (which I missed when responding to review over there) and https://github.com/getsentry/sentry-docs/pull/5194#discussion_r905784314.


chadwhitacre:
I don't think we have such a problem with duplicate issues that we should add friction, let's send people to the new issue form.

PR #5233, opened 2022-06-29
fix(sentry apps): Fix url -> uri typo

ceorourke:
Fix the example where it says "url" instead of "uri". Double checked this on the [source of truth](https://github.com/getsentry/integration-platform-example/blob/main/integration-schema.json#L79-L82) 😉.


leeandher (approved):
😍  i feel seen

PR #5249, opened 2022-07-05
update Strapi link

imatwawana:
Update dead link to new Strapi plugin for v4+


imatwawana:
@remidej - Does this update look good to you?

imatwawana:
@remidej - nudge

remidej (commented):
Yes thanks it's a good fix for this broken link. I just think that we can remove the v4+ text but as you want

imatwawana:
> Yes thanks it's a good fix for this broken link. I just think that we can remove the v4+ text but as you want

Thank you!!

PR #5271, opened 2022-07-08
Add MAUI Wizard page

mattjohnsonpint:
Adding MAUI to the Wizard

### DIFF
--- a/src/wizard/dotnet/maui.md
+++ b/src/wizard/dotnet/maui.md
@@ -40,6 +40,10 @@ public static MauiApp CreateMauiApp()
             // This option is not recommended when deploying your application.
             options.Debug = true;
 
+            // Set TracesSampleRate to 1.0 to capture 100% of transactions for performance monitoring.
+            // We recommend adjusting this value in production.
+            options.TracesSampleRate = 1.0;
+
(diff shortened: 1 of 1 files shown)


src/wizard/dotnet/maui.md
```diff
@@ -0,0 +1,104 @@
...
+        {
+            // The DSN is the only required setting.
+            options.Dsn = "___PUBLIC_DSN___";
+
+            // Use debug mode if you want to see what the SDK is doing.
+            // Debug messages are written to stdout with Console.Writeline,
+            // and are viewable in your IDE's debug console or with 'adb logcat', etc.
+            // This option is not recommended when deploying your application.
+            options.Debug = true;
+
```
bruno-garcia, line 42: Even though we currently don't send any transactions auto of the box, if users already add this, once we add auto transactions, it'll already work automatically. Then we can also avoid adding it below in the perf step:

```suggestion
            // Set TracesSampleRate to 1.0 to capture 100% of transactions for performance monitoring.
            // We recommend adjusting this value in production.
            options.TracesSampleRate = 1.0;
```

bruno-garcia (approved):

PR #5277, opened 2022-07-11
chore(integrations): Update Vercel integration docs + screenshots

leeandher:
This PR will update the docs for installing Vercel on Sentry SaaS. There was some outdated information and screenshots. To be safe, I left in the troubleshooting section with another older looking screenshot, but I never encountered that Issue with an ad blocker, so it might be safe to remove.


src/docs/product/integrations/deployment/vercel/index.mdx
```diff
@@ -63,13 +63,9 @@ Use Vercel to [link projects](#project-linking) for uploading source maps and no
 
 ## Uninstallation
 
-1. The integration must be uninstalled from Vercel. Visit the Vercel integration configuration page (**Organization Settings** > **Integrations** > **Vercel** > **Configurations**) and click "Uninstall" then "Visit Vercel" in the modal that pops up.
+1. You can uninstall the integration from Vercel or Sentry. To do so in Sentry, navigate to **Organization Settings** > **Integrations** > **Vercel** > **Configurations**, click "Uninstall" and confirm.
```
ceorourke, line 66: Why say it can be done from Sentry or Vercel but remove the Vercel instructions? Is it preferred to uninstall from Sentry?

ceorourke (approved):
Looks good other than my one tiny comment which is non blocking.

src/docs/product/integrations/deployment/vercel/index.mdx
leeandher: I guess I just mentioned that because the docs before were explicit in that you needed to uninstall from Vercel. That isn't the case any more but you'll still have to remove the internal integration afterward

Issue #5274, opened 2022-07-11
Invalid cmake command given


connorjclark:
### Core or SDK?

Platform/SDK

### Which part? Which one?

https://docs.sentry.io/platforms/native/

### Description

https://docs.sentry.io/platforms/native/

Sorry, I couldn't figure out where this page exists in the docs source. The "Edit this page" link at the bottom takes you to the wrong file.

`cmake -B build -D SENTRY_BACKEND=crashpad --config RelWithDebInfo -S .` errors in cmake 3.20+

### Suggested Solution

should drop --config parameter

it was never an option, and starting in cmake 3.20 invalid params are errors instead of being ignore.

getsentry-release:
Routing to @getsentry/team-mobile for [triage](https://develop.sentry.dev/processing-tickets/#3-triage). ⏲️

getsentry-release:
Routing to @getsentry/owners-native for [triage](https://develop.sentry.dev/processing-tickets/#3-triage). ⏲️

connorjclark:
To be clear the rest of the commands are fine, it's just this one that doesn't need the config option.

PR #5283, opened 2022-07-12
fix(native): clean up install/build instructions

supervacuus:
This fixes https://github.com/getsentry/sentry-docs/issues/5274. It seems the docs got a bit out of sync with the repository documentation. I took the liberty of cleaning up the required build steps as follows:

We can remove the CMAKE_BUILD_TYPE entirely for the purpose of this documentation because it uses our default build type. In addition to that, it would have to be passed as `-DCMAKE_BUILD_TYPE=RelWithDebInfo` and not via `--config` at the build-generator stage. This leads to an error in CMake [starting with version 3.20](https://cmake.org/cmake/help/latest/release/3.20.html#other-changes).

In the actual compile/build stage it could be passed via `--config` for build-generators that provide all build-types under a single hierarchy (like Visual Studio solutions). This is also not applicable for a
macOS example, but was silently ignored. I removed it too.

Last but not least, I got rid of the explicit `-S .` (=source) argument, because the [default value when not specifying the source directory is also the cwd](https://cmake.org/cmake/help/latest/manual/cmake.1.html#generate-a-project-buildsystem).

@Swatinem as always, welcome your input.

### DIFF
--- a/src/includes/getting-started-install/native.mdx
+++ b/src/includes/getting-started-install/native.mdx
@@ -6,9 +6,9 @@ For example, `CMake` can be used like this (on macOS):
 
 ```shell
 # configure the cmake build into the `build` directory, with crashpad (on macOS)
-cmake -B build -D SENTRY_BACKEND=crashpad --config RelWithDebInfo -S .
+cmake -B build -D SENTRY_BACKEND=crashpad
 # build the project
-cmake --build build --config RelWithDebInfo --parallel
+cmake --build build --parallel
 # install the resulting artifacts into a specific prefix
 cmake --install build --prefix install
 # which will result in the following (on macOS):
--- a/src/includes/getting-started-install/native.mdx
+++ b/src/includes/getting-started-install/native.mdx
@@ -5,8 +5,11 @@ To build the SDK, download the latest sources from the [Releases page](https://g
 For example, `CMake` can be used like this (on macOS):
 
 ```shell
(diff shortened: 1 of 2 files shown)


Swatinem (approved):
I think these updates are fair to do.
If I remember correctly, you do need an explicit `--config` on Windows which generates all the configurations at once. I remember @jan-auer had strong opinions about guiding users to build optimized code when they are lazy and just wanna do copy-paste.
Also certain ancient cmake versions actually required the explicit `-S` option, but we might as well update the examples to be based on more up-to-date versions.

supervacuus:
I agree that the docs can be used to help people take the easy path and teach them something at the same time. I can also understand if sentry prefers explicit parameters (to raise the awareness for `RelWithDebInfo` for instance) vs. implicit defaults (even though we rely on many defaults already).

On the other hand, I think it is a good idea to minimize the noise in the first example they are confronted with, especially since the CMake configuration already went out of its way to use sensible defaults. I am certainly willing to add additional examples that provide better defaults for our main target OSes (windows being a good counter-example).

In any case, just explaining my thoughts, I have no problem with making everything explicit again as long as it doesn't fail.

Swatinem:
That does sound like a good idea. Anything that helps people along so they won’t open support issues is a good idea ;-)

Swatinem (approved):

connorjclark:
Thanks for addressing the issue I opened!

Issue #5124, opened 2022-06-07
Aspnetcore Setting a Sampling Function docs incorret key value


Rassilion:
### Core or SDK?

Platform/SDK

### Which part? Which one?

Aspnetcore

### Description

In the Aspnetcore "Setting a Sampling Function" docs example script is not working, looks like key value "url" changed to "__HttpPath"

document page: https://docs.sentry.io/platforms/dotnet/guides/aspnetcore/configuration/sampling/#setting-a-sampling-function

I'm trying send a pull request but I could not found that page in this repo.

### Suggested Solution

Change 
```
context.CustomSamplingContext.GetValueOrDefault("url")
```

to 

```
context.CustomSamplingContext.GetValueOrDefault("__HttpPath")
```

bruno-garcia:
Thanks for raising. The snippets are here: https://github.com/getsentry/sentry-docs/tree/master/src/includes
We'd appreciate a PR

mattjohnsonpint:
@lucas-zimerman - want to improve this doc if you have some time?

lucas-zimerman:
Will do :)

mattjohnsonpint:
That's not quite the best way. There's an extension method that should be used instead.

lucas-zimerman:
> Actually, it looks like the example doc referenced is showing _custom_ context information. The build in one is available with just `context.TryGetHttpPath()`.
> 
> The docs could still be a little clearer about this.

Thing is, TryGetHttpPath is only available on Asp.NET where the mentioned doc is shared across all .NET targets
Maybe mention `TryGetHttpPath` if using with ASP.NET Core, otherwise the longer method.

mattjohnsonpint:
I see.  Yeah - no need to split the docs up.  Just put them both in the same example with a comment such as `// assuming ASP.NET Core` on one and `// or from your own custom data` on the other.

lucas-zimerman:
@bruno-garcia  could you close this Issue?

PR #5286, opened 2022-07-13
(.NET): Update TracerSampling Key reference

lucas-zimerman:
Aims to fix the following issue: https://github.com/getsentry/sentry-docs/issues/5124

Sampling context was retrieving the wrong key on the Documentation.

Also, only ASP.NET Core has `TryGetHttpPath` so for other platforms I only fixed the correct key where in ASP.NET Core I used the appropriate extension.

### DIFF
--- a/src/includes/performance/traces-sampler-as-sampler/dotnet.aspnetcore.mdx
+++ b/src/includes/performance/traces-sampler-as-sampler/dotnet.aspnetcore.mdx
@@ -17,7 +17,7 @@ options.TracesSampler = context =>
     }
 
     // Otherwise, sample based on URL (exposed through custom sampling context)
-    return context.CustomSamplingContext.GetValueOrDefault("__HttpPath") switch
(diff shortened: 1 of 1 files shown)


src/includes/performance/traces-sampler-as-sampler/dotnet.aspnetcore.mdx
```diff
@@ -0,0 +1,38 @@
+```csharp
+using Sentry;
```
bruno-garcia, line 2: We need the using to access some extension method? Otherwise there's no need here since we're just doing "options."

bruno-garcia (approved):

src/includes/performance/traces-sampler-as-sampler/dotnet.aspnetcore.mdx
bruno-garcia: Merging for now as this fixes a big issue. Please remove the using if needed in a follow up pr

PR #5306, opened 2022-07-19
(.NET) Remove redundant using

lucas-zimerman:
The using on the ASP.NET Core doc is not used so I just removed it.


imatwawana:
@lucas-zimerman is is ready to merge?

PR #5312, opened 2022-07-20
Add sentry wizard setup instructions to sentry docs

Neo-Zhixing:
This adds a new frontmatter field, `wizard_setup`, that allows the author to specify simplified setup instructions. This will be used on the onboarding view to give simplified setup instructions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,9 @@
     "react-instantsearch-dom": "^6.4.0",
     "react-popper": "^2.2.4",
     "react-select": "^3.1.0",
+    "remark": "^10.0.1",
     "remark-deflist": "^0.2.1",
+    "remark-html": "^8.0.0",
     "ts-node": "^9.0.0",
     "typescript": "^3.9.7"
   },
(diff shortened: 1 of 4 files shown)


imatwawana:
Please update this page in the [contributor guide](https://docs.sentry.io/contributing/pages/frontmatter/) as well with an explanation of how to use this new frontmatter field.

Neo-Zhixing:
@imatwawana This is not a complete feature just yet, we'll have to do an A/B test to decide if we want to keep this. And even if this passes, it's specific to onboarding wizard pages so it's not applicable to all pages.

package.json
```diff
@@ -73,7 +73,9 @@
     "react-instantsearch-dom": "^6.4.0",
     "react-popper": "^2.2.4",
     "react-select": "^3.1.0",
+    "remark": "^10.0.1",
```
Neo-Zhixing, line 76: We're not exactly adding new dependencies here - these two packages are already being used by `gatsby-transformer-remark`

scefali (approved):

imatwawana:
> @imatwawana This is not a complete feature just yet, we'll have to do an A/B test to decide if we want to keep this. And even if this passes, it's specific to onboarding wizard pages so it's not applicable to all pages.

Understood!

PR #5318, opened 2022-07-21
Add extra snippets for transaction names 

lucas-zimerman:
Adds the code snippet for each platform added (Except Apple because it seems like you cannot alter it).


src/includes/enriching-events/set-transaction-name/android.mdx
```diff
@@ -0,0 +1,3 @@
+```java
```
bruno-garcia, line 1: We're trying to go Kotlin first for Android. Can we have the tabbed option with Kotlin, first?

src/includes/enriching-events/set-transaction-name/java.mdx
```diff
@@ -0,0 +1,3 @@
+```java
```
bruno-garcia, line 1: Here I believe we're using Kotlin second, need to confirm

bruno-garcia (approved):

bruno-garcia:
Thanks Lucas

PR #5328, opened 2022-07-23
Hide warning for Ruby

jordan-brough:
I might not be understanding this part of the docs, but I don't think this section applies to the way the Ruby library works?

(I also might not understand how this documentation format works. Can you nest a "PlatformSection" inside a "PlatformSection"?)

### Legal Boilerplate

### DIFF
--- a/src/platforms/common/enriching-events/scopes.mdx
+++ b/src/platforms/common/enriching-events/scopes.mdx
@@ -120,12 +120,14 @@ made will stay isolated within the `with-scope` callback function. This allows y
 more easily isolate pieces of context information to specific locations in your code or
 even call `clear` to briefly remove all context information.
 
+<PlatformSection notSupported={["ruby"]}>
 <Alert level="info" title="Important">
 
 Any exceptions that occur within the callback function for `with-scope` will not be
 caught, and all errors that occur will be silently ignored and **not** reported.
 
 </Alert>
+</PlatformSection>
 
 </PlatformSection>
 


imatwawana:
Looks like a platformsection within a platformsection works! I've tagged our Ruby dev for technical review on this.

sl0thentr0py:
So this warning can be removed for ruby but it also exists because of a caveat with the way `with_scope` works.

If you do the following, the exception will be captured but the tag won't be used (because that scope is thrown away in [an ensure](https://github.com/getsentry/sentry-ruby/blob/332315907185e6ffe05aa64dc751cdd8ab9f5448/sentry-ruby/lib/sentry/hub.rb#L60-L61)).
```ruby
    Sentry.with_scope do |scope|
      scope.set_tags(my_tag: "my value")
      1 / 0
    end
```

If you do the following, the exception will be captured _and_ the tag will be visible.
```ruby
    Sentry.with_scope do |scope|
      scope.set_tags(my_tag: "my value")

      begin
        1 / 0
      rescue => e
        Sentry.capture_exception(e)
      end
    end
```

sl0thentr0py (approved):

jordan-brough:
What @sl0thentr0py mentioned was a gotcha for me in Ruby usage.  It would be nice if there were a way for an exception thrown inside of `with_scope` to carry it's tags & other data up to wherever it gets caught & reported. Is that something that might be possible? Would it be worth opening an Issue about it in the ruby repo or something?

sl0thentr0py:
@jordan-brough done, let's continue this conversation there.

imatwawana:
> @jordan-brough done, let's continue this conversation there.

@sl0thentr0py - shall I go ahead and merge this?

sl0thentr0py:
yes!

PR #5330, opened 2022-07-25
Add experimental flags section under common options

marandaneto:
### Legal Boilerplate

### DIFF
--- a/src/platforms/common/configuration/options.mdx
+++ b/src/platforms/common/configuration/options.mdx
@@ -497,13 +497,13 @@ The callback typically gets a second argument (called a "hint") which contains t
 
 </ConfigKey>
 
-<PlatformSection supported={["javascript", "java", "python", "node", "php", "dotnet", "flutter"]}>
+<PlatformSection supported={["javascript", "java", "python", "node", "php", "dotnet", "dart"]}>
 
 ## Transport Options
 
 Transports are used to send events to Sentry. Transports can be customized to some degree to better support highly specific deployments.
 
-<ConfigKey name="transport" supported={["javascript", "php", "python", "java", "dotnet", "flutter", "rust"]}>
+<ConfigKey name="transport" supported={["javascript", "php", "python", "java", "dotnet", "dart", "rust"]}>
 
 Switches out the transport used to send events. How this works depends on the SDK. It can, for instance, be used to capture events for unit-testing or to send it through some more complex setup that requires proxy authentication.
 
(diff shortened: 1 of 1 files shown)


src/platforms/common/configuration/options.mdx
```diff
@@ -497,13 +497,13 @@ The callback typically gets a second argument (called a "hint") which contains t
 
 </ConfigKey>
 
-<PlatformSection supported={["javascript", "java", "python", "node", "php", "dotnet", "flutter"]}>
+<PlatformSection supported={["javascript", "java", "python", "node", "php", "dotnet", "dart"]}>
```
marandaneto, line 500: Ignore those changes, I just fixed a bug while checking out the preview.

src/platforms/common/configuration/options.mdx
```diff
@@ -591,6 +591,15 @@ An optional property that configures which downstream services receive the `sent
 
 <PlatformSection supported={["react-native", "javascript.cordova", "javascript.capacitor", "flutter"]}>
 
+## Experimental Features
```
marandaneto, line 594: This is the new/common docs for experimental features.
It's empty because no SDKs have it yet.

src/platforms/common/configuration/options.mdx
```diff
@@ -591,6 +591,15 @@ An optional property that configures which downstream services receive the `sent
...
+
+<ConfigKey name="experimental" supported={[]}>
+
+An optional property that configures which features are under experimental mode, this property is either an `Object Type` with properties or a key/value `TypedDict` depending the language. Experimental features are still in-progress and may have bugs. We recognize the irony.
```
imatwawana, line 598:
```suggestion
An optional property that configures which features are in experimental mode. This property is either an `Object Type` with properties or a key/value `TypedDict`, depending the language. Experimental features are still in-progress and may have bugs. We recognize the irony.
```

src/platforms/common/configuration/options.mdx
imatwawana: Do you mean ignore all the updates where you've updated the supported list for a platform section? Or just the one above?

marandaneto: All the updates, the only important one is https://github.com/getsentry/sentry-docs/pull/5330#discussion_r928810993 that you already reviewed, thanks @imatwawana

vladanpaunovic (approved):
LGTM @marandaneto!

It would be good if there could be some list of experimental features to point people to so they know what they can expect under this object.

marandaneto:
> LGTM @marandaneto!
> 
> It would be good if there could be some list of experimental features to point people to so they know what they can expect under this object.

Yep, when the first SDK adds a new experimental feature, that will be visible and not empty, so I'd rather just wait :)

PR #5331, opened 2022-07-25
Describe --version-name for iOS

kopax-polyconseil:
Solution given by @marandaneto https://github.com/getsentry/sentry-react-native/issues/2372#issuecomment-1194020493

### Legal Boilerplate

### DIFF
--- a/src/platforms/react-native/manual-setup/codepush.mdx
+++ b/src/platforms/react-native/manual-setup/codepush.mdx
@@ -72,7 +72,7 @@ sentry-cli react-native appcenter {APP} ios ./build/CodePush --deployment {DEPLO
 For non-macOS devices building for iOS, you will need to pass extra parameters that can't be computed automatically (`--bundle-id` and `--version-name`).
 Within your `Info.plist`, `--version-name` corresponds to `CFBundleShortVersionString` .
 
- on iOS.
+On iOS:
 
 ```bash
 sentry-cli react-native appcenter {APP} ios ./build/CodePush --deployment {DEPLOYMENT} --dist {DIST} --bundle-id {BUNDLE_ID} --version-name {VERSION_NAME}
--- a/src/platforms/react-native/manual-setup/codepush.mdx
+++ b/src/platforms/react-native/manual-setup/codepush.mdx
@@ -78,7 +78,7 @@ For non-macOS devices building for iOS, you will need to pass extra parameters t
 sentry-cli react-native appcenter {APP} ios ./build/CodePush --deployment {DEPLOYMENT} --dist {DIST} --bundle-id {BUNDLE_ID} --version-name {VERSION_NAME}
 ```
 
-If you don't want to use the automatically computed release format, instead of passing `--bundle-id` and `--version-name` you will need to pass the `--release-name` parameter.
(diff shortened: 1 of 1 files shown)


src/platforms/react-native/manual-setup/codepush.mdx
```diff
@@ -70,12 +70,13 @@ sentry-cli react-native appcenter {APP} ios ./build/CodePush --deployment {DEPLO
 ```
 
 For non-macOS devices building for iOS, you will need to pass extra parameters that can't be computed automatically (`--bundle-id` and `--version-name`).
+`--version-name` comes from appcenter and has the `{}@{}+codepush:{}` pattern (see [here](https://github.com/getsentry/sentry-cli/blob/e08e7c31d93a29797edf7d1e5ec42ee0453ecc49/src/utils/appcenter.rs#L191-L194)), it should be the value within the 2nd `{}`.
```
marandaneto, line 73: I'd rather prefer to mention that version name is the equivalent of `CFBundleShortVersionString` on iOS.

kopax-polyconseil: I'll update that

marandaneto (approved):

src/platforms/react-native/manual-setup/codepush.mdx
```diff
@@ -70,12 +70,15 @@ sentry-cli react-native appcenter {APP} ios ./build/CodePush --deployment {DEPLO
 ```
 
 For non-macOS devices building for iOS, you will need to pass extra parameters that can't be computed automatically (`--bundle-id` and `--version-name`).
+`--version-name` correspond to `CFBundleShortVersionString` within your `Info.plist`.
```
imatwawana, line 73: To avoid starting a sentence with text in code format, which is a bit hard to read, I've rearranged it:

```suggestion
Within your `Info.plist`, `--version-name` corresponds to `CFBundleShortVersionString` .
```

src/platforms/react-native/manual-setup/codepush.mdx
```diff
@@ -70,12 +70,15 @@ sentry-cli react-native appcenter {APP} ios ./build/CodePush --deployment {DEPLO
...
+
+ on iOS.
```
imatwawana, line 75:
```suggestion
On iOS:
```

src/platforms/react-native/manual-setup/codepush.mdx
```diff
@@ -70,12 +70,15 @@ sentry-cli react-native appcenter {APP} ios ./build/CodePush --deployment {DEPLO
...
 
 ```bash
 sentry-cli react-native appcenter {APP} ios ./build/CodePush --deployment {DEPLOYMENT} --dist {DIST} --bundle-id {BUNDLE_ID} --version-name {VERSION_NAME}
 ```
 
-If you don't want to use the automatically computed release format, you will need to pass the `--release-name` parameter.
+If you don't want to use the automatically computed release format, instead of passing `--bundle-id` and `--version-name` you will need to pass the `--release-name` parameter.
```
imatwawana, line 81:
```suggestion
If you don't want to use the automatically computed release format, instead of passing `--bundle-id` and `--version-name`, you'll need to pass the `--release-name` parameter.
```

imatwawana (approved):
Made some wording edits

PR #5332, opened 2022-07-25
CodePush: do not upload manually source maps

kopax-polyconseil:
Solution given by @marandaneto https://github.com/getsentry/sentry-react-native/issues/2372#issuecomment-1191103821

We tested it and it gives wrong sourcemaps for both iOS and Android, see https://github.com/getsentry/sentry-react-native/issues/2372 for more details

### Legal Boilerplate

### DIFF
--- a/src/platforms/react-native/manual-setup/codepush.mdx
+++ b/src/platforms/react-native/manual-setup/codepush.mdx
@@ -93,4 +93,3 @@ You can find more information about this in the [AppCenter CLI](https://docs.mic
 
 </Alert>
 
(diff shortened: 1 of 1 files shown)


marandaneto:
That should not be removed, there is still a lot of important info in that link.

kopax-polyconseil:
@marandaneto I don't get it. You said we should have not at first tried to use the manually generated sourcemaps. 

If it should not be removed, perhaps we should add something? Because this is the bloc that lead us in the wrong direction, and reason is simple: we did want to find a way that worked with hermes and this part was not explaining how to, so we though this bloc that I removed concerned us.

I suggest to improve this, by removing or by adding information so people don't do like we did.

PR #5336, opened 2022-07-26
doc: Add docs for negative inbound filters

untitaker:


src/docs/product/data-management-settings/filtering/index.mdx
```diff
@@ -73,3 +73,13 @@ Some symbols, such as the `*` character, receive special meaning as meta charact
 - `*` matches zero or more characters.
 - `\` escapes the next character making it literal. If it precedes a non-meta character, the backslash is ignored.
 - `[`, `]`, `{`, and `}` are reserved meta characters.
+- `!` works for negation, but only within brackets, such as in `[!1-4]` to match any character that is not `[1-4]`.
+
+Generally `bash`'s out-of-the-box behavior for globbing comes pretty close to what Sentry supports:
```
imatwawana, line 78: Suggest making it a bit less informal

```suggestion
Generally, `bash`'s out-of-the-box behavior for globbing is close to what Sentry supports:
```

PR #5346, opened 2022-07-27
fix: Only highlight first term in paragraph [INGEST-1262]

untitaker:


src/gatsby/plugins/gatsby-remark-terms/index.js
```diff
@@ -25,22 +25,28 @@ const CUSTOM_LINK_START = new RegExp("^<([a-zA-Z]+Link|a) ");
...
   // This currently doesn't handle nesting of e.g.
   // <a><strong><abbr>... but we don't have that in docs.
-  if (node.type == "link") return;
+  if (node.type == "link") return "skip";
 
   let insideCustomLink = false;
 
+  const alreadyExplained = {};
+
+  const newChildren = [];
```
untitaker, line 48: When I made the initial change, the build never terminated. I had to run Instruments.app against gatsby to figure that most of the time was spent in array splicing. Fixing this gross misuse of datastructures made the build terminate again.

It seems that the second-heaviest stacktrace is about regex replacements. Will create a followup PR.

untitaker: sike, the thing i wanted to do doesn't actually work. I think we would want to spend some time thinking about how to make gatsby-remark-variables faster, which I think is dominating our build times. cc @dcramer

imatwawana:
In one of our discussions, you suggested highlighting the first instance after each heading, and I thought that was a good idea, so that's what I thought we were doing. Is that a hard change to make at this point? 

Also, I didn't realize that doing this based on paragraphs would mean that any other type of text (e.g., text in lists or tables) wouldn't use the highlighting (see the word "quota"): 
https://sentry-docs-git-fix-terms-first-in-paragraph.sentry.dev/product/accounts/quotas/
Would this be different if we went with first instance after each heading?

untitaker:
>so that's what I thought we were doing. Is that a hard change to make at this point?

I completely forgot, I was messing around with multiple versions of this due to performance problems, but yeah that's what we agreed on. Fixed.

>Also, I didn't realize that doing this based on paragraphs would mean that any other type of text (e.g., text in lists or tables) wouldn't use the highlighting (see the word "quota"):

That's a bug I introduced, also fixed.

imatwawana (approved):
Looks good! Thanks for getting this into the quarter!

PR #5350, opened 2022-07-27
feat(nextjs): Add `transpileClientSDK` option and build config page

lobsterkatie:
In https://github.com/getsentry/sentry-javascript/pull/5472, a `transpileClientSDK` option was added to the nextjs SDK to force SDK code to be transpiled along with other app code, primarily for the use of folks wanting to use the SDK in older browsers without support for ES6 or ES6+ features like object spread. This documents that option by:
- Adding a section to the manual setup page (which is where other build-time options live)
- Adding a new 'Build Options' page redirecting people to the relevant section of of the manual setup page, for greater discoverability
- Adding a nextjs-specific include to a new legacy browser support section on the Troubleshooting page. (A generic include has also been added for the non-nextjs JS docs, detailing how to do manually what the added option causes the nextjs SDK to do automatically.)

### DIFF
--- a/src/platforms/javascript/common/troubleshooting/index.mdx
+++ b/src/platforms/javascript/common/troubleshooting/index.mdx
@@ -424,6 +424,6 @@ Remember to pass in `true` as the second parameter to `addEventListener()`. With
 
 ## Supporting Older Browsers
 
-Starting with version 7.0.0, the Sentry JavaScript SDK uses ES6 syntax, along with a few other ES6+ language features like object spread. If you are down-compiling your code in order to target older browsers that don't support such syntax, you'll need to include the Sentry SDK in that process.
+Starting with version 7.0.0, the Sentry JavaScript SDK uses ES6 syntax, along with a few other ES6+ language features, like object spread. If you are down-compiling your code in order to target older browsers that don't support such syntax, you'll need to include the Sentry SDK in that process.
 
 <PlatformContent includePath="troubleshooting/older-browser-support" />
--- a/src/platforms/javascript/common/troubleshooting/index.mdx
+++ b/src/platforms/javascript/common/troubleshooting/index.mdx
@@ -70,7 +70,7 @@ To fix this, change the `tracingOrigins` option during SDK initialization. For m
 
 ## `instrument.js` Line Numbers for Console Log statements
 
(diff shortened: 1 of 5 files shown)


src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -421,3 +421,9 @@ document.body.addEventListener(
 ```
 
 Remember to pass in `true` as the second parameter to `addEventListener()`. Without it, the event handler won't be called, since it's being added to the event target's ancestor rather than the event target itself, and unlike JavaScript runtime errors, `error` events resulting from load failures don't bubble, and therefore must be captured during the `capture` phase. For more information, see [the W3C spec](https://www.w3.org/TR/2003/NOTE-DOM-Level-3-Events-20031107/events.html#Events-phases).
+
+## Supporting Older Browsers
+
+Starting with version 7.0.0, the Sentry JavaScript SDK uses ES6 syntax, along with a few other ES6+ language features like object spread. If you are down-compiling your code in order to target older browsers that don't support such syntax, you'll need to include the Sentry SDK in that process.
```
AbhiPrasad, line 427: We can also leave a note here to use an ES5 CDN bundle, since our default is now ES6

AbhiPrasad (approved):
Technical stuff ✅

src/platforms/javascript/common/troubleshooting/index.mdx
lobsterkatie: Added a note.

imatwawana, line 427:
```suggestion
Starting with version 7.0.0, the Sentry JavaScript SDK uses ES6 syntax, along with a few other ES6+ language features, like object spread. If you are down-compiling your code in order to target older browsers that don't support such syntax, you'll need to include the Sentry SDK in that process.
```

imatwawana: Wondering if it would work to make this a platform section that's only supported for next.js and then you don't need the include that's for all the other JS frameworks (if the only reason it's been provided is so this page displays properly for the rest of JS). Not a hard requirement for this PR, but might be worth checking.

imatwawana (approved):
Added a comma and posed a structural question, but otherwise LGTM!

src/platforms/javascript/common/troubleshooting/index.mdx
lobsterkatie: Yeah, I wasn't super clear in my wording, sorry about that. Everyone but nextjs should have thing A (which actually _is_ meaningful), nextjs should have thing B. I'll fix the PR description.

PR #5378, opened 2022-08-03
feat(integration): Add Atlas to integration docs

leeandher:
This PR adds the [Atlas](https://www.getatlas.io/) integration with installation instructions. These will be linked to from the integration directory page in [getsentry/sentry](https://github.com/getsentry/sentry)


leeandher:
[New Page Preview](https://sentry-docs-git-leander-add-atlas-api-2644.sentry.dev/product/integrations/session-replay/atlas/) and [Edited Page Preview](https://sentry-docs-git-leander-add-atlas-api-2644.sentry.dev/product/integrations/). 

I noticed Asayer wasn't updated to OpenReplay, so I made that change as well

ceorourke (approved):
Approving but this almost feels like the second half should live on their own docs and we link to it, what do you think?

leeandher:
@ceorourke I think that makes sense though since this is open source, if they direct customers here they can also maintain these instructions. I think for now though, if they don't change much we can leave them as such for now.

PR #5381, opened 2022-08-04
Fix imported user on java samples

marandaneto:
`io.sentry.core.protocol` was a long time ago renamed to `io.sentry.protocol`


adinauer (approved):

imatwawana:
@marandaneto - can we go ahead and merge this?

PR #5382, opened 2022-08-04
Add trace-propagation-targets to unified API options

lforst:
Adds documentation for the discussed `tracePropagationTargets` option. More context: [DACI](https://www.notion.so/sentry/DACI-Control-Sending-Trace-Data-63fd04ea885443ee99514c15a1e91499) (internal)

Related to changes in the JavaScript SDK: https://github.com/getsentry/sentry-javascript/pull/5521

### DIFF
--- a/src/platforms/common/configuration/options.mdx
+++ b/src/platforms/common/configuration/options.mdx
@@ -437,7 +437,7 @@ If the callback is not set, or it returns `undefined`, the default naming scheme
 
 <ConfigKey name="enabled" supported={["javascript", "node"]}>
 
-Specifies whether this SDK should send events to Sentry. Defaults to `true`.  Setting this to `enabled: false` doesn't prevent all overhead from Sentry instrumentation. To disable Sentry completely, depending on environment, call `Sentry.init` conditionally.
+Specifies whether this SDK should send events to Sentry. Defaults to `true`. Setting this to `enabled: false` doesn't prevent all overhead from Sentry instrumentation. To disable Sentry completely, depending on environment, call `Sentry.init` conditionally.
 
 </ConfigKey>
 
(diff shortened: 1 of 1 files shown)


src/platforms/common/configuration/options.mdx
```diff
@@ -595,6 +595,14 @@ An optional property that configures which downstream services receive the `sent
 
 </ConfigKey>
 
+<ConfigKey name="trace-propagation-targets" supported={["node"]}>
+
+An optional property that controls which downstream services receive tracing data, in the form of a `sentry-trace` and `baggage` headers attached to any outgoing HTTP requests.
+It contains a list of URLs or regex against which URLs are matched.
+If this option is not provided, trace data is attached to every outgoing request from the instrumented client.
```
Lms24, line 602: I know that this an optional property but should we still add a sentence that by default/if it's not specified, trace data is sent everywhere?

Lms24 (commented):
LGTM so far. Just wondering if we should add a small example (similar to the one in the dev spec maybe?) to show what to specify here? The example could show a mix of regex and string and just a few URLs that would (not) match

src/platforms/common/configuration/options.mdx
lforst: Are you suggesting that we remove that last sentence, or are you suggesting that we add a sentence that says it attaches trace data to every request? If latter, we already have that last sentence (line 602).

lforst:
> LGTM so far. Just wondering if we should add a small example (similar to the one in the dev spec maybe?) to show what to specify here? The example could show a mix of regex and string and just a few URLs that would (not) match

@Lms24 How would you go about putting an example here? I thought about it a bit but we can't really rely on JS syntax on this page because it is more or less generic to all SDKs so we'd have to put different examples for all the different SDKs here. Do we want to do that?

src/platforms/common/configuration/options.mdx
Lms24: Oh sorry, I must have missed that. Massive brainfart

Lms24:
Good point. So, I took another look at the file and it seems like there are some options (e.g. in the deprecated section) with code examples. However, I'm realizing that this isn't simple for our case because ultimately this option should make it to all SDKs. I think we can do two things:

1. Within the `tracePropagationTargets` section, we use another component to show the code example in the specific language (I think `PlatformSection` is the one we use for that?). When SDKs add the new property, they'd have to add an example in their language
2. We don't specify an example and just improve the description a little. We should explain that the list can have a mix of strings and regexes and that for strings a "contains" condition is enough (i.e. it doesn't have to be an exact match). 

I'd vote for 2 as it seems the more straight forward way

Lms24 (approved):
LGTM, thanks for addressing my feedback!

PR #5394, opened 2022-08-08
docs(metrics): Add search bar documentation

DominikB2014:
For MEP, our search bar functionality will have to change, this PR documents the changes.
It includes a
1. A summary of the changes/functionality under "performance monitoring > filters and display"
image

2. A link/alert from 2 locations of the documentation, "performance monitoring" and "sentry basics > search" (because performance search is mentioned here as well).
image


src/docs/product/performance/filters-display/index.mdx
```diff
@@ -38,6 +38,23 @@ When you select "All Transactions", the table includes all transactions across a
...
+
+This feature is in beta and is only available if your organization is participating in its limited release. Features in beta are still in-progress and may have bugs. We recognize the irony. If you’re interested in participating, [join the waitlist](https://sentry.io/for/performance/#updates-signup).
+
+</Note>
+
+The transaction filter on the performance page has been simplified to support metric enhanced performance. The search bar can only be used to filter by transaction name.
+
+![Updated transaction search](transaction-name-search.png)
+
+When you begin typing a list of transaction names that match your search appear. There are two ways to interact with the search results.
```
0Calories, line 53:
```suggestion
When you begin typing, a list of transaction names that match your search appear. There are two ways to interact with the search results:
```

src/docs/product/performance/filters-display/index.mdx
```diff
@@ -38,6 +38,23 @@ When you select "All Transactions", the table includes all transactions across a
...
+
+1. Clicking on a transaction name will apply a filter on the performance page to only display transaction data for the selected transaction.
```
0Calories, line 55:
```suggestion
1. Clicking on a transaction name will apply a filter on the Performance page to only display transaction data for the selected transaction.
```

src/docs/product/performance/filters-display/index.mdx
```diff
@@ -38,6 +38,23 @@ When you select "All Transactions", the table includes all transactions across a
...
+2. Clicking the upwards arrow on a search result will take you to that transactions summary page.
```
0Calories, line 56:
```suggestion
2. Clicking the upwards arrow icon on a search result will take you to that transaction's summary page.
```

src/docs/product/performance/index.mdx
```diff
@@ -26,6 +26,12 @@ Using the information on this page, you can trace issues back through services (
 
 ## Filter Performance Data
 
+<Alert level="info" title="Important">
+
+For organizations participating in the limited release, the transaction filter on the performance page has been simplified to support metric enhanced performance. The search bar can only be used to filter by transaction name. See [transaction search](/product/performance/filters-display/#transaction-search).
```
0Calories, line 31:
```suggestion
For organizations participating in the limited release, the transaction filter on the Performance page has been simplified to support metric enhanced performance. The search bar can only be used to filter by transaction name. See [transaction search](/product/performance/filters-display/#transaction-search).
```

Also, I think Metrics Enhanced Performance is a term that we use internally only, so maybe we should say something else here instead?

nanoburd:
Can you replace "metrics enhanced performance" with "metrics" and link to https://docs.sentry.io/product/sentry-basics/metrics/

nanoburd:
Please also include the beta banner. That you see in other documentation that's already in prod.

src/docs/product/sentry-basics/search/index.mdx
```diff
@@ -12,6 +12,12 @@ Search is available on several features throughout [sentry.io](https://sentry.io
 
 ## Query Syntax {#syntax}
 
+<Alert level="info" title="Important">
+
+For organizations participating in the limited release, the transaction filter on the performance page has been simplified to support metric enhanced performance. The search bar can only be used to filter by transaction name. See [transaction search](/product/performance/filters-display/#transaction-search).
+
+</Alert>
+
```
gggritso, line 20: I think it would be good to deemphasize this by moving it lower in the document. The transaction search limitation is only in one place, and it only affects the raw string! No need to scare people at the very top of the page. Maybe this could live under the "The token `example error`" paragraph?

src/docs/product/performance/index.mdx
```diff
@@ -26,6 +26,12 @@ Using the information on this page, you can trace issues back through services (
 
 ## Filter Performance Data
 
+<Alert level="info" title="Important">
+
+For organizations participating in the limited release, the transaction filter on the Performance page has been simplified to support metric enhanced performance. The search bar can only be used to filter by transaction name. See [transaction search](/product/performance/filters-display/#transaction-search).
+
+</Alert>
+
```
gggritso, line 34: Might be good to clarify the limited release, and to explain the difference from other searches! Maybe something like

"The transaction filter has been simplified for organizations participating in the limited release of Metrics Enhanced Performance. See blah blah"

Then the "transaction search" section can go into more detail

src/docs/product/performance/filters-display/index.mdx
```diff
@@ -38,6 +38,23 @@ When you select "All Transactions", the table includes all transactions across a
 
 "Mobile" puts an emphasis on [Mobile Vitals](/product/performance/mobile-vitals/). Note that if you've selected a non-mobile project, this mode also displays all of those transactions.
 
+## Transaction Search
```
gggritso, line 41: This section feels like it fits nicely under the heading just below! You could rename "Search Condition Filters" to "Transaction Search", and combine them. Both sections are about the search bar, and this `<Note>` could be a caveat

src/docs/product/performance/filters-display/index.mdx
```diff
@@ -38,6 +38,23 @@ When you select "All Transactions", the table includes all transactions across a
...
+
+<Note>
+
+This feature is in beta and is only available if your organization is participating in its limited release. Features in beta are still in-progress and may have bugs. We recognize the irony. If you’re interested in participating, [join the waitlist](https://sentry.io/for/performance/#updates-signup).
+
+</Note>
+
+The transaction filter on the performance page has been simplified to support metric enhanced performance. The search bar can only be used to filter by transaction name.
```
gggritso, line 49: Maybe helpful to clarify the difference from the regular search bar. Maybe something like "The search bar can only be used to filter by transaction name, rather than to search for transactions matching the query".

gggritso: Also, I wonder if MEP is a marketing-visible feature, in which case it should be capitalized?

gggritso (commented):
Fair's fair! Nits and questions only!

src/docs/product/sentry-basics/search/index.mdx
DominikB2014: I was mixed between putting it at the top, the reason I put it at the top is because it is what you see when you click "read the docs" in the search bar.
image

So it might be confusing for a user, if they click read the docs while searching for a transaction, only to figure out later in the documentation that the query syntax doesn't apply.

src/docs/product/performance/index.mdx
DominikB2014: Yeah I do agree, I was aiming to be consistent with the following documentation, as we have a standard blurb before any beta feature
https://sentry.slack.com/archives/C02N6582ZFA/p1659603015550699?thread_ts=1659573427.357199&cid=C02N6582ZFA

I feel like we'd have to update the blurb everywhere.

src/docs/product/performance/filters-display/index.mdx
DominikB2014: As suggested by @nanoburd we replaced the MEP with metrics, as metrics has documentation to link to. Also, I agree, and reworded the search bar docs as suggested!

src/docs/product/performance/index.mdx
gggritso: I'm confused, the banner there is different from yours, what's being kept consistent? Also, the banner there is under the "Server-side sampling" heading, which is indeed in limited release. Here, the alert implies that "Filter Performance Data" is in limited release which is ... kind of the case? Maybe PM should tiebreak this one

DominikB2014: I see the confusion, so this alert I meant to put an alert I wanted to state that this doesn't really apply if your in the MEP release. And to click the link to see the updated functionality for transaction name search.

DominikB2014:
@gggritso and @0Calories Updated docs

As you will either be able to search for transaction name only, or add filters. I've updated the docs so that both the transaction name only search and search condition filters are under the same heading. And the note mentions that if you are not in beta, you will use the condition filter instead.

image

Also, changed the alert to a note on the pages that reference search

image

gggritso:
@DominikB2014 right on, I think this is a lot clearer!

DominikB2014:
@nanoburd I've updated the banner with the changes, let me know what you think!

gggritso (approved):
🖊️ 👍🏻

0Calories (approved):

PR #5395, opened 2022-08-08
SF Install : Missing extends AbstractController

mathieu-ducrot:
Just a little missing typo that I find will integrating the SentryTestController myself :)

### Legal Boilerplate

### DIFF
--- a/src/platforms/php/guides/symfony/index.mdx
+++ b/src/platforms/php/guides/symfony/index.mdx
@@ -118,9 +118,10 @@ To test that both logger error and exception are correctly sent to sentry.io, yo
 namespace App\Controller;
 
 use Psr\Log\LoggerInterface;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\Routing\Annotation\Route;
 
-class SentryTestController
+class SentryTestController extends AbstractController
 {
     /**
      * @var LoggerInterface


src/platforms/php/guides/symfony/index.mdx
```diff
@@ -118,9 +118,10 @@ To test that both logger error and exception are correctly sent to sentry.io, yo
 namespace App\Controller;
 
 use Psr\Log\LoggerInterface;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\Routing\Annotation\Route;
 
-class SentryTestController
+class SentryTestController extends AbstractController
```
cleptric, line 124: Does the example use any methods provided by `AbstractController`?

Based on https://symfony.com/doc/current/controller.html#the-base-controller-class-services, this looks optional.

mathieu-ducrot: @cleptric I just retested it with a brand new symfony install (with the CLI `symfony new sentry_test`) and here is what happen.
The example doesn't use any method from the Abstract indeed, but the fact that example use the service LoggerInterface which is private then you will have this error from Symfony : 

> The controller for URI "/_sentry-test" is not callable: Controller "App\Controller\SentryTestController" cannot be fetched from the container because it is private. Did you forget to tag the service with "controller.service_arguments"?

The easiest to fix this is by extending the AbstractController or, as the exception explain, tag your class as a controller with the tag controller.service_arguments which is not mention either at the moment on your doc.

cleptric (approved):

PR #5402, opened 2022-08-09
docs(perf): Clarify widget overall number values

gggritso:
Follow-up to https://github.com/getsentry/sentry/pull/37470. Explains why the values in the top right of the widgets might be surprising for sparse data sets. Related to PERF-1383

### DIFF
--- a/src/docs/product/performance/metrics.mdx
+++ b/src/docs/product/performance/metrics.mdx
@@ -81,6 +81,12 @@ User Misery is a user-weighted performance metric to assess the relative magnitu
 
 You can set satisfactory thresholds for each project with [custom thresholds](#custom-thresholds).
 
+<Note>
+
+User Misery is calculated using a probability distribution. This prevents false positives and ensures reasonable values for data sets with low event volume.
+
+</Note>
+
 ## Custom Thresholds
 
 For each project, you can configure how [Apdex](#apdex) and [User Misery](#user-misery) are calculated in **[Project] > Settings > Performance**. You can override project-level settings at the transaction level in **Transaction Summary > Settings**.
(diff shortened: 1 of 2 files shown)


DominikB2014 (commented):
🤔 knowing that there could be users that don’t have a strong stats background, is it okay to leave it as probabilistic distribution? Should we link some further explanation?

src/docs/product/performance/metrics.mdx
```diff
@@ -81,6 +81,12 @@ User Misery is a user-weighted performance metric to assess the relative magnitu
 
 You can set satisfactory thresholds for each project with [custom thresholds](#custom-thresholds).
 
+<Note>
+
+User Misery is calculated using a probability distribution. This prevents false positives and ensures reasonable values for data sets with low event volume.
```
DominikB2014, line 86: Meant to add my comment to this line!

gggritso: I figure this is fine! I could have said "Beta Distribution" but I'm choosing to wave my hands a bit and say "it's statistical" so people get the idea that it's non-trivial. I think anyone who's using Sentry has at least _heard_ of statistics, right?

k-fish: I'd maybe clarify a bit, since `User Misery`, the rate / visualization are a prob. distribution, but the underlying data `# of miserable users` should always be correct.

gggritso: @k-fish clarified in https://github.com/getsentry/sentry-docs/pull/5402/commits/26bd98b99c2f30e350ce6e0a16654106aea9833d. How's that?

k-fish: It should help 👍

k-fish (approved):
The technical jargon here should be correct now, looks fine on our product side. Might want to get a docs adherence check at some point, but I think @imatwawana is still away.

gggritso:
@k-fish makes sense—I'll follow up with her next week!

PR #5412, opened 2022-08-11
show dif page for dotnet

bruno-garcia:
### Legal Boilerplate

### DIFF
--- a/src/platforms/common/data-management/debug-files/index.mdx
+++ b/src/platforms/common/data-management/debug-files/index.mdx
@@ -9,6 +9,8 @@ supported:
   - cocoa
   - apple
   - unity
+  - dotnet
+  - dotnet.maui
 description: "Learn about how debug information files allow Sentry to extract stack traces and provide more information about crash reports for most compiled platforms."
 ---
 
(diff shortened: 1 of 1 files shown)


mattjohnsonpint (approved):

mattjohnsonpint:
Actually, maybe we should wait until we resolve this?

image

bruno-garcia:
> Actually, maybe we should wait until we resolve this?
> 
> 
> 
> image
> 
> 

This documentation clarifies debug formats we rely on now for mobile. So I think we shouldn't wait.

Might be worth a Note on the top when rendering under dotnet. Saying it's just mobile right now. I'll add.

src/platforms/common/data-management/debug-files/index.mdx
```diff
@@ -9,6 +9,8 @@ supported:
   - cocoa
   - apple
   - unity
+  - dotnet
+  - dotnet.maui
```
bruno-garcia, line 13:
```suggestion
```

bruno-garcia:
Adding the page for now. It's just a general Debug Info documentation, not built to be specific to what exactly each platform can leverage (maybe just `native` other than ProGuard stuff).

I'll raise an issue to refactor this section to better include/exclude each debug info type where it makes sense to be available

Issue #5422, opened 2022-08-15
Relay: get-started: cannot follow initial howto


sebastian-philipp:
### Core or SDK?

Core Sentry product

### Which part? Which one?

Relay: get-started

### Description

```
$ docker run --rm -it                \
>   -v $(pwd)/config/:/work/.relay/  \
>   getsentry/relay                  \                                                                                                                                                            
>   config init
Initializing relay in /work/.relay
Do you want to create a new config?: Yes, create custom config
How should this relay operate?: Managed through upstream
upstream: https://sentry.io/
listen interface: 0.0.0.0
listen port: 3000
do you want listen to TLS no
error: could not write config file (file /work/.relay/config.yml)
  caused by: Permission denied (os error 13)
```
```
$ docker run --rm -it                \
>   -v $(pwd)/config/:/work/.relay/:z  \
>   getsentry/relay                  \
>   config init
Initializing relay in /work/.relay
Do you want to create a new config?: Yes, create custom config
How should this relay operate?: Managed through upstream
upstream: https://sentry.io/
listen interface: 0.0.0.0
listen port: 3000
do you want listen to TLS no
Generated new credentials
  relay id: something
  public key: something
All done!
```
Please not the `:z` mount option. My system is: "Fedora Linux 35" plus Docker version 20.10.12

### Suggested Solution

Add `:z`

getsentry-release:
Routing to @getsentry/ingest for [triage](https://develop.sentry.dev/processing-tickets/#3-triage). ⏲️

jjbayer:
@sebastian-philipp did you run both docker commands, first the one that configures ownership

```bash
docker run --rm -it                \
  -v $(pwd)/config/:/work/.relay/  \
  --entrypoint bash                \
  getsentry/relay                  \
  -c 'chown -R relay:relay /work/.relay'
  ```
  
  and then the config initialization
  
  ```bash
  docker run --rm -it                \
  -v $(pwd)/config/:/work/.relay/  \
  getsentry/relay                  \
  config init
  ```
  
  I can reproduce the error you are seeing by omitting the first command and only running the second command.

sebastian-philipp:
Right, I had to add `:z` to the first one as well. Then I faced the issue a second time and then I felt the urge to create an issue for it.

jjbayer:
@sebastian-philipp thanks for reporting this, I opened a [PR](https://github.com/getsentry/sentry-docs/pull/5427) to update the docs.

PR #5427, opened 2022-08-16
Add :z to volume mount in Relay example

jjbayer:
Apparently the [`:z` option](https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label) is necessary for users of [SELinux](https://en.wikipedia.org/wiki/Security-Enhanced_Linux).

Fixes https://github.com/getsentry/sentry-docs/issues/5422

### DIFF
--- a/src/docs/product/relay/getting-started.mdx
+++ b/src/docs/product/relay/getting-started.mdx
@@ -32,7 +32,8 @@ container and copying in the files.
 
 ```shell {tabTitle:Run in Docker}
 # Adjust permissions for the configuration directory
-# (the :z option is only necessary on selinux)
+# The :z option is only necessary on selinux.
+# See https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label.
 docker run --rm -it                  \
   -v $(pwd)/config/:/work/.relay/:z  \
   --entrypoint bash                  \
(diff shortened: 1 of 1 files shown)


olksdr (commented):
General comment on this: 
It might be worth  mentioning that this is for the cases when the Selinux is enabled (in force mode) on the host machine. Since without the context, just adding `z`  make little sense here.

jjbayer:
> General comment on this: It might be worth mentioning that this is for the cases when the Selinux is enabled (in force mode) on the host machine. Since without the context, just adding `z` make little sense here.

@olksdr agreed. I have a comment on line 35: https://github.com/getsentry/sentry-docs/blob/44a8f34a212d9c0e58641901dbcdabc7e098cb40/src/docs/product/relay/getting-started.mdx?plain=1#L35, do you think that's enough context? Open to suggestions.

olksdr:
I think the comment is a little bit hard to notice there. 
And also it would make sense to put there somewhere a link to the docker documentation, the one you posted in the description (like, https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label) which has some context on why this is necessary and what can go wrong.  What do you think?

olksdr (approved):
Approving, since my comment is not blocking

jan-auer (approved):

PR #5435, opened 2022-08-17
Add Yarn as a dependency for development

MatthiasPortzel:
I had to install yarn for this, and it would have saved me 2 seconds if it had been specified. If there's a reason to not list it explicitly, that's fine. But I figure while I have the project cloned I'll make a PR.

* Also, remove an unused Gatsby link
* Also, make the link to localhost clickable

### Legal Boilerplate

### DIFF
--- a/bin/bootstrap
+++ b/bin/bootstrap
@@ -12,7 +12,7 @@ if ! [ -x $(command -v homebrew) ]; then
 fi
 
 # Install system toolchain
-brew install volta pre-commit
+brew install volta pre-commit yarn
 
 # App dependencies
 make
--- a/src/docs/contributing/environment.mdx
+++ b/src/docs/contributing/environment.mdx
@@ -4,7 +4,7 @@ noindex: true
 sidebar_order: 1
 ---
 
-You will need [Volta][volta] and [pre-commit][pre-commit] installed. If you don't have opinions about the process, this will get you going:
+You will need [Yarn][yarn], [Volta][volta], and [pre-commit][pre-commit] installed. If you don't have opinions about the process, this will get you going:
 
 ```bash
 # Install Homebrew and everything mentioned above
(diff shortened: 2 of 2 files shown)


bin/bootstrap
```diff
@@ -12,7 +12,7 @@ if ! [ -x $(command -v homebrew) ]; then
 fi
 
 # Install system toolchain
-brew install volta pre-commit
+brew install volta pre-commit yarn
```
AbhiPrasad, line 15: yarn should be installed with volta - https://volta.sh/, hence why it's not part of the install here

MatthiasPortzel: Hm, so should there be a `volta install` in this script, then? When I installed these two things and then ran `make` I received an error that yarn wasn’t found.

MatthiasPortzel: I will also re-try the install later and see if I just did something weird the first time, because I might have.

MatthiasPortzel: I uninstalled Volta and yarn and tried actually running `. bin/bootstrap`, to the same result (yarn not found). Volta installs, but the script then immediately calls `make`, which immediately calls `yarn`, which fails.

After some digging:
`brew install volta` installs Volta but doesn't set it up (nor give any indication that it requires further setup, unfortunately).

It looks like setup consists of creating and populating `~/.volta` (or `$VOLTA_HOME`) and then adding `export VOLTA_HOME="$HOME/.volta"; export PATH="$VOLTA_HOME/bin:$PATH"` to your shell startup script.

Volta can set itself up with the `volta setup` command. The (non-homebrew) Volta installer also runs this setup command automatically.

This file is broken as is. One easy solution would be adding `volta run yarn` to the Makefile. Otherwise you need to add some lines here to ensure Volta is set up and `$VOLTA_HOME/bin` is in the front of the path. (That's assuming you want to keep using Volta and don't want to bother up-streaming any changes.)

PR #5436, opened 2022-08-17
Add User Feedback section for Rails

MatthiasPortzel:
Adds information on integrating the JavaScript User Feedback SDK into a Rails app. It's modeled off the existing Python documentation.

It also serves as a way to document the env variable `sentry.error_event_id` which was added in https://github.com/getsentry/sentry-ruby/pull/1849 (CC: @st0012).

This has the unfortunate side-effect of populating the Ruby-only version of this page, when the Ruby SDK doesn't actually support user feedback. But there's no easy way to show this page for Rails and not Ruby. Python also has the same problem, so I don't think this should block merging.

I recently implemented the described behavior in our Rails app, so I thought I'd add a couple words to the documentation. Feedback is welcome.

### Legal Boilerplate

### DIFF
--- a/src/includes/enriching-events/user-feedback/example-widget/ruby.rails.mdx
+++ b/src/includes/enriching-events/user-feedback/example-widget/ruby.rails.mdx
@@ -1,6 +1,6 @@
-In Rails, being able to serve dynamic (e.g. `.html.erb`) pages in response to errors is required in order to pass the needed `event_id` to the JavaScript SDK. Normally, this looks like overwriting the `config.exceptions_app` setting to point at your own router, and then calling your own controller to handle the error. Several tutorials or gems exist to allow for this.
+In Rails, being able to serve dynamic (for example, `.html.erb`) pages in response to errors is required to pass the needed `event_id` to the JavaScript SDK. Typically, this is done by overwriting the `config.exceptions_app` setting to point to your own router, and then calling your own controller to handle the error. You can find more detailed instructions for this configuration online or install a gem to handle this for you.
 
-Once you are able to serve dynamic exception pages, supporting user feedback is straightforward.
+Once you're able to serve dynamic exception pages, you can support user feedback.
 
 Make sure you've got the JavaScript SDK available:
 
@@ -12,7 +12,7 @@ Make sure you've got the JavaScript SDK available:
 ></script>
 ```
 
-And the template that brings up the dialog:
+Additionally, you need the template that brings up the dialog:
 
 ```ERB
 <% sentry_id = request.env["sentry.error_event_id"] %>
--- a/src/includes/enriching-events/user-feedback/example-widget/ruby.rails.mdx
+++ b/src/includes/enriching-events/user-feedback/example-widget/ruby.rails.mdx
@@ -1,0 +1,25 @@
+In Rails, being able to serve dynamic (e.g. `.html.erb`) pages in response to errors is required in order to pass the needed `event_id` to the JavaScript SDK. Normally, this looks like overwriting the `config.exceptions_app` setting to point at your own router, and then calling your own controller to handle the error. Several tutorials or gems exist to allow for this.
+
+Once you are able to serve dynamic exception pages, supporting user feedback is straightforward.
+
+Make sure you've got the JavaScript SDK available:
+
+```ERB
+<script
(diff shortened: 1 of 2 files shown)


st0012:
@sl0thentr0py can you take a look? thx

sl0thentr0py (approved):
technical stuff is ok, @imatwawana leaving the wording/grammar to you :)

imatwawana:
> This has the unfortunate side-effect of populating the Ruby-only version of this page, when the Ruby SDK doesn't actually support user feedback. But there's no easy way to show this page for Rails and not Ruby. Python also has the same problem, so I don't think this should block merging.

I'm gonna pull in your branch into my local to test a way around this; if it works, I'll push it. I'll circle back to do review of this PR for wording after that!

imatwawana:
I've made a very ugly hack, but it resolves the issue of having the content show up for Ruby when it's only supported on Rails.

src/includes/enriching-events/user-feedback/example-widget/ruby.rails.mdx
```diff
@@ -0,0 +1,25 @@
+In Rails, being able to serve dynamic (e.g. `.html.erb`) pages in response to errors is required in order to pass the needed `event_id` to the JavaScript SDK. Normally, this looks like overwriting the `config.exceptions_app` setting to point at your own router, and then calling your own controller to handle the error. Several tutorials or gems exist to allow for this.
```
imatwawana, line 1: I'm not clear on what you mean in the last sentence here. Do you mean there are resources available to teach people how to do what you're describing? And are you using "gems" in a colloquial way or is there some other meaning here?

```suggestion
In Rails, being able to serve dynamic (for example, `.html.erb`) pages in response to errors is required to pass the needed `event_id` to the JavaScript SDK. Typically, this is done by overwriting the `config.exceptions_app` setting to point to your own router, and then calling your own controller to handle the error. Several tutorials or gems exist to allow for this.
```

src/includes/enriching-events/user-feedback/example-widget/ruby.rails.mdx
```diff
@@ -0,0 +1,25 @@
...
+
+Once you are able to serve dynamic exception pages, supporting user feedback is straightforward.
```
imatwawana, line 3: We try to avoid telling people what their experience of carrying out a task will be in case it doesn't end up that way for them :-)

```suggestion
Once you're able to serve dynamic exception pages, you can support user feedback.
```

src/includes/enriching-events/user-feedback/example-widget/ruby.rails.mdx
```diff
@@ -0,0 +1,25 @@
...
+
+```ERB
+<script
+  src="https://browser.sentry-cdn.com/{{ packages.version('sentry.javascript.browser') }}/bundle.min.js"
+  integrity="sha384-{{ packages.checksum('sentry.javascript.browser', 'bundle.min.js', 'sha384-base64') }}"
+  crossorigin="anonymous"
+></script>
+```
+
+And the template that brings up the dialog:
```
imatwawana, line 15:
```suggestion
Additionally, you need the template that brings up the dialog:
```

imatwawana (commented):
Thanks for submitting this @MatthiasPortzel! Mostly the wording looks good, but I've left a couple questions where I'm not clear on what you're trying to convey.

src/includes/enriching-events/user-feedback/example-widget/ruby.rails.mdx
MatthiasPortzel: "gem" is the term for a Ruby library or package.
What I'm trying to say here is that people should consult external resources, whether that's following a tutorial to do it themselves, or installing a gem to handle this. This is in part because the details are hairy, and partly because I don't think these docs should be opinionated about how to do this.
There's definitely a less-awkward way of wording it. Maybe, "More detailed instructions on this configuration are available online. Alternatively, you can install a gem to handle this for you."

imatwawana: Thanks for the clarification. I've gone ahead and updated my suggestion.

imatwawana:
@MatthiasPortzel - can I commit my suggestions and get this merged for you?

MatthiasPortzel:
Yes, your changes sound good.

PR #5445, opened 2022-08-18
Make sure users do not set `none` for performance metrics

AbhiPrasad:


AbhiPrasad:
Nvm we can do this for custom perf metrics - ignore!

PR #5451, opened 2022-08-19
format links as internal

imatwawana:
Update links that are set as full URLS (external) to relative internal links


gggritso (approved):
🧹

PR #5452, opened 2022-08-19
clarify initial premise of issue alerts

souredoutlook:
Some users miss this point and expect issue alerts to be constantly scanning. It is reiterated further down after a screenshot though.

### Legal Boilerplate

### DIFF
--- a/src/docs/product/alerts/alert-types.mdx
+++ b/src/docs/product/alerts/alert-types.mdx
@@ -18,7 +18,7 @@ You can create two types of alerts:
 
 ## Issue Alerts
 
-Issue alerts trigger whenever a new event is received and any issue in a project matches the specified criteria. These criteria might be, for example, a resolved issue re-appearing or an issue affecting many users.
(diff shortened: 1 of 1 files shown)


src/docs/product/alerts/alert-types.mdx
```diff
@@ -18,7 +18,7 @@ You can create two types of alerts:
 
 ## Issue Alerts
 
-Issue alerts trigger whenever any issue in a project matches the specified criteria. These criteria might be, for example, a resolved issue re-appearing or an issue affecting many users.
+Issue alerts trigger whenever a new event is received and any issue in a project matches the specified criteria. These criteria might be, for example, a resolved issue re-appearing or an issue affecting many users.
```
imatwawana, line 21: It almost sounds like the two states are disconnected, so I'm trying to create some connective tissue between them. Does this work?

```suggestion
Issue alerts trigger whenever a new event is received for any issue in a project matching the specified criteria. These criteria might be, for example, a resolved issue re-appearing or an issue affecting many users.
```

imatwawana (commented):
Made a suggestion that I think works

Issue #5441, opened 2022-08-18
Clarify Global scope for Node SDK


rbisol:
### Core or SDK?

Platform/SDK

### Which part? Which one?

NodeJS SDK

### Description

The current text here is: https://docs.sentry.io/platforms/node/enriching-events/context/#clearing-context

```
This will be changed for all future events:

Sentry.setUser(someUser); //SetContext as well
```

However, the scope in Node is per router and if the context or user is added inside a router, it will not be available to other routers. This is expected and can be reproduced with:

```
Sentry.init({....});

app.use(Sentry.Handlers.requestHandler());

Sentry.setContext("TrueGlobalContext",{foo:'This context will be present in all future events'});

// All controllers should live here
app.get("/", function rootHandler(req, res) {
  Sentry.setContext("FakeGlobalContext",{foo:'This context will be present only in events from this request'});
  res.end("End request")
});

// The error handler must be before any other error middleware and after all controllers
app.use(Sentry.Handlers.errorHandler());
```

### Suggested Solution

The doc currently has a single text for all platforms/frameworks. The suggestion is to add notes or comments for frameworks where the behaviour is different.

imatwawana:
Looks like a platformsection just needs to be set up for that section of the page to address this.

getsentry-release:
Routing to @getsentry/team-web-sdk-frontend for [triage](https://develop.sentry.dev/processing-tickets/#3-triage). ⏲️

PR #5461, opened 2022-08-29
Clarify Global scope for Node SDK

lforst:
I tried my best at cleaning up the "clearing scope" section. I have absolutely no clue what concept it was trying to convey. The documentation's description was just *so so* far off compared to what the SDK actually does.

Resolves https://github.com/getsentry/sentry-docs/issues/5441


Lms24 (approved):
Nice improvement!

PR #5468, opened 2022-08-30
fix(mep): Update metric fields

wmak:
- Some fields & functions were marked incorrectly for metrics.
- Removing os.name & browser.name for now instead of having a TBA
  definition

### DIFF
--- a/src/docs/product/sentry-basics/search/searchable-properties.mdx
+++ b/src/docs/product/sentry-basics/search/searchable-properties.mdx
@@ -89,17 +89,16 @@ Release properties overlap with event and issue search properties, so they don't
 | Key/Token | Description | E | I | M | Type |
 |---|---|---|---|---|---|
 | age | Returns issues created since the time defined by the value. The syntax is similar to the Unix find command. Supported suffixes: `m - minutes`, `h - hours`, `d - days`, `w - weeks`. For example, `age:-24h` returns isssues that are new in the last 24 hours, while `age:+12h` returns ones that are older than 12 hours. Entering `age:+12h age:-24h` would return issues created between 12 and 24 hours ago. |  | x |  | relative time |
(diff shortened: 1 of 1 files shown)


src/docs/product/sentry-basics/search/searchable-properties.mdx
```diff
@@ -144,34 +143,33 @@ Release properties overlap with event and issue search properties, so they don't
...
 | os.build | The internal build revision of the operating system. | x | x |  | string |
-| os.name | TBA |  |  | x | TBA |
 | os.kernel_version | The independent kernel version string. This is typically the entire output of the `uname` syscall. | x | x |  | string |
-| percentile(field,level) | Returns results with an approximate percentile of the field to the level. The level can be between `0` and `1`. For example, if you wanted to find the 50th percentile of transaction durations, you would enter `percentile(transaction.duration, 0.5)`. | x |  |  | number |
-| platform | Name of the platform. This is only a metrics property for [valid platforms](https://github.com/getsentry/relay/blob/9d0d2dc54233d46043a369e1dacf7db900953c14/relay-general/src/protocol/constants.rs#L1-L21), defaulting to `other`. | x |  | x | string |
+| percentile(field, level) | Returns results with an approximate percentile of the field to the level. The level can be between `0` and `1`. For example, if you wanted to find the 50th percentile of transaction durations, you would enter `percentile(transaction.duration, 0.5)`. | x |  | x | number |
```
imatwawana, line 168: So we actually don't include a space on any of the entries where we have multiple words in the parentheses. That's the way these were entered in the list that we worked from in Notion. I assume it's because in the application, there's no space there and if you were writing that query (though obviously it auto-populates for you), you wouldn't include a space. I'll defer to you on this though. Do what you think makes sense.

imatwawana (approved):
Looks great! I left one comment, but otherwise, it LGTM. Thanks!

PR #5471, opened 2022-08-31
fix(react): Update React Router v4/v5 to make paramaterization more clear

AbhiPrasad:
To make sure we prevent possible mis-configurations with the React Router v4/v5 integration, this PR updates our docs around it. We first move the usage of `withSentryRouting` higher up, and encourage it as the default approach to use for paramaterization. We then move down the route config option below that, and offer it as a way for users to define paths outside of the usage of `withSentryRouting`.

### DIFF
--- a/src/platforms/javascript/guides/react/configuration/integrations/react-router.mdx
+++ b/src/platforms/javascript/guides/react/configuration/integrations/react-router.mdx
@@ -79,9 +79,9 @@ The React router instrumentation uses the React router library to create `pagelo
 
 ### Parameterized Transaction Names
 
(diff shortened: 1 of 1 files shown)


src/platforms/javascript/guides/react/configuration/integrations/react-router.mdx
```diff
@@ -79,29 +79,26 @@ The React router instrumentation uses the React router library to create `pagelo
 
 ### Parameterized transaction names
```
imatwawana, line 80:
```suggestion
### Parameterized Transaction Names
```

imatwawana (approved):
Made some wording edits, but otherwise looks good!

k-fish (approved):

PR #5478, opened 2022-09-01
Emphasize the supported platforms on the product page

jernejstrasner:
We started getting inbound email originating from docs but not understanding that only iOS and Android are supported currently. This should help with that. See screenshot.
Screen Shot 2022-09-01 at 17 38 18


imatwawana (approved):
LGTM!

PR #5483, opened 2022-09-05
Mention Sentry wrap and code push for RN

marandaneto:
Reason: https://stackoverflow.com/questions/73273418/how-to-wrap-app-component-while-using-sentry-with-codepush-react-native

### DIFF
--- a/src/platforms/react-native/manual-setup/codepush.mdx
+++ b/src/platforms/react-native/manual-setup/codepush.mdx
@@ -46,7 +46,7 @@ codePush.getUpdateMetadata().then(update => {
 });
 ```
 
-In case you are wrapping your Root Component with `Sentry.wrap` to gain Performance features, `codePush` should be the outer most function because it needs access to the very root component in order to swap out the bundle.
+If you're wrapping your root component with `Sentry.wrap` to access Performance features, `codePush` should be the outer most function because it needs access to the very root component in order to swap out the bundle.
 
 ```javascript
 export default codePush(Sentry.wrap(App));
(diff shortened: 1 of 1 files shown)


src/platforms/react-native/manual-setup/codepush.mdx
```diff
@@ -46,6 +46,12 @@ codePush.getUpdateMetadata().then(update => {
 });
 ```
 
+In case you are wrapping your Root Component with `Sentry.wrap` to gain Performance features, `codePush` should be the outer most function because it needs access to the very root component in order to swap out the bundle.
```
imatwawana, line 49: I'm don't understand what you mean by "the very root component". Can you explain that or try wording it another way?

```suggestion
If you're wrapping your root component with `Sentry.wrap` to access Performance features, `codePush` should be the outer most function because it needs access to the very root component in order to swap out the bundle.
```

imatwawana (commented):
There's one part of what you've added that I don't understand. See my comments.

src/platforms/react-native/manual-setup/codepush.mdx
```diff
@@ -46,6 +46,12 @@ codePush.getUpdateMetadata().then(update => {
 });
 ```
 
+If you're wrapping your root component with `Sentry.wrap` to access Performance features, `codePush` should be the outer most function because it needs access to the very root component in order to swap out the bundle.
```
marandaneto, line 49:
```suggestion
If you're wrapping your root component with `Sentry.wrap` to access Performance features, `codePush` should be the outer most function because it needs access to the root component in order to swap out the bundle.
```

marandaneto:
> There's one part of what you've added that I don't understand. See my comments.

I've removed the `very`, it was a ctrl c ctrl v problem I guess.

imatwawana (approved):
Looks good!

PR #5486, opened 2022-09-06
(docs) Add RN SDK to supported list for DS

krystofwoldrich:
After DS support for RN is released this needs to be updated and published.

Related pull requests:
- https://github.com/getsentry/sentry-react-native/pull/2454

### DIFF
--- a/src/docs/product/data-management-settings/server-side-sampling/getting-started.mdx
+++ b/src/docs/product/data-management-settings/server-side-sampling/getting-started.mdx
@@ -24,6 +24,7 @@ Server-side sampling is currently only available for the following SDKs:
 - JavaScript: 7.6.0 or later
 - Apple: 7.23.0 or later
 - Android: 6.2.1 or later
+- React Native: 4.3.0 or later
 
 If your application relies on any Sentry SDK that isn't specified above, then you won’t be able to use server-side sampling.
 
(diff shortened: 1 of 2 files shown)


imatwawana:
That update will also need to be made here: 
https://docs.sentry.io/product/data-management-settings/server-side-sampling/current-limitations/

krystofwoldrich:
> That update will also need to be made here: https://docs.sentry.io/product/data-management-settings/server-side-sampling/current-limitations/

Thanks. I've missed that. Now it should be ready. And only waiting for the RN release.

marandaneto (approved):

src/docs/product/data-management-settings/server-side-sampling/getting-started.mdx
```diff
@@ -24,6 +24,7 @@ Server-side sampling is currently only available for the following SDKs:
 - JavaScript: 7.6.0 or later
 - Apple: 7.23.0 or later
 - Android: 6.2.1 or later
+- React Native: 4.3.0 or later
```
bruno-garcia, line 27: @imatwawana How about we have this list in an include? I feel we'll eventually go out of sync here

imatwawana: Yeah, good call, I can set that up. Coincidentally, we only got the ability to use an include on the product side of our docs a couple weeks ago.

PR #5147, opened 2022-06-15
docs: add alternative snippet that blocks on sentry-cli errors

armcknight:
Some developers may want errors from `sentry-cli` to break their builds, like if it's part of a CI deployment that shouldn't succeed, lest they encounter sentry issues with unsymbolicated symbols. This gives an example of each as copyable snippets, with a quick explainer on the difference.

Screen Shot 2022-06-15 at 6 10 27 PM
Screen Shot 2022-06-15 at 6 10 57 PM

PR #5500, opened 2022-09-09
fix: --force-foreground to break builds if upload fails

armcknight:
I added this snippet in #5147 but we never added the important flag that prevents the process from executing in the background, defeating the purpose of the snippet. I just noticed this behavior going through a new test app setup.

### DIFF
--- a/src/platform-includes/debug-symbols-apple/_default.mdx
+++ b/src/platform-includes/debug-symbols-apple/_default.mdx
@@ -196,7 +196,7 @@ if which sentry-cli >/dev/null; then
 export SENTRY_ORG=___ORG_SLUG___
 export SENTRY_PROJECT=___PROJECT_SLUG___
 export SENTRY_AUTH_TOKEN=YOUR_AUTH_TOKEN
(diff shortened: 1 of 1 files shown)


src/platform-includes/debug-symbols-apple/_default.mdx
```diff
@@ -196,7 +196,7 @@ if which sentry-cli >/dev/null; then
 export SENTRY_ORG=___ORG_SLUG___
 export SENTRY_PROJECT=___PROJECT_SLUG___
 export SENTRY_AUTH_TOKEN=YOUR_AUTH_TOKEN
-ERROR=$(sentry-cli upload-dif "$DWARF_DSYM_FOLDER_PATH" 2>&1 >/dev/null)
+ERROR=$(sentry-cli upload-dif "$DWARF_DSYM_FOLDER_PATH" 2>&1 >/dev/null --force-foreground)
```
kamilogorek, line 199:
```suggestion
ERROR=$(sentry-cli upload-dif "$DWARF_DSYM_FOLDER_PATH" --force-foreground 2>&1 >/dev/null)
```

kamilogorek (changes requested):
Redirects should be at the end of the command

kamilogorek (approved):

PR #5501, opened 2022-09-09
fix rails user feedback integration snippet

imatwawana:
Moves the Rails user feedback code snippet to the correct folder so it's displayed

Before (default snippet): 

After (Rails snippet):


imatwawana:
@MatthiasPortzel - Just FYI. I think we changed our folder structure around the time you created this PR and we missed that the Rails snippet wasn't displaying properly in the preview.

sl0thentr0py (approved):

MatthiasPortzel:
👍 Thanks for following through.

PR #5507, opened 2022-09-12
add to custom performance metrics section

imatwawana:
Add more content to Custom Performance Metrics section


src/docs/product/performance/metrics.mdx
```diff
@@ -95,4 +95,31 @@ The calculation method determines if duration is defined as the entire length of
...
+
+Although you can set custom metrics in the SDK, the ability to see the data generated by these metrics is is available only if you're in the Early Adopter program and on a plan that includes Performance Monitoring. Features available to Early Adopters are still in-progress and may have bugs. We recognize the irony. If you’re interested in being an Early Adopter, you can turn your organization’s Early Adopter status on/off in General Settings. This will affect all users in your organization and can be turned back off just as easily.
+
+</Note>
+
+In addition to the automatic performance metrics described above, Sentry supports setting custom performance metrics on transactions. Custom performance metrics allow you to define metrics (beyond the ones mentioned above) that are important to your application and send them to Sentry.
+
+For example, you might want to set a custom metric to track...
+
+<!-- need some use cases/examples here -->
```
wmak, line 108:
```suggestion
- Total memory usage during a transaction
- The amount of time being queried
- How many times a user performed an action during the transaction
```

jas-kas: Re: @wmak's commit suggestion, we might want to say "during **a** transaction" for both examples 1 and 3 for consistency.

For example 2, is that number of queries or actual query time? Actual query time seems like it should be a span? 

It might be helpful if we include an example for each custom perf metric with their associated unit (if applicable).

wmak (approved):
lgtm

src/docs/product/performance/metrics.mdx
```diff
@@ -95,4 +95,33 @@ The calculation method determines if duration is defined as the entire length of
...
+<Note>
+
+Although you can set custom metrics in the SDK, the ability to see the data generated by these metrics is is available only if you're in the Early Adopter program and on a plan that includes Performance Monitoring. Features available to Early Adopters are still in-progress and may have bugs. We recognize the irony. If you’re interested in being an Early Adopter, you can turn your organization’s Early Adopter status on/off in General Settings. This will affect all users in your organization and can be turned back off just as easily.
+
+</Note>
+
+In addition to the automatic performance metrics described above, Sentry supports setting custom performance metrics on transactions. Custom performance metrics allow you to define metrics (beyond the ones mentioned above) that are important to your application and send them to Sentry.
+
+For example, you might want to set a custom metric to track...
```
imatwawana, line 106:
```suggestion
For example, you might want to set a custom metric to track:
```

PR #5516, opened 2022-09-13
feat(perf-issues): Performance Issues EA documentation

gggritso:
First documentation for Performance Issues. In the future, we'll probably have a "Error Issue Details Page" and a "Performance Issue Details Page". We'll also updated all the related documentation, like search filters and definition. For now, adds a new "Performance Issues" page that mirrors the "Issue Details" page, but explains the general properties of the feature, usage, and limitations of performance issues.

- [x] improve screenshots
- [x] add screenshot for solved N+1 queries


src/docs/product/issues/performance-issues/index.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+## Performance Issue Limitations
+
+Performance issues have a few limitations:
+
+- Sentry only detects one type of issue: [N+1 Queries](n-one-queries)
+- You cannot create alerts for performance issues
+- Performance issues cannot be merged or deleted
+- Custom fingerprinting and grouping do not apply to performance issues
+- Performance issues only support simple ignore rules
```
DominikB2014, line 43: These are temporary limitation?

src/docs/product/issues/performance-issues/n-one-queries.mdx
```diff
@@ -0,0 +1,52 @@
...
+    books = Book.objects.all()[:10]
+    book_list = [book.title + " by " + book.author.name for book in books]
+    return HttpResponse((", ").join(book_list))
+```
+
+This code has a subtle performance problem. Each call to `book.author.name` makes a query to fetch the book's author! In total, this code makes 11 queries: one query to fetch the list of books, and 10 more queries to fetch the author of each book. This results in a characteristic query span waterfall:
+
+![N+1 queries in an example application](n-plus-one-database-queries-in-booker.png)
+
+In order to fix this issue, you could use the `selected_related` method in Django, like so:
```
DominikB2014, line 41: I read about using prefetch related as well to help solve N+1
https://stackoverflow.com/questions/31237042/whats-the-difference-between-select-related-and-prefetch-related-in-django-orm

tbh, i'm not super familiar with Django, so I'm not proposing any change to docs, but perhaps something to consider.

src/docs/product/issues/performance-issues/index.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+## How Performance Issues Work
+
+Sentry detects performance issues by scanning incoming transaction events. It checks various properties of the transaction (span durations, span arrangement, span types, etc.) to detect likely problems. Sentry then checks these properties against corresponding thresholds to determine whether a performance issue exists. Sentry then generates a unique problem fingerprint based on the problem type and location in the application, and uses that fingerprint to create a performance issue.
+
+## Tags
+
+Tags are key/value string pairs that are both indexed and searchable. For example, a tag provides you with information such as the browser, device, or user associated with the event. Tags are the diagnostic information sent by the SDK for the individual event.
+
```
DominikB2014, line 28: Tags are described in its own section as well https://docs.sentry.io/platforms/javascript/enriching-events/tags/#:~:text=Tags%20are%20key%2Fvalue%20string,for%20a%20set%20of%20events

Wondering if it would make sense we can share the definition or link one to another.

DominikB2014 (commented):
Some minor comments, but lgtm! But i'd say we should have more of performance chime in, as we all worked on different parts of the code!

src/docs/product/issues/index.mdx
```diff
@@ -42,6 +42,14 @@ Learn more about triaging issues and their different states in [Issue States and
 
 You can also add issues data to your [custom dashboards](/product/dashboards/custom-dashboards/), as widgets, using the [dataset selector](/product/dashboards/widget-builder/#choose-your-dataset).
 
+## Performance Issues
+
+<Include name="early-adopter-note.mdx" />
+
+In addition to collecting application errors, Sentry can also collect performance problems. If your application is configured for [Performance Monitoring](/product/performance), Sentry will detect common performance problems, and group them into issues just like it does with errors.
```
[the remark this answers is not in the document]
gggritso: I'm trying to disambiguate them on purpose: a performance problem is something your code has, a performance issue is something we create from many events that have performance problems. Confusing?

src/docs/product/issues/performance-issues/index.mdx
gggritso: Yep, they are, why?

src/docs/product/issues/performance-issues/n-one-queries.mdx
gggritso: Noted, thanks!

src/docs/product/issues/performance-issues/index.mdx
```diff
@@ -0,0 +1,43 @@
+---
+title: "Performance Issues"
+sidebar_order: 15
+description: "Learn more about Performance Issues and how to use them."
+---
+
+<Include name="early-adopter-note.mdx" />
+
+The **Issue Details** page for performance issues helps you to gain further insight into the source of the issue and the impact it has on your application's users. The main area of the page displays information about a specific transaction event that's part of an issue. The top panel and right-hand sidebar of the page display a summary of all the events grouped together in this issue.
```
imatwawana, line 9:
```suggestion
The **Issue Details** page for performance issues helps you to gain further insight into the source of an issue and the impact it has on your application's users. The main area of the page displays information about a specific transaction event that's part of an issue. The top panel and right-hand sidebar of the page display a summary of all the events grouped together in this issue:
```

src/docs/product/issues/performance-issues/index.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+<Note>
+
+You can find performance issues by searching `issue.category:performance` in **Issues**, **Discover** and **Dashboards**.
+
+</Note>
+
```
imatwawana, line 16: I've moved the note to below the image so that the image follows the description of the image.
```suggestion
```

src/docs/product/issues/performance-issues/index.mdx
```diff
@@ -0,0 +1,43 @@
...
+![Performance issue](performance-issue-details.png)
+
```
imatwawana, line 18:
```suggestion

<Note>

You can find performance issues by searching `issue.category:performance` in **Issues**, **Discover**, and **Dashboards**.

</Note>

```

src/docs/product/issues/performance-issues/index.mdx
```diff
@@ -0,0 +1,43 @@
...
+While the **Issue Details** page displays information about a specific transaction event that's part of an issue, you can navigate between the events of an issue using the "Older" and "Newer" buttons. You'll see information about the issue as well as the details of the most recent event. You'll be able to see the summary of when the issue was seen, a breakdown of the tags, and breadcrumbs if they are available.
```
imatwawana, line 19:
```suggestion
While the **Issue Details** page displays information about a specific transaction event that's part of an issue, you can navigate between the events of an issue using the "Older" and "Newer" buttons. You'll see information about the issue as well as the details of the most recent event. You'll be able to see the summary of when the issue was seen, a breakdown of the tags, span evidence, and breadcrumbs if they are available.
```

src/docs/product/issues/performance-issues/index.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+## How Performance Issues Work
+
+Sentry detects performance issues by scanning incoming transaction events. It checks various properties of the transaction (span durations, span arrangement, span types, etc.) to detect likely problems. Sentry then checks these properties against corresponding thresholds to determine whether a performance issue exists. Sentry then generates a unique problem fingerprint based on the problem type and location in the application, and uses that fingerprint to create a performance issue.
```
imatwawana, line 23: to avoid repetition and calling out Sentry in the third person, switched to "we" after the first sentence

```suggestion
Sentry detects performance issues by scanning incoming transaction events. First we check various properties of the transaction (span durations, span arrangement, span types, and so on) to detect likely problems. We then check these properties against corresponding thresholds to determine whether a performance issue exists. Lastly, we generate a unique problem fingerprint based on the problem type and location in the application, and use that fingerprint to create a performance issue.
```

src/docs/product/issues/performance-issues/index.mdx
imatwawana: I'm fine with using the definition from the SDK docs if you want to @gggritso

src/docs/product/issues/performance-issues/index.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+## Tags
+
+Tags are key/value string pairs that are both indexed and searchable. For example, a tag provides you with information such as the browser, device, or user associated with the event. Tags are the diagnostic information sent by the SDK for the individual event.
+
+The tags displayed in the main section of this page are specific to the event that you're viewing. In contrast, the tags displayed in the right-hand sidebar are a summary of all tag values for all events included in the issue. You can set your own tags to make them more useful for debugging as described in Customize Tags.
+
+## Span Evidence
+
+Span evidence is information that explains the nature of the performance problem, and how it manifests in a transaction event. This section lists the spans involved in the issue and their descriptions, and shows a span tree that visually explains the problem's impact on the transaction. Different performance problems have slightly different kinds of evidence.
```
imatwawana, line 33: Wonder if some examples might be helpful to explain this section a bit more. They don't have to be long, even just a parenthetical (for example, x). This just feels very conceptual in a way that I find hard to hang on to.

src/docs/product/issues/performance-issues/index.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+## Performance Issue Limitations
+
+Performance issues have a few limitations:
```
imatwawana, line 37:
```suggestion
Performance issues currently have the following limitations:
```

src/docs/product/issues/performance-issues/index.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+- Sentry only detects one type of issue: [N+1 Queries](n-one-queries)
+- You cannot create alerts for performance issues
+- Performance issues cannot be merged or deleted
+- Custom fingerprinting and grouping do not apply to performance issues
+- Performance issues only support simple ignore rules
```
imatwawana, line 43:
```suggestion
- Performance issues only support simple ignore rules

We're working on removing some or all of these limitations.
```

src/docs/product/issues/performance-issues/n-one-queries.mdx
```diff
@@ -0,0 +1,54 @@
+---
+title: "N+1 Queries"
+sidebar_order: 10
+description: "Learn more about N+1 queries and how to diagnose and fix them."
+---
+
+<Include name="early-adopter-note.mdx" />
+
+## N+1 Queries
```
imatwawana, line 9: Don't need to repeat the title in a heading
```suggestion
```

src/docs/product/issues/performance-issues/index.mdx
DominikB2014: Suggesting that we document they are temporary limitations

src/docs/product/issues/performance-issues/n-one-queries.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+"N+1 Queries" is a performance problem where the application makes database queries in a loop, instead of making a single query that returns all the information at once. Each database connection has some amount of performance overhead, so querying the database in a loop can be many times slower than doing it just once. This problem often occurs when using an Object-relational mapping (ORM) tool in web frameworks like Django or Ruby on Rails.
```
imatwawana, line 11: Trying to avoid the plural/singular issue in the first sentence by making query singular
Attempting to simplify the language in the second sentence - hopefully I haven't changed the meaning. If I have then ignore that change.

```suggestion
An _N+1 query_ is a performance problem in which the application makes database queries in a loop, instead of making a single query that returns all the information at once. Each database connection takes some amount of time, so querying the database in a loop can be many times slower than doing it just once. This problem often occurs when you use an object-relational mapping (ORM) tool in web frameworks like Django or Ruby on Rails.
```

src/docs/product/issues/performance-issues/n-one-queries.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+## Span Evidence
+
+The evidence for a N+1 queries has four main aspects:
```
imatwawana, line 15: Are you talking about in general or in the context of the performance issue detail? If the latter, then change this to "N+1 query issue"

```suggestion
The evidence for an N+1 query has four main aspects:
```

src/docs/product/issues/performance-issues/n-one-queries.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+- the transaction name
+- the parent span. This can be a view, a serializer, or another span that logically groups the queries
+- the source span. This is the query that caused the N+1 queries
+- the repeating span. This is the "N" of "N+1 queries". This is the looped query that should have been part of the source
```
imatwawana, line 20:
```suggestion
- Transaction name
- Parent span -  This can be a view, a serializer, or another span that logically groups the queries.
- Source span - This is the query that caused the N+1 queries.
- Repeating span - This is the "N" of N+1 queries. This is the looped query that should have been part of the source.
```

src/docs/product/issues/performance-issues/n-one-queries.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+![N+1 Query span evidence](n-plus-one-span-evidence.png)
+
+## Example
+
+Consider a book review website. It has two ORM models: `Book` and `Author`, each with a corresponding database table. The website shows a list of ten oldest books and their respective authors. The code might look like this:
```
imatwawana, line 26:
```suggestion
Consider a book review website. It has two ORM models, `Book` and `Author`, each with a corresponding database table. The website shows a list of the ten oldest books and their respective authors. The code might look like this:
```

src/docs/product/issues/performance-issues/n-one-queries.mdx
```diff
@@ -0,0 +1,54 @@
...
+```python
+from django.http import HttpResponse
+
+def books(request):
+    books = Book.objects.all()[:10]
+    book_list = [book.title + " by " + book.author.name for book in books]
+    return HttpResponse((", ").join(book_list))
+```
+
+This code has a subtle performance problem. Each call to `book.author.name` makes a query to fetch the book's author! In total, this code makes 11 queries: one query to fetch the list of books, and 10 more queries to fetch the author of each book. This results in a characteristic query span waterfall:
```
imatwawana, line 37: No exclamation points in docs ;-)

```suggestion
This code has a subtle performance problem. Each call to `book.author.name` makes a query to fetch the book's author. In total, this code makes 11 queries: one query to fetch the list of books, and 10 more queries to fetch the author of each book. This results in a characteristic query span waterfall:
```

src/docs/product/issues/performance-issues/n-one-queries.mdx
```diff
@@ -0,0 +1,54 @@
...
+
+![N+1 queries in an example application](n-plus-one-queries-before.png)
+
+In order to fix this issue, you could use the `selected_related` method in Django, like so:
```
imatwawana, line 41:
```suggestion
In order to fix this performance issue, you could use the `selected_related` method in Django, like so:
```

imatwawana (commented):
I'd like to take one more look after you've made any updates just to make sure I haven't missed anything, but overall this is in a really good place.

src/docs/product/issues/performance-issues/index.mdx
gggritso: Clarification added!

src/docs/product/issues/performance-issues/index.mdx
gggritso: Hm, definitely maybe! I like the definition in the SDK docs, but the definition here is better at explaining the significance of tags in this specific context 🤔

src/docs/product/issues/performance-issues/n-one-queries.mdx
gggritso: Applied this change, _except_ I kept it as "N+1 Queries are a performance problem" for consistency

src/docs/product/issues/performance-issues/n-one-queries.mdx
gggritso: Changed to "The evidence for an N+1 queries problem has..." for consistency!

src/docs/product/issues/performance-issues/index.mdx
gggritso: Took a shot at re-phrasing this, with some specific examples

src/docs/product/issues/performance-issues/index.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+![Performance issue](performance-issue-details.png)
+
+<Note>
+
+You can find performance issues by searching `issue.category:performance` in **Issues**, **Discover**, and **Dashboards**.
+
+</Note>
+
+While the **Issue Details** page displays information about a specific transaction event that's part of an issue, you can navigate between the events of an issue using the "Older" and "Newer" buttons. You'll see information about the issue as well as the details of the most recent event. You'll be able to see the summary of when the issue was seen, a breakdown of the tags, span evidence, and breadcrumbs if they are available.
```
imatwawana, line 19:
```suggestion
While the **Issue Details** page displays information about a specific transaction event that's part of an issue, you can navigate between the events of an issue using the "Older" and "Newer" buttons. You'll see information about the issue as well as the details of the most recent event. You'll also be able to see the summary of when the issue was seen, a breakdown of the tags, span evidence, and breadcrumbs if they are available.
```

src/docs/product/issues/performance-issues/index.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+## Tags
+
+Tags are key/value string pairs that are both indexed and searchable. For example, a tag provides you with information such as the browser, device, or user associated with the event. Tags are the diagnostic information sent by the SDK for the individual event.
+
+The tags displayed in the main section of this page are specific to the event that you're viewing. In contrast, the tags displayed in the right-hand sidebar are a summary of all tag values for all events included in the issue. You can set your own tags to make them more useful for debugging as described in Customize Tags.
+
+## Span Evidence
+
+Span evidence is information that explains the performance problem in the context of the current event. It may include details like the specific database queries that cause the problem, the spans that contain the problem, and the slow spans that have an impact on performance. This section also has a span tree which shows how the offending spans relate to the rest of the event. Different performance problems will have slightly different kinds of evidence.
```
imatwawana, line 33: This is better, much clearer. The word "offending" is maybe a little too fancy here. 

```suggestion
Span evidence is information that explains the performance problem in the context of the current event. It may include details like the specific database queries that cause the problem, the spans that contain the problem, and the slow spans that have an impact on performance. This section also has a span tree which shows how the spans with the longest durations relate to the rest of the event. Different performance problems will have slightly different kinds of evidence.
```

imatwawana (approved):
A couple small things, but otherwise LGTM!

src/docs/product/issues/performance-issues/index.mdx
gggritso: How do you feel about "problematic" or some other synonym? "with the longest duration" is probably too specific. Sometimes they're fast, there's just a _lot_ of them, etc.

imatwawana: Was trying to avoid the word "problematic" but we've used in the docs before, so go for it. I can't think of a better word.

DominikB2014 (approved):

PR #5524, opened 2022-09-15
feat(native): clarify filtering hook behavior in the context of fast-fail crashes...

supervacuus:
... when using crashpad on Windows together with the WER (Windows Error Reporting) module.

### DIFF
--- a/src/platform-includes/configuration/before-send/native.mdx
+++ b/src/platform-includes/configuration/before-send/native.mdx
@@ -76,7 +76,7 @@ The Crashpad backend on macOS doesn't currently support notifying the crashing p
 </Alert>
 
 <Alert level="warning" title="Bypassed in Crashpad on Windows when capturing fast-fail crashes">
-  
-The Crashpad backend on Windows supports fast-fail crashes, which bypass SEH (Structured Exception Handling) primarily for security reasons. `sentry-native` registers a WER (Windows Error Reporting) module and signals the `crashpad_handler` to send a minidump to sentry. But since this process bypasses SEH, the application local exception handler is no longer invoked, which also means that for these kinds of crashes, `before_send` and `on_crash` will not be invoked before sending the minidump and thus have no effect.
+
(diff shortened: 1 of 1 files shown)


supervacuus:
@Swatinem, as discussed, that limitation should also be in the docs.

Swatinem (approved):

src/platform-includes/configuration/before-send/native.mdx
```diff
@@ -74,3 +74,9 @@ gradual migration from existing `before_send` implementations:
 The Crashpad backend on macOS doesn't currently support notifying the crashing process and thus can't correctly terminate sessions or call the registered `before_send` or `on_crash` hooks. It will also lose any events queued for sending at the time of the crash.
 
 </Alert>
+
+<Alert level="warning" title="Bypassed in Crashpad on Windows when capturing fast-fail crashes">
```
imatwawana, line 78: Fixed title casing
```suggestion
<Alert level="warning" title="Bypassed in Crashpad on Windows When Capturing Fast-fail Crashes">
```

imatwawana:
@supervacuus, I was just reviewing the wording on this as it was merged, but I'm wondering, should the first word of the title of the note be "Bypassed" or "Bypasses"? It doesn't quite make sense to me as the past tense.

supervacuus:
@imatwawana, the filtering hooks are the context of the alert box. What I tried to express: they (the filtering hooks) are bypassed (ugh, passive voice) when capturing fast-fail crashes. I wanted to continue in the style of the alert box above, which says: "Not Supported in Crashpad on macOS," meaning: "Those hooks are not supported in[...]", whereas in my case, the long form should be: "Those hooks are bypassed in." I can create a new PR that corrects my failed attempt at consistency :-) and also includes the correct title caps.

imatwawana:
Ah I understand. The format works well in the first note, but it's less obvious what's meant when we try to use that format in the second note. Can you go ahead and open another PR so we can fix the title casing and hammer out a better way of phrasing that title?

PR #5527, opened 2022-09-16
feat(react-native): Add Attachments

krystofwoldrich:
- https://github.com/getsentry/sentry-react-native/pull/2463


krystofwoldrich:
@marandaneto Besides the import the rest it the same as in js.

marandaneto:
@krystofwoldrich awesome, so we can release both together.

marandaneto (approved):

PR #5536, opened 2022-09-19
update platformicons to 5.2.3

robinrendle:
Bumping platformicons to 5.2.3 – adds wasm icons to the docs.

CleanShot 2022-09-19 at 16 04 58@2x


doralchan (approved):
Smash

PR #5539, opened 2022-09-21
fix(docs): Typo missing third underscore around DSN

Zylphrex:
Unsure what the distinction between `___DSN___` and `___PUBLIC_DSN___` is but all `__DSN__` (2 underscores) should all be `___DSN___` (3 underscores).


imatwawana (approved):
Thanks for the fix!

PR #5554, opened 2022-09-23
fix(apple): Explain why no duration for app hangs

philipphofmann:
Explain why the Cocoa SDK can't report the exact duration of app hangs.


brustolin:
TBH we kind could measure duration for most of the app hangs. And for those that cause the app to be kill we could report in the next session.

PR #5556, opened 2022-09-23
fix(native): Cut before_breadcrumb + add on_crash

supervacuus:
There is no support for the `before_breadcrumb`-hook in `sentry-native`, but there is an `on_crash`-hook which hasn't been mentioned yet in the __Basic Options__.


supervacuus:
Please, @Swatinem, have a look. Thx!

Swatinem (approved):

PR #5557, opened 2022-09-23
fix(native): set debug option in introductory examples

supervacuus:
As discussed with @Swatinem and raised in https://github.com/getsentry/sentry-native/issues/754, we should enable the debug option in all introductory examples because many new users are unaware of that option. I left out examples that discuss only one specific option.

### DIFF
--- a/src/wizard/native/index.md
+++ b/src/wizard/native/index.md
@@ -17,6 +17,7 @@ Import and initialize the Sentry SDK early in your application setup:
 int main(void) {
   sentry_options_t *options = sentry_options_new();
   sentry_options_set_dsn(options, "___PUBLIC_DSN___");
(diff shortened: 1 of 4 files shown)


src/wizard/native/index.md
```diff
@@ -17,6 +17,7 @@ Import and initialize the Sentry SDK early in your application setup:
 int main(void) {
   sentry_options_t *options = sentry_options_new();
   sentry_options_set_dsn(options, "___PUBLIC_DSN___");
+  sentry_options_set_debug(options, 1);
```
Swatinem, line 20: I wonder if the wizard should also contain a `release`?

Swatinem (approved):

Swatinem (approved):

PR #5560, opened 2022-09-25
remove weird usages of non breaking space <0xa0>

SimonCropp:
I assume the random use of non-breaking space `0xa0` is not intentional? they render as `&nbsp;` and will prevent sensible wrapping of text

image

they show up if u use a better text editor 

image


SimonCropp:
i would suggest logging an error when `0xa0` is found in mdx. Since it is too difficult to see the intent when most editors render them as normal spaces. and if people really want a non breaking space `&nbsp;` can be explicitly used

Swatinem (approved):
rubberstamping. not sure why native team was tagged for review?

imatwawana:
Thanks very much @SimonCropp!

PR #5583, opened 2022-09-29
fix(apple): Custom error desc min version

philipphofmann:
Specify which min version is needed for customized error descriptions.


imatwawana (approved):
LGTM!

PR #5590, opened 2022-10-04
Clarify python environment docs

antonpirker:
This confused one of our users and he is right about it.


imatwawana:
@antonpirker - you may want to try pushing an empty commit to this to get the checks running. They seemed to be not working very well yesterday, but seem to be running well today.

PR #5592, opened 2022-10-04
Add correct value for max_request_body_size option on PHP SDK

tkorakas:
max-request-body-size does not accept never as an argument on PHP SDK. 

`"max_request_body_size" with value "never" is invalid. Accepted values are: "none", "small", "medium", "always"`

Maybe, the other libraries accept never instead of none but I think it should be mentioned that for PHP the correct argument is `none` and not `never`

### Legal Boilerplate

### DIFF
--- a/src/platforms/common/configuration/options.mdx
+++ b/src/platforms/common/configuration/options.mdx
@@ -314,7 +314,7 @@ This parameter controls if integrations should capture HTTP request bodies. It c
 
 This parameter controls whether integrations should capture HTTP request bodies. It can be set to one of the following values:
 
-- `never`: Request bodies are never sent.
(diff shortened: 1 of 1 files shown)


cleptric:
@tkorakas Thanks for pointing this out 🙂
Instead of this change in the docs, I opened a PR to actually add the missing option to the PHP SDK. https://github.com/getsentry/sentry-php/pull/1397

tkorakas:
> @tkorakas Thanks for pointing this out 🙂 Instead of this change in the docs, I opened a PR to actually add the missing option to the PHP SDK. [getsentry/sentry-php#1397](https://github.com/getsentry/sentry-php/pull/1397)

Thanks a lot. Should I close this one?

cleptric:
Sure, thanks again!

PR #5593, opened 2022-10-04
ref: Clarify JS tracingOrigins default values

Lms24:
This PR clarifies the meaning of the JS SDK's default values for the `tracingOrigins` option of the `BrowserTracing` integration. As pointed out in https://github.com/getsentry/sentry-javascript/issues/5868, it is not obvious on first glance, to which requests tracing headers are attached if the default values are used. This PR therefore adds an explanation and slightly changes the ordering of the `tracingOrigins` description.

### DIFF
--- a/src/platforms/javascript/common/performance/instrumentation/automatic-instrumentation.mdx
+++ b/src/platforms/javascript/common/performance/instrumentation/automatic-instrumentation.mdx
@@ -52,7 +52,9 @@ Supported options:
 
 ### tracingOrigins
 
-The default value of `tracingOrigins` is `['localhost', /^\//]`. The JavaScript SDK will attach the `sentry-trace` and `baggage` headers to all outgoing XHR/fetch requests whose destination contains a string in the list or matches a regex in the list. If your frontend is making requests to a different domain, you will need to add it there to propagate the `sentry-trace` and `baggage` headers to the backend services, which is required to link transactions together as part of a single trace. **The `tracingOrigins` option matches against the whole request URL, not just the domain. Using stricter regex to match certain parts of the URL ensures that requests do not unnecessarily have the additional headers attached.**
+A list of strings and regular expressions. The JavaScript SDK will attach the `sentry-trace` and `baggage` headers to all outgoing XHR/fetch requests whose destination contains a string in the list or matches a regex in the list. If your frontend is making requests to a different domain, you will need to add it there to propagate the `sentry-trace` and `baggage` headers to the backend services, which is required to link transactions together as part of a single trace. **The `tracingOrigins` option matches against the whole request URL, not just the domain. Using stricter regex to match certain parts of the URL ensures that requests do not unnecessarily have the additional headers attached.**
+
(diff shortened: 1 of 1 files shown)


src/platforms/javascript/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -52,7 +52,9 @@ Supported options:
...
+A list of strings and regular expressions. The JavaScript SDK will attach the `sentry-trace` and `baggage` headers to all outgoing XHR/fetch requests whose destination contains a string in the list or matches a regex in the list. If your frontend is making requests to a different domain, you will need to add it there to propagate the `sentry-trace` and `baggage` headers to the backend services, which is required to link transactions together as part of a single trace. **The `tracingOrigins` option matches against the whole request URL, not just the domain. Using stricter regex to match certain parts of the URL ensures that requests do not unnecessarily have the additional headers attached.**
```
imatwawana, line 55:
```suggestion
A list of strings and regular expressions. The JavaScript SDK will attach the `sentry-trace` and `baggage` headers to all outgoing XHR/fetch requests whose destination contains a string in the list or matches a regex in the list. If your frontend is making requests to a different domain, you'll need to add it there to propagate the `sentry-trace` and `baggage` headers to the backend services, which is required to link transactions together as part of a single trace. 

**The `tracingOrigins` option matches against the whole request URL, not just the domain. Using stricter regex to match certain parts of the URL ensures that requests do not unnecessarily have the additional headers attached.**
```

src/platforms/javascript/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -52,7 +52,9 @@ Supported options:
...
+
+The default value of `tracingOrigins` is `['localhost', /^\//]`. This means that by default, tracing headers are only attached to requests that contain `localhost` in their URL or to requests whose URL starts with a `'/'` (for example `GET /api/v1/users`).
```
imatwawana, line 57:
```suggestion
The default value of `tracingOrigins` is `['localhost', /^\//]`. This means that by default, tracing headers are only attached to requests that contain `localhost` in their URL or requests whose URL starts with a `'/'` (for example `GET /api/v1/users`).
```

src/platforms/javascript/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -100,19 +102,19 @@ This option flags transactions when tabs are moved to the background with "cance
 
 The default is `true`.
 
+### \_experiments
 
-### _experiments
-
-This is an object containing experimental flags for features that may involve breaking changes that haven't yet landed in a new major version. 
+This is an object containing experimental flags for features that may involve breaking changes that haven't yet landed in a new major version.
```
imatwawana, line 107:
```suggestion
This is an object containing experimental flags for features that haven't yet landed in a new major version and might include breaking changes.
```

imatwawana (approved):
Just a few wording edits, but otherwise LGTM!

PR #5610, opened 2022-10-06
feat(discover): Add interval option docs

wmak:
- This adds interval option as a section in the query builder

### DIFF
--- a/src/docs/product/discover-queries/query-builder.mdx
+++ b/src/docs/product/discover-queries/query-builder.mdx
@@ -68,7 +68,7 @@ Each query has a interactive graph that reflects the data presented in the table
 
 ### Interval Options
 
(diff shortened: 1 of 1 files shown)


src/docs/product/discover-queries/query-builder.mdx
```diff
@@ -66,6 +66,10 @@ Click on any of these sections to further refine your search. For example, click
 
 Each query has a interactive graph that reflects the data presented in the table below. To zoom in, simply click and drag over the area you want to investigate. You can also customize the display and the Y-Axis. This will be saved with your query.
 
+### Interval Options
+
+The Interval selector will allow you to change the date groupings in the graph. The interval selector has bounds that are based on the currently selected time range, this means for example if you're doing a query over 90 days the largest interval you can select is 45 days and the smallest you can pick is 1 hour.
```
imatwawana, line 71:
```suggestion
The "Interval" selector allows you to change the date groupings in the graph. The selector has limits that are based on the currently selected time range. This means, for example, if you're querying over 90 days, the largest interval you can select is 45 days and the smallest is one hour.
```

imatwawana (approved):
Just some wording tweaks. Otherwise looks good!

PR #5619, opened 2022-10-07
ref: Improve JS SDK Component Tracking Documentation

Lms24:
This PR makes structural, naming and content changes to the documentation of a feature we support in four JS framework SDKs - component tracking. Previously, this documentation differed drastically between these four SDKs, as described in greater detail in #5582. Not only did the location differ, but also, how we called the feature. With this PR, we're introducing  one name that describes this feature, while we still find a balance between "component tracking" and however the feature is called in the SDK's code (e.g. "trace helpers" in Angular or "Profiler" in React).

Impacted SDKs: Angular, React, Vue, Svelte

As of this PR, we make the following changes:

**All SDKs**
* Adjusted the naming and location of the documentation. In the menu on the left, it is now located under "\<Framework> Features" -> "Component Tracking"
* Added a sentence in all "\<framwork> Features" index pages explaining that these features are specifically built for the respective framework. For example, in React we don't only have component tracking as such a feature but also the Sentry error boundary. 
* Added a small paragraph on top of the component tracking page, explaining the feature and why it's helpful
* Mentioned the names of the spans that this feature adds to the transactions so that users are aware of what it is actually doing. The React docs were already doing this and I thought this would be valuable information for every SDK.

**Angular**:
* Rephrased some paragraphs and sentences that were previously confusing
* Added sub-headings for each utility (previously it was a list)
* Added an "Advanced" usage section, explaining that the utilities can be used together
* Adjusted the code examples and corrected some minor mistakes

**Vue**:

The Vue SDK documentation didn't have a proper page for component tracking previously. Instead it was spread across three pages. 

* Moved component tracking documentation from the "Vue Router Integration" to the new component tracking page
* Adjusted this documentation by
  * Removing unnecessary technical details
  * Better explaining each of the three options
  * Removing lifecycle hooks that weren't even supported by the SDK
* Since component tracking was mentioned in two other pages (Getting started and Performance Setup), I decided to shorten the description in the respective places to only leave the gist there. In both places, I added links to the component tracking page for more details.
* Left the part that is actually about the Vue router on the "Vue Router Integration" page.
  * I'm not too happy with this, because technically speaking we don't have a Vue router integration but an _instrumentation_ but we do the same thing in the React SDK. In both cases, the pages actually contain relevant information. So we can leave it for now as is and revisit this if it comes up.

fixes #5582

Issue #5582, opened 2022-09-29
Improve JS SDK Component Tracking Documentation


Lms24:
### Core or SDK?

Platform/SDK

### Which part? Which one?

JS Framework SDKs (Angular, React, Vue, Svelte)

### Description

This issue was extracted from https://github.com/getsentry/sentry-docs/pull/5470

Currently, our documentation on component tracking differs between SDKs that support the feature:

* In the [Angular docs](https://docs.sentry.io/platforms/javascript/guides/angular/components/tracehelpers/), the content lives under "Components"->"Tracking Angular components". I like the placement but the items that we expose in the SDK for component tracking are not components. Also, we call these items "tracing helpers" which IMO is an overloaded term and doesn't describe component tracking well. It might lead to people thinking that component tracking has something to do with distributed tracing.
* The [React docs](https://docs.sentry.io/platforms/javascript/guides/react/) place the component tracking content under "Components"->"Profiler". Here, the term "components" actually makes sense because the profiler is a component. However, we call the same feature here entirely differently ("Profiling" vs "Component Tracking")
* The Vue docs have no "Components" page at all. The component tracking content is mostly a few paragraphs in two places:
  * [Performance setup](https://docs.sentry.io/platforms/javascript/guides/react/): Here we talk about "configuring [the] tracing instrumentation". Again IMO an overloaded term that doesn't fit for component tracking very well.
  * [Vue router integration](https://docs.sentry.io/platforms/javascript/guides/vue/configuration/integrations/vue-router/): Component tracking again makes a somewhat disguised appearance. I find this page a little weird in gerneral. We don't have a Vue routing integration. We describe the `BrowserTracing` integration but for some reason, the code examples don't contain the Vue routing instrumentation (which would be in fact vue-specific but not an integration). In the "Getting Started" page, we show the correct configuration (e.g. for Vue 3). Perhaps this page needs some rework in general (might be a leftover from before we had a proper Vue SDK?)
  * The `trackComponents` option is also mentioned in the "[Getting Started](https://docs.sentry.io/platforms/javascript/guides/vue/)" page. I think that's generally fine but from what I understand, "Getting started" should show how to get going and the other sections should show the advanced stuff, no?
[...]

PR #5625, opened 2022-10-10
feat(components): Add `additionalPages` prop to `PageGrid` component

Lms24:
## ⚠️ Only merge this after #5619 was merged!

(This might cause build errors until #5619 is merged. Opening as a draft for now)

This PR adds a new optional prop to the `PageGrid` component (`additionalPages`) that allows us to pass a list of additional page links that should be added to the list of pages generated by the component. This is useful when we want to link to a page that is not a sub-page of the MDX file where `PageGrid` was placed. 

Furthermore, this PR adds an index page to JS SDK's performance instrumentation section that makes use of the new prop. 

ref: #5582

### DIFF
--- a/src/components/pageGrid.tsx
+++ b/src/components/pageGrid.tsx
@@ -20,12 +20,25 @@ const query = graphql`
   }
 `;
 
+type PageLink = {
+  title: string;
+  path: string;
+  description?: string;
+};
+
 type Props = {
   nextPages: boolean;
   header?: string;
+
+  /** Adds list items of page links that are not automatically found by the PageGrid component */
+  additionalPages?: PageLink[];
 };
 
-export default ({ nextPages = false, header }: Props): JSX.Element => {
+export default ({
+  nextPages = false,
+  header,
+  additionalPages,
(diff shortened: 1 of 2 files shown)


src/components/pageGrid.tsx
```diff
@@ -53,11 +66,21 @@ export default ({ nextPages = false, header }: Props): JSX.Element => {
 
   if (!matches.length) return null;
 
+  const addedPages =
+    (additionalPages &&
```
AbhiPrasad, line 70: if we default `additionalPages` to be an empty array we don't need this check.

Lms24: Good idea, added in [6f1dc8d](https://github.com/getsentry/sentry-docs/pull/5625/commits/6f1dc8d6460e713d502f683b097bf7e22b4faa55)

Lms24:
That's a good point @imatwawana. I think the potential danger of people abusing this additional prop to link to ...wherever is greater than the advantage it offers for this particular case. Closing this in favour of consistency.

PR #5630, opened 2022-10-11
ref(android): Use platform link for perf metrics

philipphofmann:
Use platform links.


adinauer (approved):

markushi:
Oh I guess I missed that, thanks for fixing this!

PR #5634, opened 2022-10-12
Add information about missing java stack trace on rn new arch

krystofwoldrich:
Inform users of the new arch that we are aware of the issues with the jave stack trace.


krystofwoldrich:
@marandaneto Something like this or a bit more in-depth?

marandaneto:
> @marandaneto Something like this or a bit more in-depth?

it's alright for now.

lizokm (approved):
Hi there! This looks good to me.

FYI @imatwawana

PR #5652, opened 2022-10-17
Add SENTRY_ALLOW_FAILURE into react-native troubleshooting page

krystofwoldrich:
Closes https://github.com/getsentry/sentry-react-native/issues/2246

### DIFF
--- a/src/platforms/react-native/troubleshooting.mdx
+++ b/src/platforms/react-native/troubleshooting.mdx
@@ -245,7 +245,7 @@ If you're using the new architecture on React Native, you might be missing the J
 
 ## Failing Build Due to `sentry-cli`
 
-If your build fails due to Sentry CLI error, because you don't have internet connection or [sentry.io](https://sentry.io) is down, you can allow failure of the `sentry-cli` step by adding an environmental variable `SENTRY_ALLOW_FAILURE` equal to `true`. The CLI will still run, but won't cause the build to fail:
+If your build fails due to Sentry CLI error, because you don't have internet connection, or because [sentry.io](https://sentry.io) is down, you can allow failure of the `sentry-cli` step by adding an environmental variable `SENTRY_ALLOW_FAILURE` equal to `true`. The CLI will still run, but won't cause the build to fail:
 
 ```bash
 # iOS
@@ -255,6 +255,6 @@ SENTRY_ALLOW_FAILURE=true npx react-native run-ios
 SENTRY_ALLOW_FAILURE=true npx react-native run-android
 ```
 
-Note that if source-maps upload failed it will have to be done [manually](/platforms/react-native/sourcemaps) later when the connection is restored.
+Note that if the source maps upload fails, you'll have to do it [manually](/platforms/react-native/sourcemaps) later when the connection is restored.
 
-When you remove the environmental variable in front of your build command your build will work as before.
+When you remove the environmental variable in front of your build command, your build will work as before.


src/platforms/react-native/troubleshooting.mdx
```diff
@@ -242,3 +242,15 @@ React Native 0.60.3 tooling was unexpectedly moving source maps files, which cau
 ## Missing Java stacktrace on the New Architecture
 
 If you're using the new architecture on React Native, you might be missing the Java stacktrace. Follow the [issue](https://github.com/facebook/react-native/issues/34923) for more information.
+
+## Failing build due to `sentry-cli`
```
imatwawana, line 246:
```suggestion
## Failing Build Due to `sentry-cli`
```

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -242,3 +242,15 @@ React Native 0.60.3 tooling was unexpectedly moving source maps files, which cau
...
+
+In cases, you are building your app without access to `sentry.io` or some other unexpected error occurs. You can allow failure of the `sentry-cli` step by adding an environmental variable `SENTRY_ALLOW_FAILURE` equal to `true`. The CLI will still run, but will not fail the build.
```
imatwawana, line 248:
```suggestion
If you're building your app without access to `sentry.io` or some other unexpected error occurs, you can allow failure of the `sentry-cli` step by adding an environmental variable `SENTRY_ALLOW_FAILURE` equal to `true`. The CLI will still run, but won't cause the build to fail:
```

imatwawana (commented):
Made some style and wording edits

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -242,3 +242,15 @@ React Native 0.60.3 tooling was unexpectedly moving source maps files, which cau
 ## Missing Java stacktrace on the New Architecture
 
 If you're using the new architecture on React Native, you might be missing the Java stacktrace. Follow the [issue](https://github.com/facebook/react-native/issues/34923) for more information.
+
+## Failing Build Due to `sentry-cli`
+
+If you're building your app without access to `sentry.io` or some other unexpected error occurs, you can allow failure of the `sentry-cli` step by adding an environmental variable `SENTRY_ALLOW_FAILURE` equal to `true`. The CLI will still run, but won't cause the build to fail:
```
imatwawana, line 248: This may just be me, but "some other unexpected error" doesn't seem like a specific enough use case. An unexpected error occurring seems to be something that can happen in many situations. Can you add more context there?

imatwawana, line 248: Fixing formatting of sentry app reference

```suggestion
If you're building your app without access to [sentry.io](https://sentry.io) or some other unexpected error occurs, you can allow failure of the `sentry-cli` step by adding an environmental variable `SENTRY_ALLOW_FAILURE` equal to `true`. The CLI will still run, but won't cause the build to fail:
```

imatwawana (commented):
I left a question because I'm not clear what you mean in one part.

src/platforms/react-native/troubleshooting.mdx
krystofwoldrich: I just want to point out that the environment variable allows any failure of the cli not only connections to Sentry. It could be some issues with processing debug files. But maybe that’s obvious and I can leave that part out?

imatwawana: Ah okay. I think it's just the word CLI that's missing there. Maybe something like:

> If you're building your app without access to `sentry.io` or some other unexpected CLI error occurs, ...

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -242,3 +242,15 @@ React Native 0.60.3 tooling was unexpectedly moving source maps files, which cau
...
+## Failing Build Due to `sentry-cli`
+
+If you're building your app without access to [sentry.io](https://sentry.io) or some other unexpected error occurs, you can allow failure of the `sentry-cli` step by adding an environmental variable `SENTRY_ALLOW_FAILURE` equal to `true`. The CLI will still run, but won't cause the build to fail:
+
+```bash
+# iOS
+SENTRY_ALLOW_FAILURE=true npx react-native run-ios
+
+# Android
+SENTRY_ALLOW_FAILURE=true npx react-native run-android
```
marandaneto, line 255: I'd jump a line after exporting the env. var.

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -242,3 +242,15 @@ React Native 0.60.3 tooling was unexpectedly moving source maps files, which cau
 ## Missing Java stacktrace on the New Architecture
 
 If you're using the new architecture on React Native, you might be missing the Java stacktrace. Follow the [issue](https://github.com/facebook/react-native/issues/34923) for more information.
+
+## Failing Build Due to `sentry-cli`
+
+If you're building your app without access to [sentry.io](https://sentry.io) or some other unexpected error occurs, you can allow failure of the `sentry-cli` step by adding an environmental variable `SENTRY_ALLOW_FAILURE` equal to `true`. The CLI will still run, but won't cause the build to fail:
```
marandaneto, line 248: H: I'd give an example, such as internet issues or sentry.io is down.
I'd also explain that if this fails when eg uploading source maps, this should be done later manually once the connection is back.
I'd also explain how to remove the env. var. once everything is back to normal.

src/platforms/react-native/troubleshooting.mdx
krystofwoldrich: I created the example without exporting the ENV and instead ran the single build command with the ENV.

Like this, it's less likely that people will forget about it. Also, it's not persistent, so there won't be trouble with forgotten FAILURE=true.

marandaneto: mmm that makes a lot of sense, lets keep then.

src/platforms/react-native/troubleshooting.mdx
krystofwoldrich: Okay, I've adjusted the text.

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -242,3 +242,19 @@ React Native 0.60.3 tooling was unexpectedly moving source maps files, which cau
 ## Missing Java stacktrace on the New Architecture
 
 If you're using the new architecture on React Native, you might be missing the Java stacktrace. Follow the [issue](https://github.com/facebook/react-native/issues/34923) for more information.
+
+## Failing Build Due to `sentry-cli`
+
+If your build fails due to Sentry CLI error, because you don't have internet connection or [sentry.io](https://sentry.io) is down, you can allow failure of the `sentry-cli` step by adding an environmental variable `SENTRY_ALLOW_FAILURE` equal to `true`. The CLI will still run, but won't cause the build to fail:
```
imatwawana, line 248:
```suggestion
If your build fails due to Sentry CLI error, because you don't have internet connection, or because [sentry.io](https://sentry.io) is down, you can allow failure of the `sentry-cli` step by adding an environmental variable `SENTRY_ALLOW_FAILURE` equal to `true`. The CLI will still run, but won't cause the build to fail:
```

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -242,3 +242,19 @@ React Native 0.60.3 tooling was unexpectedly moving source maps files, which cau
...
+
+```bash
+# iOS
+SENTRY_ALLOW_FAILURE=true npx react-native run-ios
+
+# Android
+SENTRY_ALLOW_FAILURE=true npx react-native run-android
+```
+
+Note that if source-maps upload failed it will have to be done [manually](/platforms/react-native/sourcemaps) later when the connection is restored.
```
imatwawana, line 258:
```suggestion
Note that if the source maps upload fails, you'll have to do it [manually](/platforms/react-native/sourcemaps) later when the connection is restored.
```

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -242,3 +242,19 @@ React Native 0.60.3 tooling was unexpectedly moving source maps files, which cau
...
+
+When you remove the environmental variable in front of your build command your build will work as before.
```
imatwawana, line 260:
```suggestion
When you remove the environmental variable in front of your build command, your build will work as before.
```

imatwawana (approved):
Just made some wording tweaks, but otherwise LGTM!

PR #5655, opened 2022-10-17
Document that issue-link hook `actor.id` is a number

mrcljx:
I had a dictionary of `[String: User]` and was trying to look up additional user information (like email address), but for some reason that the ID was never found in my list.

Eventually I realized that, while the APIs return IDs as strings (of numbers), the field here seems to be a number directly. Code in question is at:

https://github.com/getsentry/sentry/blob/da7e0f5c3deffa6987059744a1d9df6cd9ea2443/src/sentry/mediators/external_requests/issue_link_requester.py#L116

I guess since the hook contract can't be modified easily, this means the documentation needs to be updated?

### Legal Boilerplate


mrcljx:
It looks like I might actually have tricked myself here. Closing until I reconfirm.

PR #5657, opened 2022-10-17
ref(tools): Fix volta on docs

k-fish:
### Summary
Checking if this works to fix volta on builds.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ jobs:
       JEKYLL_ENABLE_PLATFORM_API: false
     steps:
       - uses: actions/checkout@v2
-      - uses: volta-cli/action@v1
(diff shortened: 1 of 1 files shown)


gggritso (approved):
🔧

.github/workflows/test.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       JEKYLL_ENABLE_PLATFORM_API: false
     steps:
       - uses: actions/checkout@v2
-      - uses: volta-cli/action@v1
+      - uses: getsentry/action-setup-volta@d6f6ebfc4046feb3cfb6049e885185a01afd82b7  # v1.0.0
```
asottile-sentry, line 17: you'll probably want to bump to 1.1.0 which upgrades volta as well

asottile-sentry (commented):
lg2m (after the fact!)

PR #5663, opened 2022-10-18
Laravel 3.0.0

cleptric:
Updates the Laravel docs to be in line with our new major [v3.0.0](https://github.com/getsentry/sentry-laravel/releases/tag/3.0.0).
As Laravel 5.x is now deprecated, I moved it to its own page.

### DIFF
--- a/src/platforms/php/guides/laravel/other-versions/laravel5.mdx
+++ b/src/platforms/php/guides/laravel/other-versions/laravel5.mdx
@@ -23,7 +23,7 @@ Install the `sentry/sentry-laravel` package:
 composer require "sentry/sentry-laravel:^2.14"
 ```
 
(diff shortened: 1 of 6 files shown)


src/platforms/php/guides/laravel/other-versions/laravel5.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+## Install
+
+Install the `sentry/sentry-laravel` package:
+
+```bash
+composer require "sentry/sentry-laravel:^2.14"
+```
+
+If you're on Laravel 5.5 or higher, the package will be auto-discovered. Otherwise you will need to manually configure it in your `config/app.php`.
```
lizokm, line 26:
```suggestion
If you're on Laravel 5.5 or higher, the package will be auto-discovered. Otherwise, you will need to manually configure it in your `config/app.php`.
```

lizokm (commented):
Just added a comma, but otherwise GTM

PR #5668, opened 2022-10-19
fix: Change casing of javascript webhook property

rbisol:
Updating the casing in the Javascript Verifying signature example. 

from `Sentry-Hook-Signature` to `sentry-hook-signature`

The header is received in my test webhook in all lowercase `sentry-hook-signature`, and it is also used like that in the python example. One customer reported getting errors when copying the code due to the wrong case.


scttcper (approved):

scttcper:
thanks! @rbisol

PR #5675, opened 2022-10-21
Fix misleading description of `config.logger` default.

sengi:
This document is transcluded in the [docs for `sentry-rails`](https://docs.sentry.io/platforms/ruby/guides/rails/configuration/options/) as well as [those for `sentry-ruby`](https://docs.sentry.io/platforms/ruby/configuration/options/). This results in the statement about the default setting for `config.logger` being misleading in the Rails case, because the reader might reasonably assume that it's safe to mutate the default logger.

For example, based on these docs I might reasonably expect to be able to set `config.logger.level = Sentry::Logger::WARN` when initialising Sentry, but if I'm in a Rails app I'd be shooting myself in the foot by unwittingly mutating `Rails.logger` through a shared reference.

### Legal Boilerplate

### DIFF
--- a/src/platforms/ruby/common/configuration/options.mdx
+++ b/src/platforms/ruby/common/configuration/options.mdx
@@ -121,7 +121,7 @@ config.inspect_exception_causes_for_exclusion = true
 
 `logger`
 
-: The logger used by Sentry. Default is an instance of Sentry::Logger.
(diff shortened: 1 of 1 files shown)


sl0thentr0py (approved):
thx @sengi

PR #5676, opened 2022-10-21
fix(icons): updated large remix icon

robinrendle:
- Updated to the proper version of platform icons
- Fixes large Remix icon to make it more readable 

### Before 

CleanShot 2022-10-21 at 10 36 41@2x

### After
CleanShot 2022-10-21 at 10 41 07@2x


imatwawana (approved):
Looks great! Thank you!

PR #5692, opened 2022-10-26
fix extension for typescript with jsx file

chrisweb:
My smallest contribution ever but it makes a big difference ;)

In the docs for the nextjs manual setup (https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/) there is a tiny error which has a big impact, the extension for the "_error" file in a nextjs project when using typescript should be ".tsx" and not ".ts", so this PR adds the x to the "_error.ts" filename and fixes the compilation error one will get if using the .ts extension instead of the .tsx extension

### Legal Boilerplate

### DIFF
--- a/src/platforms/javascript/guides/nextjs/manual-setup.mdx
+++ b/src/platforms/javascript/guides/nextjs/manual-setup.mdx
@@ -93,7 +93,7 @@ CustomErrorComponent.getInitialProps = async contextData => {
 export default CustomErrorComponent;
 ```
 
-```typescript {filename:pages/_error.ts}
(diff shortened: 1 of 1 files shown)


lforst (approved):

lforst:
Thank you for your contribution!

chrisweb:
@lforst thank you for reviewing it 👍

lobsterkatie:
@chrisweb - I'm curious: What compilation error was it giving you without the `x`? In my test app it works fine with it being `_error.ts` (which is how the docs came to be that way).

PR #5695, opened 2022-10-27
Auto perf metrics for Flutter

marandaneto:
### Legal Boilerplate


imatwawana:
I don't see this content showing up in the preview; you need to update the supported list for the PlatformSection this include is in on the common Performance Metrics page.

marandaneto:
> I don't see this content showing up in the preview; you need to update the supported list for the PlatformSection this include is in on the common Performance Metrics page.

yep, indeed, good catch, thanks.

mattjohnsonpint (approved):

PR #5697, opened 2022-10-27
fix: Fix broken draft: frontmatter, make it apply to PageGrid too

untitaker:
Everybody seemed to use draft:false to hide pages, which reads like nonsense. Fix all queries wrt draft and fix a mismatch in expectations where drafts were still showing in PageGrid


imatwawana:
Thank you! Thank you! Thank you!

imatwawana (approved):

Issue #5647, opened 2020-12-09
Improve docs for `extractTraceparentData` function


abrahamq:
- [x] Review the documentation: https://docs.sentry.io/
- [x] Search for existing issues: https://github.com/getsentry/sentry-javascript/issues
have seen this [issue](https://github.com/getsentry/sentry-javascript/issues/2970) but it doesn't contain much more info.
- [x] Use the latest release: https://github.com/getsentry/sentry-javascript/releases
- [x] Provide a link to the affected event from your Sentry account - https://sentry.io/organizations/tierra/issues/2081900024/?project=1841969&query=is%3Aunresolved

## Package + Version

- [ ] `@sentry/browser`
- [x] `@sentry/node` - 5.29.0
- [ ] `raven-js`
- [ ] `raven-node` _(raven for node)_
- [x] other: `@sentry/tracing` - 5.29.0

### Version:

```
5.29.0
```

## Description
I'm trying to write a graphql-middleware to integrate with our app. Following the instructions [here](https://github.com/getsentry/sentry-javascript/issues/2860) I tried to modify the express middleware already in sentry for this use; however, when I try to call `Span.fromTraceparent` I receive the titular error. I've looked through the compiled javascript and that function is nowhere to be found even though it is present on the typedoc here: https://getsentry.github.io/sentry-javascript/classes/tracing.span-2.html#fromtraceparent 

Thanks!

PR #3693, opened 2021-06-05
Review / restructure source maps content.

onurtemizkan:
- [x] Move out and merge JS and Node source map content inside `includes`.
- [x] Separate 'Generate' and 'Upload' content for clarity.
- [x] Add source map generation example for Rollup. (Used: https://blog.sentry.io/2019/02/20/debug-node-source-maps)
- [x] Review/Update `RewriteFrames` content.
- [x] Add in-depth summaries and examples for CLI/sentry-webpack-plugin options.

There are another couple of things to improve our source maps content further (after we finalize this PR):
- #3831
- #3615

PR #5705, opened 2022-11-01
Add Distributed Tracing to Node docs

onurtemizkan:
Potentially Resolves: #5647 (Do we need to open another PR on [getsentry/develop](https://github.com/getsentry/develop)?)

Adds basic examples about:

- How to obtain a trace header from a span using `toTraceparent`
- How to use `extractTraceparentData` to continue upstream service traces.

Also removed the part that mentions `continueFromHeaders` function (which is not available in `sentry-javascript`) for JavaScript and Node docs. 

---

As an aside note, Node docs in general seem to need some restructuring, and I just remembered there is unfinished work in https://github.com/getsentry/sentry-docs/issues/3605.

We previously made a few updates in https://github.com/getsentry/sentry-docs/pull/3693, which have already crossed out some sub-tasks there.

I'm up to finish that next, if the team agrees.

### DIFF
--- a/src/platform-includes/performance/connect-services/node.mdx
+++ b/src/platform-includes/performance/connect-services/node.mdx
@@ -1,6 +1,6 @@
 ## Distributed Tracing
 
-To obtain a trace from a span, use `toTraceParent()` method. This method returns a string that can be used as a value of the `traceparent` HTTP header.
+To obtain a trace from a span, use `toTraceParent()` method. This method returns a string that can be used as a value of the `traceparent` HTTP header:
 
 ```javascript
 const sentryTraceHeader = span.toTraceparent();
--- a/src/platform-includes/performance/connect-services/node.mdx
+++ b/src/platform-includes/performance/connect-services/node.mdx
@@ -1,0 +1,27 @@
+## Distributed Tracing
+
+To obtain a trace from a span, use `toTraceParent()` method. This method returns a string that can be used as a value of the `traceparent` HTTP header.
+
+```javascript
+const sentryTraceHeader = span.toTraceparent();
+
+requestOptions.headers = {
+  ...requestOptions.headers,
+  'sentry-trace': sentryTraceHeader,
+};
+```
+
+To create a span as a continuation of the trace retrieved from the upstream service, use `extractTraceparentData` function from `@sentry/tracing`:
+
+```javascript
+import { extractTraceparentData } from "@sentry/tracing";
+
+// The request headers sent by your upstream service to your backend.
+const traceparentData = extractTraceparentData(request.headers['sentry-trace']);
+
+const transaction = Sentry.startTransaction({
+  op: 'transaction_op',
+  name: 'transaction_name',
+  ...traceparentData,
+});
+```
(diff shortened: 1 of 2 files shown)


src/platform-includes/performance/connect-services/node.mdx
```diff
@@ -0,0 +1,27 @@
+## Distributed Tracing
+
+To obtain a trace from a span, use `toTraceParent()` method. This method returns a string that can be used as a value of the `traceparent` HTTP header.
```
imatwawana, line 3:
```suggestion
To obtain a trace from a span, use `toTraceParent()` method. This method returns a string that can be used as a value of the `traceparent` HTTP header:
```

src/platforms/common/performance/connect-services.mdx
```diff
@@ -10,9 +10,13 @@ When tracing is enabled, the SDK will attach an outgoing header called `sentry-t
 
 SDKs with performance monitoring support listen to incoming requests and typically automatically pick up the incoming `sentry-trace` header to continue the trace (with the same `trace-id`) from there, connecting backend and frontend transactions into a single coherent trace using the `trace_id` value. Depending on the circumstance, this ID is transmitted either in a request header or in an HTML `<meta>` tag. All your transactions that have the same `trace-id` are connected. Linking transactions in this way makes it possible to navigate among them in [sentry.io](https://sentry.io) to better understand how the different parts of your system are affecting one another. You can learn more about this model in our [Distributed Tracing](/product/sentry-basics/tracing/distributed-tracing/) docs.
```
imatwawana, line 11: In the preview, this paragraph is really long. Does it make logical sense to break it up as I have below? If not, can you suggest a good break in that section?

```suggestion
SDKs with performance monitoring support listen to incoming requests and typically automatically pick up the incoming `sentry-trace` header to continue the trace (with the same `trace-id`) from there, connecting backend and frontend transactions into a single coherent trace using the `trace_id` value. Depending on the circumstance, this ID is transmitted either in a request header or in an HTML `<meta>` tag. 

All your transactions that have the same `trace-id` are connected. Linking transactions in this way makes it possible to navigate among them in [sentry.io](https://sentry.io) to better understand how the different parts of your system are affecting one another. You can learn more about this model in our [Distributed Tracing](/product/sentry-basics/tracing/distributed-tracing/) docs.
```

src/platforms/common/performance/connect-services.mdx
```diff
@@ -10,9 +10,13 @@ When tracing is enabled, the SDK will attach an outgoing header called `sentry-t
...
 
+<PlatformSection notSupported={["javascript", "node"]}>
+
 If the instrumentation you are using doesn't automatically pick up the `sentry-trace` header, you can also continue a trace manually by using the `continueFromHeaders` function on a `Transaction`, which you can learn more about in our content for [the Transaction Interface](https://develop.sentry.dev/sdk/performance/#new-span-and-transaction-classes).
```
imatwawana, line 15:
```suggestion
If the instrumentation you're using doesn't automatically pick up the `sentry-trace` header, you can also continue a trace manually by using the `continueFromHeaders` function on a `Transaction`. Learn more about this in our [Transaction Interface](https://develop.sentry.dev/sdk/performance/#new-span-and-transaction-classes) documentation.
```

imatwawana (approved):
Made some tweaks to the original content and one small edit in what you've added.

src/platforms/common/performance/connect-services.mdx
onurtemizkan: Yes, breaking like this makes sense to me.

AbhiPrasad (approved):

PR #5714, opened 2022-11-03
docs: Total transactions baseline docs

shruthilayaj:


src/docs/product/discover-queries/query-builder.mdx
```diff
@@ -66,6 +66,17 @@ Click on any of these sections to further refine your search. For example, click
...
+
+There are a few reasons why the 'Total Transactions' toggle might be disabled:
+- Selected display is not supported, only available on Total Period display.
+- Selected y-axis is not supported, not available for equations, `apdex()`, `user_misery()` and aggregations performed on fields and tags (for example `count_unique(device.os)`)
+- You are querying both error and transaction data (which is default Discover behaviour), ensure you are querying only transactions data by adding `event.type:transaction` to the query search bar
+- Your project has an outdated SDK which is incompatible with [Dynamic Sampling](/product/data-management-settings/dynamic-sampling/), confirm that your application relies on [supported SDKs](/product/data-management-settings/dynamic-sampling/getting-started/#current-limitations)
+- You have some unparameterized transactions which are incompatible with [Dynamic Sampling](/product/data-management-settings/dynamic-sampling/)
```
lizokm, line 78:
```suggestion
Discover queries use sampled transactions, or events-based data. You can compare your query results against the data, based on total transactions, by enabling the "Total Transactions" toggle in the chart footer. For example, if you were investigating performance issues in your latest release, you could display `p95(transaction.duration)` for your release and the total transactions baseline to visualize how much worse it was compared to your baseline.

There are a few reasons why the "Total Transactions" toggle might be disabled:
- Selected display is not supported and is only available on "Total Period" display.
- Selected y-axis is not supported, is not available for equations, `apdex()`, `user_misery()`, or aggregations performed on fields and tags (for example `count_unique(device.os)`).
- You're querying both error and transaction data (which is default Discover behavior). To fix this, query transactions data only by adding `event.type:transaction` to the query search bar.
- Your project has an outdated SDK which is incompatible with [Dynamic Sampling](/product/data-management-settings/dynamic-sampling/). To fix this, confirm that your application relies on [supported SDKs](/product/data-management-settings/dynamic-sampling/getting-started/#current-limitations).
- You have some unparameterized transactions which are incompatible with [Dynamic Sampling](/product/data-management-settings/dynamic-sampling/).
```

lizokm (approved):
@shruthilayaj Thanks so much for this! I've made a few edits. Just want to double check that in bullet points 3 & 4 you were suggesting solutions for the potential problems? I've added "to fix this" and just want to make sure it makes sense with what you were trying to say.

@imatwawana would love your eyes on this as a second opinion :)

src/docs/product/discover-queries/query-builder.mdx
imatwawana: Your edits mostly make sense to me @lizokm! I would make a couple tweaks. This first one because I think "data based on total transactions" is the full noun phrase (please confirm, @shruthilayaj). So, to make it a bit shorter:

> You can compare your query results against total transactions data by enabling the "Total Transactions" toggle in the chart footer.

Or if it feels like it needs to be spelled out for clarity, then:

> You can compare your query results against data based on total transactions by enabling the "Total Transactions" toggle in the chart footer.

I'm pretty sure this needs a comma after transactions:

> You have some unparameterized transactions, which are incompatible with [Dynamic Sampling](/product/data-management-settings/dynamic-sampling/).

This next bit is trickier. I think the edit is probably an improvement, but I'm not actually sure what you're trying to say here @shruthilayaj. Do you mean that the y-axis isn't supported/isn't available or are those first two clauses supposed to work in some other way? Is the rest of list starting at _equations_ a list of all the things that the y-axis isn't supported or available for?

> Selected y-axis is not supported, not available for equations, `apdex()`, `user_misery()` and aggregations performed on fields and tags (for example `count_unique(device.os)`)

Lastly, just the word _Discover_ should always be bolded, but otherwise, these edits look good to me.

shruthilayaj:
This feature is going to removed from the UI

PR #5720, opened 2022-11-04
move note from snippets into common content

imatwawana:
Removes notes about fallback behavior from snippets to common content. Clarified language of note by using the name of the option/property vs a descriptor.

### DIFF
--- a/src/platforms/common/enriching-events/identify-user.mdx
+++ b/src/platforms/common/enriching-events/identify-user.mdx
@@ -17,14 +17,28 @@ Users consist of a few critical pieces of information that construct a unique id
 
 ### `id`
 
+<Note>
+
+If you don't provide an `id`, the SDK falls back to `installationId`, which the SDK randomly generates once during an app's installation.
+
+</Note>
+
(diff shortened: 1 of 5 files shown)


src/platforms/common/enriching-events/identify-user.mdx
```diff
@@ -17,14 +17,28 @@ Users consist of a few critical pieces of information that construct a unique id
...
 
 <PlatformSection supported={["apple"]}>
 
 ### `userId`
 
+<Note>
+
+Since version 6.0.1, if you don't provide a `userId`, the SDK falls back to `installationId`, which the SDK randomly generates once during an app's installation.
+
+</Note>
```
philipphofmann, line 38: This only applies to Apple. I don't think it should stay here.

imatwawana: It's in a platform section that's only supported for Apple (see line 30), so it will only display for Apple.

philipphofmann (approved):
LGTM

PR #5753, opened 2022-11-09
Update example to follow HMAC best practice

mdtro:
Update the Python example for the webhook signature comparison to follow best practices.

Using `hmac.compare_digest()` mitigates [timing-based attacks](https://en.wikipedia.org/wiki/Timing_attack) on the signature verification.

I've also updated the line to pull the signature header value to use the `.get()` method in order to avoid a `KeyError` exception.


leeandher (approved):
Looks good :+1:

PR #5761, opened 2022-11-10
update top and bottom nav

imatwawana:
Adds Product link to the persistent top nav 
Removes Product Guides and API from the bottom nav since it's already at the top


lizokm (approved):
This looks great!

PR #5764, opened 2022-11-11
added platform section hidden from dotnet for server name

imatwawana:
Adds a platform section within the description for server-name to hide the last sentence from .NET.

### DIFF
--- a/src/platforms/common/configuration/options.mdx
+++ b/src/platforms/common/configuration/options.mdx
@@ -242,8 +242,13 @@ If possible, we recommended turning on this feature to send all such data by def
 
 <ConfigKey name="server-name" supported={["python", "node", "ruby", "php", "java", "dart", "dotnet"]} notSupported={["android"]}>
 
-This option can be used to supply a "server name." When provided, the name of the server is sent along and persisted in the event. For many integrations the server name actually corresponds to the device hostname, even in situations where the machine is not actually a server. Most SDKs
-will attempt to auto-discover this value.
+This option can be used to supply a "server name." When provided, the name of the server is sent along and persisted in the event. For many integrations the server name actually corresponds to the device hostname, even in situations where the machine is not actually a server.
+
+<PlatformSection notSupported={["dotnet"]}>
+
+Most SDKs will attempt to auto-discover this value.
+
+</PlatformSection>
 
 </ConfigKey>
 


src/platforms/common/configuration/options.mdx
```diff
@@ -242,8 +242,13 @@ If possible, we recommended turning on this feature to send all such data by def
 
 <ConfigKey name="server-name" supported={["python", "node", "ruby", "php", "java", "dart", "dotnet"]} notSupported={["android"]}>
 
-This option can be used to supply a "server name." When provided, the name of the server is sent along and persisted in the event. For many integrations the server name actually corresponds to the device hostname, even in situations where the machine is not actually a server. Most SDKs
-will attempt to auto-discover this value.
+This option can be used to supply a "server name." When provided, the name of the server is sent along and persisted in the event. For many integrations the server name actually corresponds to the device hostname, even in situations where the machine is not actually a server.
+
+<PlatformSection notSupported={["dotnet"]}>
```
bruno-garcia, line 247: .NET also does this but hides behind `SendDefaultPii` if not on ASPNET (when we know for sure it's a server).

The reason is taht it could be a desktop app, so that means we'd be sending a device name, potentiall "Bruno's Computer". That's why the opt-in through htat PII guard

imatwawana: Can you suggest an edit @bruno-garcia?

mattjohnsonpint: I pushed an update that explains this better.

src/platforms/common/configuration/options.mdx
```diff
@@ -242,8 +242,19 @@ If possible, we recommended turning on this feature to send all such data by def
...
+
+Most SDKs will attempt to auto-discover this value.
+
+</PlatformSection>
+
+<PlatformSection supported={["dotnet"]}>
+
+For ASP.NET and ASP.NET Core applications, the value will default to the server's name.  For other application types, the value will default to the computer's name only when the `SendDefaultPii` is set to `true`, because the computer's name can be considered personally identifiable information in the case of a desktop or mobile application.
```
imatwawana, line 255:
```suggestion
For ASP.NET and ASP.NET Core applications, the value will default to the server's name.  For other application types, the value will default to the computer's name only when the `SendDefaultPii` is set to `true`, because the computer's name can be considered personally identifiable information (PII) in the case of a desktop or mobile application.
```

imatwawana:
I'm happy with the update, so I'll go ahead and merge. Thanks for your help @mattjohnsonpint!

Issue #5723, opened 2022-11-07
Square brackets break fingerprinting rules


spaspala:
### Core or SDK?

Core Sentry product

### Which part? Which one?

Issue Grouping > Fingerprinting Rules

### Description

When setting fingerprinting rules, if I want to include a message that has `[` and `]` in it, the rule will not work.

Example: `error.type:UnhandledRejection message:"Non-Error promise rejection captured with value: FirebaseError: [code=already-exists]: Document already exists: projects/{org slug}/databases/(default)/documents/snippets/*" -> {rule name}`

Found out that the workaround to this is replacing the square brackets with asterisks (`*`). So the example above would look like: `error.type:UnhandledRejection message:"Non-Error promise rejection captured with value: FirebaseError: *code=already-exists*: Document already exists: projects/{org slug}/databases/(default)/documents/snippets/*" -> {rule name}`

### Suggested Solution

Add in the documentation a section that explains that when setting issue grouping rules, square brackets will not work and asterisks should be used in their place.

imatwawana:
Hi @spaspala! Thanks for bringing this up. I think GitHub might be interfering with what you're trying to input, so I'm not super clear if you're trying to indicate quotation marks or asterisks. Can you try formatting your before and after examples as code using the angled brackets in the toolbar of the comment field?
![image](https://user-images.githubusercontent.com/76437239/200618072-a74a23b2-1e17-4305-a432-66bbc8276d41.png)

spaspala:
Thank you for pointing this out @imatwawana, I have updated it!

imatwawana:
Much clearer, thank you!

PR #5765, opened 2022-11-11
add note about square brackets for fingerprinting

imatwawana:
Adds note about replacing square brackets with asterisks in fingerprinting rules.
Updates a link about syntax to point to search docs.

### DIFF
--- a/src/docs/product/data-management-settings/event-grouping/fingerprint-rules.mdx
+++ b/src/docs/product/data-management-settings/event-grouping/fingerprint-rules.mdx
@@ -15,7 +15,9 @@ These rules can be configured on a per-project basis in **[Project] > Settings >
 1. Identify the match logic for grouping issues together.
 1. Set the match logic and the fingerprint for it.
 
-The syntax follows the syntax from [Discover queries](/product/discover-queries/). If you want to negate the match, prefix the expression with an exclamation mark (`!`).
+The syntax follows the [syntax](/product/sentry-basics/search/#syntax) from **Discover** queries. If you want to negate the match, prefix the expression with an exclamation mark (`!`).
+
+If you want to create a rule that includes square brackets `[ ]`, replace them with asterisks `*`.
 
 Sentry attempts to match against all values that are configured in the fingerprint. In the case of stack traces, all frames are considered. If the event data matches all the values in a line for a matcher and expression, then the fingerprint is applied:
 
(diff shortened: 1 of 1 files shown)


src/docs/product/data-management-settings/event-grouping/fingerprint-rules.mdx
```diff
@@ -15,7 +15,9 @@ These rules can be configured on a per-project basis in **[Project] > Settings >
 1. Identify the match logic for grouping issues together.
 1. Set the match logic and the fingerprint for it.
 
-The syntax follows the syntax from [Discover queries](/product/discover-queries/). If you want to negate the match, prefix the expression with an exclamation mark (`!`).
+The syntax follows the [syntax](/product/sentry-basics/search/#syntax) from **Discover** queries. If you want to negate the match, prefix the expression with an exclamation mark (`!`).
+
+If you want to create a rule that includes square brackets `[ ]`, replace them with asterisks `*`.
```
jjbayer, line 20: Was there an issue with square brackets that we or a customer ran into?

imatwawana: Yep, see issue [5723](https://github.com/getsentry/sentry-docs/issues/5723)

jjbayer: Thanks, sorry I somehow missed that linked issue. I will investigate why the square brackets don't work, maybe we can actually fix this in the code.

imatwawana: No worries! I'll leave this open and wait for you to circle back.

jjbayer: Turns out that brackets need to be escaped because they are part of [glob syntax ](https://en.wikipedia.org/wiki/Glob_(programming)#Syntax). Globbing is already mentioned under `## Matchers`, so instead of adding this line here, I would expand the explanation there to

```markdown
## Matchers

Matchers allow you to use [glob patterns](https://en.wikipedia.org/wiki/Glob_(programming)).
If you want to create a rule that includes square brackets `[ ]` or a literal `*`, escape them with a backslash, that is, `\[`, `\]` or `\*`. 

The following matchers are available:
```

imatwawana: Makes sense to me!

jjbayer (approved):

PR #5769, opened 2022-11-11
feat(perf): Note wildcard support in transaction name search

gggritso:
Wildcards are supported in the UI as of https://github.com/getsentry/sentry/pull/41268! I opted to link to the "Wildcards" section, but I think that might be a mistake, since that section talks about wildcards in tag values and so on. Maybe it'd be better to do something like "(e.g., '/transaction/*')" instead of the link?

### DIFF
--- a/src/docs/product/performance/filters-display/index.mdx
+++ b/src/docs/product/performance/filters-display/index.mdx
@@ -42,7 +42,7 @@ When you select "All Transactions", the table includes all transactions across a
 
 ### Transaction Name Search
 
-Rather than searching for transactions by matching event properties, the search bar on the **Performance** page filters by transaction name only. You can search for specific transaction names or use wildcards, which can take the place of one or more characters in a search term.
+Rather than searching for transactions by matching event properties, the search bar on the **Performance** page filters by transaction name only. You can search for specific transaction names or use wildcards, which can take the place of one or more characters in a search term (e.g., `/api/*/organization`).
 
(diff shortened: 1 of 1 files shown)


src/docs/product/performance/filters-display/index.mdx
```diff
@@ -42,7 +42,7 @@ When you select "All Transactions", the table includes all transactions across a
 
 ### Transaction Name Search
 
-On the **Performance** page, the search bar filters by transaction name only, rather than searching for transactions by matching event properties.
+On the **Performance** page, the search bar filters by transaction name only, rather than searching for transactions by matching event properties. You can search for specific transaction names as well as [wildcards](/product/sentry-basics/search/#wildcards).
```
lizokm, line 45:
```suggestion
Rather than searching for transactions by matching event properties, the search bar on the **Performance** page filters by transaction name only. You can search for specific transaction names or use wildcards, which can take the place of one or more characters in a search term.
```
I agree that it would make more sense to add a bit more of a description of how to incorporate wildcards in search here, rather than linking out to a different section of docs. I'd add "For example....[specific example]." At the end of the paragraph above.

lizokm (commented):
Changed the formatting slightly and left a comment. I think adding an example of a search that uses a wildcard would be helpful here :)

gggritso:
@lizokm makes sense to me, thanks! I added a small example

lizokm (approved):
LGTM!

PR #5770, opened 2022-11-11
update page title and remove old notes

imatwawana:
Change _Set Up_ to _Set Up Performance_ so that it's clearer what this page is when it's displayed in a PageGrid

### DIFF
--- a/src/components/platformSidebar.tsx
+++ b/src/components/platformSidebar.tsx
@@ -86,7 +86,7 @@ export const PlatformSidebar = ({
       <DynamicNav
         root={`/${pathRoot}/performance`}
         title="Performance Monitoring"
-        prependLinks={[[`/${pathRoot}/performance/`, "Set Up"]]}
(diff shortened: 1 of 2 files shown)


imatwawana:
@untitaker - I've tried updating the page name [here](https://github.com/getsentry/sentry-docs/pull/5770/files#diff-87349e90c2fa6f749f3e4beab7aba25ba2e30b3c797722acfa737c0d32574009R2), but in the [preview](https://sentry-docs-git-fix-platforms-performance-page-fixes.sentry.dev/platforms/javascript/performance/), it doesn't seem to work. It shows up as changed on the page, but not in the left hand navigation. But I can't figure out what I'm missing to make that update as well.

untitaker (approved):
sorry for the late response, you need to change it here; https://github.com/getsentry/sentry-docs/blob/1a20cf56566d53fb08b28808d04c882309abf93e/src/components/platformSidebar.tsx#L89

PR #5772, opened 2022-11-13
fix(native): hide "Enable Tracing" section in...

supervacuus:
... Performance Monitoring setup. `sentry-native` has no separate "Enabling" setup step (beyond configuring the trace sampling). This currently renders a misleading message raised here: https://github.com/getsentry/sentry-native/issues/601#issuecomment-1259977174.

@Swatinem pls have a look.

### DIFF
--- a/src/platforms/common/performance/index.mdx
+++ b/src/platforms/common/performance/index.mdx
@@ -66,7 +66,7 @@ Automatic instrumentation for monitoring the performance of your application is 
 
 </PlatformSection>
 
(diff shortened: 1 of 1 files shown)


Swatinem (approved):
I believe there is also the `performance/enable-tracing` snippet that can now be removed?

supervacuus:
> I believe there is also the `performance/enable-tracing` snippet that can now be removed?

maybe I misunderstand, but I think the error message appears exactly because that snippet is missing for `native`. If we had something to put into the "Enable Tracing" section, this would require a file `native.mdx` under `src/platform-includes/performance/enable-tracing/'. But there currently is none.

Swatinem:
:-D okay, then I completely misunderstood that

PR #5776, opened 2022-11-14
feat(profiling): add nodejs profiling docs

JonasBa:
First pass on adding the docs for profiling to node. Since this is an alpha package for now, should we indicate this somewhere in the docs? 

Other than that, feedback very much wanted :D


src/platforms/node/common/profiling/index.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+
+## How does it work?
+
+Under the hood, sentry profiler leverages V8's [CpuProfiler](https://v8docs.nodesource.com/node-18.2/d2/d34/classv8_1_1_cpu_profiler.html) to collect stack samples. This means that sentry/profiling-node is written as a [native addon](https://nodejs.org/docs/latest-v18.x/api/addons.html) for Node and will not run in environments like Deno or Bun. Profiling enhances tracing by providing profiles for the individual transactions - this allows you to look at higher level performance information like transaction and span durations before diving deeper and looking at profiles.
+
+## Runtime flags
+
+Profiling package enables you to run the profiler in either eager or lazy [mode](https://v8docs.nodesource.com/node-18.2/d2/dc3/namespacev8.html#a874b4921ddee43bef58d8538e3149374). During our initial tests, we have seen that the default profiling mode (lazy) can cause calls to startProfiling to take a long time - sometimes in the range of couple hundred milliseconds. We found that this occurs when we are starting a profile when no other profiles are started. This is the benefit of lazy logging, it adds no overhead when no profiles are running.
+
```
lizokm, line 50:
```suggestion
## Runtime Flags

The Profiling package enables you to run the profiler in either [eager or lazy mode](https://v8docs.nodesource.com/node-18.2/d2/dc3/namespacev8.html#a874b4921ddee43bef58d8538e3149374). During initial tests, when we started a profile while no other profiles were running, we saw that the default profiling mode (lazy) sometimes caused startProfiling calls to take a long time, (sometimes in the range of a couple hundred milliseconds). The benefit of lazy logging is that it doesn't add any overhead when profiles aren't running.

```
Hoping that my edits didn't change the meaning of what you were trying to say :)

src/platforms/node/common/profiling/index.mdx
```diff
@@ -0,0 +1,60 @@
...
+If you find this behavior prohibitive for your use case, there is an environment flag that you can set which will cause the profiler to run in eager mode. Eager mode improves the performance of calls to startProfiling at the tradeoff of small constant CPU overhead.
+
```
lizokm, line 52:
```suggestion
If this behavior is prohibitive for your use case, there's an environment flag that you can set which will cause the profiler to run in eager mode. Eager mode improves the performance of calls to startProfiling, but the tradeoff is small, constant CPU overhead.

```

src/platforms/node/common/profiling/index.mdx
```diff
@@ -0,0 +1,60 @@
...
+```bash
+// Run profiler in eager mode
+SENTRY_PROFILER_LOGGING_MODE=eager node script.js
+```
+
+It is always advised that you have your own CPU resource monitoring in place as the actual resource use could be environment dependant.
```
lizokm, line 58:
```suggestion
It's always advisable that you have your own CPU resource-monitoring in place, because the actual resource use could be environment-dependent.
```

lizokm (approved):
Hi there! I'm the new docs manager :) I've made a few tweaks, but overall this is looking solid.

JonasBa:
Thanks so much @lizokm - I agree with all suggestions and they are all applied now

src/platforms/node/common/profiling/index.mdx
```diff
@@ -0,0 +1,60 @@
...
+  op: "transaction",
+  name: "My Transaction",
+});
+
+// Any code executed between startTransaction and transaction.finish
+// will now be automatically profiled.
+
+transaction.finish();
+```
+
```
imatwawana, line 41: Removing extra line break
```suggestion
```

src/platforms/node/common/profiling/index.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+## How Does It Work?
+
+Under the hood, the sentry profiler leverages V8's [CpuProfiler](https://v8docs.nodesource.com/node-18.2/d2/d34/classv8_1_1_cpu_profiler.html) to collect stack samples. This means that sentry/profiling-node is written as a [native add-on](https://nodejs.org/docs/latest-v18.x/api/addons.html) for Node and won't run in environments like Deno or Bun. Profiling enhances tracing by providing profiles for individual transactions. This allows you to look at higher level performance information like transaction and span durations before diving deeper and looking at profiles.
```
imatwawana, line 45: Should "Sentry" in the first sentence be capitalized? I get why it isn't in the second sentence when it's part of the a larger term. I changed "leverages" to "uses" to try to simplify the language as much as possible.

```suggestion
Under the hood, the Sentry profiler uses V8's [CpuProfiler](https://v8docs.nodesource.com/node-18.2/d2/d34/classv8_1_1_cpu_profiler.html) to collect stack samples. This means that sentry/profiling-node is written as a [native add-on](https://nodejs.org/docs/latest-v18.x/api/addons.html) for Node and won't run in environments like Deno or Bun. Profiling enhances tracing by providing profiles for individual transactions. This allows you to look at higher level performance information like transaction and span durations before diving deeper and looking at profiles.
```

src/platforms/node/common/profiling/index.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+## Runtime Flags
+
+The Profiling package enables you to run the profiler in either [eager or lazy mode](https://v8docs.nodesource.com/node-18.2/d2/dc3/namespacev8.html#a874b4921ddee43bef58d8538e3149374). During initial tests, when we started a profile while no other profiles were running, we saw that the default profiling mode (lazy) sometimes caused startProfiling calls to take a long time, (sometimes in the range of a couple hundred milliseconds). The benefit of lazy logging is that it doesn't add any overhead when profiles aren't running.
```
imatwawana, line 49: Just wondering if we can make this more concise:
```suggestion
The **Profiling** package enables you to run the profiler in either [eager or lazy mode](https://v8docs.nodesource.com/node-18.2/d2/dc3/namespacev8.html#a874b4921ddee43bef58d8538e3149374). When a profile is started while no other profiles are running, the default profiling mode (lazy) sometimes causes `startProfiling` calls to take a long time (sometimes in the range of a couple hundred milliseconds). The benefit of lazy logging is that it doesn't add any overhead when profiles aren't running.
```

src/platforms/node/common/profiling/index.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+If this behavior is prohibitive for your use case, there's an environment flag that you can set which will cause the profiler to run in eager mode. Eager mode improves the performance of calls to startProfiling, but the tradeoff is small, constant CPU overhead.
```
imatwawana, line 51:
```suggestion
If this behavior is prohibitive for your use case, you can set the environment flag shown below, which will cause the profiler to run in eager mode. Eager mode improves the performance of calls to `startProfiling`, but the tradeoff is small, constant CPU overhead.
```

src/platforms/node/common/profiling/index.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+```bash
+// Run profiler in eager mode
+SENTRY_PROFILER_LOGGING_MODE=eager node script.js
+```
+
+It's always advisable that you have your own CPU resource-monitoring in place, because the actual resource use could be environment-dependent.
```
imatwawana, line 58:
```suggestion
We recommend you have your own CPU resource-monitoring in place, because the actual resource use could be environment-dependent.
```

src/platforms/node/common/profiling/index.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+If you have feedback or experience different behavior please open an issue on sentry/profiling-node repository.
```
imatwawana, line 60:
```suggestion
If you have feedback or experience different behavior, please open an issue on sentry/profiling-node repository.
```

imatwawana (approved):
I've made some suggestions to make the text more concise. @lizokm, you may want to review my suggestions and make sure you agree. This is also fine as is though apart from a couple formatting tweaks for style. Apart from those, you can take or leave my suggestions.

PR #5779, opened 2022-11-14
feat(js): Add note about cookies to troubleshooting guide

lobsterkatie:
Fairly frequently, new customers using any of our browser-based SDKS get themselves into this situation:

- They create a Sentry account and log into it.
- The Sentry front end sets some cookies on their machine.
- They create a dummy test app, and open it in their browser.
- They send a test event to Sentry.
- They look at their cookie storage, see Sentry cookies, and get worried that they were set by the SDK (which they weren't).

This adds a note to the troubleshooting section of the JS docs explaining the true origin of the cookies.

### DIFF
--- a/src/platforms/javascript/common/troubleshooting/index.mdx
+++ b/src/platforms/javascript/common/troubleshooting/index.mdx
@@ -62,6 +62,12 @@ Most community CDNs properly set an `Access-Control-Allow-Origin` header.
  Access-Control-Allow-Origin: *
 ```
 
+## Sentry Cookies
+
+If you are testing your setup and see Sentry cookies in your local cookie storage, note that these cookies are NOT set by any Sentry SDK, and therefore will not be set on your users' machines. Those cookies come from the Sentry frontend, and do things like allow you to stay logged in for a period of time after you close a tab containing Sentry.
+
+If you instead use an incognito window to send test events to Sentry, it will mimic more closely what will happen for your users (assuming they are not themselves Sentry customers, of course!), and you can confirm that no Sentry cookies are set.
+
 ## Unexpected OPTIONS request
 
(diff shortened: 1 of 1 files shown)


src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -62,6 +62,12 @@ Most community CDNs properly set an `Access-Control-Allow-Origin` header.
  Access-Control-Allow-Origin: *
 ```
 
+## Sentry Cookies
+
+If you are testing your setup and see Sentry cookies in your local cookie storage, note that these cookies are NOT set by any Sentry SDK, and therefore will not be set on your users' machines. Those cookies come from the Sentry frontend, and do things like allow you to stay logged in for a period of time after you close a tab containing Sentry.
+
+If you instead use an incognito window to send test events to Sentry, it will mimic more closely what will happen for your users (assuming they are not themselves Sentry customers, of course!), and you can confirm that no Sentry cookies are set.
```
lizokm, line 69:
```suggestion
If you're testing your setup and see Sentry cookies in your local cookie storage, note that they're NOT set by any Sentry SDK and won't be set on your users' machines. Sentry cookies come from our frontend, and do things like allow you to stay logged in for a period of time after you close a Sentry tab.

If you want to see what your users see, (assuming they're not Sentry customers), use an incognito window to send test events to Sentry instead. That way, you can confirm that no Sentry cookies are set.
```

lizokm (approved):
Hi Katie! Made a few tweaks, but otherwise this looks good to me.

src/platforms/javascript/common/troubleshooting/index.mdx
mdtro: Do the SDKs send the Sentry cookies if the user visiting the website is a Sentry customer (and logged in)? I would not expect this to be the case.

Since the cookies are scoped to our domain (`.sentry.io`), I would expect the user's request to the website utilizing the Sentry SDK to only include relevant cookies for said website.

Issue #5784, opened 2022-11-15
[sentry-unplugin] Restructure Source Maps Upload Docs


lforst:
Ref https://github.com/getsentry/sentry-docs/issues/5718

With the upcoming generic bundler plugin https://github.com/getsentry/sentry-javascript-bundler-plugins it makes sense for us to revisit the source maps upload docs.

### What we wanna do

Currently, the source maps upload is described in two steps: 1. generating source maps 2. uploading source maps. While this makes sense from a maintainer POV users only care about what the process is for their own toolchain so instead, we propose removing the "Generating Source Maps" section and only having complete guides for each individual bundler/build-tool in the "Uploading Source Maps" section.

PR #5786, opened 2022-11-15
Move multiple origins page into source maps troubleshooting page

lforst:
This PR is part of a bigger restructuring https://github.com/getsentry/sentry-docs/issues/5784 of the source maps upload docs.

Currently, the "Multiple Origins" page lives under the "Uploading Source Maps" section which is not really fitting and in conflict with future pages we want to have under the "Uploading Source Maps" section.

This PR moves that page into the troubleshooting page as a subheading, as it is more of an edge case for users. Additionally, this PR adds a redirect in the `vercel.json` that points from the old page to the new heading on the troubleshooting page


Lms24 (approved):

vercel.json
```diff
@@ -375,6 +375,14 @@
     {
       "source": "/platforms/node/guides/connect/event-processors/",
       "destination": "/platforms/node/enriching-events/event-processors/"
+    },
+    {
+      "source": "/platforms/javascript/sourcemaps/uploading/hosting-publicly/",
+      "destination": "/platforms/javascript/sourcemaps/troubleshooting_js/#multiple-origins"
+    },
```
imatwawana, line 382: I think this is supposed to be a redirect for the multiple-origins page, yes?

```suggestion
    {
      "source": "/platforms/javascript/sourcemaps/uploading/multiple-origins/",
      "destination": "/platforms/javascript/sourcemaps/troubleshooting_js/#multiple-origins"
    },
```

imatwawana (commented):
I think one of the redirects has a typo in it, but otherwise, this looks good to me.

vercel.json
lforst: Phew thanks for catching this!

PR #5789, opened 2022-11-15
feat(nextjs): Add `excludeServerRoutes` option to manual setup page

lobsterkatie:
This adds the new `excludeServerRoutes` option in the nextjs SDK (added in https://github.com/getsentry/sentry-javascript/pull/6207) to the docs. It also pulls all of the auto-instrumentation config into its own section in the manual setup page.

(Note: The only part of this which is actually new content is the `Opt Out of Auto-instrumentation on Specific Routes` section. Everything else is just stuff moving around.)

### DIFF
--- a/src/platforms/javascript/guides/nextjs/manual-setup.mdx
+++ b/src/platforms/javascript/guides/nextjs/manual-setup.mdx
@@ -156,14 +156,17 @@ const moduleExports = {
 
   // Optional build-time configuration options
   sentry: {
-    // See the 'Configure Source Maps' and 'Configure Legacy Browser Support'
-    // sections below for information on the following options:
-    //   - disableServerWebpackPlugin
-    //   - disableClientWebpackPlugin
-    //   - autoInstrumentServerFunctions
-    //   - hideSourceMaps
-    //   - widenClientFileUpload
-    //   - transpileClientSDK
+    // See the sections below for information on the following options:
+    //   'Configure Source Maps':
+    //     - disableServerWebpackPlugin
+    //     - disableClientWebpackPlugin
+    //     - hideSourceMaps
+    //     - widenClientFileUpload
+    //   'Configure Legacy Browser Support':
+    //     - transpileClientSDK
+    //   'Configure Serverside Auto-instrumentation':
+    //     - autoInstrumentServerFunctions
+    //     - excludeServerRoutes
   },
 };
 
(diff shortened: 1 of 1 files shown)


lforst (approved):

src/platforms/javascript/guides/nextjs/manual-setup.mdx
```diff
@@ -355,3 +310,82 @@ const moduleExports = {
 ```
 
 (This assumes you are using [the `next.config.js` setup shown above](#extend-nextjs-configuration).)
+
+## Configure Serverside Auto-instrumentation
```
lizokm, line 314:
```suggestion
## Configure Server-side Auto-instrumentation
```

src/platforms/javascript/guides/nextjs/manual-setup.mdx
```diff
@@ -355,3 +310,82 @@ const moduleExports = {
...
+
+The SDK will automatically instrument API routes and server-side [Next.js data fetching methods](https://nextjs.org/docs/basic-features/data-fetching/overview) with error and performance monitoring.
+
+### Disable API Route and Data Fetching Auto-instrumentation Entirely
+
+_(New in version 7.14.0)_
+
+To disable the automatic instrumentation of API route handlers and serverside data fetching functions, set the `autoInstrumentServerFunctions` to `false`.
```
lizokm, line 322:
```suggestion
To disable the automatic instrumentation of API route handlers and server-side data fetching functions, set the `autoInstrumentServerFunctions` to `false`.
```

src/platforms/javascript/guides/nextjs/manual-setup.mdx
```diff
@@ -355,3 +310,82 @@ const moduleExports = {
...
+
+```javascript {filename:next.config.js}
+const moduleExports = {
+  sentry: {
+    autoInstrumentServerFunctions: false,
+  },
+};
+```
+
+Under the hood, when using this option, the SDK is using a Webpack loader to wrap all your API route handlers and data fetching methods.
```
lizokm, line 332:
```suggestion
With this option, under the hood, the SDK is using a Webpack loader to wrap all your API route handlers and data fetching methods.
```

src/platforms/javascript/guides/nextjs/manual-setup.mdx
```diff
@@ -355,3 +310,82 @@ const moduleExports = {
...
+
+### Opt In to Auto-instrumentation on Specific Routes
+
+_(New in version 7.14.0)_
+
+If the automatic instrumentation doesn't work for your use case, you can turn it off globally and choose instead to only wrap specific API route handlers or data fetching functions.
```
lizokm, line 338:
```suggestion
If the automatic instrumentation doesn't work for your use case, you can turn it off globally and choose to only wrap specific API route handlers or data fetching functions instead.
```

src/platforms/javascript/guides/nextjs/manual-setup.mdx
```diff
@@ -355,3 +310,82 @@ const moduleExports = {
...
+- `withSentryServerSideErrorGetInitialProps` for `getInitialProps` in [custom Error pages](https://nextjs.org/docs/advanced-features/custom-error-page)
+- `withSentryServerSideAppGetInitialProps` for `getInitialProps` in [custom `App` components](https://nextjs.org/docs/advanced-features/custom-app)
+- `withSentryServerSideDocumentGetInitialProps` for `getInitialProps` in [custom `Document` components](https://nextjs.org/docs/advanced-features/custom-document)
+
+### Opt Out of Auto-instrumentation on Specific Routes
+
+_(New in version 7.20.0)_
+
+If you want auto-instrumenatation to apply by default, but want to exclude certain routes, use the `excludeServerRoutes` option in the `sentry` object in your `next.config.js`:
+
```
lizokm, line 377:
```suggestion
If you want auto-instrumenatation to apply by default, but also want to exclude certain routes, use the `excludeServerRoutes` option in the `sentry` object in your `next.config.js`:

```

src/platforms/javascript/guides/nextjs/manual-setup.mdx
```diff
@@ -355,3 +310,82 @@ const moduleExports = {
...
+      "/some/excluded/route",
+      "/excluded/route/with/[parameter]",
+      /^\/route\/beginning\/with\/some\/prefix/,
+      /\/routeContainingASpecificPathSegment\/?/,
+    ],
+  },
+};
+```
+
+Excluded routes can be specified either as strings (which must exactly match the full route you want to exclude, with a leading slash but no trailing one) or regexes.
```
lizokm, line 391:
```suggestion
Excluded routes can be specified either as regexes or strings, (which must match the full route you want to exclude exactly and have a leading slash, but no trailing one).
```

lizokm (approved):
Made a few small stylistic tweaks, but otherwise this looks great!

src/platforms/javascript/guides/nextjs/manual-setup.mdx
lobsterkatie:
```suggestion
Excluded routes can be specified either as regexes or strings. When using a string, make sure that it matches the route exactly, and has a leading slash but no trailing one.
```

PR #5778, opened 2022-11-14
feat(profiling): add wizard onboarding docs

eliashussary:
This PR adds "wizard" docs for profiling.

Related PR: https://github.com/getsentry/sentry/pull/41322

PR #5792, opened 2022-11-16
Revert "feat(profiling): add wizard onboarding docs"

Zylphrex:
Reverts getsentry/sentry-docs#5778

These steps are showing up in the onboarding experience as separate platforms now.


Zylphrex:
looks like @eliashussary found a fix for it so no need to revert getsentry/sentry#41432

PR #5793, opened 2022-11-16
fix(native): hub abstraction not implemented

supervacuus:
Let's clarify in the explanation of Hub & Scope that the Hub abstraction is not yet implemented in the Native SDK. While the tracing docs mention this shortcoming, changes to the scope from multiple threads are currently not coordinated (besides locking for mutually exclusive access) and affect other use cases too.

Please, @Swatinem and @imatwawana, I welcome your input.

### DIFF
--- a/src/platforms/common/enriching-events/scopes.mdx
+++ b/src/platforms/common/enriching-events/scopes.mdx
@@ -64,7 +64,7 @@ then merge the event with the topmost scope's data.
 
 <Alert level="warning" title="Hub Abstraction Is Not Implemented">
 
-The Native SDK currently only provides a single scope shared by all threads in the program and maintains no thread-local state by way of a hub. Because of this limitation, you must make sure that your threads don't logically interfere with each other's changes to the global scope, even though they are safe to do so purely from a data-race perspective. This, for example, affects use cases where you want to track many users in server-like programs while capturing events or when you want to trace multiple threads in performance monitoring.
+The Native SDK currently only provides a single scope shared by all threads in the program and maintains no thread-local state by way of a hub. Because of this limitation, you must make sure that your threads don't logically interfere with each other's changes to the global scope, even though they are safe to do so purely from a data-race perspective. This affects use cases where, for example, you want to track many users in server-like programs while capturing events or when you want to trace multiple threads in performance monitoring.
 
 For details on how this affects performance monitoring, check out the [Connect Errors With Spans](/platforms/native/performance/instrumentation/custom-instrumentation/#connect-errors-with-spans) section of "Custom Instrumentation".
 
--- a/src/platforms/common/enriching-events/scopes.mdx
+++ b/src/platforms/common/enriching-events/scopes.mdx
@@ -60,6 +60,14 @@ When you call a global function such as `capture_event` internally Sentry
 discovers the current hub and asks it to capture an event. Internally the hub will
 then merge the event with the topmost scope's data.
 
+<PlatformSection supported={["native"]}>
+<Alert level="warning" title="Hub abstraction is not implemented">
+The Native SDK currently only provides a single Scope shared by all threads in the program and maintains no thread-local state via a Hub. As a result of this limitation, client code must ensure that threads do not interfere with their changes to the global scope, even though it is safe to do so from a data-race perspective. This, for example, affects use cases where you want to track many users in server-like programs while capturing events or when you want to trace multiple threads in performance monitoring.
+
+For details on how this affects performance monitoring, please consult the [Connect Errors With Spans](/platforms/native/performance/instrumentation/custom-instrumentation/#connect-errors-with-spans) section of "Custom Instrumentation".
+</Alert> 
+</PlatformSection>
+
 ## Configuring the Scope
 
 The most useful operation when working with scopes is the `configure-scope` function. It can be used to reconfigure the current scope.
--- a/src/platforms/common/enriching-events/scopes.mdx
+++ b/src/platforms/common/enriching-events/scopes.mdx
@@ -61,10 +61,13 @@ discovers the current hub and asks it to capture an event. Internally the hub wi
 then merge the event with the topmost scope's data.
 
 <PlatformSection supported={["native"]}>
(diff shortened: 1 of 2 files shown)


src/platforms/common/enriching-events/scopes.mdx
```diff
@@ -60,6 +60,14 @@ When you call a global function such as `capture_event` internally Sentry
 discovers the current hub and asks it to capture an event. Internally the hub will
 then merge the event with the topmost scope's data.
 
+<PlatformSection supported={["native"]}>
+<Alert level="warning" title="Hub abstraction is not implemented">
+The Native SDK currently only provides a single Scope shared by all threads in the program and maintains no thread-local state via a Hub. As a result of this limitation, client code must ensure that threads do not interfere with their changes to the global scope, even though it is safe to do so from a data-race perspective. This, for example, affects use cases where you want to track many users in server-like programs while capturing events or when you want to trace multiple threads in performance monitoring.
+
+For details on how this affects performance monitoring, please consult the [Connect Errors With Spans](/platforms/native/performance/instrumentation/custom-instrumentation/#connect-errors-with-spans) section of "Custom Instrumentation".
+</Alert>
```
imatwawana, line 68:
```suggestion
<PlatformSection supported={["native"]}>

<Alert level="warning" title="Hub Abstraction Is Not Implemented">

The Native SDK currently only provides a single scope shared by all threads in the program and maintains no thread-local state by way of a hub. Because of this limitation, you must make sure that threads don't interfere with your client code changes to the global scope, even though it's safe to do so from a data-race perspective. This, for example, affects use cases where you want to track many users in server-like programs while capturing events or when you want to trace multiple threads in performance monitoring.

For details on how this affects performance monitoring, check out the [Connect Errors With Spans](/platforms/native/performance/instrumentation/custom-instrumentation/#connect-errors-with-spans) section of "Custom Instrumentation".

</Alert> 
```

imatwawana, line 68: Is not following this direction going to break anything? If not, then this shouldn't be in an alert like this as these are meant for warnings of that level. I'd suggest just making this an H3 section if you don't want it to get lost.

supervacuus: This is a big issue and warrants a warning, especially considering the paragraph above describes context and behavior that does not exist in the Native SDK. In that sense, what I describe is not optional but a warning that - for now - the expectation of an SDK that takes care of thread-local data for the user is not met. If we would make it an H3 section, it should probably replace the previous one.

supervacuus: That makes sense; thx. I would reword the following slightly to get closer to what I meant initially:

> Because of this limitation, you must make sure that threads don't interfere with your client code changes to the global scope, even though it's safe to do so from a data-race perspective.

to

Because of this limitation, you must make sure that your threads don't logically interfere with each other's changes to the global scope, even though they are safe to do purely from a data-race perspective.

imatwawana: Works for me. Updating my edit.

imatwawana: So if the content under How the Scope and Hub Work doesn't actually apply to Native, it seems like we shouldn't display it for that platform. What about adding a PlatformSection around that content and marking Native as unSupported and then having the content of the note be the Native content for that section.

supervacuus: I had something like that in mind when I mentioned replacing the previous section. But it requires a bit more merging of the contents; the concepts should still be introduced to the user, but with the context of them not being available as expected from other platforms. Will do this tomorrow.

imatwawana: Maybe before you do that work, let's wait for @Swatinem to give his opinion. And then see where we land.

Swatinem: I believe a warning like that is a good idea, as this can definitely cause surprises to the developer. Introducing the concept of Hubs as "how its done in other SDKs" might still be a fair thing to do. If there is enough incentiv / customer demand, we might eventually introduce hubs to the native SDK as well, but it is a lot of effort.

Swatinem (approved):

src/platforms/common/enriching-events/scopes.mdx
imatwawana: Okay so we'll leave it as a warning for now then.

src/platforms/common/enriching-events/scopes.mdx
```diff
@@ -60,6 +60,17 @@ When you call a global function such as `capture_event` internally Sentry
 discovers the current hub and asks it to capture an event. Internally the hub will
 then merge the event with the topmost scope's data.
 
+<PlatformSection supported={["native"]}>
+
+<Alert level="warning" title="Hub Abstraction Is Not Implemented">
+
+The Native SDK currently only provides a single scope shared by all threads in the program and maintains no thread-local state by way of a hub. Because of this limitation, you must make sure that your threads don't logically interfere with each other's changes to the global scope, even though they are safe to do so purely from a data-race perspective. This, for example, affects use cases where you want to track many users in server-like programs while capturing events or when you want to trace multiple threads in performance monitoring.
```
imatwawana, line 67:
```suggestion
The Native SDK currently only provides a single scope shared by all threads in the program and maintains no thread-local state by way of a hub. Because of this limitation, you must make sure that your threads don't logically interfere with each other's changes to the global scope, even though they are safe to do so purely from a data-race perspective. This affects use cases where, for example, you want to track many users in server-like programs while capturing events or when you want to trace multiple threads in performance monitoring.
```

imatwawana (approved):
I missed one thing in my edit, but otherwise, this is good to go!

imatwawana:
Shall I merge this for you now @supervacuus?

supervacuus:
Yes, @imatwawana, I think it makes sense the way it is now. It sharply contrasts the expected behavior from other SDKs with the reality in the Native SDK.

imatwawana:
Thanks for the contribution!!

PR #5800, opened 2022-11-17
feat(unity): IL2CPP & troubleshooting

bitsandfoxes:
The feature is enabled by default and can be opt out.


imatwawana (approved):
Made some wording tweaks and left one content suggestion

imatwawana (approved):

PR #5802, opened 2022-11-17
moved and updated troubleshooting

bitsandfoxes:
The troubleshooting page grew kind of wild.
I tried splitting it into two categories: `Runtime Issues` and `Editor & Buildtime Issues`.
I also updated parts where we now have proper solution to previous workarounds.

Waiting on https://github.com/getsentry/sentry-docs/pull/5800


imatwawana:
I understand the value in splitting the page up into two major sections, but I'm not sure I see the value in moving most of the content to a new page and leaving so little on the original page. What if you kept it all on one page with the following three headings:

Events Issues
Editor Issues - Installation
Editor Issues - Buildtime

PR #5803, opened 2022-11-17
feat(sdk): Add `addGlobalEventProcessor` to Event Processors page

lobsterkatie:
This adds the top-level `addGlobalEventProcessor` method to the Enriching Events -> Event Processors page alongside `addEventProcessor`.


src/platforms/common/enriching-events/event-processors.mdx
```diff
@@ -22,27 +22,40 @@ notSupported:
...
-Also, `eventProcessors` optionally receive the hint (see: <PlatformLink to="/configuration/filtering/#using-hints">hints</PlatformLink>).
+With event processors, you can enrich events with additional data. You can add your own event processors to either the current scope or the global one. The difference between `beforeSend` and event processors is that `beforeSend` is guaranteed to be run last, after all other event processors. In addition, event processors added with `addEventProcessor` run on the scope level, whereas `beforeSend` (and event processors added with `addGlobalEventProcessor`) run globally, no matter which scope you're in. Like `beforeSend`, event processors are passed both the event itself and <PlatformLink to="/configuration/filtering/#using-hints">a `hint`` object</PlatformLink> containing extra metadata.
+
+Event processors added to the global scope will run on every event sent after they are added.
```
lizokm, line 27:
```suggestion
You can enrich an event with additional data by running `eventProcessors` on either the scope level or globally. The difference between `beforeSend` and `eventProcessors` is that `beforeSend` always runs last (after all other event processors), and globally. While `eventProcessors` added with `addEventProcessor` run on the scope level, and `eventProcessors` added with `addGlobalEventProcessor`, run globally on every event sent after they are added, no matter which scope you’re in. 

Both `beforeSend` and `eventProcessors` are passed the event itself and  [a `hint` object](https://sentry-docs-git-kmclb-add-addglobaleventprocessor.sentry.dev/platforms/javascript/configuration/filtering/#using-hints)  containing extra metadata.
```

lizokm (approved):
Hi Katie! I've made some tweaks to make this read a bit more smoothly. Hopefully it still conveys the same meaning. I also felt that writing "event processors" in code made it easier to read. Please let me know if you disagree with this or any of my other updates :)

src/platforms/common/enriching-events/event-processors.mdx
lobsterkatie: Yup, I was consolidating what was there and changing what I needed to more than really wordsmithing, so thanks for cleaning it up! A few things I'm going to push back on, though:

- The user adds the event processors, then the SDK runs them. So that word choice is deliberate (and matches the method names `addXXX`).
- Can "while" be used on its own like that? I think of it as something which can only introduce a subordinate clause.
- On the writing-event-processors-in-code question, I don't think it's correct to treat them as code. `eventProcessors` is essentially a proper noun, the name of a variable. 'Event processors' is either an adjective-improper-noun pair or a compound improper noun, and that's how we're using them, to talk about the behavior of the overall class of things.

I don't disagree that the third sentence could be simplified and broken up, though. Lemme give it another shot.

src/platforms/common/enriching-events/event-processors.mdx
```diff
@@ -22,27 +22,45 @@ notSupported:
   - unreal
 ---
 
-With `eventProcessors` you can hook into the process of enriching the event with additional data. You can add your own `eventProcessor` on the current scope. The difference between `beforeSend` and `eventProcessors` is that `eventProcessors` run on the scope level whereas `beforeSend` runs globally, no matter which scope you're in.
-Also, `eventProcessors` optionally receive the hint (see: <PlatformLink to="/configuration/filtering/#using-hints">hints</PlatformLink>).
+You can enrich events with additional data by adding your own event processors, either on the scope level or globally. Though event processors are similar to `beforeSend`, there are two key differences:
+
+- `beforeSend` is guaranteed to be run last, after all other event processors, which means it gets the final version of the event version of the event right before it's sent (hence the name!). Event processors added with either of the methods below run in an undetermined order, meaning changes to the event may still be made after the event processor runs.
```
lizokm, line 27:
```suggestion
- `beforeSend` is guaranteed to be run last, after all other event processors, which means it gets the final version of the event right before it's sent (hence the name). Event processors added with either of the methods below run in an undetermined order, which means changes to the event may still be made after the event processor runs.
```

src/platforms/common/enriching-events/event-processors.mdx
```diff
@@ -22,27 +22,45 @@ notSupported:
...
+- While `beforeSend` and processors added with `Sentry.addGlobalEventProcessor` run globally, regardless of scope, processors added with `scope.addEventProcessor` only run on events captured while the scope in question is active.
+
```
lizokm, line 29:
```suggestion
- While `beforeSend` and processors added with `Sentry.addGlobalEventProcessor` run globally, regardless of scope, processors added with `scope.addEventProcessor` only run on events captured while that scope is active.

```

lizokm (approved):
Made a few tiny tweaks, but I think this looks great!

src/platforms/common/enriching-events/event-processors.mdx
lobsterkatie: Yikes. Proofreading. Thanks.

PR #5804, opened 2022-11-18
ref(filtering): Revamp SDK filtering docs

lobsterkatie:
Note: Converting to draft now that it's clear this is going to be more work than I thought. See https://github.com/getsentry/sentry-docs/issues/5811 for somewhat of a todo list.

---------

This is some general wordsmithing and reorganization of the SDK filtering page, specifically the `beforeSend` and hints sections, extracted from a future PR adding `beforeSendTransaction` to the docs.

Key Changes:

- Consolidate explanatory text and examples for hints into a single `Hints` section to avoid repetition.
- Fix JS `before-send-hint` snippet to not be identical to the `before-send-fingerprint` snippet and instead match the other languages' `before-send-hint` snippets. (UPDATE: Some other languages need this done, too.)
- Flesh out the list of properties available in hints. (UPDATE: Most other languages need this done.)
- Move list of properties to subpage.
- Make event properties use an include the way breadcrumb properties already do.
- Rename the `Decluttering Sentry` heading to something more descriptive.

TODO:
- This is all mushed together in one giant commit for now. Can clean that up a bit once we know what, if anything, we're going to do from here.

PR #5814, opened 2022-11-19
feat(js): Add `beforeSendTransaction` option

lobsterkatie:
This adds the new `beforeSendTransaction` option to the docs, both on the options page and alongside `beforeSend` in many places in the text. Because you can't selectively turn certain phrases on and off for different platforms, I tried to indicate where I could that it's not universally supported. (Right now, I believe JS is in fact the only platform to support it, though I expect that to change in the near future.) It also does a tiny bit of wordsmithing of the text around the additions, courtesy of @lizokm and @imatwawana.

Note that the one spot where updates haven't been done in this PR is the text of the Filtering page, as it's being worked on separately in https://github.com/getsentry/sentry-docs/pull/5804. Once the changes there are done, https://github.com/getsentry/sentry-docs/pull/5838 will add `beforeSendTransaction`.

### DIFF
--- a/src/docs/product/accounts/quotas/manage-event-stream-guide.mdx
+++ b/src/docs/product/accounts/quotas/manage-event-stream-guide.mdx
@@ -324,7 +324,7 @@ The SDK sample rate is not dynamic; changing it requires re-deployment. Also, ke
 
 ## 7. SDK Filtering: beforeSend {#1-sdk-filtering-beforesend}
 
-All Sentry SDKs support the `beforeSend` callback method. Once implemented, the method is invoked when the SDK captures an event, right before sending it to your Sentry account. It receives the event object as a parameter, so you can use that to modify the event's data or drop it completely (by returning `null`) based on your custom logic and the data available on the event, like _tags_, _environment_, _release version_, _error attributes_, and so on. Learn more in [Filtering Events](/platform-redirect/?next=/configuration/filtering/).
+All Sentry SDKs support the `beforeSend` callback method. Once implemented, the method is invoked when the SDK captures an error event, right before sending it to your Sentry account. It receives the event object as a parameter, so you can use that to modify the event's data or drop it completely (by returning `null`) based on your custom logic and the data available on the event, like _tags_, _environment_, _release version_, _error attributes_, and so on. Note that only error and message events pass through `beforeSend`. Tansaction events have a separate method, `beforeSendTransaction`, though it is not yet supported in all SDKs. Learn more about both methods in [Filtering Events](/platform-redirect/?next=/configuration/filtering/).
 
 ## 8. SDK Configuration {#2-sdk-configuration}
 
--- a/src/docs/product/accounts/quotas/manage-transaction-quota.mdx
+++ b/src/docs/product/accounts/quotas/manage-transaction-quota.mdx
@@ -138,6 +138,10 @@ If you have a rogue client, Sentry supports blocking an IP from sending data. Na
 
 If you discover a problematic release causing excessive noise, Sentry supports ignoring all events and attachments from that release. Navigate to **[Project] > Settings > Inbound Filters**, then add the release to the "Releases" field.
 
-## 3. SDK Configuration: Tracing Options {#2-sdk-configuration-tracing-options}
+## 3. SDK Filtering: beforeSendTransaction {#1-sdk-filtering-beforesendtransaction}
+
+Some Sentry SDKs support the `beforeSendTransaction` callback method. Once implemented, the method is invoked when the SDK captures a transaction event, right before sending it to your Sentry account. It receives the transaction event object as a parameter, so you can use that to modify the event's data or drop it completely (by returning `null`) based on your custom logic and the data available on the event, like _tags_, _environment_, _release version_, _transaction name_, and so on. Note that only transaction events pass through `beforeSendTransaction`. Error and message events have a separate method, `beforeSend`, which is supported in all SDKs. Learn more about both methods in [Filtering Events](/platform-redirect/?next=/configuration/filtering/).
+
+## 4. SDK Configuration: Tracing Options {#2-sdk-configuration-tracing-options}
 
 When you configure your the SDK, you can control the number of transactions that are sent to Sentry by setting the [tracing options](/platform-redirect/?next=/configuration/options/%23tracing-options), including either setting a sample rate or providing a function for sampling. You can also set up [custom instrumentation](/platform-redirect/?next=/performance/instrumentation/custom-instrumentation/) for performance monitoring to capture certain types of transactions.
--- a/src/platforms/common/data-management/sensitive-data/index.mdx
+++ b/src/platforms/common/data-management/sensitive-data/index.mdx
@@ -49,7 +49,7 @@ If you _do not_ wish to use the default PII behavior, you can also choose to ide
 
 ## Scrubbing Data
 
-SDKs provide a <PlatformIdentifier name="before-send" /> hook, which is invoked before an event is sent and can be used to modify event data to remove sensitive information. Using <PlatformIdentifier name="before-send" /> in the SDKs to **scrub any data before it is sent** is the recommended scrubbing approach, so sensitive data never leaves the local environment.
+SDKs provide a <PlatformIdentifier name="before-send" /> hook, which is invoked before an error or message event is sent and can be used to modify event data to remove sensitive information. (Some SDKs also provide a <PlatformIdentifier name="before-send-transaction" /> hook which does the same thing for transactions.) Using <PlatformIdentifier name="before-send" /> and <PlatformIdentifier name="before-send-transaction" /> in the SDKs to **scrub any data before it is sent** is the recommended scrubbing approach, so sensitive data never leaves the local environment.
 
 <PlatformContent includePath="configuration/before-send" />
 
--- a/src/platforms/common/enriching-events/event-processors.mdx
+++ b/src/platforms/common/enriching-events/event-processors.mdx
@@ -22,12 +22,12 @@ notSupported:
   - unreal
 ---
 
-You can enrich events with additional data by adding your own event processors, either on the scope level or globally. Though event processors are similar to `beforeSend`, there are two key differences:
+You can enrich events with additional data by adding your own event processors, either on the scope level or globally. Though event processors are similar to `beforeSend` and `beforeSendTransaction`, there are two key differences:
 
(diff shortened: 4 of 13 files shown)


src/docs/product/accounts/quotas/manage-event-stream-guide.mdx
```diff
@@ -324,7 +324,7 @@ The SDK sample rate is not dynamic; changing it requires re-deployment. Also, ke
...
+All Sentry SDKs support the `beforeSend` callback method. Once implemented, the method is invoked when the SDK captures an error event, right before sending it to your Sentry account. It receives the event object as a parameter, so you can use that to modify the event's data or drop it completely (by returning `null`) based on your custom logic and the data available on the event, like _tags_, _environment_, _release version_, _error attributes_, and so on. Note that only error and message events pass through `beforeSend`. Tansaction events have a separate method, `beforeSendTransaction`, though it is not yet supported in all SDKs. Learn more about both methods in [Filtering Events](/platform-redirect/?next=/configuration/filtering/).
```
imatwawana, line 327: I don't think you need the two sentences starting at "Note" because this is already a page devoted to error events. It's a given that things on this page may not apply to transactions. That kind of caveat only needs to exist on the parent page.

```suggestion
All Sentry SDKs support the `beforeSend` callback method. Once implemented, the method is invoked when the SDK captures an error event, right before sending it to your Sentry account. It receives the event object as a parameter, so you can use that to modify the event's data or drop it completely (by returning `null`) based on your custom logic and the data available on the event, like _tags_, _environment_, _release version_, _error attributes_, and so on. Learn more in [Filtering Events](/platform-redirect/?next=/configuration/filtering/).
```

src/docs/product/accounts/quotas/manage-transaction-quota.mdx
```diff
@@ -138,6 +138,10 @@ If you have a rogue client, Sentry supports blocking an IP from sending data. Na
 
 If you discover a problematic release causing excessive noise, Sentry supports ignoring all events and attachments from that release. Navigate to **[Project] > Settings > Inbound Filters**, then add the release to the "Releases" field.
 
-## 3. SDK Configuration: Tracing Options {#2-sdk-configuration-tracing-options}
+## 3. SDK Filtering: beforeSendTransaction {#1-sdk-filtering-beforesendtransaction}
```
imatwawana, line 141: Is this heading alias from an ancient version of this doc?
```suggestion
## 3. SDK Filtering: beforeSendTransaction {#1-sdk-filtering-beforesendtransaction}
```

src/docs/product/accounts/quotas/manage-transaction-quota.mdx
```diff
@@ -138,6 +138,10 @@ If you have a rogue client, Sentry supports blocking an IP from sending data. Na
...
+
+Some Sentry SDKs support the `beforeSendTransaction` callback method. Once implemented, the method is invoked when the SDK captures a transaction event, right before sending it to your Sentry account. It receives the transaction event object as a parameter, so you can use that to modify the event's data or drop it completely (by returning `null`) based on your custom logic and the data available on the event, like _tags_, _environment_, _release version_, _transaction name_, and so on. Note that only transaction events pass through `beforeSendTransaction`. Error and message events have a separate method, `beforeSend`, which is supported in all SDKs. Learn more about both methods in [Filtering Events](/platform-redirect/?next=/configuration/filtering/).
```
imatwawana, line 143: Same comment as for the similar section in the error events page.

```suggestion
Some Sentry SDKs support the `beforeSendTransaction` callback method. Once implemented, the method is invoked when the SDK captures a transaction event, right before sending it to your Sentry account. It receives the transaction event object as a parameter, so you can use that to modify the event's data or drop it completely (by returning `null`) based on your custom logic and the data available on the event, like _tags_, _environment_, _release version_, _transaction name_, and so on. Learn more in [Filtering Events](/platform-redirect/?next=/configuration/filtering/).
```

src/platforms/common/data-management/sensitive-data/index.mdx
```diff
@@ -49,7 +49,7 @@ If you _do not_ wish to use the default PII behavior, you can also choose to ide
 
 ## Scrubbing Data
 
-SDKs provide a <PlatformIdentifier name="before-send" /> hook, which is invoked before an event is sent and can be used to modify event data to remove sensitive information. Using <PlatformIdentifier name="before-send" /> in the SDKs to **scrub any data before it is sent** is the recommended scrubbing approach, so sensitive data never leaves the local environment.
+SDKs provide a <PlatformIdentifier name="before-send" /> hook, which is invoked before an error or message event is sent and can be used to modify event data to remove sensitive information. (Some SDKs also provide a <PlatformIdentifier name="before-send-transaction" /> hook which does the same thing for transactions.) Using <PlatformIdentifier name="before-send" /> and <PlatformIdentifier name="before-send-transaction" /> in the SDKs to **scrub any data before it is sent** is the recommended scrubbing approach, so sensitive data never leaves the local environment.
```
imatwawana, line 52: It's odd to introduce the idea of beforeSendTransaction hooks in a parenthetical and then follow up with a non-parenthetical that talks about them.

```suggestion
SDKs provide a <PlatformIdentifier name="before-send" /> hook, which is invoked before an error or message event is sent and can be used to modify event data to remove sensitive information. Some SDKs also provide a <PlatformIdentifier name="before-send-transaction" /> hook which does the same thing for transactions. We recommend using <PlatformIdentifier name="before-send" /> and <PlatformIdentifier name="before-send-transaction" /> in the SDKs to **scrub any data before it is sent**, so sensitive data never leaves the local environment.
```

imatwawana (commented):
I've made some overall wording tweaks and left a couple questions.

src/platforms/common/data-management/sensitive-data/index.mdx
lizokm, line 52:
```suggestion
SDKs provide a <PlatformIdentifier name="before-send" /> hook, which is invoked before an error or message event is sent and can be used to modify event data to remove sensitive information. (Some SDKs also provide a <PlatformIdentifier name="before-send-transaction" /> hook which does the same thing for transactions.) Using <PlatformIdentifier name="before-send" /> and <PlatformIdentifier name="before-send-transaction" /> in the SDKs to **scrub any data before it is sent** is the recommended scrubbing approach to ensure that sensitive data never leaves the local environment.
```

src/platforms/common/enriching-events/event-processors.mdx
```diff
@@ -22,12 +22,12 @@ notSupported:
...
+You can enrich events with additional data by adding your own event processors, either on the scope level or globally. Though event processors are similar to `beforeSend` and `beforeSendTransaction`, there are two key differences:
 
-- `beforeSend` is guaranteed to be run last, after all other event processors, which means it gets the final version of the event right before it's sent (hence the name). Event processors added with either of the methods below run in an undetermined order, which means changes to the event may still be made after the event processor runs.
-- While `beforeSend` and processors added with `Sentry.addGlobalEventProcessor` run globally, regardless of scope, processors added with `scope.addEventProcessor` only run on events captured while that scope is active.
+- `beforeSend` and `beforeSendTransaction` are guaranteed to be run last, after all other event processors, which means they gets the final version of the event right before it's sent (hence the name). Event processors added with either of the methods below run in an undetermined order, which means changes to the event may still be made after the event processor runs.
```
lizokm, line 27:
```suggestion
- `beforeSend` and `beforeSendTransaction` are guaranteed to be run last, after all other event processors, which means they get the final version of the event right before it's sent (hence the name). Event processors added with either of the methods below run in an undetermined order, which means changes to the event may still be made after the event processor runs.
```

lizokm (approved):
Woot! Looking good.

smeubank:
https://sentry-docs-git-kmclb-add-beforesendtransaction-to-js-options.sentry.dev/platforms/javascript/configuration/options/#before-send-transaction

src/docs/product/accounts/quotas/manage-event-stream-guide.mdx
lobsterkatie: I disagree. The fact that transactions don't go through `beforeSend` is a perennial source of confusion for users, which is fair given that a) it's not called `beforeSendError`, and b) we have a bad habit of using "event" to sometimes mean "error event" and sometimes mean "error or transaction event," a holdover from the days when there were no transaction events, so it's easy to mess up which things apply only to errors and which things apply to all events. 

Also, this is 327 lines in, so if you got here by searching `beforeSend`, or a following a link, you easily might not realize what page you're on, especially since the lefthand sidebar doesn't automatically scroll to show the highlighted section. Check out the screenshot below, where there's only the faintest indication that what we're looking at is specific to error events:

image

So I think it can't hurt to be extra clear on this point.

src/docs/product/accounts/quotas/manage-transaction-quota.mdx
lobsterkatie: I have to assume so? IDK, I just went with it, because fixing it really should be a separate PR if we're going to do it.

src/docs/product/accounts/quotas/manage-transaction-quota.mdx
lobsterkatie: And same answer. 🙂

src/platforms/common/enriching-events/event-processors.mdx
lobsterkatie: Yikes. Good catch, thanks.

lobsterkatie:
Thanks for the suggestions, @imatwawana  and @lizokm! I took most of them, and @imatwawana, I tried to answer your questions above. In order to unblock folks working on `beforeSendTransaction` in other SDKs, I'm going to merge this, but if we need to come back to anything in a future PR, we totally can.

P.S. Just for the record (since it came up again in this PR), in hopes of preventing us from going back and forth every time I use the phrase "Note that xxxx" (which I'll admit I perhaps do too frequently): 

No disrespect intended, but you can't note-comma-that. It's just not grammatical. Either "note" is an interjection (akin to "look" or "see" in the phrases "look, here's the thing" or "see, that's the problem"), in which case it gets a comma, or it's a prepositional verb (akin to "look" or "see" in the phrases "look at the sky" or "see that you do"), in which case it doesn't. But if it's an interjection, you have to be able to get rid of it and still have a full sentence after you've done so. Removing "Note," from, "Note, that xyz is true" just leaves you with "that xyz is true," which clearly isn't a fully-formed sentence. So we can argue the stylistic merits of "Note, xyz is true" vs. "Note that xyz is true" (I gravitate towards the latter, but am open to switching to the former if you think it fits the overall docs style better), but the hybrid doesn't work.

(Other than that, though, I really do appreciate the help tightening up my prose. My brain thinks in many layers of parentheses, and I try to keep it out of my writing, but I know I don't always succeed. So keep 'em coming!)

PR #5815, opened 2022-11-19
fix(quota): Remove `sampleRate` section in transaction quota guide

lobsterkatie:
This removes the SDK Sample Rate section from the guide on transaction quota management, because `sampleRate` doesn't apply to transactions, and `tracesSampleRate` is covered by the next heading, SDK Tracing options.

### DIFF
--- a/src/docs/product/accounts/quotas/index.mdx
+++ b/src/docs/product/accounts/quotas/index.mdx
@@ -86,7 +86,7 @@ If an inbound filter is applied for a type of error, transaction, or attachment,
 
 ### 6. SDK Sample Rate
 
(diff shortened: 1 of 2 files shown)


AbhiPrasad (approved):

imatwawana:
So your edit for the Transactions quota mgmt page totally makes sense, but it affects [this page](https://docs.sentry.io/product/accounts/quotas/) as well. I'm not sure what the solution is. Maybe instead of Update SDK Sample Rate, two items called Update Error Sample Rate and Update Transaction Sample Rate?

lobsterkatie:
> So your edit for the Transactions quota mgmt page totally makes sense, but it affects [this page](https://docs.sentry.io/product/accounts/quotas/) as well. I'm not sure what the solution is. Maybe instead of Update SDK Sample Rate, two items called Update Error Sample Rate and Update Transaction Sample Rate?

Good point. I'm surprised it didn't fail the link checker.

In any case, I think I fixed it. LMK what you think. (TL;DR, I mentioned sample rate in the section on tracing config, and changed the link to point there.)

imatwawana:
> > So your edit for the Transactions quota mgmt page totally makes sense, but it affects [this page](https://docs.sentry.io/product/accounts/quotas/) as well. I'm not sure what the solution is. Maybe instead of Update SDK Sample Rate, two items called Update Error Sample Rate and Update Transaction Sample Rate?
> 
> Good point. I'm surprised it didn't fail the link checker.
> 
> In any case, I think I fixed it. LMK what you think. (TL;DR, I mentioned sample rate in the section on tracing config, and changed the link to point there.)

Works for me! I suspect the link checker doesn't catch that because it can still resolve to a page even if it can't find the heading.

Issue #5718, opened 2022-11-04
[sentry-unplugin] Update Source Maps Docs for newly supported JS Bundler


smeubank:
### Core or SDK?

Platform/SDK

### Which part? Which one?

sentry-javascript

### Description

JavaScript SDK team is working on a new feature which will support 3 new JS bundlers in the uploading of Source Maps to Sentry. ideally with same feature parity of what is supported through webpack today.

More information can be found here: https://github.com/getsentry/sentry-javascript-bundler-plugins

Today via sentry-webpack-plugin we only support webpack, and for other JS Bundlers we direct people down an advnaced usage path, by setting everything up manually with sentry-cli.

### Suggested Solution

Docs shall be updated to include guides for Vite, esbuild, and Rollup

Docs to be updated: (multiple issues can be created from this one)

imatwawana:
@Lms24 - I'm concerned about making the wizard pages significantly longer. How much more content would that entail? All of the content on the upload pages that I'm seeing on #5799? Or less? Or more?

Lms24:
@imatwawana good point. This probably depends a little on the JS framework. Ideally, I think it would be just one more step, showing how to set up bundler plugins. This however doesn't work in all frameworks (e.g. Angular, Ember?, Remix?). So for the frameworks where it doesn't work, I guess our best strategy would be to link to Sentry-CLI/Release action instructions. Which I'm realizing is far from ideal.

imatwawana:
What about only doing it for the bundler plugins and if there isn't an ideal strategy, not adding the step?

PR #5819, opened 2022-11-21
Make source maps troubleshooting page bundler-agnostic

lforst:
This PR checks and adjusts all places in the source maps troubleshooting docs that assume our webpack plugin is the only plugin, which is not the case anymore since we support multiple bundlers now: https://github.com/getsentry/sentry-javascript-bundler-plugins

Also, adjusts some random wrong backticks.

Ref https://github.com/getsentry/sentry-docs/issues/5718

### DIFF
--- a/src/platforms/javascript/common/sourcemaps/troubleshooting_js.mdx
+++ b/src/platforms/javascript/common/sourcemaps/troubleshooting_js.mdx
@@ -6,7 +6,7 @@ description: "Troubleshooting for source maps."
 sidebar_order: 8
 ---
 
-Source maps can sometimes be tricky to get going. If you’re having trouble, you can try using our verification tool inside `sentry-cli`, or go through the steps listed below yourself. Please keep in mind, that the `sentry-cli` is a work-in-progress, and we’re actively working to improve it.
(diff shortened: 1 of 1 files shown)


src/platforms/javascript/common/sourcemaps/troubleshooting_js.mdx
```diff
@@ -185,33 +185,33 @@ Alternatively, if you are using Sentry CLI to upload source maps to Sentry, you
...
   sourceMap = require("source-map");
 
-// file output by Webpack, Uglify, and so forth
+// Path to file that is generated by your build tool (webpack, tsc, ...)
 var GENERATED_FILE = path.join(".", "app.min.js.map");
 
-// line and column located in your generated file (for example, the source of your error
-// from your minified file)
+// Line and column located in your generated file (for example, the source
+// of error from your minified file)
```
Lms24, line 207: l: not sure if that's correct either...

```suggestion
// Line and column located in your generated file (for example, the source
// of the error from your minified file)
```

Lms24 (approved):

PR #5828, opened 2022-11-21
Enable node and electron to custom perf metrics docs

AbhiPrasad:
Node and Electron both work since it is enabled for JavaScript.


AbhiPrasad:
with https://sentry-docs-git-node-electron-custom-perf-metrics.sentry.dev/platforms/javascript/guides/electron/performance/instrumentation/performance-metrics/ and https://sentry-docs-git-node-electron-custom-perf-metrics.sentry.dev/platforms/node/performance/instrumentation/performance-metrics/, seems to work fine!

PR #5832, opened 2022-11-22
Add RN Screenshots

krystofwoldrich:
Docs for:
- https://github.com/getsentry/sentry-react-native/pull/2610


src/platforms/common/enriching-events/screenshots.mdx
```diff
@@ -25,6 +26,12 @@ Screenshots may contain <PlatformLink to="/data-management/sensitive-data/">PII<
 
 <PlatformContent includePath="enriching-events/attach-screenshots" />
 
+<PlatformSection supported={["react-native"]}>
+
+Screenshots are transfered from native SDKs to React Native over the Bridge. This can be a performance bottleneck on lower end devices. This happens only when capturing error event with the screenshots enabled. React Native New Architecture will improve the performance.
```
imatwawana, line 31: Can we rearrange it like so?

Also, it doesn't look like bridge needs to be capitalized based on these docs [here](https://reactnative.dev/docs/native-modules-ios#reserved-method-names). But we can also get around this by using 
```suggestion
Screenshots are transferred from native SDKs to React Native over the `Bridge`, and this happens only when an error event is captured with screenshots enabled. This can negatively impact performance on lower end devices, however, React Native New Architecture will improve performance.
```

imatwawana (commented):
It seems like the two parts about performance are separated when they should be together. I've made a suggestion, but if this doesn't work, we can figure something else out.

src/platforms/common/enriching-events/screenshots.mdx
krystofwoldrich, line 31: Thank you for the comment, how does this sound?

```suggestion
When `attachScreenshot` is enabled and an error event is captured the screenshots are transferred from native SDKs to React Native over the bridge. This can negatively impact performance on lower-end devices, however, React Native New Architecture will improve the performance of the transfer as the data are not serialized.
```

imatwawana: That's much better. A couple needed commas, but otherwise this looks good:

```suggestion
When `attachScreenshot` is enabled and an error event is captured, the screenshots are transferred from native SDKs to React Native over the bridge. This can negatively impact performance on lower-end devices, however, React Native New Architecture will improve the performance of the transfer, as the data are not serialized.
```

imatwawana (approved):

src/platforms/common/enriching-events/screenshots.mdx
krystofwoldrich: Thanks.

marandaneto (approved):
Approving it but let's release the new version before merging this.
Thanks @krystofwoldrich

PR #5833, opened 2022-11-22
add key transaction section

andriisoldatenko:


src/docs/product/data-management-settings/dynamic-sampling/index.mdx
```diff
@@ -43,6 +44,10 @@ While prioritizing data from the latest release, we also take into consideration
 
 We prioritize data from development environments because these environments typically generate a relatively small number of transactions compared to your production environments. As a result, to gain meaningful insights during your testing phase, we prioritize transaction data from these environments.
 
+### Key Transactions
+
+We prioritize my Key Transactions to 5x times from blended rate. This rule applies to single transactions.
```
imatwawana, line 49: I have a few questions: 

1. Do you mean _5x times_ or _5 times_? 
2. What do we mean by "blended rate"? Is this something users should already know?
3. Why do we mean by this applying to single transactions. Maybe this is obvious to a developer, but I don't understand what people might think it would apply to instead? An entire trace? 

```suggestion
We prioritize key, or starred, transactions up to five times over the blended rate. This rule applies to single transactions.
```

imatwawana (commented):
I think we need to clarify a few things before we publish this. I've left some questions.

src/docs/product/data-management-settings/dynamic-sampling/index.mdx
andriisoldatenko: yeah i agree let me rework it. Thanks for review!

iambriccardo: I would personally phrase it similarly to the other descriptions. Remember that as @jan-auer correctly says, we always need to think about the problem we are solving for customers, therefore I would add to the description the rationale behind this feature.

iambriccardo, line 49:
```suggestion
For most projects there are only a handful or transactions that you really care about and we purposefully designed the starring feature to give you a visual prioritization of them. With dynamic sampling, starred transactions are even more useful because we will prioritize them up to five times over the blended rate. This rule applies to single transactions.
```

iambriccardo (changes requested):

src/docs/product/data-management-settings/dynamic-sampling/index.mdx
iambriccardo: What do you think @imatwawana?

andriisoldatenko: thanks!

iambriccardo (approved):

src/docs/product/data-management-settings/dynamic-sampling/index.mdx
```diff
@@ -43,6 +44,10 @@ While prioritizing data from the latest release, we also take into consideration
 
 We prioritize data from development environments because these environments typically generate a relatively small number of transactions compared to your production environments. As a result, to gain meaningful insights during your testing phase, we prioritize transaction data from these environments.
 
+### Key Transactions
+
+For most projects there are only a handful of transactions that you really care about. We prioritize key, or starred, transactions over others, so you can have more visibility into your important data, rather than sample all transactions in equivalent manner.
```
imatwawana, line 49: This is much clearer. The sentence gets too long with that last clause and it's kind of implied anyway by the fact that we're saying we prioritize them, so I removed it. Also, I added the word "typically" so that it leaves room for the fact that this might not be the case for some folks.

```suggestion
Typically, for most projects there are only a few transactions that you really care about. We prioritize key, or starred, transactions over others, so you can have more visibility into your important data.
```

imatwawana (approved):
Just made some wording tweaks, but otherwise, this looks good!

src/docs/product/data-management-settings/dynamic-sampling/index.mdx
```diff
@@ -43,6 +44,10 @@ While prioritizing data from the latest release, we also take into consideration
 
 We prioritize data from development environments because these environments typically generate a relatively small number of transactions compared to your production environments. As a result, to gain meaningful insights during your testing phase, we prioritize transaction data from these environments.
 
+### Key Transactions
+
+Typically, for most projects there are only a few transactions that you really care about. We prioritize key, or starred, transactions over others, so you can have more visibility into your important data.
```
priscilawebdev, line 49: "we prioritize key" is not the same as "starred"? I feel like we could like the section where we explain what is a starred transaction here

priscilawebdev:
can we please update the screenshot? 

image


the new one has the "health checks" highlighted and I think we don't want that

src/docs/product/data-management-settings/dynamic-sampling/index.mdx
andriisoldatenko: AFAIK "starred" means -> 
image

priscilawebdev:
here is the link for the "starring key transaction" https://docs.sentry.io/product/performance/transaction-summary/#starring-key-transactions

src/docs/product/data-management-settings/dynamic-sampling/index.mdx
priscilawebdev: yes  "starred" is clear to me but what does mean "we prioritized by _key **or** starred_"?

andriisoldatenko: @ale-cota can you please help here?

ale-cota:
How about this then, @priscilawebdev, @andriisoldatenko, @imatwawana?  

Typically, for most projects there are only a few transactions that you really care about. We prioritize [key transactions](https://docs.sentry.io/product/performance/transaction-summary/#starring-key-transactions) over others, so you can have more visibility into your important data.

priscilawebdev (approved):
lgtm

src/docs/product/data-management-settings/dynamic-sampling/index.mdx
iambriccardo: I think key transactions is the internal term, whereas starred is the external one.

Issue #5811, opened 2022-11-18
[sdk config] Clean up filtering config docs (`beforeSend` and hints)


lobsterkatie:
### Core or SDK?

Platform/SDK

### Which part? Which one?

all of them

### Description

It started out as a quick edit - it always does - and now it's turned into actually fixing the problems with the [filtering](https://docs.sentry.io/platforms/javascript/configuration/filtering/) docs. Writing this up here because it's now bigger than me and not a project I can fully take on at the moment, but I've done some of the work and don't want it to be wasted. Maybe we can crowd-source the parts that still need doing, as they mostly have to do with figuring out and writing down which SDKs pass which properties in the hint, and updating code snippets in various languages.

Problems:
- Hint text is repetitive, including having the same example twice (at least in words - see below).
- Hint text is split between a subheading (under Using `beforeSend` and its own heading, Using Hints)
- The `before-send-hint` and `before-send-fingerprint` snippets are introduced with the same words, and in some languages they have the same contents. In other languages, one has an example that matches the words (fingerprinting) and one has another example of using the hint.
- The list of things the hints contain is both missing things and, for many languages, includes things which don't apply.
- Some languages list the hint properties on the main filtering page, whereas at least one language (Python) splits them between the main filtering page and a Hints sub-page.

### Suggested Solution

Fixes:

- [x] Consolidate hint text and clean up repetition
- [x] Move list of hint contents to subpage. The page is long enough, even without adding whatever properties are missing.
- [x] Use includes for listing both event and breadcrumb hint contents. (The latter already has an include folder, but it only has JS in it.)
- [ ] Fix all the links to whatever's been moved around.
- [ ] Collect hint contents lists for both `beforeSend` and breadcrumbs hints for all platforms which support hints. `beforeSend` hint lists should note whether each property is available in message events, error events, and transaction events. See `before-send-hint-contents/javascript.mdx` in the draft PR below for one way to organize this. (Not wedded to that way - could break it up for each kind of event, but that might lead to a lot of repetition.)
- [ ] Decide whether it's better for languages whose data hasn't yet been collected to have nothing listed, or have JS hint contents listed.
- [ ] Update the `before-send-hint` snippets for the languages where the contents there is the same as in the `before-send-fingerprint` snippet, so that the former shows an event being dropped because of something in the hint (just to match the ones which are already different).
- [x] Rewrite text introducing the non-database example to match what the example is actually doing.

[...]

PR #5838, opened 2022-11-22
feat(sdks): Add `beforeSendTransaction` to filtering page

lobsterkatie:
Note: Given the [on-going work](https://github.com/getsentry/sentry-docs/pull/5804) on the Filtering page, I'm splitting this off of [the main PR adding `beforeSendTransaction` to the docs](https://github.com/getsentry/sentry-docs/pull/5814). I'll update this as the filtering page work progresses, but leaving it in draft state for now. (If you're curious about what this PR will actually do in and of itself, look at the second commit below. Not expecting any reviews at this point, though.)

-------

This adds the new `beforeSendTransaction` option to the Filtering page.

### DIFF
--- a/src/platform-includes/configuration/before-send-hint/javascript.mdx
+++ b/src/platform-includes/configuration/before-send-hint/javascript.mdx
@@ -3,17 +3,13 @@ Sentry.init({
   // ...
 
   beforeSend(event, hint) {
-    const error = hint.originalException;
     if (
-      error &&
-      error.message &&
-      error.message.match(/database unavailable/i)
+      hint.originalException instanceof DOMException &&
+      hint.originalException.name === "NotSupportedError"
     ) {
-      event.fingerprint = ["database-unavailable"];
+      return null;
     }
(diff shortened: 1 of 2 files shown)


cleptric:
@lobsterkatie What's the status here?

lobsterkatie:
> @lobsterkatie What's the status here?

It's currently waiting on https://github.com/getsentry/sentry-docs/pull/5804, which is actually done save dealing with language-specific snippets and hint properties (see https://github.com/getsentry/sentry-docs/issues/5811). I'm about to go on sabbatical, so cc @lizokm @imatwawana to help decide what the best path forward is. I think with not too much work, we could get the wordsmithing part of https://github.com/getsentry/sentry-docs/pull/5804 merged (assuming they seem like good changes), which would then unblock this PR, and chase the snippets/properties separately. Thoughts?

imatwawana:
> #5804

I think that's fine. I'll review #5804 today so we can unstick that.

PR #5642, opened 2022-10-13
Add a reminder on the supported languages for Stack Trace Linking

vkedwardli:
https://github.com/getsentry/sentry/issues/35608

Since Stack Trace Linking is not supported in languages except JS and Python? Adding a reminder so that others won't spent days to find out why it is not working

### Legal Boilerplate

Look, I get it. The entity doing business as "Sentry" was incorporated in the State of Delaware in 2015 as Functional Software, Inc. and is gonna need some rights from me in order to utilize my contributions in this here PR. So here's the deal: I retain all rights, title and interest in and to my contributions, and by keeping this boilerplate intact I confirm that Sentry can use, modify, copy, and redistribute my contributions, under Sentry's choice of terms.

PR #5860, opened 2022-11-28
Update list of platforms that support stacktrace linking

armenzg:
This is to update the list added in #5642.


armenzg:
@imatwawana this is unrelated to our conversation on Slack

brianthi (approved):

brianthi:
LGTM, I took a quick look for other docs which reference this feature and couldn't find anywhere else worth calling out.

PR #5864, opened 2022-11-30
Add support for PHP to beforeSendTransaction

cleptric:


lobsterkatie:
Man, I'm totally raining on your parade today, but can we separate the "adding PHP support" part of this from the "editing the filtering page" part? The latter is already 90% addressed in https://github.com/getsentry/sentry-docs/pull/5804.

cleptric:
Yes we can

lobsterkatie (approved):
I can't speak to the accuracy of the PHP code, but otherwise LGTM!

PR #5865, opened 2022-11-30
Remove platform specific “New in”

cleptric:
This approach doesn't scale well, as this page is platform agnostic, hence I removed it.

cc @lobsterkatie

### DIFF
--- a/src/platforms/common/configuration/options.mdx
+++ b/src/platforms/common/configuration/options.mdx
@@ -571,12 +571,6 @@ This function is called with an SDK-specific message or error event object, and 
 
 This function is called with an SDK-specific transaction event object, and can return a modified transaction event object, or `null` to skip reporting the event. This can be used, for instance, for manual PII stripping before sending.
 
-<PlatformSection supported={["javascript", "node"]}>
-
-_(New in version 7.18.0)_
-
-</PlatformSection>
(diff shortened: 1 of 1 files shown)


lizokm (approved):
Thanks for making this edit.

lobsterkatie:
I don't actually think we should do this.

I 100% agree about the scalability of our current "new in" solution - it's pretty awful. But the fact remains that it's what we've done for other options (albeit inconsistently). See [the `sendClientReports` entry](https://github.com/getsentry/sentry-docs/blob/cb27660f0537ead54f6ac20f80f9efd32565f0f5/src/platforms/common/configuration/options.mdx?plain=1#L501-L523), for example. And until we version our docs (which IMHO we should do at some point), I do think it's really helpful for people to know when a certain feature was introduced.

I have a half-baked idea for fixing this, though. Would need to play around to see if it would work, but regardless, if we do make a change here, I think it should be to change our solution for all of the options, rather than to just remove this particular "new in" entry.

lizokm:
Let's pause on this and talk more about it tomorrow when we meet @lobsterkatie!

lobsterkatie:
Sounds good. Come to think of it, I actually might have two half-baked ideas...

PR #5870, opened 2022-11-30
Implement membership check before routing logic

hubertdeng123:
This implements a membership check so that the routing logic doesn't happen for issues created by Sentry members


ethanhs (approved):
LGTM

PR #5799, opened 2022-11-17
Restructure Source Maps Upload page

lforst:
Ref: https://github.com/getsentry/sentry-docs/issues/5784

First, this PR adds guides on how to upload source maps with additional bundlers ([Rollup](https://rollupjs.org/guide/en/), [Vite](https://vitejs.dev/), [esbuild](https://esbuild.github.io/)) in combination with our new [bundler plugins](https://github.com/getsentry/sentry-javascript-bundler-plugins).

Additionally, this PR changes our docs flow for source maps from a 2-step guide ("generating sourcemaps" and "uploading sourcemaps") to full guides for each individual bundler we have solutions for. This should make it easier for people to follow the guide because it involves way less navigating through various pages.

Comprehensive list of changes:
- Shorten source maps primer page
- Add image that should describe what source maps will do to primer page
- Add links to guides on primer page
- Remove "Generating Source Maps" pages
- Move all information that was contained in "Generating Source Maps" to other relevant sections (either to "Troubleshooting Sourcemaps" or to the respective guides)

PR #5872, opened 2022-12-01
Add back note about the source-map-support package

lforst:
Somewhere during https://github.com/getsentry/sentry-docs/pull/5799 we dropped a note about the `source-map-support` package.

This PR adds that note back in the source maps troubleshooting section.

### DIFF
--- a/src/platform-includes/sourcemaps/troubleshooting/javascript.mdx
+++ b/src/platform-includes/sourcemaps/troubleshooting/javascript.mdx
@@ -228,3 +228,7 @@ Sometimes build scripts and plugins produce pre-compressed minified files (for e
 ## Verify workers are sharing the same volume as web (if running self-hosted Sentry via Docker)
 
 Sentry does source map calculation in its workers. This means the workers need access to the files uploaded through the front end. Double check that the cron workers and web workers can read/write files from the same disk.
+
+## Verify you are not using the `source-map-support` package
+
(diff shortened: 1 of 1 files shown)


src/platform-includes/sourcemaps/troubleshooting/javascript.mdx
```diff
@@ -228,3 +228,7 @@ Sometimes build scripts and plugins produce pre-compressed minified files (for e
 ## Verify workers are sharing the same volume as web (if running self-hosted Sentry via Docker)
 
 Sentry does source map calculation in its workers. This means the workers need access to the files uploaded through the front end. Double check that the cron workers and web workers can read/write files from the same disk.
+
+## Verify you are not using the `source-map-support` package
+
+To rely on Sentry's source map resolution, your code cannot use the [source-map-support](https://www.npmjs.com/package/source-map-support) package. That package overwrites the captured stack trace in a way that prevents our processors from correctly parsing it.
```
kamilogorek, line 234: It'd be nice to add a note regarding possible `ContextLines` workaround, but we'd need to verify it's working as expected first.

kamilogorek (approved):

src/platform-includes/sourcemaps/troubleshooting/javascript.mdx
lforst: Ok so I tried this out. `ContextLines` + `source-map-support` works surprisingly well. I'll slightly adjust this section to say that `source-map-support` will mess up processing *if* they have uploaded source maps for their project.

kamilogorek (approved):

PR #5874, opened 2022-12-01
Update Svelte source maps docs to refer to new bundler plugins

lforst:
Updates the Svelte source maps docs to refer to the new vite bundler plugin. Also removes tsc as a recommended tool and adds a section that was accidentally removed.


Lms24 (approved):
Nice!

PR #5875, opened 2022-12-01
Add source maps step to getting started pages

lforst:
Adds a section to the JavaScript/Node.js "Getting Started" pages to remind people to set up source maps.


lforst:
Screen Shot 2022-12-01 at 15 33 19

@smeubank Is this what we want or something more in-depth?

src/platform-includes/getting-started-sourcemaps/javascript.mdx
```diff
@@ -0,0 +1,7 @@
+## Get Readable Stack Traces in your Sentry Errors
```
imatwawana, line 1: Can we say this more succinctly? This sounds very much like we're marketing to the user which we want to avoid in the docs. I've made a suggestion, but I'm not sure if that changes the meaning at all. 
```suggestion
## Add Readable Stack Traces to Errors
```

src/platform-includes/getting-started-sourcemaps/javascript.mdx
```diff
@@ -0,0 +1,7 @@
...
+
+You're not quite done yet!
```
imatwawana, line 3:
```suggestion
```

imatwawana (commented):
I'm always hesitant about making changes to the Getting Started page, so I want to give this change some thought, but I'm wondering, is there consensus on your team about adding this here? What's the thinking behind adding this content (vs adding anything else to this page)?

For now, I've only made edits on one of the includes; once we nail that one down (and decide that this content should definitely be added to this page), then you can just apply the same changes to the second include.

lforst:
> I'm always hesitant about making changes to the Getting Started page, so I want to give this change some thought, but I'm wondering,

Right there with you. I was actually procrastinating on this change for a week because I was thinking that the getting started pages are so important and adding something there without careful consideration is maybe not the best thing.

> is there consensus on your team about adding this here?

There is unanimous consensus within the team. We had some discussions around what *kind* of content we should add on source maps (just a section with a link vs. framework-specific code examples) but everybody agreed that we should definitely have some section about source maps here since we consider it a big part of the user experience and it would be a shame if people didn't set it up. We are aware of the complexity that this section adds to the getting started pages - maybe it's even a bit too much. We can always remove it later I guess.

> What's the thinking behind adding this content (vs adding anything else to this page)?

As mentioned above I think everybody is aware that the getting started pages are super important. I feel like we very rarely want to add to this page (if at all) unless we consider something important enough. We think source maps are important enough to the user experience that we should add them here. If we ever want to add something else to this page we'll probably have the same introspective discussion again if it is important enough. I doubt that we will add anything here soon though.

> For now, I've only made edits on one of the includes; once we nail that one down (and decide that this content should definitely be added to this page), then you can just apply the same changes to the second include.

Perfect! Thank you so much!

imatwawana:
Okay that makes sense to me. Thank you for explaining more fully. Let's go ahead with these changes. I'm thinking that this change should be made in the wizard as well. That'll probably give us a better indication of how well it's working.

lforst:
Sorry for the pings - was by accident. For some reason my brain sometimes does `git pull master` instead of `git merge master` 🤦

imatwawana:
Don't forget to update the wizard as well!

lforst:
> Don't forget to update the wizard as well!

It is something we're discussing right now. The growth, and telemetry experience teams are also on this.

PR #5892, opened 2022-12-05
fix: Code samples for react components

HazAT:
When going through the process of instrumenting Sentry on Javascript (React as an example) I uncovered this small annoyance.
To verify if Sentry is instrumented you’re asked to intentionally create an error with this snippet of code.
`return <button onClick={methodDoesNotExist}>Break the world</button>;`
The Problem here is that, you get a blank white screen by adding this `-> onClick={methodDoesNotExist}` because your in fact not declaring anything. I know a developer would know this, but I don’t think we should rely on people reading the snippet of code to get the intended value.


Jesse-Box (approved):

vladanpaunovic (approved):
Nice catch @HazAT !

Issue #5894, opened 2022-12-05
Add Basic Replay Integration documentation


Lms24:
### Core or SDK?

Platform/SDK

### Which part? Which one?

JS Browser SDKs

### Description

As a first step towards having proper documentation for Sentry Replay, we should provide a small code snippet in the "Pluggable Integrations" section of the JS SDK documentation to document how to set up the Replay integration. Surely, to document the entire Replay feature we'll want more documentation but for now, this should do. Also, it adds instructions on how to use Replay as a CDN bundle, which is something, we haven't yet documented properly. 

### Suggested Solution

Add an entry to "Pluggable Integrations" analogously to how we document other pluggable integrations

Lms24:
Closing as we now have [proper docs](https://docs.sentry.io/platforms/javascript/session-replay/) for Replay.

PR #5899, opened 2022-12-06
Initial Session Replay SDK Docs

ryan953:
Resolves https://github.com/getsentry/sentry-replay/issues/174

Relates to https://github.com/getsentry/sentry-docs/issues/5894
Relates to https://github.com/getsentry/sentry-docs/issues/5903


src/pages/index.tsx
```diff
@@ -211,6 +213,11 @@ const IndexPage = () => {
                   Alerts & Notifications
                 </a>
               </li>
+              <li>
+                <a href="/product/session-replay/">
+                  Session Replay
+                </a>
+              </li>
```
ryan953, line 220: New footer link on the homepage: http://localhost:3000/

SCR-20221205-no0

src/platform-includes/performance/configure-sample-rate/javascript.electron.mdx
```diff
@@ -16,7 +16,7 @@ Sentry.init({
   integrations: [new BrowserTracing()],
 
   // To set a uniform sample rate
-  tracesSampleRate: 0.2
+  tracesSampleRate: 0.2,
```
ryan953, line 19: perf examples are messed up, will pull this out to it's own PR though

src/docs/product/session-replay/index.mdx
```diff
@@ -0,0 +1,19 @@
+---
+title: "Session Replay"
+sidebar_order: 3
+redirect_from:
+  - /session-replay/
+description: "Session Replay. TODO"
+---
```
ryan953, line 7: New item in the sidebar of the Product section:
SCR-20221205-nph

src/docs/product/session-replay/getting-started.mdx
```diff
@@ -0,0 +1,28 @@
+---
+title: Set Up
+sidebar_order: 1
+redirect_from:
+  - /session-replay/setup/
+  - /session-replay/getting-started/
+description: "Get started with Session Replay, TODO."
+---
```
ryan953, line 8: Child page in the Product section:

SCR-20221205-nrc

src/components/platformSidebar.tsx
```diff
@@ -90,6 +91,15 @@ export const PlatformSidebar = ({
...
       />
+      <DynamicNav
+        root={`/${pathRoot}/session-replay`}
+        title="Session Replay"
+        prependLinks={[
+          [`/${pathRoot}/session-replay/`, "Set Up Session Replay"],
+        ]}
+        suppressMissing
+        tree={tree}
+      />
```
ryan953, line 102: SDK nav: 
(the exclude list above means that the link will not appear (again) under the "Sentry for JavaScript" heading)

SCR-20221205-nsp

src/pages/index.tsx
bruno-garcia: I wonder if we can move that to a more 'prime location'.
How about we swap it with `Sentry-CLI` 😏 
Or another arrangement that makes more sense?

src/docs/product/session-replay/getting-started.mdx
```diff
@@ -0,0 +1,28 @@
...
+title: Set Up
+sidebar_order: 1
+redirect_from:
+  - /session-replay/setup/
+  - /session-replay/getting-started/
```
bruno-garcia, line 6: To replace: https://github.com/getsentry/sentry-docs/blob/7197fd3dea0a7ffc92a0d283ebd147ad76356547/src/platforms/javascript/common/configuration/integrations/rrweb.mdx?plain=1#L6

Would require removing this page from the docs on this PR: https://github.com/getsentry/sentry-docs/blob/7197fd3dea0a7ffc92a0d283ebd147ad76356547/src/platforms/javascript/common/configuration/integrations/rrweb.mdx?plain=1#L6

`/platforms/javascript/configuration/integrations/rrweb/` being the current path
and `/platforms/javascript/integrations/rrweb/` the old one
```suggestion
  - /session-replay/getting-started/
  - /platforms/javascript/integrations/rrweb/
  - /platforms/javascript/configuration/integrations/rrweb/
```

src/docs/product/session-replay/getting-started.mdx
bruno-garcia: I imagine Set Up is mainly a client-side/SDK thing? If so, we could probably get by without the sub page to start? Basically add all content on the landing Session Replay page for now

src/platforms/common/session-replay/index.mdx
```diff
@@ -0,0 +1,46 @@
+---
+title: Set Up Session Replay
```
bruno-garcia, line 2: Since this content will be exclusively for JavaScript, I wonder if the best location is under:

`src/platforms/javascript/common` instead of `src/platforms/common`. The latter causes inheritance to all platforms

src/platform-includes/session-replay/enable-session-replay/javascript.mdx
```diff
@@ -0,0 +1,22 @@
...
+npm install --save @sentry/replay
+```
+
+```html {tabTitle: CDN}
+<script
+  <!--
+    Note that `bundle.replay.min.js` contains both `@sentry/browser` AND
+    `@sentry/replay`, and should therefore be used in place of
+    `@sentry/browser`'s bundle rather than in addition to it.
+  -->
```
Lms24, line 17: This bundle doesn't exist and the bundle that's going to be released is an addon bundle, containing **only** the replay integration. So to install the SDK plus Replay, users have to add the follwing bundles:

* `bundle(.tracing)?(.min)?.js`
* `replay(.min)?.js`

So, the setup is identical to how to use our other [pluggable integrations](https://docs.sentry.io/platforms/javascript/configuration/integrations/plugin/#extraerrordata).

I was going to add instructions to the README today. I can also update this example if you want.

src/docs/product/session-replay/getting-started.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+Session Replay is currently in beta. Beta features are still in-progress and may have bugs. We recognize the irony. If you have any questions or feedback, please email us at [session-replay@sentry.io](mailto:session-replay@sentry.io).
+
+</Note>
+
+To set up the session replay integration, use the links for supported SDKs below. TODO.
+
+## Supported SDKs
+
+- JavaScript
```
Lms24, line 20: We should add the plain Browser SDK (@sentry/browser) here for users who don't rely on a framework

src/platform-includes/session-replay/configure-sample-rate/javascript.mdx
```diff
@@ -0,0 +1,19 @@
...
+import * as Sentry from "@sentry/browser";
+
+import { Replay } from "@sentry/replay";
+
+Sentry.init({
+  dsn: "___PUBLIC_DSN___",
+
+  // This enables automatic instrumentation (highly recommended), but is not
+  // necessary for purely manual usage
+  integrations: [new Replay()],
```
Lms24, line 13: We'll also need to add a CDN tab here, as the setup isn't 100% identical (the namespace for the integration is `Sentry.Integrations.Replay`). As already said below, I can add this in a follow-up PR later if you want.

src/platforms/common/session-replay/index.mdx
```diff
@@ -0,0 +1,46 @@
...
+  - javascript.react
+  - javascript.remix
+  - javascript.svelte
+  - javascript.vue
+description: "Learn how to enable session replay in your app if it is not already set up."
+---
+
+With [session replay](/product/session-replay/), TODO.
+
+<PlatformSection supported={["javascript", "javascript.vue", "java.spring", "react-native"]} notSupported={["java", "go", "android", "ruby", "python", "dotnet", "dart", "native"]}>
```
Lms24, line 19: You have way more context here but just out of curioisity, `java.spring` really supports Replay? Are they using the Browser SDK?

Also, do we need to explicilty add `javascript.vue` or is it automatically included by `javascript`?

src/platforms/common/session-replay/index.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+## Enable Session Replay
+
+<PlatformContent includePath="session-replay/enable-session-replay" />
+
+</PlatformSection>
+
+## Configure the Sample Rate
```
Lms24, line 27: I think we should change the order here a little in the sense that it's most important for users to understand how to set up Replay. So under "Enable Session Replay" I think we should already show both code examples (installing the package and the `Sentry.init` call). 

Below then, we can describe the impact of the sample rates and (maybe later on) the other configuration options. 

Side note (no relevant for this PR): I just saw that our docs for Performance setup have the same structure so we might wanna refactor it a bit there as well.

src/platforms/common/session-replay/index.mdx
```diff
@@ -0,0 +1,46 @@
...
+The two options can both be set. If a session is not captured by the <PlatformIdentifier name="replays-session-sample-rate" />, and an error happens, then the <PlatformIdentifier name="replays-on-error-sample-rate" /> will have a chance to start sampling.
+
+<PlatformContent includePath="session-replay/configure-sample-rate" />
+
+## Verify
+
+While you're testing, set <PlatformIdentifier name="replays-session-sample-rate" /> to `1.0`, as that ensures that every user session will be sent to Sentry.
+
+Once testing is complete, **we recommend lowering this value in production** by either lowering your <PlatformIdentifier name="replays-session-sample-rate" /> value. We still recommend keeping <PlatformIdentifier name="replays-on-error-sample-rate" /> set to `1.0`.
+
```
Lms24, line 43: Should we add another subsection here (e.g. `## Advanced Configuration` or something similar) and link to the Readme for the time being? This way, users can be made aware that there are additional configuration options.

Lms24 (commented):
LGTM so far, thanks for writing all of this up!

src/platform-includes/session-replay/configure-sample-rate/javascript.mdx
```diff
@@ -0,0 +1,19 @@
...
+// substitute its name for `@sentry/browser` here
+import * as Sentry from "@sentry/browser";
+
+import { Replay } from "@sentry/replay";
+
+Sentry.init({
+  dsn: "___PUBLIC_DSN___",
+
+  // This enables automatic instrumentation (highly recommended), but is not
+  // necessary for purely manual usage
```
Lms24, line 12:
```suggestion
```

I would honestly just delete this to keep it as simple as possible for the majority of our users. (see my comment about "Advanced Configuration", as we're describing more manual usage in the Readme).

src/docs/product/session-replay/getting-started.mdx
jas-kas: Agree that Set Up should in the SDK docs ^ Not in the Product Guide section

jas-kas: Ah just took a look at the "Performance Monitoring" section again, it looks like Ryan is following the pattern of what's done there. The "Set Up" just includes a list of links to the SDK docs and that makes sense to me. I wouldn't want to clutter product docs w/ installation instructions

src/platform-includes/session-replay/enable-session-replay/javascript.mdx
Lms24: Update - Here are the readme docs for the CDN bundle: https://github.com/getsentry/sentry-javascript/blob/master/packages/replay/MIGRATION.md

src/pages/index.tsx
ryan953: totally agree. this was an 'empty' spot in the grid, but the columns/order is hard-coded as html, so we can move things around.

Might need to get buy-in from some other team. Lets split this off into a separate change.

src/docs/product/session-replay/getting-started.mdx
ryan953: yup! copy+pasted from Perf

src/platforms/common/session-replay/index.mdx
ryan953: ah, this is all copy+paste that i missed. none of these are really supported

src/pages/index.tsx
ryan953: I'll rip this change out of this PR, replaced with : https://github.com/getsentry/sentry-docs/pull/5911

src/docs/product/session-replay/getting-started.mdx
ryan953: I can pull these Product pages out of this PR and do them in a separate PR

Based on https://github.com/getsentry/sentry-docs/issues/5903 we don't have a specific product pages drafted yet (maybe faq?)

src/platforms/common/session-replay/custom-instrumentation/privacy-configuration.mdx
```diff
@@ -0,0 +1,40 @@
...
+
+### Blocking
+Blocking replaces the element with a placeholder that has the same dimensions. The recording will show an empty space where the content was. In this example the relevant html code is: `<table data-sentry-block>...</table>`.
+![Blocking example](session-replay-blocking.png)
+
+
+### Ignoring
+Ignoring only applies to form inputs. Events will be ignored on the input element so that the replay does not show what occurs inside of the input. In the below example, notice how the results in the table below the input changes, but no text is visible in the input.
+
+![Ignoring Example](session-replay-ignore-input.mp4)
```
ryan953, line 22: This isn't playing, it was originally a `.mov` file, i renamed it because some googling suggested that... it didn't work

Next approach: convert it to gif

src/platforms/common/session-replay/index.mdx
ryan953: Moved. 

Native is on the roadmap, but it'll be sooo different that i don't think it'll be a problem (and we could move it back if we wanted)

src/platform-includes/session-replay/setup-session-replay/javascript.mdx
```diff
@@ -0,0 +1,52 @@
+To set up the integration, add the following to your Sentry initialization. Several options are supported and passable via the integration constructor. See the [configuration sections](/platforms/javascript/session-replay/custom-instrumentation/) for more details.
+
+```javascript {tabTitle: ESM}
+import * as Sentry from "@sentry/browser";
+import { Replay } from "@sentry/replay";
```
Lms24, line 5: Here it's exactly the same as above, the `@sentry/browser` import wouldn't be correct for Framework SDKs. WDYT if (for now) we just add a comment to use the framework SDK here and in a follow-up PR later, we add platform-specific code snippets? This is how we do it for the Performance monitoring docs.
```suggestion
```javascript {tabTitle: ESM}
// import Sentry from your framework SDK (e.g. @sentry/react) instead of @sentry/browser 

import { Replay } from "@sentry/replay";
```

src/platforms/javascript/common/session-replay/custom-instrumentation/index.mdx
```diff
@@ -0,0 +1,24 @@
+---
+title: Custom Instrumentation
```
Lms24, line 2: nit (feel free to ignore): Is "Custom Instrumentation" the correct title here? To me, the content here seems to be more around additional advanced configuration than doing custom things (maybe apart from the manual `Replay` start/stop section). No strong opinion here, though. We can also revisit this later.

Lms24 (approved):
Looks good to me! Thanks for addressing my previous feedback. I just have two more comments but nothing blocking.

Issue #5901, opened 2022-12-06
`sendDefaultPii` does not affect the collection of ip address.


larsqa:
### Core or SDK?

Platform/SDK

### Which part? Which one?

JavaScript related SDK

### Description

From [sentry-javascript issue #6375](https://github.com/getsentry/sentry-javascript/issues/6375), the `sendDefaultPii` option does not affect the collection of ip address, as opposed to what's written in the docs.

> That might require sendDefaultPii set to true in the SDK options.

On a personal note - I'm still convinced (based on tests) that `sendDefaultPii` allows me to show/hide the ip address of an event, in sentry dashboard.

### Suggested Solution

Update the documentation with what the `sendDefaultPii` actually affects.

PR #5902, opened 2022-12-06
Clarify how IPs are collected for JS SDKs

lforst:
Fixes https://github.com/getsentry/sentry-docs/issues/5901

### DIFF
--- a/src/platforms/common/enriching-events/identify-user.mdx
+++ b/src/platforms/common/enriching-events/identify-user.mdx
@@ -66,7 +66,8 @@ The user segment, for apps that divide users in user segments.
 
 ### `ip_address`
 
-The user's IP address. If the user is unauthenticated, Sentry uses the IP address as a unique identifier for the user. Sentry will attempt to pull this from the HTTP request data (`request.env.REMOTE_ADDR` field in JSON), if available. That might require <PlatformIdentifier name="send-default-pii" /> set to `true` in the SDK options.
+The user's IP address. If the user is unauthenticated, Sentry uses the IP address as a unique identifier for the user.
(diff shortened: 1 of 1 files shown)


src/platforms/common/enriching-events/identify-user.mdx
```diff
@@ -66,7 +66,8 @@ The user segment, for apps that divide users in user segments.
 
 ### `ip_address`
 
-The user's IP address. If the user is unauthenticated, Sentry uses the IP address as a unique identifier for the user. Sentry will attempt to pull this from the HTTP request data (`request.env.REMOTE_ADDR` field in JSON), if available. That might require <PlatformIdentifier name="send-default-pii" /> set to `true` in the SDK options.
+The user's IP address. If the user is unauthenticated, Sentry uses the IP address as a unique identifier for the user.
+Serverside SDKs that instrument incoming requests will attempt to pull the IP address from the HTTP request data (`request.env.REMOTE_ADDR` field in JSON), if available. <PlatformSection notSupported={["javascript", "node"]}>That might require <PlatformIdentifier name="send-default-pii" /> set to `true` in the SDK options.</PlatformSection><PlatformSection supported={["javascript", "node"]}>This requires you to set the `include.ip` option to `true` in the [`RequestData`](https://docs.sentry.io/platforms/node/guides/connect/configuration/integrations/default-integrations/#requestdata) integration.</PlatformSection>
```
AbhiPrasad, line 70: > This requires you to set the `include.ip` option to `true` in the [`RequestData`]

Doesn't this only apply to node?

AbhiPrasad: Hmm - I'd rather have it be the correct source of truth, this split is something we should solve as an aside (cc @lizokm)

lforst: Aight will update!

lforst: Should I remove the entire paragraph for node though?

AbhiPrasad: It's still technically true for node - the ip address will just be of the server then though. Let's keep it, maybe we can come back to it later.

lforst: Sorry, edited my previous comment. I meant the paragraph is technically not relevant for browser.

AbhiPrasad: Yes, I think that's a good idea.

AbhiPrasad (approved):

src/platforms/common/enriching-events/identify-user.mdx
lforst: Discussed in person: too much headache in the include-exclude logic so we'll leave it in.

PR #5913, opened 2022-12-07
replay: alternative structure

bruno-garcia:
Alternative structure. Comments inline


src/platforms/javascript/common/session-replay/configuration.mdx
```diff
@@ -19,35 +19,6 @@ notSupported:
...
-
-```javascript {tabTitle: CDN}
-const replay = new Sentry.Integrations.Replay(); // This will *NOT* begin recording replays
-
-replay.start(); // Start recording
-
-replay.stop(); // Stop recording
-```
-
 ## General Integration Configuration
```
bruno-garcia, line 22: Starting with the sampling options, which are the main ones/everyone will use

src/platforms/javascript/common/session-replay/configuration.mdx
```diff
@@ -74,6 +45,17 @@ The following options can be configured as options to the integration, in `new R
 | inlineStylesheet | boolean | `true`  | Should inline stylesheets used in the recording |
 | recordCanvas     | boolean | `false` | Should record `<canvas>` elements |
 
+## Identifying Users
```
bruno-garcia, line 48: Identifying user is an SDK specific thing. It'll improve the user experience for sure, so makes total sense to add here. I tried to word it a bit as "here are the actual docs, but here's a quick snippet to get you started"

src/platforms/javascript/common/session-replay/configuration.mdx
```diff
@@ -83,3 +65,20 @@ new Replay({
   // any further configuration here is passed directly to rrweb
 });
 ```
+
+## Start and Stop Recording Manually
```
bruno-garcia, line 69: I'm assuming here we prefer users keep with defaults and let the SDK start/stop. Is that valid? For that reason, moving to the bottom.

These APIs can become a source of support issues, where the SDK is already start/stopping but users decide to call it too and then are confused by the duration of replays. For that reason, if `debug=true` and `start()` is called but there's already a replay running, worth logging out, for discoverability sake.

Do we know of some use cases when users want to start/stop manually? Is it worth adding here to the docs "you for example want to xyz"?

src/platforms/javascript/common/session-replay/privacy.mdx
```diff
@@ -1,10 +1,12 @@
 ---
-title: Privacy Configuration
+title: Privacy
 sidebar_order: 4200
 description: "Configuring Session Replay to maintain user and data privacy."
 ---
 
-There are several ways to deal with PII (personally identifiable information). By default, the integration will mask all text content with `*` and block all media elements (`img, svg, video, object, picture, embed, map, audio`). This can be disabled by setting `maskAllText` to `false`. It is also possible to add the following CSS classes to specific DOM elements to prevent recording its contents: `sentry-block`, `sentry-ignore`, and `sentry-mask`. The following sections will show examples of how content is handled by the differing methods.
+Sentry's Replay product is geared toward developers. By default, it'll mask all fiends, so no PII (personally identifiable information) is collected, eliminating any concerns with GDPR.
```
bruno-garcia, line 7: The idea here was some general, more product-focus' spiel about how we see privacy. And that we default in a way that ppl can trust and use this thing. We could even talk about server side PII stripping here and link to docs

src/platforms/javascript/common/session-replay/configuration.mdx
```diff
@@ -74,6 +45,17 @@ The following options can be configured as options to the integration, in `new R
...
 
+## Identifying Users
+
+You can use the Sentry SDK to set the user of a session, in order to associate a user identity to a session replay, refer to the <PlatformLink to="/enriching-events/identify-user">Identifying User Docs</PlatformLink>.
+
+Here's a short example using e-mail:
+
+```javascript
+import * as Sentry from "@sentry/browser";
+Sentry.setUser({ email: "sentaur@sentry.dev" });
```
Lms24, line 56:
```suggestion
Sentry.setUser({ email: "sentaur@sentry.dev" });
```

To spare us another place where we should add platform-specific imports (see my [comment](https://github.com/getsentry/sentry-docs/pull/5899#discussion_r1041993025)), I'd vote to just remove the import here. WDYT?

Lms24 (approved):
I like this flatter structure of the docs. It also addresses a [concern](https://github.com/getsentry/sentry-docs/pull/5899#discussion_r1042026697) I had about the "Custom Instrumentation" section.

src/platforms/javascript/common/session-replay/configuration.mdx
bruno-garcia: This is a challenge indeed. We went with "always add imports to snippets" a while back because it happens that users come back with "I get an error of Sentry is undefined" or the equivalent in other languages. But the added overhead of having to maintain multiple snippets is indeed an issue.

Your code comment (in the link you shared) can be a good approach. We can also add a couple of examples like:

Lms24: Sure, def not a blocker, merge away 👍 
And I agree, we should discuss this in one place and not multiple 😅

src/platforms/javascript/common/session-replay/configuration.mdx
```diff
@@ -83,3 +65,20 @@ new Replay({
   // any further configuration here is passed directly to rrweb
 });
 ```
+
+## Start and Stop Recording Manually
+
+Replay recording only starts automatically when it is included in the integrations key when calling `Sentry.init`. Otherwise you can initialize the plugin and manually call the `start()` method on the integration instance. To stop recording you can call the `stop()`.
```
imatwawana, line 71:
```suggestion
Replay recording only starts automatically when it's included in the integrations key when calling `Sentry.init`. Otherwise, you can initialize the plugin and manually call the `start()` method on the integration instance. To stop recording you can call the `stop()`:
```

src/platforms/javascript/common/session-replay/privacy.mdx
imatwawana, line 7:
```suggestion
Sentry's Replay product is geared toward developers. By default, it'll mask all fiends, so no personally identifiable information (PII) is collected, eliminating any concerns with General Data Protection Regulation (GDPR).
```

src/platforms/javascript/common/session-replay/privacy.mdx
```diff
@@ -1,10 +1,12 @@
...
+
+There are several ways to deal with PII. By default, the integration will mask all text content with `*` and block all media elements (`img, svg, video, object, picture, embed, map, audio`). This can be disabled by setting `maskAllText` to `false`. It is also possible to add the following CSS classes to specific DOM elements to prevent recording its contents: `sentry-block`, `sentry-ignore`, and `sentry-mask`. The following sections will show examples of how content is handled by the differing methods.
```
imatwawana, line 9:
```suggestion
There are several ways to deal with PII. By default, the integration will mask all text content with `*` and block all media elements (`img, svg, video, object, picture, embed, map, audio`). This can be disabled by setting `maskAllText` to `false`. It's also possible to add the following CSS classes to specific DOM elements to prevent recording their contents: `sentry-block`, `sentry-ignore`, and `sentry-mask`. The following sections will show examples of how content is handled by the differing methods.
```

bruno-garcia:
Thanks @imatwawana 🙏 will take this changes in a follow up PR (target was merged into master)

src/platform-includes/session-replay/install-session-replay/javascript.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+```html {tabTitle: CDN}
+<!--
+Note that the Replay bundle only contains the Replay integration and not the
+entire Sentry SDK. You have to add it in addition to the Sentry Browser SDK bundle:
+-->
+
+<script
+  src="https://browser.sentry-cdn.com/{{ packages.version('sentry.javascript.browser') }}/bundle.min.js"
+  integrity="sha384-{{ packages.checksum('sentry.javascript.browser', 'bundle.min.js', 'sha384-base64') }}"
```
AbhiPrasad, line 19: drive by review: we should make this the browser + performance bundle so it's easy for folks to use performance monitoring if they want to in the future.

src/platforms/javascript/common/session-replay/configuration.mdx
```diff
@@ -74,6 +45,17 @@ The following options can be configured as options to the integration, in `new R
 | inlineStylesheet | boolean | `true`  | Should inline stylesheets used in the recording |
 | recordCanvas     | boolean | `false` | Should record `<canvas>` elements |
 
+## Identifying Users
+
+You can use the Sentry SDK to set the user of a session, in order to associate a user identity to a session replay, refer to the <PlatformLink to="/enriching-events/identify-user">Identifying User Docs</PlatformLink>.
```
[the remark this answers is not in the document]
bruno-garcia: done in new branch, PR incoming.

src/platforms/javascript/common/session-replay/configuration.mdx
bruno-garcia: In new branch, PR incoming

bruno-garcia:
Superseded by https://github.com/getsentry/sentry-docs/pull/5927

PR #5919, opened 2022-12-07
fix(apple): Delete experimental note GA features

kevinrenskers:
Remove experimental suffixes for user interaction tracing, which is now GA. See https://github.com/getsentry/sentry-cocoa/pull/2503/


philipphofmann (approved):
LGTM

PR #5923, opened 2022-12-07
Update general-configuration - remove Canvas

jas-kas:
Removing reference to Canvas since it's not supported in Session Replay.


bruno-garcia (approved):
Tks

PR #5938, opened 2022-12-09
feat(apple): Rename the OOM integration

kevinrenskers:
OOM is now called watchdog terminations. See https://github.com/getsentry/sentry-cocoa/pull/2499, which will be part of the 8.0.0 release of sentry-cocoa.


src/platforms/apple/common/configuration/watchdog-terminations.mdx
```diff
@@ -0,0 +1,43 @@
+---
+title: Watchdog Terminations
+sidebar_order: 10
+description: "Learn how to turn off Watchdog Termination tracking"
+---
+
+This integration tracks <a href="https://developer.apple.com/documentation/xcode/addressing-watchdog-terminations">watchdog terminations</a> based on heuristics. This feature is available for iOS, tvOS, and Mac Catalyst, works only if the application was in the foreground, and doesn't track watchdog terminations for unit tests.
```
philipphofmann, line 7: `m`: I think we should add a note here that this was called OOM before 8.0.0.

src/platforms/apple/common/configuration/watchdog-terminations.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+When a typical unhandled error occurs, the Apple SDK writes a report to disk with the current state of your application, with details like the stack trace, tags, breadcrumbs, and so on, before the app terminates. When the watchdog terminates your app this happens without further notice, which means the SDK can't write a report to disk.
+As a result, in the Apple SDK, we track watchdog terminations during the app start based on heuristics, but getting the state of the app when a watchdog termination occurs is challenging. To provide the same state as we do for a typical crash, we would periodically need to store the app state to disk. This would cause a lot of I/O, which in turn could negatively affect the application's performance. To avoid this, watchdog termination events have less context than typical unhandled errors.
```
philipphofmann, line 10: > To provide the same state as we do for a typical crash, we would periodically need to store the app state to disk. This would cause a lot of I/O, which in turn could negatively affect the application's performance.

We should adapt this part as we have now breadcrumbs for them.

src/platforms/apple/common/configuration/watchdog-terminations.mdx
```diff
@@ -0,0 +1,43 @@
+---
+title: Watchdog Terminations
+sidebar_order: 10
+description: "Learn how to turn off Watchdog Termination tracking"
+---
+
+This integration tracks <a href="https://developer.apple.com/documentation/xcode/addressing-watchdog-terminations">watchdog terminations</a> based on heuristics. This feature is available for iOS, tvOS, and Mac Catalyst, works only if the application was in the foreground, and doesn't track watchdog terminations for unit tests.
+
+When a typical unhandled error occurs, the Apple SDK writes a report to disk with the current state of your application, with details like the stack trace, tags, breadcrumbs, and so on, before the app terminates. When the watchdog terminates your app this happens without further notice, which means the SDK can't write a report to disk.
```
philipphofmann, line 9: `m`: I think we should mention that one of the reasons are OOMs, and we don't create a watchdog termination issue if the app was hanging. We create AppHangs for that. Please put that info into the docs here.

philipphofmann (commented):
Thank you @kevinrenskers. We only can merge this once 8.0.0 is out, or we add a 8.0.0 beta note.

src/platforms/apple/common/configuration/out-of-memory.mdx
```diff
@@ -1,24 +1,28 @@
 ---
 title: Out Of Memory
 sidebar_order: 10
-description: "Learn how to turn off Out Of Memory"
+description: "Learn how to turn off Out Of Memory tracking"
 ---
 
+<Note>
+This integration is only available before 8.0.0. We renamed it to <PlatformLink to="/configuration/watchdog-terminations/">watchdog terminations</PlatformLink> in the 8.0.0 release.
+</Note>
```
imatwawana, line 9: Maybe this is obvious, but can we be more explicit about what we're talking about here? The SDK version? The iOS version?
```suggestion
<Note>

This integration is only available before version 8.0.0. It's now called <PlatformLink to="/configuration/watchdog-terminations/">Watchdog Terminations</PlatformLink> in the 8.0.0 release.

</Note>
```

src/platforms/apple/common/configuration/watchdog-terminations.mdx
```diff
@@ -0,0 +1,48 @@
+---
+title: Watchdog Terminations
+sidebar_order: 10
+description: "Learn how to turn off Watchdog Termination tracking"
+---
+
+<Note>
+This integration is available since 8.0.0. It was named <PlatformLink to="/configuration/out-of-memory/">Out of Memory</PlatformLink> before the 8.0.0 release.
+</Note>
```
imatwawana, line 9:
```suggestion
<Note>

This integration is available starting with version 8.0.0. It was called <PlatformLink to="/configuration/out-of-memory/">Out of Memory</PlatformLink> before the 8.0.0 release.

</Note>
```

src/platforms/apple/common/configuration/watchdog-terminations.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+This integration tracks <a href="https://developer.apple.com/documentation/xcode/addressing-watchdog-terminations">watchdog terminations</a> based on heuristics. This feature is available for iOS, tvOS, and Mac Catalyst, works only if the application was in the foreground, and doesn't track watchdog terminations for unit tests.
```
imatwawana, line 11: Changing this to a markdown link
```suggestion
This integration tracks [watchdog terminations](https://developer.apple.com/documentation/xcode/addressing-watchdog-terminations) based on heuristics. This feature is available for iOS, tvOS, and Mac Catalyst, works only if the application was in the foreground, and doesn't track watchdog terminations for unit tests.
```

src/platforms/apple/common/configuration/watchdog-terminations.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+When a typical unhandled error occurs, the Apple SDK writes a report to disk with the current state of your application, with details like the stack trace, tags, breadcrumbs, and so on, before the app terminates. When the watchdog terminates your app this happens without further notice, which means the SDK can't write a report to disk. A common reason the watchdog can terminate your app is an Out Of Memory problem. In the case the app is terminated because it hangs we don't create a watchdog termination event, instead an AppHangs event is created.
```
imatwawana, line 13:
```suggestion
When a typical unhandled error occurs, the Apple SDK writes a report to disk with the current state of your application, with details like the stack trace, tags, breadcrumbs, and so on, before the app terminates. When the watchdog terminates your app this happens without further notice, which means the SDK can't write a report to disk. A common reason the watchdog can terminate your app is an out of memory problem. If the app is terminated because it hangs, we don't create a watchdog termination event, but instead an `AppHangs` event is created.
```

src/platforms/apple/common/configuration/watchdog-terminations.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+As a result, in the Apple SDK, we track watchdog terminations during the app start based on heuristics, but getting the state of the app when a watchdog termination occurs is challenging. To provide the same state as we do for a typical crash, we would periodically need to store the app state to disk. This would cause a lot of I/O, which in turn could negatively affect the application's performance. To avoid this, watchdog termination events have less context than typical unhandled errors. Breadcrumbs are stored to disk and are added to watchdog termination events to help you figure out where in the app the event happened. This is done by appending the breadcrumbs to an open file and should have a marginal impact on your app's performance.
```
imatwawana, line 15:
```suggestion
As a result, in the Apple SDK, we track watchdog terminations during the app start based on heuristics, but getting the state of the app when a watchdog termination occurs is challenging. To provide the same state as we do for a typical crash, we would periodically need to store the app state to disk. This would cause a lot of I/O, which in turn could negatively affect the application's performance. To avoid this, watchdog termination events have less context than typical unhandled errors. Breadcrumbs are stored to disk and are added to watchdog termination events to help you figure out where in the app the event happened. This is done by appending the breadcrumbs to an open file, which should have a marginal impact on your app's performance.
```

src/platforms/apple/common/configuration/watchdog-terminations.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+When a user launches the app, the SDK checks the following statements and reports a watchdog termination if all of them are true:
+
+- The app didn't crash on the previous run.
+- The app was in the foreground/active.
+- The user launched the app at least once. When your app watch killed by the watchdog during the first launch, the SDK can't detect it.
```
imatwawana, line 21: It seems like a word is missing, but I'm not sure if I've got this right.
```suggestion
- The user launched the app at least once. When your app watch is killed by the watchdog during the first launch, the SDK can't detect it.
```

src/platforms/apple/common/configuration/watchdog-terminations.mdx
```diff
@@ -0,0 +1,48 @@
...
+- There was no debugger attached to the process of the app. If there was, our logic would falsely report OOMs whenever you restarted the app in development.
```
imatwawana, line 22: This bit is a little confusing. In the second sentence, are you describing the behaviour that will occur or the behaviour we're trying to avoid? If the latter, then I would put this in parentheses to make that clearer.
```suggestion
- There was no debugger attached to the process of the app. If there was, our logic would falsely report OOMs whenever you restarted the app in development.
```

src/platforms/apple/common/configuration/watchdog-terminations.mdx
```diff
@@ -0,0 +1,48 @@
...
+- The app is not running on a simulator.
```
imatwawana, line 23: Since these have all been in the past tense so far, let's keep it consistent.
```suggestion
- The app was not running on a simulator.
```

src/platforms/apple/common/configuration/watchdog-terminations.mdx
```diff
@@ -0,0 +1,48 @@
...
+- The OS did not update your app.
```
imatwawana, line 24:
```suggestion
- The OS didn't update your app.
```

src/platforms/apple/common/configuration/watchdog-terminations.mdx
kevinrenskers: We call this `Out Of Memory` everywhere else, like in the old docs. I think we should keep that.

src/platforms/apple/common/configuration/out-of-memory.mdx
kevinrenskers: It's the SDK version, which I don't know how we call that in the docs. `sentry-cocoa`? `The Sentry SDK`?

src/platforms/apple/common/configuration/watchdog-terminations.mdx
kevinrenskers: This is a direct copy from the existing OOM docs (https://docs.sentry.io/platforms/apple/configuration/out-of-memory/). To me it's obvious, but happy to see a suggestion, the suggested change here doesn't seem to actually change anything?

imatwawana: I hadn't made the change since I wasn't sure what the meaning was. I'm still not clear on what it means, lol. Is this the expected behaviour or the behaviour we're trying to avoid?

kevinrenskers: If the debugger was attached, we won't report watchdog terminations. Because if the debugger was attached we'd otherwise falsely report watchdog terminations. I've added the word "otherwise" in there which should clear things up I think.

src/platforms/apple/common/configuration/out-of-memory.mdx
imatwawana: I think just saying "version 8.0.0 of the Sentry Apple SDK" is fine.

src/platforms/apple/common/configuration/out-of-memory.mdx
```diff
@@ -1,24 +1,28 @@
...
-* The user launched the app at least once. When your app produces an OOM during the first launch, the SDK can't detect it.
-* There was no debugger attached to the process of the app. If there was, our logic would falsely report OOMs whenever you restarted the app in development.
-* The app is not running on a simulator.
-* The OS did not update your app.
-* The user didn't update the OS of their device.
-* The user didn't terminate the app manually, and neither did the OS.
+- The app didn't crash on the previous run.
+- The app was in the foreground/active.
+- The user launched the app at least once. When your app produces an OOM during the first launch, the SDK can't detect it.
+- There was no debugger attached to the process of the app. If there was, our logic would falsely report OOMs whenever you restarted the app in development.
```
imatwawana, line 21:
```suggestion
- There was no debugger attached to the process of the app. (If there was, our logic would falsely report OOMs whenever you restarted the app in development.)
```

src/platforms/apple/common/configuration/watchdog-terminations.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+When a user launches the app, the SDK checks the following statements and reports a watchdog termination if all of them are true:
+
+- The app didn't crash on the previous run.
+- The app was in the foreground/active.
+- The user launched the app at least once. When your app was killed by the watchdog during the first launch, the SDK can't detect it.
+- There was no debugger attached to the process of the app. If there was, our logic would otherwise falsely report watchdog terminations whenever you restarted the app in development.
```
imatwawana, line 22: I actually don't think "otherwise" works super well here after all that. Suggest just doing this instead:
```suggestion
- There was no debugger attached to the process of the app. (If there was, our logic would falsely report watchdog terminations whenever you restarted the app in development.)
```

src/platforms/apple/common/configuration/out-of-memory.mdx
imatwawana: Just not super clear why you committed the change for the other instance of this but not this one. It's pretty much the same language.

kevinrenskers: Just missed it.

imatwawana (approved):

philipphofmann (approved):
Thanks @kevinrenskers

PR #5940, opened 2022-12-09
Add last event id to the feedback example react-native

krystofwoldrich:
Not released yet.

- https://github.com/getsentry/sentry-react-native/pull/2675

### DIFF
--- a/src/platform-includes/enriching-events/user-feedback/sdk-api-example/react-native.mdx
+++ b/src/platform-includes/enriching-events/user-feedback/sdk-api-example/react-native.mdx
@@ -2,7 +2,8 @@
 import * as Sentry from '@sentry/react-native';
 import { UserFeedback } from '@sentry/react-native';
 
-const sentryId = Sentry.captureMessage('My Message');
+const sentryId = Sentry.lastEventId();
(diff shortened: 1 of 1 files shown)


src/platform-includes/enriching-events/user-feedback/sdk-api-example/react-native.mdx
```diff
@@ -2,7 +2,8 @@
 import * as Sentry from '@sentry/react-native';
 import { UserFeedback } from '@sentry/react-native';
 
-const sentryId = Sentry.captureMessage('My Message');
+const sentryId = Sentry.lastEventId();
+// OR: const sentryId = Sentry.captureMessage('My Message');
```
marandaneto, line 6:
```suggestion
const sentryId = Sentry.captureMessage('My Message');
// OR: const sentryId = Sentry.lastEventId();
```

marandaneto: `lastEventId` might be null or invalid if there's no event captured before using this function.

marandaneto (approved):

PR #5941, opened 2022-12-09
Add note to disable ad-blocker when installing JIRA

armenzg:
There's a note in the troubleshooting section but it is too common for customers to use ad-blockers.

### DIFF
--- a/src/docs/product/integrations/issue-tracking/jira/index.mdx
+++ b/src/docs/product/integrations/issue-tracking/jira/index.mdx
@@ -29,6 +29,8 @@ Sentry owner, manager, or admin permissions, and Jira admin permissions are requ
 2. Install the Sentry app through the [Jira marketplace](https://marketplace.atlassian.com/apps/1219432/sentry-for-jira?hosting=cloud&tab=overview).
 
 3. Select which Sentry organizations you’d like to use with Jira, and **Save Settings**.
+  * If you have an ad-blocker, disabled it now before you try to save the settings
+  * A link back to Sentry will be presented
 
 4. In Sentry, you’ll see a new Jira instance appear on the Integrations page.
 
--- a/src/docs/product/integrations/issue-tracking/jira/index.mdx
+++ b/src/docs/product/integrations/issue-tracking/jira/index.mdx
@@ -28,9 +28,9 @@ Sentry owner, manager, or admin permissions, and Jira admin permissions are requ
 
 2. Install the Sentry app through the [Jira marketplace](https://marketplace.atlassian.com/apps/1219432/sentry-for-jira?hosting=cloud&tab=overview).
 
-3. If you have an ad-blocker enabled you will not be able to complete step number 4.
+3. If you have an ad-blocker enabled, disable it now or you won't be able to complete step number 4.
 
-4. Select which Sentry organizations you’d like to use with Jira. A link back to Sentry will be presented.
+4. Select which Sentry organizations you’d like to use with Jira. Then click the presented link to return to Sentry.
 
 5. In Sentry, you’ll see a new Jira instance appear on the Integrations page.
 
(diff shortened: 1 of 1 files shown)


src/docs/product/integrations/issue-tracking/jira/index.mdx
```diff
@@ -29,6 +29,8 @@ Sentry owner, manager, or admin permissions, and Jira admin permissions are requ
 2. Install the Sentry app through the [Jira marketplace](https://marketplace.atlassian.com/apps/1219432/sentry-for-jira?hosting=cloud&tab=overview).
 
 3. Select which Sentry organizations you’d like to use with Jira, and **Save Settings**.
+  * If you have an ad-blocker, disabled it now before you try to save the settings
+  * A link back to Sentry will be presented
```
imatwawana, line 33:
```suggestion
  * If you have an ad-blocker, disable it now before you try to save the settings. A link back to Sentry will be presented.
```

imatwawana (commented):
I've just made a small change (typo and structure), but really we shouldn't be telling someone to save settings and then tell them, after they've read that, to disable the ad-blocker. Can we tell them this as step 1 or 2 instead? Are there downsides to that approach?

imatwawana:
> @imatwawana does this look better? image

It's better...but basically, this is a pre-requisite; you won't be able to complete this process with an ad-blocker enabled. So to reduce frustration, we want to let the user know that as early in the process as possible. I don't want to put it in a note before the steps because people sometimes just skip over notes, so I think rather than making one step longer to read, we just make it its own step. Does that make more sense?

armenzg:
> > @imatwawana does this look better? image
> 
> It's better...but basically, this is a pre-requisite; you won't be able to complete this process with an ad-blocker enabled. So to reduce frustration, we want to let the user know that as early in the process as possible. I don't want to put it in a note before the steps because people sometimes just skip over notes, so I think rather than making one step longer to read, we just make it its own step. Does that make more sense? Or is there a downside that you perceive to making this its own step?

How about this?

src/docs/product/integrations/issue-tracking/jira/index.mdx
```diff
@@ -28,9 +28,11 @@ Sentry owner, manager, or admin permissions, and Jira admin permissions are requ
 
 2. Install the Sentry app through the [Jira marketplace](https://marketplace.atlassian.com/apps/1219432/sentry-for-jira?hosting=cloud&tab=overview).
 
-3. Select which Sentry organizations you’d like to use with Jira, and **Save Settings**.
+3. If you have an ad-blocker enabled you will not be able to complete step number 4.
```
imatwawana, line 31:
```suggestion
3. If you have an ad-blocker enabled, disable it now or you won't be able to complete step number 4.
```

src/docs/product/integrations/issue-tracking/jira/index.mdx
```diff
@@ -28,9 +28,11 @@ Sentry owner, manager, or admin permissions, and Jira admin permissions are requ
...
 
-4. In Sentry, you’ll see a new Jira instance appear on the Integrations page.
+4. Select which Sentry organizations you’d like to use with Jira. A link back to Sentry will be presented.
```
imatwawana, line 33:
```suggestion
4. Select which Sentry organizations you’d like to use with Jira. Then click the link back to [sentry.io](https://sentry.io) that's presented.
```

src/docs/product/integrations/issue-tracking/jira/index.mdx
```diff
@@ -28,9 +28,11 @@ Sentry owner, manager, or admin permissions, and Jira admin permissions are requ
...
+
+5. In Sentry, you’ll see a new Jira instance appear on the Integrations page.
```
imatwawana, line 35:
```suggestion
5. In [sentry.io](https://sentry.io), you’ll see a new Jira instance appear on the **Integrations** page.
```

imatwawana (commented):
Good stuff. I've tweaked the wording to make step 3 an action/step rather than just information as well as some other style edits.

armenzg:
@imatwawana I decided to take most of your changes but I did not add the links since that's not where the redirection would have taken them to

imatwawana:
> @imatwawana I decided to take most of your changes but I did not add the links since that's not where the redirection would have taken them to

Gotcha. One of our style guidelines is to use the linked version of sentry to denote the application, but if it avoids confusion not to follow the guidelines, then so be it.

PR #5946, opened 2022-12-09
structural, wording, and styling changes

imatwawana:
Several structural, wording and style edits for the Session Replay docs


src/platform-includes/session-replay/setup/javascript.mdx
```diff
@@ -1,7 +1,7 @@
-To set up the integration, add the following to your Sentry initialization. Several options are supported and passable via the integration constructor. See the [configuration sections](/platforms/javascript/session-replay/) for more details.
+To set up the integration, add the following to your Sentry initialization. Several options are supported and passable using the integration constructor. See the [configuration sections](/platforms/javascript/session-replay/configuration/) for more details.
```
imatwawana, line 1: This link was pointing to the same page. I'm not sure if that's intentional (seems odd) or if this was meant to point to the configuration content.

src/platforms/javascript/common/session-replay/index.mdx
```diff
@@ -70,14 +70,14 @@ Sampling occurs when the session is first started. <PlatformIdentifier name="rep
 
 ## Error Linking
 
-Currently, errors that happen on the page while a replay is running are linked to the Replay, making it as easy as possible to jump between related issues/replays. However, please note that it is __possible__ that the error count reported on the Replay Detail page does not match the actual errors that have been captured. The reason for that is that errors can be lost, e.g. a network request fails, or similar. This should not happen to often, but be aware that it is theoretically possible.
+Currently, errors that happen on the page while a replay is running are linked to the Replay, making it as easy as possible to jump between related issues and replays. However, it is **possible** that the error count reported on the **Replays Details** page does not match the actual errors that have been captured. This is because errors can be lost; for example, a network request fails, or similar. This should not happen to often, but be aware that it is theoretically possible.
```
imatwawana, line 73:
```suggestion
Currently, errors that happen on the page while a replay is running are linked to the Replay, making it as easy as possible to jump between related issues and replays. However, it is **possible** that the error count reported on the **Replays Details** page does not match the actual errors that have been captured. This is because errors can be lost; for example, a network request may fail or similar. This shouldn't happen often, but it's theoretically possible.
```

src/platforms/javascript/common/session-replay/configuration.mdx
```diff
@@ -15,25 +15,25 @@ description: "Learn about the general Session Replay configuration fields."
 
 The following options can be configured on the root level of your browser-based Sentry SDK, in `init({})`:
```
imatwawana, line 17: On all the tables (on this page and on the Privacy page), should we be using PlatformIdentifier components to ensure correct casing with in the tables?

bruno-garcia: Even though this is top-level menu, it's only JS-configuration. Even if/when we expand to other platforms, we wouldn't be opt'ing in this page to those platforms, as too much of it wouldn't apply, including most options.

For that reason I believe this is fine as-is

bruno-garcia (approved):
Thank you @imatwawana

PR #5957, opened 2022-12-13
Add description for Next.js SDK `tunnelRoute` option

lforst:
Adds a description for the `tunnelRoute` option introduced in https://github.com/getsentry/sentry-javascript/pull/6425.

### DIFF
--- a/src/platforms/javascript/guides/nextjs/manual-setup.mdx
+++ b/src/platforms/javascript/guides/nextjs/manual-setup.mdx
@@ -167,6 +167,8 @@ const moduleExports = {
     //   'Configure Serverside Auto-instrumentation':
     //     - autoInstrumentServerFunctions
     //     - excludeServerRoutes
+    //   'Configure Tunneling to avoid Ad-Blockers':
+    //     - tunnelRoute
   },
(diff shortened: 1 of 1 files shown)


src/platforms/javascript/guides/nextjs/manual-setup.mdx
```diff
@@ -391,3 +393,23 @@ const moduleExports = {
 ```
 
 Excluded routes can be specified either as regexes or strings. When using a string, make sure that it matches the route exactly, and has a leading slash but no trailing one.
+
+## Configure Tunneling to avoid Ad-Blockers
+
+_(New in version 7.25.0)_
+
+You might notice that Sentry events are sometimes blocked by Ad-Blockers.
+Ad-blockers can be circumvented by using <PlatformLink to="/troubleshooting/#dealing-with-ad-blockers">tunneling</PlatformLink>.
```
Lms24, line 402: l: Should we move this line to the bottom of this section? Something like

If you don't want to use the automatic option, you can also circumvent ad blockers by using...

Lms24 (approved):

PR #5964, opened 2022-12-14
SPM: package(name:url:from:) is deprecated

edigaryev:
Problem: package(name:url:from:) is [deprecated since SPM 5.6](https://github.com/apple/swift-package-manager/blob/dd7e9cc68fccd6aad5db11aca2becfc24746905a/Sources/PackageDescription/PackageDependency.swift#L256).

Solution: switch to the new syntax.

### DIFF
--- a/src/platforms/apple/common/install/swift-package-manager.mdx
+++ b/src/platforms/apple/common/install/swift-package-manager.mdx
@@ -14,7 +14,7 @@ You can define your dependency rule by selecting the SDK version (or branch), an
 Alternatively, when using SPM version 5.2 or later and your project uses `Package.swift` file to manage dependencies, you can specify the target with name:
 
 ```swift {tabTitle:Swift}
-.package(name: "Sentry", url: "https://github.com/getsentry/sentry-cocoa", from: "{{ packages.version('sentry.cocoa') }}"),
+.package(url: "https://github.com/getsentry/sentry-cocoa", from: "{{ packages.version('sentry.cocoa') }}"),
 ```
 
 <Note>
--- a/src/platforms/apple/common/install/swift-package-manager.mdx
+++ b/src/platforms/apple/common/install/swift-package-manager.mdx
@@ -11,7 +11,7 @@ https://github.com/getsentry/sentry-cocoa.git
 
 You can define your dependency rule by selecting the SDK version (or branch), and then click the "Add Package" button.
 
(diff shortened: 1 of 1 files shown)


src/platforms/apple/common/install/swift-package-manager.mdx
```diff
@@ -14,7 +14,7 @@ You can define your dependency rule by selecting the SDK version (or branch), an
 Alternatively, when using SPM version 5.2 or later and your project uses `Package.swift` file to manage dependencies, you can specify the target with name:
```
philipphofmann, line 14:
```suggestion
Alternatively, when your project uses a `Package.swift` file to manage dependencies, you can specify the target with:
```

src/platforms/apple/common/install/swift-package-manager.mdx
```diff
@@ -14,7 +14,7 @@ You can define your dependency rule by selecting the SDK version (or branch), an
...
 
 ```swift {tabTitle:Swift}
-.package(name: "Sentry", url: "https://github.com/getsentry/sentry-cocoa", from: "{{ packages.version('sentry.cocoa') }}"),
+.package(url: "https://github.com/getsentry/sentry-cocoa", from: "{{ packages.version('sentry.cocoa') }}"),
```
philipphofmann, line 17: We could also change it to upToNextMajor
```suggestion
.package(url: "https://github.com/getsentry/sentry-cocoa", .upToNextMajor(from: "{{ packages.version('sentry.cocoa') }}"))
```

philipphofmann (commented):
Good catch, thanks. Two comments please.

src/platforms/apple/common/install/swift-package-manager.mdx
edigaryev: It seems that the `from:` argument already has this logic, from [`.package(...)` documentation](https://github.com/apple/swift-package-manager/blob/dd7e9cc68fccd6aad5db11aca2becfc24746905a/Sources/PackageDescription/PackageDependency.swift#L230-L231):

>Adds a package dependency that uses the version requirement, starting
>with the given minimum version, going up to the next major version.

philipphofmann: Okay, then we can leave it.

philipphofmann (approved):
LGTM

PR #5967, opened 2022-12-15
Include scopes for auth token for uploading Dsyms

Cronay:
When incorporating the automatic Debug Symbols Upload into our iOS-CI I created an Auth Token. I wanted to keep the permissions for the token to the minimum, but finding out which scopes are required for the token would now require a lot of trial and error. So it would be nice if this information could be included in this doc. I have included the scopes which our now working token has, but I expect some to be unnecessary. I hope you guys have this information available easily. I'm completely fine to editing the PR or you even closing it and add the permission scopes yourself.

Thanks!

### Legal Boilerplate

### DIFF
--- a/src/docs/clients/cocoa/dsym.mdx
+++ b/src/docs/clients/cocoa/dsym.mdx
@@ -105,6 +105,7 @@ api_host: 'https://mysentry.invalid/'
 Your project’s dSYM can be upload during the build phase as a “Run Script”. For this you need to st the _DEBUG_INFORMATION_FORMAT_ to be _DWARF with dSYM File_. By default, an Xcode project will only have _DEBUG_INFORMATION_FORMAT_ set to _DWARF with dSYM File_ in _Release_ so make sure everything is set in your build settings properly.
 
 You need to have an Auth Token for this to work. You can [create an Auth Token here](https://sentry.io/api/).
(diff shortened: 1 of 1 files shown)


src/docs/clients/cocoa/dsym.mdx
```diff
@@ -105,6 +105,7 @@ api_host: 'https://mysentry.invalid/'
 Your project’s dSYM can be upload during the build phase as a “Run Script”. For this you need to st the _DEBUG_INFORMATION_FORMAT_ to be _DWARF with dSYM File_. By default, an Xcode project will only have _DEBUG_INFORMATION_FORMAT_ set to _DWARF with dSYM File_ in _Release_ so make sure everything is set in your build settings properly.
 
 You need to have an Auth Token for this to work. You can [create an Auth Token here](https://sentry.io/api/).
```
philipphofmann, line 107: Thanks for trying to clarify that. Please link to this doc instead.
```suggestion
You need to have an Auth Token with the [correct permissions](https://docs.sentry.io/product/cli/dif/#permissions) for this to work. You can [create an Auth Token here](https://sentry.io/api/).
```

philipphofmann (commented):
Thanks @Cronay for opening this PR.

PR #5982, opened 2022-12-20
ref(android): Clarify initial Android SDK gradle setup step

markushi:
We got some feedback that it's not clear if the Sentry SDK should be applied on the project root or on the app level. This is an effort to make it more clear that the SDK needs be added on a module level.


romtsn:
maybe we should put `com.android.application` to the plugins block to make it more prominent?

```
plugins {
  id("com.android.application")
  id("io.sentry.android.gradle") version "{{ packages.version('sentry.java.android.gradle-plugin', '3.0.0') }}"
}
```

Could you also update [this page](https://docs.sentry.io/platforms/android/gradle/#install)?

romtsn (approved):

PR #5986, opened 2022-12-21
View hierarchy docs for Flutter

marandaneto:
Depends on https://github.com/getsentry/sentry-dart/pull/1189


src/platforms/common/enriching-events/viewhierarchy.mdx
```diff
@@ -13,6 +14,8 @@ notSupported:
   - dotnet.xamarin
 ---
 
+<Include name="alpha-note.mdx" />
```
krystofwoldrich, line 17: L: Should this be displayed for all platforms?

marandaneto: Yes, every platform is alpha, it was missing.

PR #6022, opened 2023-01-05
Add docs for JavaScript HttpClient Integration.

onurtemizkan:
Documents https://github.com/getsentry/sentry-javascript/pull/6500

### DIFF
--- a/src/platform-includes/configuration/http-client/javascript.mdx
+++ b/src/platform-includes/configuration/http-client/javascript.mdx
@@ -1,0 +1,54 @@
+```javascript {tabTitle: JavaScript}
+import * as Sentry from "@sentry/browser";
+import { HttpClient as HttpClientIntegration } from "@sentry/integrations";
+
+Sentry.init({
+  dsn: "___PUBLIC_DSN___",
+  integrations: [new HttpClientIntegration({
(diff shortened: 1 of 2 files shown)


src/platform-includes/configuration/http-client/javascript.mdx
```diff
@@ -0,0 +1,62 @@
+```javascript {tabTitle: JavaScript}
+import * as Sentry from "@sentry/browser";
+import { HttpClient as HttpClientIntegration } from "@sentry/integrations";
+
+Sentry.init({
+  dsn: "___PUBLIC_DSN___",
+  integrations: [
+    new HttpClientIntegration({
```
lforst, line 8: Looking at this again, I am wondering if we should have made the default [400, 599] instead. What was our thought process here again?

onurtemizkan: `[500, 599]` is defined in the spec: 
https://develop.sentry.dev/sdk/features/#http-client-errors

If I remember correctly, this also came up a few times as issues in Remix and Next SDKs (their request auto wrappers). It seems users don't normally expect 4xx's on their dashboard.

lforst (approved):

PR #6028, opened 2023-01-06
feat(perf-issues): Clarify N+1 Queries issues

gggritso:
Based on some feedback, more detailed explanations:

- Clarify span truncation
- Explain the source span
- Add note about fingerprinting

### DIFF
--- a/src/docs/product/issues/issue-details/performance-issues/n-one-queries.mdx
+++ b/src/docs/product/issues/issue-details/performance-issues/n-one-queries.mdx
@@ -29,6 +29,10 @@ The evidence for an N+1 queries problem has four main aspects:
 
 ![N+1 Query span evidence](n-plus-one-span-evidence.png)
 
+## Fingerprinting
+
+Sentry computes the issue fingerprint based on the parent span, the source span, and the repeating spans. If you're finding that Sentry is failing to correctly group N+1 query issues, make sure that the database spans in your transactions have parameterized queries as the span description. e.g., the span description `SELECT * FROM books WHERE books.id = 1` is not parameterized correctly, since it includes the book ID. The book ID parameter would vary between queries, and would create different fingerprints and therefore separate issues. The span description should look like this: `SELECT * FROM books WHERE books.id = %s`, or similar. Sentry attempts to parameterize queries, but it may miss some cases.
+
 ## Example
 
 Consider a book review website. It has two ORM models, `Book` and `Author`, each with a corresponding database table. The website shows a list of the ten oldest books and their respective authors. The code might look like this:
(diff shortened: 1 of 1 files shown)


src/docs/product/issues/issue-details/performance-issues/n-one-queries.mdx
```diff
@@ -28,6 +29,10 @@ The evidence for an N+1 queries problem has four main aspects:
 
 ![N+1 Query span evidence](n-plus-one-span-evidence.png)
 
+## Fingerprinting
+
+Sentry computes the issue fingerprint based on the parent span, the source span, and the repeating spans. If you're finding that Sentry is failing to correctly group N+1 query issues, make sure that the database spans in your transactions have parameterized queries as the span description. e.g., the span description `SELECT * FROM books WHERE books.id = 1` is not parameterized correctly, since it includes the book ID. The book ID parameter would vary between queries, and would create different fingerprints and therefore separate issues. The span description should look like this: `SELECT * FROM books WHERE books.id = %s`, or similar. Sentry attempts to parameterize queries, but it may miss some cases.
+
```
lizokm, line 35:
```suggestion
Sentry computes the issue fingerprint based on the parent span, the source span, and the repeating spans. If you're finding that Sentry is failing to correctly group N+1 query issues, make sure that the database spans in your transactions have parameterized queries as the span description. For example, the span description `SELECT * FROM books WHERE books.id = 1` is not parameterized correctly because it includes the book ID. The book ID parameter should vary for each query. This would create different fingerprints and separate issues. A correct span description should look something like this: `SELECT * FROM books WHERE books.id = %s`. While Sentry attempts to parameterize queries, it may miss some cases.

```

gggritso:
@lizokm thank you!

src/docs/product/issues/issue-details/performance-issues/n-one-queries.mdx
```diff
@@ -28,6 +29,10 @@ The evidence for an N+1 queries problem has four main aspects:
 
 ![N+1 Query span evidence](n-plus-one-span-evidence.png)
 
+## Fingerprinting
+
+Sentry computes the issue fingerprint based on the parent span, the source span, and the repeating spans. If you're finding that Sentry is failing to correctly group N+1 query issues, make sure that the database spans in your transactions have parameterized queries as the span description. For example, the span description `SELECT * FROM books WHERE books.id = 1` is not parameterized correctly because it includes the book ID. The book ID parameter should vary for each query. This would create different fingerprints and separate issues. A correct span description should look something like this: `SELECT * FROM books WHERE books.id = %s`. While Sentry attempts to parameterize queries, it may miss some cases.
```
mjq-sentry, line 34: > While Sentry attempts to parameterize queries,

Could we consider something like

> While Sentry attempts to parameterize queries while fingerprinting,

I'm trying to avoid any confusion around whether we actually change span descriptions (we don't, it's just during hashing). But I don't think we talk about span hashes in the docs elsewhere so "during fingerprinting" maybe gets that across? Just a thought.

gggritso: Yep, that makes sense!

mjq-sentry (approved):
🎸

PR #6057, opened 2023-01-12
docs(nextjs): Add info on opting out of client/server bundles

navin-moorthy:
Update the Next.js Manual Setup to include the following based on the issue https://github.com/getsentry/sentry-javascript/issues/6729

## Opt Out of Sentry SDK bundling in Client or Server side

If you want the `sentry` to be available in your server side & not in client side, you can make your `sentry.client.config.js` empty. This will prevent webpack from pulling in the Sentry related files when generating the browser bundle. The same goes the opposite for opting out of server side bundle by emptying `sentry.server.config.js`.

You cannot delete the respective config files because the SDK requires you to have it.

### Legal Boilerplate

### DIFF
--- a/src/platforms/javascript/guides/nextjs/manual-setup.mdx
+++ b/src/platforms/javascript/guides/nextjs/manual-setup.mdx
@@ -415,3 +415,9 @@ const moduleExports = {
 Please note that this option will tunnel Sentry events through your Next.js application so you might experience increased server usage.
 
 Learn more about tunneling in the <PlatformLink to="/troubleshooting/#dealing-with-ad-blockers">troubleshooting section</PlatformLink>.
+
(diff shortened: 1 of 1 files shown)


lforst:
Hi, thanks for writing in! Instead of documenting this, I think we should generally allow people not to have these empty config files laying around.

AbhiPrasad:
IMO I think we can add this for now, and then just change it later when we update how the config files work. We can always revert this PR in the future.

lforst:
> IMO I think we can add this for now, and then just change it later when we update how the config files work. We can always revert this PR in the future.

Yeah I guess that's true.

lforst:
@navin-moorthy Thank you for your contribution!

PR #5859, opened 2022-11-28
ref(apple): Rename APM tracking feature flags to tracing

kevinrenskers:
Rename APM tracking feature flags to tracing
     - Rename `SentryOptions.enableAutoPerformanceTracking` to `enableAutoPerformanceTracing`
     - Rename `SentryOptions.enableUIViewControllerTracking` to `enableUIViewControllerTracing`
     - Rename `SentryOptions.enablePreWarmedAppStartTracking` to `enablePreWarmedAppStartTracing`
     - Rename `SentryOptions.enableFileIOTracking` to `enableFileIOTracing`
     - Rename `SentryOptions.enableCoreDataTracking` to `enableCoreDataTracing`

See https://github.com/getsentry/sentry-cocoa/pull/2450. This will be released with 8.0.0 of sentry-cocoa.

PR #6059, opened 2023-01-12
fix(apple): Renamings from tracking to tracing

philipphofmann:
We renamed a couple of features from tracking to tracing for Cocoa 8.0.0. We missed a couple of renamings, which this PR addresses. Furthermore, we can remove some of the renamed features in the wizard and getting started as they are now enabled by default. Finally, this PR also clarifies that core data spans are enabled by default.

I did a bunch of different changes in one PR, as splitting them up would be complicated.

This is a follow-up on https://github.com/getsentry/sentry-docs/pull/5859.


lizokm (approved):
LGTM

PR #6070, opened 2023-01-13
feat(Android): Add docs for View Hierarchy

markushi:
Add documentation on how to enable the View Hierarchy feature for Android.
Our `sentry-java` needs to be shipped first before merging this, will probably be part of the `6.12.0` release.


romtsn (approved):
I guess we should also update the page with the new look once it's out 
image

PR #6073, opened 2023-01-13
chore(desp): match node version to sentry

evanpurkhiser:
my m1 mac was struggling to run the node 14 it has installed.

It's probably a "my system" issue, but I also see no reason this can't just be on node 18.


joshuarli (approved):

evanpurkhiser:
Going to do node 16 actually. It's not the active LTS like we are on sentry, but they changed some SSL stuff that we need to upgrade a bunch of gatsby stuff for

PR #6082, opened 2023-01-16
feat(replay): Update replay setup docs

mydea:
This updates the docs for Replay to ensure we reference the "new" installation style everywhere.


Lms24 (approved):
LGTM, I like the simplification here. I think it's safe to assume that users at this point have already installed the base SDK.

billyvg (approved):
LGTM

PR #6084, opened 2023-01-16
tracingOrigins changed to tracePropagationTargets

liamjones:
This property has been renamed in sentry-javascript and the original name is deprecated and marked for removal in the next major version.

See: https://github.com/getsentry/sentry-javascript/blob/461afdfc3740e19536b0c6fe515aeff40b130bd0/packages/tracing/src/browser/request.ts#L31

### Legal Boilerplate


krystofwoldrich (approved):
We can change this right away.

And maybe on the migration page can refer to the JS migration when relevant?
https://github.com/getsentry/sentry-javascript/blob/38f49ca22176d82646fa08d46cdfeecf1c0b7e98/MIGRATION.md#deprecated-tracingorigins-since-7190---6176

PR #6088, opened 2023-01-17
feat(replay): Update replay CDN usage docs

### LINKED ISSUE #5972: [JS] Make Browser+Tracing CDN bundle the main bundle
### Core or SDK?

Platform/SDK

### Which part? Which one?

JavaScript Browser SDK

### Description

We should consider "promoting" the Browser+Tracing bundle to be the first bundle mentioned in our [CDN bundle installation](https://docs.sentry.io/platforms/javascript/install/cdn/) page. Right now, it is only mentioned as the second bundle from the top (in the "Performance Bundle" section). 

### Suggested Solution

Proposal:

* In the top-most code example, show the Browser+Tracing bundle
* Replace "Performance Bundle" with "Errors-only Bundle" section, which will show the errors-only bundle.

Pros:
+ Promotes our performance product

Cons:
'- The browser+tracing bundle has a higher bundle size (~7Kb more, minified, gzipped). 

It's important to note though that this is only a purely cosmetic docs change. People who load right now the `bundle.min.js` (i.e. errors-only bundle) will not be affected at all.

mydea:
This updates the CDN usage docs to use the new full replay bundle by default, streamlining this.

While at it, I also reordered the "CDN installation" page generally a bit:

1. I moved the tracing bundle to top, as that is the default experience we want to provide.
2. After that, I added a new tracing + replay section
3. After that, there is the error only bundle
4. Moved the "Updates to naming scheme in SDK version 7" section further down, as it is not really very up to date anymore

Related to https://github.com/getsentry/sentry-docs/pull/6082, and dependent on https://github.com/getsentry/sentry-javascript/pull/6818.

Closes https://github.com/getsentry/sentry-javascript/issues/6573
Closes #5972

### DIFF
--- a/src/platform-includes/session-replay/install/javascript.mdx
+++ b/src/platform-includes/session-replay/install/javascript.mdx
@@ -11,20 +11,10 @@ yarn add @sentry/browser
 ```
 
 ```html {tabTitle: CDN}
-<!--
-Note that the Replay bundle only contains the Replay integration and not the
-entire Sentry SDK. You have to add it in addition to the Sentry Browser SDK bundle:
--->
-
 <script
-  src="https://browser.sentry-cdn.com/{{ packages.version('sentry.javascript.browser') }}/bundle.tracing.min.js"
-  integrity="sha384-{{ packages.checksum('sentry.javascript.browser', 'bundle.min.js', 'sha384-base64') }}"
+  src="https://browser.sentry-cdn.com/{{ packages.version('sentry.javascript.browser') }}/bundle.tracing.replay.min.js"
+  integrity="sha384-{{ packages.checksum('sentry.javascript.browser', 'bundle.tracing.replay.min.js', 'sha384-base64') }}"
   crossorigin="anonymous"
 ></script>
 
-<script
-  src="https://browser.sentry-cdn.com/{{ packages.version('sentry.javascript.browser') }}/replay.min.js"
-  integrity="sha384-{{ packages.checksum('sentry.javascript.browser', 'replay.min.js', 'sha384-base64') }}"
-  crossorigin="anonymous"
-></script>
 ```
--- a/src/platforms/javascript/common/install/cdn.mdx
+++ b/src/platforms/javascript/common/install/cdn.mdx
@@ -22,7 +22,7 @@ To use Sentry for error and performance monitoring, you can use the following bu
 
 ## Performance & Replay Bundle
 
-To use Sentry for error monitoring, performance tracing, and Replay, you can use the following bundle:
+To use Sentry for error and performance monitoring, as well as for [Session Replay](../session-replay), you can use the following bundle:
 
 ```html {tabTitle: CDN}
 <script
--- a/src/platforms/javascript/common/install/cdn.mdx
+++ b/src/platforms/javascript/common/install/cdn.mdx
@@ -10,7 +10,7 @@ Sentry supports loading the JavaScript SDK from a CDN. Generally we suggest usin
 
 ## Performance Bundle
 
-To use Sentry for error monitoring and performance tracing, you can use the following bundle:
+To use Sentry for error and performance monitoring, you can use the following bundle:
 
 ```html {tabTitle: CDN}
 <script
--- a/src/platforms/javascript/common/install/cdn.mdx
+++ b/src/platforms/javascript/common/install/cdn.mdx
@@ -8,24 +8,9 @@ import JsBundleList from "~src/components/jsBundleList";
 
 Sentry supports loading the JavaScript SDK from a CDN. Generally we suggest using our npm package (`@sentry/browser`) instead, as using the CDN can create scenarios where Sentry is unable to load due to networking issues or common extensions like ad blockers. If you _must_ use a CDN, take a look at [loading Sentry lazily with our JS loader](../lazy-load-sentry/), which provides a deferred version of our minified ES5 browser bundle. To see what other bundles are available, see [Available Bundles](#available-bundles) below.
 
-```html {tabTitle: CDN}
-<script
-  src="https://browser.sentry-cdn.com/{{ packages.version('sentry.javascript.browser') }}/bundle.min.js"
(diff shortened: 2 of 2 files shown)


src/platforms/javascript/common/install/cdn.mdx
```diff
@@ -8,38 +8,43 @@ import JsBundleList from "~src/components/jsBundleList";
 
 Sentry supports loading the JavaScript SDK from a CDN. Generally we suggest using our npm package (`@sentry/browser`) instead, as using the CDN can create scenarios where Sentry is unable to load due to networking issues or common extensions like ad blockers. If you _must_ use a CDN, take a look at [loading Sentry lazily with our JS loader](../lazy-load-sentry/), which provides a deferred version of our minified ES5 browser bundle. To see what other bundles are available, see [Available Bundles](#available-bundles) below.
 
+## Performance Bundle
+
+To use Sentry for error monitoring and performance tracing, you can use the following bundle:
```
Lms24, line 13:
```suggestion
To use Sentry for error and performance monitoring, you can use the following bundle:
```

src/platforms/javascript/common/install/cdn.mdx
```diff
@@ -8,38 +8,43 @@ import JsBundleList from "~src/components/jsBundleList";
...
   crossorigin="anonymous"
 ></script>
 ```
 
-<Alert level="info" title="Updates to naming scheme in SDK version 7">
+## Performance & Replay Bundle
 
-Version 7 of the Sentry JavaScript SDKs changed the bundles to be ES6 by default.
-Previously, the default bundles were compiled to ES5. If you need to support ES5, see [Available Bundles](#available-bundles) or the [Migration Guide to Version 7](https://github.com/getsentry/sentry-javascript/blob/master/MIGRATION.md#upgrading-from-6x-to-7x).
+To use Sentry for error monitoring, performance tracing, and Replay, you can use the following bundle:
```
Lms24, line 25:
```suggestion
To use Sentry for error and performance monitoring, as well as Session Replay, you can use the following bundle:
```

Lms24, line 25: Second idea: We could link to replay here as users might not be aware of the feature (didn't try the actual links, pls double check)
```suggestion
To use Sentry for error and performance monitoring, as well as [Session Replay](platforms/javascript/session-replay/), you can use the following bundle:
```

src/platforms/javascript/common/install/cdn.mdx
```diff
@@ -8,38 +8,43 @@ import JsBundleList from "~src/components/jsBundleList";
 
 Sentry supports loading the JavaScript SDK from a CDN. Generally we suggest using our npm package (`@sentry/browser`) instead, as using the CDN can create scenarios where Sentry is unable to load due to networking issues or common extensions like ad blockers. If you _must_ use a CDN, take a look at [loading Sentry lazily with our JS loader](../lazy-load-sentry/), which provides a deferred version of our minified ES5 browser bundle. To see what other bundles are available, see [Available Bundles](#available-bundles) below.
 
+## Performance Bundle
```
Lms24, line 11: I'm just thinking out lout here so feel fee to disregard but maybe we should make it clear that this is what we consider the default bundle? Right now it seems like we're not really providing a quick and easy "this is the bundle you most probably wanna use" code snippet. But people actually have to go in and figure out which one of the three they need. 

```suggestion
## Base Bundle
```

Lms24 (approved):
Looks good and I like the simplification from the previous version of the page! I just had a couple of suggestions for wording but feel free to go with whatever you think makes sense.

Lms24:
Added #5972 to the list of issues closed by this PR

src/platform-includes/session-replay/install/javascript.mdx
```diff
@@ -11,20 +11,10 @@ yarn add @sentry/browser
...
-<!--
-Note that the Replay bundle only contains the Replay integration and not the
-entire Sentry SDK. You have to add it in addition to the Sentry Browser SDK bundle:
--->
-
 <script
-  src="https://browser.sentry-cdn.com/{{ packages.version('sentry.javascript.browser') }}/bundle.tracing.min.js"
-  integrity="sha384-{{ packages.checksum('sentry.javascript.browser', 'bundle.min.js', 'sha384-base64') }}"
+  src="https://browser.sentry-cdn.com/{{ packages.version('sentry.javascript.browser') }}/bundle.tracing.replay.min.js"
+  integrity="sha384-{{ packages.checksum('sentry.javascript.browser', 'bundle.tracing.replay.min.js', 'sha384-base64') }}"
```
bruno-garcia, line 16: Lovely! Do we also add the `new Replay` and sample rates now on the default `Sentry.init` call?

src/platforms/javascript/common/install/cdn.mdx
bruno-garcia: Isn't the one with Replay the one we want users to use?

Downgrade from there if they don't want a feature for the sake of a smaller bundle size?

mydea: Yeah, I guess that's a general question to solve 😅 IMHO for now, especially while replay is still in beta, it makes sense to not have it the default bundle? I'd call it "Default bundle" maybe, as Base seems to me like it is the "minimum" bundle?

src/platform-includes/session-replay/install/javascript.mdx
mydea: Not yet, I think - I only changed the CDN bundles for now :)

src/platforms/javascript/common/install/cdn.mdx
Lms24: > IMHO for now, especially while replay is still in beta, it makes sense to not have it the default bundle? 

Agreed. Also, the bundle size footprint should be considered. Not saying we shouldn't promote it as the default bundle in the future but let's discuss this when it's stable.

> I'd call it "Default bundle" 

Sure, sounds also good to me 👍

bruno-garcia: Agreed, thanks folks

bruno-garcia (approved):

PR #6089, opened 2023-01-17
Enable before_send_transaction for python

antonpirker:
Python is now also able to specify a `before_send_transaction` option.


lizokm (approved):
LGTM

cleptric:
@antonpirker You should also update `src/platforms/common/configuration/filtering.mdx` and add some code snippets, like https://docs.sentry.io/platforms/php/configuration/filtering/#using-platformidentifier-namebefore-send-transaction-

PR #6093, opened 2023-01-17
fix(apple): Disabling crashes disables watchdog

philipphofmann:
Explain that disabling the crash handler on Apple also disables the watchdog termination integration.

### DIFF
--- a/src/platforms/apple/common/configuration/watchdog-terminations.mdx
+++ b/src/platforms/apple/common/configuration/watchdog-terminations.mdx
@@ -46,3 +46,6 @@ SentrySDK.start { options in
     options.enableWatchdogTerminationTracking = NO;
 }];
 ```
+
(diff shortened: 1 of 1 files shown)


src/platforms/apple/common/configuration/watchdog-terminations.mdx
```diff
@@ -46,3 +46,6 @@ SentrySDK.start { options in
     options.enableWatchdogTerminationTracking = NO;
 }];
 ```
+
+If you disable the `enableCrashHandler` option, the SDK disables the watchdog termination tracking as well, cause
+the SDK would report false positive watchdog terminations for each crash.
```
lizokm, line 51:
```suggestion
If you disable the `enableCrashHandler` option, the SDK will disable watchdog termination tracking. This will prevent false positive watchdog termination reporting for every crash.
```

lizokm (approved):
LGTM

PR #6102, opened 2023-01-18
feat(apple): More detail to when screenshots are captured

brustolin:
Add the information the screenshots are capture not only with errors, but also when the app crashes

### Legal Boilerplate


bruno-garcia (approved):

bruno-garcia:
"exception or crash" would still fall under 'an error' but lgtm

lizokm (approved):
LGTM

brustolin:
> "exception or crash" would still fall under 'an error'  ...

Thats also my understanding. But is not that clear for some people.

PR #6130, opened 2023-01-20
feat(n1-api-call): N+1 API Calls documentation

gggritso:
N+1 API Calls are a new performance issue type, and EA is coming soon! This document explains the issue and so on and so forth. Looking for all levels of feedback on this, from structure (though it matches the other issues) to grammar and so on!


src/docs/product/issues/issue-details/performance-issues/n-one-api-calls.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+- applies to `GET` requests only
+- a minimum of 10 network calls need to happen at the same time
+- the network calls must be a minimum of 50s in duration, and within 5ms of each other
+- the calls must not be to a GraphQL endpoint, or to `_next/data`
+- the calls must not be fetching a static resource (e.g., `.png` files)
+
+## Span Evidence
+
+The evidence for an N+1 queries problem has three main aspects:
```
lizokm, line 23:
```suggestion
You can identify N+1 problems by three main aspects in the "Span Evidence" section:
```

src/docs/product/issues/issue-details/performance-issues/n-one-api-calls.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+- Transaction name
+- Repeating spans - This is the URL that's receiving the repeated network calls
+- Parameters - These are the query parameters that vary between the network calls
+
+![N+1 API Calls span evidence](n-plus-one-api-calls-evidence.png)
```
lizokm, line 29:
```suggestion
![N+1 API Calls span evidence](n-plus-one-api-calls-evidence.png)

View it by going to the **Issues** page in Sentry, selecting your Android project, clicking on the N+1 API Call error you want to examine, then scrolling down to the "Span Evidence" section in the "Details" tab.


```
I recommend adding this section to all 3 performance issue types (N+1 Queries, N+1 API calls, and File IO). I think it's helpful for the user to know where exactly to go to see this in the product. I'm assuming these errors will be called N+1 API Calls?

src/docs/product/issues/issue-details/performance-issues/n-one-api-calls.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+## Fingerprinting
+
+Sentry computes the issue fingerprint based on the URLs of the repeating network calls. Sentry parameterizes the URLs by stripping integers, UUIDs, SHA1 hashes, and MD5 hashes. If you're finding that Sentry is failing to correctly group N+1 API Calls issues, you can parameterize the `http.client` spans in your transactions by using the <PlatformLink to="/configuration/options#before-send-transaction"><PlatformIdentifier name="before-send-transaction" /></PlatformLink> hook if your platform supports it.
+
```
lizokm, line 34:
```suggestion
Sentry computes the issue fingerprint based on the URLs of the repeating network calls. Sentry parameterizes the URLs by stripping integers, UUIDs, SHA1 hashes, and MD5 hashes. If you're finding that Sentry is failing to correctly group N+1 API Call issues, you can parameterize the `http.client` spans in your transactions by using the <PlatformLink to="/configuration/options#before-send-transaction"><PlatformIdentifier name="before-send-transaction" /></PlatformLink> hook if your platform supports it.

```

src/docs/product/issues/issue-details/performance-issues/n-one-api-calls.mdx
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <div>
+      <span>{book.title}</span>
+    </div>
+  );
+};
+```
+
+If the list has 100 books in it, the page will make 100 network calls simultaneously. The 100 is the "N" of "N+1 API Calls". This effect is visible on the span waterfall, which shows all the simultaneous network requests:
+
```
lizokm, line 56:
```suggestion
If the list has 100 books in it, the page will make 100 network calls simultaneously. The number 100 is the "N" in "N+1 API Calls". This effect is visible in the span waterfall, which shows all the simultaneous network requests:

```

src/docs/product/issues/issue-details/performance-issues/n-one-api-calls.mdx
```diff
@@ -0,0 +1,64 @@
...
+![N+1 API Calls in an example application](n-plus-one-api-calls-in-book-application.png)
+
+## Fixing N+1 API Call Issues
+
+Fixes for this issue type depend on the application context. In some cases, this behavior is intentional and even desired. It's up to you to decide what the appropriate fix is, if any. Here are some common approaches:
+
+- fetch all the information ahead of time, in one network calls. In the example above, you might update the `/books` endpoint to accept a list of IDs (e.g., `/books?id=78&id=89&id=120`)
+- lazy-load the information. In the book example, you could change the `<Book>` component to only fetch the information once the component is visible on the screen
```
lizokm, line 64:
```suggestion
- Fetch all the information ahead of time in one network call. In the example above, you might update the `/books` endpoint to accept a list of IDs (e.g., `/books?id=78&id=89&id=120`).
- Lazy-load the information. In the book example, you could change the `<Book>` component to only fetch the information once the component is visible on the screen.
```

lizokm (approved):
Made some wording tweaks and added an explanation of how to get to the Span Evidence section in the Sentry app. Let me know if you have any questions or want to chat.

src/docs/product/issues/issue-details/performance-issues/n-one-api-calls.mdx
```diff
@@ -0,0 +1,64 @@
+---
+title: "N+1 API Calls"
+sidebar_order: 15
+description: "Learn more about N+1 API calls and how to diagnose and fix them."
+---
+
+<Include name="early-adopter-note.mdx" />
+
+_N+1 API Calls_ are a performance problem in which an application makes many simultaneous network requests to fetch a given type of resource. For example, an application might make one network request for each item in a list, instead of fetching all the information in a combined network call. Each call to the server incurs some performance overhead, and if the number of calls is high this can lead to severe performance problems. This problems commonly occurs when rendering lists or grids of items, where each item needs to fetch additional information from the server.
+
```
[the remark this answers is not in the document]
gggritso: "Each item on a list" sound strange to me vs. "each item in a list" but taking this suggestion!

src/docs/product/issues/issue-details/performance-issues/n-one-api-calls.mdx
gggritso: Do you have any advice on how to correctly use "N+1 API Calls" vs. "N+1 API Call"? There are opinions about this flying around, and it's not very consistent. e.g., in the Sentry app the page title is "N+1 Query" and "N+1 API Call" on the issue details page, but the documentation often says "N+1 Queries" and "N+1 API Calls". Is there a good way to figure out what to use and when? Should it always be consistent?

src/docs/product/issues/issue-details/performance-issues/n-one-api-calls.mdx
gggritso: 🤔 this section of the doc is trying to explain the contents of the "Span Evidence" section (as in, here's what you get to help you solve this issue), does it make sense to say that this is how they're identified?

src/docs/product/issues/issue-details/performance-issues/n-one-api-calls.mdx
gggritso: They will be called that yeah, but I want to ask about this suggestion! People will land on this page from either:

1. Clicking the "Read More About N+1 API Call Issues" _from_ the issue details page
2. Going through the docs from Issues → Issue Details → Performance Issues → N+1 API Calls
3. Googling?

I think it might be weird to make a point of explaining how to get to a page that you probably came from?

gggritso:
@lizokm thank you for the review, all the suggestions look great! I just had a few questions before I merge some of them, to understand why they're there

src/docs/product/issues/issue-details/performance-issues/n-one-api-calls.mdx
gggritso: I like that! Reference to "root cause" helps here, I think!

gggritso: Took a shot, let me know if it's decent

src/docs/product/issues/issue-details/performance-issues/n-one-api-calls.mdx
lizokm: Super valid point. One of my goals is to get better analytics set up that trace how people go from the product to docs and back. That would answer these kinds of questions. Absolutely agree that it doesn't make sense for 1. & 2. But I think it does make sense for 3. To me, people who are coming from the product will just ignore the instructions for where to find it, but it may be SUPER useful to people who are coming from Google and may not know where to go. Plus, I think the product UI is not super intuitive and it can be hard to find things unless you've done it over and over again. So to me, the reward of helping people find the thing we're talking about in the product, outweighs the risk of being redundant for those who are coming from there. What do you think?

src/docs/product/issues/issue-details/performance-issues/n-one-api-calls.mdx
lizokm: I think we should be consistent with the Sentry app whenever possible. So if the app calls it "query" and "call" singular, we should refer to it as singular in docs too, at least when we're referring to the feature name. When we're not referring to it as the feature name, (like the sentence: If you're finding that Sentry is failing to correctly group N+1 API Call issues), then we should just go with what's grammatically correct :)

lizokm:
Thanks for the great questions!!

src/docs/product/issues/issue-details/performance-issues/n-one-api-calls.mdx
gggritso: Fair enough! So, should this page be called "N+1 API Query", then? And similarly for "N+1 Query"?

src/docs/product/issues/issue-details/performance-issues/n-one-api-calls.mdx
gggritso: That makes sense! The only strange aspect, I think, is that we're telling people to go to a page that probably doesn't exist. I'm worried that the doc has specific instructions that probably go nowhere, since most people won't have this kind of issue, or will need to search for it, or might not even be a Sentry customer! This isn't really a _feature_ that is always present, it's an explainer for a type of issue they might have 🤔

Would something like this work?

> You can find the "Span Evidence" section on the issue details page of an N+1 API Call issue, in the "Details" tab.

Or is that too sparse?

src/docs/product/issues/issue-details/performance-issues/n-one-api-calls.mdx
gggritso: Updated the pluralization so it says "N+1 API Call" whenever it explicitly mentions the detector!

src/docs/product/issues/issue-details/performance-issues/n-one-api-calls.mdx
gggritso: I added your suggestion minus the "Android" part, since it's not relevant here 🙏🏻

gggritso:
@lizokm integrated all your suggestions, thank you very much!

PR #6137, opened 2023-01-23
fix(file-io): Improve the file io screenshot

wmak:
- Followed https://www.notion.so/How-to-optimize-a-screenshot-211014f618b247049eadf0114baee2be to get a better screenshot, also narrowed my viewport when taking the screenshot so there's less padding


lizokm (approved):
LGTM

PR #6138, opened 2023-01-23
Grammar fix to dsn-explainer - DSN Utilization section

sdzhong:
Change 'advice' to 'advise'


lizokm (changes requested):
"Advice" is actually the correct word here :) Let's leave as is please.

PR #6142, opened 2023-01-24
Update outdated set-transaction-name python code

carlosa54:
Update the `set-transaction-name` outdated doc to use the latest `scope.set_transaction_name("UserListView")` api 

### Legal Boilerplate


AbhiPrasad:
Hey! This seems to have been addressed by https://github.com/getsentry/sentry-docs/pull/6145, but thanks for opening a PR.

PR #6177, opened 2023-01-30
feat: order platforms by their natural name

vaind:
Sorts `.NET` as `NET`

| Before | After |
| ----------- | ----------- |
| old      | new       |


lizokm (approved):
Thanks for this awesome fix!

PR #6184, opened 2023-01-31
fix(unity): Screenshot capture mechanim

bitsandfoxes:
The Unity SDK does not simply "capture a screenshot" but instead captures what is currently visible on screen from there game's perspective. This means that overlays from native plugins won't be visible on the screenshot. Which is important for PII reasons.


kahest:
Fixes https://github.com/getsentry/sentry-unity/issues/1161

kahest (approved):
LGTM

PR #6185, opened 2023-01-31
ref(pii): update the docs on PII scribbing and objects removal

olksdr:
related https://github.com/getsentry/relay/pull/1789

### DIFF
--- a/src/docs/product/data-management-settings/scrubbing/server-side-scrubbing.mdx
+++ b/src/docs/product/data-management-settings/scrubbing/server-side-scrubbing.mdx
@@ -43,7 +43,8 @@ credentials = {
 }
 ```
 
-Using the default options would cause the _object_ `credentials` to be set to `null`, and the object itself will not be entered. To prevent this from happening the keyname can be added to Safe Fields, which will leave the `credentials` object intact with all its contents.
+Using the default options would cause the _object_ `credentials` to be set to `null`, and the object itself will not be entered. To prevent this from happening the keyname can be added to Safe Fields (`credentials` in this case), which will leave the `credentials` object intact, but all of its entries would be subject to scrubbing.
+So `password` would be redacted by default, and adding `cats`, `username`, and/or `lastLogin` to the list of additional fields would cause those values to be redacted as well.
 
 <Note>
 
--- a/src/docs/product/data-management-settings/scrubbing/server-side-scrubbing.mdx
+++ b/src/docs/product/data-management-settings/scrubbing/server-side-scrubbing.mdx
@@ -31,7 +31,7 @@ With it enabled, Sentry will scrub the following:
   - An entry in "Additional Sensitive Fields" such as `mysekret`, for example, will cause the removal of any field named `mysekret`, but also removes any field _value_ that contains `mysekret`. Sentry does this to protect against sensitive data leaking as part of structured data that has been sent as a single string to Sentry (such as, a JSON object that is stringified and embedded as JSON string in another JSON structure).
   - As an extreme example where this behavior can become surprising, the string `"Unexpected error"` will be removed from events if the entry `exp` is in "Additional Sensitive Fields".
 
-Sentry will redact arrays and literal values, but also the entire objects. For example:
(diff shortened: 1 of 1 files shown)


src/docs/product/data-management-settings/scrubbing/server-side-scrubbing.mdx
```diff
@@ -31,7 +31,7 @@ With it enabled, Sentry will scrub the following:
   - An entry in "Additional Sensitive Fields" such as `mysekret`, for example, will cause the removal of any field named `mysekret`, but also removes any field _value_ that contains `mysekret`. Sentry does this to protect against sensitive data leaking as part of structured data that has been sent as a single string to Sentry (such as, a JSON object that is stringified and embedded as JSON string in another JSON structure).
   - As an extreme example where this behavior can become surprising, the string `"Unexpected error"` will be removed from events if the entry `exp` is in "Additional Sensitive Fields".
 
-Sentry will redact arrays and literal values, but not entire objects. For example:
+Sentry will redact arrays and literal values, but also the entire objects. For example:
```
jjbayer, line 34:
```suggestion
Sentry will redact arrays and literal values, but also entire objects. For example:
```

src/docs/product/data-management-settings/scrubbing/server-side-scrubbing.mdx
```diff
@@ -43,7 +43,7 @@ credentials = {
 }
 ```
 
-Using the default options would not cause the _object_ `credentials` to be redacted in its entirety; rather, all of its entries would be subject to scrubbing. So `password` would be redacted by default, and adding `cats`, `username`, and/or `lastLogin` to the list of additional fields would cause those values to be redacted as well.
+Using the default options would cause the _object_ `credentials` to be set to `null`, and the object itself will not be entered. To prevent this from happening the keyname can be added to Safe Fields, which will leave the `credentials` object intact with all its contents.
```
jjbayer, line 46: Did you verify this? I am not 100% sure if putting `credentials` in `Safe Fields` will leave the entire object untouched, or if it will descend and scrub the `password` field anyway.

olksdr: Added test in https://github.com/getsentry/relay/pull/1789/commits/05caa3442f50a2fe3b3b65a21d4a752504905a84 to verify. Also updated the doc accordingly.

src/docs/product/data-management-settings/scrubbing/server-side-scrubbing.mdx
```diff
@@ -43,7 +43,8 @@ credentials = {
 }
 ```
 
-Using the default options would not cause the _object_ `credentials` to be redacted in its entirety; rather, all of its entries would be subject to scrubbing. So `password` would be redacted by default, and adding `cats`, `username`, and/or `lastLogin` to the list of additional fields would cause those values to be redacted as well.
+Using the default options would cause the _object_ `credentials` to be set to `null`, and the object itself will not be entered. To prevent this from happening the keyname can be added to Safe Fields (`credentials` in this case), which will leave the `credentials` object intact, but all of its entries would be subject to scrubbing.
+So `password` would be redacted by default, and adding `cats`, `username`, and/or `lastLogin` to the list of additional fields would cause those values to be redacted as well.
```
lizokm, line 47:
```suggestion
Using the default options will cause the _object_ `credentials` to be set to `null`, and the object itself will not be entered. To prevent this from happening add the keyname to Safe Fields (`credentials` in this case). This will leave the `credentials` object intact, while making all of its entries subject to scrubbing.
So `password` will be redacted by default, and adding `cats`, `username`, and/or `lastLogin` to the list of additional fields will cause those values to be redacted as well.
```

lizokm (approved):
Small copy tweak.

src/docs/product/data-management-settings/scrubbing/server-side-scrubbing.mdx
jjbayer: Thanks for double checking!

jjbayer (approved):

PR #6187, opened 2023-01-31
fix(unity): Native screenshots

bitsandfoxes:
Went into more detail on the difference of which SDK tries to capture a screenshot.


mattjohnsonpint (approved):

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-unity/issues/1161

PR #6200, opened 2023-02-01
ref(heroku): Update Heroku docs

ceorourke:
Update the Heroku docs to include a line about copying the webhooks signing secret into your Sentry plugin configuration, as well as some small cleanups that confused me when I read through it.


lizokm (approved):
LGTM :)

PR #6206, opened 2023-02-02
ref(data-scrubbing): Clarify rule preference

iker-barriocanal:
Rules from organizations have preference over rules from projects, see: 
https://github.com/getsentry/sentry/blob/bc89d976a0dbe0d169d7dedc072975befb6d2ff6/src/sentry/datascrubbing.py#L45

### DIFF
--- a/src/docs/product/data-management-settings/scrubbing/server-side-scrubbing.mdx
+++ b/src/docs/product/data-management-settings/scrubbing/server-side-scrubbing.mdx
@@ -7,7 +7,7 @@ keywords: ["pii", "gdpr", "personally identifiable data", "compliance"]
 description: "Learn about the data scrubbing Sentry enables by default."
 ---
 
-If you cannot - or do not want to - <PlatformLink to="/data-management/sensitive-data/">scrub data within the SDK</PlatformLink>, Sentry provides a number of options to achieve this in the product itself. You'll find these settings under the **Data Scrubber** option within both your Organization and individual Project settings. By default this is enabled, and we highly recommend you keep it that way.
+If you cannot - or do not want to - <PlatformLink to="/data-management/sensitive-data/">scrub data within the SDK</PlatformLink>, Sentry provides a number of options to achieve this in the product itself. You'll find these settings under the **Data Scrubber** option within both your Organization and individual Project settings (organization settings have preference over project settings). By default this is enabled, and we highly recommend you keep it that way.
 
 With it enabled, Sentry will scrub the following:
 


jjbayer (changes requested):
Is precedence the right word here? Looking at the linked function, each flag is `OR`ed, and the list options are joined (union), so to me it looks like both have equal precedence. Could you link to the discussion that triggered this change?

jjbayer (approved):

src/docs/product/data-management-settings/scrubbing/server-side-scrubbing.mdx
```diff
@@ -7,7 +7,7 @@ keywords: ["pii", "gdpr", "personally identifiable data", "compliance"]
 description: "Learn about the data scrubbing Sentry enables by default."
 ---
 
-If you cannot - or do not want to - <PlatformLink to="/data-management/sensitive-data/">scrub data within the SDK</PlatformLink>, Sentry provides a number of options to achieve this in the product itself. You'll find these settings under the **Data Scrubber** option within both your Organization and individual Project settings. By default this is enabled, and we highly recommend you keep it that way.
+If you cannot - or do not want to - <PlatformLink to="/data-management/sensitive-data/">scrub data within the SDK</PlatformLink>, Sentry provides a number of options to achieve this in the product itself. You'll find these settings under the **Data Scrubber** option within both your Organization and individual Project settings. If organization settings exist, those will be applied and project-level settings will be ignored; to enable a subset of projects, disable the organization-level settings and enable each project individually. By default data scrubbing is enabled, and we highly recommend you keep it that way.
```
lizokm, line 10:
```suggestion
If you can't or don't want to <PlatformLink to="/data-management/sensitive-data/">scrub data within the SDK</PlatformLink>, you can choose from several data scrubbing options in the Sentry app on either an organization-wide or a project level. Data scrubbing is enabled by default and we highly recommend you keep it that way.

To set data scrubbing rules for your entire organization, go to **[Settings] > Security & Privacy,** and scroll down to the "DATA SCRUBBING" section.

To set them on a per-project basis, go to **[Settings] > Projects**, select a project, click "Security & Privacy" and scroll down to the "DATA SCRUBBING" section. If organization-wide settings exist, they’ll override any project settings, so be sure to disable them first. 




```

lizokm (approved):
I think it helps to give users a more direct path for where they can find these different data-scrubbing options in the product since it's not very obvious :)

PR #6215, opened 2023-02-02
feat(crons): Document timezone config

evanpurkhiser:
image


davidenwang (approved):

src/docs/product/crons/getting-started.mdx
```diff
@@ -25,6 +25,10 @@ Sentry Cron Monitoring allows you to monitor any job running on a scheduled inte
      Example of a Crontab for every hour.
      ![Example of an every hour Crontab](cron-setup-crontab-example-2.png)
      Example of a Crontab for every month on the 1st and 15th day.
+
+     You will also want to ensure the timezone matches the timezone of the machine your cron is running on. This is important to account for daylight saving time in some cases.
+     ![Crontab timezone setting](cron-setup-crontab-timezone.png)
+
```
lizokm, line 31:
```suggestion
     To account for daylight saving time, you'll want to make sure that the timezone you set matches the timezone of the machine your cron is running on.
     ![Crontab timezone setting](cron-setup-crontab-timezone.png)

```

lizokm (approved):
Let's rethink the structure. It would be good to strip the steps down to their bare minimum and add any additional information in a different section.

src/docs/product/crons/getting-started.mdx
evanpurkhiser: Let's follow up with these changes. I think we want to leave the DST note in there since it is an important edge case.

lizokm: Yup, would love to work together to figure out what's most important to keep. We can also add this as a note rather than leaving it as part of the step itself.

PR #6225, opened 2023-02-06
Remove non existing link to our Expo troubleshooting page

krystofwoldrich:
The link in our docs doesn't exist anymore.

### DIFF
--- a/src/platform-includes/getting-started-install/react-native.mdx
+++ b/src/platform-includes/getting-started-install/react-native.mdx
@@ -14,7 +14,7 @@ If you are using Expo, see [How to Add Sentry to Your Expo Project](https://docs
 
 <Note>
 
-Make sure that the version of `@sentry/react-native` matches what `sentry-expo` depends on, [read more on our troubleshooting page](/platforms/react-native/troubleshooting/#expo).
+Make sure that the version of `@sentry/react-native` matches what `sentry-expo` depends on, [read more in Expo documentation](https://docs.expo.io/guides/using-sentry/).
 
 </Note>
 


marandaneto (approved):

src/platform-includes/getting-started-install/react-native.mdx
```diff
@@ -14,7 +14,7 @@ If you are using Expo, see [How to Add Sentry to Your Expo Project](https://docs
 
 <Note>
 
-Make sure that the version of `@sentry/react-native` matches what `sentry-expo` depends on, [read more on our troubleshooting page](/platforms/react-native/troubleshooting/#expo).
+Make sure that the version of `@sentry/react-native` matches what `sentry-expo` depends on, [read more in Expo documentation](https://docs.expo.io/guides/using-sentry/).
```
lizokm, line 17: Thanks for updating the link. The troubleshooting one is definitely not helpful :)
But I'm also not super clear on what exactly the note is saying. What exactly does `sentry-expo` depend on? I read through the Expo docs and am still unsure. Could we clarify or point to a specific section of the Expo docs that talks about this??

lizokm (commented):
This note could use a bit more clarification :)

src/platform-includes/getting-started-install/react-native.mdx
krystofwoldrich: I don't think it's mentioned on Expo docs. But their package depends on an exact version of our SDK.

https://github.com/expo/sentry-expo/blob/main/package.json#LL43

But sometimes users install a different version than the compatible one and that can cause issues.

PR #6235, opened 2023-02-07
fix(nextjs): move installation instructions

scefali:
Explicitly installing the sentry package is no long necessary with the wizard


lizokm (approved):
LGTM

PR #6245, opened 2023-02-07
Add Session Replay to Sentry docs

AbhiPrasad:
Update to use https://docs.sentry.io/platforms/javascript/guides/gatsby/#sentry-configuration-file, so we can use the Replay integration.

I'm going to remove the plugin config docs, since folks should not be using this anymore (and add a deprecation message as well).


bruno-garcia (approved):
Nice!! Thanks Abhi

PR #6247, opened 2023-02-07
feat(ruby): Add before_send_transaction

sl0thentr0py:


cleptric:
You may also want to update https://docs.sentry.io/platforms/ruby/configuration/filtering/, see https://docs.sentry.io/platforms/javascript/configuration/filtering/#using-platformidentifier-namebefore-send-transaction-

src/platforms/ruby/common/configuration/options.mdx
```diff
@@ -52,10 +52,16 @@ end
...
-: Provide a lambda or proc. This will be `called` before sending an event to Sentry. Receives an `event` and `hint` as parameter. `hint` is a dict `{:exception => ex | nil, :message => message | nil}`. It is possible to mutate the event, also if this function returns `nil` the event will be dropped and not sent.
+: Provide a lambda or proc. This will be `called` before sending an error event to Sentry. Receives an `event` and `hint` as parameter. `hint` is a dict `{:exception => ex | nil, :message => message | nil}`. It is possible to mutate the event, also if this function returns `nil` the event will be dropped and not sent.
 
 <PlatformContent includePath="configuration/before-send" />
 
+`before_send_transaction`
+
+: Provide a lambda or proc. This will be `called` before sending a transaction event to Sentry. Receives an `event` and `hint` as parameter. `hint` is currently empty but might be augmented in the future. It is possible to mutate the event, also if this function returns `nil` the event will be dropped and not sent.
+
+<PlatformContent includePath="configuration/before-send-transaction" />
```
lizokm, line 63:
```suggestion

: Provide a lambda or proc. This will be `called` before sending a transaction event to Sentry. It receives an `event` and `hint` as a parameter. The `hint` event is currently empty, but might be augmented in the future. It's possible to mutate the event. Also, if this function returns `nil`, the event will be dropped and not sent.

<PlatformContent includePath="configuration/before-send-transaction" />
```
I wonder if it's possible to add a bit of additional context here and also for the "before_send_transaction" the same way we have additional context for "Background_worker_threads" and "Backtrace_cleanup_callback".  It's really helpful to talk about why a user might want to do the things we're describing.

lizokm (approved):
Made a few suggestions.

antonpirker (approved):

sl0thentr0py:
added @cleptric ty!

PR #6267, opened 2023-02-09
feat(replay): Add note about potential Angular performance impact

billyvg:
image

### DIFF
--- a/src/platforms/javascript/common/session-replay/index.mdx
+++ b/src/platforms/javascript/common/session-replay/index.mdx
@@ -31,9 +31,11 @@ By default, it'll mask all fields, so no personally identifiable information (PI
 
 <PlatformContent includePath="session-replay/pre-requisites" />
 
-<Alert level="warning">
-  We’re investigating possible performance degradation in some **Angular applications**. We recommend testing your application before releasing. Please reach out via [this GitHub ticket](https://github.com/getsentry/sentry-javascript/issues/6946) if you have issues.
+<PlatformSection supported={["javascript.angular"]}>
+<Alert level="warning" title="Warning">
+We’re investigating possible performance degradation in <b>some</b> Angular applications. We recommend testing your application before releasing. Please reach out via <a href="https://github.com/getsentry/sentry-javascript/issues/6946">this GitHub ticket</a> if you have issues.
 </Alert>
+</PlatformSection>
 
 ## Install
 
--- a/src/platforms/javascript/common/session-replay/index.mdx
+++ b/src/platforms/javascript/common/session-replay/index.mdx
@@ -31,6 +31,10 @@ By default, it'll mask all fields, so no personally identifiable information (PI
 
 <PlatformContent includePath="session-replay/pre-requisites" />
 
+<Alert level="warning">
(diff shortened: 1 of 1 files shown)


src/platforms/javascript/common/session-replay/index.mdx
```diff
@@ -31,6 +31,10 @@ By default, it'll mask all fields, so no personally identifiable information (PI
 
 <PlatformContent includePath="session-replay/pre-requisites" />
 
+<Alert level="warning">
+  We’re investigating possible performance degradation in some Angular applications. We recommend testing your application before releasing. Please reach out via [this GitHub ticket](https://github.com/getsentry/sentry-javascript/issues/6946) if you have issues.
```
Lms24, line 35: Just to make this a little more obvious, wdyt? 

```suggestion
  We’re investigating possible performance degradation in some **Angular applications**. We recommend testing your application before releasing. Please reach out via [this GitHub ticket](https://github.com/getsentry/sentry-javascript/issues/6946) if you have issues.
```

Lms24 (commented):
Hmm, not sure if we actually need this, given that we've only received two reports so far? 
But I might be missing something/not be up to date, so feel free to merge it if y'all discussed this already.

billyvg:
@Lms24 We talked about this briefly in a meeting, this was a potential solution in order to be extra cautious. It's hard for us to determine actual impact since we're relying on 2 layers of feedback.

Lms24 (approved):
Fair enough, thanks for explaining

jas-kas (changes requested):
@billyvg Can we only show this on the Session Replay instructions for Angular - https://docs.sentry.io/platforms/javascript/guides/angular/session-replay/

Right now, it shows up on every single instructions for each platform (Ember, Gatsby, etc.) which I think is unnecessary and seems alarming. 

Also, the bolding and hyperlinking seems to be broken in the note : (

Screen Shot 2023-02-09 at 3 21 04 PM

src/platforms/javascript/common/session-replay/index.mdx
```diff
@@ -31,6 +31,10 @@ By default, it'll mask all fields, so no personally identifiable information (PI
 
 <PlatformContent includePath="session-replay/pre-requisites" />
 
+<Alert level="warning">
+  We’re investigating possible performance degradation in some **Angular applications**. We recommend testing your application before releasing. Please reach out via [this GitHub ticket](https://github.com/getsentry/sentry-javascript/issues/6946) if you have issues.
+</Alert>
+
```
lizokm, line 37:
```suggestion
<Alert level="warning">

 We’re investigating possible performance degradation in some **Angular applications**. We recommend testing your application before releasing. Please reach out via <a class href=“https://github.com/getsentry/sentry-javascript/issues/6946”>this GitHub ticket</a> if you have issues.
 
</Alert>

```
Links haven't been working in alerts/notes for some reason. ^^ they do seems to work if we write them this way.

lizokm (approved):
Updated the link, hopefully it's fix that and the bolding issue... but definitely agree with @jas-kas that we should only have this show up in Angular.

src/platforms/javascript/common/session-replay/index.mdx
billyvg: Ah looks like markdown is not supported inside of react components, thanks!

jas-kas (approved):

Issue #6150, opened 2023-01-25
Update Sample rate to be the same for JS setup


smeubank:
### Core or SDK?

Platform/SDK

### Which part? Which one?

JS Browser Docs

### Description

[Getting Started](https://docs.sentry.io/platforms/javascript/guides/react/)
vs
[Set Up Performance](https://docs.sentry.io/platforms/javascript/guides/react/performance/)

we have 2 different rates in the snippets

### Suggested Solution

Should be the same everywhere

sdzhong:
In addition--in the JS Browser docs, [Setting a Sampling Function](https://docs.sentry.io/platforms/javascript/configuration/sampling/#setting-a-sampling-function)--I believe the `tracesSampler` option should have an else `return 1;` instead of 0.1 

> } else {
>   // Default sample rate
>   return 0.1;
> }

PR #6288, opened 2023-02-13
ref(perf): Update JS sample rates to 1.0

AbhiPrasad:
resolves https://github.com/getsentry/sentry-docs/issues/6150

Change to `tracesSampleRate: 1.0`.


src/platform-includes/performance/traces-sample-rate/javascript.mdx
```diff
@@ -2,6 +2,6 @@
 Sentry.init({
   // ...
 
-  tracesSampleRate: 0.2,
+  tracesSampleRate: 0.5,
```
Lms24, line 5: l: Just noticed that in the [text above this include](https://sentry-docs-git-abhi-sample-rate-js.sentry.dev/platforms/javascript/configuration/sampling/#setting-a-uniform-sample-rate) we still talk about 0.2. I think we should update this as well.

Lms24 (approved):

src/platform-includes/performance/traces-sample-rate/javascript.mdx
AbhiPrasad: Done!

PR #6291, opened 2023-02-13
Update Replay README.md wrong default values

lucas-zimerman:
https://github.com/getsentry/sentry-javascript/pull/6878 altered on version 7.33.0 the default to 0 but the docs still point as if the default was 1.0

### DIFF
--- a/src/platforms/javascript/common/session-replay/configuration.mdx
+++ b/src/platforms/javascript/common/session-replay/configuration.mdx
@@ -16,7 +16,7 @@ The following options can be configured on the root level of your browser-based 
 
 | Key                      | Type   | Default | Description                                                                                                                                                                                                                                                       |
 | ------------------------ | ------ | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


src/platforms/javascript/common/session-replay/configuration.mdx
```diff
@@ -17,7 +17,7 @@ The following options can be configured on the root level of your browser-based
 | Key                      | Type   | Default | Description                                                                                                                                                                                                                                                       |
 | ------------------------ | ------ | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | replaysSessionSampleRate | number | `0.1`   | The sample rate for replays that begin recording immediately and last the entirety of the user's session. `1.0` collects all replays, and `0` collects none.                                                                                                      |
```
billyvg: This should be zero as well.

```suggestion
| replaysSessionSampleRate | number | `0`   | The sample rate for replays that begin recording immediately and last the entirety of the user's session. `1.0` collects all replays, and `0` collects none.                                                                                                      |
```

billyvg (approved):
Thanks, good catch!

Lms24 (approved):

src/platforms/javascript/common/session-replay/configuration.mdx
lucas-zimerman: i have just fixed it 😅

PR #6292, opened 2023-02-13
feat(browser): Add offline transport docs

timfish:
This PR:
- Removes reference to the `Offline` integration since this will be deprecated
- Add a `Transports` section to the JavaScript configuration
  - Excludes this from sub-plartforms which already have their own specialised offline support
- Documents how to use `makeBrowserOfflineTransport` to create an offline transport

Does this need separate platform includes with differing imports for Angular/React/etc/etc?

### DIFF
--- a/src/platforms/javascript/common/configuration/transports.mdx
+++ b/src/platforms/javascript/common/configuration/transports.mdx
@@ -1,0 +1,67 @@
+---
+title: Transports
+sidebar_order: 30
+description: "Transports let you change the way in which events are delivered to Sentry."
+notSupported:
+  - javascript.capacitor
+  - javascript.cordova
+  - javascript.electron
+  - javascript.wasm
+---
+
+The JavaScript SDK uses a `transport` to send events to Sentry. On modern
+browsers, the default transport uses the browsers `fetch` API to send events.
+
+The default transports drop events if they fail to send due to a lack of connection.
+
+### Offline Caching
+
+If your JavaScript application is designed to continue working offline,
+dropping events when no connection is available could mean you are missing
+events that occur in that mode.
+
+To enable offline caching of events, `makeBrowserOfflineTransport` can be used
+to wrap existing transports and queue events in the browsers IndexedDB storage.
+Sending of events is then retried later.
+
+<PlatformContent includePath="configuration/offline" />
+
+This adds a number of optional properties to `transportOptions`:
+
+```javascript
+transportOptions:{
+  /**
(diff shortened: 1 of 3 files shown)


src/platforms/javascript/common/configuration/transports.mdx
```diff
@@ -0,0 +1,67 @@
...
+  - javascript.capacitor
+  - javascript.cordova
+  - javascript.electron
+  - javascript.wasm
+---
+
+The JavaScript SDK uses a `transport` to send events to Sentry. On modern
+browsers, the default transport uses the browsers `fetch` API to send events.
+
+The default transports drop events if they fail to send due to a lack of connection.
```
lizokm, line 15:
```suggestion

The JavaScript SDK uses a `transport` to send events to Sentry. On modern browsers, most transports use the browsers' `fetch` API to send events. Transports will drop an event if it fails to send due to a lack of connection.
```
Let me know if this changes the meaning too much. If it's important to differentiate between a transport and a "default" transport, it might be good to explain the difference a bit more :)

src/platforms/javascript/common/configuration/transports.mdx
```diff
@@ -0,0 +1,67 @@
...
+dropping events when no connection is available could mean you are missing
+events that occur in that mode.
+
+To enable offline caching of events, `makeBrowserOfflineTransport` can be used
+to wrap existing transports and queue events using the browsers IndexedDB storage.
+Sending of events is then retried later.
+
+<PlatformContent includePath="configuration/offline" />
+
+This adds a number of optional properties to `transportOptions`:
```
lizokm, line 29:
```suggestion
If your JavaScript application is designed to continue working offline,
dropping events when no connection is available and missing
offline events, could mean you're missing important information.

To enable offline events caching, use `makeBrowserOfflineTransport` to wrap existing transports and queue events 
using the browsers' IndexedDB storage. Once your application comes back online, all events will be sent together.

<PlatformContent includePath="configuration/offline" />

Here are a few optional properties you can add to `transportOptions`:
```

Again, please let me know if I've changed the meaning too much.

lizokm (approved):
I've made a few changes to help clarify. Happy to revise if they change the meaning too much :)

AbhiPrasad (approved):

AbhiPrasad:
> Does this need separate platform includes with differing imports for Angular/React/etc/etc?

I think we are fine for now, we can come back and re-evaluate if it becomes too confusing.

PR #6298, opened 2023-02-14
Fix tracingOrigins was renamed to tracePropagationTargets

krystofwoldrich:
Renamed in RN SDK 4.9.0

### DIFF
--- a/src/platform-includes/performance/enable-tracing/react-native.mdx
+++ b/src/platform-includes/performance/enable-tracing/react-native.mdx
@@ -1,6 +1,6 @@
 <Alert level="Info" title="Prior Versions">
 
-If you use a version of our SDK prior to version `3.0.0`, you will need to include the `ReactNativeTracing` integration to use automatic instrumentation. You do not need to do this in versions `3.0.0` and above.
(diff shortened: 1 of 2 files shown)


src/platform-includes/performance/enable-tracing/react-native.mdx
```diff
@@ -1,6 +1,6 @@
 <Alert level="Info" title="Prior Versions">
 
-If you use a version of our SDK prior to version `3.0.0`, you will need to include the `ReactNativeTracing` integration to use automatic instrumentation. You do not need to do this in versions `3.0.0` and above.
+If you use a version of our SDK prior to version `3.0.0`, you will need to include the `ReactNativeTracing` integration to use automatic instrumentation. You do not need to do this in versions `3.0.0` and above. The `tracingOrigins` option was renamed to `tracePropagationTargets` in version `4.9.0`.
```
bitsandfoxes, line 3: Could go on the migration guide instead?

krystofwoldrich: Sure. It's actually already there. 

```suggestion
If you use a version of our SDK prior to version `3.0.0`, you will need to include the `ReactNativeTracing` integration to use automatic instrumentation. You do not need to do this in versions `3.0.0` and above.
```

bitsandfoxes (approved):

PR #6299, opened 2023-02-14
feat(Apple): SentrySwiftUI from cocoapod

brustolin:
Updated the docs on how to install SentrySwiftUI to include cocoapod.

### Legal Boilerplate

### DIFF
--- a/src/platforms/apple/common/performance/instrumentation/swiftui-instrumentation.mdx
+++ b/src/platforms/apple/common/performance/instrumentation/swiftui-instrumentation.mdx
@@ -17,7 +17,7 @@ target 'YourApp' do
 end
 ```
 
-> If you plan to add _SentrySwiftUI_, you don`t need to also add _Sentry_ library since the former dependents on the latter.
+> If you plan to add _SentrySwiftUI_, you don`t need to also add _Sentry_ library since the former depends on the latter.
 
 In order to start monitoring the performance of your views, you need to wrap it with `SentryTracedView`, like this:
 ```swift {tabTitle:Swift}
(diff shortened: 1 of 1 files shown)


kahest (commented):
Seems like this info should be on the CocoaPods page (https://docs.sentry.io/platforms/apple/install/cocoapods/) and we should link there - wdyt?

brustolin:
> Seems like this info should be on the CocoaPods page (https://docs.sentry.io/platforms/apple/install/cocoapods/) and we should link there - wdyt?

Personally I prefer when every information required to start using something is in the same place.
Our [wizard](https://docs.sentry.io/platforms/apple/guides/ios/#install) uses this approach.

src/platforms/apple/common/performance/instrumentation/swiftui-instrumentation.mdx
```diff
@@ -6,7 +6,17 @@ description: "Learn how to monitor the performance of your SwiftUI views."
 
 <Include name="beta-note.mdx" />
 
-You can monitor the performance of your views in a SwiftUI project with the SentrySwiftUI library, which you can install via <PlatformLink to="/install/swift-package-manager/">SPM (Swift package manager)</PlatformLink>.
+You can monitor the performance of your views in a SwiftUI project with the SentrySwiftUI library, which you can install via <PlatformLink to="/install/swift-package-manager/">SPM (Swift package manager)</PlatformLink> or cocoapod:
```
brustolin, line 9:
```suggestion
You can monitor the performance of your views in a SwiftUI project with the SentrySwiftUI library, which you can install via <PlatformLink to="/install/swift-package-manager/">SPM (Swift package manager)</PlatformLink> or CocoaPods:
```

kahest:
> > Seems like this info should be on the CocoaPods page (https://docs.sentry.io/platforms/apple/install/cocoapods/) and we should link there - wdyt?
> 
> Personally I prefer when every information required to start using something is in the same place. Our [wizard](https://docs.sentry.io/platforms/apple/guides/ios/#install) uses this approach.

I can get behind that. Should we also explicitly recommend CocoaPods here?

brustolin:
> Should we also explicitly recommend CocoaPods here? 

Sure.

kahest (approved):
Small suggestions, otherwise LGTM! 🚀

PR #6302, opened 2023-02-14
docs(perf-issues): Add documentation for Slow DB Query Issues

0Calories:
Adds information on the new Slow DB Query Issue type to the docs:

image
image
image


src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+If it is supported by your version of SQL, you can run an `EXPLAIN` command on your query which will present you with a query plan. The plan will lay out the individual steps that your database will perform in order to execute the query, along with accompanying information with each step, such as its cost. The cost is an arbitrary value, but it will be a higher value depending on how "expensive" that portion of the query is. You can contrast the cost of each step to determine which steps of your query are the most expensive. Additionally, you can also run `EXPLAIN ANALYZE`, which will execute the query and report back with the same information as a regular `EXPLAIN`, but also with the actual time taken by each step of your query (be careful not to run this on any destructive queries, as they _will_ be fully executed!).
+
+You can learn more about `EXPLAIN` from the following resources:
+
+- [PostgreSQL Documentation: EXPLAIN](https://www.postgresql.org/docs/current/sql-explain.html)
+- [MySQL Documentation: Optimizing Queries with EXPLAIN](https://dev.mysql.com/doc/refman/8.0/en/using-explain.html)
```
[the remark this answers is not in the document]
k-fish: I think linking to MySql / Postgres should be fine

src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
```diff
@@ -0,0 +1,45 @@
...
+## Detection Criteria
+
+The detector for this performance issue looks at singular database spans with the following criteria:
+
+- The span operation starts with `db` and the query is stored within the span description
+- The query is specifically a `SELECT` query
+- The span duration is at least `1000ms`
+
+Keep in mind that an issue will not be created unless the query in question is _consistently_ taking longer than `1000ms`. More specifically, the issue is created only if the same query is detected as a slow query
+in 100 events within a 24 hour period.
```
k-fish, line 22: I'm a bit reluctant to put the noise thresholds in the docs but we're probably better off being explicit. I think if you do include it you should mention that it's collected only once the `100` event threshold is surpassed, since that might be surprising. This also means that sometimes the issue won't correlate as well to a release since the `100` threshold could delay the first issue creation.

src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+One common cause of slow queries is that the table that is being queried from is un-indexed, so it is recommended to confirm that your table is properly indexed first and foremost. Aside from that, there's no immediately obvious way to fix slow queries, but there are some steps you can take to narrow down what specifically is causing your query to be slow.
+
+If it is supported by your version of SQL, you can run an `EXPLAIN` command on your query which will present you with a query plan. The plan will lay out the individual steps that your database will perform in order to execute the query, along with accompanying information with each step, such as its cost. The cost is an arbitrary value, but it will be a higher value depending on how "expensive" that portion of the query is. You can contrast the cost of each step to determine which steps of your query are the most expensive. Additionally, you can also run `EXPLAIN ANALYZE`, which will execute the query and report back with the same information as a regular `EXPLAIN`, but also with the actual time taken by each step of your query (be careful not to run this on any destructive queries, as they _will_ be fully executed!).
```
k-fish, line 40: iirc MySQL doesn't include cost, it's rows, maybe that's changed in an update

src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
```diff
@@ -0,0 +1,45 @@
...
+- Slow DB Query
+- Duration impact - The duration of the query with respect to the entire transaction, to put into perspective how much of the transaction is being taken up by the query
+
+![Slow DB Query span evidence](span-evidence.png)
+
+View it by going to the **Issues** page in Sentry, selecting your project, clicking on the Slow DB Query issue you want to examine, then scrolling down to the "Span Evidence" section in the "Details" tab.
+
+## Fixing Slow DB Queries
+
+One common cause of slow queries is that the table that is being queried from is un-indexed, so it is recommended to confirm that your table is properly indexed first and foremost. Aside from that, there's no immediately obvious way to fix slow queries, but there are some steps you can take to narrow down what specifically is causing your query to be slow.
```
k-fish, line 38: The more high level gist of why a query is slow is that you are performing too much work on too much data (or there is a resource contention), an index simplifies some of the work by having some of the lookup work done for you, at the cost of more storage etc.

k-fish (approved):
Looks good from a product standpoint

src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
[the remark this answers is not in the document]
DominikB2014: Yeah, we've discussed that before, and external links to offical docs should be good

src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
0Calories: Hmmm I'm not sure anymore, initially I saw [this article](https://www.infoworld.com/article/3234637/7-keys-to-better-mysql-performance.html#:~:text=One%20component%20you%20should%20look,cost%20of%20less%20than%201%2C000.) and it mentions cost, but I don't see cost anywhere in MySQL's official documentation 😕

src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
0Calories: That's true, although we have plans of fixing this so we include all previously collected events once the threshold is passed

src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
lizokm, line 38:
```suggestion

Slow queries often happen when un-indexed tables are queried. To avoid this, make sure that your table is properly indexed. While there aren’t any other obvious fixes, there are steps you can take to narrow down the causes of your query being slow.
```

src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
lizokm, line 40: ### 
```suggestion
If it's supported by your version of SQL, you can run an EXPLAIN command on your query. This will present you with a query plan. The plan will lay out individual steps that your database will perform in order to execute the query, along with additional information, such as its cost. 

The cost is an arbitrary value, but it will fluctuate based on how “expensive” that portion of the query is. You can compare the cost of each step to determine which ones are the most expensive. Additionally, you can run EXPLAIN ANALYZE, which will execute the query and report back with the same information as a regular EXPLAIN, plus the actual time taken by each step of your query (be careful not to run this on any destructive queries, as they /will/ be fully executed!).
```

lizokm (approved):
Tweaked the language a bit, but otherwise this looks good!

src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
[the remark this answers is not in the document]
0Calories: Thanks for the suggestion, gonna commit this: 

it's the same as your suggestion except I removed "that" after "Keep in mind"

```suggestion
Keep in mind, an issue won't be created unless the query in question is _consistently_ taking longer than `1000ms`. More specifically, the issue will be created only if the same query is detected as a slow query
in 100 events within a 24-hour period.
```

src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
```diff
@@ -0,0 +1,45 @@
...
+- The span operation starts with `db` and the query is stored within the span description
+- The query is specifically a `SELECT` query
+- The span duration is at least `1000ms`
+
+Keep in mind that an issue will not be created unless the query in question is _consistently_ taking longer than `1000ms`. More specifically, the issue is created only if the same query is detected as a slow query
+in 100 events within a 24 hour period.
+
+## Span Evidence
+
+The "Span Evidence" section of a Slow DB Query issue contains the following fields:
```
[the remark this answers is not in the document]
0Calories: I think this might be a bit misleading to say, since the Span Evidence section does not point out the root cause of the issue, it just surfaces relevant info. The root cause of the slow query issue runs deeper than that and requires further investigation.

src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
0Calories: I kept your suggestion but restored the markdown symbols for code and italics if that's okay! It helps out with the readability of the documentation 😄 

```suggestion
If it's supported by your version of SQL, you can run an `EXPLAIN` command on your query. This will present you with a query plan. The plan will lay out individual steps that your database will perform in order to execute the query, along with additional information, such as its cost. 

The cost is an arbitrary value, but it will fluctuate based on how “expensive” that portion of the query is. You can compare the cost of each step to determine which ones are the most expensive. Additionally, you can run `EXPLAIN ANALYZE`, which will execute the query and report back with the same information as a regular `EXPLAIN`, plus the actual time taken by each step of your query (be careful not to run this on any destructive queries, as they _will_ be fully executed!).
```

src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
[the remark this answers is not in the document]
lizokm: sg!

src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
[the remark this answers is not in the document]
lizokm: Totally makes sense, but it would be good to point out the benefit of them looking at the span evidence to make this read better for the user. How about something like: You can get additional information about your Slow DB Query by looking at the following fields in the "Span Evidence" section:   ??

src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
lizokm: Absolutely!

src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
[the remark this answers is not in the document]
0Calories: That sounds good to me!

PR #6314, opened 2023-02-16
feat(docs): add search analytics

JonasBa:
Integrate click analytics with algolia so we can get some info on queries and tune them

### DIFF
--- a/src/components/search.tsx
+++ b/src/components/search.tsx
@@ -1,4 +1,5 @@
-import React, { useState, useEffect, useRef } from "react";
+import React, { useState, useEffect, useRef, useCallback } from "react";
+import algoliaInsights from 'search-insights';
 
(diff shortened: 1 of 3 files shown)


src/components/search.tsx
```diff
@@ -1,15 +1,39 @@
...
+      dt = Math.floor(dt/16);
+      return (c=='x' ? r :(r&0x3|0x8)).toString(16);
+  });
+  return uuid;
+}
+
+// Initialize Algolia Insights
+algoliaInsights('init', {
+  appId: 'OOK48W9UCL',
+  apiKey: '2d64ec1106519cbc672d863b0d200782',
```
JonasBa, line 29: This is the same search only API key we expose in global search pkg

lizokm (approved):
LGTM

PR #6329, opened 2023-02-21
add louder alert

smeubank:
adding louder alert

### DIFF
--- a/src/platforms/javascript/common/configuration/tree-shaking/index.mdx
+++ b/src/platforms/javascript/common/configuration/tree-shaking/index.mdx
@@ -23,9 +23,13 @@ To make optional code eligible for tree shaking, you can replace various flags i
 
 `__SENTRY_TRACING__`
 
-: Replacing this flag with `false` will tree shake all code in the SDK that is related to performance monitoring.
-**Attention:** `__SENTRY_TRACING__` must not be replaced with `false` when you're directly using the `@sentry/tracing` package, or when you're using any performance monitoring related SDK features (e.g. `Sentry.startTransaction()`).
-This flag is intended to be used in combination with packages like `@sentry/next` where the `@sentry/tracing` package is used implicitly.
+: Replacing this flag with `false` will tree shake all code in the SDK that is related to performance monitoring, including transcations and tracing.
+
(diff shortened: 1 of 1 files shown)


src/platforms/javascript/common/configuration/tree-shaking/index.mdx
```diff
@@ -23,9 +23,13 @@ To make optional code eligible for tree shaking, you can replace various flags i
...
-: Replacing this flag with `false` will tree shake all code in the SDK that is related to performance monitoring.
-**Attention:** `__SENTRY_TRACING__` must not be replaced with `false` when you're directly using the `@sentry/tracing` package, or when you're using any performance monitoring related SDK features (e.g. `Sentry.startTransaction()`).
-This flag is intended to be used in combination with packages like `@sentry/next` where the `@sentry/tracing` package is used implicitly.
+: Replacing this flag with `false` will tree shake all code in the SDK that is related to performance monitoring, including transcations and tracing.
+
+<Alert level="info" title="Using Performance or tracing?">
+  
+__SENTRY_TRACING__` must not be replaced with `false` when you're directly using the `@sentry/tracing` package, or when you're using any performance monitoring related SDK features (e.g. `Sentry.startTransaction()`). This flag is intended to be used in combination with packages like `@sentry/next` where the `@sentry/tracing` package is used implicitly.
+
+</Alert>
```
lizokm, line 32:
```suggestion
<Alert level="info" title="Not for use with performance monitoring SDK features or tracing package.">
  
__SENTRY_TRACING__` shouldn't be replaced with `false` when you're directly using the `@sentry/tracing` package or any performance-monitoring-related SDK features (such as `Sentry.startTransaction()`). This flag is intended to be used in combination with packages like `@sentry/next` where the `@sentry/tracing` package is used implicitly.

</Alert>
```
It may just be me, but I'm still pretty unclear on when to use this flag vs. not after reading this note.

Is the difference between directly using the @sentry/tracing vs implicitly using it clear enough to the average developer? Also, will they understand using performance monitoring SDK features versus other types of performance monitoring? If not, it might be good to rephrase this or add some context as to when to use this versus not.

lizokm (commented):
Left a question.

PR #6332, opened 2023-02-21
feat(replay): Update Replay PII images (masking/ignore/blocking)

### LINKED ISSUE #5978: Docs: Update screen shots on docs/privacy page so they're clearer and more obviously represent a customers app (not sentry-in-sentry)
>[...] more illustrative/abstract route for the supplemental screenshots? i found it a tiny bit disorienting seeing the Sentry UI as an illustration of what I should expect to see in my own UI
> i think the scale of this is a bit off, but i think something like this would've helped it click a hair sooner

Image

The only places where we have images right now is on the privacy page: https://docs.sentry.io/platforms/javascript/session-replay/privacy/ and on the marketing page: https://sentry.io/for/session-replay/

The Docs page should be updated first to make it clear that it's a customers app being masked/blocked out.

ryan953:
Fixes #5978

In-situ:

localhost_3001_platforms_javascript_session-replay_privacy_


billyvg (approved):
Looks great

lizokm (approved):
LGTM!

PR #6336, opened 2023-02-22
feat(Performance): Add time to initial display and time to full display spans

stefanosiano:
Android (and others soon) has now the `time-to-initial-display` and `time-to-full-display` spans in ui Instrumentation.


src/platforms/android/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -342,3 +342,49 @@ fun loadUserDataOnClick() {
...
+
+```kotlin
+import io.sentry.Sentry
+
+Sentry.reportFullDisplayed()
+```
+
+Time to initial display depends on having an active transaction bound to the scope, and ideally it'd be used along with [Android's Activity Instrumentation](/platforms/android/performance/instrumentation/automatic-instrumentation/#androids-activity-instrumentation), which starts a transaction and binds it to the scope automatically.
+The span `status` is set to `SpanStatus.OK` if it's finished through the API. If the span isn't finished after 30 seconds it gets automatically finished by the SDK and its `status` is set to `SpanStatus.DEADLINE_EXCEEDED`.
+If the span is finished before the Activity's is first drawn and displayed as measured by the `Time to initial display`, the reported time here will be shifted to the `Time to initial display` measured time.
```
lizokm, line 390:
```suggestion
Time to initial display is dependent on having an active transaction bound to the scope. Ideally, it should be used along with [Android's Activity Instrumentation](/platforms/android/performance/instrumentation/automatic-instrumentation/#androids-activity-instrumentation), which starts a transaction and binds it to the scope automatically. 

If the span finishes through the API, its `status` is set to `SpanStatus.OK`. If the span doesn't finish after 30 seconds, it is finished by the SDK automatically, and its `status` is set to `SpanStatus.DEADLINE_EXCEEDED`.
If the span finishes before the Activity is first drawn and displayed as measured by the `Time to initial display`, the reported time will be shifted to `Time to initial display` measured time.
```

lizokm (approved):
Made a few changes.

PR #6339, opened 2023-02-22
Remove the EA note from file IO

wmak:
- Removes the EA note from file io, planning to GA it once this merges: https://github.com/getsentry/getsentry/pull/9610/files


lizokm (approved):
This LGTM!

PR #6342, opened 2023-02-23
feat(view-hierarchy): Update View Hierarchy section in SDK docs

narsaynorath:
Add information r.e. the view hierarchy feature in Issue Details and include react native . I call out some use cases and add a screenshot as an example.


wmak (approved):
lgtm

lizokm:
@narsaynorath Hi! Yes, it makes sense to only add to the SDK for now since it's not a lot of information. As for the screenshot, yes, if possible it'd be great to either have a different screenshot for each SDK or if you could just mask out the apple symbol, you could use that same one for all of them.

narsaynorath:
Thanks for the feedback @lizokm! I've gone with a change similar your suggestion of masking out the icon and replaced it with our generic platform icon, so it shouldn't be evidently tied to a specific SDK.

PR #6364, opened 2023-02-25
docs: adding url filtering example for python

iloveitaly:
### Legal Boilerplate


antonpirker (changes requested):
Hey @iloveitaly 

Thanks for the contribution! Great idea about showing how to filter out health checks, because this is a very common use case.

Could you please add your example instead to the bottom of this file: https://github.com/getsentry/sentry-docs/blob/master/src/platform-includes/configuration/before-send-transaction/python.mdx

(because our docs are intertwined with all the platforms we support, and this way it will only show up in the Python docs and not also in all the other platforms. Thanks!)

cleptric:
Quickly did this in https://github.com/getsentry/sentry-docs/pull/6509

PR #6375, opened 2023-02-27
feat(perf-issues): Add MN+1 Queries docs

mjq-sentry:
Our new MN+1 Query detector creates N+1 Query issues, the same as our existing N+1 Query detector. To a user's perspective, they will just see more types of N+1s being detected. It still seemed useful to give this pattern its own section on the N+1 Query page as a way to distinguish this pattern from the primary N+1 pattern matched by the existing detector.


src/docs/product/issues/issue-details/performance-issues/n-one-queries.mdx
```diff
@@ -64,3 +64,34 @@ def books(request):
...
+    book_list = [book.title + " by " + book.author.name + ", " + book.publisher.name for book in books]
+    return HttpResponse((", ").join(book_list))
+```
+
+This code will perform two database queries for each iteration of the loop over books: one to fetch the author, and another to fetch the publisher. Since there are two queries in each iteration of the loop, we could call this a 2N+1 issue. Sentry will find and report these issues, too. They are categorized in the application as N+1 Queries, but may involve multiple repeating spans. MN+1 queries can be solved using similar strategies to N+1s, as [described above](#example).
+
+### Detection Criteria
+
+The detector for performance issues looks for a set of sequential spans that includes at least one database span and follows a recurring pattern. It also uses the following criteria:
+
```
lizokm, line 88:
```suggestion
The detector for performance issues looks for a set of sequential spans, including at least one database span, and follows a recurring pattern. It also looks for the following criteria:

```

src/docs/product/issues/issue-details/performance-issues/n-one-queries.mdx
```diff
@@ -64,3 +64,34 @@ def books(request):
...
+- Total duration of involved spans must exceed a threshold (usually 100 ms)
+- Total count of involved spans must exceed a threshold (usually ten spans)
+- At least one involved spans must be a database span with a full query as its descriptions (some SDKs truncate queries and add an ellipsis to the end)
```
lizokm, line 91:
```suggestion
- At least one involved span must be a database span with a full query as its description. (Some SDKs truncate queries and add an ellipsis at the end.)
```

src/docs/product/issues/issue-details/performance-issues/n-one-queries.mdx
```diff
@@ -64,3 +64,34 @@ def books(request):
...
+
+If Sentry is not detecting an MN+1 issue where you expect one, it's probably because the transaction didn't meet one of the above criteria.
```
lizokm, line 93:
```suggestion
If Sentry isn't detecting an MN+1 issue where you expected one, it's probably because the transaction didn't meet one of the above criteria.
```

src/docs/product/issues/issue-details/performance-issues/n-one-queries.mdx
```diff
@@ -64,3 +64,34 @@ def books(request):
...
+
+### Fingerprinting
+
+Sentry computes the fingerprint for MN+1 issues similarly to standard N+1s [described above](#fingerprinting). MN+1s do not require or use a source span in their fingerprints - instead they are based only on the parent span and the repeating spans. Repeating spans may be a combination of `db` and non-`db` spans. `db` spans are fingerprinted by their operation and the query in their span description, while non-`db` spans are fingerprinted by their span operation.
```
lizokm, line 97:
```suggestion
Because both are based on parent and repeating spans, and don't require a source span, Sentry computes the fingerprints of MN+1 and standard [N+1s](#fingerprinting) similarly. 

```
I think this section could use a bit more context, such as why this information is important to the user. Kind of like the fingerprinting section for the N+1 queries. If we're just trying to list different kinds of spans, it may be better to do it as a table. Happy to chat about this more, maybe I'm just misunderstanding something.

lizokm (commented):
I've made some edits, and would love to talk a bit more about the "fingerprinting" section. Happy to have a quick video chat :)

src/docs/product/issues/issue-details/performance-issues/n-one-queries.mdx
```diff
@@ -64,3 +64,34 @@ def books(request):
 Django will `JOIN` the tables ahead of time, and preload the author information. That way, calling `book.author.name` does not need to make an extra query. Instead of a long waterfall, there is a single `SELECT` query:
 
 ![Solved N+1 queries in an example application](n-plus-one-queries-after.png)
+
+## MN+1 Queries
+
+<Include name="early-adopter-note.mdx" />
```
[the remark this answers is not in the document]
mjq-sentry: For sure, done!

src/docs/product/issues/issue-details/performance-issues/n-one-queries.mdx
```diff
@@ -64,3 +64,34 @@ def books(request):
...
+from django.http import HttpResponse
+
+def books(request):
+    books = Book.objects.select_related("author").all()[:10]
+    book_list = [book.title + " by " + book.author.name + ", " + book.publisher.name for book in books]
+    return HttpResponse((", ").join(book_list))
+```
+
+This code will perform two database queries for each iteration of the loop over books: one to fetch the author, and another to fetch the publisher. Since there are two queries in each iteration of the loop, we could call this a 2N+1 issue. Sentry will find and report these issues, too. They are categorized in the application as N+1 Queries, but may involve multiple repeating spans. MN+1 queries can be solved using similar strategies to N+1s, as [described above](#example).
+
```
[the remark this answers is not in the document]
mjq-sentry: > Can we use N+1 queries, N+1 issues, and N+1s interchangeably?

* Colloquially (outside of Sentry), these are typically called N+1 DB queries, N+1 queries, or just N+1s. Once we establish the context that we're dealing with N+1 DB queries (as opposed to N+1 API calls or some other N+1 pattern), I think N+1 is reasonable as a shorthand.
* The label we've applied in the UI for issues of this type is "N+1 Query".
* Internally we've tried to avoid the word "issue" unless we actually mean an Issue (in the Sentry sense). We've been using the word "problem" as a substitute when "issue" would be ambiguous. (Sentry detects a performance problem in your code, and creates a Performance Issue to inform you and track its resolution).

I didn't do a particularly good job of following the above when writing the updates to this doc however 😅 

Does that make sense as a baseline?

> Does it make sense to say; "solving an N+1 issue?"

I think so, with the caveat above that "issue" can be confusing. If we mean a problem identified by Sentry (i.e. there exists an Issue for it) then it's fine; if we mean solving an issue in general, "solving an N+1 problem" might be less ambiguous.

PR #6376, opened 2023-02-27
chore(docs): Bump min python sdk for profiling to 1.16.0

Zylphrex:
1.16.0 contains some bug fixes that ensure things work out of the box on more configurations.


lizokm (approved):
LGTM!

PR #6381, opened 2023-02-28
feat(algolia): send analytics tags

JonasBa:
Send analytics tags so we can distinguish from different sources of traffic in algolia analytics


lizokm (approved):
This LGTM!

PR #6388, opened 2023-03-01
Chore(wizard)react

Jesse-Box:
- Each step is broken down
- Describes what the instructions will set up

### Legal Boilerplate

### DIFF
--- a/src/wizard/javascript/react.md
+++ b/src/wizard/javascript/react.md
@@ -22,7 +22,7 @@ npm install --save @sentry/react @sentry/tracing
 ```
 
 ## Configure
-Initialise Sentry as early as possible in your application's lifecycle.
+Initialize Sentry as early as possible in your application's lifecycle.
 
 ```javascript
 import { createRoot } React from "react-dom/client";
--- a/src/wizard/javascript/react.md
+++ b/src/wizard/javascript/react.md
@@ -43,7 +43,7 @@ root.render(<App />)
 ```
 
 > **tracesSampleRate: 1.0**
-> The example above ensures every transaction will be to Sentry, but we recommend lowering this value in production.
+> The above setting ensures that 100% of transactions will be sent to Sentry, but we recommend lowering this number in production.
 
 ## Verify
 This snippet includes an intentional error, so you can test that everything is working as soon as you set it up.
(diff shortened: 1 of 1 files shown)


src/wizard/javascript/react.md
```diff
@@ -15,39 +21,39 @@ yarn add @sentry/react @sentry/tracing
 npm install --save @sentry/react @sentry/tracing
 ```
 
-Next, import and initialize the Sentry module as early as possible, before initializing React:
+## Configure
+Initialise Sentry as early as possible in your application's lifecycle.
```
lizokm, line 25:
```suggestion
Initialize Sentry as early as possible in your application's lifecycle.
```
Same comment as before - I'm wondering if "application's lifecycle" is a bit vague? Maybe it's just vague to me, but I'm wondering if there's a way to call this something more specific?

src/wizard/javascript/react.md
```diff
@@ -15,39 +21,39 @@ yarn add @sentry/react @sentry/tracing
...
+const container = document.getElementById(“app”);
+const root = createRoot(container);
+root.render(<App />)
 ```
 
-The above configuration captures both error and performance data. To reduce the volume of performance data captured, change `tracesSampleRate` to a value between 0 and 1.
-
-After this step, Sentry will report any uncaught exceptions triggered by your application.
+> **tracesSampleRate: 1.0**
+> The example above ensures every transaction will be to Sentry, but we recommend lowering this value in production.
```
lizokm, line 46:
```suggestion
> The above setting ensures that 100% of transactions will be sent to Sentry, but we recommend lowering this number in production.
```

src/wizard/javascript/react.md
```diff
@@ -15,39 +21,39 @@ yarn add @sentry/react @sentry/tracing
...
 
-You can trigger your first event from your development environment by raising an exception somewhere within your application. An example of this would be rendering a button whose `onClick` handler attempts to invoke a method that does not exist:
+## Verify
+This snippet includes an intentional error, so you can test that everything is working as soon as you set it up.
```
lizokm, line 49: This snippet contains an intentional error and can be used as a test to make sure that everything's working as expected.

lizokm (approved):
Have a few suggestions.

PR #6411, opened 2023-03-03
nit pick

smeubank:
less text, more bold

### DIFF
--- a/src/wizard/javascript/replay-onboarding/nextjs/2.configure.md
+++ b/src/wizard/javascript/replay-onboarding/nextjs/2.configure.md
@@ -9,7 +9,7 @@ type: language
 
 Add the following to your SDK config. Several privacy and sampling options are available, all of which can be set using the `integrations` constructor. Learn more about configuring Session Replay by reading the [configuration docs](https://docs.sentry.io/platforms/javascript/guides/nextjs/session-replay/).
 
-Note: The Replay integration *must* be added to your *`sentry.client.config.js`* file. Adding it into `sentry.server.config.js` or `sentry.edge.config.js` will do nothing.
+Alert: The Replay integration *must* be added to your *`sentry.client.config.js`* file. Adding it into `sentry.server.config.js` or `sentry.edge.config.js` may break your build.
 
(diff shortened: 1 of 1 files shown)


Lms24 (approved):

src/wizard/javascript/replay-onboarding/nextjs/2.configure.md
```diff
@@ -7,7 +7,9 @@ type: language
 
 #### Configure
 
-Add the following to your SDK config. There are several privacy and sampling options available, all of which can be set using the `integrations` constructor. Learn more about configuring Session Replay by reading the [configuration docs](https://docs.sentry.io/platforms/javascript/guides/nextjs/session-replay/).
+Add the following to your SDK config. Several privacy and sampling options are available, all of which can be set using the `integrations` constructor. Learn more about configuring Session Replay by reading the [configuration docs](https://docs.sentry.io/platforms/javascript/guides/nextjs/session-replay/).
+
+Note: The Replay integration *must* be added to your *`sentry.client.config.js`* file. Adding it into `sentry.server.config.js` or `sentry.edge.config.js` will do nothing.
```
mydea, line 12: I am actually not 100% sure, but maybe we better not word this as `will do nothing` but as `will break your build`? This sounds as if you could add it and it will just be ignored, but IMHO it should def. not be added.

mydea: To be clear, it _should_ actually just be ignored, but maybe from a docs perspective it is safer to make it clear that people should really not be adding this there?

lizokm, line 12:
```suggestion
Alert: The Replay integration *must* be added to your *`sentry.client.config.js`* file. Adding it into `sentry.server.config.js` or `sentry.edge.config.js` may break your build.
```

lizokm: Bringing the DRAMA!

lizokm (approved):
LGTM

PR #6427, opened 2023-03-06
(Performance) reconsider onboarding wording

smeubank:
This is a shared component for documentation. So merging this would impact all SDKs across the docs which reference this page.

This PR proposes 2 changes:
1. around turning on tracing and configuring a sample rate
Is:

> Sampling for transactions must also be configured before tracing

To be (something simpler):
> Enable tracing and configure the sampling rate for transacations with....

2. dynamic sampling has become too loaded
Updating to avoid that this is not conflated with "dynamic sampling" which is something on server side, and instead proposing to call it "selective" sampling, based on contextual data, Which is more accurate to what it actually is

[Example:](https://sentry-docs-git-smeubank-perf-tracing-config.sentry.dev/platforms/javascript/guides/react/performance/#configure-the-sample-rate)

### DIFF
--- a/src/platforms/common/performance/index.mdx
+++ b/src/platforms/common/performance/index.mdx
@@ -94,11 +94,11 @@ Test out tracing by starting and finishing a transaction, which you _must_ do so
 
 While you're testing, set <PlatformIdentifier name="traces-sample-rate" /> to `1.0`, as that ensures that every transaction will be sent to Sentry.
 
-Once testing is complete, **we recommend lowering this value in production** by either lowering your <PlatformIdentifier name="traces-sample-rate" /> value, or switching to using <PlatformIdentifier name="traces-sampler" /> to selectively sample and filter your transactions, based on contextual data.
+Once testing is complete, you may want to consider setting a lower <PlatformIdentifier name="traces-sample-rate" /> value, or switching to using <PlatformIdentifier name="traces-sampler" /> to selectively sample and filter your transactions, based on contextual data. 
 
 <PlatformSection supported={["javascript"]} notSupported={["react-native"]}>
 
-Leaving the sample rate at `1.0` means that automatic instrumentation will send a transaction each time a user loads any page or navigates anywhere in your app, which is a lot of transactions. Sampling enables you to collect representative data without overwhelming either your system or your Sentry transaction quota.
+Leaving the sample rate at `1.0` means that automatic instrumentation will send a transaction each time a user loads any page or navigates anywhere in your app, which can be a lot of transactions. Sampling enables you to collect representative data without overwhelming either your system or your Sentry transaction quota. For lower traffic applications there may not be a concern at all. It is mainly of interest to high load backend applications, where you may want to explore sampling consider load on the network or system resources are high.
 
 </PlatformSection>
 
--- a/src/platforms/common/performance/index.mdx
+++ b/src/platforms/common/performance/index.mdx
@@ -94,7 +94,7 @@ Test out tracing by starting and finishing a transaction, which you _must_ do so
 
 While you're testing, set <PlatformIdentifier name="traces-sample-rate" /> to `1.0`, as that ensures that every transaction will be sent to Sentry.
 
-Once testing is complete, **we recommend lowering this value in production** by either lowering your <PlatformIdentifier name="traces-sample-rate" /> value, or switching to using <PlatformIdentifier name="traces-sampler" /> to dynamically sample and filter your transactions.
+Once testing is complete, **we recommend lowering this value in production** by either lowering your <PlatformIdentifier name="traces-sample-rate" /> value, or switching to using <PlatformIdentifier name="traces-sampler" /> to selectively sample and filter your transactions, based on contextual data.
 
 <PlatformSection supported={["javascript"]} notSupported={["react-native"]}>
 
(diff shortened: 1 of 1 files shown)


src/platforms/common/performance/index.mdx
```diff
@@ -67,7 +67,7 @@ Automatic instrumentation for monitoring the performance of your application is
 
 ## Configure the Sample Rate
 
-Sampling for transactions must also be configured before tracing is enabled in your app. Set the sample rate for your transactions by either:
+First, enable both tracing and configure the sampling rate for transactions. Set the sample rate for your transactions by either:
```
lizokm, line 70:
```suggestion
Before tracing can be enabled, you have to set the sample rate for your transactions by either:
```
Can sample rates be set BEFORE the tracing package is installed? If so, it would make sense to switch the order and move "configure sample rate" above "enable tracing". It seems illogical to tell people to do something and then say, before that can be done, you should do this.

src/platforms/common/performance/index.mdx
```diff
@@ -94,11 +94,11 @@ Test out tracing by starting and finishing a transaction, which you _must_ do so
 
 While you're testing, set <PlatformIdentifier name="traces-sample-rate" /> to `1.0`, as that ensures that every transaction will be sent to Sentry.
 
-Once testing is complete, **we recommend lowering this value in production** by either lowering your <PlatformIdentifier name="traces-sample-rate" /> value, or switching to using <PlatformIdentifier name="traces-sampler" /> to dynamically sample and filter your transactions.
+Once testing is complete, you may want to consider setting a lower <PlatformIdentifier name="traces-sample-rate" /> value, or switching to using <PlatformIdentifier name="traces-sampler" /> to selectively sample and filter your transactions, based on contextual data.
```
lizokm, line 97:
```suggestion
Once testing is complete, you may want to set a lower <PlatformIdentifier name="traces-sample-rate" /> value, or switch to using <PlatformIdentifier name="traces-sampler" /> to selectively sample and filter your transactions, based on contextual data. 
```

In case you want a record of Ian's comments: 
"Lets just say something more balanced >
Like for small applicaitons under X TxNs you may not need any sampling.  For FE apps there is no scale concern and sampling is optional.  For extermely hhigh scale BE apps, you may wish to explore sampling if load on the network or system resources are high."

src/platforms/common/performance/index.mdx
```diff
@@ -94,11 +94,11 @@ Test out tracing by starting and finishing a transaction, which you _must_ do so
...
 
 <PlatformSection supported={["javascript"]} notSupported={["react-native"]}>
 
-Leaving the sample rate at `1.0` means that automatic instrumentation will send a transaction each time a user loads any page or navigates anywhere in your app, which is a lot of transactions. Sampling enables you to collect representative data without overwhelming either your system or your Sentry transaction quota.
+Leaving the sample rate at `1.0` means that automatic instrumentation will send a transaction each time a user loads any page or navigates anywhere in your app, which can be a lot of transactions. Sampling enables you to collect representative data without overwhelming either your system or your Sentry transaction quota. For lower traffic applications there may not be a concern at all. It is mainly of interest to high load backend applications, where you may want to explore sampling consider load on the network or system resources are high.
```
lizokm, line 101:
```suggestion
If you leave your sample rate at `1.0`, a transaction will be sent every time a user loads a page or navigates within your app. Depending on the amount of traffic your application gets, this may mean a lot of transactions. If you have a high-load, backend application, you may want to consider setting a lower <PlatformIdentifier name="traces-sample-rate" /> value, or switching to using <PlatformIdentifier name="traces-sampler" /> to selectively sample and filter your transactions, based on contextual data.
```
I recommend consolidating the above paragraph with this one into one.

src/platforms/common/performance/index.mdx
lizokm: I recommend consolidating this paragraph and the one below it into one.

lizokm (approved):
Made some suggestions. Happy to talk through this more :)

src/platforms/common/performance/index.mdx
smeubank: essentially setting a sample rate (directly or indirectly) does both at the same time, there is no really first or before

the tracing package is the performance package, generates spans/transactions and propagates traces

but i understand it is confusing, I would stick with "first, do this to do both..."

src/platforms/common/performance/index.mdx
```diff
@@ -94,11 +94,11 @@ Test out tracing by starting and finishing a transaction, which you _must_ do so
 
 While you're testing, set <PlatformIdentifier name="traces-sample-rate" /> to `1.0`, as that ensures that every transaction will be sent to Sentry.
```
ale-cota, line 95: I have a suggestion here - how about we introduce what sampling does generically first and then go into recommended settings for pre-prod vs. prod environments? 

"Sampling enables you to collect representative data without overwhelming your system or your [transaction quota] (/product/accounts/quotas/). While you're testing, you're likely doing so in a pre-production environment with limited traffic, so we recommend setting <PlatformIdentifier name="traces-sample-rate" /> to `1.0`, as that ensures that every transaction will be sent to Sentry.

Once testing is complete, you may consider changing your sampling configuration when rolling out to production if overhead is a concern. If you have a high-load, backend application, you may want to consider setting a lower <PlatformIdentifier name="traces-sample-rate" /> value, or switching to using <PlatformIdentifier name="traces-sampler" /> to selectively sample and filter your transactions, based on contextual data."

ale-cota (changes requested):

PR #6432, opened 2023-03-07
chore(android): Disable NDK integration through Gradle

romtsn:


src/platforms/android/using-ndk.mdx
```diff
@@ -96,7 +96,21 @@ You can disable the NDK integration by adding the following to your `AndroidMani
 You can use Sentry's Android SDK without the Android Native Development Kit (NDK) by either:
 
 - [Disabling the NDK](#disable-ndk-integration)
-- Using `sentry-android-core`, which doesn't contain the NDK and can be used separately, instead of `sentry-android` when adding the dependency. The minimal required API level for `sentry-android-core`  is 14.
+- Using `sentry-android-core`, which doesn't contain the NDK and can be used separately, instead of `sentry-android` when adding the dependency. The minimal required API level for `sentry-android-core` is 14.
+
+If you're using our [Gradle plugin](/platforms/android/gradle/), it will still pull the NDK integration in as part of the auto-installation feature. To disable, you can remove the `sentry-android-ndk` dependency from the app configurations in `app/build.gradle`:
+
```
lizokm, line 102:
```suggestion
- Using `sentry-android-core`, which doesn't contain the NDK and can be used separately instead of `sentry-android`, when adding the dependency. The minimal required API level for `sentry-android-core` is 14.

If you're using our [Gradle plugin](/platforms/android/gradle/), it'll still pull the NDK integration in as part of the auto-installation feature. To disable it, remove the `sentry-android-ndk` dependency from the app configurations in `app/build.gradle`:

```

lizokm (approved):
LGTM

PR #6439, opened 2023-03-07
[Python] Improve "custom instrumentation" docs

antonpirker:
Made it easier comprehensible (less special words) 
Also added the new way of adding spans (with the decorator)

Before:
https://docs.sentry.io/platforms/python/performance/instrumentation/custom-instrumentation/

Now:
https://sentry-docs-git-antonpirker-performance-custom-instrumentation.sentry.dev/platforms/python/performance/instrumentation/custom-instrumentation/

### DIFF
--- a/src/platforms/python/common/performance/instrumentation/custom-instrumentation.mdx
+++ b/src/platforms/python/common/performance/instrumentation/custom-instrumentation.mdx
@@ -12,7 +12,7 @@ Adding transactions will allow you to instrument and capture certain regions of 
 
 <Note>
 
-If you are using one of the Sentry SDK integrations, those integrations will create transactions for you!
+If you're using one of Sentry's SDK integrations, transactions will be created for you automatically.
 
 </Note>
 
--- a/src/platforms/python/common/performance/instrumentation/custom-instrumentation.mdx
+++ b/src/platforms/python/common/performance/instrumentation/custom-instrumentation.mdx
@@ -35,7 +35,7 @@ def eat_pizza(pizza):
 If you want to have more fine grained performance monitoring, you can add so called child spans to your transaction. You can do so in multiple ways:
 
 - Using a context manager or
-- Using a decorator
+- Using a decorator (works on sync and `async` functions.)
 
 Calling `sentry_sdk.start_span()` will find the current active transaction, and attach the span to it.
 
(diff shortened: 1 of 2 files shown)


src/platforms/python/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -0,0 +1,143 @@
...
+@sentry_sdk.trace
+def eat_slice(slice):
+    chew()
+    swallow()
+```
+
+## Accessing the Current Transaction
+
+In cases where you want to change data in an already ongoing Transaction you can use `Hub.current.scope.transaction`. This property will return a Transaction in case there is a running Transaction otherwise it returns `None`.
+
```
lizokm, line 112:
```suggestion
To change data in an already ongoing transaction, use `Hub.current.scope.transaction`. This property will return a transaction if there's one running, otherwise it will return `None`.

```

src/platforms/python/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -0,0 +1,143 @@
...
+    while pizza.slices > 0:
+        eat_slice(pizza.slices.pop())
+```
+
+## Accessing the Current Span
+
+In cases where you want to change data in the current Span you can use `sentry_sdk.Hub.current.scope.span`. This property will return a Span in case there is a running Span otherwise it returns `None`.
+
+In this example we will set a tag in the Span created by the `@sentry_sdk.trace` decorator.
+
```
lizokm, line 131:
```suggestion

To change data in the current span, use `sentry_sdk.Hub.current.scope.span`. This property will return a span if there's one running, otherwise it will return `None`.

In this example, we'll set a tag in the span created by the `@sentry_sdk.trace` decorator.

```

lizokm (approved):
Made a few suggestions.

sl0thentr0py:
@antonpirker also update this please, it gets injected into `Missing Instrumentation` spans in the product.
https://github.com/getsentry/sentry-docs/blob/master/src/wizard/python/tracing.md

antonpirker:
Is there a way to see the wizard content in the documentation (with a secret URL or something?) Does anyone know @sl0thentr0py or @lizokm ?

The "create new project" wizard is displayed on Sentry.io without doing a request to the documentation, so I guess this is baked in at build time..

PR #6448, opened 2023-03-09
fix/rate-limiting-language

souredoutlook:
Clarify how concurrent rate limiting works

### DIFF
--- a/src/api/ratelimits.mdx
+++ b/src/api/ratelimits.mdx
@@ -26,9 +26,9 @@ Every API request response includes the following headers:
 - `X-Sentry-Rate-Limit-Reset`
   - The time when the next rate limit window begins and the count resets, measured in UTC seconds from epoch
 - `X-Sentry-Rate-Limit-ConcurrentLimit`
-  - The maximum number of concurrent requests allowed within the window
+  - The maximum number of concurrent ("in flight") requests allowed at any given time
 - `X-Sentry-Rate-Limit-ConcurrentRemaining`
-  - The number of concurrent requests this caller has left on this endpoint within the current window
+  - The number of concurrent requests this caller has left
 
 ## Additional Information
 


src/api/ratelimits.mdx
```diff
@@ -26,9 +26,9 @@ Every API request response includes the following headers:
 - `X-Sentry-Rate-Limit-Reset`
   - The time when the next rate limit window begins and the count resets, measured in UTC seconds from epoch
 - `X-Sentry-Rate-Limit-ConcurrentLimit`
-  - The maximum number of concurrent requests allowed within the window
+  - The maximum number of concurrent ("in flight") requests allowed at any given time
```
nhsiehgit, line 29: 🤔 

'concurrent' does not mean 'in flight', just means happening at the same time.
So this _could_ just be `The maximum number of concurrent, in flight requests allowed at any given time`. But this also becomes slightly redundant since `in flight` and any given time are both sort of timing-specific participles 🤔 

'concurrent' and 'at the same time' are redundant too.
in-flight is a given and is unnecessary as any request in the pipeline is "in flight" already.

src/api/ratelimits.mdx
```diff
@@ -26,9 +26,9 @@ Every API request response includes the following headers:
...
 - `X-Sentry-Rate-Limit-ConcurrentRemaining`
-  - The number of concurrent requests this caller has left on this endpoint within the current window
+  - The number of concurrent requests this caller has left
```
nhsiehgit, line 31: this makes it sound like there's a finite pool here that is being filled.

`within the current window` makes it more clear that this is a queue.

nhsiehgit (commented):
some comments.

I don't think these copy changes make this any more clear :(

src/api/ratelimits.mdx
nhsiehgit: if the confusion is around the term `concurrent` we should just remove that from the doc and clarify it in more laymans terms

src/api/ratelimits.mdx
volokluev: There is indeed a finite pool being filled. here is the implementation, let me know if you have any questions:

https://github.com/getsentry/sentry/blob/master/src/sentry/scripts/ratelimits/api_limiter.lua

Here is the design document:

https://www.notion.so/sentry/Concurrent-Rate-Limiter-ebd5b7562aed4d01b102d83774a06b11

src/api/ratelimits.mdx
souredoutlook: The confusion around concurrent rate limiting typically stems from the fact that it is not actually window based. 

In flight is redundant for sure (the specific report was resolved by explaining these are all requests currently in-progress/in-flight).

I am open for using whatever describes concurrency in more laymans terms.

The root problem is the original language "within the window"

src/api/ratelimits.mdx
souredoutlook: Any updates on where we want to go with this particular change?

nhsiehgit: Ah sure, there's a finite amount of concurrent requests, but its still a queue right? so it will replenish?
Maybe we could tack on "caller has left in the queue"?

nhsiehgit (approved):

Issue #6459, opened 2023-03-12
Sentry's Vite documentation suggests leaking private authentication tokens


Manouchehri:
### Core or SDK?

Core Sentry product

### Which part? Which one?

@sentry/vite-plugin

### Description

Creating a `VITE_SENTRY_AUTH_TOKEN` will cause the the private authentication token to be exposed publicly. This behavior of `VITE_` prefixed variables is documented and expected. https://vitejs.dev/guide/env-and-mode.html#env-files

```typescript
import { defineConfig, loadEnv } from "vite";
import { sentryVitePlugin } from "@sentry/vite-plugin";

export default defineConfig(({ command, mode }) => {
  // Load env file based on `mode` in the current working directory.
  // Set the third parameter to '' to load all env regardless of the `VITE_` prefix.
  const env = loadEnv(mode, process.cwd(), '')

  return {
    build: {
      sourcemap: true, // Source map generation must be turned on
    },
    plugins: [
      // Put the Sentry vite plugin after all other plugins
      sentryVitePlugin({
        org: "example-org",
        project: "example-project",

        // Specify the directory containing build artifacts
        include: "./dist",

        // Auth tokens can be obtained from https://sentry.io/settings/account/api/auth-tokens/
        // and needs the `project:releases` and `org:read` scopes
        authToken: env.VITE_SENTRY_AUTH_TOKEN, // <------ THIS WILL BE EXPOSED PUBLICLY DUE TO THE VITE_ PREFIX

        // Optionally uncomment the line below to override automatic release name detection
        // release: env.RELEASE,
      }),
    ],
  }
});
```
Source: https://docs.sentry.io/platforms/node/guides/connect/sourcemaps/uploading/vite/

### Suggested Solution

The documentation over here is fine, it uses `process.env.SENTRY_AUTH_TOKEN` instead (which is safe as far as I know).

https://github.com/getsentry/sentry-javascript-bundler-plugins/tree/main/packages/vite-plugin

mydea:
Hey, thanks for reporting this! You are right, we will update the docs accordingly.

PR #6460, opened 2023-03-13
fix: Do not recommend using `env.VITE_XXX`

mydea:
As anything prefixed with `VITE_` will be made available to the client, which we do not want for the auth token - see:
https://vitejs.dev/guide/env-and-mode.html#env-files

Closes https://github.com/getsentry/sentry-docs/issues/6459


Lms24 (approved):
LGTM

PR #6469, opened 2023-03-14
fix: in-app docs sentry.io reference

vaind:


src/platforms/common/configuration/options.mdx
```diff
@@ -297,7 +297,7 @@ This value is used only if Sentry can not find the origin of the frame.
 
 A list of string prefixes of module names that belong to the app. This option takes precedence over `in-app-exclude`.
 
-Sentry differentiates stack frames that are directly related to your application ("in application") from stack frames that come from other packages such as the standard library, frameworks, or other dependencies. The application package is automatically marked as `inApp`. The difference is visible in the sentry.io, where only the "in application" frames are displayed by default.
+Sentry differentiates stack frames that are directly related to your application ("in application") from stack frames that come from other packages such as the standard library, frameworks, or other dependencies. The application package is automatically marked as `inApp`. The difference is visible on [sentry.io](https://sentry.io), where only the "in application" frames are displayed by default.
```
lizokm, line 300:
```suggestion
Sentry differentiates stack frames that are directly related to your application ("in application") from stack frames that come from other packages such as the standard library, frameworks, or other dependencies. The application package is automatically marked as `inApp`. The difference is visible in [sentry.io](https://sentry.io), where only the "in application" frames are displayed by default.
```

lizokm (approved):
Thanks for the update!

PR #6471, opened 2023-03-15
user feedback remove note unsupported

bruno-garcia:
1. React Native actually supports it

During the platform selection flow we'll show in case something isn't supported. 99% people getting here will pick a supported platform so not worth having Perl and Native there

### DIFF
--- a/src/docs/product/user-feedback/index.mdx
+++ b/src/docs/product/user-feedback/index.mdx
@@ -24,10 +24,4 @@ User feedback is only applicable for [error issues](/product/issues/issue-detail
 
 Sentry provides two ways to collect user feedback, an API or an embeddable widget, which you can [configure](/platform-redirect/?next=/enriching-events/user-feedback/) to collect feedback when a user experiences an error.
 
-This feature is not currently supported for the following SDKs:
-
-- Native
-- Perl
-- React Native
-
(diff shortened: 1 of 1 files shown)


src/docs/product/user-feedback/index.mdx
```diff
@@ -24,10 +24,4 @@ User feedback is only applicable for [error issues](/product/issues/issue-detail
 
 Sentry provides two ways to collect user feedback, an API or an embeddable widget, which you can [configure](/platform-redirect/?next=/enriching-events/user-feedback/) to collect feedback when a user experiences an error.
```
lizokm, line 25:
```suggestion

There are two ways to collect user feedback in Sentry: via an API, or by using an embeddable widget, which you can [configure](/platform-redirect/?next=/enriching-events/user-feedback/) to collect feedback when a user experiences an error. For JVM-related use of Kotlin, check out our [Java](/platforms/java/enriching-events/user-feedback/) or [Android](/platforms/android/enriching-events/user-feedback/) SDKs.
```

lizokm: Since we're removing the unsupported list, we should also remove the first part of the paragraph below it:

"Clicking the "configure" link above for these SDKs may result in a "Page Not Found" error."

lizokm (approved):
Couple of small changes.

bruno-garcia:
Oh missed the suggestions. I'll add it in a separate pr

PR #6473, opened 2023-03-15
angular: remove single quote in double

bruno-garcia:
I imagine the `"'..'"` is an accident here?


bruno-garcia:
Seems like it's actually needed. Closing.

PR #6486, opened 2023-03-17
docs (dyn-sampling): Add Low Volume Projects section

andriisoldatenko:


matejminar (approved):

src/docs/product/data-management-settings/dynamic-sampling/index.mdx
```diff
@@ -37,6 +37,7 @@ When your data reaches high volumes, Sentry begins to automatically prioritize y
 | Latest Release       | Health Checks |
 | Dev Environments     |               |
 | Starred Transactions |               |
+| Low Volume Projects  |               |
```
ale-cota, line 40: I think we need an asterisk (*) here to mention that this one differently from the others can't be turned off. You can maybe add it below in the paragraph that explains the bias.

src/docs/product/data-management-settings/dynamic-sampling/index.mdx
```diff
@@ -60,6 +61,14 @@ Typically, for most projects there are only a few transactions that you really c
...
 _Health check_ type transactions, while important for checking the stability of your application, don't have any value for you beyond the task associated with them. They also tend to be high in volume, potentially using up a lot of you quota if retained. To avoid using up your volume on this type of transactions, we ignore these when sampling your data.
 
+
+### Low Volume Projects
+
+Sentry stores transactions from all projects at an equal rate, based on the fidelity target
+(ratio of stored vs. total) of a customer package. But some projects are very high volume compared to others
+and this means some low-volume projects might be "drowned out", especially at lower fidelity rates.
+We launched new rebalance algorithms which take the edge off of very larger projects and "redistribute" it to smaller ones.
+
```
ale-cota, line 71: Internally this description makes a lot of sense, but externally we're not communicating anywhere and explaining what things like "fidelity target" and "stored" vs. "total" transactions is... we don't want customers to think about that. So I would suggest to rephrase it slightly:

Some projects within your organisation are very high volume compared to others and this means some low-volume projects might be "drowned out" if we were to treat them all equally. To avoid this problem, we prioritize retaining transactions from lower volume projects, in order to ensure that you have enough samples at any project scale.

ale-cota (changes requested):

src/docs/product/data-management-settings/dynamic-sampling/index.mdx
```diff
@@ -1,3 +1,4 @@
+import Note from "../../../../components/note";
```
andriisoldatenko, line 1:
```suggestion
```

PR #6502, opened 2023-03-21
adjust qualifications for grace period

souredoutlook:
non-paid accounts have access to grace period. 

This PR changes the grace period section of the quota management docs to reflect this

### DIFF
--- a/src/docs/product/accounts/quotas/manage-event-stream-guide.mdx
+++ b/src/docs/product/accounts/quotas/manage-event-stream-guide.mdx
@@ -51,7 +51,7 @@ Once you know where your quota is being used, you can start taking the steps bel
 
 ## Grace Period
 
-We include a one-time grace period for error events on all paid accounts. This means that if an account depletes its error event volume and has never previously triggered a grace period, we continue to accept events for the next three days at a maximum rate of 10,000 events per hour. If you choose not to increase your error volume, Sentry will stop accepting events for the remainder of the current billing month. However, if you increase your volume, error events accepted during the grace period will be counted as consumed volume.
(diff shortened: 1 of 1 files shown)


mbauer404 (approved):

src/docs/product/accounts/quotas/manage-event-stream-guide.mdx
```diff
@@ -51,7 +51,7 @@ Once you know where your quota is being used, you can start taking the steps bel
 
 ## Grace Period
 
-We include a one-time grace period for error events on all paid accounts. This means that if an account depletes its error event volume and has never previously triggered a grace period, we continue to accept events for the next three days at a maximum rate of 10,000 events per hour. If you choose not to increase your error volume, Sentry will stop accepting events for the remainder of the current billing month. However, if you increase your volume, error events accepted during the grace period will be counted as consumed volume.
+All accounts have access to a one-time grace period. This means that if an account depletes its error event volume and has never previously triggered a grace period, we continue to accept events for the next three days at a maximum rate of 10,000 events per hour. If you choose not to increase your error volume, Sentry will stop accepting events for the remainder of the current billing month. However, if you increase your volume, error events accepted during the grace period will be counted as consumed volume.
```
lizokm, line 54:
```suggestion
All accounts get a one-time grace period. This means that if an account depletes its error event volume and has never previously triggered a grace period, we continue to accept events for the next three days at a maximum rate of 10,000 events per hour. If you choose not to increase your error volume, Sentry will stop accepting events for the remainder of the current billing month. However, if you increase your volume, error events accepted during the grace period will be counted as consumed volume.
```

PR #6514, opened 2023-03-23
unity: note on concat exceptions

bruno-garcia:
I dunno the APIs by heart so would appreciate review.

Should we point out `Debug.LogException` too? Any other API?


bruno-garcia:
@bitsandfoxes based on ur "approval" via slack, feel free to adjust this when u're online. Thanks

PR #6519, opened 2023-03-23
feat(apple): Time to Initial display and Time to Full display measurement

brustolin:
Added documentation for TTID and TTFD spans during view controller automatic transactions.

This docs should not be merged before [this PR](https://github.com/getsentry/sentry-cocoa/pull/2724).

### DIFF
--- a/src/platforms/android/common/performance/instrumentation/automatic-instrumentation.mdx
+++ b/src/platforms/android/common/performance/instrumentation/automatic-instrumentation.mdx
@@ -385,7 +385,7 @@ import io.sentry.Sentry
 Sentry.reportFullyDisplayed()
 ```
 
-Time to initial display is dependent on having an active transaction bound to the scope. Ideally, it should be used along with [Android's Activity Instrumentation](/platforms/android/performance/instrumentation/automatic-instrumentation/#androids-activity-instrumentation), which starts a transaction and binds it to the scope automatically.
+Time to full display is dependent on having an active transaction bound to the scope. Ideally, it should be used along with [Android's Activity Instrumentation](/platforms/android/performance/instrumentation/automatic-instrumentation/#androids-activity-instrumentation), which starts a transaction and binds it to the scope automatically.
 
 If the span finishes through the API, its `status` is set to `SpanStatus.OK`. If the span doesn't finish after 30 seconds, it is finished by the SDK automatically, and its `status` is set to `SpanStatus.DEADLINE_EXCEEDED`.
 If the span finishes before the Activity is first drawn and displayed as measured by the `Time to initial display`, the reported time will be shifted to `Time to initial display` measured time.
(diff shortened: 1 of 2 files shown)


src/platforms/android/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -385,7 +385,7 @@ import io.sentry.Sentry
 Sentry.reportFullyDisplayed()
 ```
 
-Time to initial display is dependent on having an active transaction bound to the scope. Ideally, it should be used along with [Android's Activity Instrumentation](/platforms/android/performance/instrumentation/automatic-instrumentation/#androids-activity-instrumentation), which starts a transaction and binds it to the scope automatically.
+Time to full display is dependent on having an active transaction bound to the scope. Ideally, it should be used along with [Android's Activity Instrumentation](/platforms/android/performance/instrumentation/automatic-instrumentation/#androids-activity-instrumentation), which starts a transaction and binds it to the scope automatically.
```
brustolin, line 388: I believe this was wrong. @stefanosiano can you confirm this?

stefanosiano: both of them depends on an active transaction, but yeah, this one was a copy/paste from the one above 😅

src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -306,3 +306,59 @@ SentrySDK.start { options in
...
+
+The span starts when the view of a view controller is loaded, and there is no other view controller transaction happening at the moment.
+
+The span finishes after the view appeared on the screen.
+
+### Time to Full Display
+
+_(New in version 8.4.0)_
+
+Time to full display provides insight into how long it would take your view controller to launch and load all of their content, by adding a span for each launch of an view controller. The SDK sets the span operation to `ui.load.full-display` and the span description to the view controllers's name followed by `full display`, for example, `MainActivity full display`.
```
philipphofmann, line 325:
```suggestion
Time to full display provides insight into how long it takes your view controller to launch and load all of its content, by adding a span for each launch of a view controller. The SDK sets the span operation to `ui.load.full-display` and the span description to the view controllers' name followed by `full display`, for example, `MainActivity full display`.
```

src/platforms/apple/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -306,3 +306,59 @@ SentrySDK.start { options in
...
+```objc {tabTitle:Objective-C}
+@import Sentry;
+
+[SentrySDK startWithConfigureOptions:^(SentryOptions *options) {
+    options.dsn = @"___PUBLIC_DSN___";
+    options.enableTimeToFullDisplay = YES;
+}];
+```
+
+_The span has to be finished manually, by using:_
```
philipphofmann, line 349: `m`: I think we should explain that we can't detect when your UIViewController is fully loaded. Only you, the user, can achieve this. Therefore, you have to manually call the API to get proper statistics.

philipphofmann (approved):
LGTM

PR #6535, opened 2023-03-27
Update quota management page

lizokm:
Make the quota management page easier to read and understand.


src/docs/product/accounts/quotas/index.mdx
```diff
@@ -127,6 +117,18 @@ This table outlines whether or not an an event or attachment that you send to Se
 
 <!-- prettier-ignore-end -->
 
+## Key Terms
+
+Let's clarify a few terms to start:
```
lizokm, line 122:
```suggestion
Let's clarify a few terms:
```

PR #6551, opened 2023-03-29
feat(sourcemaps): Update source maps docs

iambriccardo:
This PR updates the new docs for source maps by adding explanations of how Debug IDs work.

Closes: https://github.com/getsentry/sentry/issues/45633


src/platform-includes/sourcemaps/upload/cli/javascript.mdx
```diff
@@ -47,7 +39,108 @@ export SENTRY_ORG=___ORG_SLUG___
...
+
+### 2. Enable The Debug IDs In Sentry CLI
+
+Since Debug IDs are currently opt-in only, you should enable the following environment variable, which will tell the CLI to use the new Debug IDs system for uploading your source files.
+
+```bash
+export SENTRY_FORCE_ARTIFACT_BUNDLES=1
+```
+
+### 2. Inject Debug IDs Using Sentry CLI
```
matejminar, line 58: Looks like we have #2 two times.

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
```diff
@@ -47,7 +39,108 @@ export SENTRY_ORG=___ORG_SLUG___
...
+sentry-cli sourcemaps upload --release=<release_name> --dist=<dist_name> /path/to/file
+```
+
+<Note>
+
+The difference between the old release-based system and the new system with weak release association is that in the new one we don't require you to create a release but you can just optimistically tell Sentry the name of your "future" release.
+
+</Note>
+
+## Upload with Release
```
matejminar, line 125: Do we want to call out in the title that this is the beginning of another (old) approach?
I just looked at how the document flows, and I can see people just copy-pasting commands and not realizing that they are done and they don't need to continue with this step - it's just another, older approach.

matejminar (commented):
Do we need to update our [Best Practices](https://docs.sentry.io/platforms/javascript/sourcemaps/best-practices/) or [Troubleshooting](https://docs.sentry.io/platforms/javascript/sourcemaps/troubleshooting_js/) pages as well?

iambriccardo:
Good point @matejminar, the new system doesn't need some of those best practices but still, some of them might be useful. As of now, I cannot think of a best practice for our new system, since it aims at making everything effortless and working with minimal struggle.

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
iambriccardo: You are indeed right, I will try to think of a way to make it apparent.

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
```diff
@@ -1,20 +1,29 @@
...
+If you choose [`sentry-cli`](/product/cli/) as the method to upload your artifacts, we assume you already know how to generate source maps within your build toolchain.
 
-Sentry uses [releases](/product/releases/) to match the correct uploaded source maps to your events. You can use `sentry-cli` to create releases and upload source maps to Sentry.
+Sentry has two ways of matching the correct uploaded source maps to your events:
 
-## 1. Update SDK Options
+1. The first is to use Debug IDs, which allow Sentry to automatically associate minified sources with their respective source maps without needing to create a release. _This method is the preferred way of uploading artifacts._
 
-To be able to match Sentry errors with the correct release, you must provide a `release` property in the SDK options:
+2. The second is to use a [release](/product/releases/) to identify the artifacts you uploaded containing all your source files. _This method will require the creation of a release._
```
Jesse-Box, line 7: So, you've articulated how these methods differ in uploading source maps, but I think it's better to focus on the implications of using releases here.

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
```diff
@@ -24,30 +33,119 @@ For more info on Sentry CLI configuration visit the [Sentry CLI configuration do
...
-```ìni {filename:.sentryclirc}
-[auth]
-token=your-auth-token
+## Upload with Debug ID (Preferred)
 
-[defaults]
-org=___ORG_SLUG___
-project=___PROJECT_SLUG___
+Debug IDs are a new feature we have released to account for most of the problems that users were incurring in with the release-based system. Since having events not symbolicated is big problem for our users, we designed Debug IDs to provide a reliable system that would solve the problems people were incurring.
+
```
Jesse-Box, line 45: I would avoid explaining why we built this feature because I doubt people care. It might be better to use this time to explain what debug IDs are! 

For example; summarizing the process of injecting to uploading.

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
```diff
@@ -24,30 +33,119 @@ For more info on Sentry CLI configuration visit the [Sentry CLI configuration do
...
+Since Debug IDs are currently opt-in only, you should enable the following environment variable, which will tell Sentry CLI to use the new Debug IDs system for uploading your artifacts.
+
+```bash
+export SENTRY_FORCE_ARTIFACT_BUNDLES=1
 ```
 
+### 3. Inject Debug IDs Using Sentry CLI
+
+Debug IDs require an additional but very important step, the injection. In this step, [`sentry-cli`](/product/cli/) will go over all the minified sources and their respective source maps and it will start adding custom metadata that is required by Sentry for correctly symbolicating the event.
```
Jesse-Box, line 66: As a user, I assume every step is important unless specified otherwise, so let's remove the first sentence from this paragraph.

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
```diff
@@ -24,30 +33,119 @@ For more info on Sentry CLI configuration visit the [Sentry CLI configuration do
...
+<Note>
+
+Sentry CLI will notify you in case there are files that don't contain Debug IDs (e.g., you forgot to run the `inject` command). In that case, it will prompt you to weakly associate a release to the artifact. This weakly association is done in order to allow Sentry to properly symbolicate events without Debug IDs.
+
+</Note>
+
+#### Weak Release Association
+
+Since Debug IDs might cause issues for some users, we extended the new Debug IDs system with the concept of _Weak Release Association_. With such system you are able to tell Sentry to weakly associate the artifacts you upload with a `release`, and optionally `dist`.
+
```
Jesse-Box, line 109: What problems? Good to be specific or don't say it at all as it sounds like the feature is half-baked.

Jesse-Box (changes requested):
First round of feedback here

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
iambriccardo: Will fix it thanks!

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
iambriccardo: Great feedback, will do!

shanamatthews (approved):
I agreed with much of @Jesse-Box 's feedback and it looks like @iambriccardo already did a good job implementing much of it.

I did a pretty thorough review of this doc since it's so important (everyone struggles with source maps!) My suggested edits are intended to make the document clearer and more concise.

shanamatthews (approved):
I agreed with much of @Jesse-Box 's feedback and it looks like @iambriccardo already did a good job implementing much of it.

I did a pretty thorough review of this doc since it's so important (everyone struggles with source maps!) My suggested edits are intended to make the document clearer and more concise.

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
```diff
@@ -1,20 +1,31 @@
-If you choose [`sentry-cli`](/product/cli/) as the method to upload your source files, we assume you already know how to generate source maps within your build tool chain.
+If you choose [`sentry-cli`](/product/cli/) as the method to upload your artifacts, we assume you already know how to generate source maps within your build toolchain.
 
-Sentry uses [releases](/product/releases/) to match the correct uploaded source maps to your events. You can use `sentry-cli` to create releases and upload source maps to Sentry.
+Sentry has two ways of matching the correct uploaded source maps to your events:
```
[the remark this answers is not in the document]
lizokm: Added a colon and a comma :)

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
```diff
@@ -1,20 +1,31 @@
...
 
-## 1. Update SDK Options
+1. The first is to use Debug IDs, which allow Sentry to automatically associate minified sources with their respective source maps without needing to create a release. _This method is the preferred way of uploading artifacts._
```
lizokm, line 5:
```suggestion
1. Using Debug IDs allows Sentry to automatically associate minified sources with their respective source maps without needing to create a release. _This is the preferred method of uploading artifacts._
```
Since there are just two methods and we mention it above, saying the first and the second in addition to the numbers feels repetitive.

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
```diff
@@ -1,20 +1,31 @@
...
 
-To be able to match Sentry errors with the correct release, you must provide a `release` property in the SDK options:
+2. The second is to use a [release](/product/releases/) to identify the artifacts you uploaded containing all your source files. _This method will require the creation of a release._
```
lizokm, line 7:
```suggestion
2. Using a [release](/product/releases/) to identify the artifacts you uploaded containing all your source files. _This method will require the creation of a release._
```

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
```diff
@@ -1,20 +1,31 @@
...
 
-```javascript
-Sentry.init({
-  // Make sure this value is identical to the name you give the release that you
-  // create below using Sentry CLI
-  release: "<release_name>",
-});
+The implications of having to create a release are mainly related to the fact that sometimes you don't want to create a release for a change in your application. For example, if you want to release a development build to test internally, you would like to have the errors symbolicated but you may not want to create a release that is going to be shown in the UI.
```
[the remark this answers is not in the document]
lizokm: Added a few commas and change "which shows up" to "which would show up" since the rest of the sentence is in conditional tense.

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
```diff
@@ -24,30 +35,143 @@ For more info on Sentry CLI configuration visit the [Sentry CLI configuration do
...
+
+Since Debug IDs are currently opt-in only, you should enable the following environment variable, which will tell Sentry CLI to use the new Debug IDs system for uploading your artifacts.
+
+```bash
+export SENTRY_FORCE_ARTIFACT_BUNDLES=1
 ```
 
+### 3. Inject Debug IDs Using Sentry CLI
+
+During injection, [`sentry-cli`](/product/cli/) will go over all the minified sources and their respective source maps and it will start adding Debug IDs that are required by Sentry for correctly symbolicating the event.
```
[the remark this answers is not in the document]
lizokm: Added "the" (the Debug IDs)

lizokm (approved):
LGTM

iambriccardo:
Thanks a lot for the amazing review @shanamatthews and @lizokm! Now the doc flows way better, need to improve my verbose Italian-English 😅

iambriccardo:
To everyone following this PR, we decided to change the UX for enabling debug ids, with a UI toggle, thus until that is shipped, docs will remain stale.

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
```diff
@@ -1,53 +1,163 @@
...
+## Option 2: Upload with Release (Legacy)
+
+Sentry can also use [releases](/product/releases/) to match uploaded source maps to events. You can use `sentry-cli` to create releases and upload their source maps to Sentry.
 
 <Note>
 
-Visit the [auth token user settings page](https://sentry.io/settings/account/api/auth-tokens/) to find your auth tokens.
+We strongly suggest to use the new Debug IDs system for uploading your artifacts since it provides more reliable symbolication and the old release-based system will be faded out in the future.
 
 </Note>
```
lforst, line 144: Instead of only saying that we recommend the new version, we should just say in which cases the legacy upload is useful. eg:

- bundler plugins don't work for some reason
- bundler plugins make bundle size too large
- integrity issues
- ...

> the old release-based system will be faded out in the future

This is not correct I think. We cannot phase out the old system. There are cases where debug id will not work. We shouldn't unnecessarily scare users with deprecation notices.

lforst (commented):
To me the text feels way too technical. IMO we should not mention the term "debug id" even once. It is an implementation detail.

We can reduce the instructions for users to the following:

- if you want releases, set a release value
- if not, set the env var and run the inject command before uploading

No mention of debug ids necessary. I get that it is interesting from a technical standpoint but users shouldn't care.

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
```diff
@@ -1,53 +1,163 @@
...
   release: "<release_name>",
 });
 ```
 
-## 2. Upload Source Maps Using Sentry CLI
+Add these parameters to the `upload` command to weakly associate artifacts with a `release` and optionally `dist`:
 
-<Note>
+```bash
+sentry-cli sourcemaps upload --release=<release_name> --dist=<dist_name> /path/to/file
```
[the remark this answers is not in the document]
iambriccardo: Good point, how do we represent optional in docs?

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
iambriccardo: We are already using an updated fetching system that doesn't use debug ids, thus everything will use the new endpoint and logic, irrespective of debug ids. The plan is to fade out the old system and use the weakly associated system or debug id one based on the user information.

iambriccardo:
I don't agree with you @lforst. Users should know that their files are being modified due to an internal system we use, if they are unaware of the inner workings, it's gonna be super difficult for them to even start debugging the problem.

lforst:
> I don't agree with you @lforst. Users should know that their files are being modified due to an internal system we use, if they are unaware of the inner workings, it's gonna be super difficult for them to even start debugging the problem.

Hm, putting myself into the shoes of the user, even if I knew what's happening to my files, I wouldn't know how to debug this.

We're not building a source mapping library you need to understand the technical details of, we're building a product you want to spend minimal time on reading docs and setting it up.

iambriccardo:
> > I don't agree with you @lforst. Users should know that their files are being modified due to an internal system we use, if they are unaware of the inner workings, it's gonna be super difficult for them to even start debugging the problem.
> 
> Hm, putting myself into the shoes of the user, even if I knew what's happening to my files, I wouldn't know how to debug this.
> 
> We're not building a source mapping library you need to understand the technical details of. We're building a product you want to spend minimal time on reading docs and setting it up.

We didn't touch on technical details in depth, it just describes what the system does. It's not interrupting the reader in any impactful way. If a user doesn't even want to spend time reading two lines, the problem is somewhere else in my opinion. A simpler option would be to make some paragraphs collapsible.

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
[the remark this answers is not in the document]
shanamatthews: I don't think there's a way to represent this as optional in this code snippet. We say it's optional in the preceding text though, so I think it's fine.

src/platform-includes/sourcemaps/upload/cli/javascript.mdx
shanamatthews: This the current level of detail is fine. If you don't explain debug ids at all, the command `sentry-cli sourcemaps inject /path/to/folder` doesn't make sense. What are you injecting? Why?

It might sense to take another pass (can be in a later PR) at this and pull out all but the bare minimum explanation of debug ids into a separate doc.

Jesse-Box:
@iambriccardo here is my final suggestion for the docs.

My goal here was 
- To make each section easier to scan
- Consistently format verify steps
- Contextualize weakly associating releases.*

* You'll probably notice most of my changes focus on associating releases with artifact bundles. The reason why I was most aggressive with these edits was because I felt the original text doesn't contextualize when or why debug IDs would be missing. Practically speaking, weakly associating a release has the same process as setting up Release bundles, and so I want to emphasize when it's best, rather than its possible. We want to avoid giving the impression the new system is more complex than the legacy system.

---
### Inject Debug IDs into artifacts
During injection, `sentry-cli` will evaluate all minified sources and respective source maps (artifacts), adding the Debug IDs required to correctly de-minify an event.

To inject Debug IDs, use the new `inject` command:

```bash
sentry-cli sourcemaps inject /path/to/directory
```

<Note>
Please note that the `inject` command **injects code at the end of your bundle**, thus if an error occurs before that code, resolving with Debug IDs won't work. This more commonly occurs if an error is thrown during application initialization. (We are currently working to overcome this limitation)
</Note>

#### Verify Debug IDs were injected in artifacts
Minified source files should contain at the end a comment named `debugId` like:

```javascript {filename:exam[plminified-file.js}
...
//# debugId=<debug_id>
//# sourceMappingURL=<sourcemap_url>
```

Source maps should contain a field named `debug_id` like:

```json {filename:source-map.js.map}
{
    ...
    "debug_id":"<debug_id>",
    ...
}
```

### Upload artifact bundle
After you've injected your artifact files, upload them using the new `--use-artifact-bundle` option.

```bash
sentry-cli sourcemaps upload --use-artifact-bundle /path/to/directory
```

#### Verify artifact bundles were uploaded
Open up Sentry and navigate to **Project Settings > Source Maps**. If you choose “Artifact Bundles” in the tabbed navigation, you'll see all the artifact bundles that have been successfully uploaded to Sentry.

#### Uploading Artifact Bundles to a CDN
If you're using a CDN that builds based on which filenames were changed, we suggest that you associate a release to an artifact bundle. 

First, you must provide a release property in your SDK options.

```javascript
Sentry.init({  
// This value must be identical to the release name specified during upload  
// with the `sentry-cli`.  
release: "<release_name>",
});
```
After that, run the `sourcemaps upload` command with the additional `--release` option. Please ensure that the value specified for `<release_name>` is the same value specified in your SDK options.

```bash
sentry-cli sourcemaps upload --use-artifact-bundle --release=<release_name> /path/to/directory
```

#### Associating dist with Artifact Bundle (Optional)
Run the `sourcemaps upload` command with the additional  `--dist` option.

```bash
sentry-cli sourcemaps upload --use-artifact-bundle --dist=<dist_name> /path/to/directory
```

Jesse-Box (changes requested):

Jesse-Box (approved):

PR #6580, opened 2023-03-31
ci(dependencies): add dependency review

mdtro:
Add a dependency review workflow that will detect when new dependencies are added with existing vulnerabilities. If a high or critical exists, the action will report as a failure.


shanamatthews:
@mdtro - do we know why the vercel build failed for this? LMK if you're having issues.

mdtro:
@shanamatthews Sorry for the late reply. I have no idea why Vercel failed, it's not giving me much. I assume it's because I didn't actually make any changes to the content source?

I'll rebase and push again.

shanamatthews:
@mdtro - thanks! I think it's failing now due to some changes in upgrading Gatsby... Evan's working on it now. 

Since this PR is just for gh actions, is it cool if we just merge it so it's not hanging around?

mdtro:
@shanamatthews 

Whichever you prefer. I'd probably suggest merging this _after_ the Gatsby upgrade. Just in case there more dependency changes involved. I wouldn't want to make something more complicated than it already is. :)

Issue #6588, opened 2023-04-03
Improve SEO - Hide preview deployments


cleptric:
We noticed that old Vercel previews are indexed by, e.g. Google.

https://sentry-docs-git-patch-1.sentry.dev

Current preview deployments contain the correct `x-robots-tag: noindex` response header, but old preview deployments do not.

Action items

```[tasklist]
### Tasks
- [x] Get rid of all stale preview deploys like https://sentry-docs-git-patch-1.sentry.dev. We'll likely do this via the Vercel API, as we have thousands of them. https://github.com/getsentry/sentry-docs/pull/6592
- [ ] Purge Google cache via https://search.google.com/search-console/, to get rid of old indexed pages. OPs ticket https://getsentry.atlassian.net/browse/OPS-3096
- [x] Add a GitHub action that deletes all preview deploys for branches that were merged/deleted.
```

shanamatthews:
Quick update - Google was still indexing (some of?) the deleted preview deploys, Michi went into our Google Search console and started a reindex of the 404s, which should remove them from search results once the reindex finishes (should be a couple days).

PR #6592, opened 2023-04-04
feat(docs): Add preview delete script

AbhiPrasad:
ref: https://github.com/getsentry/sentry-docs/issues/6588

To try to respect Vercel rate limits, this script is pretty slow. I will leave it overnight to completely delete all the previews.

Pushing this up so @shanamatthews if you need to run again in the future you can!

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "remark-deflist": "^0.2.1",
     "sass": "^1.56.2",
     "search-insights": "^2.2.3",
-    "ts-node": "^9.0.0",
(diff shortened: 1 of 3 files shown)


cleptric (approved):
Is the bump of `ts-node` necessary, has any side effects?

AbhiPrasad:
> Is the bump of ts-node necessary, has any side effects?

Should be completely safe. This biggest change in the new version is dropping support for Node 10, and we are on Node 16 locally in the repo so we should be fine.

PR #6607, opened 2023-04-06
Highlight limitations of transaction filters

ale-cota:
It's pretty easy to miss the limitations on filtering for transactions, trying to make it a bit easier to spot.


matejminar (approved):

gggritso (approved):
Makes sense, thanks!

PR #6615, opened 2023-04-07
feat: Improve & streamline Loader & CDN instructions

mydea:
A few code snippets were outdated and not really needed anymore.

In other places, I added instructions to install stuff via the Loader.

I also streamlined tab titles to be more consistent.


src/platform-includes/configuration/capture-console/javascript.mdx
```diff
@@ -14,27 +14,47 @@ Sentry.init({
 });
 ```
 
-```javascript {tabTitle: CDN}
+```html {tabTitle: Loader}
+<!-- In addition to the Loader: -->
```
AbhiPrasad, line 18: Can we not include the loader URL in here?

src/platform-includes/configuration/enable-pluggable-integrations/node.mdx
shanamatthews, line 1: Something is up with this file in docs. 

/platforms/javascript/common/configuration/integrations/plugin.mdx doesn't generate the page for node in docs, instead it's generated from /platforms/node/common/configuration/integrations/pluggable-integrations.mdx

So none of the changes to this file show up: https://sentry-docs-pvmsapnvh.sentry.dev/platforms/node/configuration/integrations/default-integrations/

src/platform-includes/configuration/capture-console/javascript.mdx
mydea: I wasn't 100% sure if we want to include this in every one of this 🤔 As it kind of drowns out the "relevant" change a bit, IMHO. But not sure, happy to put it in if we think that makes more sense!

src/platform-includes/configuration/enable-pluggable-integrations/node.mdx
mydea: This is correct, because this is only for the browser, not for node! It shows up correctly here: https://sentry-docs-git-fn-remove-cdn-docs.sentry.dev/platforms/javascript/configuration/integrations/plugin/

src/platform-includes/configuration/capture-console/javascript.mdx
AbhiPrasad: I would like to see it to make it extra clear what is going on - also signifies order to users.

src/platform-includes/configuration/enable-pluggable-integrations/node.mdx
shanamatthews: Sorry, still confused. There's
- /enable-pluggable-integrations/javascript.mdx
- /enable-pluggable-integrations/node.mdx (this file)

The first one builds to https://sentry-docs-git-fn-remove-cdn-docs.sentry.dev/platforms/javascript/configuration/integrations/plugin/ 

Where does the second one build to?

mydea: Ah, right, sorry, I misunderstood you. I see, maybe this file is actually not used - I think this can be removed! I'll try it locally & remove it.

shanamatthews: Yes, I think so too! If you remove it, would you mind making sure that the content [here](https://sentry-docs-pvmsapnvh.sentry.dev/platforms/node/configuration/integrations/pluggable-integrations/) is correct? I think folks have been updating /enable-pluggable-integrations/node.mdx and thinking it's being published to the Node pluggable integrations page, but it's not.

mydea:
> 

This is already in the docs here: 
https://docs.sentry.io/platforms/javascript/install/loader/#sdk-configuration

AbhiPrasad (approved):

mydea:
So I updated this:

* Removed the unused node page
* Added back the base Loader & CDN instructions for each integration

src/platform-includes/configuration/enable-pluggable-integrations/node.mdx
mydea: it looks good 👍

PR #6635, opened 2023-04-11
feat(crons): Environment support in PHP

evanpurkhiser:
Need to confirm this is correct


cleptric:
We already default to production.
https://github.com/getsentry/sentry-php/blob/master/src/CheckIn.php#L57

Laravel uses the correct env set in the SDK.
https://github.com/getsentry/sentry-laravel/blob/master/src/Sentry/Laravel/Features/ConsoleIntegration.php#L121

shanamatthews (approved):

evanpurkhiser:
This needs changed I think

PR #6636, opened 2023-04-11
feat(crons): Environment support in python

evanpurkhiser:
Need to confirm this is correct


antonpirker:
This is not how the `environment` is set. The `monitor` decorator takes the `environment` from the environment that is set in the init call: `sentry_sdk.init(..., environment="staging")`.

antonpirker (changes requested):
Environment is taken from `sentry_sdk.init()` call, there is no `environment` parameter in the `monitor` decorator.

PR #6641, opened 2023-04-12
feat(node): Add docs for async context apis

AbhiPrasad:
ref https://github.com/getsentry/sentry-javascript/issues/7691

ref https://github.com/getsentry/sentry-javascript/pull/7817

In the new upcoming version of the Node SDK, we now expose an API allowing users to isolate their requests to prevent breadcrumbs/tags/spans from leaking across requests. This PR documents that.

Before users had to manually use Node's domains API to do this, but now we've abstracted this away from them. To make this easier, we've updated the Koa docs to stop using domains directly, but instead use the new `runWithAsyncContext` API.

Note: only merge when `7.48.0` is released.

### DIFF
--- a/src/platforms/node/common/configuration/async-context.mdx
+++ b/src/platforms/node/common/configuration/async-context.mdx
@@ -11,14 +11,9 @@ const Sentry = require("@sentry/node");
 
 function requestHandlerMiddleware(req, res, next) {
   // Any breadcrumbs or tags added will be isolated to the request
-  return Sentry.runWithAsyncContext(
-    () => {
-      return next(req, res);
-    },
-    // emitters are event emitters that need to be tracked by async
-    // context since they emit async events
-    { emitters: [req, res] }
-  );
+  return Sentry.runWithAsyncContext(() => {
+    return next(req, res);
+  });
 }
 ```
 
@@ -31,11 +26,6 @@ const domain = require("domain");
 
 function myRequestHandler(req, res, next) {
   const localDomain = domain.create();
-
-  // track event emitters in domain
-  localDomain.add(req);
-  localDomain.add(res);
-
   return localDomain.bind(() => {
     return next(req, res);
   })();
--- a/src/platforms/node/common/configuration/async-context.mdx
+++ b/src/platforms/node/common/configuration/async-context.mdx
@@ -1,0 +1,40 @@
+---
+title: Async Context
+sidebar_order: 80
+description: "Learn more about how to isolate Sentry scope and breadcrumbs across requests."
+---
+
+You can use the `runWithAsyncContext` method to isolate Sentry scope and breadcrumbs to a single request if you are using SDK v7.48.0 or higher. This is useful if you are finding that breadcrumbs and scope are leaking across requests.
+
+```js
+const Sentry = require("@sentry/node");
+
+function requestHandlerMiddleware(req, res, next) {
+  // Any breadcrumbs or tags added will be isolated to the request
+  return Sentry.runWithAsyncContext(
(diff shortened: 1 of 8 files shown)


src/platforms/node/common/configuration/async-context.mdx
```diff
@@ -0,0 +1,40 @@
...
+  return Sentry.runWithAsyncContext(
+    () => {
+      return next(req, res);
+    },
+    { emitters: [req, res] }
+  );
+}
+```
+
+Under the hood the SDK uses Node's [AsyncLocalStorage API](https://nodejs.org/api/async_context.html#class-asynclocalstorage) to perform the isolation.
```
shanamatthews, line 23:
```suggestion
Under the hood the SDK uses Node's [AsyncLocalStorage API](https://nodejs.org/api/async_context.html#class-asynclocalstorage) to perform the isolation.
```
```suggestion
Under the hood, the SDK uses Node's [AsyncLocalStorage API](https://nodejs.org/api/async_context.html#class-asynclocalstorage) to perform the isolation.
```

shanamatthews (approved):
Looks great! I'm curious why you chose Koa as the example to show this on.

src/platforms/node/common/configuration/async-context.mdx
```diff
@@ -0,0 +1,40 @@
...
+```js
+const Sentry = require("@sentry/node");
+
+function requestHandlerMiddleware(req, res, next) {
+  // Any breadcrumbs or tags added will be isolated to the request
+  return Sentry.runWithAsyncContext(
+    () => {
+      return next(req, res);
+    },
+    { emitters: [req, res] }
```
lforst, line 18: Should we explain what users should pass as `emitters` and what they are used for?

lforst (approved):

src/platforms/node/common/configuration/async-context.mdx
AbhiPrasad: I can add a note.

src/platforms/node/common/configuration/async-context.mdx
```diff
@@ -0,0 +1,40 @@
...
+On lower SDK versions you'll have to use [domains](https://nodejs.org/api/domain.html) to isolate Sentry scope and breadcrumbs to a single request.
+
+```js
+const domain = require("domain");
+
+function myRequestHandler(req, res, next) {
+  const localDomain = domain.create();
+
+  localDomain.add(req);
+  localDomain.add(res);
```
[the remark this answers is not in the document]
AbhiPrasad: We are doing this in our production code atm - it's why the `emitters` field exists.

src/platforms/node/guides/koa/index.mdx
```diff
@@ -64,30 +63,25 @@ Sentry.init({
...
-    local.run(async () => {
-      Sentry.getCurrentHub().configureScope(scope =>
-        scope.addEventProcessor(event =>
-          Sentry.addRequestDataToEvent(event, ctx.request, {
-            include: {
-              user: false,
-            },
-          });
+    Sentry.runWithAsyncContext(
+      async hub => {
```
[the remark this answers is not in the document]
AbhiPrasad: We discussed offline and decided this was fine, since very few people are going to use this API anyway.

src/platforms/node/common/configuration/async-context.mdx
AbhiPrasad: Actually I decided to remove this, since 99.9% of users won't need it. We can add additional notes if users ask for it.

PR #6642, opened 2023-04-12
feat: Add docs for `ignoreErrors` & `ignoreTransactions`

mydea:
This adds proper docs for `ignoreErrors` & `ignoreTransactions` as ways to filter events.

I've only added them for JavaScript for now, but I believe it also exists for other languages already - we may also add these then?

### DIFF
--- a/src/platforms/common/configuration/options.mdx
+++ b/src/platforms/common/configuration/options.mdx
@@ -268,13 +268,13 @@ For ASP.NET and ASP.NET Core applications, the value will default to the server'
 
 <ConfigKey name="ignore-errors" supported={["javascript", "node"]} notSupported={[]}>
 
-A list of strings or regex patterns that match error messages that should not be sent to Sentry. This expects a list of the messages to be filtered out before being sent to Sentry as either regular expressions or strings. When using strings, they’ll partially match the messages, so if you need to achieve an exact match, use RegExp patterns instead. By default, all errors are sent.
+A list of strings or regex patterns that match error messages that shouldn't be sent to Sentry. Messages that match these strings or regular expressions will be filtered out before they're sent to Sentry. When using strings, partial matches will be filtered out, so if you need to filter by exact match, use regex patterns instead. By default, all errors are sent.
 
 </ConfigKey>
 
 <ConfigKey name="ignore-transactions" supported={["javascript", "node"]} notSupported={[]}>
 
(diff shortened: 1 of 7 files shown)


cleptric (approved):

src/platforms/common/configuration/options.mdx
```diff
@@ -266,6 +266,18 @@ For ASP.NET and ASP.NET Core applications, the value will default to the server'
 
 </ConfigKey>
 
+<ConfigKey name="ignore-errors" supported={["javascript", "node"]} notSupported={[]}>
+
+A list of strings or regex patterns that match error messages that should not be sent to Sentry. This expects a list of the messages to be filtered out before being sent to Sentry as either regular expressions or strings. When using strings, they’ll partially match the messages, so if you need to achieve an exact match, use RegExp patterns instead. By default, all errors are sent.
```
shanamatthews, line 271:
```suggestion
A list of strings or regex patterns that match error messages that shouldn't be sent to Sentry. Messages that match these strings or regular expressions will be filtered out before they're sent to Sentry. When using strings, partial matches will be filtered out, so if you need to filter by exact match, use regex patterns instead. By default, all errors are sent.
```

src/platforms/common/configuration/options.mdx
```diff
@@ -266,6 +266,18 @@ For ASP.NET and ASP.NET Core applications, the value will default to the server'
...
+
+</ConfigKey>
+
+<ConfigKey name="ignore-transactions" supported={["javascript", "node"]} notSupported={[]}>
+
+A list of strings or regex patterns that match transaction names that should not be sent to Sentry. This expects a list of the names to be filtered out before being sent to Sentry as either regular expressions or strings. When using strings, they’ll partially match the name, so if you need to achieve an exact match, use RegExp patterns instead. By default, all transactions are sent.
```
shanamatthews, line 277:
```suggestion
A list of strings or regex patterns that match transaction names that shouldn't be sent to Sentry. Transactions that match these strings or regular expressions will be filtered out before they're sent to Sentry. When using strings, partial matches will be filtered out, so if you need to filter by exact match, use regex patterns instead. By default, all transactions are sent.
```

shanamatthews (approved):

shanamatthews:
Looked great, just a couple of minor phrasing tweaks for clarity.

PR #6655, opened 2023-04-14
Remove alert

ale-cota:
After verifying with the DnD and Ingest teams, we confirmed that custom performance metrics are NOT gated by AM2 and are also available on older plans, so I am removing the alert.


priscilawebdev (approved):

shanamatthews (approved):

lizokm (commented):
LGTM. I'll go ahead and push this change out.

PR #6656, opened 2023-04-14
docs(profiling): add flamechart and flamegraph docs

JonasBa:


shanamatthews:
included in #6767 as discussed!

PR #6681, opened 2023-04-19
style(js): Apply existing prettier config to everything

evanpurkhiser:


src/docs/clients/cocoa/dsym.mdx
```diff
@@ -125,8 +125,9 @@ fi
 
 3.  You also need to add the following line to the _Input Files_ section in the _Run Script_ from step 2:
 
-[//]: # (Don't use bash here. Clicking the copy button on the code sample with bash)
-[//]: # (removes the leading $.)
+[//]: # "Don't use bash here. Clicking the copy button on the code sample with bash"
+[//]: # "removes the leading $."
```
evanpurkhiser, line 129: urgh

src/docs/clients/cocoa/dsym.mdx
```diff
@@ -125,8 +125,9 @@ fi
...
+
```
shanamatthews, line 130: this is still here and very weird. not sure which rule is causing it?

src/docs/clients/python/api.mdx
```diff
@@ -23,21 +23,21 @@ This gives you an overview of the public API that raven-python exposes.
...
+        'url': '...',
+        'data': {},
+        'query_string': '...',
+        'method': 'POST',
+    },
+    'logger': 'logger.name',
+}, extra={
+    'key': 'value',
+})
+```
```
shanamatthews, line 40: we also want to keep these leading spaces since they cause the code block to be indented in the text

I think a similar thing is happening with bullets but haven't run across the example in this PR

evanpurkhiser: Let's confirm in the build preview (assuming it builds)

evanpurkhiser: nvm it also failed

PR #6682, opened 2023-04-19
style(js): Apply prettier to js/ts files

evanpurkhiser:


shanamatthews (approved):
this all looks reasonable to me

PR #6686, opened 2023-04-19
fix: Rename javascript ember.mdx -> javascript.ember.mdx

evanpurkhiser:
Spaces no good. afaict this might not be used?


AbhiPrasad (approved):
Right now it doesn't display here: https://docs.sentry.io/platforms/javascript/guides/ember/#verify, so this change is needed.

PR #6691, opened 2023-04-20
feat(search): Make it better

evanpurkhiser:
- Search now supports keyboard navigation

   This is lifted from [my own project][0]
 - Search now uses gatsby's smart <Link /> and navigate function, so search will now correctly work in dev
 - Auto focus on the index page
 - General code cleanup

[0]: https://github.com/evanpurkhiser/waitress/blob/main/components/useKeyboardNavigate.tsx

### DIFF
--- a/src/components/search.tsx
+++ b/src/components/search.tsx
@@ -1,4 +1,6 @@
 import React, { useState, useEffect, useRef, useCallback } from "react";
+import { Link, navigate } from "gatsby";
+
 import algoliaInsights from "search-insights";
 
 import Logo from "./logo";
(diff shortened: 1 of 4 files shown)


src/components/search.tsx
```diff
@@ -72,61 +75,90 @@ const useClickOutside = (
   });
 };
 
+function relativizeUrl(url: string) {
+  console.log(url);
```
scttcper, line 79: cleanup?

evanpurkhiser: 👍

evanpurkhiser, line 79:
```suggestion
```

scttcper (approved):

PR #6692, opened 2023-04-20
ref(ui): Improve bannar close position

evanpurkhiser:
Align center


evanpurkhiser:
nvm this is used in some other places

PR #6695, opened 2023-04-20
fix(rn): Use npx for all rn wizard examples

krystofwoldrich:
Previously we used `npx` and `yarn` interchangeably but since the new one-line command for RN we should use only `npx` to avoid `wizard` not installed errors.


kahest (approved):
LGTM 🚀

PR #6696, opened 2023-04-20
feat(relay): Add spooling configuration options

olksdr:
Adding new spooling configuration options:
* path to the file
* max memory size
* max disk size
* max / min connections

### DIFF
--- a/src/docs/product/relay/options.mdx
+++ b/src/docs/product/relay/options.mdx
@@ -199,6 +199,42 @@ See also `cache.envelope_expiry`.
 
 Interval for evicting outdated project configs from memory.
 
+## Spooling
+
+These settings control on-disk spooling. Currently only spooling for incoming envelopes is supported.
+
+`spool.envelopes.path`
+
+: _String, default: null_
+
+  The path to the spool DB file, which is used to persist the envelopes to disk.
+  If provided enables on disk spooling.
+
+`spool.envelopes.max_memory_size`
+
+: _String, default: `500MB`_
+
+  The maximum in-memory buffer for incoming envelopes.
+  The buffer will be spooled to disk when the configured size is reached.
+
+`spool.envelopes.max_disk_size`
+
+: _String, default: `500MB`_
+
+  The maximum size on disk spool.
+
+`spool.envelopes.max_connections`
+
+: _Integer, default: `20`_
+
+  The maximum number of connections in the pool to manage the on-disk spool.
+
+`spool.envelopes.min_connections`
+
+: _Integer, default: `10`_
+
+  The minimum number of connections in the pool to manage the on-disk spool.
+
 ## Size Limits
 
 These settings control HTTP-related limits. All values are either integers, or are
--- a/src/docs/product/relay/options.mdx
+++ b/src/docs/product/relay/options.mdx
@@ -201,14 +201,14 @@ Interval for evicting outdated project configs from memory.
 
 ## Spooling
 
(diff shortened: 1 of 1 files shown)


src/docs/product/relay/options.mdx
```diff
@@ -199,6 +199,42 @@ See also `cache.envelope_expiry`.
...
 
+## Spooling
+
+These settings control on-disk spooling. Currently only spooling for incoming envelopes is supported.
+
+`spool.envelopes.path`
+
+: _String, default: null_
+
+  The path to the spool DB file, which is used to persist the envelopes to disk.
```
iker-barriocanal, line 210: nit: I think DB is too much details for this.
```suggestion
  The path to the spool file, which is used to persist the envelopes to disk.
```

src/docs/product/relay/options.mdx
```diff
@@ -199,6 +199,42 @@ See also `cache.envelope_expiry`.
...
+  If provided enables on disk spooling.
+
+`spool.envelopes.max_memory_size`
+
+: _String, default: `500MB`_
```
iker-barriocanal, line 215: Units are separated from the value (see examples above), so either
```suggestion
: _String, default: `500` (MB)_
```
or
```suggestion
: _String, default: `500` (500MB)_
```
Same with the unit below.

iker-barriocanal (approved):
Please fix the unit before merging.

src/docs/product/relay/options.mdx
olksdr: Provided values are correct, see for example how the `limits` below configured. 
This is the format (human readable) which is used to configure this options.

src/docs/product/relay/options.mdx
lizokm, line 210:
```suggestion
These settings control on-disk spooling. Currently, only spooling for incoming envelopes is supported.

`spool.envelopes.path`

: _String, default: null_

  The path to the spool DB file, which is used to persist the envelopes to disk.
```
Not familiar with this concept, but wan to double check that "persist" is the correct word to be used here?

src/docs/product/relay/options.mdx
```diff
@@ -199,6 +199,42 @@ See also `cache.envelope_expiry`.
...
+## Spooling
+
+These settings control on-disk spooling. Currently only spooling for incoming envelopes is supported.
+
+`spool.envelopes.path`
+
+: _String, default: null_
+
+  The path to the spool DB file, which is used to persist the envelopes to disk.
+  If provided enables on disk spooling.
```
lizokm, line 211:
```suggestion
  If provided, enables on-disk spooling.
```

src/docs/product/relay/options.mdx
```diff
@@ -199,6 +199,42 @@ See also `cache.envelope_expiry`.
...
+: _String, default: `500MB`_
+
+  The maximum in-memory buffer for incoming envelopes.
+  The buffer will be spooled to disk when the configured size is reached.
+
+`spool.envelopes.max_disk_size`
+
+: _String, default: `500MB`_
+
+  The maximum size on disk spool.
```
lizokm, line 224:
```suggestion
  The maximum size on-disk spool.
```

lizokm (approved):
A few tiny tweaks :)

src/docs/product/relay/options.mdx
olksdr: Thanks! Changed it to `save` , which precisely describes what's going on.

PR #6697, opened 2023-04-20
Clean up Next.js docs

lforst:
I gave the Next.js docs a pass and a lot of it doesn't make sense anymore or is too verbose.


src/platforms/javascript/guides/nextjs/manual-setup.mdx
```diff
@@ -6,40 +6,57 @@ description: "Learn how to set up the SDK manually."
...
-```typescript {filename:sentry.server.config.ts/sentry.client.config.ts/sentry.edge.config.ts}
+```typescript {filename:sentry.(client|server|edge).ts}
 import * as Sentry from "@sentry/nextjs";
 
-const SENTRY_DSN = process.env.SENTRY_DSN || process.env.NEXT_PUBLIC_SENTRY_DSN;
-
 Sentry.init({
-  dsn: SENTRY_DSN || "___PUBLIC_DSN___",
+  dsn: "___PUBLIC_DSN___",
+
```
AbhiPrasad, line 53:
```suggestion
```

so that the spacing is the same between js/ts

AbhiPrasad (approved):
Should we move `Configure Tunneling to avoid Ad-Blockers` higher up the page so it's more likely to be seen?

PR #6705, opened 2023-04-20
style(prettier): Consistent italics and bold styling

evanpurkhiser:


JoshFerge (approved):
is there a prettier rule for this? how will it be enforced?

evanpurkhiser:
Yeah this is from prettier. I’m slowly fixing everything in batches with prettier so I can enforce it after it doesn’t change everything

PR #6720, opened 2023-04-24
ref(onboarding): Remove backbone platform and not used doc

priscilawebdev:
As backbone is just an integration and we don't have a Sentry guide for it, we decided to remove it from our onboarding

also removing the file `src/wizard/javascript/nextjs_wizard.md` as it is not used


AbhiPrasad (approved):
end of an era

PR #6741, opened 2023-04-27
fix(ui): Avoid scroll chaining on code block selector

evanpurkhiser:
When picking a DSN etc, do not let the scroll chain outside of the
selector box


shanamatthews (approved):
I like it!

PR #6749, opened 2023-04-27
fix: Silence 'Error executing variable-like construct'

evanpurkhiser:
This warning is produced far to much for it to be of any value

### DIFF
--- a/src/gatsby/plugins/gatsby-remark-variables/index.js
+++ b/src/gatsby/plugins/gatsby-remark-variables/index.js
@@ -61,7 +61,9 @@ module.exports = ({markdownAST, markdownNode}, options) => {
             result = await result;
           }
         } catch (err) {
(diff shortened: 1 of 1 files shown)


src/gatsby/plugins/gatsby-remark-variables/index.js
```diff
@@ -61,7 +61,9 @@ module.exports = ({markdownAST, markdownNode}, options) => {
             result = await result;
           }
         } catch (err) {
-          console.warn(`Error executing variable-like construct: ${match[0]}`);
+          // no-op. We previously had a warning here, but the we have so many
+          // "variable-like" cosntructs in our codeblocks that this becomes
```
gggritso, line 65:
```suggestion
          // "variable-like" constructs in our codeblocks that this becomes
```

gggritso (approved):

PR #6764, opened 2023-04-27
chore(deps): Bump postcss

evanpurkhiser:


evanpurkhiser:
Purely a node version bump https://github.com/postcss/postcss/wiki/PostCSS-8-for-end-users

PR #6801, opened 2023-05-02
chore(prettier): Apply prettier to all code blocks

evanpurkhiser:
If we want to apply prettier to our markdown files automatically we'll
need to do this.


src/docs/product/cli/configuration.mdx
```diff
@@ -36,9 +36,11 @@ Visit your [auth token user settings page](https://sentry.io/settings/account/ap
...
   ```
 - export it as an environment variable:
-  ```bash
-  export SENTRY_AUTH_TOKEN=your-auth-token
-  ```
+```bash
+export SENTRY_AUTH_TOKEN=your-auth-token
+``
+
+  `
```
shanamatthews, line 43: this doesn't look right

src/platform-includes/enriching-events/user-feedback/example-widget/php.laravel.mdx
```diff
@@ -15,13 +15,13 @@ Next, create `resources/views/errors/500.blade.php`, and embed the feedback code
...
-        eventId: '{{ app('sentry')->getLastEventId() }}'
-      });
-    </script>
+  <div class="subtitle">Error ID: {{ app('sentry')->getLastEventId() }}</div>
+  <script>
+    Sentry.init({ dsn: '___PUBLIC_DSN___' });
+    Sentry.showReportDialog({
+      eventId: '{{ app('sentry')->getLastEventId() }}'
+    });
+  </script>
```
shanamatthews, line 24: seems like this isn't ideal but not a big deal

src/platform-includes/session-replay/setup/javascript.capacitor.mdx
```diff
@@ -1,15 +1,13 @@
 Several options are supported and passable using the integration constructor. See the [configuration documentation](/platforms/javascript/guides/capacitor/session-replay/configuration/) for more details. To set up the integration, add the following to your Sentry initialization:
 
-<Note>
-The minimal supported version of Sentry's SDK for Capacitor is 0.11.1.
-</Note>
+<Note>The minimal supported version of Sentry's SDK for Capacitor is 0.11.1.</Note>
```
shanamatthews, line 3: Does this work okay like this?

Our internal docs on docs say notes have to be formatted like this to work with Gatsby:

```suggestion
<Note>

The minimal supported version of Sentry's SDK for Capacitor is 0.11.1.

</Note>
```

evanpurkhiser: In this scenario either way is fine. since there's no markdown in the note.

If there was then it would not work since the markdown would not be parsed.

src/docs/product/cli/configuration.mdx
evanpurkhiser: Wacky. I'll take a look at this

evanpurkhiser: It's becasue of the trailing `:`

evanpurkhiser: Actually I'm not 100% sure lol

src/platforms/javascript/guides/electron/index.mdx
```diff
@@ -88,24 +93,24 @@ The default transport automatically handles offline events caching. Following, a
...
-    queuedLengthChanged: (length) => { },
+    queuedLengthChanged: length => {},
     // Called before attempting to send an event to Sentry. Used to override queuing behavior.
     //
     // Return 'send' to attempt to send the event.
     // Return 'queue' to queue and persist the event for sending later.
     // Return 'drop' to drop the event.
-    beforeSend: (request) => isOnline() ? 'send' : 'queue'
-  }
+    beforeSend: request => (isOnline() ? 'send' : 'queue'),
```
shanamatthews, line 112: I don't think this is wrong, but this formatting doesn't feel conventional imo

evanpurkhiser: Yeah it's definitely not wrong, it's prettier conventional. It's wrapped like that for clarity, since otherwise it can be slightly unclear that it doesn't mean

```tsx
(request => isOnline()) ? 'send' : 'queue'`
```

evanpurkhiser: I agree though, unfortunately we probably don't want to turn off prettier just for stuff like this 😢

shanamatthews (approved):
I think the weird bash code fence was the only actual problem here!

PR #6810, opened 2023-05-03
feat(lint): Automatic fix + commit in CI

evanpurkhiser:
This will auto-commit any automatic lint fixes from prettier / eslint


shanamatthews (approved):
psyched for this!

evanpurkhiser:
There is still a tad more we have to do to get this working, but just relying on dev-infra now I think :-)

PR #6834, opened 2023-05-04
Revert "chore(prettier): Apply prettier to all code blocks"

HazAT:
Reverts getsentry/sentry-docs#6801


lforst (approved):

kamilogorek:
Why is that? Please leave some context what went wrong when we do reverts :3

lforst:
> Why is that? Please leave some context what went wrong when we do reverts :3

It introduced "Sentry code style" to all the code snippets didn't bode well with the SDK teams. Right now the code snippets are curated with a particular style - we can argue about whether we want prettier to run over these.

IMO we should have prettier, BUT it should be with prettier default rules so we don't alienate users with our style.

PR #6839, opened 2023-05-04
Uploading sourcemaps without debug ids

matejminar:
Resolves https://github.com/getsentry/sentry/issues/47237

### DIFF
--- a/src/platform-includes/sourcemaps/uploading-without-debug-ids/javascript.mdx
+++ b/src/platform-includes/sourcemaps/uploading-without-debug-ids/javascript.mdx
@@ -1,0 +1,73 @@
+We've recently updated all our guides to help users upload source maps using Artifact Bundles with Debug IDs. We think that users who have failed to upload source maps in the past will get a much better experience if they follow our current guides.
+
+However, this does require your application to use a very recent SDK and bundler plugin (or sentry-cli) version. We know that this isn't always possible, so here we'll explain what to expect when uploading source maps without Debug IDs.
+
+## Artifact Bundle & Debug ID Requirements
+
+You need a valid version of our SDK and the bundler plugin (or sentry-cli) to use Artifact Bundles with Debug IDs.
+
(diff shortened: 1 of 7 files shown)


src/platform-includes/sourcemaps/uploading-without-debug-ids/javascript.mdx
```diff
@@ -0,0 +1,73 @@
...
+  release: '<release_name>',
+});
+```
+
+### 2. Create a New Release
+You must first create a release with `sentry-cli`. The release name must be unique within your organization:
+
+```bash
+sentry-cli releases new <release_name>
+```
```
iambriccardo, line 50: With the new system you don't need to manually create a `release`.

iambriccardo: With the new system you don't need to manually create a release.

iambriccardo (commented):
Is this doc showing the old system or the new system with weak release association, I don't understand. Currently everyone should use the new artifact bundles system with weak release association, unless they have old cli versions.

src/platform-includes/sourcemaps/uploading-without-debug-ids/javascript.mdx
matejminar: @iambriccardo 
> Since we've updated our sentry-cli guides to focus entirely on debug IDs, users with older SDKs need an alternative guide to support their specific needs.
https://github.com/getsentry/sentry/issues/47237

This is a guide for folks who don't want / are unable to update their CLIs therefore they can't use the new system.

matejminar:
@iambriccardo 
> Since we've updated our sentry-cli guides to focus entirely on debug IDs, users with older SDKs need an alternative guide to support their specific needs.
https://github.com/getsentry/sentry/issues/47237

This is a guide for folks who don't want / are unable to update their CLIs therefore they can't use the new system.

PR #6854, opened 2023-05-04
chore(prettier): A little more minor formatting cleanup

evanpurkhiser:
These were missed in earlier PRs and snuck into the code only formatting
PR

### DIFF
--- a/src/platform-includes/session-replay/setup/javascript.capacitor.mdx
+++ b/src/platform-includes/session-replay/setup/javascript.capacitor.mdx
@@ -1,6 +1,8 @@
 Several options are supported and passable using the integration constructor. See the [configuration documentation](/platforms/javascript/guides/capacitor/session-replay/configuration/) for more details. To set up the integration, add the following to your Sentry initialization:
 
-<Note>The minimal supported version of Sentry's SDK for Capacitor is 0.11.1.</Note>
+<Note>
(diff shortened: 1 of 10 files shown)


src/platform-includes/session-replay/setup/javascript.capacitor.mdx
```diff
@@ -1,6 +1,8 @@
 Several options are supported and passable using the integration constructor. See the [configuration documentation](/platforms/javascript/guides/capacitor/session-replay/configuration/) for more details. To set up the integration, add the following to your Sentry initialization:
 
-<Note>The minimal supported version of Sentry's SDK for Capacitor is 0.11.1.</Note>
+<Note>
+  The minimal supported version of Sentry's SDK for Capacitor is 0.11.1.
+</Note>
```
shanamatthews, line 5: whoa new note formatting dropped. I feel like I've asked this before, but will it mess up our spacing for the notes that contain markdown & require newlines?

evanpurkhiser: So if your note contains markdown, when you put the newlines before and after the prose then remark will treat it as markdown. Without the newlines then it is treated fully as a react javascript component (as is the case here), in which case no markdown formatting will be applied.

shanamatthews: makes sense, thanks!

Issue #6843, opened 2023-05-04
Add SvelteKit Onboarding Wizard Docs


Lms24:
### Core or SDK?

Platform/SDK

### Which part? Which one?

SvelteKit

### Description

Now that docs are merged (#6808), we should create onboarding wizard docs

### Suggested Solution

* general onboarding
* replay onboarding
* profiling onboarding (?)

shanamatthews:
Just double checking you synced with @priscilawebdev on the plan & timing to move wizard docs to the product repo: getsentry/sentry#47470

Lms24:
Yes, Priscila is going to move the wizard next week. I'll open a PR tomorrow.

PR #6868, opened 2023-05-05
fix(js): Fix directory name for Remix profiling onboarding

Lms24:
While working on #6843 I noticed that the Remix profiling onboarding wizrad docs aren't in the correct directory that's expected from the Sentry frontend:

https://github.com/getsentry/sentry/blob/90e0d01d05e85946e1ffac6c8f55a8d33f72094e/static/app/components/profiling/ProfilingOnboarding/util.ts#L44

I also checked the Sentry UI with a remix project and the docs weren't displayed correctly.

This PR adjusts the directory name which should fix this.


priscilawebdev (approved):
lgtm 🙂

shanamatthews (approved):
This looks right to me!

PR #6867, opened 2023-05-05
feat(js): Add SvelteKit Onboarding Wizard docs

Lms24:
## Pre-merge checklist

- [x] Checked Vercel preview for correctness, including links
- [x] PR was reviewed and approved by any necessary SMEs
- [x] PR was reviewed and approved by a member of the [Sentry docs team](https://github.com/orgs/getsentry/teams/docs)

## Description of changes

This PR adds Onboarding Wizard docs for the new SvelteKit SDK:

* General onboarding (with the performance and replay permutations)
* Replay onboarding
* Profiling onboarding

closes #6843

PR #6869, opened 2023-05-05
ref(nextjs): Adjust wording and links in onboarding wizard docs

Lms24:
This adjusts the wording about "configuring" the SDK with the onboarding wizard in the NextJS wizard docs. 
Extracted from #6867 

Happy to discuss this but I think we should improve this in some way at least :)

### DIFF
--- a/src/wizard/javascript/nextjs/index.md
+++ b/src/wizard/javascript/nextjs/index.md
@@ -11,7 +11,7 @@ Add Sentry automatically to your app with the [Sentry wizard](https://docs.sentr
 npx @sentry/wizard@latest -i nextjs
 ```
 
-Sentry wizard will automatically patch your application:
+The Sentry wizard will automatically patch your application:
 
 - create `sentry.client.config.js` and `sentry.server.config.js` with the default `Sentry.init`.
 - create `next.config.js` with the default configuration.
--- a/src/wizard/javascript/nextjs/index.md
+++ b/src/wizard/javascript/nextjs/index.md
@@ -5,7 +5,7 @@ support_level: production
 type: framework
 ---
 
-Configure your app automatically with [Sentry wizard](https://docs.sentry.io/platforms/javascript/guides/nextjs/#configure).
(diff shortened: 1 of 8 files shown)


src/wizard/javascript/nextjs/index.md
```diff
@@ -5,7 +5,7 @@ support_level: production
 type: framework
 ---
 
-Configure your app automatically with [Sentry wizard](https://docs.sentry.io/platforms/javascript/guides/nextjs/#configure).
+Add Sentry automatically to your app with the [Sentry wizard](https://docs.sentry.io/platforms/javascript/guides/nextjs/#install):
```
Lms24, line 8: My motivation for making this change was that the wizard isn't mentioned in the `#configure` section of the "Getting Started" docs but rather in the `#install` section above it. 

Given that the installation command is mentioned directly below, once could argue that we don't need this sentence at all. In this case, though, I'd argue to link to the wizard (#install) in the section below where we talk about what changes the wizard actually makes. Happy to make this change if reviewers agree that this would be better.

priscilawebdev (approved):

src/wizard/javascript/nextjs/index.md
priscilawebdev: I think this sentence makes more sense, but I will be happy to bring this topic to a discussion afterward...In any case, thanks for improving it 🙂

shanamatthews: Thanks for making the change! I like having this intro sentence to orient folks who haven't heard of sentry wizard before.

We don't seem to be consistent with whether we refer to the wizard as "the Sentry wizard" or just "Sentry wizard".

I prefer with the "the". Do you mind adding "The" in line 14 as well so we're consistent in this doc at least?

shanamatthews (approved):

src/wizard/javascript/nextjs/index.md
Lms24: +1 for "the"

PR #6872, opened 2023-05-05
ref(prettier): Configure prettier to format mdx w/ defaults

evanpurkhiser:
Most of our users will be using the default prettier config, so we want
to format our codeblocks using this default


ryan953 (approved):

shanamatthews (approved):
👍

PR #6904, opened 2023-05-09
fix: fix bug with links being considered external

scefali:


src/components/smartLink.tsx
```diff
@@ -48,7 +48,7 @@ export default function SmartLink({
         urlObj.searchParams.set(key, value);
       }
     }
-    setForcedUrl(urlObj.toString());
+    setForcedUrl(urlObj.toString().replace(window.location.origin, ''));
```
evanpurkhiser, line 51: Let's add a comment, it's unclear to me what this is for

evanpurkhiser (approved):

PR #6913, opened 2023-05-10
ref(onboarding): Removing serveless cloud from the onboarding

priscilawebdev:
closes https://github.com/getsentry/sentry/issues/48687

@smeubank do we still need to keep documentation for serverless-cloud or shall I delete it too? https://docs.sentry.io/platforms/node/guides/serverless-cloud/


smeubank:
https://www.npmjs.com/package/@serverless/cloud

I am just realizing this "platform" seems to actually be for another "framework". it is not very popular but then the "problem" is actually just that the icon was wrong https://npmtrends.com/@serverless/cloud and we don't need to remove it, just add a different icon. Either way it seems almost no one creates projects with this platform icon

shanamatthews:
@priscilawebdev @smeubank - this looks straightforward, but let me know if you need any help 😄

PR #6922, opened 2023-05-10
ref(registries): Ensure registry fetching happens once

evanpurkhiser:
With the current implementation it was possible to fetch the app registry multiple times if multiple calls were made to `getList` before any one of them completed (since the cached data would not be populated yet)

This change abstracts the registries in such a way so that we ensure only one request is ever made, any subsiquent calls will wait for that request to complete and then use the resulting data.

Also abstracts the helpers for the generalized 'makeGenericRegistry' which is shared between Apps and SDKs


src/gatsby/utils/fetchCache.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    // If we're not already making a request do so now
+    if (activeFetch === null) {
+      activeFetch = fetchData();
+    }
+
+    await activeFetch;
+
+    return dataCache;
+  }
```
evanpurkhiser, line 46: This is where the magic happens, calling this function will only trigger the data fetch once, which was not true in the previous implementation

ehfeng: I thought you only got to use the phrase "This is where the magic happens" when you're on MTV Cribs.

scttcper (approved):
it looks the same

PR #6926, opened 2023-05-11
ref(js): use `sentry-docs` prefix

evanpurkhiser:
This matches sentry, and getsentry


src/components/pageGrid.tsx
```diff
@@ -2,9 +2,9 @@ import React from 'react';
 import {useLocation} from '@reach/router';
 import {graphql, useStaticQuery} from 'gatsby';
 
-import SmartLink from './smartLink';
+import {sortPages} from 'sentry-docs/utils';
 
-import {sortPages} from '~src/utils';
+import SmartLink from './smartLink';
```
ryan953, line 7: Can we use the prefix in all the cases incl these? Makes it much easier to grep for places that import a file when the import paths are all the same

ryan953 (approved):

shanamatthews (approved):
I don't have strong feelings about this, but looked ok to me!

PR #6937, opened 2023-05-11
feat(replay): Add "Custom Sampling" session

billyvg:
Add an example of when to defer loading of Replay SDK and setting sample rates at a later point in time.

### DIFF
--- a/src/platforms/javascript/common/session-replay/understanding-sessions.mdx
+++ b/src/platforms/javascript/common/session-replay/understanding-sessions.mdx
@@ -69,3 +69,24 @@ await replay.flush();
 ```
 
 In `session` mode, this will upload any pending recording data to Sentry. In `buffer` mode, this will upload any pending recording data to Sentry and then continue recording, the same as when an error is sampled with `replaysOnErrorSampleRate`.
+
+## Custom Sampling
+
+It is also possible to defer loading of the Replay integration. This is helpful to decouple the SDK initialization from Replay sampling rates. For example, if you have an external sampling service that is not immediately available.
+
+
+```javascript
+async function init(sessionSampleRate, errorSampleRate) {
+  const client = Sentry.getCurrentHub().getClient();
+  const options = client.getOptions() as BrowserClientReplayOptions;
+  options.replaysSessionSampleRate = sessionSampleRate;
+  options.replaysOnErrorSampleRate = errorSampleRate;
+
+  const replay = new Sentry.Replay({
+    maskAllText: true,
+    // additional SDK config, see https://docs.sentry.io/platforms/javascript/session-replay/configuration/
+  });
+
+  client.addIntegration(replay);
+}
+```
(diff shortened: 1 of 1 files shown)


src/platforms/javascript/common/session-replay/understanding-sessions.mdx
```diff
@@ -69,3 +69,24 @@ await replay.flush();
 ```
 
+
+## Custom Sampling
```
billyvg, line 73: I don't love this title -- thoughts @ryan953 ?

src/platforms/javascript/common/session-replay/understanding-sessions.mdx
```diff
@@ -69,3 +69,23 @@ await replay.flush();
...
+  options.replaysOnErrorSampleRate = errorSampleRate;
+
+  const replay = new Sentry.Replay({
+    maskAllText: true,
+    // additional SDK config, see https://docs.sentry.io/platforms/javascript/session-replay/configuration/
+  });
+
+  client.addIntegration(replay);
+}
+```
```
ryan953, line 91: just throwing in some edits to see what works.
i'd like to point out where the `Sentry` import comes from, but it would be better if it showed sdk specific import paths.. 

```suggestion
import * as Sentry from '@sentry/javascript';

function init(sessionSampleRate, errorSampleRate) {
  const client = Sentry.getCurrentHub().getClient();
  const options = client.getOptions() as BrowserClientReplayOptions;
  
  options.replaysSessionSampleRate = sessionSampleRate;
  options.replaysOnErrorSampleRate = errorSampleRate;

  const replay = new Sentry.Replay({
    maskAllText: true,
    // additional SDK config, see https://docs.sentry.io/platforms/javascript/session-replay/configuration/
  });

  client.addIntegration(replay);
}
```

billyvg: Yeah I took off the import because it depends on what framework lib they use :/

src/platforms/javascript/common/session-replay/understanding-sessions.mdx
ryan953: hrm, 
Async Sampling?

It mentions "defer loading" right away... but i think that's the solution
the problem is:
> [...] they asked if there is a way to set the sample rate for session replays after Sentry is instantiated.

.... so wait a sec... idk if this is even the correct page for this doc.

This could be in the config/setup pages, because we're talking about setting the sample rate after the error monitoring has been enabled.

I just found: https://docs.sentry.io/platforms/javascript/session-replay/#lazy-loading-replay which has the core mechanism here too (totally fine to duplicate).

ryan953: What if we edited the file a little further up:
```diff
  ## Default Session Initialization

- By default, Replay will immediately start a session in either `session` or `buffer` mode, depending on your `replaysSessionSampleRate` and `replaysOnErrorSampleRate`.
+ By default, Replay will immediately start a session when the `new Replay()` integration is added to the SDK client. The session will be in either `session` or `buffer` mode, depending on your `replaysSessionSampleRate` and `replaysOnErrorSampleRate`.
```

and put this new section above the `Manually Starting Replay` section. Title this new one `Manually call addIntegration()` maybe? Or `Defer Adding Replay Integration`

ryan953 (approved):

PR #6939, opened 2023-05-11
ref(node): Prefer built in node18 fetch

evanpurkhiser:


src/gatsby/utils/fetchCache.ts
```diff
@@ -24,8 +24,8 @@ export function makeFetchCache<DataType>({dataFetch, name}: Options) {
 
     async function fetchData() {
       try {
-        const result = await dataFetch();
-        dataCache = result.data;
```
evanpurkhiser, line 28: was kinda mixed up before having `dataFetch` return it's response but was typed as `any`

lforst (approved):
nice

shanamatthews (approved):
👍

PR #6948, opened 2023-05-12
feat(mute alerts): Add mute alerts docs

ceorourke:
Add a blurb about the mute alerts feature - will merge this Monday w/ GA.

### DIFF
--- a/src/docs/product/alerts/index.mdx
+++ b/src/docs/product/alerts/index.mdx
@@ -44,6 +44,10 @@ You can find a full list of available metric alerts in [Metric Alerts](/product/
 
 When you create a new project in [sentry.io](https://sentry.io), you can select a default issue alert. However, you can also [create your own alerts](/product/alerts/create-alerts/) to suit your team’s needs, using these [best practices](/product/alerts/best-practices/) as a guide.
 
+## Muting Alerts
+
+Issue alerts can be muted on the alert rule details page by clicking the "Mute" button. This will mute the alert from notifying you until you click "Unmute". If you want to mute the alert from firing entirely, select "Mute for everyone" from the dropdown.
+
 ## Notifications
 
 Besides alerts, Sentry sends you notifications about various things like [issue state changes](/product/issues/states-triage/), [release deploys](/product/releases/), and [quota usage](/product/accounts/quotas/). You can fine tune these notifications, as well as your personal alert settings, in **User Settings > Notifications**. Learn more about notifications and adjusting their associated settings in [the full documentation](/product/alerts/notifications/).


src/docs/product/alerts/index.mdx
```diff
@@ -44,6 +44,10 @@ You can find a full list of available metric alerts in [Metric Alerts](/product/
 
 When you create a new project in [sentry.io](https://sentry.io), you can select a default issue alert. However, you can also [create your own alerts](/product/alerts/create-alerts/) to suit your team’s needs, using these [best practices](/product/alerts/best-practices/) as a guide.
 
+## Muting Alerts
+
+Issue alerts can be muted on the alert rule details page by clicking the "Mute" button. This will mute the alert from notifying you until you click "Unmute". If you want to mute the alert from firing entirely, select "Mute for everyone" from the dropdown.
```
scefali, line 49: @ceorourke Might want to mention the required permission for `mute for everyone`

roggenkemper, line 49: maybe mention that if you unmute then it will first try to unmute the one for all then the one for you. seems like an edge case but might be good to have that written down

ceorourke: Hmm this feels like we should have a visual indication of whether it's muted for just you or everyone

roggenkemper: we have a visual indicator but looking at the code (https://github.com/getsentry/sentry/blob/master/src/sentry/api/endpoints/project_rule_details.py#LL97C36-L97C36) we don't give priority to everyone/just for yourself when determining the banner to show. the endpoint code for deleting does delete both right now (everyone/yourself) if they exist

ceorourke: Hmm looks like it was changed to anyone w/ `alerts:write` which is .... [members and up](https://github.com/getsentry/sentry/blob/master/src/sentry/conf/server.py#L1995)? Is that right? That doesn't seem good.

ceorourke: Oh somehow I totally missed the banner. Ehh I don't think it's necessary since the only way to get into a state where you'd be worried about what would happen would be if you used the API, right? Like if you muted the same alert for yourself and everyone and then clicked unmute?

roggenkemper: yeah that makes sense.

roggenkemper: we changed it to alerts:write since that's the permission you need to create/delete alerts. if they can delete an alert, it seems like muting is a lesser version of that (especially since it tracks who muted in the banner so people can see)

scefali: @ceorourke So it depends on the org. Some orgs let members create alerts, some orgs don't. It's an organization setting. Richard has the right idea: if you have the permission to delete an alert you should definitely have permission to mute.

ceorourke: 😬 okay I'm on board w/ matching that, separate conversation about whether that permission is a good idea lol. I wouldn't want some random person on another team to be able to delete an alert I depend on. Regardless, do we need to mention permissions when there is no restriction?

ceorourke: Oh I just saw this banner: https://docs.sentry.io/product/alerts/create-alerts/ I can copy that

scefali: @ceorourke yes exactly!

ceorourke: The wording seems wrong on that banner though, "The minimum role required to create alerts is member. " when actually it's only member if you toggle on that option ... shouldn't it be admin?

scefali: @ceorourke the default now is to allow members to create alerts so it's only if they explicitly turn it off (or they are a legacy org). I'll leave the language up to you.

src/docs/product/alerts/index.mdx
```diff
@@ -44,6 +44,16 @@ You can find a full list of available metric alerts in [Metric Alerts](/product/
 
 When you create a new project in [sentry.io](https://sentry.io), you can select a default issue alert. However, you can also [create your own alerts](/product/alerts/create-alerts/) to suit your team’s needs, using these [best practices](/product/alerts/best-practices/) as a guide.
 
+## Muting Alerts
+
+<Note>
+
+By default all users can mute alerts for everyone. Sentry users with manager or owner permissions can change the minimum role requirement in **Organization Settings > General Settings > Let Members Create and Edit Alerts**.
```
ceorourke, line 51: I put "organization settings" here but I noticed we don't explicitly say this elsewhere in the docs, we just call it "settings" which seems wrong? We have org and user level settings, and nothing just called settings ...

ceorourke: Nvm just noticed the " ⚙️ Settings" thing on the sidebar, idk if it's new or I've literally never noticed it before cause I always click the dropdown and then org settings.

roggenkemper: I think "settings" might just come from the sidebar which says settings but takes you to org settings.

roggenkemper (approved):

roggenkemper (approved):

src/docs/product/alerts/index.mdx
ceorourke: Yeah I am losing my mind here feeling like a mandela effect is happening except I'm the only person who thinks we never had ⚙️ Settings before.

PR #6963, opened 2023-05-16
fix(js): Eager load registry data

evanpurkhiser:
The problem here was the registry data was being loaded lazily **inside of a `eval` expression**. There was some logic to `await` the result of the eval, but it's unclear how well that works in a JavaScript VM and is likely cause of all of the intermittent 'Socket closed' errors we were seeing.

This explicitly updates the gatsby-remark-variables plugin to first load the data in the registries, and THEN handle the replacements using synchronous accessor functions from the registry.

### DIFF
--- a/src/gatsby/utils/fetchCache.ts
+++ b/src/gatsby/utils/fetchCache.ts
@@ -1,32 +1,3 @@
-/**
- * For some reason we sometimes see errors fetching from our registry, because
- * of that we retry the fetches
- *
- * The error was:
- *
- *   Client network socket disconnected before secure TLS connection was established
- */
-async function fetchRetry(url: string, opts: RequestInit & {retry?: number}) {
-  let retry = opts?.retry ?? 1;
-
-  while (retry > 0) {
-    try {
-      return await fetch(url, opts);
-    } catch (e) {
-      if (retry !== 0) {
-        // eslint-disable-next-line no-console
-        console.warn(`failed to fetch \`${url}\`. Retrying for ${retry} more times`);
-        retry = retry - 1;
-        continue;
-      }
-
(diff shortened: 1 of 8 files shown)


src/gatsby/utils/fetchCache.ts
```diff
@@ -1,32 +1,3 @@
-/**
- * For some reason we sometimes see errors fetching from our registry, because
- * of that we retry the fetches
- *
- * The error was:
- *
- *   Client network socket disconnected before secure TLS connection was established
- */
-async function fetchRetry(url: string, opts: RequestInit & {retry?: number}) {
```
evanpurkhiser, line 9: don't need this hack anymore.

src/gatsby/utils/fetchCache.ts
```diff
@@ -39,26 +10,26 @@ interface Options {
...
 /**
- * Creates a `getData` function that fetches with dataFetch only once.
- * Subsiquent calls will used the already fetched data.
+ * Creates a `ensureData` function that fetches from a URL only once. Subsequent
+ * calls will used the already fetched data.
  */
 export function makeFetchCache<DataType>({dataUrl, name}: Options) {
   let activeFetch: Promise<any> | null = null;
-  let dataCache: DataType | null = null;
+  let data: DataType | null = null;
```
evanpurkhiser, line 18: probably didn't need to rename a bunch of stuff in here but it's fine

shanamatthews (approved):
whew! let's wait for @lforst 's thoughts to merge, but this looks like a good fix to me

evanpurkhiser:
Gonna merge since this fixes stuff

PR #6965, opened 2023-05-16
fix(registry): Actually resolve registries

evanpurkhiser:


src/gatsby/utils/fetchCache.ts
```diff
@@ -1,3 +1,32 @@
+/**
+ * For some reason we sometimes see errors fetching from our registry, because
```
evanpurkhiser, line 2: Brought back the retry logic... it does seem like we actually need it for some reason?

shanamatthews (approved):
🙏

PR #6966, opened 2023-05-16
ref(variables): Update variable syntax to be less common

evanpurkhiser:


src/gatsby/plugins/gatsby-remark-variables/index.js
```diff
@@ -57,9 +57,8 @@ module.exports = async ({markdownAST, markdownNode}, options) => {
         try {
           result = scopedEval(expr, {...scope, page});
         } catch (err) {
-          // no-op. We previously had a warning here, but the we have so many
-          // "variable-like" constructs in our codeblocks that this becomes
-          // very spammy.
+          console.error(`Failed to interpolate expression: "${expr}"`);
+          throw err;
```
evanpurkhiser, line 61: This means builds will fail if you try and use variable interpolation without it able to resolve the variable.

shanamatthews (approved):
Let's see how this goesss!

PR #6968, opened 2023-05-16
ref(js): Replace deprecated `substr`

evanpurkhiser:


ryan953 (approved):
Args changed slightly between `substr` and `substring`... but there's no difference when the first arg is `0`.

https://stackoverflow.com/questions/3745515/what-is-the-difference-between-substr-and-substring

PR #6975, opened 2023-05-17
feat(kmp): improve docs

buenaflor:
Changes:
 - Remove Android prerequisite
 - Doc on how to upload dsym and proguard mapping files
 - Doc on how to access the SDK directly in the native platforms without shared code
 - Improved code sample in Shared initializers using expect/actual
 - Add notes that we recommend using Sentry Cocoa version 8.4.0 (this needs to be changed whenever we bump the version in the SDK)


src/platforms/kotlin-multiplatform/native-access-sdk.mdx
```diff
@@ -0,0 +1,66 @@
...
+If you need to access the SDK directly in your native platform without the shared module, you first need to change the dependency configuration from `implementation(...)` to `api(...)`:
+
+```kotlin {filename:shared/build.gradle.kts}
+val commonMain by getting {
+  dependencies {
+    api("io.sentry:sentry-kotlin-multiplatform:{{@inject packages.version('sentry.kotlin.kmp', '0.0.1-alpha.2') }}")
+  }
+}
+```
+
```
shanamatthews, line 16:
```suggestion

### Export the Framework
```

The sidebar gets confused if we don't follow markdown heading standards, so we need an H2 for these H3s

shanamatthews (approved):
Hey @buenaflor - thanks for adding these docs! 

There were some parts I found confusing so I tried to make suggestions to make it more obvious what you meant, but please check all my suggestions to ensure they make things make more sense, not less.

Also feel free to ping me if you have any questions or want to discuss!

src/platforms/kotlin-multiplatform/common/install/cocoapods.mdx
```diff
@@ -0,0 +1,50 @@
...
+title: Cocoapods
+description: "Learn about installing the Sentry Kotlin Multiplatform SDK with Cocoapods."
+---
+
+To install the Kotlin Multiplatform SDK and setup your Apple targets with Cocoapods, you need to add the following to your `build.gradle.kts` file in your shared module:
+
+<Note>
+  We recommend you to use the Sentry Cocoa version 8.4.0 as we cannot guarantee
+  correctness with other versions.
+</Note>
```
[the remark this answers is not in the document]
buenaflor: The KMP SDK uses 8.4.0 underneath (which should be bumped to the lastest with time) and the best case is that the user wouldn't need to take care of that, but sadly there are currently KMP platform limitations so the user needs to provide that dependency themselves and the guaranteed correctness is only ensured if the user uses the version used in the KMP SDK.

src/platforms/kotlin-multiplatform/common/install/cocoapods.mdx
```diff
@@ -0,0 +1,50 @@
+---
+title: Cocoapods
```
marandaneto, line 2: Why do we need this file if the `getting-started-install` section already have the cocoapods setup?

src/platforms/kotlin-multiplatform/common/install/cocoapods.mdx
```diff
@@ -0,0 +1,50 @@
...
+description: "Learn about installing the Sentry Kotlin Multiplatform SDK with Cocoapods."
+---
+
+To install the Kotlin Multiplatform SDK and setup your Apple targets with Cocoapods, you need to add the following to your `build.gradle.kts` file in your shared module:
+
+<Note>
+  We recommend using version 8.4.0 of the Sentry Cocoa SDK, as we cannot
+  guarantee correctness with other versions.
+</Note>
```
marandaneto, line 11: I'd say this is not a recommendation but rather required.

src/platforms/kotlin-multiplatform/common/install/swift-package-manager.mdx
```diff
@@ -0,0 +1,52 @@
...
+<Note>
+  We recommend using version 8.4.0 of the Sentry Cocoa SDK, as we cannot
+  guarantee correctness with other versions.
+</Note>
+
+Add the Sentry Cocoa SDK as a package in Xcode in your iOS app via **File > Add Packages**.
+Enter the Git repo URL in the search field:
+
+```text
+https://github.com/getsentry/sentry-cocoa.git
```
marandaneto, line 15: Can we pin point the 8.4.0 tag instead? otherwise there will be version mismatch

src/platforms/kotlin-multiplatform/dsym.mdx
```diff
@@ -0,0 +1,55 @@
+---
+title: Upload Debug Symbols
```
marandaneto, line 2: The file is called `dsym` but it's generic to debug symbols for Android/iOS, I'd rather rename the file.

src/platforms/kotlin-multiplatform/dsym.mdx
```diff
@@ -0,0 +1,55 @@
...
+```kotlin {filename:androidApp/build.gradle.kts}
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+}
+
+plugins {
+    id("com.android.application")
+    id("io.sentry.android.gradle") version "{{@inject packages.version('sentry.java.android.gradle-plugin', '3.0.0') }}"
```
marandaneto, line 29: I'd add either a code snippet to the plugin's feature `uploadNativeSymbols` or at least mention it.

src/platforms/kotlin-multiplatform/dsym.mdx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// Prevent Sentry from being included in the Android app through the AGP.
+configurations {
+    compileOnly {
+        exclude(group = "io.sentry", module = "sentry")
+        exclude(group = "io.sentry", module = "sentry-android")
+    }
+}
```
marandaneto, line 38: I'd rather add a plugin's code snippet for disabling the `autoInstallation.enabled` feature.

src/platforms/kotlin-multiplatform/native-access-sdk.mdx
```diff
@@ -0,0 +1,68 @@
+---
+title: Native Access to SDK
+sidebar_order: 8
+description: "How to directly access the SDK without the shared module."
+---
+
+If you need to access the SDK directly in your native platform without the shared module, you first need to change the dependency configuration from `implementation(...)` to `api(...)`:
```
marandaneto, line 7: Maybe add the SDK as `api` directly instead? so this section isn't even needed.

marandaneto:
@buenaflor the vercel preview isn't working for some reason.

src/platforms/kotlin-multiplatform/native-access-sdk.mdx
buenaflor: This might be confusing since using api only allows direct access to the sdk on java/android but not on iOS. That needs an extra setup in which the user needs to export the framework. 

The way the doc is currently set up (with api and exporting framework in the install docs) users will think this is mandatory to make the SDK work but that's just optional

src/platforms/kotlin-multiplatform/dsym.mdx
buenaflor: autoInstallation disable won't exclude `io.sentry.Sentry`, maybe a mix of both? something like this:

```
configurations {
    compileOnly {
        exclude(group = "io.sentry", module = "sentry")
    }
}

sentry {
  autoInstallation {
    enabled.set(false)
  }
}
```

src/platforms/kotlin-multiplatform/common/install/cocoapods.mdx
buenaflor: I assume users would expect all installation methods to be listed under `Apple Installation Methods` so if only SPM is present there it might not be obvious that it's in getting started especially if someone skips the getting started tutorial or tries searching for `sentry kmp cocoapods installation` through google.

src/platforms/kotlin-multiplatform/common/install/swift-package-manager.mdx
buenaflor: Added some extras to make this clear

src/platforms/kotlin-multiplatform/common/install/cocoapods.mdx
marandaneto: For me it's confusing that the [Install step](https://sentry-docs-fwte4vsy1.sentry.dev/platforms/kotlin-multiplatform/#install) already installs the Apple dep via `cocoapods` but just after comes the `Next Steps` with `Apple Installation Methods`, feel like you have to do it again, or additional steps.
We can improve this later if users find it confusing.

src/platforms/kotlin-multiplatform/dsym.mdx
romtsn: it should actually, if it's not, then it's a bug most likely

marandaneto: I'd add the code snippet for `autoInstallation` yes.

> autoInstallation disabled won't prevent installing io.sentry.Sentry

I don't follow that, disabling the `autoInstallation` is pretty much doing that, maybe @romtsn knows?

src/platforms/kotlin-multiplatform/native-access-sdk.mdx
marandaneto: Alright, let's keep it for now, we can improve it later if needed.

src/platforms/kotlin-multiplatform/initialization-strategies.mdx
```diff
@@ -62,42 +65,48 @@ func application(
...
+
+```kotlin {filename:commonMain/AppSetup.kt}
+import io.sentry.kotlin.multiplatform.Context
+
+expect fun initializeSentry(context: Context? = null)
+```
 
 ### Android
 
-```kotlin {filename:MainActivity.kt}
```
marandaneto, line 69: By removing the snippet from `YourApplication : Application` (Android) and `func application` (iOS), I wonder how users would know when to call `initializeSentry`.
It's important to know where this should be placed, can we keep this?

src/platforms/kotlin-multiplatform/dsym.mdx
buenaflor: I tested it out a couple times and could still use `io.sentry.Sentry` with `autoInstallation` false but not `SentryAndroid`, but I'll double check to make sure

src/platforms/kotlin-multiplatform/initialization-strategies.mdx
buenaflor: Yup makes sense

src/platforms/kotlin-multiplatform/common/install/cocoapods.mdx
shanamatthews: I agree this is a bit confusing. Maybe a note that says if you've followed the "Getting Started" setup you will have already done this would help.

src/platforms/kotlin-multiplatform/native-access-sdk.mdx
```diff
@@ -0,0 +1,68 @@
+---
+title: Native Access to SDK
```
shanamatthews, line 2:
```suggestion
title: Native Access to the SDK
```

shanamatthews (approved):
Looks great, thanks for the improvements @buenaflor !

marandaneto:
@buenaflor do you intend to finish this PR or should we take it? asking cus you are back only in July, right.

buenaflor:
Yeah, I'll take care of this one. In this case as an open source contribution :)

src/platforms/kotlin-multiplatform/dsym.mdx
buenaflor: continued in https://github.com/getsentry/sentry-docs/pull/7024
It works as intended and only 
```
sentry {
  autoInstallation {
    enabled.set(false)
  }
}
```

is needed

marandaneto:
Closing in favor of https://github.com/getsentry/sentry-docs/pull/7024

PR #6979, opened 2023-05-17
docs(express-sdk): Fix Express SDK wizard example code

spalmurray:
Fix `Tracing is not defined` when you try to run the example code here.


HazAT (approved):
Thank you 💯

PR #7004, opened 2023-05-23
ref: Remove deprecated client pages

HazAT:
I opened this PR to remove the old deprecated `/client` SDK pages.
It's shitty for our users that they land on these pages when visiting from Google - they don't provide any value anymore because if people have troubles or bugs in their old SDKs we tell them to upgrade.


Lms24 (approved):

cleptric (changes requested):
Sorry for pulling a ❌ here.

We still have users using old versions of our SDKs, mainly the Raven ones. These users still want to have access to the docs.

HazAT:
I am fine with not deleting, but we have to solve this - every time our internal search or a search engine guides a user to these pages, we fail.

They should only be accessible with direct links and contain a warning that this is a legacy SDK.

HazAT:
Closing in favor of https://github.com/getsentry/sentry-docs/pull/7007

PR #7033, opened 2023-05-26
Remove node primer

lforst:
I believe we should not immediately confront people landing on the getting started page with asynchronous context, which is a very advanced concept + we're already mentioning errors right below the note anyhow.

I would also like to remove the note about the legacy SDK documentation because the legacy sdk is REALLY old and doesn't deserve a spot on this page. The legacy SDK docs should really only be accessible via external links.


shanamatthews (approved):
This looks reasonable to me!

PR #7039, opened 2023-05-26
update JSON payload image

vivianyentran:
Event JSON payload image was outdated.


lizokm (approved):
Thanks for updating!

lforst:
Thank you!

PR #7052, opened 2023-05-30
Remove critical level

rodolfoBee:
As per dev docs, critical level is not supported: https://develop.sentry.dev/sdk/event-payloads/#optional-attributes

An event sent with level:critical has the level changed to fatal during server side processing


shanamatthews:
@rodolfoBee this isn't just true for Ruby, right?

It seems like we need to change this for all the SDKs.

rodolfoBee:
I did not notice it was on all SDKs, but yes it can be removed from all.

shanamatthews (approved):

Issue #7050, opened 2023-05-30
Out of date documentation: relay configuration options


rodolfoBee:
### Core or SDK?

Core Sentry product

### Which part? Which one?

Relay

### Description

The following pages are out of date in regards to a release made 2023-04-17:
https://docs.sentry.io/product/relay/options/
https://docs.sentry.io/product/relay/operating-guidelines/

The release has breaking changes in terms of reverse proxy usage and connections options:
https://github.com/getsentry/relay/releases
https://github.com/getsentry/relay/commit/e8b2acc2cedd38db4ab5bc5d5d13f9996ceca528

### Suggested Solution

Revise and update the documentation.

jan-auer:
The docs should also list the names of environment variables for options that can be configured through one.

PR #7062, opened 2023-05-31
fix(relay): Update relay configuration

TBS1996:
Update the docs regarding the deletion of some config options, as well as the removal of SSL from relay. 
Also lists the environmental variables that can be used to override the relay config options listed in the docs.

references: #7050 

preview: https://sentry-docs-git-relayconfig.sentry.dev/product/relay/options/

not sure how I should format the environment variable notes. 

## Pre-merge checklist


src/docs/product/relay/options.mdx
```diff
@@ -329,25 +337,6 @@ The total number of threads spawned will roughly be `2 * limits.max_thread_count
...
-
-The maximum number of pending connects to Relay. This corresponds to the
-backlog param of `listen(2)` in POSIX.
-
-`limits.max_connections`
-
-: _Integer: default: `25_000`_
-
-The maximum number of open incoming connections to Relay.
-
```
TBS1996, line 350: Deleted due to the following note in relay release `23.4.0`:

```
Connection config options max_connections, max_pending_connections, and max_connection_rate no longer have an effect. Instead, configure the reverse proxy to handle connection concurrency as needed.
```

src/docs/product/relay/options.mdx
```diff
@@ -106,7 +114,7 @@ response body is covered in this timeout.
 
 Timeout for establishing connections with the upstream in seconds.
 
-This includes SSL handshakes. Relay reuses connections when the upstream
```
TBS1996, line 109: Since SSL is no longer supported, this sentence is not relevant any longer.

src/docs/product/relay/options.mdx
```diff
@@ -6,7 +6,7 @@ redirect_from:
   - /meta/relay/options/
 ---
 
-Configuration for Relay is recorded in the file `.relay/config.yml`. To change this location, pass the `--config` option to any Relay command:
+Configuration for Relay is recorded either in the file `.relay/config.yml`, or when noted, will be overridden through an environment variable. To change the config location, pass the `--config` option to any Relay command:
```
TBS1996, line 9: @jan-auer  suggested that the docs should list the options that can be chosen from environment variables. Since this page doesn't show all the relay config options (by design i assume), I've only noted on the current shown options where they can be also configured by env-vars. 

I'm looking for feedback on whether we should have a separate page for all the environmental variables regardless, or keep it like this.

src/docs/product/relay/options.mdx
```diff
@@ -26,12 +26,16 @@ and `capture`
 Controls how Relay obtains the project configuration for events. For detailed
 explanation of these modes, see [Relay Modes](/product/relay/modes/).
 
+> enviroment variable: "RELAY_MODE".
```
iker-barriocanal, line 29: I'd put this next to the config name `relay.mode`, but this is probably something the docs team has an opinion on.

iker-barriocanal (approved):
Content-wise LGTM.

src/docs/product/relay/options.mdx
lizokm, line 9:
```suggestion
Configuration for Relay is recorded in either the file `.relay/config.yml` or gets overridden through an environment variable. To change the config location, pass the `--config` option to any Relay command:
```

src/docs/product/relay/options.mdx
lizokm: I'm wondering if it would be possible to turn all the lists on this page into tables (like this: https://docs.sentry.io/platforms/javascript/session-replay/configuration/) - I think it would make it easier to read.

TBS1996: I think that should be a separate PR, could you check out this draft?

https://github.com/getsentry/sentry-docs/pull/7084

src/docs/product/relay/options.mdx
```diff
@@ -6,34 +6,42 @@ redirect_from:
...
 
 ```shell
 ❯ ./relay run --config /path/to/folder
 ```
 
 All configuration keys are `snake_case`.
 
+## Credentials
+
+Credentials can be initiated as described in the [getting started page](https://sentry-docs-git-relayconfig.sentry.dev/product/relay/getting-started/?original_referrer=https%3A%2F%2Fgithub.com%2Fgetsentry%2Fsentry-docs%2Fpull%2F7062#creating-credentials), but can be overriden by setting the following environment variables:
```
iker-barriocanal, line 19: I don't know what's the best practice for links here, but you don't _need_ to write the full URL, see:
https://github.com/getsentry/sentry-docs/blob/55c4d20aaded7449ac21ec792f615a541952293e/src/platforms/android/using-ndk.mdx?plain=1#L19
I think we can get rid of the referrer parameter too?
@lizokm may be able to provide some guidance here.

src/docs/product/relay/options.mdx
```diff
@@ -6,34 +6,42 @@ redirect_from:
...
+
+secret key: `"RELAY_SECRET_KEY"`
+public key: `"RELAY_PUBLIC_KEY"`
+id: `"RELAY_ID"`
```
iker-barriocanal, line 23: This renders in a row. What do you think about creating an actual list for this?

src/docs/product/relay/options.mdx
```diff
@@ -6,34 +6,42 @@ redirect_from:
...
-| relay.port                 | Integer | default value: `3000` optional: `false`                                                       | The port to which Relay should bind for the unencrypted Relay HTTP server.                                                                                                                                                                                                                                                                                                                                                                      |
+| relay.mode                 | String  | default value: `managed` possible values: `managed, static, proxy, capture` optional: `false` | Controls how Relay obtains the project configuration for events. For detailed explanation of these modes, see [Relay Modes](/product/relay/modes/). `Environment variable: "RELAY_MODE."`                                                                                                                                                                                                                                                       |
```
iker-barriocanal, line 29: > optional: `false`

Seems confusing to me. What about `optional: no`? Typing the backticks suggests you can set `false`.

iker-barriocanal, line 29: If we're creating a table, I think it'd be less confusing if the environment variable is another column like `key`, instead of a note in the description. See how we do with [Next.js docs](https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/#use-configuration-files-for-source-map-upload), as an example.

src/docs/product/relay/options.mdx
lizokm, line 19:
```suggestion
Credentials can be initiated as described in the [getting started page](/product/relay/getting-started/), but can be overridden by setting the following environment variables:
```
:)

src/docs/product/relay/options.mdx
lizokm: I think a list will be easier to quickly read :)

src/docs/product/relay/options.mdx
lizokm: Ah yes, I tend to agree. I think the more we can split out the information the easier it will be to read <3

lizokm (approved):
LGTM!

src/docs/product/relay/options.mdx
TBS1996: i made it now an unordered list

src/docs/product/relay/options.mdx
TBS1996: this was my plan, but with 5 columns it will crash with the "ON THIS PAGE" index, is there a way to go around this? @lizokm

src/docs/product/relay/options.mdx
```diff
@@ -6,34 +6,42 @@ redirect_from:
...
 
 All configuration keys are `snake_case`.
 
+## Credentials
+
+Credentials can be initiated as described in the [getting started page](/product/relay/getting-started/), but can be overridden by setting the following environment variables:
+
+- secret key: `"RELAY_SECRET_KEY"`
+- public key: `"RELAY_PUBLIC_KEY"`
+- id: `"RELAY_ID"`
```
iker-barriocanal, line 23: Starting the sentences with uppercase look better to me, and I don't think we need the `"`s:
```suggestion
- Secret key: `RELAY_SECRET_KEY`.
- Public key: `RELAY_PUBLIC_KEY`.
- ID: `RELAY_ID`.
```

src/docs/product/relay/options.mdx
lizokm: @TBS1996 fixing the overlap of the "on this page" index is in our backlog to figure out, but we haven't fixed it yet, unfortunately. So I think we should just leave it as 4 columns for now and then we don't have to worry about the Size Limits table having one lonely entry :) 

In any case, I think this change will make the page much easier to read even if we don't have a separate column for environment variables!

PR #7063, opened 2023-05-31
remove critical level from JS

rodolfoBee:


shanamatthews (approved):
Thanks for fixing this!

PR #7073, opened 2023-06-01
ref(sveltekit): Update SvelteKit SDK compatibility note

Lms24:
Now that we support more adapters, we should update the compatibility note accordingly

ref https://github.com/getsentry/sentry-javascript/issues/8085

### DIFF
--- a/src/platforms/javascript/guides/sveltekit/manual-setup.mdx
+++ b/src/platforms/javascript/guides/sveltekit/manual-setup.mdx
@@ -240,7 +240,42 @@ export default {
 };
 ```
 
-Using the `sourceMapsUploadOptions` object is useful if the default source maps upload doesn't work out of the box, for instance, if you have a customized build setup or if you're using the SDK with a SvelteKit adapter other than the Node adapter.
+Using the `sourceMapsUploadOptions` object is useful if the default source maps upload doesn't work out of the box, for instance, if you have a customized build setup or if you're using the SDK with a SvelteKit adapter other than the [supported adapters](../#compatibility).
+
+### Overriding SvelteKit Adapter detection
+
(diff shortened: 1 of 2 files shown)


src/platforms/javascript/guides/sveltekit/manual-setup.mdx
```diff
@@ -240,7 +240,42 @@ export default {
 };
 ```
 
-Using the `sourceMapsUploadOptions` object is useful if the default source maps upload doesn't work out of the box, for instance, if you have a customized build setup or if you're using the SDK with a SvelteKit adapter other than the Node adapter.
+Using the `sourceMapsUploadOptions` object is useful if the default source maps upload doesn't work out of the box, for instance, if you have a customized build setup or if you're using the SDK with a SvelteKit adapter other than the [supported adapters](../#compatibility).
+
+### Overriding SvelteKit Adapter detection
+
+By default, `sentrySvelteKit` will try to detect the used SvelteKit adapter to configure source maps upload correctly for the respective adapter.
```
vivianyentran, line 247:
```suggestion
By default, `sentrySvelteKit` will try to detect your SvelteKit adapter to configure source maps upload correctly.
```

src/platforms/javascript/guides/sveltekit/manual-setup.mdx
```diff
@@ -240,7 +240,42 @@ export default {
...
+If we detect the wrong adapter our you're using another adapter than the [supported adapters](../#compatibility), you can override the adapter detection by passing the `adapter` option to `sentrySvelteKit`:
```
vivianyentran, line 248:
```suggestion
If you're not using one of the [supported adapters](../#compatibility) or the wrong one is detected, you can override the adapter detection by passing the `adapter` option to `sentrySvelteKit`:
```

vivianyentran (approved):
Thanks for updating the docs! Added some small suggestions to make it a little easier to read.

PR #7088, opened 2023-06-05
feat(alerts): Add new common tags

olksdr:
Add new common tags to the metric alerts. 

This adds:
- `os.name`
- `browser.name`
- `geo.country_code`
- `http.status_code`

PR #7096, opened 2023-06-05
Revert "feat(alerts): Add new common tags"

shanamatthews:
Reverts getsentry/sentry-docs#7088


shanamatthews:
@olksdr @lizokm - I'm reverting this change as the feature isn't live in Sentry yet, so we don't want to confuse users by having it in docs.

PR #7099, opened 2023-06-06
Rename "Uploading without Debug IDs" pages

lforst:
The rationale behind renaming "uploading without debug IDs" to "Legacy Uploading Methods" are two things:

- "Debug IDs" is not a term people in the ecosystem are familiar with.
- "Legacy Uploading Methods", in my opinion, is more what the page is actually about. I also thought about calling it "Uploading Sourcemaps with old Tools" which I am still open to change.

Q for reviewers: Do you also share the opinions above? Would you name the page differently?


mydea (approved):
IMHO this makes sense, as it indicates that we do not want people to do this anymore.

shanamatthews (approved):
I agree - good change.
I added redirects to your files in this PR, just as a best practice.

Issue #7105, opened 2023-06-07
Troubleshoot source maps dropdown menu showing the wrong page


rodolfoBee:
### Core or SDK?

Platform/SDK

### Which part? Which one?

JS - source maps

### Description

By expanding the left side menu in the JS SDK page Source Maps -> Troubleshooting source maps I expect to see the options to troubleshoot issues when uploading with and without debug ids.

I can find the option with debug ids, but instead of the page on how to troubleshoot release source maps i have the guide on how to upload source maps:
![Screenshot 2023-06-07 at 10 40 17](https://github.com/getsentry/sentry-docs/assets/64906376/c7accfe1-0a28-40e2-a8c3-5c48280600e7)

### Suggested Solution

Correct the page to "Troubleshooting without debug ids", and have a similar page to the old guide: https://web.archive.org/web/20230330234155/https://docs.sentry.io/platforms/javascript/sourcemaps/troubleshooting_js/

PR #7049, opened 2023-05-30
Rework sourcemap troubleshooting session for debug IDs

lforst:
I gave the current source map troubleshooting page a pass and updated it for the new debug ID process.

Disclaimer: We'll probably run into issues with debug IDs in the future which we don't know of yet. The current troubleshooting page is probably not very in-depth but I am planning to extend it whenever users run into issues.

PR #7126, opened 2023-06-12
Re-add troubleshooting guide for legacy source map upload

lforst:
Aims to fix https://github.com/getsentry/sentry-docs/issues/7105

This PR re-adds troubleshooting docs that were removed in https://github.com/getsentry/sentry-docs/pull/7049 to streamline docs around sourcemaps for debug IDs.

What I didn't consider back then was that some platforms can't use debug IDs and still need some troubleshooting guide for the legacy process.

### DIFF
--- a/src/platform-includes/sourcemaps/legacy-uploading-methods/javascript.mdx
+++ b/src/platform-includes/sourcemaps/legacy-uploading-methods/javascript.mdx
@@ -251,7 +251,7 @@ To use the automated verification process, install and configure the Sentry [Com
 
 For example, "Event ID: c2ad049f":
 
-![Image highlighting where to find the ID of an event on Sentry](event_id.png)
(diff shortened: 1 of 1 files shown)


shanamatthews:
Hmm, I know we're already nested many levels deep, but what would you think about putting the legacy troubleshooting content in a sub-page?

lforst:
> Hmm, I know we're already nested many levels deep, but what would you think about putting the legacy troubleshooting content in a sub-page?

@shanamatthews I thought about it but the nesting felt weird. I generally don't like it when "folders" in the side bar have a lot of content, and we do that like 3 times in the source maps flow already.

The nesting would be `source maps > troubleshooting > legacy upload > troubleshooting` which I don't find optimal.

The whole docs situation around source maps docs sucks so much right now but I don't know how to do it properly. It's like every solution has a giant drawback 😢

shanamatthews (approved):

PR #7136, opened 2023-06-13
Add React Router Info to Onboarding Wizard

ale-cota:
Since setting up React Router is key for getting parameterised transaction names, adding this suggestion to the next steps when the Performance product is selected.

### DIFF
--- a/src/wizard/javascript/react/with-error-monitoring-performance-and-replay.md
+++ b/src/wizard/javascript/react/with-error-monitoring-performance-and-replay.md
@@ -62,4 +62,4 @@ return <button onClick={() => methodDoesNotExist()}>Break the world</button>;
 
 - [Source Maps](https://docs.sentry.io/platforms/javascript/guides/react/sourcemaps/): Learn how to enable readable stack traces in your Sentry errors.
 - [React Features](https://docs.sentry.io/platforms/javascript/guides/react/features/): Learn about our first class integration with the React framework.
-- [Set Up React Router](https://docs.sentry.io/platforms/javascript/guides/react/configuration/integrations/react-router/): Configure routing, so Sentry can generate parameterized transaction names for a better overview in Performance Monitoring.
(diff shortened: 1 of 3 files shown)


priscilawebdev (approved):
left a suggestion but it looks good to me

src/wizard/javascript/react/with-error-monitoring-performance-and-replay.md
```diff
@@ -62,3 +62,4 @@ return <button onClick={() => methodDoesNotExist()}>Break the world</button>;
 
 - [Source Maps](https://docs.sentry.io/platforms/javascript/guides/react/sourcemaps/): Learn how to enable readable stack traces in your Sentry errors.
 - [React Features](https://docs.sentry.io/platforms/javascript/guides/react/features/): Learn about our first class integration with the React framework.
+- [Set Up React Router](https://docs.sentry.io/platforms/javascript/guides/react/configuration/integrations/react-router/): Configure routing, so Sentry can generate parameterized transaction names for a better overview in Performance Monitoring.
```
ale-cota, line 65:
```suggestion
- [React Router](https://docs.sentry.io/platforms/javascript/guides/react/configuration/integrations/react-router/): Configure routing, so Sentry can generate parameterized transaction names for a better overview in Performance Monitoring.
```

smeubank:
How much work would it be to add a toggle which says with or without react router? can be off by default

stretch: link to doc "How to detect if your app uses react-router"

ale-cota:
> How much work would it be to add a toggle which says with or without react router? can be off by default
What should this control? Would it slap the setup instructions on the page directly? I see they're also version dependent, so I think it requires a better design approach. I'd leave just the link in "Next Steps" for now because it's already an improvement to the existing and will create a separate ticket to investigate how we could introduce the instructions too.

smeubank:
> What should this control?
you are totally right next steps is good first win

2nd: the argument would be to just add it to the snippet with a comment that says your can remove or check other version in docs

the extra mile
The toggle (or really radial because it can't be more than one selected at once) would allow you turn on react router and select a version

vivianyentran (approved):

PR #7142, opened 2023-06-13
Fix link to Apple debug symbols

kahest:
Fix link to Apple debug symbols page from the RN manual setup docs. This is the only remaining reference to a `/client` page from a non-`/client` page that I could find.

EDIT: since the section was specific to bitcode, and bitcode has been deprecated months ago, we decided to remove the section

### DIFF
--- a/src/platforms/react-native/manual-setup/manual-setup.mdx
+++ b/src/platforms/react-native/manual-setup/manual-setup.mdx
@@ -78,7 +78,7 @@ $SENTRY_CLI debug-files upload "$INCLUDE_SOURCES_FLAG" "$DWARF_DSYM_FOLDER_PATH"
 ```
 
 For bitcode enabled builds via iTunes Connect, additional steps are required.
-Follow the instructions in [Sentry's bitcode documentation](/clients/cocoa/dsym/#dsym-with-bitcode) to set up uploads of symbols for all build variants.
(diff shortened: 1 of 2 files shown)


src/platforms/react-native/manual-setup/manual-setup.mdx
```diff
@@ -78,7 +78,7 @@ $SENTRY_CLI debug-files upload "$INCLUDE_SOURCES_FLAG" "$DWARF_DSYM_FOLDER_PATH"
 ```
 
 For bitcode enabled builds via iTunes Connect, additional steps are required.
-Follow the instructions in [Sentry's bitcode documentation](/clients/cocoa/dsym/#dsym-with-bitcode) to set up uploads of symbols for all build variants.
+Follow the instructions in [Sentry's bitcode documentation](/platforms/apple/dsym/) to set up uploads of symbols for all build variants.
```
kahest, line 81: @krystofwoldrich maybe we should even get rid of this section (or rather link to https://docs.sentry.io/platforms/apple/dsym/#dsym-with-bitcode) since bitcode is deprecated

krystofwoldrich, line 81: As you said, I also think we can remove it.
```suggestion
```

krystofwoldrich: I also think we can remove it.
https://github.com/getsentry/sentry-docs/pull/7142#discussion_r1228357628

shanamatthews (approved):

PR #7171, opened 2023-06-15
fix(ui): keyword dropdown selector

evanpurkhiser:
It was not opening because the clicked element was the referenceEl

### DIFF
--- a/src/components/codeBlock.tsx
+++ b/src/components/codeBlock.tsx
@@ -13,7 +13,7 @@ import {CodeContext} from './codeContext';
 
 const KEYWORDS_REGEX = /\b___(?:([A-Z_][A-Z0-9_]*)\.)?([A-Z_][A-Z0-9_]*)___\b/g;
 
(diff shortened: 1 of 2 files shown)


src/components/codeBlock.tsx
```diff
@@ -35,9 +35,10 @@ function makeKeywordsClickable(children: React.ReactChildren) {
       }
       arr.push(
         <KeywordSelector
+          key={lastIndex}
```
evanpurkhiser, line 38: key was missing

src/components/codeBlock.tsx
```diff
@@ -90,7 +91,11 @@ function KeywordSelector({keyword, group, index}: KeywordSelectorProps) {
     ],
   });
 
-  useOnClickOutside({current: dropdownEl}, () => isOpen && setIsOpen(false));
+  useOnClickOutside({
+    ref: {current: referenceEl},
```
evanpurkhiser, line 95: This is the actual fix here

priscilawebdev (approved):

PR #7178, opened 2023-06-15
unify troubleshooting location

stephanie-anderson:
Unifies the location of troubleshooting help
- always a separate page (so users find it easily through the navigation)
- always on the bottom of the section (as last navigation item)


lizokm (approved):
LGTM! Thanks @stephanie-anderson!

PR #7197, opened 2023-06-16
fix: useNavigate -> gatsby/navigate

evanpurkhiser:
useNavigate seems to no longer be a thing in some newish version of gatsby

Fixes DOCS-5YW


shanamatthews (approved):
🤞

PR #7227, opened 2023-06-21
Update ratelimits.mdx

t-h-man:
There is no default value for rate limits and every endpoint can have a different one.


shanamatthews:
Left a grammar tweak. @ceorourke - is this correct? I can see that you added the sentence that's being removed here.

Issue #7206, opened 2023-06-19
DSN isn't enough to use the cli


vasilvestre:
### Core or SDK?

Platform/SDK

### Which part? Which one?

Next/Node

### Description

I use the DSN to initialize my Sentry, it seems like an auth token is wanted.

sentry reported an error: Authentication credentials were not provided.

### Suggested Solution

Use the DSN or add to documentation that an auth token is required

vasilvestre:
> DSNs are public and can therefore not be used to authenticate when uploading source maps. Can you share a bit more about the problem you were facing. What version of the Next.js sdk were you using? What setup flow were you using? Did you use the wizard?

I was misled reading the documentation, I'm not asking to create a new release but it does anyway and ask for a TOKEN that isn't in the guide. Maybe adding a note to inform about this would be good?

lforst:
Ok, I think I understand.

We should add references to auth tokes to these docs here: https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/#configure-source-maps

vasilvestre:
> Ok, I think I understand.
> 
> We should add references to auth tokes to these docs here: https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/#configure-source-maps

Yes exactly, the DX will be even greater :)

PR #7254, opened 2023-06-23
 Add auth token to manual Next.js setup snippet

lforst:
Fixes https://github.com/getsentry/sentry-docs/issues/7206#issuecomment-1602174581

The SDK will scream at you if you don't provide an auth token so we should add it to the manual setup guide.


shanamatthews (approved):
language nits

Issue #3611, opened 2021-05-20
Document if Mobile SDKs are supported for library authors


marandaneto:
## Issue

Document if Mobile SDKs are supported for library authors.
Often we get questions like https://forum.sentry.io/t/restrict-sentry-events-just-from-the-android-sdk-library/13977/3

## Description

There are a few issues when using the Android SDK for libraries.
eg you can't use SentryAndroid#init, you also have to disable the auto-init mode.
You can't use the Sentry#capture Events.
You can't add the Sentry Android Gradle plugin for libs, hence you can't demangle events if minified by the final App.
Ideally, you'd create your own instance of Hub and pass it around to capture your own events, but Hub does not add all the default integrations, so you lose Device context, uncaught handler, etc...

### Suggested Solution

Either we documented as non supported and explain why or we document how to do it, creating a SentryOptions and Hub

getsentry-release:
Routing to @getsentry/team-mobile for [triage](https://develop.sentry.dev/processing-tickets/#3-triage). ⏲️

PeloWriter:
Looking at the history of this issue, I'm adding the label `Needs More Info` so we can capture the need for discussion on this topic.

sagrawal31:
Looking forward to this.

marandaneto:
@PeloWriter we've decided that no Sentry SDKs should be embedded into any 3rd party libraries/SDKs.
Only final apps should embed the Sentry SDKs.

Where should we place this info in the docs? this should be visible across all the SDKs, maybe something your team can help us with? thanks

sagrawal31:
Thanks, @marandaneto, for the update. May I also know the reason for this decision, please?

marandaneto:
@sagrawal31 there are a few, they will probably be available on this page, but a TLDR is: The final App owns the data, not SDKs/libraries, so only them should decide whether to capture or not.
The final App authors can file issues for libraries/SDKs with proper stack traces and erased PII if necessary.

sagrawal31:
Sure @marandaneto. I would like to add just in case you guys might want to reconsider. 

As an SDK provider, we want to make sure that the app is not crashing or behaving differently because of our SDK that's where we thought of integrating Sentry (and we did).

I read your earlier comments as well (about using Hub) but [this](https://github.com/letscooee/cooee-android-sdk/blob/master/cooee-android-sdk/src/main/java/com/letscooee/utils/SentryHelper.java#L85-L105) is how we have got it integrated into our SDK.

[...]

PR #7278, opened 2023-06-27
Add section to sentry SDKs and third party SDKs

marandaneto:
## Pre-merge checklist

## Description of changes

Closes https://github.com/getsentry/sentry-docs/issues/3611

I've not found a common doc to place this, so added it only to Mobile data privacy for Mobile right now, ideas where I can place this that is common content for all SDKs?

## Legal Boilerplate

## Extra resources


src/docs/product/security/mobile-privacy.mdx
```diff
@@ -54,3 +54,12 @@ No. Sentry does not cause your app to contain any ads.
 ### What do I need to disclose to Google Play?
 
 You would need to disclose and/or include in your app privacy policy all types of data you are collecting through your app, including data you are sending to Sentry.
+
+## Sentry SDKs for third-party SDKs and libraries
+
+Sentry SDKs are not officially supported and recommended for third-party SDKs and libraries due to some technical reasons.
```
cleptric, line 60:
```suggestion
Sentry SDKs are not officially supported and recommended for use in third-party SDKs and libraries due to technical limitations.
```

src/docs/product/security/mobile-privacy.mdx
```diff
@@ -54,3 +54,12 @@ No. Sentry does not cause your app to contain any ads.
...
+
+- Sentry SDKs are not designed to be used in third-party SDKs and libraries.
+- Version conflicts may occur between Sentry SDKs.
+- [Signal handlers](https://en.wikipedia.org/wiki/C_signal_handling) may overwrite each other leading to one of the Sentry SDKs not operating normally.
+- third-party SDKs and libraries may leak users' PII data to the third-party Sentry's organization.
```
kamilogorek, line 65:
```suggestion
- Third-party SDKs and libraries may leak users' PII data to the third-party Sentry's organization.
```

src/docs/product/security/mobile-privacy.mdx
```diff
@@ -54,3 +54,12 @@ No. Sentry does not cause your app to contain any ads.
 ### What do I need to disclose to Google Play?
 
 You would need to disclose and/or include in your app privacy policy all types of data you are collecting through your app, including data you are sending to Sentry.
+
+## Sentry SDKs for third-party SDKs and libraries
+
+Sentry SDKs are not officially supported and recommended for third-party SDKs and libraries due to some technical reasons.
+
+- Sentry SDKs are not designed to be used in third-party SDKs and libraries.
+- Version conflicts may occur between Sentry SDKs.
```
cleptric, line 63:
```suggestion
- Version conflicts may occur if multiple Sentry SDKs are installed in the same application.
```

src/docs/product/security/mobile-privacy.mdx
```diff
@@ -54,3 +54,12 @@ No. Sentry does not cause your app to contain any ads.
 ### What do I need to disclose to Google Play?
 
 You would need to disclose and/or include in your app privacy policy all types of data you are collecting through your app, including data you are sending to Sentry.
+
+## Sentry SDKs for third-party SDKs and libraries
```
lforst, line 58: I really like that we're documenting this! 👌 

I have one small point of feedback though: I would differentiate between "SDK use inside libraries" and "third-party SDKs" a bit more and separate the issues they introduce.

The main drawback of 3rd party SDKs is that they may not be as careful with PII, they will lag behind in feature support, and may not always support other tooling we have.

The main drawback of SDKs in libraries is, like you said, that multiple Sentry instances may conflict with each other, however, that is not a problem per se in 3rd party SDKs.

What do you think about splitting this up into two separate sections?

In general, this is very nice though so I won't block and I'll slap approve :)

lforst (approved):

src/docs/product/security/mobile-privacy.mdx
marandaneto: Not sure if it's needed, the title should be clear about that, otherwise, I'd apply this to the title.
Every item of the bullet point list is about having multiple Sentry SDKs in the same application.

src/docs/product/security/mobile-privacy.mdx
marandaneto: @lforst not sure I understand your point, SDKs, and libraries are pretty much a synonym here, if a 3rd party SDK or a 3rd party library embeds a  Sentry SDK in it, all the problems occur anyway.
Are you maybe referring to a Sentry SDK integration? which is not dependent on the core SDK itself? Or maybe something specific to the JS world?

src/docs/product/security/mobile-privacy.mdx
marandaneto: Actually, maybe it makes sense to apply this sentence to the 2nd and 3rd bullet points.
the 1st and 4th items are not specific to multiple sentry SDKs in the same application.

src/docs/product/security/mobile-privacy.mdx
```diff
@@ -54,3 +54,12 @@ No. Sentry does not cause your app to contain any ads.
...
+
+Sentry SDKs are not officially supported and recommended for use in third-party SDKs and libraries due to technical limitations.
+
+- Sentry SDKs are not designed to be used in third-party SDKs and libraries.
+- Version conflicts may occur if multiple Sentry SDKs are installed in the same application.
+- [Signal handlers](https://en.wikipedia.org/wiki/C_signal_handling) may overwrite each other leading to one of the Sentry SDKs not operating normally if multiple Sentry SDKs are installed in the same application.
+- Third-party SDKs and libraries may leak users' PII data to the third-party Sentry's organization.
```
bitsandfoxes, line 65:
```suggestion
- Third-party SDKs and libraries may leak users' PII data to the third-party's Sentry organization.
```

bitsandfoxes (approved):

src/docs/product/security/mobile-privacy.mdx
lforst: Well there are 3rd party SDKs that consume first party SDKs as dependencies. For example: https://github.com/nuxt-community/sentry-module

Libraries (unrelated to error monitoring or Sentry) that include Sentry SDKs are in a different class of problem because downstream users probably don't expect libraries to interfere with their Sentry setups. I think there is an inherent difference between third party SDKs (like the Nuxt SDK) and random libraries calling `Sentry.init()`.

marandaneto: @lforst so if the `sentry-module` calls `Sentry.init` https://github.com/nuxt-community/sentry-module/blob/63d79061de0ac1edc2a933cb25b530c34f297e64/src/hooks.ts#L157
Someone adds `sentry-module` into their own app, they also add the Sentry JS SDK directly and also call `Sentry.init`, what's gonna be the behavior? where does the event go? the `sentry-module`s DSN or the App's DSN that called `Sentry.init` again?

marandaneto: Anyway, since this section is Mobile only for now, we can tailor it to other platforms if needed, I don't think the `sentry-module` use case applies to Mobile in general.

shanamatthews:
> I've not found a common doc to place this, so added it only to Mobile data privacy for Mobile right now, ideas where I can place this that is common content for all SDKs?

Chatted with @vivianyentran about this - I don't think this is the best spot for these docs.

Could we put general info about using Sentry in SDKs on the platforms index page? src/platforms/index.mdx
If there's SDK-specific info, we could put that in a note on SDK getting started pages.

src/docs/product/security/mobile-privacy.mdx
vivianyentran: Most of this information except for the last bullet isn't relevant to Data Privacy for Mobile. I think it makes more sense to move this to the [platforms page](https://docs.sentry.io/platforms/), perhaps as a note before the second platforms heading listing all platforms.

src/docs/product/security/mobile-privacy.mdx
```diff
@@ -54,3 +54,12 @@ No. Sentry does not cause your app to contain any ads.
...
 You would need to disclose and/or include in your app privacy policy all types of data you are collecting through your app, including data you are sending to Sentry.
+
+## Sentry SDKs for third-party SDKs and libraries
+
+Sentry SDKs are not officially supported and recommended for use in third-party SDKs and libraries due to technical limitations.
+
+- Sentry SDKs are not designed to be used in third-party SDKs and libraries.
+- Version conflicts may occur if multiple Sentry SDKs are installed in the same application.
+- [Signal handlers](https://en.wikipedia.org/wiki/C_signal_handling) may overwrite each other leading to one of the Sentry SDKs not operating normally if multiple Sentry SDKs are installed in the same application.
+- Third-party SDKs and libraries may leak users' PII data to the third-party's Sentry organization.
```
vivianyentran, line 65: Since this bullet is the only one related to the topic of this page, I think it could be moved to the general section, perhaps included as a warning under the first question.

src/docs/product/security/mobile-privacy.mdx
```diff
@@ -54,3 +54,12 @@ No. Sentry does not cause your app to contain any ads.
 ### What do I need to disclose to Google Play?
 
 You would need to disclose and/or include in your app privacy policy all types of data you are collecting through your app, including data you are sending to Sentry.
+
+## Sentry SDKs for third-party SDKs and libraries
+
+Sentry SDKs are not officially supported and recommended for use in third-party SDKs and libraries due to technical limitations.
```
vivianyentran, line 60:
```suggestion
Sentry SDKs are not officially supported nor recommended for use in third-party SDKs and libraries due to technical limitations:
```

src/docs/product/security/mobile-privacy.mdx
marandaneto: Moved to the suggested page.

marandaneto:
> > I've not found a common doc to place this, so added it only to Mobile data privacy for Mobile right now, ideas where I can place this that is common content for all SDKs?
> 
> Chatted with @vivianyentran about this - I don't think this is the best spot for these docs.
> 
> Could we put general info about using Sentry in SDKs on the platforms index page? src/platforms/index.mdx If there's SDK-specific info, we could put that in a note on SDK getting started pages.

Sounds good, added it here as suggested https://github.com/getsentry/sentry-docs/pull/7278/commits/c3c8c43514a831bb676ff1a5be3b5413672444b7

marandaneto:
Screenshot 2023-06-28 at 09 55 54

@vivianyentran @shanamatthews let me know what you think, feel free to hit the merge button if all is good :) thanks for the feedback.

src/platforms/index.mdx
```diff
@@ -13,6 +13,17 @@ To report to Sentry you’ll need to use a language-specific SDK. The Sentry tea
 
 If your use-case is very specific, or not covered by Sentry, you'll find documentation on how to report events on your own using our [_SDK API_](https://develop.sentry.dev/sdk/overview/).
 
+<Note>
+
+Sentry SDKs are not officially supported nor recommended for use in third-party SDKs and libraries due to technical limitations:
+
+- Sentry SDKs are not designed to be used in third-party SDKs and libraries.
```
lforst, line 20:
```suggestion
- Sentry SDKs are not designed to be used in plugins, packages, and libraries (for example payment SDKs or widgets to be embedded on websites)
```

@marandaneto and I clarified that "third-party SDKs" here meant SDKs completely unrelated to Sentry or error monitoring. Since I find "third-party SDKs" a bit ambiguous in this context, I vote we rephrase this a bit.

marandaneto, line 20: @shanamatthews @vivianyentran would be possible to rephrase this sentence less ambiguously?

I'll explain what we want to achieve with this sentence:

We don't officially support/recommend Sentry SDKs being added to 3rd party SDKs and libraries, that is correct, for example, let's say that Stripe SDK embeds the Sentry SDK, customers add the Stripe SDK, customers also add the Sentry SDK, so technically there are 2 Sentry SDKs installed and this is a problem.

On the other hand, it's ok that 3rd party SDKs leverage Sentry SDKs, for example, https://github.com/nuxt-community/sentry-module, technically `nuxt/sentry` is also an SDK/Library, and this is fine, because in this case, `nuxt/sentry` acts as a Sentry SDK itself (but maintained by the community since we don't offer support to Nuxt ourselves), people would not add Sentry SDK directly again because this is the function of `nuxt/sentry`, and in this case, we do support and recommend.

In our sentences, we wanna address only the first use case, but not the 2nd, how can we phrase this better?

Maybe @lforst can add more context, let us know if this makes sense.

src/platforms/index.mdx
```diff
@@ -13,6 +13,17 @@ To report to Sentry you’ll need to use a language-specific SDK. The Sentry tea
 
 If your use-case is very specific, or not covered by Sentry, you'll find documentation on how to report events on your own using our [_SDK API_](https://develop.sentry.dev/sdk/overview/).
 
+<Note>
+
+Sentry SDKs are not officially supported nor recommended for use in third-party SDKs and libraries due to technical limitations:
```
bruno-garcia, line 18: The reasons are explained below and are beyond technical limitations so simpler is better here:

```suggestion
Sentry SDKs are not officially supported nor recommended for use in third-party SDKs and libraries:
```

src/platforms/index.mdx
vivianyentran: It sounds like the problem really comes from having more than one Sentry SDK installed on an app. What do you think about reframing the entire note like this:

It is bad practice to use Sentry SDKs in third-party plugins, packages, and libraries (such as payment SDKs or embeddable widgets) meant to be consumed by other apps. If an app uses a Sentry SDK, as well as a third-party package with its own Sentry SDK installed, then the following issues can occur: 

- Version conflicts may occur from having multiple Sentry SDKs installed in the same application.
- Signal handlers may overwrite each other, leading to one of the Sentry SDKs operating incorrectly. 
- Third-party SDKs and libraries may leak users’ PII to the third-party’s Sentry organization.

marandaneto: Thanks, applied the suggestion.

PR #7296, opened 2023-06-29
fix(apple): Clarify starting SDK on main thread

philipphofmann:
## Pre-merge checklist

## Description of changes

Clarify we recommend starting the SDK on the main thread.

### DIFF
--- a/src/platform-includes/getting-started-config/apple.mdx
+++ b/src/platform-includes/getting-started-config/apple.mdx
@@ -1,4 +1,4 @@
-You should also initialize the SDK as soon as possible, such as in your AppDelegate `application:didFinishLaunchingWithOptions` method:
+We recommend initializing the SDK on the main thread as soon as possible, such as in your AppDelegate `application:didFinishLaunchingWithOptions` method:
 
 ```swift {tabTitle:Swift}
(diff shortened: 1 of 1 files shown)


src/platform-includes/getting-started-config/apple.mdx
```diff
@@ -1,4 +1,4 @@
-You should also initialize the SDK as soon as possible, such as in your AppDelegate `application:didFinishLaunchingWithOptions` method:
+We recommend initializing the SDK on the main thread as soon as possible, such as in your AppDelegate `application:didFinishLaunchingWithOptions` method:
```
brustolin, line 1: m: Should we clarify that we will access ui related classes when automatic tracing is enabled, and if the SDK is initialized in a background thread UIKit will complain about it?

brustolin (approved):
LGTM.

Only one suggestion

src/platform-includes/getting-started-config/apple.mdx
philipphofmann: I don't think we should mention that. Even without these features, you should call us on the main thread. Otherwise, you might miss some crashes.

shanamatthews (approved):

PR #7297, opened 2023-06-29
[Python] Added deprecation message to `propagate_traces` in CeleryIntegration

antonpirker:
The option `propagate_traces` in `CeleryIntegration` is deprecated. 

This adds a message plus a link to the documentation where one can learn how to replace the option.


antonpirker:
We will not deprecate this yet, I will close this but keep it for later.

Issue #7293, opened 2023-06-29
Profiling default mode


nakanaa:
### Core or SDK?

Platform/SDK

### Which part? Which one?

Profiling

### Description

The page at https://docs.sentry.io/platforms/node/profiling/ says that "the default profiling mode (lazy) ..." but the default seems to have been changed already in February: https://github.com/getsentry/profiling-node/pull/102

### Suggested Solution

Fix accordingly to changes

JonasBa:
@nakanaa You are absolutely correct - we missed that when we changed it. We'll get that updated asap, thanks so much!

PR #7317, opened 2023-07-04
fix(profiling): update profiler mode docs

JonasBa:
Fixes https://github.com/getsentry/sentry-docs/issues/7293

### DIFF
--- a/src/platforms/node/common/profiling/index.mdx
+++ b/src/platforms/node/common/profiling/index.mdx
@@ -64,7 +64,7 @@ Under the hood, the Sentry profiler uses V8's [CpuProfiler](https://v8docs.nodes
 
 ## Runtime Flags
 
-The default mode of the v8 CpuProfiler is [kEagerLoggin](https://v8docs.nodesource.com/node-18.2/d2/dc3/namespacev8.html#a874b4921ddee43bef58d8538e3149374) which enables the profiler even when no profiles are active - this is good because it makes calls to startProfiling fast at the tradeoff for constant CPU overhead. The behavior can be controlled via the SENTRY_PROFILER_LOGGING_MODE environment variable with values of eager|lazy. If you opt to use the lazy logging mode, calls to startProfiling may be slow (depending on environment and node version, it can be in the order of a few hundred ms).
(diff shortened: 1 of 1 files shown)


src/platforms/node/common/profiling/index.mdx
```diff
@@ -64,13 +64,13 @@ Under the hood, the Sentry profiler uses V8's [CpuProfiler](https://v8docs.nodes
 
 ## Runtime Flags
 
-The **Profiling** package enables you to run the profiler in either [eager or lazy mode](https://v8docs.nodesource.com/node-18.2/d2/dc3/namespacev8.html#a874b4921ddee43bef58d8538e3149374). When a profile is started while no other profiles are running, the default profiling mode (lazy) sometimes causes `startProfiling` calls to take a long time (sometimes in the range of a couple hundred milliseconds). The benefit of lazy logging is that it doesn't add any overhead when profiles aren't running.
+The default mode of the v8 CpuProfiler is [kEagerLoggin](https://v8docs.nodesource.com/node-18.2/d2/dc3/namespacev8.html#a874b4921ddee43bef58d8538e3149374) which enables the profiler even when no profiles are active - this is good because it makes calls to startProfiling fast at the tradeoff for constant CPU overhead. The behavior can be controlled via the SENTRY_PROFILER_LOGGING_MODE environment variable with values of eager|lazy. If you opt to use the lazy logging mode, calls to startProfiling may be slow (depending on environment and node version, it can be in the order of a few hundred ms).
```
Zylphrex, line 67: Is this a typo?

```suggestion
The default mode of the v8 CpuProfiler is [kEagerLogging](https://v8docs.nodesource.com/node-18.2/d2/dc3/namespacev8.html#a874b4921ddee43bef58d8538e3149374) which enables the profiler even when no profiles are active - this is good because it makes calls to startProfiling fast at the tradeoff for constant CPU overhead. The behavior can be controlled via the SENTRY_PROFILER_LOGGING_MODE environment variable with values of eager|lazy. If you opt to use the lazy logging mode, calls to startProfiling may be slow (depending on environment and node version, it can be in the order of a few hundred ms).
```

JonasBa: Oh yes, def

PR #7320, opened 2023-07-05
Document Remix v2 support.

onurtemizkan:
Documents: https://github.com/getsentry/sentry-javascript/pull/8415

### DIFF
--- a/src/platform-includes/getting-started-config/javascript.remix.mdx
+++ b/src/platform-includes/getting-started-config/javascript.remix.mdx
@@ -94,7 +94,7 @@ const createSentryRequestHandler =
 app.all("*", createSentryRequestHandler(/* ... */));
 ```
 
-Also, wrap your Remix root with `withSentry` to get parameterized router transactions. For Remix applications that do *not* use the [v2 ErrorBoundary](https://remix.run/docs/en/main/route/error-boundary-v2), this will also wrap your root with an `ErrorBoundary` that captures errors in your application. If you use the v2 ErrorBoundary, see the [v2 ErrorBoundary](#v2-errorboundary) section below.
+Also, wrap your Remix root with `withSentry` to get parameterized router transactions. For Remix applications that do _not_ use the [v2 ErrorBoundary](https://remix.run/docs/en/main/route/error-boundary-v2), this will also wrap your root with an `ErrorBoundary` that captures errors in your application. If you use the v2 ErrorBoundary, see the [v2 ErrorBoundary](#v2-errorboundary) section below.
 
 ```typescript {filename: root.tsx}
 import {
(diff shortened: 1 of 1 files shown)


src/platform-includes/getting-started-config/javascript.remix.mdx
```diff
@@ -144,6 +144,46 @@ withSentry(App, {
 });
 ```
 
+## Remix v2 features
+
+_Available from version [...]_
+
+[Remix v2](https://remix.run/docs/en/main/pages/v2) will introduce new features that require additional configuration to work with Sentry. These features are also available from version [1.17.0](https://github.com/remix-run/remix/releases/tag/remix%401.17.0) with [future flags](https://remix.run/docs/en/main/pages/api-development-strategy#current-future-flags). Sentry automatically detects if you are using `v2_errorBoundary` flag and changes its behaviour accordingly.
```
Lms24, line 151: l: Given that we're talking about the error boundary in the subsection below, I'd remove the sentence here.
In general, I'm wondering if this (the detection of the v2 error boundary) is something users need to know about? Or if it causes confusion and they think that they don't need to explicitly call `captureRemixErrorBoundary` as shown below.

```suggestion
[Remix v2](https://remix.run/docs/en/main/pages/v2) will introduce new features that require additional configuration to work with Sentry. These features are also available from version [1.17.0](https://github.com/remix-run/remix/releases/tag/remix%401.17.0) with [future flags](https://remix.run/docs/en/main/pages/api-development-strategy#current-future-flags). 
```

src/platform-includes/getting-started-config/javascript.remix.mdx
```diff
@@ -144,6 +144,46 @@ withSentry(App, {
...
+export function handleError(
+  error: unknown,
+  { request }: DataFunctionArgs
+): void {
+  if (error instanceof Error) {
+    Sentry.captureRemixServerException(error, "remix.server", request);
+  } else {
+    // Optionally capture non-Error objects
+    Sentry.captureException(error);
+  }
```
Lms24, line 183: Not directly relevant for this PR but:
WDYT about making this distinction within the `captureRemixServerException` in the future? I'm not sure what other types of `error`s are passed to this function though. Perhaps we need to filter some of them out if this is too noisy.

src/platform-includes/getting-started-config/javascript.remix.mdx
```diff
@@ -144,6 +144,46 @@ withSentry(App, {
...
+
+  captureRemixErrorBoundaryError(error);
+
+  return <div> ... </div>;
+};
+```
+
+## v2 Server-side Errors
+
+When using `v2_errorBoundary` future flag, Sentry will not automatically capture your server-side errors. Instead, the recommended way is to define a [`handleError`](https://remix.run/docs/en/main/file-conventions/entry.server#handleerror) function in your server entry point. Then, you should use `Sentry.captureRemixServerError` to capture errors in your server-side code.
```
Lms24, line 171: l: Is there another approach than the recommended one? If not, I'd shorten this a little:
```suggestion
When using `v2_errorBoundary` future flag, Sentry will not automatically capture your server-side errors. Instead, define a [`handleError`](https://remix.run/docs/en/main/file-conventions/entry.server#handleerror) function in your server entry point. Then, you should use `Sentry.captureRemixServerError` to capture errors in your server-side code.
```

Lms24 (commented):
Looking good so far! Had a couple of suggestions but once they're addressed I think we're good for now. 

In general, I think it'd be worthwhile to go over this page and reduce the text/code snippets a little. But this is out of scope for this PR. I'll open an issue with some concrete suggestions to follow up on that.

src/platform-includes/getting-started-config/javascript.remix.mdx
onurtemizkan: Good point, yes that's too much detail, and might create confusion.

src/platform-includes/getting-started-config/javascript.remix.mdx
onurtemizkan: My concern was its potential of getting too noisy, as well. Anything thrown in the server also ends up here. Making a distinction inside with an extra configuration parameter sounds good to me actually.

Lms24: Yeah that's Fair. Let's leave it as-is then.

AbhiPrasad (approved):

PR #7334, opened 2023-07-06
fix: Use `___PUBLIC_DSN___` instead of `___DSN___` for all current SDKs

mydea:
Based on a discussion on Slack, we noticed that in a bunch of places we were using `___DSN___`, which includes a private part, instead of `___PUBLIC_DSN___`. I updated places except for legacy SDK places I found.

There were also some places where we've been using `__DSN__` (with two `_`), which is not replaced at all.

For the future, make sure to use the public dsn placeholder.

### DIFF
--- a/src/includes/platforms/configuration/integrations/redux.mdx
+++ b/src/includes/platforms/configuration/integrations/redux.mdx
@@ -11,7 +11,7 @@ const store = createStore(
 
 <Alert level="info" title="Note">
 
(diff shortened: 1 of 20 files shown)


HazAT (approved):

tonyo (approved):

priscilawebdev (approved):

AbhiPrasad (approved):

AbhiPrasad:
We should just lint against using `__DSN__` via https://eslint.org/docs/latest/rules/id-denylist

PR #7348, opened 2023-07-07
feat: Do not show sign in note while loading user

mydea:
* Do not show signed in note until the user has loaded (or not).
* Extract the signed in check out (we'll need it for the org auth tokens as well)
* Make sure the sign in link does not open in a new tab, as the redirect back to docs already works 🎉

### DIFF
--- a/src/components/codeContext.tsx
+++ b/src/components/codeContext.tsx
@@ -1,6 +1,6 @@
 import {createContext, useEffect, useState} from 'react';
 
-type CodeContextStatus = 'LOADING' | 'LOADED';
+type CodeContextStatus = 'loading' | 'loaded';
 
 type ProjectCodeKeywords = {
   API_URL: string;
@@ -180,16 +180,16 @@ export function useCodeContextState(fetcher = fetchCodeKeywords): CodeContextTyp
   const [codeKeywords, setCodeKeywords] = useState(cachedCodeKeywords ?? DEFAULTS);
 
   const [status, setStatus] = useState<CodeContextStatus>(
-    cachedCodeKeywords ? 'LOADED' : 'LOADING'
+    cachedCodeKeywords ? 'loaded' : 'loading'
   );
 
   useEffect(() => {
     if (cachedCodeKeywords === null) {
-      setStatus('LOADING');
+      setStatus('loaded');
       fetcher().then((config: CodeKeywords) => {
         cachedCodeKeywords = config;
         setCodeKeywords(config);
-        setStatus('LOADED');
+        setStatus('loading');
       });
     }
   }, [setStatus, setCodeKeywords, fetcher]);
--- a/src/components/codeContext.tsx
+++ b/src/components/codeContext.tsx
@@ -1,5 +1,7 @@
 import {createContext, useEffect, useState} from 'react';
 
+type CodeContextStatus = 'LOADING' | 'LOADED';
+
 type ProjectCodeKeywords = {
   API_URL: string;
   DSN: string;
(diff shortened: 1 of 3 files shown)


stephanie-anderson (approved):
awesome!!

shanamatthews (approved):
🙌 thank you for fixing this! It was driving me bonkers

src/components/codeContext.tsx
```diff
@@ -1,5 +1,7 @@
 import {createContext, useEffect, useState} from 'react';
 
+type CodeContextStatus = 'LOADING' | 'LOADED';
```
priscilawebdev, line 3: nit: shouldn't this be lowercase?

src/components/signedInCheck.tsx
```diff
@@ -0,0 +1,27 @@
+import React, {Fragment, useContext} from 'react';
+
+import {CodeContext} from './codeContext';
+
+export function SignedInCheck({
+  children,
+  ifIsAuthenticated,
+}: {
+  children: any;
```
priscilawebdev, line 9:
```suggestion
  children: React.ReactNode;
```

src/components/signedInCheck.tsx
```diff
@@ -0,0 +1,27 @@
...
+  ifIsAuthenticated: boolean;
+}): JSX.Element {
+  const {codeKeywords, status} = useContext(CodeContext);
+
+  const user = codeKeywords.USER;
+
+  // Never render until loaded
+  if (status !== 'LOADED') {
+    return null;
+  }
```
priscilawebdev, line 19:
```suggestion
  // Never render until loaded
  if (status !== 'LOADED') {
    return null;
  }
  
   const user = codeKeywords.USER;
```

src/components/signedInCheck.tsx
```diff
@@ -0,0 +1,27 @@
+import React, {Fragment, useContext} from 'react';
+
+import {CodeContext} from './codeContext';
+
+export function SignedInCheck({
+  children,
+  ifIsAuthenticated,
```
priscilawebdev, line 7:
```suggestion
  isUserAuthenticated,
```

🤷‍♀️

src/components/codeContext.tsx
mydea: good question, not sure about the conventions there 🤔 But I guess you're right, I'll make it lowercase! 👍

evanpurkhiser: Why didn't we just make this `isLoading`?

src/components/codeContext.tsx
```diff
@@ -80,6 +82,7 @@ type CodeContextType = {
   codeKeywords: CodeKeywords;
   sharedCodeSelection: any;
   sharedKeywordSelection: any;
+  status: CodeContextStatus;
```
evanpurkhiser, line 85: Caan we have just done `isLoading`?

src/components/signedInCheck.tsx
```diff
@@ -0,0 +1,27 @@
+import React, {Fragment, useContext} from 'react';
+
+import {CodeContext} from './codeContext';
+
+export function SignedInCheck({
```
evanpurkhiser, line 5: This component feels kind of poorly named and under-documented.

Can we add a doc comment to this component to better describe what it does.

src/components/codeContext.tsx
mydea: I initially thought about wanting `error` state or something like this as well 🤔 ended up ignoring this for now. I can adjust this a bit.

src/components/signedInCheck.tsx
mydea: sure thing!

Issue #7226, opened 2023-06-21
Auth Token Creation in Docs


ale-cota:
### Problem ###

We are working on an initiative to improve the workflow for retrieving/managing auth tokens in Sentry. At the moment we’re going for the specific use case needed for uploading source maps, but this will likely be expanded in the future for other scopes as well. You can find the epic [here](https://github.com/getsentry/team-sdks/issues/6) and all the details in [this RFC](https://github.com/getsentry/rfcs/blob/main/text/0091-ci-upload-tokens.md).

One of the very important touch points for this is actually docs. Today we have the source maps upload guides in docs and that’s where people follow instructions and copy code snippets from ([for example](https://docs.sentry.io/platforms/javascript/guides/react/sourcemaps/uploading/webpack/)). The customer is directed to login to Sentry and visit the Auth Token page in order to generate it, sometimes via a call out, sometimes in a code comment.

### Solution ###
In order to remove additional friction in the process of setting up source maps we can maybe expose this new endpoint we’re creating for token generation and actually have a button directly in the docs where a logged-in user could create a new auth token.

We know there’s already other sentry info rendered in code snippets when a user is logged in, like the org slug or DSN, so it should be possible. The novelty would be that we need to have an explicit action there in order to generate the token, we don’t want to create one every time the user navigates to the docs.

PR #7385, opened 2023-07-10
feat: Show org token selector

mydea:
This adds a new org token generator for the source maps setup. For now, we only add it in one place (the webpack plugin configuration), if it works well we can roll it out to all places that use `SENTRY_AUTH_TOKEN`.

Screenshot 2023-07-10 at 15 36 54

If not signed in, we use this instead:

Screenshot 2023-07-10 at 15 34 25

Note that this is blocked by the org tokens being rolled out to all orgs.

Closes https://github.com/getsentry/sentry-docs/issues/7226


src/components/orgAuthTokenNote.tsx
```diff
@@ -0,0 +1,49 @@
...
+  query SignInNoteQuery {
+    site {
+      siteMetadata {
+        siteUrl
+      }
+    }
+  }
+`;
+
+export function OrgAuthTokenNote(): JSX.Element {
```
evanpurkhiser, line 19: There's no need to return `JSX.Element`

src/components/codeBlock.tsx
```diff
@@ -9,50 +9,80 @@ import memoize from 'lodash/memoize';
 
 import {useOnClickOutside} from 'sentry-docs/utils';
 
-import {CodeContext} from './codeContext';
+import {CodeContext, createOrgAuthToken} from './codeContext';
 
 const KEYWORDS_REGEX = /\b___(?:([A-Z_][A-Z0-9_]*)\.)?([A-Z_][A-Z0-9_]*)___\b/g;
 
+const ORG_AUTH_TOKEN_REGEX = /__ORG_AUTH_TOKEN__/g;
```
evanpurkhiser, line 16: This is inconsistently two underscores instead of three like the rest of these keywords

src/components/codeContext.tsx
```diff
@@ -176,6 +176,65 @@ export async function fetchCodeKeywords(): Promise<CodeKeywords> {
...
+function getCookie(cookies: string, cookieName: string) {
+  const csrfCookie = cookies.split(';').find(cookie => {
+    const str = cookie.trim();
+    return str.startsWith(`${cookieName}=`);
+  });
+
+  return csrfCookie ? csrfCookie.split('=')[1] : null;
+}
+
+export async function createOrgAuthToken({
```
evanpurkhiser, line 198: Is there a better place we can put this export? Doesn't feel right to have this inside codeContext.

I know this code is a bit of a mess right now, so it's probably unclear where this should go

src/components/codeBlock.tsx
```diff
@@ -73,6 +103,59 @@ type KeywordSelectorProps = {
...
+
+  const choices = codeKeywords?.PROJECT;
+
+  // When not signed in, we just show a placeholder, as the user can't generate a token in this case
+  if (!codeKeywords.USER) {
+    return <Fragment>sntrys_YOUR_TOKEN_HERE</Fragment>;
+  }
+
+  const currentSelectionIdx = sharedSelection.PROJECT ?? 0;
+  const currentSelection = choices[currentSelectionIdx];
```
evanpurkhiser, line 125: Does it prompt you for which organization to create the token for?

I've been thinking about how we can generally make this experience better, so maybe the context would be a lot more obvious

src/components/codeContext.tsx
```diff
@@ -176,6 +176,65 @@ export async function fetchCodeKeywords(): Promise<CodeKeywords> {
...
+}
+
+function getCookie(cookies: string, cookieName: string) {
+  const csrfCookie = cookies.split(';').find(cookie => {
+    const str = cookie.trim();
+    return str.startsWith(`${cookieName}=`);
+  });
+
+  return csrfCookie ? csrfCookie.split('=')[1] : null;
+}
```
evanpurkhiser, line 196: Let's no use this please. Let's use `js-cookie`. If we don't already have it we should add it

src/components/orgAuthTokenNote.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const location = useLocation();
+
+  return (
+    <SignedInCheck isUserAuthenticated={false}>
+      <StaticQuery
```
evanpurkhiser, line 24: Are we able to `useStaticQuery` instead of this? https://www.gatsbyjs.com/docs/how-to/querying-data/use-static-query/

evanpurkhiser:
Am I missing something, where does it actually show your tokens that you've already generated?

We should add them to this endpoint https://github.com/getsentry/getsentry/blob/933c060c7efe4b90ab12cca434322f2443a4eadc/getsentry/web/docsupport.py

src/components/codeBlock.tsx
mydea: I actually did this on purpose to avoid this being captured by the _other_ regex. If we use triple underscore here, we need to make sure checking order is right (IMHO making this a bit more brittle for future refactorings), and it may also not be too clear that this does not go through the same code 🤔 maybe a _completely_ different identifier would be even better, to make it more clear this is something else? For example `+++ORG_AUTH_TOKEN+++`?

src/components/codeBlock.tsx
mydea: this just takes the org that is selected - in the UI we moved the `__ORG___` above this selector to accommodate this - it would take `example-org` in this case:

Screenshot 2023-07-10 at 15 36 54

src/components/codeContext.tsx
```diff
@@ -176,6 +176,65 @@ export async function fetchCodeKeywords(): Promise<CodeKeywords> {
...
 
+function getCsrfToken(cookies = document.cookie): string | null {
+  // is sentry-sc in production, but may also be sc in other envs
+  // So we just try both variants
+  const cookieNames = ['sentry-sc', 'sc'];
+
+  return cookieNames
+    .map(cookieName => getCookie(cookies, cookieName))
+    .find(token => token !== null);
+}
```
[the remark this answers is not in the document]
mydea: AFAIK this only exists in sentry frontend (where I kind of stole this from), not in docs yet! We have not been doing any POST stuff yet, only GET which does not need this.

src/components/codeContext.tsx
mydea: I can move it into the org auth token note component!

mydea: There isn't really a utils kind of place as far as I've seen so far? Just the `components`, and this doesn't really have a dedicated component backing it 😬 it's only used in the `codeBlock` as of now. Happy to move it to any place you think makes sense!

src/components/orgAuthTokenNote.tsx
mydea: ah, yeah, much nicer!

src/components/signedInCheck.tsx
```diff
@@ -2,17 +2,27 @@ import React, {Fragment, useContext} from 'react';
...
+ */
 export function SignedInCheck({
   children,
   isUserAuthenticated,
 }: {
   children: React.ReactNode;
   isUserAuthenticated: boolean;
 }) {
-  const {codeKeywords, status} = useContext(CodeContext);
+  const {codeKeywords, isLoading} = useContext(CodeContext);
```
mydea, line 22: @evanpurkhiser refactored this based on your feedback from the other PR!

mydea:
> Am I missing something, where does it actually show your tokens that you've already generated?
> 
> We should add them to this endpoint https://github.com/getsentry/getsentry/blob/933c060c7efe4b90ab12cca434322f2443a4eadc/getsentry/web/docsupport.py

Tokens can only be viewed once after they are created, so we always have to create a new token (which is also why we have a button that has to be manually pressed in order to create/show the token). Every time you press the button it will generate a new token!

src/components/codeBlock.tsx
evanpurkhiser: In that case we should write tests.

If we're going to start having custom behavior for different tokens, we should implement it in a way that it's clear that this component supports handling that.

Right now using a different number of underscores is just going to be **confusing**, since it's really hard to notice (I already fixed a number of places where we were using 2 instead of 3).

I think the way to go here is to implement something a little bit more robust

src/components/codeContext.tsx
evanpurkhiser: Leave it here for now, we can move it around later once this repo is a little better organized.

evanpurkhiser:
> Every time you press the button it will generate a new token!

Should we make this clear to the user? Seems a bit annoying that just clicking a button is going to pollute my token list. I'm not able to name it or give it a description here right?

src/components/codeBlock.tsx
```diff
@@ -9,50 +9,80 @@ import memoize from 'lodash/memoize';
...
+      makeOrgAuthTokenClickable(arr, child);
+    } else {
+      arr.push(child);
     }
 
     return arr;
   }, []);
 }
 
+function makeOrgAuthTokenClickable(arr: any[], str: string) {
```
priscilawebdev, line 39: is there another type instead of`any` that we can use here?

src/components/codeBlock.tsx
```diff
@@ -73,6 +103,59 @@ type KeywordSelectorProps = {
...
+      <KeywordDropdown onClick={updateToken}>
+        {tokenState === 'none'
+          ? 'Click to generate token'
+          : tokenState === 'loading'
+          ? 'Generating...'
+          : token
+          ? token
+          : 'Error generating token'}
+      </KeywordDropdown>
+    </Fragment>
```
priscilawebdev, line 155:
```suggestion
    <KeywordDropdown onClick={updateToken}>
        {tokenState === 'none'
          ? 'Click to generate token'
          : tokenState === 'loading'
          ? 'Generating...'
          : token
          ? token
          : 'Error generating token'}
      </KeywordDropdown>
```

src/components/codeContext.tsx
```diff
@@ -191,23 +190,71 @@ export async function fetchCodeKeywords(): Promise<CodeKeywords> {
...
+    if (!resp.ok) {
+      return null;
+    }
+
+    const json = await resp.json();
+
+    return json.token;
+  } catch {
+    return null;
+  }
```
priscilawebdev, line 240: will Sentry catch a bug here, in case it happens?

src/components/orgAuthTokenNote.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const url = data.site.siteMetadata.siteUrl + location.pathname;
+
+  return (
+    <SignedInCheck isUserAuthenticated={false}>
+      <Note>
+        {' '}
+        You can{' '}
+        <ExternalLink href="https://sentry.io/settings/auth-tokens/" target="_blank">
+          manually create an Auth Token
+        </ExternalLink>{' '}
```
priscilawebdev, line 32: doesn't the `ExternalLink` component already open the link in a new tab? asking because I am not familiar with the docs UI/components

priscilawebdev:
Is this PR still a draft?

mydea:
OK, so I added some tests for this for different scenarios! Thanks @priscilawebdev for helping me figure out the react test stuff 😅 But should be good now!

src/components/codeContext.tsx
mydea: No, we just ignore it in this case - not sure if it is worth it to capture this? May lead to some noise, not sure.

src/components/orgAuthTokenNote.tsx
mydea: as far as I can tell, in docs it does not open in a new tab by default..?

mydea:
> Should we make this clear to the user? Seems a bit annoying that just clicking a button is going to pollute my token list. I'm not able to name it or give it a description here right?

We generate a (hopefully explanatory) name for the generated token, so it shouldn't be too confusing, hopefully. Also, since we track which token has been used when & where, it should be easy to clean up unusued tokens from the sentry UI. We can always add a further explanation here later when needed, IMHO. We could also eventually solve this in-product (by providing some auto-cleanup functionality or whatnot).

mdtro:
> > Should we make this clear to the user? Seems a bit annoying that just clicking a button is going to pollute my token list. I'm not able to name it or give it a description here right?
> 
> 
> 
> We generate a (hopefully explanatory) name for the generated token, so it shouldn't be too confusing, hopefully. Also, since we track which token has been used when & where, it should be easy to clean up unusued tokens from the sentry UI. We can always add a further explanation here later when needed, IMHO. We could also eventually solve this in-product (by providing some auto-cleanup functionality or whatnot).

+1 to making this clear in the doc here. The comment line above it now makes it sound like I can go retrieve the token later at the link provided.

mydea:
OK, so I updated it to show this when signed in:

Screenshot 2023-07-18 at 10 22 59

vs. this when not signed in:

Screenshot 2023-07-18 at 10 23 11

PR #7402, opened 2023-07-11
ref(ts): Improve TableOfContents component

evanpurkhiser:


src/components/basePage.tsx
```diff
@@ -55,7 +55,7 @@ export function BasePage({
   const {title, excerpt, description} = pageContext;
   const hasToc = !pageContext.notoc;
 
-  const contentRef = useRef<HTMLDivElement>(null);
+  const [contentElement, setContentElement] = useState<HTMLElement | null>(null);
```
evanpurkhiser, line 58: We use state here since we care that the element reference has been stored before generating the ToC (useState will trigger a re-render). Otherwise we're just relaying on the render order of components to ensure that the ref is set before the ToC is generated.

JoshFerge (approved):
🥳

PR #7440, opened 2023-07-14
fix: Remove grouping breakdown page

scttcper:
Its this page https://docs.sentry.io/product/data-management-settings/event-grouping/grouping-breakdown/

fixes https://github.com/getsentry/sentry/issues/50988


vivianyentran (approved):
Thanks for fixing!

PR #7459, opened 2023-07-18
Add sign in notes to code samples containing placeholders

stephanie-anderson:
We developed a "sign in note", that directs users to the Sentry login page, and redirects them back to that exact page, once successfully registered/logged in.

It is only shown when users aren't logged in to Sentry.io. We've placed it above code sections, that contain the following placeholders:
- `___PUBLIC_DSN___`
- `___PROJECT_ID___`
- `___PROJECT_SLUG___`
- `___ORG_SLUG___`

When users are logged out, this is how the note looks like:

> Screenshot 2023-07-18 at 15 22 35

And when there is no placeholder in the code section, we of course haven't added the note:

> Screenshot 2023-07-18 at 15 22 41

### DIFF
--- a/src/docs/clients/cocoa/advanced.mdx
+++ b/src/docs/clients/cocoa/advanced.mdx
@@ -73,6 +73,8 @@ Client.shared?.enabled = false
 
 You can also `init` the client with options:
 
+<SignInNote />
+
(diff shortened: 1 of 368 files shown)


Lms24 (approved):
🚢 🚢

PR #7464, opened 2023-07-19
docs: add missing getCurrentHub import

rdord:
This PR will fix the missing `getCurrentHub` import in the `src/platform-includes/session-replay/setup/javascript.mdx` docs.

### DIFF
--- a/src/platform-includes/session-replay/setup/javascript.mdx
+++ b/src/platform-includes/session-replay/setup/javascript.mdx
@@ -45,6 +45,6 @@ Sentry.init({
 });
 
 // Sometime later
-const { Replay } = await import("@sentry/browser");
+const { getCurrentHub, Replay } = await import("@sentry/browser");
 getCurrentHub().getClient().addIntegration(new Replay());
 ```


Lms24 (commented):
Hi @rdord thanks for opening this PR! I agree that it might be confusing where to import this function from. 

Is there a specific reason why you'd need to dynamically import `getCurrentHub`? I'm not saying it's wrong, I'm just thinking we might instead want to adjust the snippet to call `Sentry.getCurrentHub` instead as users must have imported `Sentry` (and hence `getCurrentHub`) already anyway. Does that make sense for your use case?

rdord:
> Is there a specific reason why you'd need to dynamically import `getCurrentHub`? 

There is no need to dynamically import it, but I need to import it in the component where I want to dynamically add Replay integration. I add it depending on a feature flag condition.

> instead as users must have imported * as Sentry from "@sentry/<sdk>" (and hence getCurrentHub) already anyway

In my case, this is not available in the component. `Sentry.init()` is in a different file than adding the Replay integration.

PR #7471, opened 2023-07-20
Add org role requirement note

stellaatsentry:
It looks like the missing information caused confusion to a customer, even though they got the same message in the Sentry Web UI:
image

For completeness, it would be better to add it in our documentation too.

### DIFF
--- a/src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
+++ b/src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
@@ -23,6 +23,12 @@ You can configure detector thresholds for slow DB queries issues in **Project Se
 
 ![Slow DB Query detector threshold settings](slow-db-queries-detector-settings.png)
 
+<Note>
+
+These settings can only be edited by users with the organization-level owner, manager, or team-level admin roles.
+
+</Note
+
 ## Span Evidence
 
 You can get additional information about your slow DB query by looking at the following fields in the "Span Evidence":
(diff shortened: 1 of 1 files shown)


src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
```diff
@@ -23,6 +23,12 @@ You can configure detector thresholds for slow DB queries issues in **Project Se
 
 ![Slow DB Query detector threshold settings](slow-db-queries-detector-settings.png)
 
+<Note>
+
+These settings can only be edited by users with the organization-level owner, manager, or team-level admin roles.
```
lizokm, line 28:
```suggestion
These settings can only be edited by users with owner, manager, or team-level admin roles.
```

I think the in-app message is a bit confusing. I assume that most users don't know that we have org-level and team-level roles. Might be better to simplify a bit in both docs and the product UI?

lizokm (approved):
Left a suggestion.

src/docs/product/issues/issue-details/performance-issues/slow-db-queries/index.mdx
```diff
@@ -23,6 +23,12 @@ You can configure detector thresholds for slow DB queries issues in **Project Se
...
+
+</Note
```
shanamatthews, line 30:
```suggestion
</Note>
```

This should fix your build error :)

shanamatthews:
@stellaatsentry - I've fixed your build error here, FYI.

stellaatsentry:
Thank you for the contributions @shanamatthews , @lizokm ! Is this ok to merge?

lizokm:
Yes! Merging now :)

Issue #6677, opened 2023-04-19
Highlight Source Context Feature for Android (NDK)


romtsn:
Check if the upload debug image page mentions the [source context](https://docs.sentry.io/platforms/apple/data-management/debug-files/source-context/), ideally with a screenshot to encourage more users to use it. The [Apple docs](https://docs.sentry.io/platforms/apple/dsym/#sentry-cli) already mention source context, but not with a screenshot.

Point out the feature in the wizard. We still need to align on how to do this, but the minimum should be pointing it out as the Apple wizard does

marandaneto:
@stefanosiano would you like to address this one as well?
The source context info is there but its missing the one for the NDK bits
https://docs.sentry.io/platforms/android/enhance-errors/source-context/
We have to highlight this for NDK, See https://docs.sentry.io/platforms/android/configuration/gradle/ and the `uploadNativeSymbols` flag.

PR #7472, opened 2023-07-20
Add NDK references to source context setup

stefanosiano:
## Pre-merge checklist

## Description of changes
Fixes https://github.com/getsentry/sentry-docs/issues/6677

Added NDK references to source context setup

### DIFF
--- a/src/platform-includes/source-context/java.mdx
+++ b/src/platform-includes/source-context/java.mdx
@@ -198,7 +198,7 @@ sentry {
 
 ### Include NDK Source Context with Gradle Plugin
 
-By default the native code is not uploaded with the Java/Kotlin code. Add the following options to enable Source Context for native code:
+By default native code (C/C++) is not uploaded with the Java/Kotlin code. Add the following options to enable Source Context for native code:
 
 ```groovy
 sentry {
--- a/src/platform-includes/source-context/java.mdx
+++ b/src/platform-includes/source-context/java.mdx
@@ -355,7 +355,7 @@ options.addBundleId("A_VALID_UUID");
 
 <PlatformSection supported={["android"]}>
 
-## Manual Uploading Native Symbols and Source Code
+## Manually Uploading Native Symbols and Source Code
 
 Upload the native symbols manually using the `debug-files upload command`, here's a typical example for Android:
 
--- a/src/platform-includes/source-context/java.mdx
+++ b/src/platform-includes/source-context/java.mdx
@@ -311,18 +311,6 @@ First, create the source bundle containing your source files:
 sentry-cli debug-files bundle-jvm --output path/to/store/bundle --debug-id A_VALID_UUID path/to/source-code
 ```
 
-<PlatformSection supported={["android"]}>
-
-### Creating the Native Source Bundle
-
-If you're using NDK, create the native source bundle containing your source files:
-
-```
-sentry-cli difutil bundle-sources /path/to/files
-```
-
-</PlatformSection>
-
 ### Uploading the Source Bundle
 
 Next, upload that source bundle to Sentry:
@@ -333,12 +321,18 @@ sentry-cli debug-files upload --type jvm path/to/bundle
 
 <PlatformSection supported={["android"]}>
 
-### Uploading including the Native Source Bundle
+### Uploading Native Symbols and Source Code
 
-Include the native source bundle with `--include-sources` and upload it to Sentry:
+Upload the native symbols manually using the `debug-files upload command`, here's a typical example for Android:
 
 ```
-sentry-cli debug-files upload --include-sources /path/to/files --type jvm path/to/bundle
+sentry-cli debug-files upload app/build/intermediates/merged_native_libs/
+```
+
+In order to upload the native source files as well, you can use the `--include-sources` flag. Using this flag sentry-cli will automatically generate source bundles on the fly during the upload.
+
+```
+sentry-cli debug-files upload app/build/intermediates/merged_native_libs/ --include-sources
 ```
 
 </PlatformSection>
--- a/src/platform-includes/source-context/java.mdx
+++ b/src/platform-includes/source-context/java.mdx
@@ -194,6 +194,49 @@ sentry {
 
 </PlatformSection>
 
+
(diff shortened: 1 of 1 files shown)


src/platform-includes/source-context/java.mdx
```diff
@@ -194,6 +194,49 @@ sentry {
 
 </PlatformSection>
 
+
+<PlatformSection supported={["android"]}>
```
stefanosiano, line 198: @romtsn these bits are relevant for Android only, right?

src/platform-includes/source-context/java.mdx
```diff
@@ -277,6 +332,18 @@ Next, upload that source bundle to Sentry:
...
 ```
 
+<PlatformSection supported={["android"]}>
+
+### Uploading including the Native Source Bundle
+
+Include the native source bundle with `--include-sources` and upload it to Sentry:
+
+```
+sentry-cli debug-files upload --include-sources /path/to/files --type jvm path/to/bundle
```
stefanosiano, line 342: hope i got the command right here

src/platform-includes/source-context/java.mdx
```diff
@@ -277,6 +331,18 @@ Next, upload that source bundle to Sentry:
...
+```
+
+</PlatformSection>
```
markushi, line 344: I'm not an expert but I just tried to upload the native symbols + sources for our Android demo app. It works a bit easier for the NDK, there's no need to create a source bundle - just adding `--include-sources` does the trick.

How about the following? (GH suggestions is not working due to code blocks):

### Uploading Native Symbols and Source Code

Upload the native symbols manually using the `debug-files upload` command, here's a typical example for Android:

```
sentry-cli debug-files upload app/build/intermediates/merged_native_libs/
```

In order to upload the native source files as well, you can use the `--include-sources` flag. Using this flag sentry-cli will automatically generate source bundles on the fly during the upload.

```
sentry-cli debug-files upload app/build/intermediates/merged_native_libs/ --include-sources
```

src/platform-includes/source-context/java.mdx
```diff
@@ -269,6 +311,18 @@ First, create the source bundle containing your source files:
 sentry-cli debug-files bundle-jvm --output path/to/store/bundle --debug-id A_VALID_UUID path/to/source-code
 ```
 
+<PlatformSection supported={["android"]}>
```
markushi, line 314: I think we could skip the section if we use the `--include-sources` flag I mentioned below.

src/platform-includes/source-context/java.mdx
markushi: I don't think it works this way, at least the Android Gradle Plugin performs two uploads: One for JVM and one for Native sources / symbols - I've added a comment below how it works.

markushi (changes requested):
Looking good, please have a look at my comments

src/platform-includes/source-context/java.mdx
romtsn: right, we don't need this one

src/platform-includes/source-context/java.mdx
romtsn: that's how it works yes 👍 sentry-cli will walk the file tree on itself and search for relevant sources when `--include-sources` is supplied.

romtsn:
@stefanosiano  I still think the current structure might be confusing - there's a bunch of unrelated to NDK subsections in "Manual Upload". Should we just have two h1 headings `Source Context for Java/Kotlin` and `Source Context for NDK` so they are better separated?
image

stefanosiano:
> @stefanosiano I still think the current structure might be confusing - there's a bunch of unrelated to NDK subsections in "Manual Upload". Should we just have two h1 headings `Source Context for Java/Kotlin` and `Source Context for NDK` so they are better separated? image
> 
> E.g. UUIDs, Know Limitations, Configuring the SDK - all of that is not needed for native sources

Now there are only 2 sections related to NDK, one for gradle plugin and one for manual upload.
I get everything else is not needed for the NDK, but i think someone uploading native source wants to upload java source first.
I mean NDK is more something to upload after the other things are configured.
I can easily make a separated section for the NDK and put it at the end, if you think it's less confusing. 
My only fear is that moving to a separate section means moving both manual and gradle plugin info (only the part related to ndk) to the bottom. And if i move the gradle plugin, either i have to duplicate the whole configuration (org, project, etc.) or the user would have to go up to fetch the full configuration and back down again to continue for the ndk

markushi (approved):
LGTM!

romtsn:
@stefanosiano I mean this looks like you have to still configure the SDK after uploading NDK source - but this is not the case, the `Configuring the SDK` section is only relevant for manual upload of Java/Kotlin sources 
image

stefanosiano:
@romtsn I moved the NDK manual upload to the end as a h2 section, separated from the previous one. Having a h1 for that seems too much imho

src/platform-includes/source-context/java.mdx
```diff
@@ -310,3 +352,21 @@ sentry.bundle-ids=A_VALID_UUID`
 ```Java
 options.addBundleId("A_VALID_UUID");
 ```
+
+<PlatformSection supported={["android"]}>
+
+## Manually Uploading Native Symbols and Source Code
```
marandaneto, line 358:
```suggestion
## Manually Uploading Debug Symbols and Source Code
```

marandaneto: I'd also change every place you mention `native symbols` to `debug symbols` in this page.

src/platform-includes/source-context/java.mdx
```diff
@@ -310,3 +352,21 @@ sentry.bundle-ids=A_VALID_UUID`
...
+
+Upload the native symbols manually using the `debug-files upload command`, here's a typical example for Android:
+
+```
+sentry-cli debug-files upload app/build/intermediates/merged_native_libs/
+```
```
marandaneto, line 364: I'd rather skip this to make it simpler and go directly with the example below.
You can adapt the wording if necessary.

stefanosiano: what about now?

PR #7477, opened 2023-07-21
.NET housekeeping

bitsandfoxes:
Updated snippets and obsolete API.


antonpirker (approved):
looks good to me. (i do not know much about .net but overall I did not see stuff that obviously wrong)

PR #7547, opened 2023-07-27
feat(replay): Add example for `beforeAddRecordingEvent`

billyvg:


src/platforms/javascript/common/session-replay/privacy.mdx
```diff
@@ -69,6 +69,31 @@ new Sentry.Replay({
...
+    ) {
+      return event;
+    }
+
+    // Only capture requests that result in a 500 status code
+    if (event.data.payload.data.statusCode !== 500) {
+      return null;
+    }
+
+    return event;
```
ryan953, line 92: IMHO The negation in the first stanza, and duplicate `return event` statements, make this harder to understand if someone's not as familiar with the types.

what about this:

```suggestion
    // Do not capture fetch/xhr requests, unless the response code is 500
    if (
      event.data.tag === "performanceSpan" &&
      (event.data.payload.op === "resource.fetch" || event.data.payload.op === 'resource.xhr') &&
      event.data.payload.data.statusCode !== 500
    ) {
      return null;
    }
    
    return event;
```

billyvg: Yeah this makes more sense.

src/platforms/javascript/common/session-replay/privacy.mdx
```diff
@@ -69,6 +69,28 @@ new Sentry.Replay({
 });
 ```
 
+### Example: Capturing 500 Status Codes Only
+
+Here is an example that will only capture fetch requests that return a 500 status code. Non-fetch requests will continue to be captured as normal.
```
lizokm, line 74:
```suggestion
Here's an example showing how to only capture fetch requests that return a 500 status code. (Non-fetch requests would continue to be captured normally.)
```

src/platforms/javascript/common/session-replay/privacy.mdx
```diff
@@ -69,6 +69,28 @@ new Sentry.Replay({
...
+    ) {
+      return null;
+    }
+
+    return event;
+  },
+});
+```
+
 Note that we also have [server-side PII scrubbing](/product/data-management-settings/scrubbing/server-side-scrubbing/) for this data. It looks for certain patterns such as American social security numbers, credit cards, and private keys.
```
lizokm, line 94:
```suggestion
We also have [server-side PII scrubbing](/product/data-management-settings/scrubbing/server-side-scrubbing/) for this data. It looks for certain patterns such as American social security numbers, credit cards, and private keys.
```
Would be good to get rid of "Note" here since there's another sentence that starts with "Note" right below :)

lizokm (approved):
LGTM

PR #7576, opened 2023-08-01
Add wrong Kotlin inline function number to Android Troubleshooting

stefanosiano:
## Pre-merge checklist

## Description of changes

added a `Kotlin` section to Android Troubleshooting regarding kotlin inline function reported with wrong line numbers
Relates to https://github.com/getsentry/sentry-java/issues/2763

### DIFF
--- a/src/platforms/android/troubleshooting.mdx
+++ b/src/platforms/android/troubleshooting.mdx
@@ -73,6 +73,13 @@ For example, the following stack traces:
   at io.sentry.HubAdapter.captureEvent (HubAdapter.java:29)
 ```
 
+## Kotlin
+
+Stack trace elements referring to Kotlin inline functions from files different than the call site are incorrect, due to how they are compiled.
+A .class file with inline functions contains code from many source files. When the JVM generates a stacktrace, it will contain a different line number, because it doesn't substitute the class and line in which the error occurred with the original file and line number.
+As a result, if an exception is thrown inside an inline function, the line number of the stack trace of the Exception could be wrong.
+This is a missing feature in the JVM itself, so "Sentry can't fix this".
+
 ## NDK
 
 If you're using a version of `sentry-android` before 5.0, due to current limitations in how `sentry-native` works, it cannot identify native libraries loaded directly from `.apk` files. Instead, it needs to have access to the extracted libraries on the file system.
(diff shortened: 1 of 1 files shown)


src/platforms/android/troubleshooting.mdx
```diff
@@ -73,6 +73,13 @@ For example, the following stack traces:
   at io.sentry.HubAdapter.captureEvent (HubAdapter.java:29)
 ```
 
+## Kotlin
+
+Stack trace elements referring to Kotlin inline functions from files different than the call site are incorrect, due to how they are compiled.
+A .class file with inline functions contains code from many source files. When the JVM generates a stacktrace, it will contain a different line number, because it doesn't substitute the class and line in which the error occurred with the original file and line number.
+As a result, if an exception is thrown inside an inline function, the line number of the stack trace of the Exception could be wrong.
+This is a missing feature in the JVM itself, so "Sentry can't fix this".
```
shanamatthews, line 81:
```suggestion

Stack trace elements referring to Kotlin inline functions from files different than the call site may refer to the wrong line numbers.

This is due to a missing feature in the JVM. `.class` files with inline functions contain code from many source files, and when the JVM generates a stacktrace, it doesn't replace the class and line where the error occurred with the original file and line number. So, when an exception is thrown in an inline function, the line number for the stack trace of the Exception may be wrong.
```

shanamatthews: > Stack trace elements referring to Kotlin inline functions from files different than the call site may refer to the wrong line numbers.

Is this correct? I wanted the first sentence to tell users specifically what is wrong.

shanamatthews (approved):

markushi (approved):

romtsn (approved):
looks good, but I'd also add:
* A link to the issue on YouTrack where it's clear that it's a missing feature on the JVM/Kotlin side
* An example with an inline function and the corresponding stacktrace, because it's a bit hard to grasp what is this about without a clear example

src/platforms/android/troubleshooting.mdx
```diff
@@ -73,6 +73,13 @@ For example, the following stack traces:
   at io.sentry.HubAdapter.captureEvent (HubAdapter.java:29)
 ```
 
+## Kotlin
```
romtsn, line 76:
```suggestion
## Incorrect Line Numbers in Kotlin
```

src/platforms/android/troubleshooting.mdx
stefanosiano: yep, thanks!

PR #7578, opened 2023-08-01
fix(rn): RN Profiling is available from version 5.8.0

krystofwoldrich:
fix rn sdk version

https://github.com/getsentry/sentry-react-native/releases/tag/5.8.0


vivianyentran (approved):
thanks for keeping the docs updated!

PR #7581, opened 2023-08-01
ref: Add that auto-assignment is turned off after manual assignment

jangjodi:
Add that auto assignment of an issue is turned off after manual assignment

closes https://github.com/getsentry/sentry/issues/51612


vivianyentran (approved):
Thanks for updating!

PR #7599, opened 2023-08-07
feat(.NET): Added `EventProcessor`

bitsandfoxes:
Resolves https://github.com/getsentry/sentry-dotnet/issues/2534
Previously undocumented feature of the .NET SDK.

### DIFF
--- a/src/platform-includes/enriching-events/event-processors/dotnet.mdx
+++ b/src/platform-includes/enriching-events/event-processors/dotnet.mdx
@@ -15,20 +15,19 @@ public class CustomEventProcessor : ISentryEventProcessor
 }
 ```
 
-There are multie ways of adding the event processors. Processors run on every event sent after they were added.
-Adding it to the options ensures the processor runs with every event after initialization
+Processors run on every event sent once they've been added. While there are multiple ways of doing this, adding processors to the options ensures that they run with every event after initialization.
 
 ```csharp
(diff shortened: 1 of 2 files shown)


src/platform-includes/enriching-events/event-processors/dotnet.mdx
```diff
@@ -0,0 +1,35 @@
...
+    {
+        // Add anything to the event here
+        // returning `null` will drop the event
+        return @event;
+    }
+}
+```
+
+There are multie ways of adding the event processors. Processors run on every event sent after they were added.
+Adding it to the options ensures the processor runs with every event after initialization
```
lizokm, line 19:
```suggestion
Processors run on every event sent once they've been added. While there are multiple ways of doing this, adding processors to the options ensures that they run with every event after initialization.
```

src/platform-includes/enriching-events/event-processors/dotnet.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+```csharp
+options.AddEventProcessor(new CustomEventProcessor());
+```
+
+After adding it to the scope the processor applies to the current and following scopes
```
lizokm, line 25:
```suggestion
If `AddEventProcessor` has been added to the scope, it'll apply to both the current and the following scopes:
```

lizokm (approved):
Made some language suggestions.

PR #7617, opened 2023-08-10
Add electron as performance at scale supported SDK

AbhiPrasad:
https://github.com/getsentry/sentry-electron/pull/710 fixes a bug with baggage propagation that existed for a while, so making `4.9.0` the min version required.


lforst (approved):
seems risky bro

AbhiPrasad:
> seems risky bro

PR #7621, opened 2023-08-11
chore(relay): Remove TLS mentions from getting started guide

olksdr:
Relay does not support TLS configuration anymore and rather required reverse proxy to do the SSL offloading. 
Let's remove it from the docs altogether.


iker-barriocanal (approved):
Nice, thanks for thinking about this!

vivianyentran (approved):
Reran the vercel deployment. Looks like it all works now!

PR #7646, opened 2023-08-15
fix(releases): Add note about repo name when associating commits

lobsterkatie:
A customer recently ran into trouble because they were trying to associate commits using the form `some-repo@123abc` rather than the form `some-org/some-repo@123abc`. Though we do have a note about this in the docs, in the section on releases, it's easy to miss. We also don't mention it at all in the CLI docs.

This fixes that by pulling it out into a highlighted note in both places.


vivianyentran (approved):
Thanks for fixing  this!

PR #7648, opened 2023-08-16
feat: Rework Browser JS integration docs

mydea:
This PR rewrites & reorganizes the integration docs for Browser JS SDKs.

Now, there is a single `Integrations` page that lists all integrations available for this SDK, with a short description & a note if it is enabled by default.

Each integration has it's own detail page with eventually available options.

Screenshot 2023-08-16 at 12 32 42

The one special case is Electron, which does not have the same default integrations. I found a way to hide the parent pages for Electron, maybe a bit hacky (??) but it seems to work fine... 🤷 

Screenshot 2023-08-16 at 12 42 08

Another change I had to make to make this work/make sense is to move React/Vue Router & Redux pages, which have been inside of Integrations, into the React/Vue Features space. IMHO that makes more sense anyhow, as it is not really integrations this is about...?

Note that this _does not_ touch Node.js yet, we can do that in a follow up.


mydea:
Note: We can/should also improve this page a bit, maybe: https://sentry-docs-git-fn-js-integrations.sentry.dev/platforms/javascript/configuration/integrations/custom/

AbhiPrasad (approved):
I like this change!! The individual pages make a lot more sense to me, and hopefully is easier to maintain as well

src/platforms/javascript/guides/electron/configuration/integrations/childprocess.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+interface ChildProcessOptions {
+  /** Child process events that generate Sentry breadcrumbs */
+  breadcrumbs: ExitReason[];
+  /** Child process events that generate Sentry events */
+  events: ExitReason[];
+}
+```
+
+For example, to disable capture of breadcrumbs and only capture events for Out-Of-Memory crashes:
```
lizokm, line 28:
```suggestion
To disable the ability to capture breadcrumbs and only capture events for Out-Of-Memory crashes, for example:
```

src/platforms/javascript/guides/electron/configuration/integrations/childprocess.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+<SignInNote />
```
lizokm, line 30:
```suggestion
<SignInNote />
```
LInk goes to "page not found"

src/platforms/javascript/guides/electron/configuration/integrations/electronbreadcrumbs.mdx
```diff
@@ -0,0 +1,46 @@
+---
+title: ElectronBreadcrumbs
+excerpt: ""
+description: "Supports capturing events from `uncaughtException` while retaining Electrons default behaviour. (default)"
+sidebar_order: 10
+---
+
+Captures breadcrumbs for events for many of Electrons built-in modules. The breadcrumbs captured for each module can be configured `true` to capture all events or `false` to capture no events. Alternatively you can supply a function which is passed the event name and returns `true` or `false` depending on whether the event should be captured.
+
```
lizokm, line 9:
```suggestion
Captures breadcrumbs for events for many of Electron's built-in modules. The breadcrumbs captured for each module can be configured `true` to capture all events or `false` to capture no events. Alternatively, you can supply a function which is passed the event name and returns `true` or `false`, depending on whether the event should be captured.

```

src/platforms/javascript/guides/electron/configuration/integrations/electronbreadcrumbs.mdx
```diff
@@ -0,0 +1,46 @@
...
+The defaults for this integration are effectively:
+
+<SignInNote />
```
lizokm, line 12:
```suggestion
<SignInNote />
```
Link goes to "Page not found"

src/platforms/javascript/guides/electron/configuration/integrations/mainprocesssession.mdx
```diff
@@ -0,0 +1,32 @@
+---
+title: MainProcessSession
+excerpt: ""
+description: "Captures sessions linked to the lifetime of the Electron main process. (default)"
+sidebar_order: 10
+---
+
+Captures sessions linked to the lifetime of the Electron main process. These sessions result in [release health](/product/releases/health/) statistics being displayed in Sentry.
```
lizokm, line 8:
```suggestion
Captures sessions linked to the lifetime of the Electron main process. Capturing sessions makes it possible to show [release health](/product/releases/health/) statistics in Sentry.
```

lizokm (approved):
Still working through this :)

src/platforms/common/data-management/sensitive-data/index.mdx
```diff
@@ -99,6 +99,6 @@ As a best practice you should always avoid logging confidential information. If
 
 - Anonymize the confidential information within the log statements (for example, swap out email addresses -> for internal identifiers)
 - Use <PlatformIdentifier name="before-breadcrumb" /> to filter it out from breadcrumbs before it is attached
-- Disable logging breadcrumb integration (for example, as described [here](/platforms/javascript/configuration/integrations/default/#breadcrumbs))
+- Disable logging breadcrumb integration (for example, as described [here](/platforms/javascript/configuration/integrations/breadcrumbs/))
```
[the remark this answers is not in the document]
mydea: well then you disable _all_ breadcrumbs, you're gonna want to do this:

```js
integrations: [
  new Sentry.Integrations.Breadcrumbs({
    console: false
  })
]
```

to only disable _console_ breadcrumbs. But based on this, let's add a sentence to the breadcrumbs integration page explicitly pointing out that the options below are meant to disable specific breadcrumbs, and why you may want to do this 👍

src/platforms/javascript/common/configuration/integrations/dedupe.mdx
```diff
@@ -0,0 +1,12 @@
...
+excerpt: ""
+description: "Deduplicate certain events to avoid receiving duplicate errors. (default)"
+sidebar_order: 1
+---
+
+_Import name: `Sentry.Integrations.Dedupe`_
+
+This integration deduplicates certain events. It can be helpful if you're receiving many duplicate errors. Note, that Sentry only compares stack traces and fingerprints.
+
+<PlatformContent includePath="configuration/dedupe" />
```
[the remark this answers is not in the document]
mydea: yeah that doesn't work in the preview! sentry always redirects back to docs.sentry.io and there the page does not exist 😅 (same for all of these)

src/includes/platforms/configuration/options/ignore-transactions.mdx
```diff
@@ -0,0 +1 @@
+A list of strings or regex patterns that match transaction names that shouldn't be sent to Sentry. Transactions that match these strings or regular expressions will be filtered out before they're sent to Sentry. When using strings, partial matches will be filtered out, so if you need to filter by exact match, use regex patterns instead. By default, all transactions are sent.
```
Lms24, line 1: This is (unfortunately) not true anymore. The [default for `ignoreTransactions`](https://github.com/getsentry/sentry-javascript/blob/137c5e912acb6bfcf9893ed114a3448ea68e28a1/packages/core/src/integrations/inboundfilters.ts#L8-L17) was changed to ignore health check transactions on the SDK side (in addition to the Relay filter). Afaik this only applies to JS though. 
 
```suggestion
A list of strings or regex patterns that match transaction names that shouldn't be sent to Sentry. Transactions that match these strings or regular expressions will be filtered out before they're sent to Sentry. When using strings, partial matches will be filtered out, so if you need to filter by exact match, use regex patterns instead. By default, transactions spanning typical API health check requests are filtered out.
```

src/platforms/javascript/common/configuration/integrations/trycatch.mdx
Lms24, line 1: Random rant/No action required:
I never liked the name of this integration. We should think about renaming it to something that resembles _what_ it does, not _how_ it does things...

Lms24 (approved):
I like this change! Makes the structure much more clear and easier to navigate between individual integrations! I have a few high-level comments. Feel free to apply/disregard as you see fit. We can also tackle them as a follow-up.

* m: I'd like to see on first glance which integrations are enabled by default and which need to be added by users. Maybe we can put this directly under the title of the individual pages? Given that we're no longer distinguishing between default and pluggable integrations I think this is important. wdyt? 
* l: Great that we now have a dedicated Options section! More a cosmetic thing than anything but can we give the individual options a heading level? This way we can link to specific options and I guess it'd look a little better than the bullet points
* l: Totally optional but wdyt about adding `BrowserTracing` and `Replay` to the list? We shouldn't duplicate information there but just link to the dedicated performance/replay docs.

src/platforms/javascript/common/configuration/integrations/trycatch.mdx
mydea: good point - should we create an issue for v8?

src/platforms/javascript/guides/electron/configuration/integrations/childprocess.mdx
mydea:
```suggestion
For example, to disable capture of breadcrumbs and only capture events for Out-Of-Memory crashes, you can configure the integration as follows:
```

maybe that is better?

src/platforms/javascript/common/configuration/integrations/trycatch.mdx
Lms24: Yes, I'll take care of it. We did something [similar](https://github.com/getsentry/sentry-javascript/pull/5027) for v7 so this should be fine.

mydea:
OK, I updated this based on feedback by @Lms24:

1. Added a note at the top of each default integration stating that it is default, + linking to the section telling you how to customize default integrations.
2. Refactored all options to be headings, to make it easier to link there etc.
3. Added a page for `Replay` and `BrowserTracing` integrations, which bascially link to performance/replay setup for more details.
4. Added the `ContextLines` integration which I forgot to add before 😅 
5. I actually figured out there is a much nicer way to filter out integrations for electron, we can just provide `notSupported` to the page and it will be gone 🤦 
6. Improved the Custom Integrations page a bit with an actual example for a custom integration

src/platforms/javascript/common/configuration/integrations/browsertracing.mdx
```diff
@@ -0,0 +1,14 @@
+---
+title: BrowserTracing
+excerpt: ""
+description: "Capture performance data for the Browser."
+sidebar_order: 99
+notSupported:
+  - javascript.cordova
```
mydea, line 7: Hmm, not sure why cordova does not have a performance section? 🤔

src/platforms/javascript/common/configuration/integrations/replay.mdx
```diff
@@ -0,0 +1,13 @@
+---
+title: Replay
+excerpt: ""
+description: "Capture a video-like reproduction of what was happening in the user's browser."
+sidebar_order: 99
+notSupported:
+  - javascript.cordova
```
mydea, line 7: Not sure if there is a reason cordova does not have a replay section?

src/platforms/javascript/common/configuration/integrations/trycatch.mdx
Lms24: https://github.com/getsentry/sentry-javascript/issues/8838

Lms24 (approved):
Thanks for applying my feedback!

src/platforms/javascript/guides/electron/configuration/integrations/net.mdx
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * Whether to add 'sentry-trace' headers to outgoing requests
+   *
+   * Defaults to: true
+   */
+  tracingOrigins: boolean | (method: string, url: string) => boolean;
+}
+```
+
+For example, to disable breadcrumb capture and addition of `sentry-trace` headers and only capture spans for `POST` requests:
```
lizokm, line 35:
```suggestion
Here's a code example of how to disable breadcrumb capture and add `sentry-trace` headers to only capture spans for `POST` requests:
```

src/platforms/javascript/guides/electron/configuration/integrations/onuncaughtexception.mdx
```diff
@@ -0,0 +1,14 @@
+---
+title: OnUncaughtException
+excerpt: ""
+description: "Supports capturing events from `uncaughtException` while retaining Electrons default behaviour. (default)"
+sidebar_order: 10
+---
+
+Supports capturing events from `uncaughtException` while retaining Electrons default behaviour of displaying an error dialog if there are no other listeners.
```
lizokm, line 8:
```suggestion
Supports capturing events from `uncaughtException` while retaining Electron's default behavior of displaying an error dialog if there are no other listeners.
```

src/platforms/javascript/guides/electron/configuration/integrations/onuncaughtexception.mdx
```diff
@@ -0,0 +1,14 @@
...
+
+To disable Electrons default error dialog, simply add another listener:
```
lizokm, line 10:
```suggestion
To disable Electron's default error dialog, add another listener:
```

src/platforms/javascript/guides/electron/configuration/integrations/preloadinjection.mdx
```diff
@@ -0,0 +1,23 @@
+---
+title: PreloadInjection
+excerpt: ""
+description: "Injects a preload script via the Electron. (default)"
+sidebar_order: 10
+---
+
+Injects a preload script via the Electron [`session.setPreloads(preloads)`](https://www.electronjs.org/docs/latest/api/session#sessetpreloadspreloads) API. By default sessions are only injected into the `defaultSession`. If you are using other sessions, you can pass a function as `getSessions` to `init`.
```
lizokm, line 8:
```suggestion
Injects a preload script via the Electron [`session.setPreloads(preloads)`](https://www.electronjs.org/docs/latest/api/session#sessetpreloadspreloads) API. By default, sessions are only injected into the `defaultSession`. If you're using other sessions, you can pass a function as `getSessions` to `init`.
```

src/platforms/react-native/troubleshooting.mdx
```diff
@@ -96,7 +96,7 @@ Common fixes to this issue include:
 - Checking that source maps are correctly uploaded with the exact same `release` and `dist` values as the event. They **must match** for source maps to upload correctly.
 - If you set a custom `release` and `dist`, confirming that you manually uploaded the sourcemaps since our automatic source map upload script will not detect the custom values.
 - Make sure that the name of the bundle in the stack trace matches exactly with the name in the source map. For example, if your stack frames say `app:///index.bundle`, the artifacts must be named `index.bundle` and `index.bundle.map` respectively.
-- Checking that you are not disabling the [`RewriteFrames`](/platforms/javascript/configuration/integrations/plugin/#rewriteframes) default integration by overwriting it, filtering it out, or setting `defaultIntegrations: false`.
+- Checking that you are not disabling the [`RewriteFrames`](/platforms/javascript/configuration/integrations/rewriteframes) default integration by overwriting it, filtering it out, or setting `defaultIntegrations: false`.
```
lizokm, line 99:
```suggestion
- Checking that you're not disabling the [`RewriteFrames`](/platforms/javascript/configuration/integrations/rewriteframes) default integration by overwriting it, filtering it out, or setting `defaultIntegrations: false`.
```

lizokm (approved):
Phew!! Thanks for this amazing, labor-intensive PR!

shanamatthews (approved):
Looks good, just a few questions from me here

src/utils.ts
```diff
@@ -58,8 +58,12 @@ export const sortPages = (arr: any, extractor: (any) => Page = n => n): any[] =>
   return arr.sort((a, b) => {
     a = extractor(a);
     b = extractor(b);
-    const aso = a.context.sidebar_order >= 0 ? a.context.sidebar_order : 10;
-    const bso = b.context.sidebar_order >= 0 ? b.context.sidebar_order : 10;
+
+    const aBase = a.context.sidebar_order ?? 10;
+    const bBase = b.context.sidebar_order ?? 10;
+    const aso = aBase >= 0 ? aBase : 10;
+    const bso = bBase >= 0 ? bBase : 10;
```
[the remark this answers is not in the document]
mydea: Ah, forgot to mention this: I noticed that this was actually buggy before, because `sidebar_order` was sometimes `null` which lead to it not getting the default of 10, but null, leading to incorrect ordering etc.

src/platforms/javascript/common/configuration/integrations/index.mdx
```diff
@@ -2,6 +2,65 @@
...
+  - /platforms/javascript/configuration/integrations/plugin/
+  - /platforms/javascript/configuration/integrations/default/
 ---
 
-<PageGrid />
+Sentry provides a number of integrations which are designed to add automatic instrumentation to your application. Some of these integrations are enabled by default, while others have to be explicitly enabled.
+
+## Available Integrations
+
+<PageGrid exclude={["custom"]} />
```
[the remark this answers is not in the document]
mydea: 👍 sure, I can move it down!

src/platforms/javascript/guides/electron/configuration/integrations/maincontext.mdx
```diff
@@ -0,0 +1,10 @@
+---
+title: MainContext
+excerpt: ""
+description: "Adds app, operating system and runtime context to all events. (default)"
+sidebar_order: 10
+redirect_from:
+  - /platforms/javascript/guides/electron/configuration/integrations/default/
```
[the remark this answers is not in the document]
mydea: Good point, will replace this with a proper redirect for all!

PR #7665, opened 2023-08-17
Add browser profiling to profiling getting started

AbhiPrasad:
Noticed this when clicking docs link in https://blog.sentry.io/profiling-to-speed-up-your-apps/ but didn't see browser profiling 😢.

I wonder if we should add custom instructions for Next.js (setting it up for both server and client)

### DIFF
--- a/src/docs/product/profiling/getting-started.mdx
+++ b/src/docs/product/profiling/getting-started.mdx
@@ -17,7 +17,7 @@ Profiling depends on Sentry's performance monitoring product being enabled befor
 
 </Alert>
 
-- [Browser JavaScript](/platforms/javascript/profiling/)
+- [Browser JavaScript](/platforms/javascript/profiling/) [experimental]
 - Mobile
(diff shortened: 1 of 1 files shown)


JonasBa:
@AbhiPrasad thank you! Yes, I have been somewhat intentionally quiet on this as I'm trying to gather feedback from folks using it before I add it to all of our docs.

src/docs/product/profiling/getting-started.mdx
```diff
@@ -17,6 +17,7 @@ Profiling depends on Sentry's performance monitoring product being enabled befor
 
 </Alert>
 
+- [Browser JavaScript](/platforms/javascript/profiling/)
```
AbhiPrasad, line 20:
```suggestion
- [Browser JavaScript](/platforms/javascript/profiling/) [experimental]
```

AbhiPrasad:
Updated so it now says `[experimental]`!

Issue #7670, opened 2023-08-18
Unclear if enabling profiling via sentry-react-native will enable profiling at the native layer too


liamjones:
### Core or SDK?

Platform/SDK

### Which part? Which one?

React Native / Profiling

### Description

https://docs.sentry.io/platforms/react-native/profiling/ explains how to enable profiling for RN. It would be good to clarify on the page if doing so will also enable profiling for native Android/iOS code in the RN project or if you need to do more setup to also profile those layers.

### Suggested Solution

State if enabling profiling via the RN `Sentry.init()` will also apply to the underlying native implementations of Sentry on each platform.

PR #7680, opened 2023-08-23
fix: Clarify which layers are profiled in RN

kahest:
fixes https://github.com/getsentry/sentry-docs/issues/7670

### DIFF
--- a/src/platforms/common/profiling/index.mdx
+++ b/src/platforms/common/profiling/index.mdx
@@ -290,7 +290,7 @@ sentry_sdk.init(
 
 <Note>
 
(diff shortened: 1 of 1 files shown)


src/platforms/common/profiling/index.mdx
```diff
@@ -290,11 +290,11 @@ sentry_sdk.init(
 
 <Note>
 
-Profiling for React Native is available in alpha in SDK versions `5.8.0` and above.
+Profiling for React Native is available in alpha in SDK versions `5.8.0` and above. Currently, this supports profiling of the JavaScript layer. Combined profiling with native Android/iOS layers is not supported at this time.
```
shanamatthews, line 293:
```suggestion
Profiling for React Native is available in alpha in SDK versions `5.8.0` and above. Currently, it supports profiling the JavaScript layer. Combined profiling with native Android/iOS layers isn't supported at this time.
```

shanamatthews (approved):

PR #7686, opened 2023-08-24
design: tweak sidebar font color to have higher contrast ratio

JoshFerge:
Screenshot 2023-08-23 at 5 27 19 PM
Screenshot 2023-08-23 at 5 27 27 PM

Before and after. Our current sidebar text-to-background contrast ratio is 3.54, while 4.5 is the recommended ratio for text. This should make the sidebar text a bit easier to read https://www.w3.org/TR/WCAG20-TECHS/G18.html

lmk thoughts on if color should be tweaked.

part of replays-accessibility hackweek project

### DIFF
--- a/src/css/_includes/sidebar.scss
+++ b/src/css/_includes/sidebar.scss
@@ -75,7 +75,7 @@
   position: relative;
   line-height: 1.5;
   border-radius: 0.125em;
-  color: $desatPurple6;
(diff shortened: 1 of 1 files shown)


shanamatthews (approved):
Awesome, this looks great Josh!

lizokm (approved):
Yes please!

PR #7687, opened 2023-08-24
Update opentelemetry with diagram explaining

smeubank:
Might be useful to add a visualization showing what this is and is not, not a collector approach

[Preview link](https://sentry-docs-git-smeubank-otel-sentry-diagram.sentry.dev/platforms/node/performance/instrumentation/opentelemetry/#opentelemetry-and-sentry)


shanamatthews (commented):
I like this!

PR #7698, opened 2023-08-25
[sentry-cli] use debug-files instead of difutil

supervacuus:
While `difutil` still exists as an alias, `debug-files` seems to be the preferred sub-command name for uploading debug-files. Now, the docs inconsistently present both, which confuses users. This PR replaces all uses with the new preference.

CC @Swatinem, @bitsandfoxes, @markushi


kamilogorek (approved):

shanamatthews (approved):
LGTM! I'll leave this up until Monday in case there's additional feedback and then merge it.

bitsandfoxes (approved):
Thanks for catching and updating that!

PR #7707, opened 2023-08-29
feat(js): Document new performance APIs

AbhiPrasad:
resolves https://github.com/getsentry/sentry-javascript/issues/8724

## Motivation

Documents the new performance APIs under the custom instrumentation section.

In the future we'll want to revamp both the automatic and custom instrumentation sections, but for now this is a good starting point.

The APIs are detailed in the RFC about the [new performance API](https://github.com/getsentry/rfcs/blob/main/text/0101-revamping-the-sdk-performance-api.md), and were introduced in https://github.com/getsentry/sentry-javascript/pull/8803

## Details

`Sentry.startActiveSpan` automatically wraps a callback with a span and makes that span the active span for the execution context of the callback. The callback can be async or sync, and can return arbitrary values.

`Sentry.startSpan` just creates a span, but does not wrap it in a callback. It needs to be explicitly set on the scope just like `Sentry.startTransaction`.

```ts
const result = Sentry.startActiveSpan({ name: importantThing }, async () => {
  await someWork();
  return Sentry.startActiveSpan({ op: 'db', name: 'query-stuff' }, () => expensiveQuery());
});
// result === expensiveQuery() 
```

Under the hood these methods will create a transaction or span depending on if there is already an active span on the scope.

### DIFF
--- a/src/platform-includes/performance/add-active-span/javascript.mdx
+++ b/src/platform-includes/performance/add-active-span/javascript.mdx
@@ -1,0 +1,9 @@
+You can use the `Sentry.startActiveSpan` method to wrap a callback in a span to measure how long it will take. The span is automatically finished when the callback is finished. This will work with both sync and async callbacks.
+
+```javascript
+const result = Sentry.startActiveSpan({ name: "Important Function" }, () => {
+  return expensiveFunction();
+});
+```
+
+The span named `Important Function` will become the active span for the duration of the callback, which means that any child spans will be attached to it.
--- a/src/platform-includes/performance/add-active-span/javascript.mdx
+++ b/src/platform-includes/performance/add-active-span/javascript.mdx
@@ -4,6 +4,26 @@ You can use the `Sentry.startActiveSpan` method to wrap a callback in a span to 
 const result = Sentry.startActiveSpan({ name: "Important Function" }, () => {
   return expensiveFunction();
 });
+
+const result = await Sentry.startActiveSpan(
+  { name: "Important Function" },
+  async () => {
+    const res = Sentry.startActiveSpan({ name: "Child Span" }, () => {
+      return expensiveFunction();
+    });
+
+    return updateRes(res);
+  }
+);
+
+const result = Sentry.startActiveSpan(
+  { name: "Important Function" },
+  (span) => {
+    // Can access the span to add data or set specific status
+    span?.setData("foo", "bar");
+    return expensiveFunction();
+  }
+);
 ```
 
-The span named `Important Function` will become the active span for the duration of the callback, which means that any child spans will be attached to it.
+The span named `Important Function` will become the active span for the duration of the callback.
--- a/src/platform-includes/performance/add-active-span/javascript.mdx
+++ b/src/platform-includes/performance/add-active-span/javascript.mdx
@@ -19,7 +19,8 @@ const result = await Sentry.startActiveSpan(
 const result = Sentry.startActiveSpan(
   { name: "Important Function" },
   (span) => {
-    // Can access the span to add data or set specific status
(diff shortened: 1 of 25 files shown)


src/platform-includes/performance/span-api-version/javascript.mdx
```diff
@@ -0,0 +1,5 @@
+<Note>
+
+The span APIs require SDK version `7.65.0` or higher. If you are using an older version of the SDK, you can use the [explicit transaction APIs](#start-transaction) for custom instrumentation.
```
Lms24, line 3: l: Is there a reason why this note is different to the Node version? I'd prefer the Node version as it explicitly names the APIs.

Lms24 (approved):
LGTM generally, just had some minor comments.

Side note/no action required as too late: I'm only now realizing that `startActiveSpan` to me conveys that we only start a span but don't finish it after running the callback.  Super sorry for not bringing this up earlier! But it's also not the end of the world.

src/platform-includes/performance/add-active-span/javascript.mdx
```diff
@@ -0,0 +1,29 @@
+You can use the `Sentry.startActiveSpan` method to wrap a callback in a span to measure how long it will take. The span is automatically finished when the callback is finished. This will work with both sync and async callbacks.
+
+```javascript
+const result = Sentry.startActiveSpan({ name: "Important Function" }, () => {
+  return expensiveFunction();
+});
+
+const result = await Sentry.startActiveSpan(
+  { name: "Important Function" },
```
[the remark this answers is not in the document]
AbhiPrasad: op is optional so I didn't include it. We need to go back and expand documentation on it, but that can happen later on.

src/platforms/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -33,23 +33,63 @@ redirect_from:
...
+
+<PlatformContent includePath="performance/get-span" />
+
+## Start Independent Spans
+
+If you want to add a span that is not active, you can create a independent spans. This is useful for when you have work that is grouped together under a single parent span, but is independent from the current active span. In most cases you'll want to create and use active spans.
+
+<PlatformContent includePath="performance/add-independent-span" />
+
+## Start Transaction
```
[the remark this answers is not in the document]
AbhiPrasad: I don't necessarily want to deprecate this method for now, but once we're more comfortable with the API we can come back and adjust.

src/platform-includes/performance/span-api-version/javascript.mdx
AbhiPrasad: nope, I just forgot lol

good call!

AbhiPrasad:
> I'm only now realizing that startActiveSpan to me conveys that we only start a span but don't finish it after running the callback

Yeah I thought about this too, but the UX of just not having to think about span finish meant that it kinda just got included with the name. We can alias the name to something else in a future version if users get confused, we have room to move here.

src/platform-includes/performance/add-active-span/javascript.mdx
```diff
@@ -0,0 +1,29 @@
...
+
+    return updateRes(res);
+  }
+);
+
+const result = Sentry.startActiveSpan(
+  { name: "Important Function" },
+  (span) => {
+    // Can access the span to add data or set specific status
+    span?.setData("foo", "bar");
```
mydea, line 23: m: Should we mention/explain when/why `span` is undefined? May not be entirely clear from the outside why this may be unset and why I need to guard here.

AbhiPrasad: yup, ill add a note

PR #7708, opened 2023-08-29
fix(a11y): add more contrast to sidebar links

JoshFerge:
darker text, higher contrast -- fixes accessibility constrast minimum on sidebar links https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html
Screenshot 2023-08-29 at 4 35 48 PM
Screenshot 2023-08-29 at 4 35 38 PM

before and after.


vivianyentran (approved):
LGTM, thanks for fixing!

PR #7716, opened 2023-08-31
Add org selection dropdown to org auth token button

lforst:
Added a dropdown to select an organization for the org auth token button. This should help make it more clear for which org a token is gonna be generated. Especially on pages where we don't have `___ORG_SLUG___` or `___PROJECT_SLUG___` magic strings.

Looks like this:

Screenshot 2023-08-31 at 14 29 26


mydea:
Since this is sadly hard to test on the preview 😬 Just to clarify

* When there is only one org, we do not show the dropdown at all
* When there is more than one org, we show the same select UI as in the org/project select thingy, but only show the org (not the project) in there to select

I like it!

lforst:
> * When there is only one org, we do not show the dropdown at all

Correct! See

https://github.com/getsentry/sentry-docs/blob/4290615f5dc70876869f43001ac88b41ccbdf1c7/src/components/codeBlock.tsx#L214-L218

> * When there is more than one org, we show the same select UI as in the org/project select thingy, but only show the org (not the project) in there to select

Also correct! (See screenshot in PR description)

mydea:
Nice, that's what I figured but just to be extra clear :D Maybe @ArthurKnaus can also give it a review from a React perspective. From an outcome-perspective, I think this is great and will make this much more flexible to be used 🎉

src/components/codeBlock.tsx
```diff
@@ -106,52 +106,154 @@ type KeywordSelectorProps = {
 };
 
 function OrgAuthTokenCreator() {
-  const codeContext = useContext(CodeContext);
+  const [tokenState, setTokenState] = useState<
+    | {status: 'none'}
+    | {status: 'loading'}
+    | {status: 'success'; token: string}
+    | {status: 'error'}
```
ArthurKnaus, line 113: nit: I would probably define the type `TokenState` outside the component to keep the component cleaner 🤷 
Same goes for the popper options.

src/components/codeBlock.tsx
```diff
@@ -106,52 +106,154 @@ type KeywordSelectorProps = {
...
+  >({status: 'none'});
+  const [isOpen, setIsOpen] = useState(false);
+  const [referenceEl, setReferenceEl] = useState<HTMLSpanElement>(null);
+  const [dropdownEl, setDropdownEl] = useState<HTMLElement>(null);
+  const {codeKeywords} = useContext(CodeContext);
```
ArthurKnaus, line 118: nit: I would move the context subscription to the first line then it is easier to spot.
(as it is a dependency that component has)

src/components/codeBlock.tsx
```diff
@@ -106,52 +106,154 @@ type KeywordSelectorProps = {
...
+        status: 'error',
+      });
+    }
+  };
+
+  const orgSet = new Set<string>();
+  codeKeywords?.PROJECT?.forEach(projectKeyword => {
+    orgSet.add(projectKeyword.ORG_SLUG);
+  });
+  const orgs = [...orgSet];
```
ArthurKnaus, line 159: So it is clear further down the component that those are the slugs and not whole objects.
```suggestion
  const orgSlugs = [...orgSet];
```

src/components/codeBlock.tsx
```diff
@@ -106,52 +106,154 @@ type KeywordSelectorProps = {
...
+      <KeywordDropdown
+        ref={setReferenceEl}
+        role="button"
+        title="Click to generate token"
+        tabIndex={0}
+        onClick={() => {
+          handlePress();
+        }}
+        onKeyDown={e => {
+          if (e.key === 'Enter') {
```
ArthurKnaus, line 232: A native button also fires on `Space`.
```suggestion
          if (['Enter', 'Space'].includes(e.key)) {
```

ArthurKnaus (approved):
Nice feature, love it! 🎉 
Mostly nitpicking from my side:

PR #7724, opened 2023-08-31
fix(a11y): actually query for lang

JoshFerge:
in my previous pr #7712 it didn't actually work because i didn't update the graphql query.


shanamatthews (approved):
Classic mistake 😄

JoshFerge:
FYI this still isnt working in preview. will take a look at this again sometime soon.

PR #7732, opened 2023-09-05
THOUGHT EXPERIMENT: new performance api

antonpirker:
This is just an experiment how the new performance api could look like. 

In Python: https://sentry-docs-git-antonpirker-python-new-performance-api.sentry.dev/platforms/python/performance/instrumentation/custom-instrumentation/?original_referrer=https%3A%2F%2Fgithub.com%2Fgetsentry%2Fsentry-docs%2Fpull%2F7732

In Javascript: https://sentry-docs-git-antonpirker-python-new-performance-api.sentry.dev/platforms/node/performance/instrumentation/custom-instrumentation/?original_referrer=https%3A%2F%2Fgithub.com%2Fgetsentry%2Fsentry-docs%2Fpull%2F7732

### DIFF
--- a/src/platforms/common/performance/instrumentation/custom-instrumentation.mdx
+++ b/src/platforms/common/performance/instrumentation/custom-instrumentation.mdx
@@ -31,86 +31,114 @@ redirect_from:
   - /performance/custom-instrumentation
 ---
 
+
 <Note>
 
-To capture transactions and spans customized to your organization's needs, you must first <PlatformLink to="/performance/">set up performance monitoring.</PlatformLink>
+This is just a THOUGHT experiment and should not be merged!
 
 </Note>
 
-<PlatformSection notSupported={["javascript", "node"]}>
 
-<PlatformContent includePath="performance/enable-manual-instrumentation" />
 
-<PlatformContent includePath="performance/add-spans-example" />
+## Creating Spans
 
-</PlatformSection>
+```javascript
+function doWork() {
+  const result = Sentry.startSpan({ name: "Important Function" }, () => {
+    return expensiveFunction();
+  });
+}
+```
 
-<PlatformSection supported={["javascript", "node"]}>
+## Creacting Nested Spans
 
-To add custom performance data to your application, you need to create and use spans. Spans are a way to measure the time it takes for a specific action to occur. For example, you can create a span to measure the time it takes for a function to execute.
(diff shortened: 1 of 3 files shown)


src/platforms/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -33,84 +33,106 @@ redirect_from:
...
+childSpan.finish();
+parentSpan.finish();
+```
 
-The root span (the span that is the parent of all other spans) is known as a transaction in Sentry. This can be accessed and created separately if you need more control over the timing data or if you use a version of the SDK that does not support the top level span APIs.
+If you want to create independend spans that are not in a parent/child relation ship a as above you can create them like this:
 
-<PlatformContent includePath="performance/enable-manual-instrumentation" />
+```javascript
+const span1 = Sentry.startSpan({ name: "first-step", independend: true });
```
AbhiPrasad, line 100: I prefer not to make this an argument because the behavior changes quite a bit between the two modes.

Active span mode **requires** the use of a callback (to wrap) and returns the result of the callback - not the span.

Independent span mode returns just the span.

Certain languages cannot overload arguments/return values like this.

src/platforms/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -33,84 +33,106 @@ redirect_from:
...
 
 </Note>
 
-<PlatformSection notSupported={["javascript", "node"]}>
+## Creating Spans
 
-<PlatformContent includePath="performance/enable-manual-instrumentation" />
+```javascript
+function doWork() {
+  const result = Sentry.startSpan({ name: "Important Function" }, () => {
```
AbhiPrasad, line 44: I guess the struggle with naming here is between `startActiveSpan` vs. `startSpan`.

I can defer to whatever the majority thinks should be done but here is some of my reasoning toward choosing the existing naming in the RFC.

1. It aligns with OpenTelemetry
2. `Sentry.startSpan` maps in behavior exactly to `Sentry.startTransaction` (doesn't add itself to scope, returns a span/transaction to use, etc.)
3. `startActiveSpan` indicates something special happening to the passed in callback. `trace` is also good here, but is an overloaded word.

If we wanted to explore alternative naming, `withSpan` or `withActiveSpan` also makes sense in my head.

antonpirker:
Thanks for the input @AbhiPrasad 

This branch has now done his duty as basis for discussion. I will close this.

PR #7735, opened 2023-09-05
feat(react): Document error boundary linked error

AbhiPrasad:
I merged https://github.com/getsentry/sentry-javascript/pull/3532 around 2+ years ago at this point, but never got around to documenting it. Making up for my mistake now (also because someone asked).

image


src/platforms/javascript/guides/react/features/error-boundary.mdx
```diff
@@ -65,6 +67,12 @@ By default React [logs all errors to the console](https://github.com/facebook/re
 
 </Alert>
 
+## Linked Errors
```
Lms24, line 70: l: iirc there's something special about the parent/child relationship here. Should we mention this?

Lms24 (approved):

src/platforms/javascript/guides/react/features/error-boundary.mdx
Lms24: Found what I had in mind: https://github.com/getsentry/sentry-javascript/issues/8871

AbhiPrasad: yeah good point I'll write that down.

PR #7748, opened 2023-09-06
ref(js): Extract codeKeywords from codeBlock component

evanpurkhiser:
There's just too much stuff in here


src/components/codeKeywords.tsx
```diff
@@ -0,0 +1,544 @@
...
+  codeKeywords?.PROJECT?.forEach(projectKeyword => {
+    orgSet.add(projectKeyword.ORG_SLUG);
+  });
+  const orgSlugs = [...orgSet];
+
+  const [isAnimating, setIsAnimating] = useState(false);
+
+  if (!codeKeywords.USER) {
+    // User is not logged in - show dummy token
+    return <Fragment>sntrys_YOUR_TOKEN_HERE</Fragment>;
```
shanamatthews, line 194: unrelated to your change, but do you know why this is "sntrys_YOUR_TOKEN_HERE"? 
the "sntrys" parts specifically just looks like a typo

shanamatthews (approved):
Should maybe break this up further, but this is a good start

evanpurkhiser:
Yeah I plan to split this up even more.

src/components/codeKeywords.tsx
mydea: this is correct, the org tokens _always_ have the `sntrys_` prefix, so you can identify immediately that this is a sentry token :)

PR #7760, opened 2023-09-06
feat(python): Add backpressure option and add note in overhead section

sl0thentr0py:


vivianyentran (approved):
Added a couple of edits. Thanks for updating the docs!

PR #7774, opened 2023-09-08
Add some information about hosting environment overhead

lforst:
Adds some information about the overhead that might be caused by the sdk in certain hosting environments like AWS lambdas or Vercel edge runtime.

### DIFF
--- a/src/docs/product/performance/performance-overhead.mdx
+++ b/src/docs/product/performance/performance-overhead.mdx
@@ -32,3 +32,10 @@ Each SDK has limits on the data structures used to record all the above data. If
 ### Backpressure Management
 
 The Python SDK, starting with version `1.30.0`, manages backpressure internally by reducing the dynamic sampling rate when the system is in an unhealthy state. This ensures better system stability and lower SDK overhead under heavy load. We will gradually roll this out to other backend languages in the future.
+
+### Hosting Environment
+
+Some hosting environments have mechanisms in place that will have a negative influence on SDK overhead:
+
+- AWS Lambdas completely freeze any code execution when a response is sent. This means that the SDKs generally will block the response from being finished until it has flushed all of the recorded telemetry data to Sentry. This might cause non-trivial overhead in some situations. Here, it might be beneficial to run a Sentry Relay instance physically close to your lambdas to decrease latency.
+- SDKs that are designed to run on the Vercel Edge Runtime will stall responses to requests until the SDK has flushed all of the telemetry for that request to Sentry. This is a limitation imposed by the Edge runtime, since it doesn't allow for I/O operations to be shared or persisted across request handler invocations, meaning the SDK cannot flush its data asynchronously.
--- a/src/docs/product/performance/performance-overhead.mdx
+++ b/src/docs/product/performance/performance-overhead.mdx
@@ -35,7 +35,7 @@ The Python SDK, starting with version `1.30.0`, manages backpressure internally 
 
 ### Hosting Environment
 
-Some hosting environments have mechanisms in place that will have a negative influence on SDK overhead:
+Some hosting environments have mechanisms in place that negatively influence SDK overhead:
 
 - AWS Lambdas completely freeze any code execution when a response is sent. This means that the SDKs generally will block the response from being finished until it has flushed all of the recorded telemetry data to Sentry. This might cause non-trivial overhead in some situations. Here, it might be beneficial to run a Sentry Relay instance physically close to your lambdas to decrease latency.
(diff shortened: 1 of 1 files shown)


src/docs/product/performance/performance-overhead.mdx
```diff
@@ -32,3 +32,10 @@ Each SDK has limits on the data structures used to record all the above data. If
 ### Backpressure Management
 
 The Python SDK, starting with version `1.30.0`, manages backpressure internally by reducing the dynamic sampling rate when the system is in an unhealthy state. This ensures better system stability and lower SDK overhead under heavy load. We will gradually roll this out to other backend languages in the future.
+
+### Hosting Environment
+
+Some hosting environments have mechanisms in place that will have a negative influence on SDK overhead:
+
+- AWS Lambdas completely freeze any code execution when a response is sent. This means that the SDKs generally will block the response from being finished until it has flushed all of the recorded telemetry data to Sentry. This might cause non-trivial overhead in some situations. Here, it might be beneficial to run a Sentry Relay instance physically close to your lambdas to decrease latency.
```
mwarkentin, line 40: I think this is also the purpose of Lambda Extensions, they can be used for monitoring / observability tools to support flushing after the main response has completed: https://docs.aws.amazon.com/lambda/latest/dg/lambda-extensions.html

Not sure if we have any solutions there.

AbhiPrasad: We actually tried using lambda extensions and found that it didn't improve sdk overhead and in some cases increased it! cc @sl0thentr0py.

src/docs/product/performance/performance-overhead.mdx
```diff
@@ -32,3 +32,10 @@ Each SDK has limits on the data structures used to record all the above data. If
 ### Backpressure Management
 
 The Python SDK, starting with version `1.30.0`, manages backpressure internally by reducing the dynamic sampling rate when the system is in an unhealthy state. This ensures better system stability and lower SDK overhead under heavy load. We will gradually roll this out to other backend languages in the future.
+
+### Hosting Environment
+
+Some hosting environments have mechanisms in place that will have a negative influence on SDK overhead:
```
vivianyentran, line 38:
```suggestion
Some hosting environments have mechanisms in place that negatively influence SDK overhead:
```

src/docs/product/performance/performance-overhead.mdx
vivianyentran, line 40:
```suggestion
- AWS Lambda completely freezes any code execution when a response is sent. To keep code running, the SDKs generally block the response from being finished until it has flushed all of the recorded telemetry data to Sentry. This might cause non-trivial overhead in some situations. Here, it might be beneficial to run a Sentry Relay instance physically close to your lambdas to decrease latency.
```

I think "physically close" is a little confusing here. Do you mean in the same region or availability zone, or something else?

src/docs/product/performance/performance-overhead.mdx
```diff
@@ -32,3 +32,10 @@ Each SDK has limits on the data structures used to record all the above data. If
...
+
+- AWS Lambdas completely freeze any code execution when a response is sent. This means that the SDKs generally will block the response from being finished until it has flushed all of the recorded telemetry data to Sentry. This might cause non-trivial overhead in some situations. Here, it might be beneficial to run a Sentry Relay instance physically close to your lambdas to decrease latency.
+- SDKs that are designed to run on the Vercel Edge Runtime will stall responses to requests until the SDK has flushed all of the telemetry for that request to Sentry. This is a limitation imposed by the Edge runtime, since it doesn't allow for I/O operations to be shared or persisted across request handler invocations, meaning the SDK cannot flush its data asynchronously.
```
vivianyentran, line 41:
```suggestion
- SDKs that are designed to run on the Vercel Edge Runtime will stall responses to requests until the SDK has flushed all of the telemetry for that request to Sentry. This is a limitation imposed by the Edge Runtime, since it doesn't allow for I/O operations to be shared or persisted across request handler invocations, meaning the SDK cannot flush its data asynchronously.
```

vivianyentran (approved):
Added some small copyedits but looks good! Thanks for adding this section!

src/docs/product/performance/performance-overhead.mdx
lforst: Yes. But I don't really want to write "same region" or same "availability zone" because having servers that run physically close to eachother will generally have lower latency. They don't necessarily need to be in the same region and "availability zones" are technically something unrelated to locality.

PR #7797, opened 2023-09-13
Wrapped Key with square brackets

### LINKED ISSUE #7793: EXTRA_KEY Configuration in Micro Frontend Documentation
### Core or SDK?

Platform/SDK

### Which part? Which one?

Combining ModuleMetadata and makeMultiplexedTransport

### Description

I think there is a tiny syntax issue I encountered with this when I tried implementing it with TS.

`const EXTRA_KEY = "ROUTE_TO"`

is declared and used inside `event.extra`

But tsc does not recognize it when its written:

```
.
.
if (routeTo.length) {
        event.extra = {
          ...event.extra,
          EXTRA_KEY: routeTo,
        };
      }
    }

    return event;
  },
});
```

If the intention was to literally set the key to EXTRA_KEY then this is correct, but if the intention is to set the key to "ROUTE_TO" then it should be wrapped with square brackets like so:

```
.
.
if (routeTo.length) {
        event.extra = {
          ...event.extra,
          [EXTRA_KEY]: routeTo,
        };
      }
    }

    return event;
  },
});
```

Without the brackets, the object would literally have a property named "EXTRA_KEY", not a property named whatever string value EXTRA_KEY is holding.
(I could be totally off, did not investigate it much further)

### Suggested Solution

Wrap EXTRA_KEY with square brackets.

shakedhagag:
## Description of changes

Added square brackets for correct syntax.
fixes #7793 

## Legal Boilerplate

## Extra resources

### DIFF
--- a/src/platforms/javascript/common/configuration/micro-frontend-support.mdx
+++ b/src/platforms/javascript/common/configuration/micro-frontend-support.mdx
@@ -226,7 +226,7 @@ init({
       if (routeTo.length) {
         event.extra = {
           ...event.extra,
-          EXTRA_KEY: routeTo,
+          [EXTRA_KEY]: routeTo,
         };
       }
     }


AbhiPrasad (approved):
Thank you!

PR #7825, opened 2023-09-15
Improve Go manual instrumentation examples

cleptric:
## Pre-merge checklist

### DIFF
--- a/src/platform-includes/performance/add-spans-example/go.mdx
+++ b/src/platform-includes/performance/add-spans-example/go.mdx
@@ -4,11 +4,13 @@ The next example contains the implementation of the hypothetical `doWork` functi
 
 ```go
 func doWork(ctx context.Context) {
+	// Set the OP based on values from https://develop.sentry.dev/sdk/performance/span-operations/
 	span := sentry.StartSpan(ctx, "function")
 	span.Description = "suboperation1"
 	// omitted code ...
 	span.Finish()
 
+	// Set the OP based on values from https://develop.sentry.dev/sdk/performance/span-operations/
 	span := sentry.StartSpan(ctx, "function")
 	span.Description = "suboperation2"
 	// omitted code ...
--- a/src/platform-includes/performance/add-spans-example/go.mdx
+++ b/src/platform-includes/performance/add-spans-example/go.mdx
@@ -1,15 +1,16 @@
 ## Add More Spans to the Transaction
 
-The next example contains the implementation of the hypothetical `doWork` function called from the code snippet in the previous section. Our SDK can determine if there is currently an open transaction in the current context and add all newly created spans as child operations to that transaction. Keep in mind that each individual span also needs to be manually finished; otherwise, spans will not show up in the transaction.
-
-You can choose the value of `operation` and `description`.
+The next example contains the implementation of the hypothetical `doWork` function called from the code snippet in the previous section. Our SDK can determine if there is currently an open transaction in the current context and add all newly created spans as child operations to that transaction. Keep in mind that each individual span also needs to be manually finished. Otherwise, spans will not show up on the transaction.
 
 ```go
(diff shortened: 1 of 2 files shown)


src/platform-includes/performance/add-spans-example/go.mdx
```diff
@@ -1,15 +1,16 @@
...
-
-You can choose the value of `operation` and `description`.
+The next example contains the implementation of the hypothetical `doWork` function called from the code snippet in the previous section. Our SDK can determine if there is currently an open transaction in the current context and add all newly created spans as child operations to that transaction. Keep in mind that each individual span also needs to be manually finished. Otherwise, spans will not show up on the transaction.
 
 ```go
-func doWork(ctx context.Context, item Item) {
-	span := sentry.StartSpan(ctx, "suboperation1")
+func doWork(ctx context.Context) {
+	span := sentry.StartSpan(ctx, "function")
+	span.Description = "suboperation1"
```
tonyo, line 8: How about this:

```suggestion
	span := sentry.StartSpan(ctx, "span name 1")
	span.Description = "span description 1"
```
(and same below)

Otherwise "function" is a bit too specific (we don't only wrap functions in spans), and also both spans will be named the same.

src/platform-includes/performance/enable-manual-instrumentation/go.mdx
```diff
@@ -1,14 +1,28 @@
...
-	span := sentry.StartSpan(ctx, "doWork",
-		sentry.WithTransactionName(fmt.Sprintf("doWork: %s", item.Type)))
-	doWork(span.Context(), item) // doWork may create additional spans
-	span.Finish()
-}
+http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	hub := sentry.GetHubFromContext(ctx)
+	if hub == nil {
+		hub = sentry.CurrentHub().Clone()
```
tonyo, line 10: This example uses a couple concurrency-related patterns (whose intent might not be immediately clear to users), and also uses HTTP-related example (for that people should probably use web-related integrations).

If you really want to have this example instead, let's maybe add a comment (e.g. pointing at https://docs.sentry.io/platforms/go/concurrency/) and also a note that if "you're doing HTTP stuff, then look at the integrations".

src/platform-includes/performance/add-spans-example/go.mdx
tonyo: name -> operation

cleptric: The `op` should be from https://develop.sentry.dev/sdk/performance/span-operations/#general; more context is added with the description.

src/platform-includes/performance/enable-manual-instrumentation/go.mdx
cleptric: We could also use a more straightforward example, though I would argue that the 90% use-case is likely an HTTP request.

src/platform-includes/performance/add-spans-example/go.mdx
tonyo: Ah, confused with StartTransaction. 
Let's then link to https://develop.sentry.dev/sdk/performance/span-operations/ here? This is the part that would be confusing to me personally, it's not immediately clear what should go into that second argument.

src/platform-includes/performance/enable-manual-instrumentation/go.mdx
tonyo:
```suggestion
		// Check the concurrency guide for more details: https://docs.sentry.io/platforms/go/concurrency/
		hub = sentry.CurrentHub().Clone()
```

shanamatthews (approved):
Looks good from a language perspective!

PR #7858, opened 2023-09-18
feat: Add Bun docs

HazAT:
Add docs for Bun

### DIFF
--- a/src/platforms/javascript/guides/bun/config.yml
+++ b/src/platforms/javascript/guides/bun/config.yml
@@ -1,0 +1,8 @@
+title: Bun
+sdk: sentry.javascript.bun
+fallbackPlatform: javascript
+caseStyle: camelCase
+supportLevel: production
+categories:
+  - browser
+  - server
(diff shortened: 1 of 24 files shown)


src/platforms/javascript/guides/bun/config.yml
shanamatthews, line 1: Do these docs make more sense to live here, in the browser JS frameworks, or in the Node frameworks?

HazAT: yeah good point, we discussed this
it's either/or 😅 

At one point we might just merge Javascript and Node together.

HazAT:
This commit [Add icon to Platform](https://github.com/getsentry/sentry-docs/pull/7858/commits/4bdf91190dc630238571eddc68cde70ed5f5b094) is more or less independent from bun, makes it possible to define a `icon` attribute in the platform config if the platform icon has a different name

AbhiPrasad (approved):

PR #7859, opened 2023-09-18
Python: Getting Started socket

antonpirker:


src/platforms/python/common/configuration/integrations/socket.mdx
shanamatthews, line 1: Do you know why the title of this page is in all caps?

shanamatthews (approved):

PR #7871, opened 2023-09-18
fix(feedback): Fix `getCurrentHub` undefined

billyvg:
Possible that the loader SDK has not loaded before rendering the widget modal (seems unlikely though)

Fixes DOCS-6A1

### DIFF
--- a/src/components/feedback/feedbackModal.tsx
+++ b/src/components/feedback/feedbackModal.tsx
@@ -188,7 +188,10 @@ export function FeedbackModal({open, onClose, onSubmit}: FeedbackModalProps) {
     });
   };
 
-  const user = window.Sentry?.getCurrentHub().getScope()?.getUser();
+  const user =
+    typeof window.Sentry?.getCurrentHub === 'function'
+      ? window.Sentry?.getCurrentHub().getScope()?.getUser()
+      : null;
 
   return (
     <Dialog id="feedbackModal" open={open} ref={dialogRef} onClick={onClose}>


ryan953 (approved):

src/components/feedback/feedbackModal.tsx
```diff
@@ -188,7 +188,10 @@ export function FeedbackModal({open, onClose, onSubmit}: FeedbackModalProps) {
     });
   };
 
-  const user = window.Sentry?.getCurrentHub().getScope()?.getUser();
+  const user =
+    typeof window.Sentry?.getCurrentHub === 'function'
+      ? window.Sentry?.getCurrentHub().getScope()?.getUser()
+      : null;
```
ryan953, line 194: so many `?` in there already... should throw in more instead?

const user = window.Sentry?.getCurrentHub()?.getScope()?.getUser() ?? null;

ryan953, line 194: so many `?` in there already... should throw in more instead?

```suggestion
const user = window.Sentry?.getCurrentHub()?.getScope()?.getUser() ?? null;
```

billyvg: Ah it'd have to be this because `getCurrentHub` can be undefined (not sure I love `?.()` syntax but also [l])

```suggestion
  const user = window.Sentry?.getCurrentHub?.()?.getScope()?.getUser();
```

PR #7874, opened 2023-09-19
ref(js): Add Source Maps wizard snippet to Vue Source Maps page

Lms24:
As part of completing https://github.com/getsentry/team-sdks/issues/2, this PR adds an improvement to the Vue source maps page. Seems like we forgot to add the source maps wizard snippet to the Vue-specific version of this page. This PR adds the snippet and aligns the structure with the Angular and Svelte-specific pages (i.e. automatic and manual setup sections). 

ref: https://github.com/getsentry/team-sdks/issues/2

### DIFF
--- a/src/platform-includes/sourcemaps/overview/javascript.vue.mdx
+++ b/src/platform-includes/sourcemaps/overview/javascript.vue.mdx
@@ -1,4 +1,4 @@
-## Uploading Source Maps using Vite
+## Uploading Source Maps in a Vue Project
 
 <Note>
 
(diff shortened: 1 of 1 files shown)


Lms24:
Worth mentioning that we had a customer call where the customer (developing a Vue/Nuxt app) said that they didn't see the wizard snippet in the docs. Not entirely sure if this exactly the place but I think it'd make sense.

AbhiPrasad (approved):

shanamatthews (approved):

PR #7875, opened 2023-09-19
ref(js): Clarify release creation in Sentry CLI source maps upload guide

Lms24:
As part of completing https://github.com/getsentry/team-sdks/issues/2, this PR slightly improves wording around how to deal with releases in the new artifact bundle upload procedure. This is the result of a customer call where the customer explained how they were confused if they had to create a release with the CLI or not. It's just a slight improvement though, nothing too fancy.

ref: https://github.com/getsentry/team-sdks/issues/2


vivianyentran (approved):
LGTM, thanks for the updates!

PR #7887, opened 2023-09-20
ref(sveltekit): Update SvelteKit version compatibility

Lms24:
This PR adds a recommendation to use our SDK with SvelteKit 1.24.0 or newer. The reason is that 1.24.0 contains an important [fix](https://github.com/sveltejs/kit/pull/10576) that lets us [avoid invalidating route data ](https://github.com/getsentry/sentry-javascript/pull/9071 )that previously caused an additional round of data loading. 

To be clear, we're not breaking compatibility with our minimum supported version (1.0.0) but for everything before 1.24.0 the SDK causes a slight performance hit, hence the recommendation.

ref https://github.com/getsentry/sentry-javascript/issues/8818


vivianyentran (approved):
lgtm

AbhiPrasad (approved):

PR #7910, opened 2023-09-20
feat(feedback): New feedback button + form design

billyvg:
New design based on @Jesse-Box mox

image

image

image

Form is not so great at small heights:

image


billyvg:
@Jesse-Box Note I kept the Sentry purple on the button

ryan953 (approved):

PR #7913, opened 2023-09-21
Bumped `platformicons` to make docs nicer looking

antonpirker:
Now more of the Python guides have an icon in the sidebar! (also renamed one guide form "pyspark" to just "spark" so the right icon is taken (also added a redirect to not have 404s)


shanamatthews (approved):
This looks so nice! Great work @antonpirker !

PR #7917, opened 2023-09-21
Add Disabled Alerts section

vivianyentran:
## Pre-merge checklist

## Description of changes

- Add section on disabled alerts

## Legal Boilerplate

## Extra resources

### DIFF
--- a/src/docs/product/alerts/index.mdx
+++ b/src/docs/product/alerts/index.mdx
@@ -66,7 +66,7 @@ Metric alerts work in the same way as Issue alerts and can be muted on the **Ale
 
 Sentry has disabled duplicate alerts and alerts with no actions. If a noisy alert is sent to a communication channel and receives no engagement for an extended time, Sentry will proactively disable it to reduce unnecessary noise.
 
-Anyone with [alert edit access](product/accounts/membership/) can re-enable an alert by editing its conditions and re-saving it. Alerts need to pass Sentry’s checks before they can be saved. See [best practices](product/alerts/best-practices/) for guidance on writing useful alerts.
+Anyone with [alert edit access](/product/accounts/membership/) can re-enable an alert by editing its conditions and re-saving it. Alerts need to pass Sentry’s checks before they can be saved. See [best practices](/product/alerts/best-practices/) for guidance on writing useful alerts.
 
 ## Notifications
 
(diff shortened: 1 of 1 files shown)


src/docs/product/alerts/index.mdx
```diff
@@ -62,6 +62,16 @@ Issue alerts can be muted on the **Alerts** details page by clicking the "Mute"
 
 Metric alerts work in the same way as Issue alerts and can be muted on the **Alerts** details page by clicking the "Mute" button.
 
+## Disabled Alerts
+
+To ensure that your alerts are working properly without causing unnecessary noise, Sentry will proactively disable an alert from sending if we detect that the alert:
+
+- Is a duplicate of an existing alert rule.
```
ceorourke, line 69: For the first 2 bullet points, this is something that we block going forward, and disabled already created rules that matched the criteria in a migration. I don't think we should add docs for something we ran once - if instead we didn't block these from being created and continually ran migrations to disable them (which would be a silly idea) I could see adding this, but as it stands now I don't think these should be included.

The 3rd one I hesitate to agree should be listed because for one, we changed it to 5000, and for another, when we run it again we may change the time period and number of triggers. @rachrwang what do you think? 

It's also untrue that anyone in the organization can re-enable an alert, they must have the correct permissions (or if that is true, it's a mistake and we should fix it).

vivianyentran: Good point... Are there any other conditions for disabling an alert? How's the following revision for the entire section?

"Sentry has disabled duplicate alerts and alerts with no actions. If a noisy alert is sent to a communication channel and receives no engagement for 30 days, Sentry will proactively disable it to reduce unnecessary noise.

Alert owners can re-enable an alert by editing its conditions and re-saving it. Alerts need to pass Sentry’s checks before they can be saved. See [best practices](product/alerts/best-practices/) for guidance on writing useful alerts."

ceorourke: Can we be even more vague 😅 we might not always do the last 30 days. 

Maybe we could link to the [team permissions](https://docs.sentry.io/product/accounts/membership/#team-level-roles) docs for the last part about who can re-enable, it's just anyone who has alert edit access.

vivianyentran: Ok, introduced additional vagueness and added the link. LMK if you or @rachrwang  have any other suggestions!

PR #7918, opened 2023-09-21
fix(docs): Match user token update and prioritize internal integrations

schew2381:
### Background
As part of the change to [Introduce Scope Hierarchy](https://www.notion.so/sentry/24Q3-Permission-Scope-Hierarchy-67620f99783345bbb8bc2828b1addc64), I edited the page where you create user auth tokens to use the same dropdowns as internal integrations in https://github.com/getsentry/sentry/pull/54651. You can test this for yourself [here](https://sentry.io/settings/account/api/auth-tokens/new-token/).

### Changes
1. Clarify that user auth tokens are not editable after they are created. In order to change the scopes on them, you must create a new token.
2. Replace old picture of the create user auth page from box selectors to the new dropdowns
3. I moved user auth tokens to below integration tokens to hopefully have people favor integration > user tokens. In the future, we will probably look into eventually deprecating user auth tokens in favor of the other 2 tokens.
4. For the internal integration, I changed the wording from 
`See our docs on Internal Integrations to learn more.`
to
`To get started, see our docs on Internal Integrations`
because it wasn't immediately clear to me which link in that section leads to the steps to create the integration.


src/docs/product/accounts/auth-tokens/index.mdx
```diff
@@ -21,26 +21,26 @@ There are three key types of auth tokens in Sentry:
...
 
-[User Auth Tokens](#user-auth-tokens) should be used to interact with the Sentry API on behalf of a user. For example, to fetch all issues for a user, you would use a User auth token. We don't recommend using User auth tokens for CI tasks because if the user who created the token is removed from the Organization, the token will stop working.
-
-User auth token permissions are customizable and editable.
-
 [Internal Integrations](#internal-integrations) should be used when you need full API access (which the organization auth tokens cannot grant), and you want to interact with the Sentry API on behalf of an organization. For example, to programmatically create a new project, you would use an internal integration.
 
 Permissions for auth tokens created as part of an internal integration are customizable and editable.
 
+[User Auth Tokens](#user-auth-tokens) should be used to interact with the Sentry API on behalf of a user. For example, to fetch all issues for a user, you would use a User auth token. We don't recommend using User auth tokens for CI tasks because if the user who created the token is removed from the Organization, the token will stop working.
```
shanamatthews, line 40: Sorry, just noticed this incorrect capitalization when you moved this :)

```suggestion
[User Auth Tokens](#user-auth-tokens) should be used to interact with the Sentry API on behalf of a user. For example, to fetch all issues for a user, you would use a user auth token. We don't recommend using User auth tokens for CI tasks because if the user who created the token is removed from the Organization, the token will stop working.
```

shanamatthews (approved):
lgtm!

PR #7933, opened 2023-09-25
Remove requirement for "Business Plan" 

ale-cota:
Retention priorities are available on all AM2 plans with more than 1M tx/month, so this line is misleading. We explain the availability on latest plans in the callout on the top of the page, so I would remove it completely.

Screenshot 2023-09-25 at 11 36 58


vivianyentran (approved):
lgtm

lizokm (approved):
LGTM

PR #7943, opened 2023-09-26
feat(node): Add docs for ANR tracking

AbhiPrasad:
ref https://github.com/getsentry/sentry-javascript/issues/9108

Documents https://github.com/getsentry/sentry-javascript/pull/9079

cc @timfish for review.


src/platforms/node/common/configuration/application-not-responding.mdx
```diff
@@ -0,0 +1,108 @@
...
+   */
+  debug: boolean;
+}
+```
+
+## ANR Implementation and Overhead
+
+ANR detection with the Node SDK works with a forked child process. The child process runs the same entry point as the main app. To ensure that the main app code does not run in the child process, we use a promise that only resolves in the main process.
+
+The child process sends a heartbeat message to the main process every 50ms. If the main process does not receive a heartbeat message for 5 seconds, it triggers an ANR event. If the `captureStackTrace` option is enabled, the child process uses WebSockets to capture stack traces via the [v8 inspector API](https://nodejs.org/api/inspector.html).
```
timfish, line 104: The heartbeat message is the other direction, main to child.

src/platforms/node/common/configuration/application-not-responding.mdx
```diff
@@ -0,0 +1,108 @@
...
+  ]
+---
+
+Application Not Responding (ANR) errors, or Event Loop Stall errors, are triggered when the Node.js main thread event loop of an application is blocked for more than five seconds. The Node SDK reports ANR errors as Sentry events and can optionally attach a stacktrace of the blocking code to the ANR event.
+
+<Include name="feature-stage-beta.mdx" />
+
+_(Available in version 7.72.0 and above)_
+
+To enable ANR detection, import and use the `enableANRDetection` function from the `@sentry/node` package before you run the rest of your application code. Any application code run before `enableANRDetection` will not be detected by the SDK.
```
timfish, line 20:
```suggestion
To enable ANR detection, import and use the `enableANRDetection` function from the `@sentry/node` package before you run the rest of your application code. Any event loop blocking before calling `enableANRDetection` will not be detected by the SDK.
```

src/platforms/node/common/configuration/application-not-responding.mdx
timfish, line 104:
```suggestion
The main app process sends a heartbeat message to the child process every 50ms. If the child process does not receive a heartbeat message for 5 seconds, it triggers an ANR event. If the `captureStackTrace` option is enabled, the child process uses WebSockets to capture stack traces via the [v8 inspector API](https://nodejs.org/api/inspector.html).
```

AbhiPrasad:
Going to merge this in since we did a release - want to start sharing docs so that folks can experiment with this. We'll come back to this again when we document Electron ANR stuff.

Issue #7435, opened 2023-07-13
Make it clear RequestData is only avaliable for NextJS server side


rodolfoBee:
### Is there an existing issue for this?

- [X] I have checked for existing issues https://github.com/getsentry/sentry-javascript/issues
- [X] I have reviewed the documentation https://docs.sentry.io/
- [X] I am using the latest SDK release https://github.com/getsentry/sentry-javascript/releases

### How do you use Sentry?

Sentry Saas (sentry.io)

### Which SDK are you using?

@sentry/nextjs

### SDK Version

7.58.1

### SDK Setup

in sentry.client.js:

```
import * as Sentry from "@sentry/nextjs";

Sentry.init({
  dsn: "<secret>",
  integrations: [
      new Sentry.Integrations.RequestData({include: {data: false,cookies: false,ip: false,}})
  ],
});
```

### Steps to Reproduce

Following the docs to Modifying System Integrations
https://docs.sentry.io/platforms/javascript/guides/nextjs/configuration/integrations/default/#modifying-system-integrations

Im adding the import name from the RequestData integration to my init to edit it:

```
RequestData
Import name: Sentry.Integrations.RequestData
```

### Expected Result

The integration is modified.

### Actual Result

getting the following error in the console: 

```
sentry.client.config.js:16 Uncaught TypeError: _sentry_nextjs__WEBPACK_IMPORTED_MODULE_0__.Integrations.RequestData is not a constructor
    at eval (sentry.client.config.js:16:7)
    at ./sentry.client.config.js (_app.js?ts=1689250891003:1880:1)
    at options.factory (webpack.js?ts=1689250891003:676:31)
    at __webpack_require__ (webpack.js?ts=1689250891003:37:33)
    at __webpack_exec__ (_app.js?ts=1689250891003:1939:61)
    at _app.js?ts=1689250891003:1940:65
    at __webpack_require__.O (webpack.js?ts=1689250891003:86:23)
    at _app.js?ts=1689250891003:1941:56
    at webpackJsonpCallback (webpack.js?ts=1689250891003:1224:39)
    at _app.js?ts=1689250891003:9:61
```

as result the SDK fails to initialise and the page crashes.

PR #7948, opened 2023-09-26
Clarify runtime dependant integration usage for Next.js

lforst:
Fixes https://github.com/getsentry/sentry-docs/issues/7435

### DIFF
--- a/src/platforms/javascript/common/configuration/integrations/index.mdx
+++ b/src/platforms/javascript/common/configuration/integrations/index.mdx
@@ -9,6 +9,22 @@ redirect_from:
 
 Sentry provides a number of integrations which are designed to add automatic instrumentation to your application. Some of these integrations are enabled by default, while others have to be explicitly enabled.
 
+<PlatformSection supported={["javascript.nextjs"]}>
+
+<Alert title="Integration Usage In Next.js">
+
+Next.js can operate within three runtimes: the Node.js runtime, the browser runtime, or the Edge runtime. However, it's important to note that not all integrations are compatible with all of these runtimes.
+
+Depending on whether an integration enhances the functionality of a particular runtime, such as the BrowserTracing integration for the browser runtime or the RequestData integration for the Node.js runtime, you can only include these integrations in their respective configuration files:
+
+- For the browser runtime, add integrations to `sentry.client.config.ts`.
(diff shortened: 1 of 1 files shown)


mydea (approved):

AbhiPrasad (approved):

src/platforms/javascript/common/configuration/integrations/index.mdx
```diff
@@ -9,6 +9,22 @@ redirect_from:
...
+
+Depending on whether an integration enhances the functionality of a particular runtime, such as the BrowserTracing integration for the browser runtime or the RequestData integration for the Node.js runtime, you can only include these integrations in their respective configuration files:
+
+- For the browser runtime, add integrations to `sentry.client.config.ts`.
+- For Node.js, add integrations to `sentry.server.config.ts`.
+- For the Edge runtime, add integrations to `sentry.edge.config.ts`.
+
+</Alert>
+
+</PlatformSection>
```
shanamatthews, line 26: I think this would be better in regular text vs. a note

```suggestion
<PlatformSection supported={["javascript.nextjs"]}>

## Integration Usage in Next.js

Next.js can operate within three runtimes: the Node.js runtime, the browser runtime, or the Edge runtime. However, it's important to note that not all integrations are compatible with all of these runtimes.

Depending on whether an integration enhances the functionality of a particular runtime, such as the BrowserTracing integration for the browser runtime or the RequestData integration for the Node.js runtime, you can only include these integrations in their respective configuration files:

- For the browser runtime, add integrations to `sentry.client.config.ts`.
- For Node.js, add integrations to `sentry.server.config.ts`.
- For the Edge runtime, add integrations to `sentry.edge.config.ts`.

</PlatformSection>
```

shanamatthews (approved):
Good change! I think this would be even better if we added an easy way for users to see which runtime an integration is for at the top of each integration page.

src/platforms/javascript/common/configuration/integrations/index.mdx
lforst: > I think this would be even better if we added an easy way for users to see which runtime an integration is for at the top of each integration page.

I fully agree with this. However, this will turn into an `N times M` problem for which we don't have the time to maintain at the moment.

PR #7952, opened 2023-09-26
feat(replay): Update performance overhead

billyvg:
Updates performance overhead benchmarks + benchmark against rrweb2.

I didn't include rrweb1 benchmarks but here they are for reference:

Pasted Graphic 1

LCP is not very reliable, but here we can see total blocking time (tbt) has improved with rrweb2.


bruno-garcia (approved):
Nice!

mydea (approved):

Issue #7951, opened 2023-09-26
Session Replay on the Docs - sample rate and masking


bruno-garcia:
Increase replay sample rate
Possibly removing masking, except on anything loading data for logged users such as DSN and Auth

shanamatthews:
I'd love it if you fixed our masking! I've been meaning to, just haven't gotten around to it,

PR #7961, opened 2023-09-27
replay: mask only user data

bruno-garcia:
Resolves: https://github.com/getsentry/sentry-docs/issues/7951

Seems to work locally, lets check vercel:

image

image

Masking all code snippets which is more than we wants, but a step forward.

Thanks @billyvg for helping chase down the places to mask


shanamatthews (approved):
Nice! Thanks for this :D

PR #7964, opened 2023-09-27
Python: Updated the Intro in Getting Started

antonpirker:
Make it more calm in the sense of not having differently colored boxes and fonts in the first headline.


src/platforms/python/common/index.mdx
shanamatthews, line 1: I think it would be better to use the template index page: src/platforms/common/index.mdx

and change the intro note to your below note:

```
<Note>

If you are still using our legacy "Raven" SDK, you can access the <Link rel={`nofollow`} to={`/platforms/python/legacy-sdk/`}>legacy SDK documentation</Link>, until further notice. Migrating from older versions is documented [here](./migration/).

</Note>
```

If you want to remove the callout about frameworks you could also add Python to the ["noGuides" PlatformSection](https://github.com/getsentry/sentry-docs/blob/77246773d17e18718d04b54b48aa4c5fedc062fb/src/platforms/common/index.mdx#L5C1-L5C150) in src/platforms/common/index.mdx:

<PlatformSection noGuides notSupported={["android", "dart", "elixir", "flutter", "perl", "react-native", "unity", "unreal", "kotlin-multiplatform"]}>

shanamatthews: Or is there another reason why you prefer not to use the template?

antonpirker: ok, will use the template

antonpirker: one thing was also that I wanted to have the "On this page, we get you up and run..." as the first line and then a paragraph of explanation (without and alert box). 
So basically the primer below the "on this page". But this is not possible in the template...

shanamatthews: I think I don't follow, because wouldn't adding Python to the ["noGuides" PlatformSection](https://github.com/getsentry/sentry-docs/blob/77246773d17e18718d04b54b48aa4c5fedc062fb/src/platforms/common/index.mdx#L5C1-L5C150) in src/platforms/common/index.mdx remove the alert box you don't want? AFAICT the page would be basically identical to what this PR achieves if you add Python to that exclude.

shanamatthews (approved):
lgtm! Thanks for using the template, just trying to reduce docs complexity where possible.

PR #8006, opened 2023-09-27
fix(replay): Make "description" a required field

billyvg:
This disables the submit button until there is text in the description field

### DIFF
--- a/src/components/feedback/feedbackForm.tsx
+++ b/src/components/feedback/feedbackForm.tsx
@@ -40,7 +40,7 @@ export function FeedbackForm({onClose, onSubmit}: FeedbackFormProps) {
           type="text"
           id="sentry-feedback-name"
           name="name"
-          placeholder="Anonymous"
+          placeholder="Your name"
           defaultValue={user?.username}
         />
       </Label>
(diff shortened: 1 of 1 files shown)


src/components/feedback/feedbackForm.tsx
```diff
@@ -53,12 +55,17 @@ export function FeedbackForm({onClose, onSubmit}: FeedbackFormProps) {
...
       </Label>
       <Label htmlFor="sentry-feedback-comment">
-        Description
+        <div>
+          Description<Required>*</Required>
+        </div>
         <TextArea
           rows={5}
+          onChange={event => {
+            setHasDescription(!!event.target.value);
```
michellewzhang, line 64: is the `!!` intentional😮

billyvg: Yeah, it's a shorthand for casting the string to a boolean (rather than `Boolean()`)

bruno-garcia (approved):

src/components/feedback/feedbackForm.tsx
bruno-garcia: TIL

shanamatthews (approved):

Issue #7839, opened 2023-09-15
Replay: Document examples where `replay.start` can be used to sample sessions


bruno-garcia:
We often receive requests to sample sessions based on some conditions other than the already offered 'on error' or just by rolling a dice on a new session.

Common examples are:

* I'd like to always capture a Replay for a specific user
* If a user browsers certain page, start recording
* When a user submits feedback/ticket

These can be achieved today by calling `replay.start` programatically.

It's isn't always clear to developer how it can be achieved, who ask us for examples or best practices.

billyvg:
@bruno-garcia I added examples for the first 2 points, I don't think the third is very common

PR #8035, opened 2023-09-28
feat(replay): Add examples for custom sampling

billyvg:
Give more specific example scenarios for custom sampling.

Closes https://github.com/getsentry/sentry-docs/issues/7839
Closes https://github.com/getsentry/sentry-javascript/issues/9178

### DIFF
--- a/src/platforms/javascript/common/session-replay/understanding-sessions.mdx
+++ b/src/platforms/javascript/common/session-replay/understanding-sessions.mdx
@@ -143,7 +143,7 @@ if (loggedInUser.isEmployee) {
 
 ### Replays on Specific URLs
 
-You can also choose to record a replay every time a user lands on a specified URL. This can help developers debug problematic pages. To do this, use the [Navigation API](https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API), which can listen to navigation events for a specific URL, start the replay recording, and then stop it when the user leaves the URL.
+You can also choose to record a replay every time a user lands on a specified URL. This can help developers debug problematic pages. To do this, use the [Navigation API](https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API), which can listen to navigation events for a specific URL and start the replay recording.
 
 ```javascript
 const replay = new Sentry.Replay();
@@ -169,9 +169,6 @@ navigation.addEventListener("navigate", (event) => {
   if (url.pathname.startsWith("/buggy-page/")) {
     // User navigates to the buggy page, let's start recording
     replay.start();
-  } else if (window.location.pathname.startsWith("/buggy-page/")) {
-    // Stop replay when the user leaves the buggy page
-    replay.stop();
   }
 });
 ```
--- a/src/platforms/javascript/common/session-replay/understanding-sessions.mdx
+++ b/src/platforms/javascript/common/session-replay/understanding-sessions.mdx
@@ -97,3 +97,77 @@ await replay.flush();
 In `session` mode, this will upload any pending recording data to Sentry. In `buffer` mode, this will upload any pending recording data to Sentry and then continue recording, the same as when an error is sampled with `replaysOnErrorSampleRate`.
 
 Note that it's safe to call `flush()` at any time, even if Session Replay is stopped, in which case, it will do nothing.
+
+## Examples of Custom Sampling
+
+### Replays For Specific Users
+
+Here is an example of always recording a replay for a specific segment of your users. In this case, we would like to have error-based sampling for all users and we want to enable session-based replays for employees only. The `user` object has a field `isEmployee` to express if they are an employee or not.
+
+```javascript
+const replay = new Sentry.Replay();
+
+/**
+ * Initialize the Sentry SDK as normal.
+ * `replaysSessionSampleRate` is set to "0" so that we have manual control over session-based replays
+ * `replaysOnErrorSampleRate` is set to "1.0" to always record a replay when an error occurs
+ */
+Sentry.init({
+  dsn: "...",
+  replaysSessionSampleRate: 0,
+  replaysOnErrorSampleRate: 1.0,
+  integrations: [replay],
+});
+
+
+// ...
+// Sometime after your users are authenticated, check if they are an employee
+// If they are, we call `replay.start()` to start recording a replay for the employee
+// ...
+
+// Check if user is an employee, if so, always record a replay
+if (loggedInUser.isEmployee) {
+  // You can get a reference to the Sentry Replay integration like so:
+  const replay = Sentry.getCurrentHub().getIntegration(Sentry.Replay);
+  // or alternatively
+  // const replay = Sentry.getCurrentHub().getClient().getIntegrationById('Replay');
+
+  // Start the replay!
+  replay.start();
+}
+```
+
+### Replays on Specific URLs
+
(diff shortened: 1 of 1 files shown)


ryan953 (approved):
:shipit:

src/platforms/javascript/common/session-replay/understanding-sessions.mdx
```diff
@@ -93,3 +93,75 @@ await replay.flush();
 ```
 
+
+## Examples of Custom Sampling
+
+### Replays For Specific Users
```
lizokm, line 99:
```suggestion
### Replays for Specific Users
```

src/platforms/javascript/common/session-replay/understanding-sessions.mdx
```diff
@@ -93,3 +93,75 @@ await replay.flush();
...
+  // const replay = Sentry.getCurrentHub().getClient().getIntegrationById('Replay');
+
+  // Start the replay!
+  replay.start();
+}
+```
+
+### Replays on Specific URLs
+
+Here is an example of always recording a replay when a user lands on a specific URL. Maybe this specific page has been buggy and you would like to always record a replay when a user lands on it so that your developers have a better opportunity to debug. Using the [Navigation API](https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API) we can listen to navigation events for a specific URL, start the replay recording, and then stop the replay when the user leaves the URL.
```
lizokm, line 137:
```suggestion
You can also choose to record a replay every time a user lands on a specified URL. This can help developers debug problematic pages. To do this, use the [Navigation API](https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API), which can listen to navigation events for a specific URL, start the replay recording, and then stop it when the user leaves the URL.
```

lizokm (approved):
Made some language suggestions :)

src/platforms/javascript/common/session-replay/understanding-sessions.mdx
```diff
@@ -93,3 +93,75 @@ await replay.flush();
...
+
+// Check if user is an employee, if so, always record a replay
+if (loggedInUser.isEmployee) {
+  // You can get a reference to the Sentry Replay integration like so:
+  const replay = Sentry.getCurrentHub().getIntegration(Sentry.Replay);
+  // or alternatively
+  // const replay = Sentry.getCurrentHub().getClient().getIntegrationById('Replay');
+
+  // Start the replay!
+  replay.start();
```
mydea, line 131: This will not actually work! As since an error sample rate is set, replay is enabled and we'll run into this branch in `start()`:

```js
 if (this._isEnabled && this.recordingMode === 'buffer') {
      throw new Error('Replay buffering is in progress, call `flush()` to save the replay');
    }
```

I think you'd either need to do this:

```js
await replay.stop();
replay.start();
```

_or_ do this:

```js
replay.flush(); // <-- convert the error session to a regular session and start recording it
```

src/platforms/javascript/common/session-replay/understanding-sessions.mdx
```diff
@@ -93,3 +93,75 @@ await replay.flush();
...
+});
+
+// Listen to navigation events for when users land on our buggy URL, appropriately named: "/buggy-page/"
+// and then start recording a replay.
+navigation.addEventListener("navigate", (event) => {
+  const url = new URL(event.destination.url);
+
+  if (url.pathname.startsWith("/buggy-page/")) {
+    // User navigates to the buggy page, let's start recording
+    replay.start();
```
mydea, line 161: Same as above, you cannot start without stopping first! Or just call `flush()`, I guess.

src/platforms/javascript/common/session-replay/understanding-sessions.mdx
```diff
@@ -93,3 +93,82 @@ await replay.flush();
...
+  integrations: [replay],
+});
+
+// ...
+// Sometime after your users are authenticated, check if they are an employee
+// If they are, we call `replay.flush()` to either flush the replay as normal if session-based replay, or if it is buffering due to error sampling, the flush will send the first segment of the replay.
+
+// Since `replaysOnErrorSampleRate` is 1, we know that a replay has been buffered and `flush()` will flush the contents of that buffer. If both sample rates are < 1, then there is a chance that a replay has not been recording/buffering.
+// In this case, you can check that `replay.getReplayId()` returns a value, if it does, it means replay is active and you can call `replay.flush()`, other call `replay.start()` to start recording a new replay.
+// ...
```
billyvg, line 127: How does this sound @mydea ?

src/platforms/javascript/common/session-replay/understanding-sessions.mdx
```diff
@@ -93,3 +93,82 @@ await replay.flush();
...
+Sentry.init({
+  dsn: "...",
+  replaysSessionSampleRate: 0.1,
+  replaysOnErrorSampleRate: 1.0,
+  integrations: [replay],
+});
+
+// ...
+// Sometime after your users are authenticated, check if they are an employee
+// If they are, we call `replay.flush()` to either flush the replay as normal if session-based replay, or if it is buffering due to error sampling, the flush will send the first segment of the replay.
```
lizokm, line 123:
```suggestion
// After a user has been authenticated, check if they're an employee
// If they are, we'll call `replay.flush()` to either flush the replay as normal if it's a session-based replay, or if it's buffering due to error sampling, the flush will send the first segment of the replay.
```

lizokm (approved):
Just a small language tweak.

src/platforms/javascript/common/session-replay/understanding-sessions.mdx
mydea: it's not fully correct - the `replaysOnErrorSampleRate` only needs to be > 0. As long as this is > 0, you'll always have a buffer. So the instructions should be something like:

As long as `replaysOnErrorSampleRate` is greater than 0, we will always buffer the replay in the background. 
When you then call `flush()`, we will ensure to always send whatever has been buffered before.

It is also safe to call `flush()` in any time, really. If the replay is not enabled, it will simply do nothing. If the replay is currently in `session` mode, it will send what has been buffered, and if it is in `buffer` mode, it will also send what has been buffered _and_ convert the session to `session` mode, like when an error occurred.

src/platforms/javascript/common/session-replay/understanding-sessions.mdx
```diff
@@ -97,3 +97,83 @@ await replay.flush();
...
+
+// Check if user is an employee, if so, always record a replay
+if (loggedInUser.isEmployee) {
+  // You can get a reference to the Sentry Replay integration like so:
+  const replay = Sentry.getCurrentHub().getIntegration(Sentry.Replay);
+  // or alternatively
+  // const replay = Sentry.getCurrentHub().getClient().getIntegrationById('Replay');
+
+  // `replaysOnErrorSampleRate` is > 0, so a replay will always be buffered while it waits for an error to happen.
+  // `flush()` will create a new replay if it is in a buffering state, otherwise will flush the recording segment if a replay is currently in progress.
```
mydea, line 141:
```suggestion
```

I think this can be removed now, as this is described above?

src/platforms/javascript/common/session-replay/understanding-sessions.mdx
```diff
@@ -97,3 +97,83 @@ await replay.flush();
...
+// and then start recording a replay.
+navigation.addEventListener("navigate", (event) => {
+  const url = new URL(event.destination.url);
+
+  if (url.pathname.startsWith("/buggy-page/")) {
+    // User navigates to the buggy page, let's start recording
+    replay.start();
+  } else if (window.location.pathname.startsWith("/buggy-page/")) {
+    // Stop replay when the user leaves the buggy page
+    replay.stop();
```
mydea, line 176: I actually would not recommend that - if we started (not buffered!) there is no need to stop, IMHO it is better to just leave it recording in that case (you can disregard what was captured after the user moved away if you want).

Otherwise you need to implement awaiting on the `stop()`, ensuring stop has finished before you call `start` again etc.

Issue #7957, opened 2023-09-26
Adjustments to the OpenTelemetry Support for Next.js docs


Kobby-Bawuah:
### Core or SDK?

Core Sentry product

### Which part? Which one?

Next.js

### Description

When trying to get Otel working for Next.js, there are a few road blocks which are not mentioned by the docs. 

### Suggested Solution

There are a few things that need to be updated on the next.js page seen [here](https://docs.sentry.io/platforms/javascript/guides/nextjs/performance/instrumentation/opentelemetry/?original_referrer=https%3A%2F%2Fwww.google.com%2F) to get Otel running. 

First: 

It should be noted that the SDK needs to be on one of the latest versions. 
Users can upgrade `"@next/font"` and `"next"`in their package.json to a minimum version of 13.5.3. They will also need to add the following config 
```
experimental: {
    instrumentationHook: true,
  },
```
to their next.config.js. 

![image (5)](https://github.com/getsentry/sentry-docs/assets/89524511/474dd552-506a-4af2-8ef1-b50809370333)

Read more about the reason for the upgrade [here](https://nextjs.org/docs/app/building-your-application/optimizing/instrumentation). 

Second: 

When is comes to [this section](https://docs.sentry.io/platforms/javascript/guides/nextjs/performance/instrumentation/opentelemetry/?original_referrer=https%3A%2F%2Fwww.google.com%2F), SentrySpanProcessor() and SentryPropagator() were not defined in the code snippet for `instrumentation.node.js`. The following import is needed: 
```
import {SentrySpanProcessor, SentryPropagator} from '@sentry/opentelemetry-node'
```

PR #8080, opened 2023-09-29
Add minimum version to otel node docs and fix snippets

lforst:
Fixes https://github.com/getsentry/sentry-docs/issues/7957


AbhiPrasad (commented):
I think we should add a note about `instrumentationHook: true`, it's easy to miss otherwise

shanamatthews (approved):
agree with abhi, otherwise lgtm!

PR #8092, opened 2023-10-02
fix: Exclude wizard pages from VScode search

HazAT:
We should no longer show the wizard content in the search - so people stop editing these pages


ArthurKnaus (approved):
makes sense 👍

PR #8095, opened 2023-10-02
Added mention of requests

antonpirker:
Small fix to mention that `Requests` instrumentation is done with the default integrations.


antonpirker:
As this is such a small change, I took the liberty to merge it right away.

PR #8109, opened 2023-10-03
Add stack trace linking support for azure and bitbucket

vivianyentran:
- Add stack trace linking support for azure and bitbucket


Dhrumil-Sentry:
@vivianyentran we can probably avoid the images for the stack trace links, we don't do that for GitHub or GitLab today

Dhrumil-Sentry (approved):

PR #8119, opened 2023-10-04
Link to vite plugin options v0.6.1

lforst:
We linked to the `latest` version which is incorrect in the context of sveltekit.


Lms24 (approved):
Good catch, thanks!

PR #8121, opened 2023-10-04
add React Native to platforms with screenshots

kahest:
adds React Native to list of platforms supporting screenshots


vivianyentran (approved):
thanks for adding!

PR #8140, opened 2023-10-05
fix(replay): Update manual starting & flushing docs for replay

mydea:
This updates the session docs for replay:

1. Fix a wrong link
2. Mention how to access the replay instance from other parts of your code
3. Mention that calling `flush` is safe in any circumstance.


michellewzhang (approved):

src/platforms/javascript/common/session-replay/understanding-sessions.mdx
```diff
@@ -93,3 +95,5 @@ await replay.flush();
 ```
 
+
+Note that it is safe to call `flush()` any time, even if Session Replay is stopped. In that case, it will simply do nothing.
```
lizokm, line 99:
```suggestion
Note that it's safe to call `flush()` at any time, even if Session Replay is stopped, in which case, it will do nothing.
```

lizokm (approved):
LGTM

PR #8144, opened 2023-10-05
feat: Add express & fastify tunnel examples

mydea:
This adds docs for setting up a tunnel with express or fastify, two main JS frameworks.

If we prefer to keep this "small", we could also only add the express example.

ref https://github.com/getsentry/sentry-docs/pull/8124

### DIFF
--- a/src/platforms/javascript/common/troubleshooting/index.mdx
+++ b/src/platforms/javascript/common/troubleshooting/index.mdx
@@ -103,11 +103,11 @@ Sentry.init({
 });
 ```
 
-Once configured, all events will be sent to the `/tunnel` endpoint. This solution, however, requires an additional configuration on the server, as the events now need to be parsed and redirected to Sentry.
+Once configured, all events will be sent to the `/tunnel` endpoint. This will require an additional configuration on the server because the events will need to be parsed and redirected to Sentry.
 
 ### Using a Tunneling Server
 
-Here's an example of how to set up an endpoint on your server that tunnels events to Sentry:
+Here's an example of how to set up an endpoint on your server that will tunnel events to Sentry:
 
 ```js {tabTitle:Express}
 // Change this to suit your needs
--- a/src/platforms/javascript/common/troubleshooting/index.mdx
+++ b/src/platforms/javascript/common/troubleshooting/index.mdx
@@ -103,7 +103,93 @@ Sentry.init({
 });
 ```
 
-Once configured, all events will be sent to the `/tunnel` endpoint. This solution, however, requires an additional configuration on the server, as the events now need to be parsed and redirected to Sentry. Here's an example for your server component:
+Once configured, all events will be sent to the `/tunnel` endpoint. This solution, however, requires an additional configuration on the server, as the events now need to be parsed and redirected to Sentry.
+
+### Using a Tunneling Server
+
+Here's an example for your server component:
+
+```js {tabTitle:Express}
+// Change this to suit your needs
+const SENTRY_HOST = "oXXXXXX.ingest.sentry.io";
+const SENTRY_KNOWN_PROJECTS = ["123456"];
+
+app.post("/bugs", async (req, res) => {
+  try {
+    const envelope = req.body;
+
+    const piece = envelope.toString().split("\n")[0];
+    const header = JSON.parse(piece);
+
+    const dsn = new URL(header.dsn);
+    if (dsn.hostname !== SENTRY_HOST) {
+      throw new Error(`Invalid Sentry host: ${dsn.hostname}`);
+    }
+
+    const projectId = dsn.pathname.substring(1);
+    if (!SENTRY_KNOWN_PROJECTS.includes(projectId)) {
+      throw new Error(`Invalid Project ID: ${projectId}`);
+    }
+
+    const url = `https://${SENTRY_HOST}/api/${projectId}/envelope/`;
+    await fetch(url, {
+      method: "POST",
+      body: envelope,
+      headers: {
+        "Content-Type": "application/x-sentry-envelope",
+      },
+    });
+  } catch (error) {
+    return res.sendStatus(400).send({ message: error.message });
+  }
+  return res.sendStatus(204);
+});
+```
+
+```js {tabTitle:Fastify}
+// Change this to suit your needs
+const SENTRY_HOST = "oXXXXXX.ingest.sentry.io";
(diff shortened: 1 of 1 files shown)


src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -103,7 +103,93 @@ Sentry.init({
...
+### Using a Tunneling Server
+
+Here's an example for your server component:
+
+```js {tabTitle:Express}
+// Change this to suit your needs
+const SENTRY_HOST = "oXXXXXX.ingest.sentry.io";
+const SENTRY_KNOWN_PROJECTS = ["123456"];
+
+app.post("/bugs", async (req, res) => {
```
lforst, line 117: h: You cannot have handlers in express be async. This will mess up error handling.

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -103,7 +103,93 @@ Sentry.init({
...
+  try {
+    const envelope = req.body;
+
+    const piece = envelope.toString().split("\n")[0];
+    const header = JSON.parse(piece);
+
+    const dsn = new URL(header.dsn);
+    if (dsn.hostname !== SENTRY_HOST) {
```
lforst, line 125: m: I assume we do this so that the endpoint can't be abused. We should add a comment to describe this because strictly speaking `SENTRY_HOST` is not necessary.

Ditto for `SENTRY_KNOWN_PROJECTS`.

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -103,7 +103,93 @@ Sentry.init({
 });
 ```
 
-Once configured, all events will be sent to the `/tunnel` endpoint. This solution, however, requires an additional configuration on the server, as the events now need to be parsed and redirected to Sentry. Here's an example for your server component:
+Once configured, all events will be sent to the `/tunnel` endpoint. This solution, however, requires an additional configuration on the server, as the events now need to be parsed and redirected to Sentry.
+
+### Using a Tunneling Server
+
+Here's an example for your server component:
```
lforst, line 110:
```suggestion
Here's an example of how to set up an endpoint on your server that tunnels events to Sentry:
```

"Server component" is becoming a bit of an ambiguous term so we should avoid it in the wrong context.

src/platforms/javascript/common/troubleshooting/index.mdx
mydea: hmm, this is based on the existing tunnel examples, where they have set this up in a similar way too 🤔 not sure if it helps to explain this more, or just bloats this up? IMHO we _can_ also just leave this away if we want to keep this as simple as possible, not sure how important this check is.

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -103,7 +103,99 @@ Sentry.init({
 });
 ```
 
-Once configured, all events will be sent to the `/tunnel` endpoint. This solution, however, requires an additional configuration on the server, as the events now need to be parsed and redirected to Sentry. Here's an example for your server component:
+Once configured, all events will be sent to the `/tunnel` endpoint. This solution, however, requires an additional configuration on the server, as the events now need to be parsed and redirected to Sentry.
+
```
lizokm, line 107:
```suggestion
Once configured, all events will be sent to the `/tunnel` endpoint. This will require an additional configuration on the server because the events will need to be parsed and redirected to Sentry.

```

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -103,7 +103,99 @@ Sentry.init({
...
+### Using a Tunneling Server
+
+Here's an example of how to set up an endpoint on your server that tunnels events to Sentry:
```
lizokm, line 110:
```suggestion
Here's an example of how to set up an endpoint on your server that will tunnel events to Sentry:
```

lizokm (approved):
A few small tweaks, but this LGTM

src/platforms/javascript/common/troubleshooting/index.mdx
lforst: No hard feelings either way. Let me downgrade this to logaf l. I think it makes sense for users to be aware of abuse potential. We can just leave it as is (without explanation) and add it if we see a lot of people asking about it, which is not gonna happen anyhow.

Removing it would probably be handing people a footgun which I don't particularly like.

lforst (approved):

src/platforms/javascript/common/troubleshooting/index.mdx
lforst: Okay nevermind let's keep this for sure!

GeekCornerGH (approved):

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -103,7 +103,99 @@ Sentry.init({
...
+### Using a Tunneling Server
+
+Here's an example of how to set up an endpoint on your server that will tunnel events to Sentry:
+
+```js {tabTitle:Express}
+// Change this to suit your needs
+const SENTRY_HOST = "oXXXXXX.ingest.sentry.io";
+const SENTRY_KNOWN_PROJECTS = ["123456"];
+
+app.post("/bugs", (req, res) => {
```
GeekCornerGH, line 117:
```suggestion
app.post("/tunnel", (req, res) => {
```

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -103,7 +103,99 @@ Sentry.init({
...
+
+// Handle replay packets where no content-type header is available
+fastify.addContentTypeParser(
+  "*",
+  { parseAs: "buffer" },
+  function (req, body, done) {
+    done(null, body);
+  }
+);
+fastify.post("/bugs", async (request, reply) => {
```
GeekCornerGH, line 168:
```suggestion
fastify.post("/tunnel", async (request, reply) => {
```

GeekCornerGH (changes requested):
Figured out you forgot to change the endpoint per the line nr 106

Lms24:
Wondering if there's a reason this PR got stale. Do you still remember @mydea? 

I think it'd be great if we somewhere had an up to date example how to create the proxy in JS.

PR #8145, opened 2023-10-05
fix(perf): Relax `http.server` requirement for database view

gggritso:
As of https://github.com/getsentry/sentry/pull/57501 we show queries from all transactions, not just endpoints! Also, removes a note about database ops. That's covered in more details in the `op` requirements listed below, and isn't very correct anyway.


AbhiPrasad (approved):

vivianyentran (approved):
lgtm, thanks for updating!

PR #8149, opened 2023-10-05
fix(build): Add sharp resolution to fix build on Apple silicon

malwilley:
Without this I wasn't able to build docs on my M1 macbook.

https://sharp.pixelplumbing.com/install#apple-m1

Looks like this was previously added in https://github.com/getsentry/sentry-docs/pull/5380 and removed in https://github.com/getsentry/sentry-docs/pull/6783


malwilley:
Blowing away node_modules fixed this - no need to add back the resolution anymore

PR #8159, opened 2023-10-06
improved project selector doc updates

vivianyentran:
## Description of changes

remove mention of lock function and add info on new project selector


scttcper (approved):
thanks for updating this, should the old image get deleted?

PR #8195, opened 2023-10-13
Pre SDK power update (Python)

stephanie-anderson:


shanamatthews (approved):
👍

PR #8196, opened 2023-10-13
Pre SDK power update (Android)

stephanie-anderson:
  - moved all integrations in their own subfolder `/integrations/<integration>` in order to improve the overview of all integrations (reduce noise by potential images used on integration pages)
  - added redirects for all integrations


romtsn:
I believe we should also add redirects for those pages that are moved:
```
redirect_from:
  - /platforms/android/configuration/integrations/<integration_name>/
```

stephanie-anderson:
> I believe we should also add redirects for those pages that are moved, to not break external links:
> 
> ```
> redirect_from:
>   - /platforms/android/configuration/integrations/<integration_name>/
> ```

Thank you @romtsn! I added redirects as suggested :) I also moved the integrations in separate subfolders, to reduce the noise from images in the same folder

Screenshot 2023-10-13 at 14 02 42

<br>

Screenshot 2023-10-13 at 14 03 06

shanamatthews (approved):
Looks good, thanks for adding the redirects!

PR #8198, opened 2023-10-13
Pre SDK power update (Dart)

stephanie-anderson:
  - added redirects from old location


kahest (approved):
LGTM 🚀 Thank you @stephanie-anderson!

PR #8204, opened 2023-10-13
Pre SDK power update (Kotlin + Kotlin Multiplatform)

stephanie-anderson:
- merged `config.yml` with `index.mdx`
- move files from `common` to parent directory, since no guides exist


shanamatthews (approved):
👍

PR #8208, opened 2023-10-13
Pre SDK power update (Unreal Engine)

stephanie-anderson:
- group files and screenshots in folders for better overview


shanamatthews (approved):
👍

PR #8209, opened 2023-10-13
Pre SDK power update (Unity)

stephanie-anderson:
- group files in folders for better readabilty
- remove unnecessary `/common` folder, since Unity has no guides


shanamatthews (approved):
👍

PR #8211, opened 2023-10-13
Pre SDK power update (enriching events)

stephanie-anderson:
- moved everything under `/enriching-events/` into separate subfolders for better readabilty, e.g.
  - `src/platforms/common/enriching-events/context.mdx` to
  - `src/platforms/common/enriching-events/context/index.mdx` and grouped it with its images
- renamed `/enriching-events/identify-user` to `/enriching-events/user`
- moved `include/enriching-events/` to `include/usage/events/enriching-events/`
- removed the sidebar_order from all files in `enriching-events` so that they show up alphabetically in the left sidebar
- simplified the headlines for all subpages of `enriching-events`


shanamatthews (changes requested):
It seems that your new redirects aren't working correctly. For example this path "/platforms/go/enriching-events/attachments/" doesn't work on this build.

Also, do we need the "Events" page? Is the plan to put more subpages there, or can "Enriching Events" be promoted up a level?

stephanie-anderson:
> It seems that your new redirects aren't working correctly. For example this path "/platforms/go/enriching-events/attachments/" doesn't work on this build.
> 
> Also, do we need the "Events" page? Is the plan to put more subpages there, or can "Enriching Events" be promoted up a level?

Hey Shana, I will take a look at the redirects as soon as possible - thanks for pointing it out! Regarding the Events, yes the `Enriching Events` will be just one of many items under `Events`:
Screenshot 2023-10-17 at 12 53 58

Moving `Enriching Events` is such a huge change, that I didn't want to mingle it with other changes

mydea:
I wonder if `Events` should not be a top-level options (next to Usage, not under it) - feels like a lot of steps right now, and honestly I did not go looking for this inside of Usage 😅 IMHO if we have a bunch of items under this, it's warranted to have this on top, as it's also something people are likely to go looking for...?

stephanie-anderson:
> I wonder if `Events` should not be a top-level options (next to Usage, not under it) - feels like a lot of steps right now, and honestly I did not go looking for this inside of Usage 😅 IMHO if we have a bunch of items under this, it's warranted to have this on top, as it's also something people are likely to go looking for...?

@mydea I get that the content is really important, but is it not about the `usage` of events and what one can do with them? If we imagine we'd point users to `enriching events` whenever it makes sense, would you still be against moving it to the usage context? I guess the need for putting it in the first level of navigation is because you think users will otherwise not find the content? I'd really really like to get a consensus on that change before it goes out. Happy to also talk next week so it does not need to be async

vercel.json
shanamatthews, line 1: I can't see from a glance why /enriching-events/identify-user/ doesn't work here :/

src/platforms/common/usage/sdk-fingerprinting.mdx
```diff
@@ -2,7 +2,7 @@
 title: "SDK Fingerprinting"
 sidebar_order: 30
 redirect_from:
-  - ../../../enriching-events/context/fingerprint-override/
+  - ../../../usage/events/enriching-events/context/fingerprint-override/
```
shanamatthews, line 5: I think this change was a mistake, although it looks like even the existing redirect here doesn't work

```suggestion
  - ../../../enriching-events/context/fingerprint-override/
```

shanamatthews (changes requested):
Just a few small issues left that I found!!

PR #8236, opened 2023-10-13
feat(suspect-commits): Add docs for setting up Java code mappings

malwilley:
Ref https://github.com/getsentry/sentry/issues/54764

## Pre-merge checklist

## Description of changes

Adds a new section to the Java Source Context docs: https://docs.sentry.io/platforms/java/source-context/ 

This section is meant to explain the specifics around setting up code mappings for Java, which are different to most other platforms. I've added a link from the common docs on code mappings so that hopefully users on these platforms will now have some way of knowing how to set these up. In the future we can link directly to this page from the app, but the first step is documenting it at all.


src/platforms/java/common/source-context.mdx
```diff
@@ -5,3 +5,17 @@ description: "Learn about showing your source code as part of stack traces."
...
+
+First, navigate to a stack trace that you wish to map. Find an In App frame, which is denoted by a bubble on the right side of the frame. The `module` will be shown as the first piece of text at the left hand side of the frame header (marked as **1** in the below screenshot). In this example, it is `com.example.vu.android.MainActivity`. Hovering over the module will show the absolute path to the file (marked as **2** in the below screenshot). In this example, it is `MainActivity.java`.
+
+![Highlighting where in the stack trace to find the module and absolute path](java-code-mappings-stacktrace.png)
+
+Next, calculate the derived file path. Take the frame `module` and replace all `.` characters with `/`. Then, replace the last item (`MainActivity` in this example) with the absolute path found earlier (`MainActivity.java`). In this example, the derived file path is `com/example/vu/android/MainActivity.java`.
+
+Finally, use the derived file path to determine the **Stack Trace Root** and **Source Code Root** using the steps detailed in the [Stack Trace Root and Source Code Root](/product/issues/suspect-commits/#stack-trace-root-and-source-code-root) docs.
```
shanamatthews, line 21: Instead of bailing back to the product docs, could we just include the instructions for setting stack trace root and source code root here? Afaict it wouldn't be that much extra text. We can still link back to the product docs as a "for more information..."

yarn.lock
shanamatthews, line 1: what's going on with this? not up to date with master?

shanamatthews (commented):
Overall this is good! My main feedback is to include the final instructions for setting stack trace root and source code root on this page, rather than linking back to product docs.

yarn.lock
malwilley: Oops! Mistakenly added

src/platforms/java/common/source-context.mdx
malwilley: Makes sense! Will add

PR #8248, opened 2023-10-16
feat: Update loader docs to guard against unexisting `Sentry.onLoad()`

mydea:
It is possible that the loader script fails to load (e.g. if it is adblocked or something like this), so to be safe we should recommend to guard against `Sentry` not existing before calling `Sentry.onLoad()`.

### DIFF
--- a/gatsby-ssr.tsx
+++ b/gatsby-ssr.tsx
@@ -53,7 +53,7 @@ function SentryLoaderConfig() {
       key="sentry-loader-config"
       dangerouslySetInnerHTML={{
         __html: `
-Sentry.onLoad(function() {
+  window.Sentry && Sentry.onLoad(function() {
   Sentry.init({
     integrations: [
       new Sentry.Replay({
--- a/src/platform-includes/configuration/capture-console/javascript.mdx
+++ b/src/platform-includes/configuration/capture-console/javascript.mdx
@@ -22,11 +22,13 @@ Sentry.init({
 ></script>
 
 <script>
-  Sentry.onLoad(function () {
-    Sentry.init({
-      integrations: [new Sentry.Integrations.CaptureConsole()],
+  // Check for existence of Sentry in case Ad-blockers block the Sentry Loader Script
+  window.Sentry &&
+    Sentry.onLoad(function () {
+      Sentry.init({
+        integrations: [new Sentry.Integrations.CaptureConsole()],
+      });
     });
-  });
 </script>
 ```
 
--- a/src/platforms/javascript/common/install/loader.mdx
+++ b/src/platforms/javascript/common/install/loader.mdx
@@ -92,15 +92,17 @@ The loader script always includes a call to `Sentry.init` with a default configu
   crossorigin="anonymous"
 ></script>
 <script>
-  Sentry.onLoad(function() { ... });
+  // Check for existence of Sentry in case Ad-blockers block the Sentry Loader Script
+  window.Sentry && Sentry.onLoad(function() { ... });
 </script>
 ```
 
-`Sentry.onLoad` is a function that only the loader provides, and - as the name suggests - it sets a function to be run once the full SDK has been loaded. In that function, you can configure your SDK exactly the way you would were you using the CDN, with one difference: your `Sentry.init` call doesn't need to include your DSN, since it's already been set.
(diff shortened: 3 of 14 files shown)


gatsby-ssr.tsx
```diff
@@ -53,7 +53,7 @@ function SentryLoaderConfig() {
       key="sentry-loader-config"
       dangerouslySetInnerHTML={{
         __html: `
-Sentry.onLoad(function() {
+Sentry && Sentry.onLoad(function() {
```
lforst, line 56: This will throw if `Sentry` is not defined in the global scope yet. I think we either need to do this:

```suggestion
window.Sentry && Sentry.onLoad(function() {
```

or this:

```suggestion
typeof Sentry !== 'undefined' && Sentry.onLoad(function() {
```

src/platform-includes/configuration/capture-console/javascript.mdx
```diff
@@ -22,11 +22,12 @@ Sentry.init({
 ></script>
 
 <script>
-  Sentry.onLoad(function () {
-    Sentry.init({
-      integrations: [new Sentry.Integrations.CaptureConsole()],
+  Sentry &&
```
lforst, line 25: I feel like we should also say in a comment or something why we first check for `Sentry`'s existence:

eg `Check for existence of Sentry in case Ad-blockers block the Sentry Loader Script`

gatsby-ssr.tsx
mydea: good point! I'll update it to `window.Sentry` (a bit less verbose)

src/platform-includes/configuration/capture-console/javascript.mdx
mydea: Hmm I was thinking about this, my thought was it is a lot of repetition for all the integrations - but better safe than sorry I guess! I'll add it.

src/platforms/javascript/common/install/loader.mdx
```diff
@@ -92,15 +92,17 @@ The loader script always includes a call to `Sentry.init` with a default configu
...
 ></script>
 <script>
-  Sentry.onLoad(function() { ... });
+  // Check for existence of Sentry in case Ad-blockers block the Sentry Loader Script
+  window.Sentry && Sentry.onLoad(function() { ... });
 </script>
 ```
 
-`Sentry.onLoad` is a function that only the loader provides, and - as the name suggests - it sets a function to be run once the full SDK has been loaded. In that function, you can configure your SDK exactly the way you would were you using the CDN, with one difference: your `Sentry.init` call doesn't need to include your DSN, since it's already been set.
+`Sentry.onLoad` is a function that only the loader provides, and - as the name suggests - it sets a function to be run once the full SDK has been loaded. In that function, you can configure your SDK exactly the way you would were you using the CDN, with one difference: your `Sentry.init` call doesn't need to include your DSN, since it's already been set. Note that we check if `Sentry` exists before we call `Sentry.onLoad` to guard against the case that the loader script failed to load.
```
shanamatthews, line 100: Slight rephrase and I suggest adding some context around ad blocking.


```suggestion
`Sentry.onLoad` is a function that only the loader provides, and - as the name suggests - it sets a function to be run once the full SDK has been loaded. In that function, you can configure your SDK exactly the way you would were you using the CDN, with one difference: your `Sentry.init` call doesn't need to include your DSN, since it's already been set. We recommend checking if `Sentry` exists before calling `Sentry.onLoad` in case the loader script fails to load. One common reason for this is when Sentry is blocked by an ad-blocker. See our docs on [Dealing with Ad-Blockers](https://docs.sentry.io/platforms/javascript/troubleshooting/#dealing-with-ad-blockers) for more information.
```

shanamatthews (approved):

src/platforms/javascript/common/install/loader.mdx
mydea: Actually the ad-blocker docs will not apply to this loader usage, as the loader is always loaded from Sentry. So probably this last sentence should not be included...?

lforst (approved):

src/platforms/javascript/common/install/loader.mdx
shanamatthews: Oh, sorry about that! I've edited the suggestion.

PR #8262, opened 2023-10-17
Update health-check inbound filter documentation.

RaduW:
This updates the patterns used by the health-check inbound filter.


RaduW:
messed the branch will re-create PR

PR #8263, opened 2023-10-17
Update health-check inbound filter documentation to the new glob patterns

RaduW:
This updates the patterns used by the health-check inbound filter.

See https://github.com/getsentry/sentry/pull/58223

Resolves https://github.com/getsentry/sentry-docs/issues/8260


shanamatthews (approved):
This is good from a docs perspective, but please tag another dev if you need an accuracy review :D

RaduW:
@shanamatthews  thanks for the review.
I have implemented the change myself, I have double checked that I used the exact same globs as in code.

PR #8285, opened 2023-10-18
Fix rust automatic-instrumentation.mdx

danprueitt:
## Description of changes

Update the instrumentation docs to be inline with the sentry-tower crate.

The Hub layer should always be before the Http layer as seen here:

https://docs.rs/sentry-tower/latest/sentry_tower/#usage-with-tower-http

## Legal Boilerplate

## Extra resources

### DIFF
--- a/src/platforms/rust/common/performance/instrumentation/automatic-instrumentation.mdx
+++ b/src/platforms/rust/common/performance/instrumentation/automatic-instrumentation.mdx
@@ -50,13 +50,13 @@ fn main_span2() {
 
 The Sentry SDK offers an integration for the `tower` ecosystem which can automatically continue a trace from an incoming HTTP request.
 
+When combining both layers, take care of the ordering of both. For example with tower::ServiceBuilder, always define the Hub layer before the Http one, like so:
+
 ```rust
 use sentry_tower::{NewSentryLayer, SentryHttpLayer};
 use tower::ServiceBuilder;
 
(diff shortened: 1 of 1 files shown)


src/platforms/rust/common/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -50,13 +50,13 @@ fn main_span2() {
 
 The Sentry SDK offers an integration for the `tower` ecosystem which can automatically continue a trace from an incoming HTTP request.
 
+When combining both layers, take care of the ordering of both. For example with tower::ServiceBuilder, always define the Hub layer before the Http one, like so:
```
shanamatthews, line 53: A small suggestion to simplify wording. Could we also add a sentence explaining _why_ users need to define the hub layer first? Unless it's obvious, but it's not to me :D

```suggestion
When combining both layers, order matters. For example, with tower::ServiceBuilder, you must define the Hub layer before the Http one, like so:
```

shanamatthews:
Thanks so much for your contribution @danprueitt ! This is a great change.

src/platforms/rust/common/performance/instrumentation/automatic-instrumentation.mdx
danprueitt: I'm not sure about the reasoning. @shanamatthews

Swatinem (approved):

PR #8299, opened 2023-10-18
chore(gradle): Remove buildscript blocks

romtsn:
Removed `buildscript {}` blocks as they are added by default when creating a project in IJ/AS or calling `gradle init` anyway. It's also practically impossible to use any dependency without declaring `mavenCentral` so we are good to assume that probably all of our customers have it already


lizokm (approved):
LGTM

PR #8306, opened 2023-10-20
docs(relay): Update directory creation when enabling spooling

iker-barriocanal:
See https://github.com/getsentry/relay/pull/2636.

### DIFF
--- a/src/docs/product/relay/options.mdx
+++ b/src/docs/product/relay/options.mdx
@@ -62,13 +62,13 @@ These settings fine tune caching of project states.
 
 ## Spooling
 
-| key                             | type    | value            | description                                                                                                                                                                                                             |
-| ------------------------------- | ------- | ---------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| spool.envelopes.path            | String  | default: `null`  | The path to the spool file used to save envelopes to disk. If provided, Relay will create the directory structure and enable on-disk spooling. Note that `~` is interpreted as a literal and not as the home directory. |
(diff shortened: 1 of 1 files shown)


src/docs/product/relay/options.mdx
```diff
@@ -62,13 +62,13 @@ These settings fine tune caching of project states.
...
-| spool.envelopes.min_connections | Integer | default: `10`    | The minimum number of connections in the pool to manage on-disk spooling.                                                    |
+| key                             | type    | value            | description                                                                                                                                                                                                             |
+| ------------------------------- | ------- | ---------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| spool.envelopes.path            | String  | default: `null`  | The path to the spool file used to save envelopes to disk. If provided, Relay will create the directory structure and enable on-disk spooling. Note that `~` is interpreted as a literal and not as the home directory. |
```
olksdr, line 67:
```suggestion
| spool.envelopes.path            | String  | default: `null`  | The path to the spool file used to save envelopes to disk. If provided, enables on-disk spooling. Relay will create the directory structure and enable on-disk spooling. Note that `~` is interpreted as a literal and not as the home directory. |
```

olksdr (changes requested):
I still think we should keep the sentence as before and just add the part about the folders autocreate. See the suggested changes.

src/docs/product/relay/options.mdx
iker-barriocanal: Reworded to something similar, let me know what you think. I don't like duplicating the "enable on-disk" part.

vivianyentran (approved):
lgtm

PR #8267, opened 2023-10-17
feat(js): Add Documentation for Sentry Astro SDK

Lms24:
## Pre-merge checklist

*If you work at Sentry, you're able to merge your own PR without review, but please don't unless there's a good reason.*

- [x] Checked Vercel preview for correctness, including links
- [x] PR was reviewed and approved by any necessary SMEs
- [x] PR was reviewed and approved by a member of the [Sentry docs team](https://github.com/orgs/getsentry/teams/docs)

## Description of changes

This PR adds SDK docs for the new Sentry Astro SDK. It's still in Alpha which we clearly point out in the new docs.

This PR includes adjustments for Astro in the following pages:
* Getting Started shows the semi-automatic setup via the `astro` CLI
* Added "Manual Setup" page similarly, to SvelteKit, Remix and NextJS with all possible configuration options
* Adjusted source map docs to not display the full range of source maps upload material. The Astro SDK ships with the Vite plugin, so there shouldn't be too much need for other resources than troubleshooting. 
* Added Astro specific setup for Replay (This is the first SDK that actually enables Replay by default. However, there are two ways how users can manually customize or add Replay).
* Added various platform includes for performance monitoring snippets

ref: https://github.com/getsentry/sentry-javascript/issues/9182

PR #8312, opened 2023-10-23
fix: Remove glob syntax from `tracePropagationTargets` example

Lms24:
As discovered by @lforst in https://github.com/getsentry/sentry-docs/pull/8267#discussion_r1368295746, we showed an incorrect example for `tracePropagationTargets` values in the "Usage->Distributed Tracing" page ([example](https://docs.sentry.io/platforms/javascript/guides/sveltekit/usage/distributed-tracing/#how-to-use-distributed-tracing)). TPT values can only be String.contains matches or regex matches but glob patterns in strings like `"https://.*.otherservice.org/.*"` are not supported. 

This PR removes the glob pattern and adds a regex example to show both types of TPT values. 
(Adjusted the one for Astro in #8267)

Note: The same value is used in Flutter and iOS docs but I'm not yet sure if these SDKs support globs or not. Asked in Slack for clarification.

### DIFF
--- a/src/platform-includes/distributed-tracing/how-to-use/apple.mdx
+++ b/src/platform-includes/distributed-tracing/how-to-use/apple.mdx
@@ -7,7 +7,7 @@ SentrySDK.start { options in
     options.dsn = "___PUBLIC_DSN___"
     options.tracePropagationTargets = [
         "https://myproject.org",
-        "https://.*.otherservice.org/.*",
+        "https://api.otherservice.org/",
     ]
 }
 ```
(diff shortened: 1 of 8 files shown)


lforst (approved):
oh didnt realize we had this more often

krystofwoldrich (approved):
I think we can update Apple Regex in a follow-up PR.

Lms24:
> I think we can update Apple Regex in a follow-up PR.

Sounds good to me, thanks!

PR #8313, opened 2023-10-23
feat: Add docs for replay tree shaking flags

mydea:
This adds the new tree shaking flags for Replay, as well as the new bundler plugin configs for bundle size optimizations.

I also excluded the webpack/rollup examples for sveltekit and remix as well (??) - not sure if we should show something (?) else for these platforms?

### DIFF
--- a/src/platforms/javascript/common/configuration/tree-shaking/index.mdx
+++ b/src/platforms/javascript/common/configuration/tree-shaking/index.mdx
@@ -36,6 +36,8 @@ To make optional code eligible for tree shaking, you can replace various flags i
 : Replacing this flag with `false` will tree shake all code in the SDK that is related to performance monitoring.
 **Attention:** `__SENTRY_TRACING__` must not be replaced with `false` when you're using any performance monitoring-related SDK features (e.g. `Sentry.startTransaction()`). This flag is intended to be used in combination with packages like `@sentry/next` or `@sentry/sveltekit`, which automatically include performance monitoring functionality.
 
+<PlatformSection notSupported={["javascript.wasm", "javascript.cordova"]}>
+
 `__RRWEB_EXCLUDE_CANVAS__`
 
 : Replacing this flag with `true` will tree shake all code in the SDK that is related capturing canvas elements with Session Replay. This is only relevant when using [Session Replay](../../session-replay/). You can enable this flag if you don't have any canvas elements on your page you care to record.
@@ -47,6 +49,8 @@ To make optional code eligible for tree shaking, you can replace various flags i
 `__RRWEB_EXCLUDE_SHADOW_DOM__`
 
 : Replacing this flag with `true` will tree shake all code in the SDK that is related capturing shadow dom elements with Session Replay. This is only relevant when using [Session Replay](../../session-replay/). You can enable this flag if you don't have any shadow dom elements on your page you care to record.
+
+</PlatformSection>
 
 <PlatformSection notSupported={["javascript.nextjs", "javascript.sveltekit", "javascript.remix", "javascript.astro"]}>
 
--- a/src/platforms/javascript/common/configuration/tree-shaking/index.mdx
+++ b/src/platforms/javascript/common/configuration/tree-shaking/index.mdx
@@ -65,7 +65,7 @@ sentryPlugin({
     excludeReplayCanvas: true,
     excludeReplayIframe: true,
     excludeReplayShadowDom: true,
-  }
+  },
 });
 ```
 
--- a/src/platforms/javascript/common/configuration/tree-shaking/index.mdx
+++ b/src/platforms/javascript/common/configuration/tree-shaking/index.mdx
@@ -25,7 +25,7 @@ The Sentry SDK ships with code that is not strictly required for it to collect y
 
 ### List Of Flags
 
-To make optional code eligible for tree shaking, you can replace various flags in the Sentry SDK with `false`.
+To make optional code eligible for tree shaking, you can replace various flags in the Sentry SDK in order to remove the code from your build output. The following flags are available:
 
 `__SENTRY_DEBUG__`
 
@@ -36,7 +36,49 @@ To make optional code eligible for tree shaking, you can replace various flags i
 : Replacing this flag with `false` will tree shake all code in the SDK that is related to performance monitoring.
 **Attention:** `__SENTRY_TRACING__` must not be replaced with `false` when you're using any performance monitoring-related SDK features (e.g. `Sentry.startTransaction()`). This flag is intended to be used in combination with packages like `@sentry/next` or `@sentry/sveltekit`, which automatically include performance monitoring functionality.
 
-<PlatformSection notSupported={["javascript.nextjs"]}>
+`__RRWEB_EXCLUDE_CANVAS__`
+
+: Replacing this flag with `true` will tree shake all code in the SDK that is related capturing canvas elements with Session Replay. This is only relevant when using [Session Replay](../../session-replay/). You can enable this flag if you don't have any canvas elements on your page you care to record.
+
+`__RRWEB_EXCLUDE_IFRAME__`
+
+: Replacing this flag with `true` will tree shake all code in the SDK that is related capturing iframe content with Session Replay. This is only relevant when using [Session Replay](../../session-replay/). You can enable this flag if you don't have any iframes on your page you care to record.
+
+`__RRWEB_EXCLUDE_SHADOW_DOM__`
+
+: Replacing this flag with `true` will tree shake all code in the SDK that is related capturing shadow dom elements with Session Replay. This is only relevant when using [Session Replay](../../session-replay/). You can enable this flag if you don't have any shadow dom elements on your page you care to record.
+
+<PlatformSection notSupported={["javascript.nextjs", 'javascript.sveltekit', 'javascript.remix']}>
+
+### Tree Shaking Optional Code With Sentry Bundler Plugins
+
+If you are using one of our bundler plugins, you can use the `bundleSizeOptimizations` configuration option to tree shake optional code:
+
+```javascript
+sentryPlugin({
+  // other config
+  excludeDebugStatements: true,
+  excludePerformanceMonitoring: true,
+  excludeReplayCanvas: true,
+  excludeReplayIframe: true,
+  excludeReplayShadowDom: true,
+})
+```
+
+For more details, see the documentation for the specific bundler plugin you're using:
+
+* [Sentry Webpack Plugin](https://www.npmjs.com/package/@sentry/webpack-plugin)
+* [Sentry Vite Plugin](https://www.npmjs.com/package/@sentry/vite-plugin)
+* [Sentry Esbuild Plugin](https://www.npmjs.com/package/@sentry/esbuild-plugin)
+* [Sentry Rollup Plugin](https://www.npmjs.com/package/@sentry/rollup-plugin)
+
+<Note>
+  This configuration is available since v2.9.0 of the bundler plugins.
+</Note>
+
+</PlatformSection>
+
+<PlatformSection notSupported={["javascript.nextjs", 'javascript.sveltekit', 'javascript.remix']}>
 
 ### Tree Shaking Optional Code With Webpack
 
@@ -51,6 +93,9 @@ module.exports = {
     new webpack.DefinePlugin({
       __SENTRY_DEBUG__: false,
       __SENTRY_TRACING__: false,
+      __RRWEB_EXCLUDE_CANVAS__: true,
+      __RRWEB_EXCLUDE_IFRAME__: true,
+      __RRWEB_EXCLUDE_SHADOW_DOM__: true,
     }),
     // ... other plugins
   ],
@@ -59,7 +104,7 @@ module.exports = {
 
 </PlatformSection>
 
-<PlatformSection notSupported={["javascript.nextjs"]}>
+<PlatformSection notSupported={["javascript.nextjs", 'javascript.sveltekit', 'javascript.remix']}>
 
 ### Tree Shaking Optional Code With Rollup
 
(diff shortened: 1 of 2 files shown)


src/platforms/javascript/common/configuration/tree-shaking/index.mdx
```diff
@@ -36,7 +36,49 @@ To make optional code eligible for tree shaking, you can replace various flags i
...
+For more details, see the documentation for the specific bundler plugin you're using:
+
+- [Sentry Webpack Plugin](https://www.npmjs.com/package/@sentry/webpack-plugin)
+- [Sentry Vite Plugin](https://www.npmjs.com/package/@sentry/vite-plugin)
+- [Sentry Esbuild Plugin](https://www.npmjs.com/package/@sentry/esbuild-plugin)
+- [Sentry Rollup Plugin](https://www.npmjs.com/package/@sentry/rollup-plugin)
+
+<Note>
+  This configuration is available since v2.9.0 of the bundler plugins.
+</Note>
```
lforst, line 77: I would probably move this right below the heading for now, since it is super new. Maybe this also doesn't need to be a note but just italic text. Idk.

src/platforms/javascript/common/configuration/tree-shaking/index.mdx
```diff
@@ -36,7 +36,49 @@ To make optional code eligible for tree shaking, you can replace various flags i
...
+If you are using one of our bundler plugins, you can use the `bundleSizeOptimizations` configuration option to tree shake optional code:
+
+```javascript
+sentryPlugin({
+  // other config
+  excludeDebugStatements: true,
+  excludePerformanceMonitoring: true,
+  excludeReplayCanvas: true,
+  excludeReplayIframe: true,
+  excludeReplayShadowDom: true,
```
lforst, line 64: We are missing the surrounding `bundleSizeOptimizations` here.

src/platforms/javascript/common/configuration/tree-shaking/index.mdx
```diff
@@ -36,7 +36,49 @@ To make optional code eligible for tree shaking, you can replace various flags i
...
+
+`__RRWEB_EXCLUDE_IFRAME__`
+
+: Replacing this flag with `true` will tree shake all code in the SDK that is related capturing iframe content with Session Replay. This is only relevant when using [Session Replay](../../session-replay/). You can enable this flag if you don't have any iframes on your page you care to record.
+
+`__RRWEB_EXCLUDE_SHADOW_DOM__`
+
+: Replacing this flag with `true` will tree shake all code in the SDK that is related capturing shadow dom elements with Session Replay. This is only relevant when using [Session Replay](../../session-replay/). You can enable this flag if you don't have any shadow dom elements on your page you care to record.
+
+<PlatformSection notSupported={["javascript.nextjs", 'javascript.sveltekit', 'javascript.remix']}>
```
lforst, line 51: Astro too I believe. @Lms24 ?

src/platforms/javascript/common/configuration/tree-shaking/index.mdx
```diff
@@ -36,7 +36,49 @@ To make optional code eligible for tree shaking, you can replace various flags i
 : Replacing this flag with `false` will tree shake all code in the SDK that is related to performance monitoring.
 **Attention:** `__SENTRY_TRACING__` must not be replaced with `false` when you're using any performance monitoring-related SDK features (e.g. `Sentry.startTransaction()`). This flag is intended to be used in combination with packages like `@sentry/next` or `@sentry/sveltekit`, which automatically include performance monitoring functionality.
 
-<PlatformSection notSupported={["javascript.nextjs"]}>
+`__RRWEB_EXCLUDE_CANVAS__`
+
+: Replacing this flag with `true` will tree shake all code in the SDK that is related capturing canvas elements with Session Replay. This is only relevant when using [Session Replay](../../session-replay/). You can enable this flag if you don't have any canvas elements on your page you care to record.
```
lforst, line 41: Instead of the relative links we should likely be using platform links here right?

src/platforms/javascript/common/configuration/tree-shaking/index.mdx
Lms24: The Astro SDK currently doesn't expose Vite plugin options except for a few hand-picked ones to avoid depending on the plugin's API. I can update the SDK though to include this new API.

lforst: > (I guess that's the downside of not just passing through plugin options)

Yup that's a bit annoying, but I would say it is worth not having to break.

Stupid thought: We could have like one central layer (in utils or smth) for all the higher level JS SDKs that reexposes certain options. That way we wouldn't have to do this for all SDKs separately. (When we have Astro, Next.js, Sveltekit, and probably a million others soon (luc.js :P))

src/platforms/javascript/common/configuration/tree-shaking/index.mdx
mydea: oops, damn! 😅

src/platforms/javascript/common/configuration/tree-shaking/index.mdx
mydea: but with the relative links, it should work on the platforms, shouldn't it? E.g. if you are in react you should remain in react..?

bruno-garcia: might be a problem with the root JS docs vs guides:

> https://docs.sentry.io/platforms/javascript/session-replay/configuration/

vs

> https://docs.sentry.io/platforms/javascript/guides/angular/session-replay/configuration/

But we can check with the preview deploy (and CI should fail fwiw)

src/platforms/javascript/common/configuration/tree-shaking/index.mdx
```diff
@@ -36,7 +36,49 @@ To make optional code eligible for tree shaking, you can replace various flags i
...
+
+`__RRWEB_EXCLUDE_IFRAME__`
+
+: Replacing this flag with `true` will tree shake all code in the SDK that is related capturing iframe content with Session Replay. This is only relevant when using [Session Replay](../../session-replay/). You can enable this flag if you don't have any iframes on your page you care to record.
```
bruno-garcia, line 45: I realize the following ones also start with 'replace with true' where my suggestion above starts by clarifying this is relevant to Session Replay.

How about being clear at the top under a header 
```
### Session Replay

The following options are specific for Session Replay users. If you're not using Session Replay, (by adding `new Replay()` integration) these will not take any effect in your final bundle size
```

then we can skip being specific about it in each entry

src/platforms/javascript/common/configuration/tree-shaking/index.mdx
lforst: Hm, just checked. Seems to work fine so disregard! I could swear at one point these relative links would mess things up and end up linking to the generic guides instead.

mydea: actually we have a pretty decent link lint step which checks if the links are valid everywhere!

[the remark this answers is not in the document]
mydea: Technically you want to replace this flag in your source code, that's what the user has to do. You can't really "set" this anywhere - what you have to do is, however your build step looks like, make sure it replaces all occurrences of e.g. `__RRWEB_EXLUDE_CANVAS__` with `true`.

src/platforms/javascript/common/configuration/tree-shaking/index.mdx
mydea: hmm not 100% sure, then we need to basically reorganize all of this, as for consistency then we should have a section (?) for performance (`__SENTRY_TRACE__`) as well 🤔 There would be a bunch of headings then, e.g.:

## List of Flags

### Common Flags

`__SENTRY_DEBUG__`: 

....

### Performance

`__SENTRY_TRACE__`: 

...

### Session Replay

`__RRWEB_EXCLUDE_CANVAS__`: 

...

`__RRWEB_EXCLUDE_IFRAME__`: 

...

`__RRWEB_EXCLUDE_SHADOW_DOM__`: 

...


--> Do we like this better?

src/platforms/javascript/common/configuration/tree-shaking/index.mdx
mydea: (FWIW the `lint:links` command actually found some issue there, mainly that the link does not work for wasm and cordova 😅 fixed that! good CI!)

mydea:
All depenendencies for this have been released, so this is ready to go now!

src/platforms/javascript/common/configuration/tree-shaking/index.mdx
```diff
@@ -36,7 +36,53 @@ To make optional code eligible for tree shaking, you can replace various flags i
...
+
+`__RRWEB_EXCLUDE_SHADOW_DOM__`
+
+: Replacing this flag with `true` will tree shake all code in the SDK that is related capturing shadow dom elements with Session Replay. This is only relevant when using [Session Replay](../../session-replay/). You can enable this flag if you don't have any shadow dom elements on your page you care to record.
```
lizokm, line 51:
```suggestion
: Replacing this flag with `true` will tree shake all code in the SDK that's related to capturing shadow dom elements with Session Replay. This is only relevant when using [Session Replay](../../session-replay/). Enable this flag if you don't have any shadow dom elements on your page you want to record.
```

src/platforms/javascript/common/configuration/tree-shaking/index.mdx
```diff
@@ -36,7 +36,53 @@ To make optional code eligible for tree shaking, you can replace various flags i
...
+
+</PlatformSection>
+
+<PlatformSection notSupported={["javascript.nextjs", "javascript.sveltekit", "javascript.remix", "javascript.astro"]}>
+
+### Tree Shaking Optional Code With Sentry Bundler Plugins
+
+_This configuration is available since v2.9.0 of the bundler plugins._
```
lizokm, line 59:
```suggestion
_This configuration is available starting with v2.9.0 of the bundler plugins._
```

src/platforms/javascript/common/configuration/tree-shaking/index.mdx
```diff
@@ -36,7 +36,53 @@ To make optional code eligible for tree shaking, you can replace various flags i
...
+
+If you are using one of our bundler plugins, you can use the `bundleSizeOptimizations` configuration option to tree shake optional code:
```
lizokm, line 61:
```suggestion
If you're using one of our bundler plugins, you can use the `bundleSizeOptimizations` configuration option to tree shake optional code:
```

src/platforms/javascript/common/session-replay/troubleshooting.mdx
```diff
@@ -54,6 +54,15 @@ To mitigate these problems, SDK version 7.54.0 will [truncate console messages](
 
 If you're having any problems with the latest SDK version, we want to hear about it. Please open a [GitHub issue](https://github.com/getsentry/sentry-javascript/issues/new?assignees=&labels=Type%3A+Bug&projects=&template=bug.yml) and describe your situation.
 
+## Using Session Replay increases the bundle size of my application.
```
lizokm, line 57:
```suggestion
## Using Session Replay Increases the Bundle Size of My Application.
```

src/platforms/javascript/common/session-replay/troubleshooting.mdx
```diff
@@ -54,6 +54,15 @@ To mitigate these problems, SDK version 7.54.0 will [truncate console messages](
...
+
+Enabling Session Replay will add about 50 kb (gzipped) to your application bundle.
+While this is a significant increase, we believe the benefits of Session Replay outweigh the cost.
+Due to the inherent complexity of the browser enviornment, there is a considerable amount of code necessary in order for Session Replay to work.
+
+While we are generally working on ways to reduce the bundle size, there are a few steps you can take right now in order to reduce the size of Session Replay,
+based on your specific use case. See [Tree Shaking](../../configuration/tree-shaking/) for more information.
```
lizokm, line 64:
```suggestion
Because of the complexity of the browser environment, there's a significant amount of code necessary in order for Session Replay to work. And while enabling Session Replay will add about 50 kb (gzipped) to your application bundle, we believe the benefits will outweigh the cost.

We're always working on ways to reduce the bundle size. Additionally, there are steps you can take in order to reduce the size of Session Replay based on your specific use case. See [Tree Shaking](../../configuration/tree-shaking/) for more information.
```

src/platforms/javascript/common/session-replay/troubleshooting.mdx
```diff
@@ -54,6 +54,15 @@ To mitigate these problems, SDK version 7.54.0 will [truncate console messages](
...
+
 ## I see the message: "A large number of mutations was detected (N). This can slow down the Replay SDK and impact your customers."
```
lizokm, line 66:
```suggestion
## I See the Message: "A large number of mutations was detected (N). This can slow down the Replay SDK and impact your customers."
```

lizokm (approved):
Left a few language tweaks.

src/platforms/javascript/common/configuration/tree-shaking/index.mdx
Lms24: > We could have like one central layer (in utils or smth) for all the higher level JS SDKs that reexposes certain options

Hmm good idea! Let's see how things develop in Astro. Maybe we can extract this layer/API out for the next (!== NextJs) SDK or use it in existing SDKs when we break in v8.

src/platforms/javascript/common/configuration/tree-shaking/index.mdx
Lms24: I don't think this matters too much overall but if we go with this categorization, `__SENTRY_DEBUG__` and `__SENTRY_TRACING__` should both be under "common" because they remove bytes from the core SDK, while the Replay flags are only relevant when using replay.

Lms24 (approved):
LGTM overall!

billyvg (approved):

PR #8321, opened 2023-10-24
Replace traces_sample_rate with enable_tracing

stephanie-anderson:
Now that we have efficient backpressure management in our Python SDK, we no longer need to instruct users on how to enable performance monitoring by configuring their `traces_sample_rate` or `traces_sampler`.

When we simply talk about activating performance monitoring, we should now use `enable_tracing` instead. Attached are some screenshots from the performance monitoring getting started page.

### Before
before

### After
after

### DIFF
--- a/src/platform-includes/configuration/config-intro/python.mdx
+++ b/src/platform-includes/configuration/config-intro/python.mdx
@@ -9,11 +9,7 @@ sentry_sdk.init(
     dsn="___PUBLIC_DSN___",
     max_breadcrumbs=50,
     debug=True,
-
-    # Set traces_sample_rate to 1.0 to capture 100%
-    # of transactions for performance monitoring.
-    # We recommend adjusting this value in production,
-    traces_sample_rate=1.0,
+    enable_tracing=True,
 
     # By default the SDK will try to use the SENTRY_RELEASE
     # environment variable, or infer a git commit
--- a/src/platforms/common/performance/index.mdx
+++ b/src/platforms/common/performance/index.mdx
@@ -65,14 +65,16 @@ Automatic instrumentation for monitoring the performance of your application is 
 
 </PlatformSection>
 
-## Configure the Sample Rate
+## Configure
 
+<PlatformSection notSupported={["python"]}>
 First, enable tracing and configure the sampling rate for transactions. Set the sample rate for your transactions by either:
 
 - Setting a uniform sample rate for all transactions using the <PlatformIdentifier name="traces-sample-rate" /> option in your SDK config to a number between `0` and `1`. (For example, to send 20% of transactions, set <PlatformIdentifier name="traces-sample-rate" /> to `0.2`.)
 - Controlling the sample rate based on the transaction itself and the context in which it's captured, by providing a function to the <PlatformIdentifier name="traces-sampler" /> config option.
 
 The two options are meant to be mutually exclusive. If you set both, <PlatformIdentifier name="traces-sampler" /> will take precedence.
+</PlatformSection>
 
 <PlatformSection supported={["native"]}>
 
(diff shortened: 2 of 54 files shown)


src/platforms/python/integrations/aiohttp/aiohttp-client.mdx
```diff
@@ -29,6 +29,7 @@ from sentry_sdk.integrations.aiohttp import AioHttpIntegration
 
 sentry_sdk.init(
     dsn="___PUBLIC_DSN___",
+    enable_tracing=True,
```
stephanie-anderson, line 32: this option has been missing and had to be added

src/platforms/python/integrations/ariadne/index.mdx
```diff
@@ -27,9 +27,6 @@ from sentry_sdk.integrations.ariadne import AriadneIntegration
 
 sentry_sdk.init(
     dsn="___PUBLIC_DSN___",
-    # Set traces_sample_rate to 1.0 to capture 100%
-    # of transactions for performance monitoring.
-    traces_sample_rate=1.0,
     integrations=[
```
stephanie-anderson, line 30: Ariadne does not support performance monitoring as of now, so no need to activate it for this integration

src/platforms/python/integrations/logging/index.mdx
```diff
@@ -26,10 +26,7 @@ The logging integrations is a default integration so it will be enabled automati
 import sentry_sdk
 
 sentry_sdk.init(
-    dsn="___PUBLIC_DSN___",
-    # Set traces_sample_rate to 1.0 to capture 100%
-    # of transactions for performance monitoring.
-    traces_sample_rate=1.0,
+    dsn="___PUBLIC_DSN___"
 )
```
stephanie-anderson, line 30: The logging integration doesn't support performance monitoring

src/platform-includes/configuration/config-intro/python.mdx
```diff
@@ -9,11 +9,7 @@ sentry_sdk.init(
     dsn="___PUBLIC_DSN___",
     max_breadcrumbs=50,
     debug=True,
-
-    # Set traces_sample_rate to 1.0 to capture 100%
-    # of transactions for performance monitoring.
-    # We recommend adjusting this value in production,
-    traces_sample_rate=1.0,
+    enable_tracing=True,
```
shanamatthews, line 12: Does this set a default sample rate of 100%? 
Should we consider leaving in a similar comment that warns users?

src/platforms/python/integrations/aws-lambda/container-image/index.mdx
```diff
@@ -19,13 +19,13 @@ Set your image’s CMD value to the Sentry Handler:
...
 
-Also, set the following environment variables in AWS:
+Next, set the following environment variables in AWS:
 
 - Set `SENTRY_INITIAL_HANDLER` to the path of your function handler
 - Set `SENTRY_DSN` to your Sentry DSN
 - Set `SENTRY_TRACES_SAMPLE_RATE` to your preferred [sampling rate](/platforms/python/configuration/sampling/#sampling-transaction-events) for transactions
 
-Alternatively, you can also set the environment variables in the Dockerfile:
+Alternatively to AWS, you can also set the environment variables in your Dockerfile. Please note, that values in AWS override the values in a Dockerfile if both are provided.
```
shanamatthews, line 28:
```suggestion
Alternatively, you can also set the environment variables in your Dockerfile. Values set in AWS override the values in a Dockerfile if both are provided.
```

shanamatthews (approved):
Thanks for doing all this extra cleanup work too!

src/platform-includes/configuration/config-intro/python.mdx
stephanie-anderson: @shanamatthews  Yes, `enable_tracing=True` is equal to `traces_sample_rate=1.0`. But warning users is no longer necessary (at least in this SDK), they can also set it to 1.0 in production and the SDK will take care of tuning down in case of heavy load. Next quarter we're working on backpressure management in the Java and Ruby SDK, so that we can deliver the same (simpler and safer) experience there.

src/platforms/common/performance/index.mdx
```diff
@@ -86,8 +89,9 @@ The Native SDK doesn't currently support sampling functions (<PlatformIdentifier
 
 <PlatformContent includePath="performance/configure-sample-rate" />
 
-Learn more about how the options work in <PlatformLink to="/configuration/sampling/">Sampling Transactions</PlatformLink>.
+Learn more about performance monitoring <PlatformLink to="/configuration/options/#tracing-options">options</PlatformLink>, or how to <PlatformLink to="/configuration/sampling/">sample transactions</PlatformLink>.
```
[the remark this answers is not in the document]
stephanie-anderson: Great catch! Had the same thought, but wanted to rename the section to `Performance Monitoring` before. Will make a separate PR for that

PR #8331, opened 2023-10-26
Remove inheritance for Python's index.mdx

stephanie-anderson:


shanamatthews:
Hey @stephanie-anderson ! Is the idea to start moving away from templated docs for everything, or is this a special case?

PR #8337, opened 2023-10-26
Make it clear that dsn needs to be set when using tunnel

AbhiPrasad:
ref https://github.com/getsentry/sentry-javascript/discussions/9377

### DIFF
--- a/src/platforms/common/configuration/options.mdx
+++ b/src/platforms/common/configuration/options.mdx
@@ -174,8 +174,7 @@ By default the SDK will try to read this value from the `SENTRY_ENVIRONMENT` env
 
 <ConfigKey name="tunnel" supported={["javascript", "node"]} notSupported={["react-native"]}>
 
-Sets the URL that will be used to transport captured events, instead of using the DSN. This can be used to work around ad-blockers or to have more granular
-control over events sent to Sentry. This option **requires the implementation** of a custom server endpoint. Learn more and find examples in [Dealing with Ad-Blockers](/platforms/javascript/troubleshooting/#dealing-with-ad-blockers).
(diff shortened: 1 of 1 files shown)


src/platforms/common/configuration/options.mdx
```diff
@@ -174,8 +174,7 @@ By default the SDK will try to read this value from the `SENTRY_ENVIRONMENT` env
 
 <ConfigKey name="tunnel" supported={["javascript", "node"]} notSupported={["react-native"]}>
 
-Sets the URL that will be used to transport captured events, instead of using the DSN. This can be used to work around ad-blockers or to have more granular
-control over events sent to Sentry. This option **requires the implementation** of a custom server endpoint. Learn more and find examples in [Dealing with Ad-Blockers](/platforms/javascript/troubleshooting/#dealing-with-ad-blockers).
+Sets the URL that will be used to transport captured events. This can be used to work around ad-blockers or to have more granular control over events sent to Sentry. Usage of a DSN is still required so that needed attributes can be set on the generated Sentry data. This option **requires the implementation** of a custom server endpoint. Learn more and find examples in [Dealing with Ad-Blockers](/platforms/javascript/troubleshooting/#dealing-with-ad-blockers).
```
shanamatthews, line 177:
```suggestion
Sets the URL that will be used to transport captured events. This can be used to work around ad-blockers or to have more granular control over events sent to Sentry. Adding your DSN is still required when using this option so necessary attributes can be set on the generated Sentry data. This option **requires the implementation** of a custom server endpoint. Learn more and find examples in [Dealing with Ad-Blockers](/platforms/javascript/troubleshooting/#dealing-with-ad-blockers).
```

shanamatthews: Also not totally sure we need this part of the sentence - do users need to know why?

> so necessary attributes can be set on the generated Sentry data

shanamatthews (approved):
Just some wording thoughts.

src/platforms/common/configuration/options.mdx
AbhiPrasad: I added that section because I figured it would be nice to give some context why DSN needs to be set, but we can remove it if you think that is better.

shanamatthews: No strong feelings. The general principle is to only include info users need to know to keep things short. If you think the explanation is helpful, keep it :)

PR #8347, opened 2023-10-27
feat(crons): Document withMonitor API for JS SDK

AbhiPrasad:
Documents https://github.com/getsentry/sentry-javascript/pull/9395 released as part of https://github.com/getsentry/sentry-javascript/releases/tag/7.76.0

### DIFF
--- a/src/includes/javascript-crons-checkins.mdx
+++ b/src/includes/javascript-crons-checkins.mdx
@@ -11,6 +11,7 @@ const checkInId = Sentry.captureCheckIn({
 
 // 🟢 Notify Sentry your job has completed successfully:
 Sentry.captureCheckIn({
+  // Make sure this variable is named `checkInId`
   checkInId,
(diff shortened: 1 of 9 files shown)


shanamatthews:
@AbhiPrasad - it looks like [Astro](https://docs.sentry.io/platforms/javascript/guides/astro/) and [Bun](https://docs.sentry.io/platforms/javascript/guides/bun/) don't have troubleshooting pages for Crons. Was that an oversight?

I think they just need to be added to the supported lists in src/platforms/common/crons/troubleshooting.mdx

AbhiPrasad:
> it looks like [Astro](https://docs.sentry.io/platforms/javascript/guides/astro/) and [Bun](https://docs.sentry.io/platforms/javascript/guides/bun/) don't have troubleshooting pages for Crons. Was that an oversight?

Yes it was an oversight, good call! Will update.

PR #8362, opened 2023-10-30
fix: Fix order of npm vs yarn

mydea:
We streamlined this some while ago, but forgot a place, where we still had `yarn` before `npm`.


krystofwoldrich (approved):
LG 🚀

Lms24 (approved):

PR #8366, opened 2023-10-30
feat: Update docs for node opentelemetry

mydea:
This updates the docs for OpenTelemetry-based node instrumentation to the new `@sentry/opentelemetry` package.

* Also fix an issue with a `Verify` heading
* Also remove the beta note that was on all the otel docs pages, as it was a bit over prominent


src/platform-includes/performance/opentelemetry-setup/node.mdx
```diff
@@ -17,25 +25,50 @@ const {
...
-});
-
-const sdk = new opentelemetry.NodeSDK({
-  // Existing config
-  traceExporter: new OTLPTraceExporter(),
-  instrumentations: [getNodeAutoInstrumentations()],
-
-  // Sentry config
-  spanProcessor: new SentrySpanProcessor(),
-  textMapPropagator: new SentryPropagator(),
```
AbhiPrasad, line 37: I wonder if we should get rid of the old approach entirely. There are still some people using it. Maybe we add a "DEPRECATED" section below with this code?

mydea: Hmm, but I guess for people already using the old approach there isn't much need to go to the docs and find the same thing again? 😅 WDYT?

AbhiPrasad: yeah that's fair. We can also just link to the npm readme as a note then.

AbhiPrasad (approved):

src/platform-includes/performance/opentelemetry-setup/node.mdx
```diff
@@ -1,14 +1,22 @@
-You need to register `SentrySpanProcessor` and `SentryPropagator` with your OpenTelemetry installation:
+There are a few steps necessary in order to ensure that your OpenTelemetry setup is fully synced with your Sentry SDK.
```
shanamatthews, line 1:
```suggestion
There are a few steps necessary in order to ensure that your OpenTelemetry setup is fully synced with your Sentry SDK. The following code snippet walks through the needed steps.
```

src/platform-includes/performance/opentelemetry-setup/node.mdx
```diff
@@ -17,25 +25,50 @@ const {
...
+  setOpenTelemetryContextAsyncContextStrategy();
+
+  sdk.start();
+}
+
+setupSentry();
 ```
+
+Note that with this setup, you can _only_ use OpenTelemetry tracing for performance.
+Any Sentry performance APIs (e.g. `Sentry.startTransaction()`) will not work.
```
shanamatthews, line 74:
```suggestion
With this setup, you can _only_ use OpenTelemetry tracing for performance. Sentry performance APIs (e.g. `Sentry.startTransaction()`) will not work.
```

src/platform-includes/performance/opentelemetry-setup/node.mdx
shanamatthews: Wait, does the npm readme have the old instructions? If so, I think linking to it would be a good solution.

shanamatthews (approved):

src/platform-includes/performance/opentelemetry-setup/node.mdx
mydea: I added a section below (a note) pointing to the old readme, WDYT?

src/platform-includes/performance/opentelemetry-setup/node.mdx
```diff
@@ -17,25 +25,71 @@ const {
...
-  // Sentry config
-  spanProcessor: new SentrySpanProcessor(),
-  textMapPropagator: new SentryPropagator(),
+  sdk.start();
+}
+
+setupSentry();
+```
+
+Note that with this setup, you can _only_ use OpenTelemetry tracing for performance. `Sentry.startTransaction()` will _not_ work.
```
mydea, line 73: @shanamatthews I updated this actually and made this more explicit. I specifically call out `startTransaction()` as not working (as that is really the main thing that will not work) but also explain what to do instead!

src/platform-includes/performance/opentelemetry-setup/node.mdx
```diff
@@ -17,25 +25,81 @@ const {
...
+    target="_blank"
+    rel="noopener"
+  >
+    @sentry/opentelemetry-node
+  </a>{" "}
+  package, you can continue to do so and find usage instructions in the
+  package's README. This package has a slightly easier setup process, but
+  provides a more limited connection between Sentry and OpenTelemetry than the
+  instructions outlined above.
+</Note>
```
shanamatthews, line 105: I think if you leave extra newlines padding the contents of the note, markdown inside the note will render correctly:

```suggestion
<Note>

  If you are using the [@sentry/opentelemetry-node](https://www.npmjs.com/package/@sentry/opentelemetry-node] package, you can continue to do so. Usage instructions are in the
  package's README. This package has a slightly easier setup process but
  provides a more limited connection between Sentry and OpenTelemetry than the
  instructions outlined above.
  
</Note>
```

src/platform-includes/performance/opentelemetry-setup/node.mdx
shanamatthews: This looks good to me!

shanamatthews (approved):
I like the changes here. Just one suggestion to simplify the note so we don't have to use html.

src/platform-includes/performance/opentelemetry-setup/node.mdx
mydea: ah, nice, didn't know that! 👍

PR #8394, opened 2023-11-02
fix(remix): Add Remix-specific Performance monitoring setup code snippets

Lms24:
Currently, we only showed the generic `@sentry/browser` code snippet to set up performance monitoring for the Remix SDK. This PR adds Remix-specific code snippets that show both, client and server code for Remix projects.  

Side note: I think we can probably go the same route with Remix as with Next/Sveltekit/Astro and always add BrowserTracing under the hood. This however is a larger task for which I currently don't have time. So let's go with correct docs for now and adjust later when/if we get to simplify things in the Remix SDK.


shanamatthews (approved):
lgtm, thanks!

PR #8398, opened 2023-11-02
Move Distributed Tracing Docs

stephanie-anderson:
Split the SDK docs for distributed tracing. Left the site under /usage for the general explanation, but added a top level item that explains the setup etc.


shanamatthews:
Just adding these comments here instead of in slack 😄 

- I’m confused about why the base Python “[Set Up Distributed Tracing](https://sentry-docs-git-distributed-tracing.sentry.dev/platforms/python/distributed-tracing/)” page no longer has information about extracting & injecting trace info. Was that intentional?
- I also noticed that you removed the explicit callouts about CORS on the backend pages. I wonder if it would be better to leave it?

PR #8408, opened 2023-11-03
fix(browser-profiling): Remove `BrowserTracing` from Nextjs and Sveltekit snippets

Lms24:
Both, the NextJS and SvelteKit SDKs, already add `BrowserTracing` by default. There's no harm in adding them as a user but on the other hand, it's not necessary either. So this PR removes the intgration from the respective snippets.

The same thing applies to Astro but I don't think we have Astro specific Browser profiling docs yet. @JonasBa are you currently adding more platforms or should I quickly add a snippet for Astro?

ref: https://github.com/syntaxfm/website/pull/1346#discussion_r1381817262


shanamatthews (commented):
👏 thanks for updating this!

vivianyentran (approved):

JonasBa:
Oh yes, please feel free to add Astro docs!

Lms24:
Kinda forgot about this PR. Will merge it as-is and tackle Astro profiling later.

PR #8414, opened 2023-11-04
Ruby: add proxy config docs

natikgadzhi:
## Summary

This PR adds a little blob on configuring proxy in `sentry-ruby`. 

I worked on https://github.com/getsentry/sentry-ruby/pull/2161 and realized that proxy configuration is not documented yet — so here we go. 

@shanamatthews, in `sentry-ruby` 5.1.2, we don't yet support reading proxy from env variables, but we just pushed up a PR with it. Is it a good idea to document the env var behavior in this PR, or wait until it's merged and released?

## Legal Boilerplate

## Extra resources

### DIFF
--- a/src/platforms/ruby/common/configuration/options.mdx
+++ b/src/platforms/ruby/common/configuration/options.mdx
@@ -251,6 +251,30 @@ config.trusted_proxies = ["2.2.2.2"]
 config.transport.transport_class = MyTransportClass
 ```
 
+`proxy`
+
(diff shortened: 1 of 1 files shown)


shanamatthews:
@natikgadzhi - we try to wait until the new SDK version is released before merging docs for it, so let's keep the PRs separate!

natikgadzhi:
> so let's keep the PRs separate!

Then this one should be good to go, too!

shanamatthews:
@sl0thentr0py - mind taking a look at this to make sure it's all correct?

sl0thentr0py:
i will merge this after releasing the feature

PR #8422, opened 2023-11-06
Sanic instructions initialize in before_server_start

szokeasaurusrex:
## Pre-merge checklist

## Description of changes

Document https://github.com/getsentry/sentry-python/issues/1690#issuecomment-1787605696

## Extra resources

### DIFF
--- a/src/platforms/python/integrations/sanic/index.mdx
+++ b/src/platforms/python/integrations/sanic/index.mdx
@@ -25,18 +25,25 @@ pip install --upgrade aiocontextvars
 
 If you have the `sanic` package in your dependencies, the Sanic integration will be enabled automatically when you initialize the Sentry SDK.
 
+To ensure that errors occuring in background tasks get reported to Sentry, we always recommend enabling the [`AsyncioIntegration`](/platforms/python/integrations/asyncio/) when you are using the `SanicIntegration`. To enable the `AsyncioIntegration`, you must explicitly add the `AsyncioIntegration` to the `integrations` list and initialize the SDK within a `before_server_start` listener.
+
+The example below demonstrates the simplest recommended setup.
+
 <SignInNote />
 
 ```python
 from sanic import Sanic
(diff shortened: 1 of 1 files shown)


src/platforms/python/integrations/sanic/index.mdx
```diff
@@ -25,18 +25,25 @@ pip install --upgrade aiocontextvars
 
 If you have the `sanic` package in your dependencies, the Sanic integration will be enabled automatically when you initialize the Sentry SDK.
 
+To ensure that errors occuring in background tasks get reported to Sentry, we always recommend enabling the [`AsyncioIntegration`](/platforms/python/integrations/asyncio/) when you are using the `SanicIntegration`. To enable the `AsyncioIntegration`, you must explicitly add the `AsyncioIntegration` to the `integrations` list and initialize the SDK within a `before_server_start` listener.
```
shanamatthews, line 28:
```suggestion
To ensure that errors that occur in background tasks get reported to Sentry, enable the [`AsyncioIntegration`](/platforms/python/integrations/asyncio/) when you use the `SanicIntegration`. To enable the `AsyncioIntegration`, you must explicitly add the `AsyncioIntegration` to the `integrations` list and initialize the SDK within a `before_server_start` listener.
```

shanamatthews (approved):
optional wording tweaks

src/platforms/python/integrations/sanic/index.mdx
szokeasaurusrex: Thank you for the suggestion, but I think it would be better to keep the initial wording (or something closer to it) here, since I would like to emphasize that we recommend users to always initialize the AsyncioIntegration when using the SanicIntegration, since there is no real downside to doing so. With your proposed change, it reads more like we might only be recommending enabling the AsyncioIntegration when there are background tasks that might raise errors

antonpirker (approved):

PR #8438, opened 2023-11-07
ref(native): Add note about multi-process support

kahest:
Add note about multi-process support with crashpad using separate or shared database paths

### DIFF
--- a/src/platforms/native/configuration/backends/crashpad.mdx
+++ b/src/platforms/native/configuration/backends/crashpad.mdx
@@ -37,10 +37,11 @@ according to shell lookup rules at runtime.
 It is advised to specify both the path to the `crashpad_handler` executable
 as well as the path to the database directory manually.
 
-Multi-process use cases are also supported. Initialize Sentry individually
-in each process which spawns its own Crashpad handler either in a separate
-database path, or using a shared database path. Each run uses its individual
-run directory, and there is filesystem-level locking in place.
+You can also use Crashpad for multi-process use cases. To do this, initialize
+Sentry individually in each process, so that every process spawns its own
+Crashpad handler, either in a separate database path or using a shared database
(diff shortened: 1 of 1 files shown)


src/platforms/native/configuration/backends/crashpad.mdx
```diff
@@ -37,6 +37,11 @@ according to shell lookup rules at runtime.
 It is advised to specify both the path to the `crashpad_handler` executable
 as well as the path to the database directory manually.
 
+Multi-process use cases are also supported. Initialize Sentry individually
+in each process which spawns its own Crashpad handler either in a separate
+database path, or using a shared database path. Each run uses its individual
+run directory, and there is filesystem-level locking in place.
```
shanamatthews, line 43:
```suggestion
You can also use Crashpad for multi-process use cases. To do this, initialize Sentry individually
in each process that spawns its own Crashpad handler, either in a separate
database path or using a shared database path. Each run will use its individual
run directory, and there is filesystem-level locking in place.
```

shanamatthews: I found this original sentence to be a little confusing, so hopefully this reword makes it more clear and accurately reflects what you meant!

shanamatthews (approved):

src/platforms/native/configuration/backends/crashpad.mdx
kahest: Thanks for the suggestion! I adapted it a tiny bit, wdyt?

Swatinem (approved):

src/platforms/native/configuration/backends/crashpad.mdx
shanamatthews: This looks awesome, thanks for taking the time to make sure this is really clear!

PR #8447, opened 2023-11-08
fix(replay): Clarify streaming video capture

mydea:
This clarifies in replay troubleshooting docs that only static videos can be captured.

Closes https://github.com/getsentry/sentry-javascript/issues/9447#issuecomment-1801639171

### DIFF
--- a/src/platforms/javascript/common/session-replay/troubleshooting.mdx
+++ b/src/platforms/javascript/common/session-replay/troubleshooting.mdx
@@ -29,6 +29,8 @@ There's currently no support for `canvas`. It's being tracked in this [GitHub is
 
 The replay 'video' is actually a video-like reproduction of the HTML on your website. This means that all the external resources your site uses (CSS/Images/Fonts), will be rendered by the corresponding &lt;style&gt;, &lt;img&gt; and &lt;video&gt; tags on your site. Add `sentry.io` to your CORS policy so the iframe hosted on sentry.io can fetch and display these resources.
 
+Please note that only static videos can be captured by Replay, e.g. `src="./my-video.mp4"`. Streamed videos and similar are not supported.
+
 <Alert>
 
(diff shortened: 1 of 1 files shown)


cmanallen (approved):
Could also clarify the link must be public otherwise we will not be able to download the video for the replay.

mydea:
> Could also clarify the link must be public otherwise we will not be able to download the video for the replay.

Good point, I tweaked the language slightly to indicate that!

cmanallen:
👍 LGTM!

billyvg (approved):

src/platforms/javascript/common/session-replay/troubleshooting.mdx
```diff
@@ -29,6 +29,8 @@ There's currently no support for `canvas`. It's being tracked in this [GitHub is
 
 The replay 'video' is actually a video-like reproduction of the HTML on your website. This means that all the external resources your site uses (CSS/Images/Fonts), will be rendered by the corresponding &lt;style&gt;, &lt;img&gt; and &lt;video&gt; tags on your site. Add `sentry.io` to your CORS policy so the iframe hosted on sentry.io can fetch and display these resources.
 
+Please note that only static, publicly hosted videos can be captured by Replay, e.g. `src="./my-video.mp4"`. Streamed videos and similar are not supported.
```
vivianyentran, line 32:
```suggestion
Note that only static, publicly hosted videos (for example, `src="./my-video.mp4"`) can be captured by Replay. Streamed videos and similar are not supported.
```
We generally avoid latin acronyms (e.g.,) in docs. 

"and similar" is pretty vague, could this be clarified a bit more? If not, it's ok as is too.

vivianyentran (approved):
Added a comment

src/platforms/javascript/common/session-replay/troubleshooting.mdx
mydea: Not 100% sure about all the other cases how you can show a video 🤔 I'll merge this for now I think (with your suggestions, thank you!) and we can potentially tweak it later if we know of other specific scenarios!

PR #8448, opened 2023-11-08
fix(replay): update treeshaking docs to remove canvas flag

mydea:
This is now excluded by default, no need to manually configure anything.


billyvg (approved):

vivianyentran (approved):
thanks for updating

PR #8451, opened 2023-11-08
Remove outdated histogram section in web vitals doc

vivianyentran:
- Remove distribution histogram section (no longer shown in UI)
- Delete related image


edwardgou-sentry (approved):
Thanks!

PR #8458, opened 2023-11-09
feat(replay): Add docs for custom web worker

mydea:
This adds docs for the new `workerUrl` config for replay, along with in-depth explanation on how & when to use it.

This is a bit more involved, but I hope I managed to explain this reasonably OK.

This depends on https://github.com/getsentry/publish/issues/2952


src/platforms/javascript/common/session-replay/configuration.mdx
```diff
@@ -87,6 +88,50 @@ new Replay({
 
 <Note>Captured bodies will be truncated to 150k characters max.</Note>
 
+## Identifying Users
```
mydea, line 91: Moved this up above "Mutation limits" as IMHO mutation limits is the least important config described on this page...?

src/platforms/javascript/common/session-replay/configuration.mdx
```diff
@@ -87,21 +88,57 @@ new Replay({
...
 ```javascript
 new Replay({
-  mutationBreadcrumbLimit: 1000,
-  mutationLimit: 1500,
+  workerUrl: "/assets/worker.min.js",
 });
 ```
 
-## Identifying Users
+4. You can now [tree shake](../../configuration/tree-shaking/) the default included worker script. If you are using our bundler plugins (version 2.10.0 or later), you can do:
```
billyvg, line 122: Should we add a webpack example as well?

src/platforms/javascript/common/session-replay/index.mdx
```diff
@@ -83,6 +83,8 @@ worker-src 'self' blob:
 child-src 'self' blob:
 ```
 
+If you don't want to update your CSP policy to allow inline web workers, you can also [use a custom compression worker](./configuration#using-a-custom-compression-worker) instead.
```
billyvg, line 86: I wouldn't describe this as a "want" --

```suggestion
If you're unable to update your CSP policy to allow inline web workers, you can also [use a custom compression worker](./configuration#using-a-custom-compression-worker) instead.
```

src/platforms/javascript/common/session-replay/configuration.mdx
```diff
@@ -87,21 +88,57 @@ new Replay({
...
+You can use the Sentry SDK to link a user to a session. Read more about it in our <PlatformLink to="/enriching-events/identify-user">docs</PlatformLink>.
+
+```javascript
+Sentry.setUser({ email: "jane.doe@example.com" });
+```
+
+## Using a Custom Compression Worker
+
+By default, Session Replay will use an inlined web worker script to compress Replay data before we send it over the network. This helps to drastically reduce the amount of data transferred, improving performance and reducing network overhead. Because compressing data is CPU intensive, we use a web worker to offload this work to a separate thread.
```
shanamatthews, line 101: Avoid using "we" to refer to Sentry

```suggestion
By default, Session Replay will use an inlined web worker script to compress Replay data before sending it over the network. This drastically reduces the amount of data transferred, improving performance and reducing network overhead. Because compressing data is CPU intensive, we use a web worker to offload this work to a separate thread.
```

src/platforms/javascript/common/session-replay/configuration.mdx
```diff
@@ -87,21 +88,57 @@ new Replay({
...
+
+While an inlined worker will work well for most applications, there are two main problems with this:
+
+1. Since the worker code is inlined, it increases the main bundle size of your application.
+2. Using inline workers can cause CSP violations if your application has a very strict CSP policy. [Learn more about configuring your CSP policy.](../#content-security-policy-csp)
```
shanamatthews, line 106: Does a PlatformLink work? They're more robust than ../ if page structure changes in the future.

```suggestion
2. Using inline workers can cause CSP violations if your application has a very strict CSP policy. See our <PlatformLink to="/session-replay/#content-security-policy-csp">CSP</PlatformLink> docs for more information.
```

src/platforms/javascript/common/session-replay/configuration.mdx
```diff
@@ -87,21 +88,57 @@ new Replay({
...
+
+To solve these problems, we provide the ability to use a custom worker script. This allows you to host the worker script on your own domain, and avoid the CSP issues that come with inline workers. Additionally, the worker script can then be removed from the main application bundle, and can be served and cached separately.
+
+Follow these steps in order to use a custom compression worker:
+
+1. Download the [example worker script](https://github.com/getsentry/sentry-javascript/blob/develop/packages/replay-worker/examples/worker.min.js) from our Github repository.
+2. Host this worker script on the same domain as your application. For example, if your application is hosted at `https://example.com`, you could host the worker script at `https://example.com/assets/worker.min.js`.
+3. Configure your custom worker script in replay:
```
shanamatthews, line 114:
```suggestion
3. Configure your custom worker script in Replay:
```

src/platforms/javascript/common/session-replay/configuration.mdx
shanamatthews, line 122: Also prefer PlatformLink if it works

```suggestion
4. You can now <PlatformLink to="/configuration/tree-shaking/">Tree Shake</PlatformLink> the default included worker script. If you are using our bundler plugins (version 2.10.0 or later), you can do:
```

src/platforms/javascript/common/session-replay/configuration.mdx
```diff
@@ -87,21 +88,57 @@ new Replay({
...
+```javascript
+sentryPlugin({
+  // other config
+  bundleSizeOptimizations: {
+    excludeReplayWorker: true,
+  },
+});
+```
+
+Please note that since you are hosting this script yourself, you are responsible for keeping it up to date. We recommend that you check for updates to the worker script periodically, and update it as needed. The worker script API is guaranteed to remain stable inside of a major version, so any v7 worker script should be backwards and forwards compatible with any v7 SDK version. However, we may make improvements or bug fixes to the worker script, which you may miss out if you don't update it.
```
shanamatthews, line 133:
```suggestion
Since you are hosting this script yourself, you are responsible for keeping it up to date. We recommend that you check for updates to the worker script periodically, and update it as needed. The worker script API is guaranteed to remain stable inside of a major version, so any v7 worker script should be backward and forward compatible with any v7 SDK version. However, we may make improvements or bug fixes to the worker script, which you will miss out on if you don't update it.
```

src/platforms/javascript/common/session-replay/configuration.mdx
```diff
@@ -87,21 +88,57 @@ new Replay({
...
+
+## Mutation Limits
+
+Session Replay works by recording incremental DOM changes that occur in your web application. Generally these changes occur in small batches and cause minimal overhead. However, it can be easy to overlook cases that will cause a large amount of DOM mutations to happen in a single update. An example is a custom dropdown component that has an unbounded list of options to render. This can negatively impact performance without Session Replay and its effects will be magnified with Session Replay enabled. In order to avoid this scenario, Session Replay will stop recording if it detects a large number of mutations (default: 10,000), which can be configured by setting `mutationLimit`. Additionally, we provide breadcrumbs in the replay to warn you when a large number of mutations are detected (default: 750).
```
shanamatthews, line 137: Just taking advantage of the fact that you moved this to add a missing comma :D

```suggestion
Session Replay works by recording incremental DOM changes that occur in your web application. Generally, these changes occur in small batches and cause minimal overhead. However, it can be easy to overlook cases that will cause a large amount of DOM mutations to happen in a single update. An example is a custom dropdown component that has an unbounded list of options to render. This can negatively impact performance without Session Replay and its effects will be magnified with Session Replay enabled. In order to avoid this scenario, Session Replay will stop recording if it detects a large number of mutations (default: 10,000), which can be configured by setting `mutationLimit`. Additionally, we provide breadcrumbs in the replay to warn you when a large number of mutations are detected (default: 750).
```

shanamatthews (approved):
Some small grammar and style guide tweaks, but otherwise looks good, as long as you make sure to add that missing link.

src/platforms/javascript/common/configuration/tree-shaking/index.mdx
```diff
@@ -48,6 +48,10 @@ The following flags are available:
 
 : Replacing this flag with `true` will tree shake all code in the SDK that's related to capturing shadow dom elements with Session Replay. This is only relevant when using [Session Replay](../../session-replay/). Enable this flag if you don't have any shadow dom elements on your page you want to record.
 
+`__SENTRY_EXCLUDE_REPLAY_WORKER__`
+
+: Replacing this flag with `true` will tree shake all code in the SDK that's related to the included compression web worker for Session Replay. This is only relevant when using [Session Replay](../../session-replay/). Enable this flag if you want to host a compression worker yourself - see XXXX for details. **We do not reccomend to enable this flag unless you are providing a custom worker URL.**
```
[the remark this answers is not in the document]
mydea: oops totally forgot to update this xD will fix!

src/platforms/javascript/common/session-replay/configuration.mdx
mydea: This is for webpack as well, any of our bundler plugins can be configured like this. How could we make this clearer, without repeating this for all types of bundlers? 🤔 I added a comment above:

```js
// For example, the @sentry/webpack-plugin passed to the webpack config
```

WDYT about this?

billyvg: Ah was just thinking of the case where we don't use our bundler plugins

mydea: I'd say for this case it's OK to link to the tree shaking page, as otherwise we need to add quite a lot of context again 😅

PR #8463, opened 2023-11-09
fix(native): remove obsolete warning note...

supervacuus:
... for the missing proxy support in the crashpad handler.

We've added proxy support to the crashpad fork here:

https://github.com/getsentry/sentry-native/pull/847

This was released as part of https://github.com/getsentry/sentry-native/releases/tag/0.6.3


shanamatthews (approved):
Thanks for making this cleanup!

PR #8470, opened 2023-11-10
ref(js): Add note that Sentry Testkit is community-maintained

Lms24:
Sentry Testkit is a [community-maintained library](https://github.com/zivl/sentry-testkit), meaning we don't offer official support for it. However, we quite prominently feature them in our docs. This PR adds a note that this library is not maintained by us and that support requests should be made in the maintainer's repository.

### DIFF
--- a/src/platforms/javascript/common/configuration/sentry-testkit/index.mdx
+++ b/src/platforms/javascript/common/configuration/sentry-testkit/index.mdx
@@ -8,6 +8,13 @@ When building tests for your application, you want to assert that the right flow
 
 [Sentry Testkit](https://zivl.github.io/sentry-testkit/) is a Sentry plugin that allows Sentry's reports to be intercepted for further data inspection. It enables Sentry to work natively in your application, by overriding Sentry's default transport mechanism, which makes it so that the report isn't really sent, but rather logged locally into memory. This way, logged reports can be fetched later for your own usage, verification, or any other purpose you may have in your local developing or testing environment.
 
+<Alert level="warning">
+
+Sentry Testkit is a community-maintained 3rd party library, and is **not officially supported by Sentry**.
+If you have any questions or issues with Sentry Testkit, please open an issue in the [Sentry Testkit repository](https://zivl.github.io/sentry-testkit/)
+
+</Alert>
+
 ## Installation
 
 ```bash {tabTitle:npm}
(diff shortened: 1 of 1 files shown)


mydea (approved):

src/platforms/javascript/common/configuration/sentry-testkit/index.mdx
```diff
@@ -8,6 +8,13 @@ When building tests for your application, you want to assert that the right flow
 
 [Sentry Testkit](https://zivl.github.io/sentry-testkit/) is a Sentry plugin that allows Sentry's reports to be intercepted for further data inspection. It enables Sentry to work natively in your application, by overriding Sentry's default transport mechanism, which makes it so that the report isn't really sent, but rather logged locally into memory. This way, logged reports can be fetched later for your own usage, verification, or any other purpose you may have in your local developing or testing environment.
 
+<Alert level="warning">
+
+Sentry Testkit is a community-maintained 3rd party library, and is **not officially supported by Sentry**.
+If you have any questions or issues with Sentry Testkit, please open an issue in the [Sentry Testkit repository](https://zivl.github.io/sentry-testkit/)
+
+</Alert>
+
```
lforst, line 17:
```suggestion
[Sentry Testkit](https://zivl.github.io/sentry-testkit/) is a community-maintained Sentry plugin that allows Sentry's reports to be intercepted for further data inspection. It enables Sentry to work natively in your application, by overriding Sentry's default transport mechanism, which makes it so that the report isn't really sent, but rather logged locally into memory. This way, logged reports can be fetched later for your own usage, verification, or any other purpose you may have in your local developing or testing environment.

```

I fear that this alert could come off as a bit hostile. I think the project is really cool and useful and putting an alert here could come off as a bit silo-y while we should encourage such projects.

Lms24: I still want to at least redirect folks to their repo for issues so I'm gonna apply your text suggestion but add an Alert (level info) with the issues sentence. Currently this page looks 100% like its our product and this is problematic.

lforst: Sounds good to me! I am fully on-board with the fact that we point this out. I just wanted to make sure we get the tone right. :)

lforst (approved):

PR #8485, opened 2023-11-14
build: Remove `format` command and add `lint:fix` command

mydea:
Noticed that the `yarn format` command doesn't really work properly, so I removed it and instead added a new `yarn lint:fix` command that will run all autofixing locally.


shanamatthews (approved):
Nice improvement, thanks!

PR #8487, opened 2023-11-14
feat(loader): update loader docs for new config

mydea:
Updates the docs for the loader to use the newly added `window.sentryOnLoad` config, which is better than `onLoad()` as it has better timing guarantees and also we do not need to check for existence of `Sentry`.

### DIFF
--- a/src/platforms/javascript/common/install/loader.mdx
+++ b/src/platforms/javascript/common/install/loader.mdx
@@ -74,6 +74,16 @@ If Performance Monitoring and/or Session Replay is enabled, the SDK will immedia
 
 ## SDK Configuration
 
+While the Loader Script will work out of the box without any configuration in your application, you can still configure the SDK according to your needs.
+
+### Default Configuration
+
+For Performance Monitoring, the SDK will be initialized with `tracesSampleRate: 1` by default. This means that the SDK will capture all traces.
+
+For Session Replay, the defaults are `replaysSessionSampleRate: 0.1` and `replaysOnErrorSampleRate: 1`. This means Replays will be captured for 10% of all normal sessions and for all sessions with an error.
+
+### Release Configuration
+
 You can configure the release by adding the following to your page:
 
 ```html
@@ -83,12 +93,6 @@ You can configure the release by adding the following to your page:
   };
 </script>
 ```
-
-### Default Configuration
-
-For Performance Monitoring the SDK will be initialized with `tracesSampleRate: 1` by default. This means that the SDK will capture all traces.
-
-For Session Replay, the defaults are `replaysSessionSampleRate: 0.1` and `replaysOnErrorSampleRate: 1`. This means Replays will be captured for 10% of all normal sessions and for all sessions with an error.
 
 ### Custom Configuration
 
@@ -115,18 +119,20 @@ Inside of the `window.sentryOnLoad` function, you can configure a custom `Sentry
 ```html
 <script>
   // Configure sentryOnLoad before adding the Loader Script
-    window.sentryOnLoad = function () {
-      Sentry.init({
-        release: " ... ",
-        environment: " ... "
-      });
-      Sentry.setTag(...);
-      // etc.
-    };
+  window.sentryOnLoad = function () {
+    Sentry.init({
+      release: " ... ",
+      environment: " ... "
+    });
+    Sentry.setTag(...);
+    // etc.
+  };
 </script>
 ```
 
-In addition to `window.sentryOnLoad`, you can also use `Sentry.onLoad()` to make sure the SDK is loaded before you call any other SDK methods. The loader will by default make sure that these functions can be called directly on `Sentry` at all times:
+### Guarding SDK Function Calls
+
+The loader will by default make sure that these functions can be called directly on `Sentry` at all times, even if the SDK is not yet loaded:
 
 - `Sentry.captureException()`
 - `Sentry.captureMessage()`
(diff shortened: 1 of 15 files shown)


src/platforms/javascript/common/install/loader.mdx
```diff
@@ -84,42 +84,71 @@ You can configure the release by adding the following to your page:
...
   crossorigin="anonymous"
 ></script>
-<script>
-  // Check for existence of Sentry in case Ad-blockers block the Sentry Loader Script
-  window.Sentry && Sentry.onLoad(function() { ... });
-</script>
 ```
 
-`Sentry.onLoad` is a function that only the loader provides, and - as the name suggests - it sets a function to be run once the full SDK has been loaded. In that function, you can configure your SDK exactly the way you would were you using the CDN, with one difference: your `Sentry.init` call doesn't need to include your DSN, since it's already been set. We recommend checking if `Sentry` exists before calling `Sentry.onLoad` in case the loader script failed to load.
+Inside of the `window.sentryOnLoad` function, you can configure a custom `Sentry.init()` call. You can configure your SDK exactly the way you would were you using the CDN, with one difference: your `Sentry.init()` call doesn't need to include your DSN, since it's already been set. Inside of this function, the full Sentry SDK is guaranteed to be loaded & available.
```
Lms24, line 113: l:
```suggestion
Inside of the `window.sentryOnLoad` function, you can configure a custom `Sentry.init()` call. You can configure your SDK exactly the way you would if you were using the CDN, with one difference: your `Sentry.init()` call doesn't need to include your DSN, since it's already been set. Inside of this function, the full Sentry SDK is guaranteed to be loaded & available.
```

src/platforms/javascript/common/install/loader.mdx
```diff
@@ -84,42 +84,71 @@ You can configure the release by adding the following to your page:
...
+      Sentry.setTag(...);
+      // etc.
+    };
 </script>
 ```
 
-For Performance Monitoring, by default, the SDK will be initialized with `tracesSampleRate: 1`. This means that the SDK will capture all traces.
-
-For Session Replay, the defaults are set to `replaysSessionSampleRate: 0.1` and `replaysOnErrorSampleRate: 1`. This means Replays will be captured for 10% of all normal sessions and for all sessions with an error.
+In addition to `window.sentryOnLoad`, you can also use `Sentry.onLoad()` to make sure the SDK is loaded before you call any other SDK methods. The loader will by default make sure that these functions can be called directly on `Sentry` at all times:
```
Lms24, line 129: Hmm I'm wondering if we should even still show this. Is there a concrete advantage users gain from doing it the old way? If not wdyt about deprecating the function and
a) adding a deprecation note to this section
b) entirely removing the function from the docs?

No strong opinions here though so logaf-l for me

Lms24 (approved):
LGTM, just have an optional suggestion (or rather question)

src/platforms/javascript/common/install/loader.mdx
```diff
@@ -84,42 +84,71 @@ You can configure the release by adding the following to your page:
...
+### Default Configuration
+
+For Performance Monitoring, by default, the SDK will be initialized with `tracesSampleRate: 1`. This means that the SDK will capture all traces.
+
+For Session Replay, the defaults are set to `replaysSessionSampleRate: 0.1` and `replaysOnErrorSampleRate: 1`. This means Replays will be captured for 10% of all normal sessions and for all sessions with an error.
+
+### Custom Configuration
+
+The loader script always includes a call to `Sentry.init` with a default configuration, including your DSN. If you want to [configure your SDK](../../configuration/options/) beyond that, you can configure a custom init call by defining a `window.sentryOnLoad` function. Whatever is defined inside of this function will _always_ be called first, before any other SDK method is called. Be sure to define this function _before_ you add the loader script, to ensure it can be called at the right time:
```
shanamatthews, line 95:
```suggestion
The loader script always includes a call to `Sentry.init` with a default configuration, including your DSN. If you want to <PlatformLink to="/configuration/options/">configure your SDK</PlatformLink> beyond that, you can configure a custom init call by defining a `window.sentryOnLoad` function. Whatever is defined inside of this function will _always_ be called first, before any other SDK method is called. Be sure to define this function _before_ you add the loader script, to ensure it can be called at the right time:
```

shanamatthews (approved):
Lgtm! Just a couple small suggestions

src/platforms/javascript/common/install/loader.mdx
mydea: So `onLoad` is only useful now for guarding for the SDK to be loaded - so this should _not_ be used to call `Sentry.init()` anymore, but only if you want to call any other SDK method. I tried to make this clear but apparently it is not yet so :D I'll tweak it a bit to hopefully clarify this!

mydea: I actually moved the `onLoad()` stuff into a dedicated new section "Guarding SDK Function Calls" to make this clearer! does that make sense to you?

src/platforms/javascript/common/install/loader.mdx
```diff
@@ -84,42 +94,68 @@ You can configure the release by adding the following to your page:
...
-  });
+  };
 </script>
 ```
 
-For Performance Monitoring, by default, the SDK will be initialized with `tracesSampleRate: 1`. This means that the SDK will capture all traces.
+### Guarding SDK Function Calls
 
-For Session Replay, the defaults are set to `replaysSessionSampleRate: 0.1` and `replaysOnErrorSampleRate: 1`. This means Replays will be captured for 10% of all normal sessions and for all sessions with an error.
+The loader will by default make sure that these functions can be called directly on `Sentry` at all times, even if the SDK is not yet loaded:
```
shanamatthews, line 135: My 2 cents - this section seems pretty clear to me now. I did make one small wording suggestion to make this sentence flow better.

```suggestion
By default, the loader will make sure you can call these functions directly on `Sentry` at any time, even if the SDK is not yet loaded:
```

PR #8492, opened 2023-11-14
fix(issues): remove note about DB on main thread for Android

kahest:
Remove note that DB on main thread perf issues is not available on Android - it's generally available now, see e.g. https://demo.sentry.io/issues/4327317740/?project=1801383

Note: This PR aligns the DB on main thread page with the other pages. We generally do not document compatibility of Mobile perf issues on the docs, but availability is intuitively subject to SDK version, frameworks, and configuration used (i.e. the File/DB APIs/frameworks used need to be instrumented by the SDK for the backend detector to be able to detect perf issues).


shanamatthews (approved):
👍

PR #8495, opened 2023-11-15
fix: Update docs for manual client creation

mydea:
We noticed these docs were actually incorrect, because you basically _have to_ call `new Hub(client)` - if you don't do this, integrations will not run, which will/may lead to unintended side effects. So updating this to just document this as "the" way to do this.

While at this, I also updated this to specifically call out browser extensions, as that is a common thing we get requests for every now and then, and also point out that in addition you should not setup any global integrations in such a scenario.

### DIFF
--- a/src/platforms/javascript/common/troubleshooting/index.mdx
+++ b/src/platforms/javascript/common/troubleshooting/index.mdx
@@ -277,19 +277,18 @@ You can now customize the hub to your liking, without affecting other hubs/clien
 
 ### Setting up Sentry in shared environments (e.g. Browser Extensions)
 
-When setting up Sentry in a shared environment - for example, in a browser extension or similar - where multiple Sentry instances may run,
-you should under no circumstances use `Sentry.init()`, as this will pollute the global state. If your browser extension uses `Sentry.init()`, and the website the extension is running on also uses Sentry, the extension may send events to the website's Sentry project, or vice versa.
+When setting up Sentry in a shared environment where multiple Sentry instances may run, for example, in a browser extension or similar, you should **not use `Sentry.init()`**, as this will pollute the global state. If your browser extension uses `Sentry.init()`, and the website the extension is running on also uses Sentry, the extension may send events to the website's Sentry project, or vice versa.
 
-For such scenarios, you _have to_ setup a client manually as seen in the example above.
-In addition to this, you should also avoid adding any integrations that use global state, like `Breadcrumbs` or `TryCatch`.
-As a rule of thumb, it's best to avoid using any integrations, and to rely on manual capture of errors only in such scenarios.
+For such scenarios, you have to set up a client manually as seen in the example above.
+In addition, you should also avoid adding any integrations that use global state, like `Breadcrumbs` or `TryCatch`.
+As a rule of thumb, it's best to avoid using any integrations and to rely on manual capture of errors only in such scenarios.
 
 ### Dealing with Integrations
 
 Integrations are setup on the `Client`, if you need to deal with multiple clients and hubs you have to make sure to also do the integration handling correctly.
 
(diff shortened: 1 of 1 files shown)


lforst (approved):

Lms24 (approved):

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -264,53 +265,31 @@ const client = new BrowserClient({
...
-}
+### Setting up Sentry in shared environments (e.g. Browser Extensions)
 
-hub.withScope(function (scope) {
-  hub.addBreadcrumb({ message: "crumb 2" });
-  hub.captureMessage("test2");
-});
-```
+When setting up Sentry in a shared environment - for example, in a browser extension or similar - where multiple Sentry instances may run,
+you should under no circumstances use `Sentry.init()`, as this will pollute the global state. If your browser extension uses `Sentry.init()`, and the website the extension is running on also uses Sentry, the extension may send events to the website's Sentry project, or vice versa.
```
shanamatthews, line 281:
```suggestion
When setting up Sentry in a shared environment where multiple Sentry instances may run, for example, in a browser extension or similar, you should **not use `Sentry.init()`**, as this will pollute the global state. If your browser extension uses `Sentry.init()`, and the website the extension is running on also uses Sentry, the extension may send events to the website's Sentry project, or vice versa.
```

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -264,53 +265,31 @@ const client = new BrowserClient({
...
+
+For such scenarios, you _have to_ setup a client manually as seen in the example above.
+In addition to this, you should also avoid adding any integrations that use global state, like `Breadcrumbs` or `TryCatch`.
+As a rule of thumb, it's best to avoid using any integrations, and to rely on manual capture of errors only in such scenarios.
```
shanamatthews, line 285:
```suggestion
For such scenarios, you have to set up a client manually as seen in the example above.
In addition, you should also avoid adding any integrations that use global state, like `Breadcrumbs` or `TryCatch`.
As a rule of thumb, it's best to avoid using any integrations and to rely on manual capture of errors only in such scenarios.
```

src/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -264,53 +265,31 @@ const client = new BrowserClient({
...
 
 ### Dealing with Integrations
 
 Integrations are setup on the `Client`, if you need to deal with multiple clients and hubs you have to make sure to also do the integration handling correctly.
-Here is a working example of how to use multiple clients with multiple hubs running global integrations.
+
+You should avoid doing this if you are using Sentry in a browser extension or in similar scenarios.
+If you know what you are doing, and you want to use global integrations nonetheless (e.g. in a microfrontend application), here is a working example of how to use multiple clients with multiple hubs running global integrations.
```
shanamatthews, line 292: What do you think of this edit? I wouldn't recommend saying "If you know what you're doing", since it's so non-specific. Users usually aren't good at judging if they know what they're doing :D

```suggestion
We do not recommend doing this if you are using Sentry in a browser extension or in similar scenarios.
If you can't avoid using global integrations (e.g. in a micro frontend application), here is a working example of how to use multiple clients with multiple hubs running global integrations.
```

shanamatthews (approved):
Just some small phrasing tweaks, otherwise lgtm!

src/platforms/javascript/common/troubleshooting/index.mdx
mydea: makes sense to me 👍

PR #8500, opened 2023-11-15
Monitor upsert support docs

### LINKED ISSUE #8482: Update Sentry CLI docs with monitor upsert instructions
Document the changes made in https://github.com/getsentry/sentry-cli/pull/1807

szokeasaurusrex:
## Pre-merge checklist

## Description of changes

This PR adds documentation for the new monitor upserts feature in the Sentry CLI, which allows users to create a cron monitor directly from the command line, without having to set it up in the Sentry web UI first.

Closes #8482

### DIFF
--- a/src/docs/product/crons/getting-started/cli.mdx
+++ b/src/docs/product/crons/getting-started/cli.mdx
@@ -3,13 +3,13 @@ title: CLI
 sidebar_order: 200
 ---
 
-<Include name="feature-stage-beta-crons.mdx" />
+        <Include name="feature-stage-beta-crons.mdx" />
 
-<Alert level="note" title="Deprecation Notice">
+        <Alert level="note" title="Deprecation Notice">
 
-Starting with v2.16.1 of the Sentry CLI, the ability to monitor check-ins using an auth token for authorization has been deprecated. Read on to learn how to update your CLI and use your project's DSN instead.
 
-</Alert>
+
+                </Alert>
 
 Sentry Crons allows you to monitor the uptime and performance of any scheduled, recurring job. Once implemented, it'll allow you to get alerts and metrics to help you solve errors, detect timeouts, and prevent disruptions to your service.
 
@@ -24,7 +24,7 @@ To begin monitoring your recurring, scheduled job:
 
 The Sentry CLI uses your Monitor's project DSN to authorize check-ins. To set it up, export the `SENTRY_DSN` environment variable:
 
-<SignInNote />
+        <SignInNote />
 
 ```bash
 export SENTRY_DSN=___PUBLIC_DSN___
@@ -32,7 +32,7 @@ export SENTRY_DSN=___PUBLIC_DSN___
 
 Alternatively, you can add it to your `~/.sentryclirc` config:
 
-<SignInNote />
+        <SignInNote />
 
 ```ini {filename:~/.sentryclirc}
 [auth]
@@ -57,6 +57,29 @@ sentry-cli monitors run my-monitor-slug -- python path/to/file.py
 
 ```bash {tabTitle: Node.JS}
 sentry-cli monitors run my-monitor-slug -- node path/to/file.js
+```
+
+### Configuring Your Monitor via the CLI
+
+You can also use the Sentry CLI to configure your cron monitor when you run your job. This way, you can avoid having to first set up the monitor through the Sentry web interface.
+
+Configure the cron monitor by providing the cron schdule in crontab format using the `--schedule` or the equivalent `-s` argument when executing the `sentry cli monitors run` command. Please make sure to enclose the schedule in quotes, so that your shell parses the argument correctly, like so:
+```bash
+sentry-cli monitors run --schedule "<expected schedule>" <monitor-slug> -- <command> <args>
+```
+
+When providing the `--schedule` argument, we also support the following optional arguments to provide additional configuration:
+  - `--check-in-margin`: The allowed margin of minutes after the expected check-in time that the monitor will not be considered missed for.
+  - `--max-runtime`: The allowed duration in minutes that the monitor may be in progress for before being considered failed due to timeout.
+  - `--timezone` *(beta, use cautiously!)*: A tz database string (e.g. "Europe/Vienna") representing the monitor's execution schedule's timezone. Please be **absolutely sure that you provide a valid tz database string** when using this argument; as of writing, the CLI lacks the capability to validate this string, and so **the monitor creation will silently fail when the string is invalid!** For now, we recommend verifying that the cron monitor was created successfully in Sentry when using this argument.
+
+Below are some usage examples:
+```bash {tabTitle: Every Minute}
+sentry-cli monitors run -s "* * * * *" -- my-command
+```
+
+```bash {tabTitle: Every Hour (+ Optional Arguments)}
+sentry-cli monitors run -s "0 * * * *" --check-in-margin 10 --max-runtime 5 --timezone "Europe/Vienna" -- my-command
 ```
 
 ### Specifying Monitor Environments (Optional)


src/docs/product/crons/getting-started/cli.mdx
```diff
@@ -59,6 +59,29 @@ sentry-cli monitors run my-monitor-slug -- python path/to/file.py
...
+```bash
+sentry-cli monitors run --schedule "<expected schedule>" <monitor-slug> -- <command> <args>
+```
+
+When providing the `--schedule` argument, we also support the following optional arguments to provide additional configuration:
+  - `--check-in-margin`: The allowed margin of minutes after the expected check-in time that the monitor will not be considered missed for.
+  - `--max-runtime`: The allowed duration in minutes that the monitor may be in progress for before being considered failed due to timeout.
+  - `--timezone` *(beta, use cautiously!)*: A tz database string (e.g. "Europe/Vienna") representing the monitor's execution schedule's timezone. Please be **absolutely sure that you provide a valid tz database string** when using this argument; as of writing, the CLI lacks the capability to validate this string, and so **the monitor creation will silently fail when the string is invalid!** For now, we recommend verifying that the cron monitor was created successfully in Sentry when using this argument.
```
szokeasaurusrex: Since this `--timezone` option can cause the entire monitor upsert to silently fail if an invalid value is provided, I have labelled it as a beta feature for now. Is this appropriate?

We would need to make sure we remove the "beta" notice and the warning about the CLI not validating the string once we close getsentry/relay#2718.

src/docs/product/crons/getting-started/cli.mdx
```diff
@@ -59,6 +59,32 @@ sentry-cli monitors run my-monitor-slug -- python path/to/file.py
...
+
+```bash
+sentry-cli monitors run --schedule "<expected schedule>" <monitor-slug> -- <command> <args>
+```
+
+When providing the `--schedule` argument, we also support the following optional arguments to provide additional configuration:
+
+- `--check-in-margin`: The allowed margin of minutes after the expected check-in time that the monitor will not be considered missed for.
+- `--max-runtime`: The allowed duration in minutes that the monitor may be in progress for before being considered failed due to timeout.
+- `--timezone` _(beta, use cautiously!)_: A tz database string (e.g. "Europe/Vienna") representing the monitor's execution schedule's timezone. Please be **absolutely sure that you provide a valid tz database string** when using this argument; as of writing, the CLI lacks the capability to validate this string, and so **the monitor creation will silently fail when the string is invalid!** For now, we recommend verifying that the cron monitor was created successfully in Sentry when using this argument.
```
szokeasaurusrex, line 76: Since this `--timezone` option can cause the entire monitor upsert to silently fail if an invalid value is provided, I have labelled it as a beta feature for now. Is this appropriate?

We would need to make sure we remove the "beta" notice and the warning about the CLI not validating the string once we close getsentry/relay#2718.

vivianyentran, line 76:
```suggestion
- `--timezone` _(beta, use cautiously!)_: A [tz database identifier string](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#List) (e.g. "Europe/Vienna") representing the monitor's execution schedule's timezone. Please be **absolutely sure that you provide a valid tz database string** when using this argument; as of writing, the CLI lacks the capability to validate this string, and so **the monitor creation will silently fail when the string is invalid!** For now, we recommend verifying that the cron monitor was created successfully in Sentry when using this argument.
```
Let's include a link to the tz database strings so that users have something to reference since it's important. Feel free to change the link if you have a better one!

src/docs/product/crons/getting-started/cli.mdx
```diff
@@ -59,6 +59,32 @@ sentry-cli monitors run my-monitor-slug -- python path/to/file.py
 sentry-cli monitors run my-monitor-slug -- node path/to/file.js
 ```
 
+### Configuring Your Monitor via the CLI
+
+You can also use the Sentry CLI to configure your cron monitor when you run your job. This way, you can avoid having to first set up the monitor through the Sentry web interface.
```
vivianyentran, line 64: This line feels weird in the middle of the page since it's general to the CLI and not specific to this section. I think you can either delete it or move it up to right before the "Requirements" section.

src/docs/product/crons/getting-started/cli.mdx
```diff
@@ -59,6 +59,32 @@ sentry-cli monitors run my-monitor-slug -- python path/to/file.py
...
+sentry-cli monitors run --schedule "<expected schedule>" <monitor-slug> -- <command> <args>
+```
+
+When providing the `--schedule` argument, we also support the following optional arguments to provide additional configuration:
+
+- `--check-in-margin`: The allowed margin of minutes after the expected check-in time that the monitor will not be considered missed for.
+- `--max-runtime`: The allowed duration in minutes that the monitor may be in progress for before being considered failed due to timeout.
+- `--timezone` _(beta, use cautiously!)_: A tz database string (e.g. "Europe/Vienna") representing the monitor's execution schedule's timezone. Please be **absolutely sure that you provide a valid tz database string** when using this argument; as of writing, the CLI lacks the capability to validate this string, and so **the monitor creation will silently fail when the string is invalid!** For now, we recommend verifying that the cron monitor was created successfully in Sentry when using this argument.
+
+Below are some usage examples:
```
vivianyentran, line 78: Nice! I like how you have two examples here

src/docs/product/crons/getting-started/cli.mdx
vivianyentran: I think the beta note makes sense

vivianyentran (approved):
Added some comments, but overall looks good! Thanks for adding to the docs!

src/docs/product/crons/getting-started/cli.mdx
szokeasaurusrex: Can you please explain why you think this line is out of place here? I think it is necessary to include this line here to introduce the purpose of this feature; that is, that you can configure your monitor via the CLI instead of configuring via the web interface. The line also indicates that using this feature is optional, since monitors can also be created and configured in the Sentry web interface.

However, I agree that we should probably mention the possibility of setting up the monitor via the CLI itself in the "Requirements" section, so I will add a mention there.

vivianyentran: Ah I thought it was implied from where the page lives (Crons -> Set Up -> CLI) that you were using the CLI to create / set up a new monitor, so it felt like something that should go at the top of the page. But now I see that the original feature was just for checking in on an existing monitor.

I think to make this clearer, you could keep this line here, but rename the section header to "Create or Update Your Monitor" similar to the section for [HTTP](https://docs.sentry.io/product/crons/getting-started/http/#creating-or-updating-a-monitor-through-a-check-in-optional).

Swatinem (approved):

src/docs/product/crons/getting-started/cli.mdx
```diff
@@ -59,6 +59,32 @@ sentry-cli monitors run my-monitor-slug -- python path/to/file.py
...
+
+```bash
+sentry-cli monitors run --schedule "<expected schedule>" <monitor-slug> -- <command> <args>
+```
+
+When providing the `--schedule` argument, we also support the following optional arguments to provide additional configuration:
+
+- `--check-in-margin`: The allowed margin of minutes after the expected check-in time that the monitor will not be considered missed for.
+- `--max-runtime`: The allowed duration in minutes that the monitor may be in progress for before being considered failed due to timeout.
+- `--timezone` _(beta, use cautiously!)_: A [tz database identifier string](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#List) (e.g. "Europe/Vienna") representing the monitor's execution schedule's timezone. Please be **absolutely sure that you provide a valid tz database string** when using this argument; as of writing, the CLI lacks the capability to validate this string, and so **the monitor creation will silently fail when the string is invalid!** For now, we recommend verifying that the cron monitor was created successfully in Sentry when using this argument.
```
evanpurkhiser, line 76: Why is this labeled as Beta?

evanpurkhiser, line 76: Can we add this timezone bit as a note instead of this long string at the end of the list item?

szokeasaurusrex: I added this to call folks' attention to the fact that using this argument can cause the upsert to fail silently; however, it is probably sufficient to just have the "use cautiously" label. The "beta" label is redundant, since all features described on this page are beta features, according to the note at the top of the page

szokeasaurusrex: Sure, a note was one of the options I also considered. Might be cleaner here

szokeasaurusrex:
@evanpurkhiser I made some changes to address the concerns you brought up, does it look good now?

PR #8513, opened 2023-11-16
feat(hybridcloud) Fetch DSNs from each region

markstory:
Because we won't have all the DSNs in a single region anymore, we need to fanout across each region that a user has membership in. The results from each region are joined together to make the final DSN/project context.

Refs HC-984

### DIFF
--- a/src/components/codeContext.tsx
+++ b/src/components/codeContext.tsx
@@ -56,6 +56,11 @@ type UserApiResult = {
   name: string;
 };
 
+type Region = {
+  name: string;
+  url: string;
+};
+
 // only fetch them once
 let cachedCodeKeywords: CodeKeywords | null = null;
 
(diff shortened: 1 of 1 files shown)


src/components/codeContext.tsx
```diff
@@ -123,38 +128,68 @@ const formatApiUrl = ({scheme, host}: Dsn) => {
   return `${scheme}${apiHost}/api`;
 };
 
+function getHost(): string {
+  if (process.env.NODE_ENV === 'development') {
+    return 'http://dev.getsentry.net:8000';
```
shanamatthews, line 133: I'm not familiar with dev.getsentry.net - what's the reason for pointing to that in develop?

shanamatthews (approved):
lgtm, just the one question

dashed (approved):

src/components/codeContext.tsx
markstory: `dev.getsentry.net` is the canonical hostname for the getsentry devserver. The docs support endpoint is only available in getsentry.

PR #8521, opened 2023-11-17
Reference tunnelRoute in Next.js tunnel docs

lforst:
Adds a quick note to the tunnel option docs to say that there is a quicker setup route in Next.js.


shanamatthews (approved):
lgtm!

Lms24 (approved):

PR #8557, opened 2023-11-23
Added a guide for using Sentry with WinUI 3

jamescrosswell:
## Pre-merge checklist

## Description of changes

Added a guide for using Sentry with WinUI 3 applications. In particular, the guide contains a section for users who would like to publish their applications using Ahead-of-Time compilation or Trimming enabled, as there are some additional steps that we can't automate using our integration once AOT is enabled. 

See:
- https://github.com/getsentry/sentry-dotnet/issues/2778

### DIFF
--- a/src/platforms/dotnet/guides/winui/index.mdx
+++ b/src/platforms/dotnet/guides/winui/index.mdx
@@ -1,0 +1,133 @@
+---
+title: WinUI 3
+description: "Learn about using Sentry's .NET SDK with the Windows UI Library (WinUI 3)"
+redirect_from:
+  - /platforms/dotnet/winui3/
+---
+
+Sentry's .NET SDK works with Windows UI Library applications through the [Sentry NuGet package](https://www.nuget.org/packages/Sentry).
+
+## Configuration (for Just-In-Time compilation)
+
+If you are using Just-In-Time (JIT) compilation, which will be the case for most users, all you need to do is [initialize the SDK with `SentrySdk.Init`](/platforms/dotnet/) in the constructor of your application class (usually `App.xaml.cs`).
+
+<Note>
+
+Do not initialize the SDK in the `OnLaunched` event of the application or the `Hub` will not be initialized correctly.
+
+</Note>
+
+<SignInNote />
+
+```csharp
+// Add these to your existing using statements.
+using Sentry;
+using Sentry.Protocol;
+
+sealed partial class App : Application
+{
+    public App()
+    {
+        // Initialize Sentry in the App constructor before any other code, to ensure you capture all possible exceptions.
+        SentrySdk.Init(o =>
+        {
+            // Tells which project in Sentry to send events to:
+            o.Dsn = "___PUBLIC_DSN___";
+
+            // When configuring for the first time, to see what the SDK is doing:
+            o.Debug = true;
+
+            // Set traces_sample_rate to 1.0 to capture 100% of transactions for performance monitoring.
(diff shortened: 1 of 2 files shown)


src/platforms/dotnet/guides/winui/index.mdx
```diff
@@ -0,0 +1,133 @@
+---
+title: WinUI 3
+description: "Learn about using Sentry's .NET SDK with the Windows UI Library (WinUI 3)"
+redirect_from:
+  - /platforms/dotnet/winui3/
```
bruno-garcia, line 5: How does it differ from UWP?

Their docs say:

> The Windows UI Library (WinUI) is a native user experience (UX) framework for both Windows desktop and UWP applications.

So maybe different. But below says WinUI2 as UWP and WinUI3 can be UWP or "desktop". I'm super confused with this stuff

https://docs.sentry.io/platforms/dotnet/guides/uwp/

```suggestion
  - /platform/dotnet/guides/uwp/
```

src/platforms/dotnet/guides/winui/index.mdx
```diff
@@ -0,0 +1,133 @@
...
+---
+
+Sentry's .NET SDK works with Windows UI Library applications through the [Sentry NuGet package](https://www.nuget.org/packages/Sentry).
+
+## Configuration (for Just-In-Time compilation)
+
+If you are using Just-In-Time (JIT) compilation, which will be the case for most users, all you need to do is [initialize the SDK with `SentrySdk.Init`](/platforms/dotnet/) in the constructor of your application class (usually `App.xaml.cs`).
```
bruno-garcia, line 12: > which will be the case for most users

Is this true? UWP is only AOT. Perhaps what they say about "Desktop" means just .NET (JIT). So maybe we need to reword here to be specific when it's going to be JIT and when it's AOT (particularly now that even .NET has AOT as first-class with NativeAOT).

bruno-garcia (commented):
@lucas-zimerman knows about UWP maybe he can help review here

src/platforms/dotnet/guides/winui/index.mdx
```diff
@@ -0,0 +1,133 @@
+---
+title: WinUI 3
```
bruno-garcia, line 2: I'd also recommend dropping the version from the title.

Just mention WinUI 3 below if needed, if we don't support WinUI2 (I believe we do, and we call that UWP), we might want to have 2 sections in the doc for v2 and v3

jamescrosswell: Yeah I think WinUI 2 is UWP (which has it's own documentation). You reckon we could/should merge these? At the moment there are two different sets of guidance in this page (JIT and AOT)... I guess if we merge it with UWP then there would be 3 sets of guidance in the one page. I think it might be easier to leave the UWP instructions separate, personally.

bruno-garcia: merge and redirect yeah, that'd be best I think

shanamatthews (approved):
This all looks good from the docs side!

src/platforms/dotnet/guides/winui/index.mdx
shanamatthews: I don't have strong feelings about this either. It seems clear as-is to me, but I see bruno's argument.

bruno-garcia (approved):

bruno-garcia:
Thanks James!

PR #8562, opened 2023-11-23
fix(cordova): fix docs around Cordova Perf and Profiling support

kahest:
mark Cordova not supported by browser profiling
fix Performance support (hide auto-instrumentation, custom init snippet,...)

### DIFF
--- a/src/platform-includes/getting-started-primer/javascript.cordova.mdx
+++ b/src/platform-includes/getting-started-primer/javascript.cordova.mdx
@@ -18,4 +18,4 @@ Features:
 - On-device symbolication for JavaScript (in `Debug` mode)
 - [Ionic support](/platforms/javascript/guides/cordova/ionic/)
 - Under the hood the SDK relies on our [JavaScript SDK](/platforms/javascript/), which makes all functions available for JavaScript also available in this SDK including:
-  - [Performance Monitoring](/product/performance/) to create transactions
+  - [Performance Monitoring](/platforms/javascript/guides/cordova/performance/instrumentation/custom-instrumentation) to create transactions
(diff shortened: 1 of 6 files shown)


lucas-zimerman (approved):
LGTM!

bruno-garcia:
@kahest Nice change! 
Obviously not my call but I suggest we take a step further and remove the performance monitoring parts for cordova too.

Keeping it up-to-date with bug fixes etc makes sense but adding new things will introduce new use cases for us to debug

bruno-garcia:
We also removed Session Replay from it

lucas-zimerman:
> @kahest Nice change! Obviously not my call but I suggest we take a step further and remove the performance monitoring parts for cordova too.
> 
> Keeping it up-to-date with bug fixes etc makes sense but adding new things will introduce new use cases for us to debug

It's only the minimal docs for performance monitoring, the idea wasn't to add support to it, but to at least document a feature that already existed since 2021 https://github.com/getsentry/sentry-cordova/pull/216

kahest:
> @kahest Nice change! Obviously not my call but I suggest we take a step further and remove the performance monitoring parts for cordova too.
> 
> Keeping it up-to-date with bug fixes etc makes sense but adding new things will introduce new use cases for us to debug

As @lucas-zimerman mentioned, this PR aims to add only the custom perf docs because that's what we've been supporting for a while, and we always mentioned supporting perf on the landing page of the SDK docs, but linked to generic product perf docs. In case we decided we don't support perf for Cordova, I'm all for not exposing it in the SDK and completely removing it from the docs. Wdyt?

krystofwoldrich (approved):

bruno-garcia (approved):
makes sense. thanks everyone

PR #8528, opened 2023-11-18
feat(feedback): Add SDK docs for User Feedback

billyvg:
Adds SDK docs for new User Feedback widget (reorgs to support both old and new widgets). Adds setup page in product docs.

Closes https://github.com/getsentry/sentry-docs/issues/8494

PR #8590, opened 2023-11-27
ruby on rails: user feedback modal

bruno-garcia:
We support the modal on RoR but not on Ruby, correct?
Do we have the logic inverted?

Please note we're moving docs around: https://github.com/getsentry/sentry-docs/pull/8528/files


sl0thentr0py:
it's a double negative, so what we have is correct :)
This is saying feedback is supported on rails, not on ruby and the snippet saying so is being added to ruby and not rails.

PR #8594, opened 2023-11-28
docs(browser-starfish): add note on resource parameterization

DominikB2014:
Add a note that any changes in parameterization might result in both old and new groups being present.

PR #8595, opened 2023-11-28
docs(browser-starfish): add note on resource parameterization 

DominikB2014:
Add a note that any changes in parameterization might result in both old and new groups being present.

Update to this #8594, Didn't realize it would auto merge without approval.


AbhiPrasad (approved):
yeah you can merge without approval in docs.

Issue #8560, opened 2023-11-23
Add warning the Browser JS source map guide doesn't work for Hermes bytecode


krystofwoldrich:
### Core or SDK?

Platform/SDK

### Which part? Which one?

Browser JS/React Native

### Description

Currently, the Browser JS Source Map page doesn't point to React Native. But based on the incoming issues. Users are trying to use it on RN Hermes for which it doesn't work.

- https://github.com/getsentry/sentry-cli/issues/1824

https://docs.sentry.io/platforms/javascript/sourcemaps/

### Suggested Solution

Add links to RN guides.

Add https://docs.sentry.io/platforms/react-native/sourcemaps/

PR #8618, opened 2023-12-01
ref(sourcemaps): adds RN note to browser JS source maps page

kahest:
fixes https://github.com/getsentry/sentry-docs/issues/8560


shanamatthews (approved):
👍

PR #8633, opened 2023-12-05
[Python] On main page link to integrations list

antonpirker:
Bring back the note about frameworks in Python. Because Python has no guides anymore, we link to the integrations list instead.


shanamatthews (approved):
Looks good! Just a small tweak.

PR #8649, opened 2023-12-05
ref(replay): add onboarding docs for astro and fix capacitor doc

michellewzhang:


src/wizard/javascript/replay-onboarding/astro/2.configure.md
```diff
@@ -0,0 +1,28 @@
...
+  dsn: "___PUBLIC_DSN___",
+
+  // This sets the sample rate at 10%. You may want this to be 100% while
+  // in development, then sample at a lower rate in production.
+  replaysSessionSampleRate: 0.1,
+  // If the entire session is not sampled, use the below sample rate to sample
+  // sessions when an error occurs.
+  replaysOnErrorSampleRate: 1.0,
+
+  integrations: [new Replay()],
```
bruno-garcia, line 26: In these integrations part of the monorepo it looks like this:
Example React: https://docs.sentry.io/platforms/javascript/guides/react/session-replay/
```suggestion
import * as Sentry from "@sentry/astro";

Sentry.init({
  dsn: "___PUBLIC_DSN___",

  // This sets the sample rate at 10%. You may want this to be 100% while
  // in development, then sample at a lower rate in production.
  replaysSessionSampleRate: 0.1,
  // If the entire session is not sampled, use the below sample rate to sample
  // sessions when an error occurs.
  replaysOnErrorSampleRate: 1.0,

  integrations: [new Sentry.Replay()],
```

bruno-garcia (approved):

PR #8657, opened 2023-12-06
feat(profiling): add differential flamegraph docs

JonasBa:
New docs on differential flamegraph feature. 

@shanamatthews, do we auto optimize these images or should I do this manually?

_Please **do not** merge as the feature isn't actually released yet_


src/docs/product/profiling/differential-flamegraphs.mdx
```diff
@@ -0,0 +1,47 @@
...
+
+Let us see an example of how our module lock problem looks like when using the profiling data from after the regression as the source of truth:
+
+![Differential flamegraph new functions](./differential-flamegraph-regular.png)
+
+And now lets switch to using the data from before the regression as our source of truth
+
+![Differential flamegraph new functions](./differential-flamegraph-inverted.png)
+
+We can see here that `ModuleLock.acquire` was faster before the regression, but we can also see some of the function calls inside our `ChunkUploadEndpoint`, which are no longer being called.
```
lizokm, line 39:
```suggestion
By using the `Before -> After` toggle we can see that `ModuleLock.acquire` was faster before the regression, but we can also see some of the function calls inside our `ChunkUploadEndpoint`, which are no longer being called.
```

src/docs/product/profiling/differential-flamegraphs.mdx
```diff
@@ -0,0 +1,47 @@
...
+
+Besides the color scheme and a different source of data, differential flamegraphs follow a similar UI pattern as aggregated flamegraphs. You can pan/zoom around the chart and search for frames like in a regular flamegraph.
+
+The table under the differential flamegraph represents the largest changes in function durations. The functions on the left have increased in duration while the functions on the right have improved in duration.
+
```
lizokm, line 44:
```suggestion
While differential flamegraphs in Sentry use a different color scheme and pull from a different source of data than regular flamegraphs, you can pan and zoom around the chart and search for frames the same way you do with regular flamegraph.

The table under the differential flamegraph shows both the largest increases and the largest decreases in function duration.

```

src/docs/product/profiling/differential-flamegraphs.mdx
```diff
@@ -0,0 +1,47 @@
...
+![Differential flamegraph table](./differential-flamegraph-table.png)
+
+Differential flamegraphs are a powerful feature that allow us to quickly visualize potential sources of performance regressions or improvements and while they are currently only available as part of the automated function regression workflow, we will be adding them to other parts of the profiling workflow.
```
lizokm, line 47:
```suggestion
Differential flamegraphs are a good way to quickly visualize potential sources of performance regressions or improvements. Currently, they're only available as part of the automated function regression workflow, but we're working on adding them to other parts of the profiling workflow.
```

lizokm (approved):
This looks great! I've made some language tweaks, and there's one part I'd like to clarify so I've scheduled a few minutes for us to talk on Monday :)

src/docs/product/profiling/differential-flamegraphs.mdx
```diff
@@ -0,0 +1,53 @@
...
+sidebar_order: 110
+description: "Learn how to use and interpret differential flamegraphs"
+---
+
+There are multiple ways to visualize and use profiling data, including differential flamegraphs, which can help easily identify function regressions or changes in the execution context.
+
+## What Are Differential Flamegraphs?
+
+Differential flamegraphs are graph visualizations that highlight how the call stack frequencies of a program have changed, a task which can be tedious to figure out without them. They're often used in analysis for finding function regressions and understanding which functions got slower, faster, or were newly added. Keep in mind that differential flamegraphs aren't really a brand new concept, but just aggregate flamegraphs, drawn using the colors generated by the diff of the two.
+
```
lizokm, line 12:
```suggestion
Differential flamegraphs are graph visualizations that highlight how the call stack frequencies of a program have changed, a task which can be tedious to figure out without them. They're often used in analysis for finding function regressions and understanding which functions got slower, faster, or were newly added. Keep in mind that differential flamegraphs aren't really a new concept, but just aggregate flamegraphs, drawn using the colors generated by the diff of the two.

```

src/docs/product/profiling/differential-flamegraphs.mdx
```diff
@@ -0,0 +1,53 @@
...
+sidebar_order: 110
+description: "Learn how to use and interpret differential flamegraphs"
+---
+
+There are multiple ways to visualize and use profiling data, including differential flamegraphs, which can help easily identify function regressions or changes in the execution context.
+
+## What Are Differential Flamegraphs?
+
+Differential flamegraphs are graph visualizations that highlight how the call stack frequencies of a program have changed, a task which can be tedious to figure out without them. They're often used in analysis for finding function regressions and understanding which functions got slower, faster, or were newly added. Keep in mind that differential flamegraphs aren't really a new concept, but just aggregate flamegraphs, drawn using the colors generated by the diff of the two.
+
```
lizokm, line 12:
```suggestion
Differential flamegraphs are graph visualizations that highlight how the call stack frequencies of a program have changed, a task which can be tedious without them. They're often used in analysis for finding function regressions and understanding which functions got slower, faster, or were newly added. Keep in mind that differential flamegraphs aren't really a new concept, but just aggregate flamegraphs drawn using the colors generated by the diff of the two.

```

src/docs/product/profiling/differential-flamegraphs.mdx
```diff
@@ -0,0 +1,53 @@
...
+In Sentry, differential flamegraphs are shown whenever a function regression is detected. When a regression happens, Sentry generates two aggregate flamegraphs. One from the period leading up to the regression, and another from the time right after it. A differential flamegraph is then used to visualize the data.
+
```
lizokm, line 14:
```suggestion
In Sentry, differential flamegraphs are shown whenever a function regression is detected. When a regression happens, Sentry generates two aggregate flamegraphs. One from the period leading up to the regression, and another from the time right after it. A differential flamegraph is then used to visualize the difference between the two.

```

src/docs/product/profiling/differential-flamegraphs.mdx
```diff
@@ -0,0 +1,53 @@
...
+The differential flamegraph above is showing that the function call to `ModuleLock.acquire` has increased in duration, hinting at a lock contention problem which is slowing down the code.
+
+Let's look at another example:
+
+![Differential flamegraph new functions](./differential-flamegraph-new.png)
+
+The above differential flamegraph shows that new function calls to `OrganizationTeamsEndpoint.dispatch` were introduced, which made the code run slower, while calls to `ChunkUploadEndpoint.get` decreased.
+
+If you look closely, you'll notice that the top right corner has a `Before -> After` and `After -> Before` toggle. This is because by default, differential flamegraphs draw the aggregate flamegraph from after a regression has occured as the source of truth. This means that any code that may have been removed, will no longer be drawn (after all, it's not there anymore). This is why negating the view is useful, as it allows us to compare the data from before and peek into the future of how our code will change.
+
```
lizokm, line 34:
```suggestion

The top left corner has a `Before -> After` and `After -> Before` toggle. This is because it's helpful to be able to view data from **before a regression occurred** and compare it with data after the regression. This is a different view than a differential flamegraph, which shows the aggregate flamegraph drawn right **after a regression has occurred**. 

```
@JonasBa let me know if this makes sense :)

src/docs/product/profiling/differential-flamegraphs.mdx
```diff
@@ -0,0 +1,53 @@
...
+``
+I think I need 10 minutes of your time to talk through this with you. I'm not sure I'm understanding it correctly.
+Here's an example of what our module lock problem looks like when using the profiling data from after the regression has occurred as the source of truth:
+
```
lizokm, line 38:
```suggestion
``
Here's an example of what our module lock problem looks like when using the profiling data from after the regression has occurred as the source of truth:

```

lizokm (approved):
Looking good! Made just a few more tweaks, but I think it's ready to go.

Issue #8324, opened 2023-10-24
Document Rage and Dead Click 


bruno-garcia:
Main goals:

- [ ] The explanation for how Dead and Rage clicks are detected
- [ ] SDK options, such as `slowClickIgnoreSelectors` (https://github.com/getsentry/sentry-javascript/issues/9087)
- [ ] Built-in attempts to ignore false positives (e.g: https://github.com/getsentry/getsentry/pull/11728)
- [ ] Known false positives for dead click: Browser download
- [ ] If we can fit it in: Document good UX suggestions like loading a spinner on a paging component to avoid triggering Rage Click (if user doesn't see UI changes, it's bad UX)

Note we've added R+D Click here to product docs: https://docs.sentry.io/product/session-replay/replay-page-and-filters/

Basically: 
> Dead Clicks: User clicks on a and button tags that do not result in any page activity after 7 seconds (i.e. no HTML was added, removed, or updated; no visual changes were observed in the page).
> Rage Clicks: Five or more clicks on a dead element (it exhibits no page activity after 7 seconds.) Rage clicks are a subset of dead clicks.

And we link to that page from the [changelog announcement of the new cards](https://changelog.getsentry.com/announcements/top-selectors-with-rage-and-dead-clicks-in-session-replay).

PR #8658, opened 2023-12-07
feat(replay): Add slow click detection docs

mydea:
This adds the `slowClickIgnoreSelectors` config option, and also a section to troubleshooting explaining how this works - not sure if that is the best place, but can't think of a better one...?

Closes https://github.com/getsentry/sentry-docs/issues/8324


src/platforms/javascript/common/session-replay/troubleshooting.mdx
```diff
@@ -62,6 +62,20 @@ Because of the complexity of the browser environment, there's a significant amou
...
+
+Another concept related to this is "Rage Clicks". A "Rage Click" is when a user clicks many times on an element, which may indicate frustration that something is not working.
+
+When a Slow Click and a Rage Click come together, we have a very good indicator that something is actually wrong, and the Replay SDK will generate a "Rage Click" for this.
+
+Note that there are some limitations to this. On the one hand, there may be _false negatives_ - we cannot detect all cases, for example, imagine a user clicks a "dead" button, but then clicks a different button afterwards that leads to a DOM update. In this case, since we cannot attribute the DOM update to a specific button, we'll _not_ capture this slow click.
+
+On the other hand, there may be _false positives_. For example, if a user clicks a link that leads to a file download, there is no reliable way to detect that a download was initiated, so a Slow Click may be generated even if the button is not actually "dead". For these cases, you can configure the SDK via `slowClickIgnoreSelectors` - see <PlatformLink to='/session-replay/configuration'>Configuration</PlatformLink> for more details.
```
bruno-garcia, line 77: How about some examples?

````suggestion
On the other hand, there may be _false positives_. For example, if a user clicks a link that leads to a file download, there is no reliable way to detect that a download was initiated, so a Slow Click may be generated even if the button is not actually "dead". For these cases, you can configure the SDK via `slowClickIgnoreSelectors` - see <PlatformLink to='/session-replay/configuration'>Configuration</PlatformLink> for more details.

For example, ignoring detection of Dead and Rage clicks on selectors:

```javascript
new Sentry.Replay({
	slowClickIgnoreSelectors: [
		".download" // All buttons with class Download. 
		"a[label=download]"
	].
});
```
````

File: src/platforms/javascript/common/session-replay/troubleshooting.mdx [hunk 2]
```diff
@@ -62,6 +62,20 @@ Because of the complexity of the browser environment, there's a significant amou
 
 We're always working on ways to reduce the bundle size. Additionally, there are steps you can take in order to reduce the size of Session Replay based on your specific use case. See [Tree Shaking](../../configuration/tree-shaking/) for more information.
 
+## How does Rage Click detection work?
+
+The Replay SDK tries to detect cases where a user clicked on something, but nothing happened. This is called a "Dead Click", a click that seems to be "dead" and does nothing, from the users perspective. This can happen if a bug or similar prevents the UI from being correctly updated after a button is clicked.
```
michellewzhang, line 67:
```suggestion
The Replay SDK tries to detect cases where a user clicked on something, but nothing happened. This is called a "Dead Click" -- a click that seems to be "dead" and does nothing, from the user's perspective. This can happen if a bug or similar prevents the UI from being correctly updated after a button is clicked.
```

src/platforms/javascript/common/session-replay/troubleshooting.mdx
```diff
@@ -62,6 +62,20 @@ Because of the complexity of the browser environment, there's a significant amou
...
+
+The way the Replay SDK detects this is by checking if a click on a `<button>` or `<a>` element does not lead to any updates of the DOM, or a page scroll. If within 7 seconds no updates are detected, the SDK will generate a "Slow Click" in the Replay UI. We call it a "Slow Click" because it may be possible that something will happen after the 7s, but we chose this as a cutoff time.
```
michellewzhang, line 69:
```suggestion
The way that the Replay SDK detects this is by checking if a click on a `<button>` or `<a>` element does not lead to either updates to the DOM or a page scroll. If within 7 seconds no updates are detected, then the SDK will generate a "Slow Click" in the Replay UI. We call it a "Slow Click" because it may be possible that something will happen after the 7s, but we chose this as a cutoff time.
```

src/platforms/javascript/common/session-replay/troubleshooting.mdx
```diff
@@ -62,6 +62,20 @@ Because of the complexity of the browser environment, there's a significant amou
...
+
+Another concept related to this is "Rage Clicks". A "Rage Click" is when a user clicks many times on an element, which may indicate frustration that something is not working.
+
+When a Slow Click and a Rage Click come together, we have a very good indicator that something is actually wrong, and the Replay SDK will generate a "Rage Click" for this.
+
+Note that there are some limitations to this. On the one hand, there may be _false negatives_ - we cannot detect all cases, for example, imagine a user clicks a "dead" button, but then clicks a different button afterwards that leads to a DOM update. In this case, since we cannot attribute the DOM update to a specific button, we'll _not_ capture this slow click.
```
michellewzhang, line 75:
```suggestion
Note that there are some limitations to this. On the one hand, there may be _false negatives_ -- we cannot detect all cases; for example, imagine a user clicks a "dead" button, but then clicks a different button afterwards that leads to a DOM update. In this case, since we cannot attribute the DOM update to a specific button, we'll _not_ capture this slow click.
```

lizokm (commented):
Hey there, apologies for the delay. Shana and I are both out sick. We'll review this as soon as possible. Thanks for being patient!

lizokm (approved):
Made a few small tweaks, but otherwise this looks good!

src/platforms/javascript/common/session-replay/troubleshooting.mdx
```diff
@@ -62,6 +62,32 @@ Because of the complexity of the browser environment, there's a significant amou
...
+
+When a Slow Click and a Rage Click come together, we have a very good indicator that something is actually wrong, and the Replay SDK will generate a "Rage Click" for this.
+
+Note that there are some limitations to this. On the one hand, there may be _false negatives_ -- we cannot detect all cases; for example, imagine a user clicks a "dead" button, but then clicks a different button afterwards that leads to a DOM update. In this case, since we cannot attribute the DOM update to a specific button, we'll _not_ capture this slow click.
+
+On the other hand, there may be _false positives_. For example, if a user clicks a link that leads to a file download, there is no reliable way to detect that a download was initiated, so a Slow Click may be generated even if the button is not actually "dead". For these cases, you can configure the SDK via `slowClickIgnoreSelectors` - see <PlatformLink to='/session-replay/configuration'>Configuration</PlatformLink> for more details.
+
+For example, ignoring detection of Dead and Rage clicks on selectors:
```
mydea, line 79:
```suggestion
For example, you could ignore detection of dad and rage clicks for download links in your application like this:
```

PR #8677, opened 2023-12-09
environment: default and env var

bruno-garcia:
Ideally we mention the env var only where its applicable. Most SDKs, when running via the command line, for example. But not really mobile apps.

### DIFF
--- a/src/platforms/common/configuration/environments.mdx
+++ b/src/platforms/common/configuration/environments.mdx
@@ -6,6 +6,12 @@ description: "Learn how to configure your SDK to tell Sentry about your environm
 
 Environments tell you where an error occurred, whether that's in your production system, your staging server, or elsewhere.
 
+<Note>
+
+The default value is `production`.
+
+</Note>
+
 Sentry automatically creates an environment when it receives an event with the `environment` parameter set.
 
 Environments are case sensitive. The environment name can't contain newlines, spaces or forward slashes, can't be the string "None", or exceed 64 characters. You can't delete environments, but you can [hide](/product/sentry-basics/environments/#hidden-environments) them.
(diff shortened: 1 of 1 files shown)


src/platforms/common/configuration/environments.mdx
```diff
@@ -6,10 +6,18 @@ description: "Learn how to configure your SDK to tell Sentry about your environm
 
 Environments tell you where an error occurred, whether that's in your production system, your staging server, or elsewhere.
 
+<Note>
+
+The default value is `production`.
```
lizokm, line 11:
```suggestion
The default value is `production`.
```
I'm not sure I understand this correctly. Are we trying to tell the user that the default environment is production? If so, why would there be a default since this is supposed to tell you where an error occurred?

lizokm (commented):
Need a bit of clarification :)

src/platforms/common/configuration/environments.mdx
bruno-garcia: Because without a default value it was challenging in the production, so we needed to have a value and since Sentry runs in prod, by default we went with `production`. 

It's documented here but can't link to anchor (no anchor): https://develop.sentry.dev/sdk/event-payloads/

image

bruno-garcia: PR that added to develop:
* https://github.com/getsentry/develop/pull/656

Unfortunately I didn't add the context of the discussion (which happened on DIscord with @dcramer and @mitsuhiko)

shanamatthews (commented):
Hey @bruno-garcia - This file no longer exists due to the refactor Stephanie did to prep for the Next.js rewrite transition, and has been pulled out into separate mdx files for each platform. 

To make these docs make more sense, I think we should clarify this line:

> Sentry automatically creates an environment when it receives an event with the `environment` parameter set.

Since there's a default value set, does that mean that if a user doesn't set `environment`, all events will be tagged as `production` and Sentry will automatically create a corresponding "default" environment?

I think that's the kind of information we should add to this docs page, as well as the product page for environments (/product/sentry-basics/concepts/environments/)

bruno-garcia:
> Since there's a default value set, does that mean that if a user doesn't set environment, all events will be tagged as production and Sentry will automatically create a corresponding "default" environment?

That's correct. The default environment is `production`.

> I think that's the kind of information we should add to this docs page, as well as the product page for environments (/product/sentry-basics/concepts/environments/)

Wouldn't hurt being there too.

PR #8681, opened 2023-12-11
Add `wrapRemixHandleError` instructions

onurtemizkan:
Ref: https://github.com/getsentry/sentry-javascript/pull/9725

Updated `handleError` instructions to reflect the new feature.

Also updated external links from Remix v1 documentation to v2.


lizokm (approved):
Made a few language tweaks, but otherwise this looks great!

jgentes:
I can file a bug for this, but using v7.91.0 of @sentry/remix generates: `No matching export in "node_modules/@sentry/remix/esm/index.client.js" for import "wrapRemixHandleError"`. Surprisingly v7.87.0 also has the same error?

jgentes:
I got it, the fix is to use `export const handleError = Sentry.wrapRemixHandleError` instead.

PR #8701, opened 2023-12-12
Clarify window.SENTRY_RELEASE

lforst:
Supersedes https://github.com/getsentry/sentry-docs/pull/8679

the SDK does not read the value of window.SENTRY_RELEASE it releases the value of window.SENTRY_RELEASE.id

[getsentry/sentry-javascript@b3f8d9b/packages/browser/src/sdk.ts#L111](https://github.com/getsentry/sentry-javascript/blob/b3f8d9bcb067b391c1d0eba4f537ff09be56c1e9/packages/browser/src/sdk.ts#L111)

### DIFF
--- a/src/platforms/android/configuration/options.mdx
+++ b/src/platforms/android/configuration/options.mdx
@@ -111,7 +111,7 @@ Sets the release. Some SDKs will try to automatically configure a release out of
 
 <PlatformSection notSupported={["unity", "javascript.electron"]}>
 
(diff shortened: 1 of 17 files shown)


Lms24 (approved):
LGTM generally but naive question: Why do we mention the browser at all in all these docs?

lforst:
@Lms24 I think the reason this is everywhere is because we're in the process of making everything forwards compatible for the Next.js rewrite: https://github.com/getsentry/sentry-docs/pull/8686

Issue #8619, opened 2023-12-01
Specify the maximum number of metric alerts in docs


scefali:
We need to document the maximum number of metric alerts an organization can have (1000) has which is determined by this code: https://github.com/getsentry/sentry/blob/master/src/sentry/conf/server.py#L3760

PR #8704, opened 2023-12-12
Document metric alert org max

snigdhas:
Document the maximum number of metric alerts allowed for an organization.

Fixes https://github.com/getsentry/sentry-docs/issues/8619


lizokm (approved):
LGTM

PR #8708, opened 2023-12-13
flutter: improve getting started

buenaflor:
The getting started page is currently way too cluttered with information and it takes about 2-3 full scrolls until you get to actual productive section.

Currently the features list grows too large and creates more maintenance work for essentially duplicate information.


lizokm (approved):
Great idea to make this update!

PR #8722, opened 2023-12-14
ref: Add link to explicit tag syntax from tag pages

jangjodi:
Add link to explicit tag syntax from tag pages

related to https://github.com/getsentry/sentry/issues/61595

### DIFF
--- a/src/platforms/android/enriching-events/tags/index.mdx
+++ b/src/platforms/android/enriching-events/tags/index.mdx
@@ -19,7 +19,7 @@ Define the tag:
 
 <Note>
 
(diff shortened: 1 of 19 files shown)


isabellaenriquez (approved):
🙏🏽 

i think the slash needs to be removed from the end of the anchor (at the very least it didn't take my to the anchor when i used the version with the slash)

vivianyentran (approved):
nit, but I would change the link to `/product/reference/search/#explicit-tag-syntax` so that users don't think the link goes to an external website. Also I had the same issue as Isabella with the link, so hopefully removing the / at the end fixes it!

PR #8756, opened 2023-12-19
docs(mobile-spans): Add documentation for ttid/ttfd contributing spans

shruthilayaj:
Documentation for the TTID/TTFD contribution
column on the mobile page

Screenshot 2023-12-19 at 12 09 01 PM


src/docs/product/performance/mobile-vitals.mdx
```diff
@@ -124,6 +124,12 @@ At the bottom of the page, you’ll see a table of spans that occurred during sc
 - db.sql.query
 - db.sql.transaction
 
+### TTID and TTFD affecting spans:
+
+Spans on the **Screen Details** page are tagged if they end before TTID or TTFD, as it could mean they are affecting the respective metrics. Just because a span has been identified as TTID ir TTFD affecting it doesn't necessariliy mean it's the culprit. This is an approximation and a tool for you to identify on an aggregate level if a span may be affecting TTID or TTFD and narrow down the scope of spans to investigate.
+
```
lizokm, line 130:
```suggestion
If spans end before TTID or TTFD there's a chance they could be affecting their respective metrics. They're  tagged in Sentry and can be viewed on the **Screen Details** page. But just because a span has been identified as TTID or TTFD-affecting, it doesn't necessarily mean it's the culprit. This information is an approximation and should be used to find potentially problematic spans on an aggregate level and then narrow down the scope of spans to investigate.

```

lizokm (approved):
Made a few tweaks, but otherwise this looks great. Love the image!

PR #8760, opened 2023-12-20
kmp: add version compatibility table

buenaflor:
With more and more KMP SDK releases coming, we need to be able to track which KMP SDK version is compatible with which Cocoa SDK version. 

Ideally we can automate this in the future but now it's the best alternative we can do.

Closes https://github.com/getsentry/sentry-kotlin-multiplatform/issues/159

### DIFF
--- a/src/platform-includes/getting-started-primer/kotlin-multiplatform.mdx
+++ b/src/platform-includes/getting-started-primer/kotlin-multiplatform.mdx
@@ -1,10 +1,5 @@
 Sentry's Kotlin Multiplatform SDK builds on top of multiple Sentry SDKs, allowing developers to use the same codebase and share code between platforms while being able to integrate Sentry's features into their applications. Installing this SDK will enable native crash reporting and automatic breadcrumbs for app lifecycle and UI events.
 
-<Note>
-  ## Prerequisites Ensure compatibility in Kotlin Multiplatform projects
-  targeting Cocoa by using the version of the Cocoa SDK specified in our{" "}
-  <Link to="/platforms/kotlin-multiplatform/install/#cocoa-sdk-version-compatibility">
-    Cocoa SDK Version Compatibility Table
-  </Link>
-  .
-</Note>
+## Prerequisites
+
+Ensure compatibility in Kotlin Multiplatform projects targeting Cocoa by using the version of the Cocoa SDK specified in our [Cocoa SDK Version Compatibility Table](/platforms/kotlin-multiplatform/install/#cocoa-sdk-version-compatibility)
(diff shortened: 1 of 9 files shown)


src/platform-includes/getting-started-primer/kotlin-multiplatform.mdx
```diff
@@ -1,13 +1,10 @@
 Sentry's Kotlin Multiplatform SDK builds on top of multiple Sentry SDKs, allowing developers to use the same codebase and share code between platforms while being able to integrate Sentry's features into their applications.
 
-## Overview of Features
-
-- Native crash reporting for Android and JVM, leveraging our [Android SDK](/platforms/android) and [Java SDK](/platforms/java)
-- Native crash reporting for iOS, macOS, tvOS, and watchOS, leveraging our [Cocoa SDK](/platforms/apple)
-- Automatic breadcrumbs for app lifecycle and UI events
-
 <Note>
```
kahest, line 3: `m`: this note is placed in slightly different positions on all 3 pages where it's used - I'd suggest to always put it below the install code snippet

src/platforms/kotlin-multiplatform/install/index.mdx
```diff
@@ -1,7 +1,18 @@
...
-sidebar_order: 1
+sidebar_order: 3
 ---
 
 <PageGrid />
+
+## Cocoa SDK Version Compatibility
+
+Every version of our Kotlin Multiplatform SDK is compiled with a specific version of the [Sentry Cocoa SDK](/platforms/apple/).
+The table below shows the recommended Sentry Cocoa SDK versions for corresponding versions of the Kotlin Multiplatform SDK.
```
kahest, line 12:
```suggestion
Use the table below to find the recommended Sentry Cocoa SDK versions for specific versions of the Kotlin Multiplatform SDK.
```

src/platform-includes/getting-started-primer/kotlin-multiplatform.mdx
```diff
@@ -1,13 +1,10 @@
 Sentry's Kotlin Multiplatform SDK builds on top of multiple Sentry SDKs, allowing developers to use the same codebase and share code between platforms while being able to integrate Sentry's features into their applications.
 
-## Overview of Features
```
kahest, line 3: `l`: Generally speaking, I think these feature sections are useful, though I agree they shouldn't be the first thing users see when opening the platform docs. So food for thought: should we create a separate page for this? Maybe be a bit more descriptive? (totally fine as a separate PR)

kahest (commented):
Nice work 👍 left some small structural and wording suggestions

kahest:
Note to reviewers: the vercel previews still show version `8.17.1` because the bump to `8.17.2` was force-pushed and therefore no new preview generated

lizokm (approved):
I like these changes, but think we can take this one step further. I've made an outline of what I propose [here](https://docs.google.com/document/d/14Je4UeU-9QtkwWEh_zus5aoBY-w4MbfNYqa240jNVO4/edit). And set up time for us to chat through them tomorrow if you have time.

src/platform-includes/getting-started-primer/kotlin-multiplatform.mdx
buenaflor: 👍 will bring it up in the docs meeting

buenaflor:
merged it now as it's very relevant and the current information is outdated, but will make another PR with the new structure that we discussed

PR #8777, opened 2023-12-29
node: Updated ANR docs

timfish:
This PR updates the Node ANR docs to reflect the new API.

cc @lforst because I can't select you as a reviewer...


lforst (approved):
So clean

image

PR #8816, opened 2024-01-04
feat(cron): Add node cron library instrumentation docs

AbhiPrasad:
Documents https://github.com/getsentry/sentry-javascript/pull/9999 and https://github.com/getsentry/sentry-javascript/pull/9904

### DIFF
--- a/src/includes/javascript-crons-automatic-crons-instrumentation.mdx
+++ b/src/includes/javascript-crons-automatic-crons-instrumentation.mdx
@@ -1,10 +1,10 @@
 If you're using the [`cron`](https://www.npmjs.com/package/cron) or [`node-cron`](https://www.npmjs.com/package/node-cron) libraries to run your periodic tasks, you can use our instrumentation functions in the `Sentry.cron` export to monitor your cron jobs.
 
+Requires SDK version `7.92.0` or higher.
+
 ### Cron
 
 Use `Sentry.cron.instrumentCron` to instrument the `CronJob` constructor or `CronJob.from` method. Pass the name of the cron monitor as a second argument to the function.
-
-Requires SDK version `7.92.0` or higher.
 
 ```JavaScript
 import { CronJob } from "cron";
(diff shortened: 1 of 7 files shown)


anonrig (approved):

src/includes/javascript-crons-automatic-crons-instrumentation.mdx
```diff
@@ -0,0 +1,46 @@
...
+    console.log("You will see this message every minute");
+  },
+});
+```
+
+### Node Cron
+
+Use `Sentry.cron.instrumentNodeCron` to instrument cron export from the `node-cron` library. This returns an object with the same API as the original `cron` export, but with the `schedule` method instrumented. You can pass the name of the cron monitor and an optional timezone as part of the third options argument to the function.
+
+Requires SDK version `7.92.0` or higher.
```
shanamatthews, line 32: If both options have this as a prereq, should we move this line to directly under "Automatic Crons Instrumentation"?

shanamatthews (approved):
A few small comments, but overall looks good!

src/includes/javascript-crons-automatic-crons-instrumentation.mdx
```diff
@@ -0,0 +1,46 @@
...
+  cronTime: "* * * * *",
+  onTick: () => {
+    console.log("You will see this message every minute");
+  },
+});
+```
+
+### Node Cron
+
+Use `Sentry.cron.instrumentNodeCron` to instrument cron export from the `node-cron` library. This returns an object with the same API as the original `cron` export, but with the `schedule` method instrumented. You can pass the name of the cron monitor and an optional timezone as part of the third options argument to the function.
```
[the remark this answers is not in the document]
AbhiPrasad: I think that flows better

src/includes/javascript-crons-automatic-crons-instrumentation.mdx
AbhiPrasad: Yes I'll make the change, good idea.

Issue #8815, opened 2024-01-04
Documentation on showReportDialog showing for Android SDK when it is not available.


rodolfoBee:
### Core or SDK?

Platform/SDK

### Which part? Which one?

https://docs.sentry.io/platforms/android/user-feedback/configuration/

### Description

The page above shows options on how to configure Sentry.showReportDialog in the Android SDK, leading people to conclude this option is available when it is not.

### Suggested Solution

Hide the page for Android SDK.

kahest:
Likely culprit: https://github.com/getsentry/sentry-docs/pull/8714 

I believe this issue is relevant for all platforms listed under `notSupported` [here](https://github.com/getsentry/sentry-docs/pull/8714/files#diff-96770b55eed220e91736b5ea5fe0ef5f2179746ae7fdff316320a937eb7cb4ebL109): `"android", "apple", "java", "native", "dart", "flutter", "ruby", "unity", "react-native", "unreal", "kotlin-multiplatform"`

This change likely also produced https://docs.sentry.io/platforms/native/user-feedback

cc @stephanie-anderson

PR #8821, opened 2024-01-05
fix(user-feedback): Removes config pages for new UF modal for unsupported platforms

kahest:
Fixes https://github.com/getsentry/sentry-docs/issues/8815 by removing the `<platform>/user-feedback/configuration` for all platforms that explicitely do not support the new User Feedback modal based on the original [platform config](https://github.com/getsentry/sentry-docs/pull/8714/files#diff-96770b55eed220e91736b5ea5fe0ef5f2179746ae7fdff316320a937eb7cb4ebL109). More details in https://github.com/getsentry/sentry-docs/issues/8815#issuecomment-1877296914

Wrt https://docs.sentry.io/product/user-feedback/setup, there might be more platforms (e.g. Go, Elixir, Rust) that don't support this atm and should have the section removed - feel free to add to this PR or we can follow up with another.


shanamatthews (approved):
Thanks for fixing this! I'm not sure why these were added in the first place if the feature isn't supported for these SDKs...

Swatinem (approved):

PR #8858, opened 2024-01-10
Update Content Security Policy

Jeffreyhung:
Remove the items that was not in used or approved
Added the security team as codeowners to prevent un-approved changes to CSP
Removed youtube 

(removing kamilogorek from codeowner since he is no longer with us)


shanamatthews (approved):
Looks good from my end, thanks for updating this.

PR #8864, opened 2024-01-11
ref(apple): attachments intro and note about crash attachments

kahest:
Small updates to wording of intro and placement/wording of note about crash attachments

related: https://github.com/getsentry/sentry-cocoa/issues/1247

### DIFF
--- a/src/platforms/apple/common/enriching-events/attachments/index.mdx
+++ b/src/platforms/apple/common/enriching-events/attachments/index.mdx
@@ -3,7 +3,7 @@ title: Attachments
 description: "Learn more about how Sentry can store additional files in the same request as event attachments."
 ---
 
-On this page you will learn how you to enrich your events for further investigation by attaching additional files, such as config or log files.
+On this page, you will learn how to attach files, such as logs or config files, to your events. You'll be able to see these attachments in [sentry.io](https://sentry.io/) alongside other event details.
 
 ## Creating Attachments
 
--- a/src/platforms/apple/common/enriching-events/attachments/index.mdx
+++ b/src/platforms/apple/common/enriching-events/attachments/index.mdx
@@ -3,9 +3,7 @@ title: Attachments
 description: "Learn more about how Sentry can store additional files in the same request as event attachments."
 ---
 
-Please note that attachments don't work yet with crashes.
-
-Sentry can enrich your events for further investigation by storing additional files, such as config or log files, as attachments.
+On this page you will learn how you to enrich your events for further investigation by attaching additional files, such as config or log files.
 
 ## Creating Attachments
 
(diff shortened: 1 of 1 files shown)


src/platforms/apple/common/enriching-events/attachments/index.mdx
```diff
@@ -33,6 +31,12 @@ In addition, you can set these parameters:
 
 Attachments live on the <PlatformLink to="/enriching-events/scopes/">Scope</PlatformLink>. You can either add an attachment on the global scope to be sent with every event or add it on the <PlatformLink to="/enriching-events/scopes/#local-scopes">local Scope</PlatformLink> to just send the attachment with one specific event.
 
+<Note>
+
+Adding attachments to crashes is currently not supported.
```
kahest, line 36:
```suggestion
To add attachments to crashes, use <PlatformLink to="/configuration/options/#initial-scope">`initialScope`</PlatformLink>, and remove them after initialization of the SDK.
```

kahest: do we want to suggest this here? @philipphofmann

src/platforms/apple/common/enriching-events/attachments/index.mdx
```diff
@@ -3,9 +3,7 @@ title: Attachments
 description: "Learn more about how Sentry can store additional files in the same request as event attachments."
 ---
 
-Please note that attachments don't work yet with crashes.
-
-Sentry can enrich your events for further investigation by storing additional files, such as config or log files, as attachments.
+On this page you will learn how you to enrich your events for further investigation by attaching additional files, such as config or log files.
```
shanamatthews, line 6:
```suggestion
On this page, you will learn how to attach files, such as logs or config files, to your events. You'll be able to see these attachments in [sentry.io](https://sentry.io/) alongside other event details.
```

src/platforms/apple/common/enriching-events/attachments/index.mdx
shanamatthews: I'm not familiar with this way of adding attachments to crashes, but this sentence probably doesn't give enough context to make it possible for most users to be successful.

If we want to include this, we should definitely elaborate. I'd add a separate subheader for it.

philipphofmann: The problem is that you would add this whenever the app starts / users initialize the SDK. `SentrySDK.crashedLastRun` doesn't work for `initialScope` because the SDK must be initialized. I think this could confuse users more than it helps. Adding hints to the Cocoa SDK is the way to go IMO, and then we can update the docs.

kahest: Yeah that's what I expected - thank you both

shanamatthews (approved):

vivianyentran (approved):

PR #8889, opened 2024-01-17
feat: Remove confusing note on `withScope` exception handling

mydea:
This is a bit confusing, as what it wants to say (probably) is "errors here are not caught and swallowed, but are captured normally by Sentry error handlers & reported to Sentry", but can be read as "errors are neither caught nor reported to sentry". Overall this is not really helpful, so just removing this.


szokeasaurusrex (approved):

Lms24 (approved):

anonrig (approved):

cleptric (changes requested):
Vierfach hält besser :shipit:

cleptric (approved):

Issue #8880, opened 2024-01-16
The Sentry.setUser function is poorly documented


boxed:
### Core or SDK?

Core Sentry product

### Which part? Which one?

browser SDK

### Description

User Sentry.setUser as the docs said. But if you do that, you get an error because setUser doesn't exist yet. I think it should.

See also https://stackoverflow.com/questions/71699244/where-to-set-sentrys-setuser-in-next-js-app for others with this problem

### Suggested Solution

Make setUser available at the very start. Ideally even before Sentry.init is called.

mydea:
Hey,

I'm not sure I fully understand - `Sentry.setUser()` should always be available in recent versions of the SDK. Are you saying that this:

```js
import * as Sentry from '@sentry/browser';

Sentry.setUser(...);
```

is failing for you? If so, which package & version are you using? I'd ask you to open an issue here https://github.com/getsentry/sentry-javascript with more details, so we can debug this!

Or is it working, but not actually being set properly?

One potential reason is that you may be setting the user inside of a forked execution context - with `Sentry.setUser()`, this will set the user on the current scope, so it is only valid in there. If this is the case, you can use the new (and not yet documented very well) `getIsolationScope().setUser()` to set a user for the whole request. Or you can use `getGlobalScope().setUser()` to set a user globally (if you're using this in the client/browser environment, setting on the global scope is probably what you want).

boxed:
I had this code:

```js
Sentry.init(....)
Sentry.setUser()
```

which fails with "

If I change it to:

```js
Sentry.init(...);
Sentry.configureScope(scope => scope.setUser({email: '{{ user.email }}'}));
```

then that works.

boxed:
I copy pasted the url from the setup docs. I would assume that's the latest version you link to with the CDN.

mydea:
Which URL, from which docs? In order to be able to help you, I need a bit more specific information. Are you using the Loader Script? If so, the docs explain here:

https://docs.sentry.io/platforms/javascript/install/loader/#guarding-sdk-function-calls

that since the loader is lazy-loaded, you can only use a limited set of global methods on it. You need to guard any other calls - in your case, what you should do is:

```js
 window.sentryOnLoad = function () {
   // in here, Sentry is ensured to be loaded & fully available
    Sentry.init(...);
    Sentry.setUser(...);
  };
```

as outlined here: https://docs.sentry.io/platforms/javascript/install/loader/#custom-configuration

boxed:
[...]

PR #8890, opened 2024-01-17
feat: Update browser getting started docs to explain loader vs npm config

mydea:
Fixes https://github.com/getsentry/sentry-docs/issues/8880

It was brought up in that issue, that while we point to the Loader setup first, we show the non-loader config below on the "Getting Started" page for Browser JS. This is a bit confusing...  I updated this to specifically call out that config differs for the loader, adding tabs for npm/loader/cdn for clarity.

While at it, I also renamed the tabs from "ESM" to `npm` in the integration pages, which is IMHO clearer for there.

### DIFF
--- a/src/platform-includes/configuration/capture-console/javascript.mdx
+++ b/src/platform-includes/configuration/capture-console/javascript.mdx
@@ -1,6 +1,6 @@
 <SignInNote />
 
-```javascript {tabTitle: ESM}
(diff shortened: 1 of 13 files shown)


AbhiPrasad (approved):

Lms24 (approved):
I like it, this is much clearer!

PR #8898, opened 2024-01-18
ref: Add packagemanger and volta

HazAT:
@mjq-sentry I added key to `package.json` to manage package manager version and volta environment
Also added few things to `.gitignore`


HazAT:
I went ahead and merged this so I can continue more easily with the changelog

PR #8911, opened 2024-01-19
ref: Make "js" codeblocks --> "javascript"

billyvg:
This was a bit inconsistent, but the majority (800+ instances) of codeblocks were "javascript" vs 40 or so of "js"

### DIFF
--- a/src/docs/product/performance/resources.md
+++ b/src/docs/product/performance/resources.md
@@ -47,7 +47,7 @@ If you've manually instrumented Sentry, you'll need to make sure that your spans
 
 Here's an example snippet of creating a resource span manually with the SDK.
 
-```js
+```javascript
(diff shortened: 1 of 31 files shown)


bruno-garcia (approved):

bruno-garcia:
Does this affect the re-write to nextjs?

billyvg:
> Does this affect the re-write to nextjs?

Good question, not sure how that is being handled.... cc @mjq-sentry

shanamatthews:
@billyvg thank you for doing this! So annoying.

AFAIK this shouldn't be a problem with the rewrite, it's just a markdown change and those will all be pulled in with @mjq-sentry 's script.

shanamatthews (approved):

PR #8951, opened 2024-01-24
feat(code-mappings): Update docs to reference absolute path when setting up code mappings

malwilley:
After merging https://github.com/getsentry/sentry/pull/62768, users are able to reference the absolute path when creating code mappings, which mostly applies to native platforms. This adds a text note and screenshot about where to find that information in the native stack trace.


vivianyentran (approved):
looks good! Thanks for the updates!

PR #8954, opened 2024-01-25
fix(stacktrace): Update some stack trace screenshots

malwilley:
Our stack traces look a lot better now (and have syntax highlighting!), so I updated a few of the ones that I think are the highest visibility. I also updated the platform-specific ones to represent the platform the docs are under instead of just a generic c# stack trace like it was before.


vivianyentran (approved):
nice, thanks for updating the screenshots!

PR #8960, opened 2024-01-25
Use getIntegrationByName API for Replay docs

AbhiPrasad:
ref https://github.com/getsentry/sentry-javascript/issues/7384#issuecomment-1910863680


mydea (approved):

Amy-B-Tradition:
I actually tried getIntegrationByName, and it also did not work. It's possible it was just a TS fail, but I'm so used to deferring to TS I didn't check to see if it worked.

PR #8961, opened 2024-01-25
feat(webvitals): More browser support performance score details, and Firefox LCP support

edwardgou-sentry:
Add some more content around required browser support for performance score calculations. Also update firefox to include lcp support in the browser support table. (Firefox supports lcp as of version 122 released on Jan 23, 2024)


narsaynorath (approved):
:+1:

vivianyentran (approved):
LGTM! Thanks for updating this!

PR #8964, opened 2024-01-25
feat(canvas): Remove EA requirement for replay-canvas

billyvg:
This should actually be open beta rather than limited to EA.


vivianyentran (approved):
LGTM! Thanks for clarifying this!

PR #8965, opened 2024-01-25
feat(crons): Update docs around monitor environments

davidenwang:

PR #8967, opened 2024-01-25
fix(crons): Followup suggestions to multi-environment docs

davidenwang:
Follow up to https://github.com/getsentry/sentry-docs/pull/8965


vivianyentran (approved):
thanks for adding this!

PR #8980, opened 2024-01-29
Migrate docs site to Next.js

mjq-sentry:
## Pre-merge checklist

- [X] Checked Vercel preview for correctness, including links

## This PR is huge, how am I supposed to review this?

1. Sorry 😢 
2. Most of the size is in the content moves - every .mdx file has a new home. Git recognizes almost all of these as a file move, but nonetheless they add to the diff.

**You will have a somewhat nicer time if you check out the branch and review just the `src/` and `app/` directory changes** - that's where all of the _code_ changes live. Everything else is content moves (with minor tweaks to account for MDX version changes), and going to be very uninteresting to review.

## Description of changes

This PR migrates the docs site to Next.js from Gatsby.

This will give Sentry an opportunity to dogfood its Next.js support and (later) improve the integration between docs and the main Sentry codebase via shared React components.

The contributing experience is still very similar: pages are assembled, typically from a main `.mdx` file with additional content from other `.mdx` files included, as well as the same system for platform and guide inheritance (`PlatformSection`, `common` folders, etc). The biggest content-related change is that images must now live in the `public/` folder - they have been moved with an equivalent folder structure to maintain their current URLs, but the folder structure is arbitrary and can be simplified later if it's easier.

The new version uses SSG for everything but the changelog (which @HazAT has been working on, and is beta-ready but not yet replacing the production changelog). During the build process, every page is assembled and cached, and then that cache is used exclusively in production (no dynamic rendering).

Server components are used wherever possible. This lead to some existing components being split in two - e.g. `JsBundleList` became `JsBundleList` and `JsBundleListClient`. Data fetched from network or disk happens in the server component (and therefore happens at build time), with client-only functionality moved to the client component (most commonly React hooks or emotion).

Some changes were made to optimize for diff size (which is already huge, I realize), e.g. putting data into the same shape that Gatsby used to use so that component rendering could stay consistent. All of this can be smoothed out in the near future, but for now not introducing new bugs and a (slightly) easier code review were priorities.

The changes have been fairly extensively tested at https://docs-next.sentry.io/ from an end user perspective.


HazAT (commented):
This PR is soo huge that I think it can only reasonably be approached from an e2e testing view - or maybe point out obvious things.

I can just say since working a bit on it, generally, the page works, there might still be some small things that are not working 100%, like styling, but IMHO, we should fix those things as we move along, otherwise we'll never finish this 😅 

For me, one important thing we still need to figure out is broken links/pages, not just the ones that might come with the move -> but also when changing content in MDX files. We don't have a link checker yet as we had for Gatsy so we need to be careful until we have it when doing large refactoring so we don't break too many links.

README.md
```diff
@@ -15,3 +15,5 @@ _Bad software is everywhere, and we're tired of it. Sentry is on a mission to he
 Take a look at our [Contributing to Docs](https://docs.sentry.io/contributing/) documentation to get started.
 
 Note: The documentation for this repository is self-hosted via `src/docs/contributing/`.
+
+This is the Next.js version of our docs.
```
lforst, line 19:
```suggestion
Note: The documentation for this repository is self-hosted via `src/docs/contributing/`.
```

We can remove this now :D

.babelrc.js.bak
```diff
@@ -0,0 +1,16 @@
...
+// to output folders
+if (process.env.NODE_ENV !== `test`) {
+  ignore.push(`**/__tests__`);
+}
+
+module.exports = {
+  sourceMaps: true,
+  presets: [],
+  ignore,
+};
```
lforst, line 16: Any reason to keep the backup of `.babelrc.js` around?

.gitignore
```diff
@@ -57,7 +57,6 @@ typings/
 
 # gatsby files
 .cache/
```
lforst, line 59:
```suggestion
```

app/[[...path]]/page.tsx
```diff
@@ -0,0 +1,122 @@
...
+export const dynamicParams = false;
+export const dynamic = 'force-static';
+
+const mdxComponentsWithWrapper = mdxComponents(
+  {Include, PlatformContent},
+  ({children, frontMatter}) => <DocPage frontMatter={frontMatter}>{children}</DocPage>
+);
+
+function MDXLayoutRenderer({mdxSource, ...rest}) {
+  const MDXLayout = useMemo(() => getMDXComponent(mdxSource), [mdxSource]);
```
Comment (lforst) [inline app/[[...path]]/page.tsx:41]: I think it would be cool to properly type this function so that we can also be sure that useMemo doesn't use any dependency that has changing identity.

app/global-error.jsx
```diff
@@ -0,0 +1,23 @@
+'use client';
```
lforst, line 1: In theory we can turn this file into a tsx. We just need to slap a `ts-ignore` above the `<Error />` component.

app/[[...path]]/layout.tsx
```diff
@@ -0,0 +1,13 @@
...
+import type {Metadata} from 'next';
+
+import 'sentry-docs/styles/screen.scss';
+
+export const metadata: Metadata = {
+  title: {template: '%s | Sentry Documentation', default: 'Home'},
+};
+
+export default function DocsLayout({children}: {children: React.ReactNode}) {
+  return <div>{children}</div>;
```
Comment (lforst) [inline app/[[...path]]/layout.tsx:12]: Do we even need the div here? Or is it needed for styling?

package-lock.json
```diff
@@ -0,0 +1,10041 @@
+{
```
lforst, line 1: Correct me if I am wrong but shouldn't we have a yarn.lock instead of a package-lock.json? Or is this some yarn 4 thing? If so can we remove one of them?

next.config.js
```diff
@@ -0,0 +1,63 @@
...
+
+module.exports = withSentryConfig(
+  module.exports,
+  {
+    // For all available options, see:
+    // https://github.com/getsentry/sentry-webpack-plugin#options
+
+    // Suppresses source map uploading logs during build
+    silent: true,
+    org: 'sentry',
```
lforst, line 35:
```suggestion
    silent: !process.env.CI,
```

Annoying during build and locally but useful on Vercel and github.

lforst (approved):
I took a look at all the non-mdx changes. Super clean! I still cannot believe that I will have the docs for dogfooding soon!!

I have some minor suggestions here and there but nothing is blocking.

PR #8986, opened 2024-01-30
Changelog Coontribution Guide

smeubank:
Creating a first draft of a guide for contributing to the changelog

[CONTRIBUTING.md](https://github.com/getsentry/sentry-docs/blob/smeubank-changelog-contribution/changelog/CONTRIBUTING.md)


smeubank:
If we want a "lo-fi" template solution we could follow the RFC model with a python scrript, but it's overkill IMO.

GitHub has a template concept for creating issues, these tempaltes do not appear to be accessible for creating a new file generically unfortunately. I think copy pasta is good enough for now


https://github.com/getsentry/rfcs/blob/main/new-rfc.py

PR #8988, opened 2024-01-30
Fix draft issue "Bullet points are a bit off"

stephanie-anderson:
Fix for draft issue: "Bullet points are a bit off"

Not fixing root cause yet, which is: 
```
- <Link to="/platforms/python/integrations/gcp-functions">
      Google Cloud Functions
   </Link>
```

generates

```
<ul>
    <li>
        <a>
            <p>Google Cloud Functions</p>
        </a>
    </li>
</ul>
```

The paragraph shouldn't get generated like that


docs/platforms/python/configuration/sampling.mdx
```diff
@@ -126,6 +124,6 @@ There are multiple ways for a transaction to end up with a sampling decision.
...
-1. If <PlatformIdentifier name="traces-sampler" /> is not defined, but there's a parent sampling decision, the parent sampling decision will be used.
-1. If <PlatformIdentifier name="traces-sampler" /> is not defined and there's no parent sampling decision, <PlatformIdentifier name="traces-sample-rate" /> will be used.
+2. If <PlatformIdentifier name="traces-sampler" /> is defined, its decision will be used. It can choose to keep or ignore any parent sampling decision, use the sampling context data to make its own decision, or choose a sample rate for the transaction. We advise against overriding the parent sampling decision because it will break distributed traces)
+3. If <PlatformIdentifier name="traces-sampler" /> is not defined, but there's a parent sampling decision, the parent sampling decision will be used.
+4. If <PlatformIdentifier name="traces-sampler" /> is not defined and there's no parent sampling decision, <PlatformIdentifier name="traces-sample-rate" /> will be used.
```
shanamatthews, line 129: Does the 1's only style of numbering not work in the next site?

It's a common practice in markdown and we should support it, if we don't...

mjq-sentry: The all-1s numbering seems to work fine for me on the Next.js preview site without merging this PR (i.e. generated from the all-1s version): https://docs-next.sentry.io/platforms/python/configuration/sampling/

Maybe just a stylistic decision? cc @stephanie-anderson

mjq-sentry: I'm going to merge this as-is, but in general, I confirmed that 1's only works fine in the Next version :+1:

PR #8989, opened 2024-01-30
Fix draft issue "Colons are off"

stephanie-anderson:
This removes the indentation that was previously generated by using the following syntax

```
: some text that should be indented

```

This is now 

```
some text that should be indented

```


shanamatthews (commented):
I don't have strong feelings about eliminating these indentations, but should we check with the SDK devs if they mind?

mjq-sentry:
> I don't have strong feelings about eliminating these indentations, but should we check with the SDK devs if they mind? It seems like it's a pretty widely used feature.

Yeah this would only be temporary - I'll definitely bring this functionality back.

Issue #8982, opened 2024-01-30
Update JS docs to avoid deprecated methods


mydea:
As we are currently ramping up our efforts in the JS SDKs to prepare for v8, we are shipping a lot of deprecations there. We need to update the docs to avoid deprecated methods.

PR #9016, opened 2024-02-01
feat: Remove `startTransaction` usage in JS docs

mydea:
Except for usage in koa/express, which will be replaced by auto-instrumentation in v8.

A few notable things:

* Forcing a sampling decision like passing `sampled: true` to startTransaction will not be directly possible, so I removed this block. There _are_ ways to kind of make this work, but it's more complicated so if this ever comes up we can add it back to the docs - it's unclear how important this is. You can always use a `tracesSampler` to force sampling for some transaction.
* There is for now no built-in way to pass additional sampling data to spans (because OTEL does not support that). I think that's OK for now, if ever needed we can document some approach around attaching data to the scope or something like this.

Part of https://github.com/getsentry/sentry-docs/issues/8982


platform-includes/performance/add-spans-example/javascript.mdx
```diff
@@ -3,56 +3,47 @@ For example, if you want to create a transaction for a user interaction on your
...
-  // Assume this function makes a fetch call
-  const result = validateShoppingCartOnServer();
-
-  const span = transaction.startChild({
-    data: {
-      result,
-    },
-    op: "task",
-    description: `processing shopping cart result`,
+  return Sentry.startSpan({ name: "shopCheckout" }, () => {
```
mydea, line 6: Here you can actually see how much nicer the new APIs are in such scenarios ❤️

AbhiPrasad (approved):

vivianyentran (approved):
content LGTM, will leave it up to the sdk engineers to review the code :)

platform-includes/distributed-tracing/custom-instrumentation/javascript.mdx
```diff
@@ -58,10 +58,12 @@ const baggage = document.querySelector("meta[name=baggage]")?.content;
 const pageLoadTransaction = Sentry.continueTrace(
   { sentryTrace, baggage },
   (transactionContext) => {
-    return Sentry.startTransaction({
+    Sentry.startSpan({
```
Lms24, line 61: m (but technically out of scope): Did `Sentry.continueTrace` ever return a transaction? I thought it only returned a `TransactionContext`?

platform-includes/getting-started-verify/node.mdx
```diff
@@ -1,16 +1,16 @@
...
   op: "test",
   name: "My First Test Transaction",
+}, () => {
+  setTimeout(() => {
+    try {
+      foo();
+    } catch (e) {
+      Sentry.captureException(e);
+    } finally {
+      transaction.finish();
```
Lms24, line 12: m: This shouldn't be here anymore, no? Generally, would the `setTimeout` semantic still work with `startSpan`? I don't think so

platform-includes/performance/add-spans-example/javascript.mdx
Lms24: Yeah much cleaner, I agree.

One comment but logaf-l: Should we make this function `async` and `await` the fetch call? Might help users understand that `startSpan` supports async/Promise as well.

platform-includes/performance/custom-sampling-context/node.mdx
```diff
@@ -1,22 +1,28 @@
...
     name: 'GET /search',
     op: 'search',
-    data: {
-      queryParams: {
-        animal: "dog",
-        type: "very good"
-      }
+    attributes: {
+      queryForAnimal: "dog",
+      queryForType: "very good"
```
Lms24, line 8: l: No strong opinion but should we promote a more formalized way of setting attribute keys? Something like

```suggestion
      "queryParam.animal": "dog",
      "queryParam.type": "very good"
```

Which more directly represents the flat attributes Record style?

platform-includes/distributed-tracing/custom-instrumentation/javascript.mdx
Lms24: Either way, I think we should just remove the const here, no?

mydea: ah, well it returns what is returned by the callback, but you're right, will remove this!

platform-includes/performance/custom-sampling-context/node.mdx
mydea: makes sense to me!

platform-includes/performance/add-spans-example/javascript.mdx
mydea: good point, I think I'll just add both of these!

docs/platforms/javascript/guides/angular/features/component-tracking.mdx
```diff
@@ -184,8 +184,7 @@ User `TraceDirective` if you only want to track components in certain instances
 
 ### Track Angular Bootstrapping
 
-You can add your own custom spans by attaching them to the currently active transaction using the `getActiveTransaction`
-helper. For example, you can track the duration of the Angular bootstrapping process to find out how long your app takes to bootstrap on your users' devices:
+You can add your own custom spans using `startSpan()`. For example, you can track the duration of the Angular bootstrapping process to find out how long your app takes to bootstrap on your users' devices:
```
mydea, line 187: @Lms24 also noticed these here while going over this one more time and adjusted this, please sanity check if that all sounds reasonable for angular :D

Lms24: Just verified the snippet and it works 👍 However, we need to replace `description` with `name`

docs/platforms/javascript/guides/angular/features/component-tracking.mdx
```diff
@@ -196,23 +195,18 @@ import { AppModule } from "./app/app.module";
 
 // ...
 
-const activeTransaction = Sentry.getActiveTransaction();
-const bootstrapSpan =
-  activeTransaction &&
-  activeTransaction.startChild({
+startSpan(
+  {
     description: "platform-browser-dynamic",
```
Lms24, line 200: We need to replace description with name. Also, we might wanna change to the suggestion below because `platformBrowserDynamic` isn't used in Angular 17 anymore. 

```suggestion
    name: "bootstrap-angular-application",
```

Lms24 (approved):
Thanks, LGTM now!

docs/platforms/javascript/guides/angular/features/component-tracking.mdx
```diff
@@ -224,23 +218,18 @@ import { AppModule } from "./app/app.module";
...
```
mydea, line 223:
```suggestion
    name: "platform-browser-dynamic",
```

platform-includes/performance/add-spans-example/javascript.mdx
```diff
@@ -3,58 +3,75 @@ For example, if you want to create a transaction for a user interaction on your
...
-        span.setTag("http.status_code", response.statusCode);
-        span.setData("http.foobarsessionid", getFoobarSessionid(response));
-        span.finish();
-        resolve(response);
+```javascript
+function processItem(item) {
+  return Sentry.startSpan(
+    {
+      op: "http",
+      description: `GET /items/:item-id`,
```
mydea, line 41:
```suggestion
      name: "GET /items/:item-id",
```

platform-includes/performance/add-spans-example/javascript.mdx
```diff
@@ -3,58 +3,75 @@ For example, if you want to create a transaction for a user interaction on your
...
+```
+
+Or when using an async function:
+
+```javascript
+function processItem(item) {
+  return Sentry.startSpan(
+    {
+      op: "http",
+      description: `GET /items/:item-id`,
```
mydea, line 66:
```suggestion
      name: "GET /items/:item-id",
```

PR #9017, opened 2024-02-01
fix(.NET): Metrics

bitsandfoxes:
Updated the metrics docs by adding the `Emit Timer` part.
Also reordered and fixed headings to mirror what Python has got going.


vivianyentran (approved):
looks good! Thanks for making it consistent with python :)

PR #9018, opened 2024-02-01
fix(.NET): Added `BeforeTransaction`

bitsandfoxes:


platform-includes/configuration/before-send-transaction/dotnet.mdx
```diff
@@ -0,0 +1,17 @@
+A `Func<SentryTransaction, Hint, SentryTransaction?>` can be used to update the transaction before it's sent to Sentry, or drop it by returning `null`, for example:
```
vivianyentran, line 1:
```suggestion
A `Func<SentryTransaction, Hint, SentryTransaction?>` can be used to update the transaction or drop it by returning `null` before it gets sent to Sentry. For example:
```

vivianyentran (approved):
LGTM, thanks for adding!

PR #9035, opened 2024-02-02
feat: Ensure prettier runs for docs

mydea:
Noticed that this didn't actually run for docs files themselves anymore!

### DIFF
--- a/docs/platforms/android/enhance-errors/kotlin-compiler-plugin.mdx
+++ b/docs/platforms/android/enhance-errors/kotlin-compiler-plugin.mdx
@@ -6,8 +6,12 @@ description: "Learn about using the Sentry Kotlin Compiler Plugin."
 
 The [Sentry Kotlin Compiler Plugin](https://github.com/getsentry/sentry-android-gradle-plugin) is an addition to the Sentry Android Gradle plugin and offers improvements to existing SDK features. As of now the main focus of the plugin is to improve apps which utilize Jetpack Compose by automatically tagging `@Composable` functions.
 
-- <PlatformLink to="/enriching-events/viewhierarchy">View Hierarchy</PlatformLink>
-- <PlatformLink to="/performance/instrumentation/automatic-instrumentation/#user-interaction-instrumentation">User Interaction Tracing</PlatformLink>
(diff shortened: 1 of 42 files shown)


docs/platforms/android/enhance-errors/kotlin-compiler-plugin.mdx
```diff
@@ -6,8 +6,12 @@ description: "Learn about using the Sentry Kotlin Compiler Plugin."
...
 The [Sentry Kotlin Compiler Plugin](https://github.com/getsentry/sentry-android-gradle-plugin) is an addition to the Sentry Android Gradle plugin and offers improvements to existing SDK features. As of now the main focus of the plugin is to improve apps which utilize Jetpack Compose by automatically tagging `@Composable` functions.
 
-- <PlatformLink to="/enriching-events/viewhierarchy">View Hierarchy</PlatformLink>
-- <PlatformLink to="/performance/instrumentation/automatic-instrumentation/#user-interaction-instrumentation">User Interaction Tracing</PlatformLink>
+- <PlatformLink to="/enriching-events/viewhierarchy">
+    View Hierarchy
+  </PlatformLink>
+- <PlatformLink to="/performance/instrumentation/automatic-instrumentation/#user-interaction-instrumentation">
+    User Interaction Tracing
+  </PlatformLink>
```
mjq-sentry, line 14: Spacing changes like this actually change the markup output by the docs when the MDX is processed, and are therefore unsafe to make 😞 (This change will wrap the inner text in an extra `<p>` tag, for example).

We might be able to configure prettier to only run a subset of its lints (i.e. only markup-preserving changes - changes inside code blocks should be fine for example), but basically anything that changes tag spacing is probably a no-go.

mjq-sentry (changes requested):

docs/platforms/android/enhance-errors/kotlin-compiler-plugin.mdx
mydea: I'll close this then, no worries, just noticed that code snippets are not prettiered anymore, which was unexpected to me - if we can find a way to only run this on code or so, would be appreciated!

Issue #9049, opened 2024-02-05
Alert type `info` looks like an error


mydea:
### Steps to Reproduce

This code:

```mdx
<Alert level="info" title="Using a framework?">

Get started using a guide listed in the right sidebar.

</Alert>
```

Looks like this: 

![image](https://github.com/getsentry/sentry-docs/assets/2411343/89f95b83-ae3c-4315-864d-3d009fa36ab7)

Which I don't think is intended?

### Expected Result

Should look like an info message.

### Actual Result

Looks like an error message.

mydea:
The same is actually true for `level='warning'`, I think something is off in the CSS 😅

PR #9050, opened 2024-02-05
feat: Add deprecation info to JS docs

mydea:
This adds an explicit callout to the getting started page of all JS SDKs pointing them to the migration docs. They are maybe not super intuitive to find right now, if you go search for them on the docs... We can eventually remove them again.

Note that they shouldn't actually look like errors, this is due to https://github.com/getsentry/sentry-docs/issues/9049.

### DIFF
--- a/docs/platforms/javascript/common/index.mdx
+++ b/docs/platforms/javascript/common/index.mdx
@@ -36,6 +36,15 @@ If you prefer to follow video instructions, see [How to Install the Sentry Next.
 
 </PlatformSection>
 
+<Alert level="warning" title="Seeing deprecation warnings in your code?">
+
+We are currently preparing version 8 of the JavaScript SDKs.
+In v8, some methods and properties will be removed or renamed.
+You can refer to the [Migration docs](https://github.com/getsentry/sentry-javascript/blob/develop/MIGRATION.md#deprecations-in-7x)
+to see how you can update your code to be compatible with v8.
+
+</Alert>
+
 <PlatformSection notSupported={["javascript.deno"]}>
 
 ## Install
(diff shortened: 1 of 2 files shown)


docs/platforms/javascript/common/index.mdx
```diff
@@ -36,6 +36,15 @@ If you prefer to follow video instructions, see [How to Install the Sentry Next.
...
 </PlatformSection>
 
+<Alert level="warning" title="Seeing deprecation warnings in your code?">
+
+We are currently preparing version 8 of the JavaScript SDKs.
+In v8, some methods and properties will be removed or renamed.
+You can refer to the [Migration docs](https://github.com/getsentry/sentry-javascript/blob/develop/MIGRATION.md#deprecations-in-7x)
+to see how you can update your code to be compatible with v8.
+
+</Alert>
```
lizokm, line 46: If you're seeing **deprecation warnings in your code**, please note that we're currently working on version 8 of the JavaScript SDKs. In v8, some methods and properties will be removed or renamed. Check out the [Migration docs](https://github.com/getsentry/sentry-javascript/blob/develop/MIGRATION.md#deprecations-in-7x)
and learn how to update your code to be compatible with v8.
```
I think that adding another alert makes this section feel quite fragmented. The user has to stop and parse a lot of information before they can get to the instructions. It would flow better if rather than making this another alert, we could add it just after the "On this page, we get you up and running with Sentry's SDK." section, so it would read: "On this page, we get you up and running with Sentry's SDK. If you're seeing **deprecations warnings in your code**, please note..."

lizokm (approved):
I left a recommendation about potentially moving this note to a different spot and a bit of language tweaks, but otherwise this looks great!

docs/platforms/javascript/common/index.mdx
mydea: I updated this based on your feedback, now it is:

image

Issue #9020, opened 2024-02-01
Speed up Next.js build


mjq-sentry:
The full site Next.js build currently takes ~38 minutes in Vercel, which is annoyingly slow. I previously got it down to ~18 minutes in Vercel but it regressed to over 35 minutes in c8d5391dad419b21fb85d5b1c2deca45ca483257. Look into want happened and see how much time we can get back.

stephanie-anderson:
Optimisations are done for now, closing issue

PR #9069, opened 2024-02-06
Improve build times

mjq-sentry:
Debugged the load times using commit c8d5391d, which more than tripled local build times and more than doubled Vercel build times compared to the previous commit (baf391b1).

Initially, c8d5391d took 9:35 locally.

After caching the `apiCategories` function, the time dropped to 6:34. This method makes a `fetch` call to retrieve our OpenAPI specs - I'm suspicious that Next's fetch cache isn't working properly on this call. We only need this to run once per build anyway.

Further, caching the `getDocsFrontMatter` function dropped the build time to 5:19. This function defensively copies a large array, but the result only needs to be calculated once per build.

There's still work to be done to get all the way back down to baf391b1's build time (2:47 locally), but it's a start.

See #9020.

### DIFF
--- a/src/build/resolveOpenAPI.ts
+++ b/src/build/resolveOpenAPI.ts
@@ -4,6 +4,8 @@
 
 import {promises as fs} from 'fs';
 
+import {cache} from 'react';
+
 import {DeRefedOpenAPI} from './open-api/types';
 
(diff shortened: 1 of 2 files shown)


mjq-sentry:
Well, the first build took 24:27 compared to our recent average of ~37, so a decent improvement.

PR #9056, opened 2024-02-05
feat: Use functional integration importants for JS SDKs

mydea:
This replaces all usages of integrations for node & browser JS to their functional counterpart.

We need to wait for 7.100.0 to be released before merging this!

PR #9073, opened 2024-02-06
Fix `supported`/`notSupported` frontmatter logic

mjq-sentry:
The previous logic for whether `common` guide pages were applicable based on `supported` and `notSupported` tags was not the same as the old site. It was a little too strict, expecting `supported` to act as an allowlist and `notSupported` as a separate blocklist. Instead, the logic combined the two.

Copied the logic from the old Gatsby site exactly (just renaming the method): https://github.com/getsentry/sentry-docs/blob/405170000995bd8d98bd155f4e8d38ce4c19275a/src/gatsby/createPages/createPlatformPages.ts#L138-L161

Fixes a reported issue where some JavaScript guides were missing Automatic Instrumentation pages.

---

⚠️ Please note that #9056 created Automatic Instrumentation pages for many of the platforms that were missing as a result of this bug. However, the new pages are missing a lot of content.

Gatsby version of Ember Automatic Instrumentation: https://sentry-docs-6x97536mr.sentry.dev/platforms/javascript/guides/ember/performance/instrumentation/automatic-instrumentation/

Current prod version of Ember Automatic Instrumentation: https://sentry-docs-hd1o5x0zq.sentry.dev/platforms/javascript/guides/ember/performance/instrumentation/automatic-instrumentation/

I'm not sure this was intended. cc @lizokm, @mydea ⚠️

### DIFF
--- a/src/mdx.ts
+++ b/src/mdx.ts
@@ -30,6 +30,29 @@ const root = process.cwd();
 function formatSlug(slug) {
   return slug.replace(/\.(mdx|md)/, '');
 }
+const isSupported = (
+  frontmatter: any,
+  platformName: string,
+  guideName?: string
+): boolean => {
+  const canonical = guideName ? `${platformName}.${guideName}` : platformName;
+  if (frontmatter.supported && frontmatter.supported.length) {
+    if (frontmatter.supported.indexOf(canonical) !== -1) {
+      return true;
+    }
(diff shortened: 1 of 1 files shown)


mydea:
Yeah, the change for ember was on purpose - it was just wrong before. We can/should/will extend on this a bit - as part of the v8 updates we want to go over _all_ JS docs and update them to make sure they are fully up to date.

PR #9081, opened 2024-02-07
chore: dotnet profiling tuneups

vaind:
Replaces #8947 after docs migration.


docs/product/profiling/getting-started.mdx
```diff
@@ -20,7 +20,7 @@ Profiling depends on Sentry's performance monitoring product being enabled befor
   - [iOS](/platforms/apple/guides/ios/profiling/)
   - [React Native](/platforms/react-native/profiling/) [beta]
   - [Flutter](/platforms/flutter/profiling/) [experimental, iOS and macOS only]
-  - [.NET](/platforms/dotnet/profiling/) [experimental, iOS only]
+  - [.NET](/platforms/dotnet/guides/maui/profiling/) [experimental, iOS only]
```
vivianyentran, line 23: Not sure why you want to steer users to the maui docs instead of the general dotnet docs? the pages look identical

docs/platforms/dotnet/common/profiling/index.mdx
```diff
@@ -31,43 +31,43 @@ dotnet add package Sentry.Profiling --prerelease
...
 ```
 
-```csharp {tabTitle:Windows/Linux/macOS}
+```csharp {tabTitle:iOS/MacCatalyst}
 SentrySdk.Init(options =>
 {
     // ... usual setup options omitted for clarity (see Getting Started) ...
 
     // Sample rate for your transactions, e.g. value 0.1 means we want to report 10% of transactions.
-    // Setting 1.0 means all transactions are profiled.
```
vivianyentran, line 56: Not sure why this line was removed. Can we add it back in so the two code samples have it?

docs/product/profiling/getting-started.mdx
vaind: because these are the docs relevant to iOS profiling (what this line is about).

vivianyentran: ah gotcha

PR #9090, opened 2024-02-08
fix(native): clarify android support

supervacuus:
The Native SDK doesn't support Android directly but acts as a support library for downstream SDKs. The formulation in the documentation needs to be clarified so users can understand the level at which we can support their use cases, which often deviate severely from anything we currently implement, build, or test.

In particular, Qt for Android as a combined supported scenario can easily misinterpreted even though it is not explicitly mentioned. It is better to remove "Android" from the "Getting Started" layer for Qt altogether.

cc: @kahest


kahest (approved):
thank you for the clarifications! 🚀

platform-includes/getting-started-install/native.mdx
```diff
@@ -1,4 +1,4 @@
-The Native SDK currently supports **Windows**, **macOS**, **Linux**, and **Android**.
+The Native SDK currently supports **Windows**, **macOS**, and **Linux**. The Native SDK also acts as an **Android** NDK support library in downstream SDKs via the [Android SDK](/platforms/android/configuration/using-ndk/) but currently does not directly support the platform.
```
vivianyentran, line 1:
```suggestion
The Native SDK currently supports **Windows**, **macOS**, and **Linux**. The Native SDK also acts as an **Android** NDK support library in downstream SDKs via the [Android SDK](/platforms/android/configuration/using-ndk/), but currently does not directly support the platform.
```

vivianyentran (approved):
LGTM thanks for the clarification!

PR #9111, opened 2024-02-12
fix: (react-native): Note line breaks in user interaction

buenaflor:
Fixes the text line breaks in note https://docs.sentry.io/platforms/react-native/performance/instrumentation/user-interaction-instrumentation


lizokm (approved):
Looks good, thanks for fixing!

PR #9113, opened 2024-02-12
feat: Add docs for custom browser tracing routing instrumentation

mydea:
This fixes one forgotten line in the react docs, and adds a section to react & browser auto-instrument docs on how to implement your own navigation instrumentation using the new APIs.

### DIFF
--- a/platform-includes/performance/automatic-instrumentation-custom-routing/javascript.mdx
+++ b/platform-includes/performance/automatic-instrumentation-custom-routing/javascript.mdx
@@ -1,8 +1,8 @@
-### Custom routing
+### Custom Routing
 
-By default, the `browserTracingIntegration()` will create a `pageload` span for when the page is initially loaded, as well as a `navigation` span whenever the URL changes afterwards.
+By default, the `browserTracingIntegration()` will create a `pageload` span for when the page is initially loaded, as well as a `navigation` span for whenever the URL changes afterwards.
 
-If you want to ensure spans are created correctly for a custom routing setup, you can opt out of the default span creation by setting `instrumentNavigation: false` & `instrumentPageLoad: false` in the `browserTracingIntegration()` options. Then, you can manually create the spans like this:
+To make sure that spans are created correctly for a custom routing setup, you'll need to opt out of the default span creation by setting `instrumentNavigation: false` and `instrumentPageLoad: false` in the `browserTracingIntegration()` options. You can then manually create spans like this:
 
 ```javascript
 Sentry.init({
@@ -27,7 +27,7 @@ let pageLoadSpan = Sentry.startBrowserTracingPageLoadSpan({
 myRouter.on("routeChange", (route) => {
   const client = Sentry.getClient();
 
-  // We want to make sure to update the pageload span to use the route name
+  // Make sure that the pageload span uses the route name
   // After that, each route change should trigger a navigation span (which will automatically finish the previous one)
   if (pageLoadSpan) {
     pageloadSpan.updateName(route.name);


platform-includes/performance/automatic-instrumentation-custom-routing/javascript.mdx
```diff
@@ -0,0 +1,24 @@
...
+```javascript
+Sentry.init({
+  integrations: [
+    // disable default navigation spans
+    Sentry.browserTracingIntegration({ instrumentNavigation: false }),
+  ],
+});
+
+// Somewhere, instrument your router like this:
+myRouter.on("routeChange", (route) => {
```
AbhiPrasad, line 16: We should add a note about setting correct route name for pageload as well, we want to avoid un-paramaterized transactions.

mydea: I guess this is mostly about updating the pageload span, right? 🤔 or should we document how to emit a pageload span?

AbhiPrasad: yeah updating the pageload span, I think most frameworks will be fine with emitting pageload as soon as possible.

mydea: I opened a PR: https://github.com/getsentry/sentry-javascript/pull/10633 to make this easier, as documenting this showed that there is def. room for improvement there!

platform-includes/performance/enable-automatic-instrumentation/javascript.mdx
```diff
@@ -19,3 +19,5 @@ Sentry.init({
   tracesSampleRate: 1.0,
 });
 ```
+
+<PlatformContent includePath="performance/automatic-instrumentation-custom-routing" />
```
Lms24, line 23: l: should this documentation live under "Automatic Instrumentation"? Isn't this rather fitting for "Custom Instrumentation"? 

I guess there's a case for both, given we'Re mentioning `browserTracingIntegration` directly above but maybe it's enough to link to custom in the "Autmatic Instrumentation" page?

mydea: yeah, IMHO I'd generally restructure this a bit in a follow up - we should rethink this with a bit of time! Currently that's basically the "browser tracing" page, in reality, even though the name is not necessarily clear on that.

mydea:
I have updated this based on the changes in v7.101.0. Now it is a "full" example including updating the pageload span.

platform-includes/performance/automatic-instrumentation-custom-routing/javascript.mdx
```diff
@@ -0,0 +1,46 @@
+### Custom routing
```
lizokm, line 1:
```suggestion
### Custom Routing
```

platform-includes/performance/automatic-instrumentation-custom-routing/javascript.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+By default, the `browserTracingIntegration()` will create a `pageload` span for when the page is initially loaded, as well as a `navigation` span whenever the URL changes afterwards.
+
+If you want to ensure spans are created correctly for a custom routing setup, you can opt out of the default span creation by setting `instrumentNavigation: false` & `instrumentPageLoad: false` in the `browserTracingIntegration()` options. Then, you can manually create the spans like this:
```
lizokm, line 5:
```suggestion
By default, the `browserTracingIntegration()` will create a `pageload` span for when the page is initially loaded, as well as a `navigation` span for whenever the URL changes afterwards.

To make sure that spans are created correctly for a custom routing setup, you'll need to opt out of the default span creation by setting `instrumentNavigation: false` and `instrumentPageLoad: false` in the `browserTracingIntegration()` options. You can then manually create spans like this:
```

platform-includes/performance/automatic-instrumentation-custom-routing/javascript.mdx
```diff
@@ -0,0 +1,46 @@
...
+    [Sentry.SEMANTIC_ATTRIBUTE_SENTRY_SOURCE]: "url",
+  },
+});
+
+// Somewhere, instrument your router like this:
+myRouter.on("routeChange", (route) => {
+  const client = Sentry.getClient();
+
+  // We want to make sure to update the pageload span to use the route name
+  // After that, each route change should trigger a navigation span (which will automatically finish the previous one)
```
lizokm, line 31:
```suggestion
  // Make sure that the pageload span uses the route name
  // After that, each route change should trigger a navigation span (which will automatically finish the previous one)
```

lizokm (approved):
LGTM, just a few small tweaks

PR #9116, opened 2024-02-12
ref(js): Update and fix browser profiling integration code snippets

Lms24:
Just noticed via https://github.com/getsentry/sentry-javascript/issues/10599 that some of the browser profiling docs still mentioned the deprecated integration classes


platform-includes/profiling/automatic-instrumentation-setup/javascript.astro.mdx
Lms24, line 1: Added a new snippet for Astro as it also already includes the `browserTracingIntegration`

AbhiPrasad (approved):

PR #9126, opened 2024-02-13
feat: Cache api schema for build

HazAT:
Vercel build was error because we make a request for every page we generate, this add a simple cache.

### DIFF
--- a/src/build/resolveOpenAPI.ts
+++ b/src/build/resolveOpenAPI.ts
@@ -14,6 +14,8 @@ const SENTRY_API_SCHEMA_SHA = 'deea76e7205d4e1efb45f91796b9fc73499de226';
 
 const activeEnv = process.env.GATSBY_ENV || process.env.NODE_ENV || 'development';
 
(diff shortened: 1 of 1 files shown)


lforst (approved):

cleptric (approved):

lforst:
You may wanna use [react's `cache()` fn](https://react.dev/reference/react/cache) for this. Don't know if it works for static things like that tho.

HazAT:
@lforst  too fancy, merging now

mjq-sentry:
> You may wanna use [react's `cache()` fn](https://react.dev/reference/react/cache) for this. Don't know if it works for static things like that tho.

Actually the exported method that calls this function (`apiCategories`) already uses react's `cache` as of #9069, so tbh I'm really surprised this was still making a request per page 🤔

lforst:
The build logs mention something about something not being able to be cached bc the data is >2MB. Maybe it's not caching it because of that - which would be weird because especially in cases where the data is big, I want it cached. 🤔

PR #9139, opened 2024-02-14
Deno arguments are not in the correct order

smeubank:
When testing this Deno CLI complained about the order of arguments in the script. This is how i tried it from the docs

image

```
me@me deno-edge-hello-world % deno run index.ts --allow-net=xxxxx.ingest.sentry.io

Permission flags have likely been incorrectly set after the script argument.
To grant permissions, set them before the script argument. For example:
    deno run --allow-read=. main.js
error: Module not found "file:///Users/steveneubank/Desktop/deno-edge-hello-world/index.ts".
```

Flipping the order as instructed works

### DIFF
--- a/platform-includes/getting-started-config/javascript.deno.mdx
+++ b/platform-includes/getting-started-config/javascript.deno.mdx
@@ -15,5 +15,5 @@ To ensure the SDK can send events, you should enable network access for your
 ingestion domain:
 
 ```bash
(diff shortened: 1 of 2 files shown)


vivianyentran (approved):
Thanks for catching this and updating the code samples!

lforst (approved):

AbhiPrasad (approved):

PR #9141, opened 2024-02-14
fix(native): provide an example for the breadcrumb data property

supervacuus:
In the same vein as this: https://github.com/getsentry/sentry-native/pull/951

There is currently no example in the docs that clarifies the necessity of a wrapping object for the `data` property in breadcrumbs. Users tried to assign JSON as a string for instance.

The example is straight out of the develop docs: https://develop.sentry.dev/sdk/event-payloads/breadcrumbs/#breadcrumb-types. Not sure if we should mention this, because the surroundings docs already do.

cc: @kahest


vivianyentran (approved):
LGTM, thanks for adding this clarification!

supervacuus:
Hi @vivianyentran and @kahest. How can I help to get this example merged?

PR #9149, opened 2024-02-15
Cleanup Astro upload page

smeubank:
Astro's source map uploads includes a lot of snippet variations. I think this is distracting and we can stick what we promtoe and reduce the rest, and move the legacy token to the legacy uploading methods


Lms24 (approved):
Looks reasonable to me overall but particular strong opinions about the changes. The snippets are probably unimportant enough to omit them in favour of the "set `property` to `value` under `enclosingObject`" phrase. 

Agree that the Old auth token stuff can probably be moved to legacy by now.

PR #9167, opened 2024-02-16
feat: Fix contribution docs

HazAT:
We also need to update and sync https://github.com/getsentry/sentry-docs/blob/master/docs/contributing/environment.mdx

Not sure where we want to source of truth

- [ ]  In local dev environment we need to change that we don't upload to the google bucket and instead just store images locally.


a-hariti:
Revisiting this @HazAT 

I'm getting a 405 when I try to login locally, any idea about this?

HazAT:
@a-hariti you urls must be localhost:3000 is it the case?

a-hariti:
Yes, I'm trying to run it locally @HazAT. Adding a seeding script would be helpful as well :)

HazAT:
I couldn't figure out how to do this quickly in Next.js - I had it temporarily available in route to debug it
Go ahead and try it, I can provide you with a dump our our database if that helps

a-hariti:
Seeding is very well documented [here](https://www.prisma.io/docs/orm/prisma-migrate/workflows/seeding), I just need to be able to run the app on localhost and take it from there @HazAT

HazAT:
I added some seeds so it should be possible to at least see something.
The problem with the google login is tricky since it's locked to Sentry Google accounts and I don't want to spend time logic-change NextAuth to not use google login for local development.

I am going to merge it like this since it's better than nothing but still not complete

PR #9173, opened 2024-02-16
Add sitemap.xml file

mjq-sentry:
Generate a sitemap.xml during build time that is served statically at `/sitemap.xml`, and add a reference to it in `robots.txt` so that crawlers can find it. (See: https://developers.google.com/search/docs/crawling-indexing/sitemaps/build-sitemap#addsitemap)

The sitemap entries' `lastmod` property comes from the last modification time of the source MDX files for each page. Note that this might not be completely accurate, as updates to MDX fragments in `/includes` or `/platform-includes` won't be considered.

Pages generated from our OpenAPI specs are missing `lastmod` properties since we don't have a way to isolate the timestamp when individual pages change.

### DIFF
--- a/app/sitemap.ts
+++ b/app/sitemap.ts
@@ -1,0 +1,21 @@
+import fs from 'fs';
+import path from 'path';
+
+import type {MetadataRoute} from 'next';
+
+import {getDocsFrontMatter} from 'sentry-docs/mdx';
+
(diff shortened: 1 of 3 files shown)


lizokm (approved):
LGTM!

Jeffreyhung:
Wow we have 5300+ pages for our doc site https://sentry-docs-git-mjq-sitemap.sentry.dev/sitemap.xml

PR #9200, opened 2024-02-22
feat: Add `PlatformOrGuideName` and `PlatformSdkPackageName` components

Lms24:
This PR adds two components that return the name and SDK package name of the currently selected platform.

- `PlatformSdkPackageName` returns the package name, e.g. `@sentry/nextjs`
- `PlatformOrGuideName` returns the name of the currently selected platform or guide, e.g. `Next.JS`

Motivation: We have a lot of pages where we statically write a package name (e.g. `@sentry/browser`) which doesn't change when users select a guide (e.g. NextJS). This leads to confusion. Reportedly from SEs, thought they'd need to install _and initialize_ the browser SDK in addition to the NextJS SDK. 

Alternatives: Sure, we could also copy the guide per platform (IIRC we did this for Replay a while back) but it's arguably more work if the changes between these files are literally the name of a package or framework.

Let me know what you think, happy to adjust or close this if you have better ideas!

[Example page with dynamic sdk/platform name](https://sentry-docs-git-lms-feat-add-platform-name-and-package-c-73dfb8.sentry.dev/platforms/javascript/guides/nextjs/profiling/#prerequisites) 

PS: Please assume I know next to nothing about NextJS or RSC when reviewing this PR ;)

### DIFF
--- a/docs/platforms/javascript/common/profiling/index.mdx
+++ b/docs/platforms/javascript/common/profiling/index.mdx
@@ -26,15 +26,15 @@ Note that since the profiling API is currently only exposed in Chromium, profile
 
 To get started with JavaScript browser profiling, you'll need to:
 
-- Install the @sentry/browser SDK, minimum version 7.60.0
+- Install the <PlatformSdkPackageName fallback="@sentry/browser"/> SDK, minimum version 7.60.0
 - Configure the document response header to include `Document-Policy: js-profiling`
 - Configure the SDK to use the `BrowserProfilingIntegration` and set `profilesSampleRate`
 
-## Step 1: Install the JavaScript Browser SDK
+## Step 1: Install the <PlatformOrGuideName/> SDK
 
 <PlatformSection notSupported={["javascript.electron"]}>
 
-Install our JavaScript browser SDK using either `yarn` or `npm`, the minimum version that supports profiling is **7.60.0**.
+Install our <PlatformOrGuideName/> SDK using either `yarn` or `npm`, the minimum version that supports profiling is **7.60.0**.
 
 </PlatformSection>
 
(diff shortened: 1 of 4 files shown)


docs/platforms/javascript/common/profiling/index.mdx
```diff
@@ -26,15 +26,15 @@ Note that since the profiling API is currently only exposed in Chromium, profile
...
 - Configure the document response header to include `Document-Policy: js-profiling`
 - Configure the SDK to use the `BrowserProfilingIntegration` and set `profilesSampleRate`
 
-## Step 1: Install the JavaScript Browser SDK
+## Step 1: Install the <PlatformOrGuideName/> SDK
 
 <PlatformSection notSupported={["javascript.electron"]}>
 
-Install our JavaScript browser SDK using either `yarn` or `npm`, the minimum version that supports profiling is **7.60.0**.
+Install our <PlatformOrGuideName/> SDK using either `yarn` or `npm`, the minimum version that supports profiling is **7.60.0**.
```
Lms24, line 37: This is the concrete use case I want to fix (which was reported by our sales engineers today in Slack).

src/components/platformOrGuideName.tsx
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Displays a readable name of the currently selected platform or guide.
+ * Example: `Next.js`.
+ */
+export function PlatformOrGuideName({fallback}: PlatformOrGuideNameProps) {
+  const fallbackName = fallback || 'Sentry';
+  const {rootNode, path} = serverContext();
+  const platformOrGuide = rootNode && getCurrentPlatformOrGuide(rootNode, path);
+  if (!platformOrGuide) {
+    return 'fallbackName ';
```
mjq-sentry, line 21: Should this be the value of `fallbackName` rather than the literal string?

mjq-sentry (changes requested):
Concept sounds great to me. Commented re: one suspicious line, otherwise LGTM!

src/components/platformOrGuideName.tsx
Lms24: oops, good catch, thx! fixed it

mjq-sentry (approved):
Thanks! 🚢

PR #9204, opened 2024-02-22
chore: Add codecov bundle analysis to docs repo

AbhiPrasad:
This adds support for codecov bundle analysis to the sentry docs, so we can dogfood it.

https://app.codecov.io/gh/getsentry/sentry-docs/bundles/new

This is dependent on adding the codecov token to our docs, so tbd on that.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -20,6 +20,21 @@ const nextConfig = {
     serverComponentsExternalPackages: ['rehype-preset-minify'],
   },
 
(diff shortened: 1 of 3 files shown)


AbhiPrasad:
Alright going to merge this in! We have bundle info being reported https://app.codecov.io/gh/getsentry/sentry-docs/bundles - let's see how it looks when doing a comparison.

Does seem to impact CI times at all which is nice.

PR #9206, opened 2024-02-22
docs(metrics-extraction): Add mention of metrics extraction to docs

k-fish:
### Summary
This adds a mention of metrics extraction to the docs.


docs/product/dashboards/custom-dashboards/index.mdx
```diff
@@ -42,6 +42,12 @@ Dashboard filters are applied to all the widgets in your dashboard. They consist
 
 When adding a widget or editing an existing one, you will be taken to a new page where you can see the dashboard widget builder. The widget builder has several configuration options that you can use to shape data and add information to your dashboard. Learn more in the [Widget Builder documentation](/product/dashboards/widget-builder/).
 
+### Metrics Extraction
+
+<Include name="feature-available-for-user-group-early-adopter.mdx" />
+
+If you're using custom tags or custom filters for in your widgets, we will start collecting metrics to give you accurate counts and rates. During widget creation or editing you will see approximate data based on indexed events, but as soon as you save we will start extracting metrics for those widgets.
```
k-fish, line 49: Screenshot 2024-02-22 at 3 00 19 PM

We can maybe add an image?

armenzg: That's a good idea!

docs/product/dashboards/custom-dashboards/index.mdx
```diff
@@ -42,6 +42,16 @@ Dashboard filters are applied to all the widgets in your dashboard. They consist
...
+
+When you are using custom tags, the following features are currently not compatible with metrics extraction:
+- Equations
+- Widgets that display data from errors and transactions simultaneously
```
armenzg, line 53: Should we include any tips as to what to do?

armenzg (approved):
Some of this will not be applicable until we have opted-in their org for extraction, do we need to say anything about it?

docs/product/dashboards/custom-dashboards/index.mdx
vivianyentran: Yes, agree on adding an image!

docs/product/dashboards/custom-dashboards/index.mdx
```diff
@@ -42,6 +42,16 @@ Dashboard filters are applied to all the widgets in your dashboard. They consist
 
 When adding a widget or editing an existing one, you will be taken to a new page where you can see the dashboard widget builder. The widget builder has several configuration options that you can use to shape data and add information to your dashboard. Learn more in the [Widget Builder documentation](/product/dashboards/widget-builder/).
 
+### Metrics Extraction
+
+<Include name="feature-available-for-user-group-early-adopter.mdx" />
+
+If you're using custom tags or custom filters for in your widgets, we will start collecting metrics to give you accurate counts and rates. During widget creation or editing you will see approximate data based on indexed events, but as soon as you save we will start extracting metrics for those widgets.
```
vivianyentran, line 49:
```suggestion
If you're using custom tags or custom filters in your widgets, we will start collecting metrics to give you accurate counts and rates. During widget creation or editing, you will see approximate data based on indexed event. As soon as you save, we will start extracting metrics for those widgets. It may take up to ___ days/weeks to start seeing metrics. 
```
Not sure if we have a timeframe for when the metrics start showing or if it's immediate?

vivianyentran (commented):
Added some comments. I think we should also add some info to the widget builder page, like a short paragraph at the top before the "Widget Title" section.

docs/product/dashboards/custom-dashboards/index.mdx
k-fish: We can include these on the widget builder page I think. I'm going to add it and link there instead.

k-fish:
@vivianyentran thanks for the review! I actually think the Widget Builder is just a better location since we can elaborate on the some of the limitations more there, so I moved most of the docs and the image there and then refer to it on the custom-dashboards page with a link, let me know what you think.

docs/product/dashboards/custom-dashboards/index.mdx
```diff
@@ -42,6 +42,14 @@ Dashboard filters are applied to all the widgets in your dashboard. They consist
 
 When adding a widget or editing an existing one, you will be taken to a new page where you can see the dashboard widget builder. The widget builder has several configuration options that you can use to shape data and add information to your dashboard. Learn more in the [Widget Builder documentation](/product/dashboards/widget-builder/).
 
+### Metrics Extraction
+
+<Include name="feature-available-for-user-group-early-adopter.mdx" />
+
+If you're using custom tags or custom filters in your widgets using the transaction dataset, we will start collecting metrics to give you accurate counts and rates. During widget creation or editing, you will see approximate data based on indexed event. As soon as you save, we will start extracting metrics for those widgets. It may take up to 15 minutes to start seeing metrics.
+
+[Learn more about the feature and limitations of metrics extraction.](/product/dashboards/widget-builder/#extracted-metrics)
```
vivianyentran, line 51:
```suggestion
Learn more about the feature and limitations of [metrics extraction](/product/dashboards/widget-builder/#extracted-metrics).
```

docs/product/dashboards/custom-dashboards/index.mdx
```diff
@@ -42,6 +42,14 @@ Dashboard filters are applied to all the widgets in your dashboard. They consist
 
 When adding a widget or editing an existing one, you will be taken to a new page where you can see the dashboard widget builder. The widget builder has several configuration options that you can use to shape data and add information to your dashboard. Learn more in the [Widget Builder documentation](/product/dashboards/widget-builder/).
 
+### Metrics Extraction
+
+<Include name="feature-available-for-user-group-early-adopter.mdx" />
+
+If you're using custom tags or custom filters in your widgets using the transaction dataset, we will start collecting metrics to give you accurate counts and rates. During widget creation or editing, you will see approximate data based on indexed event. As soon as you save, we will start extracting metrics for those widgets. It may take up to 15 minutes to start seeing metrics.
```
vivianyentran, line 49:
```suggestion
If you're using custom tags or custom filters in your widgets, we will start collecting metrics to give you accurate counts and rates. During widget creation or editing, you will see approximate metrics for transaction datasets based on indexed events. As soon as you save, we will start extracting metrics for those widgets. It may take up to 15 minutes to start seeing metrics.
```

docs/product/dashboards/widget-builder/index.mdx
```diff
@@ -132,3 +132,17 @@ The option for setting thresholds is only available for [big number visualizatio
 
 ![Set Thresholds Step](/product/dashboards/widget-builder/widget-builder-thresholds.png)
 ![Widget Coloured Indicator Hover](/product/dashboards/widget-builder/widget-builder-thresholds-hover.png)
+
+## Extracted metrics
+
+<Include name="feature-available-for-user-group-early-adopter.mdx" />
+
+If you are building a widget with transaction data, and would like to use event-like filters (eg. transaction.duration:>100ms) or custom tags that you set on your transactions, we will routinely collect metrics to show you accurate count and rates for you.
```
vivianyentran, line 140:
```suggestion
If you are building a widget with transaction data, and would like to use event-like filters (eg. `transaction.duration:>100ms`) or custom tags that you set on your transactions, we will routinely collect [metrics](/product/metrics/) to show you accurate counts and rates.
```

I was thinking we should define what we mean by metrics somewhere, since it's not super clear to me how it's different from event data

docs/product/dashboards/widget-builder/index.mdx
```diff
@@ -132,3 +132,17 @@ The option for setting thresholds is only available for [big number visualizatio
...
+
+![Extracted Metrics Tooltip](/product/dashboards/widget-builder/widget-builder-extracted-metrics-tooltip.png)
+
+Not all widgets are compatible with extracted metrics however, the following cases will cause your widget to continue using event backed data:
+- Equations
+-- Using equations will show you event data instead of metrics
+- Widgets that display data from errors and transactions simultaneously, or use a custom tag or filter without also filtering on `event.type`
+-- Extraction only happens for transaction data, if you add `event.type:error` or `event.type:transaction` to your filter along with your custom tags the data will us errors or the extracted metrics as intended.
```
vivianyentran, line 148:
```suggestion
Not all widgets are compatible with metrics extraction. The following cases will cause your widget to continue using event-backed data:
- Equations — Using equations will show you event data instead of metrics.
- Widgets that display data from errors and transactions simultaneously — Extraction only happens for transaction data. You need to use the `event.type:transaction` filter along with your custom tags in order to get extracted metrics.
```
I found the last bullet with event.type really confusing. Not sure if I'm understanding this right so please feel free to correct me here!

vivianyentran (approved):
Added some comments but I think the info fits a lot better in the widget builder page! Thanks for reorganizing the content!

docs/product/dashboards/widget-builder/index.mdx
k-fish: Unfortunately it's very confusing to everyone because we've been building a lot in the last year or so 😄 . What we collect here and powering the performance pages _are_ conceptually 'metrics' but they are stored in unique rows, not really associated with the `/product/metrics` feature we've built, since they're meant to shadow performance data. You can, in theory, use `metrics` to access the metrics mentioned via transactions / performance but the app isn't really setup that way.

docs/product/dashboards/widget-builder/index.mdx
k-fish: Yeah I was struggling to make this concise. I'll explain a bit more, maybe you'll have an idea? Sorry for the incoming wall of text 😄 

If you pick the "Errors and Transactions" dataset here, it's hard for us technically to tell what to display now, since errors and transactions aren't really the same thing, they're stored differently, the get filtered differently etc. Metrics extraction works for transactions, but it doesn't work or even apply to errors since they aren't dynamically sampled in the first place. 
Screenshot 2024-02-23 at 3 05 26 PM

What data the widget should display is _ambiguous_ when a user uses custom tags. Previously we'd show entirely **event** data, which works since the events for errors and transactions are stored the same place, but when you start writing widgets that should be transactions, and specifically need metrics for them, suddenly we can't easily merge that data together. 

Widgets in general don't have a great reason to display _both_ errors & transactions at the same time, that's just how the dataset was initially created (and hopefully we split them up sometime soon since it's technical debt, imo). 

Unfortunately when a user uses a custom tag, we can't tell (without a bunch of technical work to optimize check tags) if that custom tag is intended to be for transaction data, in which case extracted metrics can be used, or if they actually were using a custom tag they set on an error.

The only workaround to this problem we have at the moment is if a user adds `event.type:error` or `event.type:transaction` in those cases, we can be sure the widget is meant to be one or the other. I've added / hidden UI to make it perhaps a bit more straightforward.
|error data|transaction data|
|-|-|
| Screenshot 2024-02-23 at 3 10 41 PM | Screenshot 2024-02-23 at 3 10 48 PM |

vivianyentran: Thanks for the context, that makes more sense now. Is the event.type:transaction filter required for metrics extraction? Or will metrics get extracted for custom tags for transactions, even without specifying the event.type?

k-fish: We extract both just in case, but when we visualize it in a table or graph, we're assuming `event.type:transaction` implicitly, so when we turn it on some people are going to be seeing a few blank widgets if it was meant to be errors. We've refined it a bit so error-like tags show you errors by default instead, but it's not perfect since we don't know their intent.

docs/product/dashboards/widget-builder/index.mdx
vivianyentran: Hmm, ok let's remove the link to the metrics product then. But can you define metrics here?  Is the difference between metrics and event data that metrics include all data whereas event data is sampled?

docs/product/dashboards/widget-builder/index.mdx
vivianyentran: Hmm ok, how's this:

> Not all widgets are compatible with metrics extraction. The following cases will cause your widget to continue using event-backed data:
> - Equations — Using equations will show you event data instead of metrics.
> - Widgets that display data from errors and transactions simultaneously — Extraction only happens for transaction data. If you see a blank widget, your dataset probably included errors. To ensure the dataset only has transaction data, you can use the `event.type:transaction` filter along with your custom tags to get extracted metrics.

lizokm:
@k-fish just a friendly reminder about this doc :)

k-fish:
@lizokm yep, I didn't forget, I'm leaving it open since I'm considering changing how some of this works, which would invalidate a portion of this.

docs/product/dashboards/widget-builder/index.mdx
k-fish: > Is the difference between metrics and event data that metrics include all data whereas event data is sampled?

That's exactly right, your visualization is only accurate in count etc. when you're using `metrics` data since it's not dynamically sampled.

PR #9207, opened 2024-02-22
Fix caching during build

mjq-sentry:
React's `cache` function does not work the way I expected, and actually wasn't caching properly. Replace usages of it with memoization, which is more verbose but definitely works.

Locally, this sped up `yarn build` to 201 seconds, which is much faster than the previous local build time.


mjq-sentry:
Reduced the Vercel build time from ~20 minutes to ~14 minutes 🚀 Getting there!

lforst (approved):
huh, so `cache` doesn't work at build time?

PR #9258, opened 2024-02-27
Move auth path top level

HazAT:
It breaks the login for changelog.


mjq-sentry (approved):
I would expect the `/api/auth` path to conflict with the doc at that URL, but according to the preview it doesn't, so 🤷🚢

PR #9299, opened 2024-03-01
fix(native): clarify use of crashpad in macOS sandbox

supervacuus:
This an outcome of a recent support discussion with one of our users: https://github.com/getsentry/sentry-native/discussions/881#discussioncomment-8640854

cc: @kahest


kahest (approved):
Thanks for adding this! LGTM

PR #9361, opened 2024-03-06
Adapt Laravel docs for version 11.0

cleptric:
Not to be merged before March 12th.

### DIFF
--- a/docs/platforms/php/guides/laravel/configuration/laravel-options.mdx
+++ b/docs/platforms/php/guides/laravel/configuration/laravel-options.mdx
@@ -91,6 +91,9 @@ These settings have no effect if `SENTRY_TRACES_SAMPLE_RATE` is set to `0.0` or 
     // Capture SQL queries as spans
     'sql_queries' => env('SENTRY_TRACE_SQL_QUERIES_ENABLED', true),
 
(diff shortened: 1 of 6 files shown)


docs/platforms/php/guides/laravel/index.mdx
```diff
@@ -17,17 +17,28 @@ Install the `sentry/sentry-laravel` package:
 composer require sentry/sentry-laravel
 ```
 
-Enable capturing unhandled exception to report to Sentry by making the following change to your `App/Exceptions/Handler.php`:
+Enable capturing unhandled exception to report to Sentry by making the following change to your `bootstrap/app.php`:
```
vivianyentran, line 20: Should we call out that this is now a different file? Do users need to do anything with `App/Exceptions/Handler.php` if they previously had it?

vivianyentran (approved):
Looks good! Thanks for adding the docs for this version!

docs/platforms/php/guides/laravel/index.mdx
cleptric: This is not a change in the SDK but instead in the framework. But the old way will still work for people upgrading from Laravel 10 to Laravel 11.

stayallive (approved):
Looks good to me 💪

PR #9373, opened 2024-03-07
feat: Format code blocks at build time

### LINKED ISSUE #9057: Format code blocks during build for consistent styling
In the past, we had prettier in the repo which had the positive side effect that all (JS) code blocks in the docs had consistent formatting. Since we identified that prettier does not work well with current MDX we will lose this benefit.

However, It should be possible to format the code blocks during build, regardless of the input. With this I don't mean to actually modify commited code, but to sortof transpile the code blocks during the build, so that they are displayed formatted on the website.

We could apply different formatters for different languages and we may even use prettier for the JS code blocks.

a-hariti:
## Pre-merge checklist

## Description of changes

Format code blocks at build time in a transpilation fashion using Prettier

Ignores code blocks with `diff` metadata to avoid confusion due to leading `+` and `-` characters

Right now supports languages that are supported of the box by Prettier, there's room for more support for other languages through plugins

resolves #9057 


src/remark-format-code.js
```diff
@@ -0,0 +1,47 @@
...
+  };
+  const parser = parsers[lang];
+  if (!parser) {
+    return;
+  }
+
+  try {
+    const formattedCode = await format(node.value, {parser: parsers[lang]});
+    // get rid of the trailing newline
+    node.value = formattedCode.trimEnd();
```
lforst, line 42: Instead of manually trimming the end, could we pass an option to `format?`

a-hariti: Probably not, [options](https://prettier.io/docs/en/options)

a-hariti:
PHP plugin seems to choke on the top level `public` keyword on [this line](https://github.com/getsentry/sentry-docs/blob/master/docs/platforms/php/guides/laravel/index.mdx#L25) although it's a [normal pattern](https://laravel.com/docs/10.x/errors#reporting-exceptions) in Laravel

I will wait on this for now until we have a clear idea of the preferred code formatting conventions among PHP developers

lforst (commented):
If this works :shipit:

lforst (approved):

Issue #5026, opened 2022-05-09
Add GH actions for image optimization


imatwawana:
### Core or SDK?

Core Sentry product

### Which part? Which one?

All (applies to SDK and Product docs)

### Description

Right now, we have to depend on folks to remember to only use images that have the correct resolution and that have been compressed in Docs. 

### Suggested Solution

Add image compression and resolution warnings as part of GitHub actions for PRs in this repo.

PR #9379, opened 2024-03-07
Image optimization

a-hariti:
## Pre-merge checklist

## Description of changes
1. use `next/image`
2. use relative paths to move images next to referring md[x] files.

**TODO**
- [ ] update contributing -> images docs

Related #5026


mjq-sentry:
Okay, very neat! So to confirm my understanding, the core changes (ignoring the image moves) are:
* adding the `remark-mdx-images` package
* updating the mdx configuration to include the `remarkMdxImages` plugin, which takes each referenced image and moves it into `/public/mdx-images/`
* updating the `DocImage` component to rewrite the relative paths in the MDX to the absolute path to the image in `/public/mdx-images/`

Cool! A couple notes:
* I tested this locally in both dev and prod mode, and both worked 👍 
* It looks like `remarkMdxImages` also takes care of adding a hash to the file name, which is great in case two different images are in different folders but with the same name (since all files ends up in one flat `mdx-images` folder).
* Before all the files were moved, originally image files were in the same folder as the MDX file that referenced them, instead of in an `./img` folder relative to the file. I don't know if this is something that the docs team has an opinion on.

It looks like there are now some merge conflicts, and also a build issue in Vercel. Please fix the conflicts to trigger another Vercel build - as long as this all works in the preview there then I think this looks good!

a-hariti:
Your understanding is mostly true @mjq-sentry 

- the remark-mdx-images package adds an import at top of the mdx file
- `mdx-bundler` + `esbuild` take care of copying the images
- I'm not sure who adds the hashes, but it's a nice feature
- `DocImage` got changed again (amidst the noise of the huge commits), but overall if checks if the image has a relative path to rewrite it to `/mdx-images`, otherwise it behaves as before

src/remark-image-size.js
```diff
@@ -0,0 +1,27 @@
+import path from 'path';
+
+import getImageSize from 'image-size';
+import {visit} from 'unist-util-visit';
+
+/**
+ * appends the image size to the image url as a hash e.g. /img.png -> /img.png#100x100
```
a-hariti, line 7: @mjq-sentry What do you think about this trick?

lizokm: @mjq-sentry would appreciate you taking a look when you have a chance!

mjq-sentry: Doing this as a `data-` attribute might be another option and a little less hacky, but I'm okay with trying this for now. The benefits outweigh the drawbacks in the short term.

a-hariti:
Maybe putting shared images under `/public/shared` should make it more explicit? @mjq-sentry 

and I agree, the more we wait the more merge conflicts arise

mjq-sentry:
> Maybe putting shared images under /public/shared should make it more explicit? @mjq-sentry

That sounds great to me. If you want to address now or in a future PR, whichever works. If you can fix the conflicts one last time, we can merge this ASAP. Thanks for your patience!

PR #9385, opened 2024-03-08
[python] Condense Profiling Setup Docs

smeubank:
Combine snippets and move the upgrade to trouble shooting


JonasBa:
@smeubank I think the yarn and npm lock file changes somehow creeped in here 😅

smeubank:
> @smeubank I think the yarn and npm lock file changes somehow creeped in here 😅

out

PR #9386, opened 2024-03-08
[ruby] Condense Profiling Setup Docs

smeubank:
Combine snippets, no stand alone for enabling performance

### DIFF
--- a/docs/platforms/ruby/common/profiling/index.mdx
+++ b/docs/platforms/ruby/common/profiling/index.mdx
@@ -44,6 +44,6 @@ end
 
 The <PlatformIdentifier name="profiles_sample_rate" /> setting is _relative_ to the <PlatformIdentifier name="traces_sample_rate" /> setting.
 
-Profiling requires Sentry’s performance monitoring to be enabled via `traces_sample_rate` in the example above. Read our <PlatformLink to="/performance/">performance setup documentation</PlatformLink> to learn how to configure sampling. Setting the sample rate to 1.0 means all transactions will be captured.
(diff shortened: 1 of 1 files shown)


JonasBa (approved):

docs/platforms/ruby/common/profiling/index.mdx
```diff
@@ -59,4 +44,6 @@ end
 
 The <PlatformIdentifier name="profiles_sample_rate" /> setting is _relative_ to the <PlatformIdentifier name="traces_sample_rate" /> setting.
 
+Profiling requires Sentry’s performance monitoring to be enabled via `traces_sample_rate` in the example above. Read our <PlatformLink to="/performance/">performance setup documentation</PlatformLink> to learn how to configure sampling. Setting the sample rate to 1.0 means all transactions will be captured.
+
```
lizokm, line 48:
```suggestion
For Profiling to work, you have to first enable Sentry’s performance monitoring via `traces_sample_rate` (like in the example above). Read our <PlatformLink to="/performance/">performance setup documentation</PlatformLink> to learn how to configure sampling. If you set your sample rate to 1.0, all transactions will be captured.

```
@smeubank I think this will actually stand out more if it's not in the format of a note. There are so many notes on this page that it makes it easy to ignore them :)

lizokm (approved):
One suggestion, otherwise this looks great.

docs/platforms/ruby/common/profiling/index.mdx
smeubank: i don't actually want it to standout more, i think this note level info, it is in the copy paste to include the sample rate and explanation below the page

PR #9396, opened 2024-03-08
use next mdx

scttcper:
was trying to figure out a hydration error in code blocks and ended up going down a really weird path


a-hariti:
It doesn't build.

Also we're exploring a path to bring images closer their referencing source files using `mdx-bundler` through `esbuild`.
Any idea how `next-mdx-remote` will behave in this regard?

scttcper:
@a-hariti this is mostly a demo to see if next-mdx-remote fixed a hydration error in the prisma rehype output. It didn't fit it. Figured the more official solution might be better, but it seems to be very similar

PR #9415, opened 2024-03-11
feat(ruby): Add metrics docs

sl0thentr0py:
to merge after release

### DIFF
--- a/docs/platforms/ruby/common/metrics/index.mdx
+++ b/docs/platforms/ruby/common/metrics/index.mdx
@@ -12,7 +12,7 @@ Metrics for Ruby are supported with Sentry Ruby SDK version `5.17.0` and above.
 
 Sentry metrics help you pinpoint and solve issues that impact user experience and app performance by measuring the data points that are important to you. You can track things like processing time, event size, user signups, and conversion rates, then correlate them back to tracing data in order to get deeper insights and solve issues faster.
 
-First, turn on metrics usage in the SDK as follows.
+First, turn on metrics usage in the SDK as follows:
 
 ```ruby
 Sentry.init do |config|
--- a/docs/platforms/ruby/common/metrics/index.mdx
+++ b/docs/platforms/ruby/common/metrics/index.mdx
@@ -95,7 +95,7 @@ Sentry.init do |config|
 end
 ```
 
-## Filtering / Modifying Tags
+## Filtering and Modifying Tags
 
 You can filter some keys or update tags on the fly with the `before_emit` callback, which will be triggered before a metric is aggregated.
 
--- a/docs/platforms/ruby/common/metrics/index.mdx
+++ b/docs/platforms/ruby/common/metrics/index.mdx
@@ -6,7 +6,7 @@ description: "Learn how to measure the data points you care about by configuring
 <Include name="feature-stage-alpha-metrics.mdx" />
 <Note>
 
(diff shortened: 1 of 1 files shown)


docs/platforms/ruby/common/metrics/index.mdx
```diff
@@ -0,0 +1,113 @@
+---
+title: Set Up Metrics
+description: "Learn how to measure the data points you care about by configuring Metrics in your Ruby app."
+---
+
+<Include name="feature-stage-alpha-metrics.mdx" />
+<Note>
+
+Metrics for Ruby are supported with Sentry Python SDK version `5.17.0` and above.
```
vivianyentran, line 9:
```suggestion
Metrics for Ruby are supported with Sentry Ruby SDK version `5.17.0` and above.
```
Assuming you mean Ruby not Python SDK

docs/platforms/ruby/common/metrics/index.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+</Note>
+
+Sentry metrics help you pinpoint and solve issues that impact user experience and app performance by measuring the data points that are important to you. You can track things like processing time, event size, user signups, and conversion rates, then correlate them back to tracing data in order to get deeper insights and solve issues faster.
+
+First, turn on metrics usage in the SDK as follows.
```
vivianyentran, line 15:
```suggestion
First, turn on metrics usage in the SDK as follows:
```

docs/platforms/ruby/common/metrics/index.mdx
```diff
@@ -0,0 +1,113 @@
...
+```ruby
+# Measure the time of execution of the `process` method in milliseconds.
+Sentry::Metrics.timing('how_long_ms', unit: 'millisecond') do
+  process
+end
+```
+
+## Code Locations
+
+The SDK will send code locations for unique metrics (defined by type, key and unit) once a day or once per startup/shutdown of your application.
```
vivianyentran, line 88:
```suggestion
By default, the SDK will send code locations for unique metrics (defined by type, key and unit) once a day and with every startup/shutdown of your application.
```
I'm assuming it's once a day and with each startup, but feel free to revert this if that's not the case!

docs/platforms/ruby/common/metrics/index.mdx
```diff
@@ -0,0 +1,113 @@
...
+```ruby
+Sentry.init do |config|
+  # ...
+  config.metrics.enable_code_locations = false
+end
+```
+
+## Filtering / Modifying Tags
+
+You can filter some keys or update tags on the fly with the `before_emit` callback that will be triggered before a metric is aggregated.
```
vivianyentran, line 100:
```suggestion
You can filter some keys or update tags on the fly with the `before_emit` callback, which will be triggered before a metric is aggregated.
```

docs/platforms/ruby/common/metrics/index.mdx
```diff
@@ -0,0 +1,113 @@
...
+You can turn this off with the following:
+
+```ruby
+Sentry.init do |config|
+  # ...
+  config.metrics.enable_code_locations = false
+end
+```
+
+## Filtering / Modifying Tags
```
vivianyentran, line 98:
```suggestion
## Filtering and Modifying Tags
```

vivianyentran (approved):
Couple of small edits but otherwise this looks good! Thanks for adding it!

PR #9424, opened 2024-03-12
fix(webvitals): Fix links to web vitals not pointing to correct page

edwardgou-sentry:
We moved the webvitals concepts into its own page. Updates links to properly point to the new page.


narsaynorath (approved):

vivianyentran (approved):
Thanks for catching & fixing these!

PR #9431, opened 2024-03-13
fix(scrubbing): Mention default data scrubbing settings

iker-barriocanal:
Link to the default scrubbed filters from the `**` selector in the data scrubbing sources docs.

Resolves https://github.com/getsentry/sentry/issues/66932.

### DIFF
--- a/docs/product/data-management-settings/scrubbing/advanced-datascrubbing.mdx
+++ b/docs/product/data-management-settings/scrubbing/advanced-datascrubbing.mdx
@@ -75,7 +75,7 @@ Sentry does not know if a local variable that looks like a credit card number ac
 
 Selectors allow you to restrict rules to certain parts of the event. This is useful to unconditionally remove certain data by event attribute, and can also be used to conservatively test rules on real data. A few examples:
 
-- `**` to scrub everything
+- `**` to scrub [almost every field](/product/data-management-settings/scrubbing/server-side-scrubbing/event-pii-fields/) (other fields require specific selectors)
 - `$error.value` to scrub in the exception message
 - `$message` to scrub the event-level log message
 - `extra.'My Value'` to scrub the key `My Value` in "Additional Data"
--- a/docs/product/data-management-settings/scrubbing/advanced-datascrubbing.mdx
+++ b/docs/product/data-management-settings/scrubbing/advanced-datascrubbing.mdx
@@ -75,7 +75,7 @@ Sentry does not know if a local variable that looks like a credit card number ac
 
 Selectors allow you to restrict rules to certain parts of the event. This is useful to unconditionally remove certain data by event attribute, and can also be used to conservatively test rules on real data. A few examples:
 
-- `**` to scrub [all default event PII fields](/product/data-management-settings/scrubbing/server-side-scrubbing/event-pii-fields/) (other fields require specific selectors)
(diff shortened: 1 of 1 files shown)


docs/product/data-management-settings/scrubbing/advanced-datascrubbing.mdx
```diff
@@ -75,7 +75,7 @@ Sentry does not know if a local variable that looks like a credit card number ac
 
 Selectors allow you to restrict rules to certain parts of the event. This is useful to unconditionally remove certain data by event attribute, and can also be used to conservatively test rules on real data. A few examples:
 
-- `**` to scrub everything
+- `**` to scrub [almost every field](/product/data-management-settings/scrubbing/server-side-scrubbing/event-pii-fields/) (other fields require specific selectors)
```
olksdr, line 78: I would say 
```suggestion
- `**` to scrub [all currently supported fields](/product/data-management-settings/scrubbing/server-side-scrubbing/event-pii-fields/) (other fields require specific selectors)
```

olksdr: or maybe `all supported by default` fields?

realkosty: "all fields (that are) scrubbed by default" ?

realkosty: that's what the linked page says
Screenshot 2024-03-13 at 9 53 35 AM

vivianyentran: What about "all event PII fields"?

```suggestion
- `**` to scrub [all event PII fields](/product/data-management-settings/scrubbing/server-side-scrubbing/event-pii-fields/) (other fields require specific selectors)
```

vivianyentran: It might also be helpful to include a link to "other fields" if possible. I have no idea what those other fields are

iker-barriocanal: "all event PII fields" seems like other fields are not PII-scrubbable so I went with "all default event PII fields", let me know what you think.

We don't maintain a list of "other fields", the closest would be the event payloads in the [dev docs](https://develop.sentry.dev/sdk/event-payloads/) (with all the linked interfaces), but that's not friendly for the regular user. PII scrubbing is already a bit complex and requires some reading from the user, so I don't think adding another link with _a lot_ of content helps. Happy to add this link if somebody thinks it'd be helpful.

vivianyentran: I like "all default event PII fields" 👍 

I agree on not overloading the user with info. Instead of linking to other fields, are there any fields we can call out that users should be aware of? A somewhat common field they might expect to be scrubbed but isn't or an important field they might need to manually scrub? If not, that's ok and we can leave it as is.

vivianyentran (approved):

docs/product/data-management-settings/scrubbing/advanced-datascrubbing.mdx
iker-barriocanal: Added `span.description` as it has raised a couple of questions in the past, including this one.

Dav1dde (approved):

PR #9439, opened 2024-03-13
Fix sitemap paths

mjq-sentry:
@lizokm reported that only some of the pages in the sitemap.xml were being indexed properly. This was caused by an extra trailing `/index` in many of the paths. This is something that the main Page component already handles, but it was missed in the sitemap. Moved this fix up a level (into `getDocsFrontMatter()`) so that it didn't have to be duplicated into both the page and the sitemap, and so that future callers don't run into this same oversight.


lizokm (approved):
LGTM

PR #9460, opened 2024-03-15
feat(webvitals): Moves overall perf score explainer up and add min latency for inp

edwardgou-sentry:
Moves up the explainer for the overall performance score and adds note about minimum latency for INP candidates which links out to the google webvitals library
image


vivianyentran (approved):
Thanks for updating!

PR #9464, opened 2024-03-15
[Python] Moved distributed tracing from 'usage' to 'performance monitoring'

antonpirker:
This moves "Distributed Tracing" from `SENTRY FOR PYTHON > Usage` into `PERFORMANCE MONITORING > Instrumentation`.

This was a request from multiple people, including support because no one finds it under "Usage".

For testing the redirects (should redirect to `/platforms/python/performance/instrumentation/distributed-tracing/*`:
- https://sentry-docs-git-antonpirker-python-distributed-tracing.sentry.dev/platforms/python/usage/distributed-tracing/
- https://sentry-docs-git-antonpirker-python-distributed-tracing.sentry.dev/platforms/python/usage/distributed-tracing/custom-instrumentation/
- https://sentry-docs-git-antonpirker-python-distributed-tracing.sentry.dev/platforms/python/usage/distributed-tracing/dealing-with-cors-issues/
- https://sentry-docs-git-antonpirker-python-distributed-tracing.sentry.dev/platforms/python/usage/distributed-tracing/limiting-trace-propagation/


dcramer:
I agree with consolidating these (its hard to find, and its duplicated content - although not truly 1:1).

_HOWEVER_, we need every customer to successfully setup trace propagation, even if they don't send span waterfalls. What do we do to achieve that? Trace IDs need ever present and propagated for errors, replays, spans, etc.

antonpirker:
Internally this is called "tracing without performance" or TwP. So trace ids are present no matter if performance is enabled or not. This is implemented in most SDKs. (I know Go is missing, but will be added soon)

We already had some discussions on where to best put documentation for distributed tracing. As we only want to have those big sections in the sidebar for actual sentry products (performance, metrics, profiling) we agreed on to put it under "performance monitoring" right now to make it easier to find. But on a second thought, maybe it is better to put "Distributed Tracing" between "Data Management" and "Migration Guide". Wdyt @dcramer ?

dcramer:
> Internally this is called "tracing without performance" or TwP. So trace ids are present no matter if performance is enabled or not. This is implemented in most SDKs. (I know Go is missing, but will be added soon)

That wont enable trace solve for propagation though?

> We already had some discussions on where to best put documentation for distributed tracing. As we only want to have those big sections in the sidebar for actual sentry products (performance, metrics, profiling) we agreed on to put it under "performance monitoring" right now to make it easier to find. But on a second thought, maybe it is better to put "Distributed Tracing" between "Data Management" and "Migration Guide". Wdyt @dcramer ?

I'd probably just call the whole thing distributed tracing (or tracing) and ignore performance. Performance is a concept not a technology or even a product ultimately.  This will be even more clear as we move towards charging for spans, as you're not buying the perf product, you're simply paying for data consumption and we're enabling product surface based on that data.

PR #9468, opened 2024-03-16
Fix mistake in tabTitle

marshallwalker:
## Description of changes

Fixed a mistake with the yarn/npm switch on the webpack source-maps docs

## Legal Boilerplate


marshallwalker:
Just noticed this was already fixed in main

PR #9472, opened 2024-03-18
feat: Add RSS feed for changelog

HazAT:
Exposes `/changelog/feed.xml` as an url for all Changelogs as an RSS feed.


antonpirker (approved):
rss ftw!

PR #9473, opened 2024-03-18
[Python] Describe precedence of `release` and `environment` configuration

antonpirker:
Make it clear how the configuration of `release` and `environment` can be set and in what order the SDK tries the different ways to set those options.


vivianyentran (approved):
Thanks for these edits! They really improve the clarity!

PR #9551, opened 2024-03-25
Update image upload instructions

lizokm:
Change recommended way to add images when contributing to docs.

### DIFF
--- a/docs/contributing/pages/images.mdx
+++ b/docs/contributing/pages/images.mdx
@@ -6,8 +6,7 @@ noindex: true
 If you want to add images to your docs, you can add them in a `./img` folder in the same directory as your `.mdx` files.
 You don't have to follow a rigid folder structure as long as the relative path (starts with `./`) to the image is correct.
 
-If an image is used in multiple places, it's best to put it in `includes/common-imgs`. You can reference it in a local `md[x]` file
-and then `Include` it in the target `mdx` files.
+If an image is used in multiple places, it's best to put it in `includes/common-imgs/imgs` and reference it in a local [`include` file ](/contributing/approach/product-docs/write-index/#includes) to be used in the target `mdx` file.
 
 While this is not our recommended method, images can also be added to the `public` folder.
 
--- a/docs/contributing/pages/images.mdx
+++ b/docs/contributing/pages/images.mdx
@@ -32,7 +32,7 @@ The image `docs/common-imgs/imgs/tags.png` is referenced in `docs/common-imgs/ta
 ![Tags](./img/tags.png)
 ```
 
-which is then included in multiple files:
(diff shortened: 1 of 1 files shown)


docs/contributing/pages/images.mdx
```diff
@@ -33,7 +33,7 @@ The image `docs/common-imgs/imgs/tags.png` is referenced in `docs/common-imgs/ta
 ![Tags](./img/tags.png)
 ```
 
-which is then included in many other files:
+which is then included in multiple files:
```
vivianyentran, line 36:
```suggestion
The image can then be used in multiple files by referencing the `include` file with the following syntax:
```

docs/contributing/pages/images.mdx
```diff
@@ -33,7 +33,7 @@ The image `docs/common-imgs/imgs/tags.png` is referenced in `docs/common-imgs/ta
...
 
 ```markdown {filename:docs/platforms/ruby/common/enriching-events/tags/index.mdx}
 <Include name="common-imgs/tags" />
```
vivianyentran, line 39: Can't comment on the last line, but it needs to be updated to not reference `public` again.

docs/contributing/pages/images.mdx
```diff
@@ -3,13 +3,13 @@ title: Images
 noindex: true
 ---
 
-If you want to add images to your docs, you can add them next to your `.mdx` files, under `./img` by convention.
+If you want to add images to your docs, you can add them in a `./img` folder under your `.mdx` files.
 You don't have to follow a rigid folder structure as long as the relative path (starts with `./`) to the image is correct.
 
-If an image is used in multiple places, it's best to put it in `includes/common-imgs`, reference it from there in a local `md[x]` file
+If an image is used in multiple places, it's best to put it in `includes/common-imgs`. You can reference it in a local `md[x]` file
 and then `Include` it in the target `mdx` files.
```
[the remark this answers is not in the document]
vivianyentran: I'm calling it an `include` file (rather than using the verb `Include`) to be consistent with other references in the contributing docs (https://docs.sentry.io/contributing/approach/sdk-docs/common_content/ & https://docs.sentry.io/contributing/approach/product-docs/write-index/). We can change this if you want to, but just want to make sure we're consistent.

vivianyentran (changes requested):
Added some comments. I think we need to make these changes:

- I looked into the current repo structure and found that the image and file paths were wrong so I noted where these should be updated. 
- I thought the way include files were explained was a little confusing, so I tried to reword it a bit.
- I think we need to add back in instructions for how to use the public folder. As a reader, it's confusing that this is mentioned without any context.

vivianyentran (approved):

PR #9568, opened 2024-03-27
mention value to track react component

bruno-garcia:
This documentation so far isn't explaining why would I want to "track components". What does that mean, etc
So I aim to add a "before an after" to help clarify that

### DIFF
--- a/docs/platforms/javascript/guides/react/features/component-tracking.mdx
+++ b/docs/platforms/javascript/guides/react/features/component-tracking.mdx
@@ -9,7 +9,7 @@ Sentry helps you track your React components and unlock additional insights in y
 So instead of looking at selectors like this:
 
 ```html
-div.row > div.col-sm-8.col-md-12.col-lg-8.col-xl-9 > div#main > h3#ignore-exceptions > a[aria-label="common options permalink"]
+button.en302zp1.app-191aavw.e16hd6vm2[role="button"][data-test-id="common-options"]
 ```
 
 You can also see exactly which React component was used, like:
--- a/docs/platforms/javascript/guides/react/features/component-tracking.mdx
+++ b/docs/platforms/javascript/guides/react/features/component-tracking.mdx
@@ -6,6 +6,18 @@ description: "Learn how Sentry's React SDK allows you to monitor the rendering p
 
 Sentry helps you track your React components and unlock additional insights in your application. You can set it up to track React component names as well as monitor the performance of your React components with component tracking.
 
+So instead of looking at selectors like this:
+
+```html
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/guides/react/features/component-tracking.mdx
```diff
@@ -6,6 +6,18 @@ description: "Learn how Sentry's React SDK allows you to monitor the rendering p
...
+So instead of looking at selectors like this:
+
+```html
+div.row > div.col-sm-8.col-md-12.col-lg-8.col-xl-9 > div#main > h3#ignore-exceptions > a[aria-label="common options permalink"]
+```
+
+You can also see exactly which React component was used, like:
+
+```html
+CommonOptions
```
c298lee, line 18: In Breadcrumbs, we still want to show the entire component "tree"

docs/platforms/javascript/guides/react/features/component-tracking.mdx
```diff
@@ -6,6 +6,18 @@ description: "Learn how Sentry's React SDK allows you to monitor the rendering p
 
 Sentry helps you track your React components and unlock additional insights in your application. You can set it up to track React component names as well as monitor the performance of your React components with component tracking.
 
+So instead of looking at selectors like this:
+
+```html
+div.row > div.col-sm-8.col-md-12.col-lg-8.col-xl-9 > div#main > h3#ignore-exceptions > a[aria-label="common options permalink"]
```
michellewzhang, line 12: might be more relevant to show a selector like this

`button.en302zp1.app-191aavw.e16hd6vm2[role="button"][data-test-id="common-options"]`

bruno-garcia: Thanks!

bruno-garcia, line 12:
```suggestion
button.en302zp1.app-191aavw.e16hd6vm2[role="button"][data-test-id="common-options"]
```

docs/platforms/javascript/guides/react/features/component-tracking.mdx
bruno-garcia: What would that look like? Could you add an example?

c298lee: oh this is good now! 
before the selector example had the entire html tree so it would make more sense if using React component example still kept the tree, but since we're only showing the selector for the before, it matches up!

bruno-garcia: Makes sense! thanks

PR #9594, opened 2024-04-01
ref(replay): update min python version

michellewzhang:
fixes https://github.com/getsentry/sentry/issues/67973

### DIFF
--- a/docs/product/session-replay/getting-started.mdx
+++ b/docs/product/session-replay/getting-started.mdx
@@ -56,7 +56,7 @@ Sentry.init({
 Currently, replays for backend errors is supported for the following SDK versions or newer:
 
 - JavaScript (or related framework) [7.51.0](https://github.com/getsentry/sentry-javascript/releases/tag/7.51.0)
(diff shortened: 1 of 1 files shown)


vivianyentran (approved):

ryan953:
Lets check the changelog/commits/release notes and make sure the version number is correct. If it's wrong for the python SDK has a wrong one then all the others could be wrong too

ryan953:
@bruno-garcia did you check if any of the other SDKs have an incorrect version listed as well?

Usually I go to the docs to verify minimum versions, if this doc was wrong, how else did you check that 1.25 was the correct version, the [release notes](https://github.com/getsentry/sentry-python/releases/tag/1.25.0) mention 5 fixes and improvements, none of them are related to tracing.

bruno-garcia:
@ryan953 I got the versions from the SDK engineers. I did see the release notes, and 1.20 actually looks right (again, I got it form the SDK engineers and has a note on `/envelope`).
I think we should be good with these min versions and not worth going double checking everything

bruno-garcia:
all those versions are ancient and folks should update so suggesting a newer version is OK

PR #9604, opened 2024-04-03
Distributed Tracing docs move

stephanie-anderson:
- moves the contents from /usage/distributed-tracing to /distributed-tracing
- adds new redirect rules, and updates existing rules to avoid multiple redirects


HazAT (approved):
💯 ship it

PR #9626, opened 2024-04-08
fix(browser-extensions): Exclude global state integrations in browser extensions

s1gr1d:
## Description of changes

Some default integrations which use the global state should not be included when using Sentry in a browser extension.

### DIFF
--- a/docs/platforms/javascript/common/best-practices/browser-extensions.mdx
+++ b/docs/platforms/javascript/common/best-practices/browser-extensions.mdx
@@ -15,7 +15,7 @@ keywords:
 When setting up Sentry in a shared environment where multiple Sentry instances may run, for example, in a browser extension or similar, you should **not use `Sentry.init()`**, as this will pollute the global state. If your browser extension uses `Sentry.init()`, and the website the extension is running on also uses Sentry, the extension may send events to the website's Sentry project, or vice versa.
 
 For such scenarios, you have to set up a client manually as seen in the example below.
-In addition, you should also avoid adding any integrations that use global state, like `Breadcrumbs` or `TryCatch`. Furthermore, some default integrations that use the global state have to be filtered as in the example below.
+In addition, you should also avoid adding any integrations that use global state, like `Breadcrumbs` or `GlobalHandlers`. Furthermore, some default integrations that use the global state have to be filtered as in the example below.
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/common/best-practices/browser-extensions.mdx
```diff
@@ -15,7 +15,7 @@ keywords:
 When setting up Sentry in a shared environment where multiple Sentry instances may run, for example, in a browser extension or similar, you should **not use `Sentry.init()`**, as this will pollute the global state. If your browser extension uses `Sentry.init()`, and the website the extension is running on also uses Sentry, the extension may send events to the website's Sentry project, or vice versa.
 
 For such scenarios, you have to set up a client manually as seen in the example below.
-In addition, you should also avoid adding any integrations that use global state, like `Breadcrumbs` or `TryCatch`.
+In addition, you should also avoid adding any integrations that use global state, like `Breadcrumbs` or `TryCatch`. Furthermore, some default integrations that use the global state have to be filtered as in the example below.
```
Lms24, line 18: l: A bit tricky as long as v8 isn't stable yet: `TryCatch` was renamed to `BrowserApiErrors`. We could mention another one that didn't change in the text and add both versions to the code snippet below, WDYT?

```suggestion
In addition, you should also avoid adding any integrations that use global state, like `Breadcrumbs` or `GlobalHandlers`. Furthermore, some default integrations that use the global state have to be filtered as in the example below.
```

docs/platforms/javascript/common/best-practices/browser-extensions.mdx
```diff
@@ -29,11 +29,16 @@ import {
   Scope,
 } from "@sentry/browser";
 
+// filter integrations that use the global variable
+const integrations = defaultIntegrations.filter(defaultIntegration => {
+  return !['BrowserApiErrors', 'Breadcrumbs', 'GlobalHandlers'].includes(defaultIntegration.name);
```
Lms24, line 34:
```suggestion
  return !['BrowserApiErrors', 'TryCatch', 'Breadcrumbs', 'GlobalHandlers'].includes(defaultIntegration.name);
```

Lms24 (approved):
Thanks for the quick fix! Just had a comment RE v7/v8 discrepancies but otherwise 👍

PR #9628, opened 2024-04-08
deps: Bump SDK to v8 alpha

lforst:
Bumps the `@sentry/nextjs` SDK to the v8 alpha. It may produce a lot of garbage performance data because we haven't built elaborate filtering logic yet.


app/changelog/page.tsx
```diff
@@ -27,8 +26,5 @@ export function generateMetadata(): Metadata {
     alternates: {
       canonical: `https://sentry.io/changelog/`,
     },
-    other: {
-      'sentry-trace': `${Sentry.getActiveSpan()?.toTraceparent()}`,
-    },
```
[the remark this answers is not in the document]
lforst: I generally don't like that we have this because it is bad dogfooding. Our users don't have this either.

src/instrumentation.ts
```diff
@@ -0,0 +1,22 @@
+import * as Sentry from '@sentry/nextjs';
+
+export function register() {
+  if (process.env.NEXT_RUNTIME === 'nodejs') {
+    Sentry.init({
+      dsn: 'https://ad63ba38287245f2803dc220be959636@o1.ingest.sentry.io/1267915',
```
HazAT, line 6:
```suggestion
      dsn: process.env.SENTRY_DSN,
```

src/instrumentation.ts
```diff
@@ -0,0 +1,22 @@
...
+      tracesSampleRate: 1,
+      debug: false,
+      environment: process.env.NODE_ENV === 'development' ? 'development' : undefined,
+      spotlight: process.env.NODE_ENV === 'development',
+    });
+  }
+
+  if (process.env.NEXT_RUNTIME === 'edge') {
+    Sentry.init({
+      dsn: 'https://ad63ba38287245f2803dc220be959636@o1.ingest.sentry.io/1267915',
```
HazAT, line 16:
```suggestion
      dsn: process.env.SENTRY_DSN,
```

HazAT (changes requested):

PR #9653, opened 2024-04-10
feat(js): Remove `Shutdown and Draining` page in Browser JS

mydea:
IMHO this is not that important in browser JS and does not warrant having its own docs page.


Lms24 (commented):
I don't see a particular usage for this myself either but maybe some people deactivate Sentry in a certain condition 🤔 It probably doesn't warrant its own page but I don't think we should delete the content completely for Browser JS. 

No super strong opinion though, if others think it's fine to delete I'm not gonna block it 😅

stephanie-anderson (commented):
Hey Francesco! Could you please add a redirect for that page? (this redirect is also needed for guides, since the file was in the /common folder)

mydea:
> I don't see a particular usage for this myself either but maybe some people deactivate Sentry in a certain condition 🤔 It probably doesn't warrant its own page but I don't think we should delete the content completely for Browser JS.
> 
> No super strong opinion though, if others think it's fine to delete I'm not gonna block it 😅

Hmm, I get that. I think what is actually missing overall is some general "API" docs/page - e.g. what methods are there that can be used etc. We don't really have any docs for that, it's all spread out everywhere etc. We can leave this page in for now, no strong feelings, just found it a bit weird that we have a dedicated page just for this 😅

PR #9655, opened 2024-04-10
fix(user-feedback): missing whitespace

buenaflor:
Missing whitespaces:
- https://docs.sentry.io/platforms/python/user-feedback/#integration
- https://docs.sentry.io/platforms/dotnet/user-feedback/#integration


vivianyentran (approved):
thanks for fixing!

PR #9660, opened 2024-04-10
feat(browser): Move transports config to best practices

mydea:
This page only contains info about setting up offline support, which IMHO can just be a best practice instead.


AbhiPrasad (approved):
IMO we should add notes about writing a custom transport here and leave the page here.

mydea:
> IMO we should add notes about writing a custom transport here and leave the page here.

I added back the transports page and added info on creating your own transport - give it a look :)

AbhiPrasad (approved):

PR #9666, opened 2024-04-10
Update deprecated option names and attachStacktrace value

Fwang36:
Update options for performance tracing and watchdog termination to reflect their new names in the v5 versions of React Native.

attachStacktrace is also On by default in React Native, whereas the docs currently listed it as Off.


vivianyentran (approved):

vivianyentran:
@Fwang36 Is this ready to publish?

PR #9669, opened 2024-04-10
fix(code-mappings): Add Go to list of languages that have support for Automatic Code mappings

MichaelSun48:
Automatic Code Mapping support for Go has just been GA'd today! This PR updates the docs to reflect that


vivianyentran (approved):
Thanks for updating!

PR #9670, opened 2024-04-10
chore(hc): EU data storage location has arrived!

mrduncan:


lizokm (approved):
LGTM! Thanks for making the update and WOO HOO

PR #9675, opened 2024-04-11
PHP doc tweaks

stayallive:
/cc @cleptric I wanted to document `trace` and in the process reworked a section. Lemme know what you think!


cleptric (approved):
Great stuff!

PR #9686, opened 2024-04-12
fix(releases): Fix broken links

ceorourke:
I was reading this page https://docs.sentry.io/product/releases/health/ and noticed the links to the crashes and sessions sections were broken because they used the singular (crash and session) but the anchor tags are plural (crashes and sessions).


vivianyentran (approved):
Thanks for fixing!

PR #9693, opened 2024-04-12
remove compatibility page

bruno-garcia:
On 4.0.0 we dropped support to very old and long deprecated versions: https://github.com/getsentry/sentry-dotnet/discussions/2776

This doc is hard to keep up-to-date and the source of truth of compatibility is NuGet itself:

See: https://www.nuget.org/packages/Sentry#dependencies-body-tab

---
For example, for version 4.3.0:

image
image


Jeffreyhung (approved):

bruno-garcia:
tested the redirects on vercel, worked out

PR #9710, opened 2024-04-15
Merge node with javascript platform

stephanie-anderson:
Things I've considered:

### General
- [x] moved all contents from `/platform/node` to `/platform/javascript/guides/node`
- [x] moved all contents from `/platform/node/guides` to `/platform/javascript/guides`
- [x] deleted all `node` specific files if they were identical with the `javascript` ones (or could be easily adapted to fit) 

### Previous Node.js Guides
All previous guides of Node.js will now be guides of the JavaScript platform. The previous inheritance for a Node.js guide's content was:
- `Any Node.js guide` -> `Node.js` ->  `JavaScript` (because the `node` platform defined `javascript` as its `fallbackPlatform` 

Now it is:
- `Node.js` (which is only a guide now) ->  `JavaScript` and also
- `Any Node.js guide` ->  `JavaScript`

### Platform Includes
Due to the previously existing `fallbackPlatform` setting of the `node` platform, it wasn't necessary for us to create a dedicated node.mdx file in the respective subfolder of `platform-includes`, e.g. `platform-includes/capture-message/node.mdx`. If `node` wasn't any different from browser JavaScript, we wouldn't have to create that file. This still holds true of course. But this is what I changed:

- [x] all include files for `node` in `/platform-includes` (`node.mdx`) have been renamed to `javascript.node.mdx`
- [x] I deleted all `node.mdx` (or `javascript.node.mdx`) files, if they weren't different from `javascript.mdx`
- [x] In order for Node guides (e.g. javascript.express.mdx) to inherit from Node (`javascript.node.mdx`) I simply created symlinks. This was easier (for me) than adding another layer of logic to the docs, so that one guide could inherit from another guide. Was just too complicated. Instead, the symlink will make sure we only need to maintain one file holding the content, but that this content is also rendered for all node-based guides as well
- [x] There were no platform includes for any node guides, otherwise they would have also been renamed

### New Component: PlatformCategorySection

I created a new component, because the only way to filter certain content before was using the PlatformSection component. 

```
<PlatformSection supported={["javascript.electron", "javascript.node", "javascript.aws-lambda", "javascript.azure-functions", "javascript.connect", "javascript.express", "javascript.gcp-functions", "javascript.koa"]}>

Special content

</PlatformSection>
```

It think maintaining this list will be painful, so I created a new component, that will render a section of content, based on the category a guide or platform has set in their frontmatter. The above could now look something like this:

```
<PlatformCategorySection supported={["server", "desktop"]}>

Special content

</PlatformCategorySection>
```
I also added the category information, if it wasn't present in any of the guides. For Express, this now looks like this:

```
---
title: Express
description: "Learn about using Sentry with Express."
sdk: sentry.javascript.node
categories:
  - server
---
```

### Add supported information to OG JavaScript pages
Some of the pages in `docs/platforms/javascript` are not meant for server runtimes/SDKs. Pages like these have now some more items on their `notSupported` list:

```
---
title: Best Practices
description: "Learn how to set up Sentry in specific scenarios with those best practice guides"
sidebar_order: 3
notSupported:
  - javascript.node
  - javascript.aws-lambda
  - javascript.azure-functions
  - javascript.connect
  - javascript.express
  - javascript.gcp-functions
  - javascript.koa
---
```

This is another use case, where filtering supported/notSupported based on categories a guide/platform has, could come in really handy.

### Platform Icons
- [x] I created a new release of platform icons, so that the logo of Node and all frameworks would be resolved correctly
- [x] Updated the dependency in package.json to use that newer version

### URLs & Redirects
- [x] A redirect for /platforms/node/guides/* is added (`src/middleware.ts`, `vercel.json`)
- [x] A redirect for /platforms/node/* is added (`src/middleware.ts`, `vercel.json`)
- [x] Old redirects that point to /platforms/node/guides/XY are updated (`src/middleware.ts`, `vercel.json`)
- [x] Old redirects that point to /platforms/node are updated (`src/middleware.ts`, `vercel.json`)
- [x] URLs across all docs have been updated to avoid as many redirects for users as possible (`docs`)


mydea (commented):
So the only downside I see here is that it is somewhat easily possible to accidentally change node docs if you want to change docs for a node-guide. E.g. if I want to overwrite page X for express that exists in node, the file exists in the express guide, but it is a symlink - but this is hard to notice, so it may easily happen that you just modify that file, and without knowing also mutate the docs for all other node guides 🤔

PR #9721, opened 2024-04-16
feat(replay): Update React Component Names

c298lee:
Splits component name tracking into two pages: React Component Names and Component Performance Monitoring.
Updates language to use capture instead of track when referring to react component names.
Adds a section about react component names in replay.


michellewzhang (approved):
gif

docs/product/reference/search/searchable-properties/session-replay.mdx
```diff
@@ -48,7 +48,7 @@ The `class` of an element that was clicked. No leading `.` is necessary. For exa
 
 The name of the frontend component that was clicked. For example, `MyAwesomeComponent` would match the React component `<MyAwesomeComponent>`.
 
-Note: This property requires that your project have [React component name tracking configured](/platforms/javascript/guides/react/features/component-tracking/#component-name-tracking).
```
bruno-garcia, line 51: do we have redirects in place?

bruno-garcia (changes requested):

0Calories (approved):
Nice!

docs/product/reference/search/searchable-properties/session-replay.mdx
c298lee: what do you mean by redirects? How do you set that up?

bruno-garcia (commented):
Please refer to this doc for redirection: https://docs.sentry.io/contributing/pages/redirects/

Not that when you delete page, it will break search engine indexing:

image

So on the PR that deletes it, also add a redirection rule to the place that makes mroe sense

docs/platforms/javascript/guides/react/features/component-names.mdx
```diff
@@ -0,0 +1,97 @@
+---
+title: React Component Names
+excerpt: ""
+description: "Learn how Sentry's React SDK allows you to monitor your components."
+---
+
+Sentry helps you capture your React components and unlock additional insights in your application. You can set it up to use React component names instead of selectors.
```
michellewzhang, line 7:
```suggestion
Sentry helps you capture your React components and unlock additional insights in your application. You can set up your Sentry configuration to use React component names instead of selectors, whenever they are available.
```

docs/platforms/javascript/guides/react/features/component-names.mdx
```diff
@@ -0,0 +1,97 @@
...
+
+You can also see exactly which React component was used, like:
+
+```html
+CommonOptions
+```
+
+## Component Name Capturing
+
+You can capture the names of React components in your application via a [Babel plugin](https://www.npmjs.com/package/@sentry/babel-plugin-component-annotate), which can unlock powerful workflows and decrease ambiguity. For example, you can search for replays where a specific React component was clicked on. You can also see the name of the component that was clicked on in breadcrumbs and spans on Sentry's **Performance** page. This significantly decreases the amount of ambiguity involved in figuring out which element was interacted with as compared to trying to figure it out by looking at the element's CSS selector alone (which becomes even more ambiguous after the minification process).
```
michellewzhang, line 23:
```suggestion
You can capture the names of React components in your application via a [Babel plugin](https://www.npmjs.com/package/@sentry/babel-plugin-component-annotate), which can unlock powerful workflows and decrease ambiguity when using various Sentry features. For example, you can search for replays where a specific React component was clicked on. You can also see the name of the component that was clicked on in breadcrumbs and spans on Sentry's **Performance** page. This significantly decreases the amount of ambiguity involved in figuring out which element was interacted with, when compared to looking at the element's CSS selector alone (which becomes even more ambiguous after the minification process).
```

docs/product/reference/search/searchable-properties/session-replay.mdx
bruno-garcia: if this is only under `guides/react` you don't need to worry about the regex group to catch all, like here: https://github.com/getsentry/sentry-docs/blob/5467a3ef3ccd4daa9b63f39051525ddbf86aba84/vercel.json#L546-L552

You can just have `/guides/react/track-..` etc

Jeffreyhung (approved):

docs/product/reference/search/searchable-properties/session-replay.mdx
c298lee: I added the redirect!

bruno-garcia (approved):

bruno-garcia:
Thanks Catherine!

PR #9742, opened 2024-04-18
Remove EA note from Reprocessing docs

Swatinem:
We have decided to enable this feature for all customers, so lets remove the early adopter note.

Turning the feature on fully happens in https://github.com/getsentry/sentry/pull/68412.


lizokm (approved):
Thanks for updating!!

PR #9749, opened 2024-04-19
try styling the words in the subtitle

armcknight:
Just wanted to try something.


lizokm:
Hey @armcknight, thanks for caring about docs and for the suggestion. We're in the process of updating our landing page with a new design. So let's hold off on making any changes until that's done :)

PR #9759, opened 2024-04-19
Direct users to `--help` on CLI DIF page

### LINKED ISSUE #9758: Direct users towards `--help` for CLI `debug-files` options
The [Sentry CLI Debug Information Files page](https://docs.sentry.io/product/cli/dif/) contains detailed documentation of several options that can be provided to the subcommands. Most of these options are irrelevant to most users, and all of the options are already documented directly in the CLI.

Users can access documentation for any CLI command by passing `--help`. This `--help` should be the single source of truth on which arguments are available and what the argument does. By also documenting these options in the docs, we increase maintenance burden and create the possibility of the docs pages diverging from the `--help` output. 

For example, #6558 reports that documentation for some `debug-files upload` options, which are displayed in `--help`, are missing from the docs.

**Rather than implement #6558's request that we add these options to the docs page, we have decided to remove documentation for all arguments from the docs page. Instead, we will direct users to the `--help` option to see all available options.**

If we find any other pages where we document individual options of a CLI command, we should follow the same principle we are applying here, and replace the documentation with directions to use `--help`. The only exception is an argument we expect most users will need to be aware of – even in this case, we should only briefly summarize the option in the docs, and direct users to `--help` for detailed information.

szokeasaurusrex:
## Pre-merge checklist

## Description of changes

The CLI's `--help` text should be the single source of truth regarding available options for commands. Closes #9758.


Swatinem (approved):

vivianyentran (approved):
LGTM. This will make maintaining the docs for this much easier too!

PR #9764, opened 2024-04-19
Add viewed_by_id to searchable replay props

aliu3ntry:


docs/product/reference/search/searchable-properties/session-replay.mdx
```diff
@@ -280,4 +280,10 @@ Username, which is typically a better label than the `user.id`.
 
 - **Type:** string
 
+### `viewed_by_id`
+
+Sentry user ids belonging to the users who have seen this Replay. Searchable as a scalar or bracketed list.
```
michellewzhang, line 285:
```suggestion
Sentry user IDs who have seen this Replay. Searchable as a scalar or bracketed list.
```

michellewzhang (approved):

docs/product/reference/search/searchable-properties/session-replay.mdx
```diff
@@ -280,4 +280,10 @@ Username, which is typically a better label than the `user.id`.
 
 - **Type:** string
 
+### `viewed_by_id`
+
+Sentry user IDs who have seen this Replay. Searchable as a scalar or bracketed list.
+
+- **Type:** (unsigned) integer
```
ryan953, line 287: None of the other fields mention scalar or list, and they use `number` or UUID as types. We don't need to be so specific/different to describe this one.

What we could/should describe, here or somewhere, is how to get an id for a given user.

```suggestion
Sentry user IDs who have seen this Replay.
- **Type:** number
```

aliu3ntry: Got it. How would an external user lookup ids?

docs/product/reference/search/searchable-properties/session-replay.mdx
```diff
@@ -280,4 +280,9 @@ Username, which is typically a better label than the `user.id`.
 
 - **Type:** string
 
+### `viewed_by_id`
+
+Sentry user IDs who have seen this Replay.
+- **Type:** number
```
ryan953:
```suggestion
Sentry user IDs who have seen this Replay.

- **Type:** number
```

oops, i messed that up the first time.

docs/product/reference/search/searchable-properties/session-replay.mdx
ryan953: Looks like it id is also in the url when you click to view a user via `/settings/members/`, but it's not printed to the page.

aliu3ntry: Ok.. would this be an acceptable way to get ids for now? I guess we could also make a PR to display user id on the members page

vivianyentran (approved):

aliu3ntry:
After https://github.com/getsentry/sentry/pull/69370 I've decided to document `seen_by_id` as the preferred prop, since it corresponds to the UI "Seen By" banner. I mention `viewed_by_id` as a valid alias (corresponds to API endpoints)

aliu3ntry:
trigger:revert

aliu3ntry:
See comment on https://github.com/getsentry/sentry/pull/69364

PR #9775, opened 2024-04-23
chore(grammar): change "the" to "your" in suspect commits page

MichaelSun48:
super small nit clarifying that we're talking about the reader's Sentry org and not _the_ Sentry org


vivianyentran (approved):
Good fix!

PR #9783, opened 2024-04-23
docs(js-profiling): Re-wording comment about startSpan

s1gr1d:
Changed the comment as wrapping in `startSpan` is not required.

### DIFF
--- a/docs/platforms/javascript/guides/aws-lambda/profiling/index.mdx
+++ b/docs/platforms/javascript/guides/aws-lambda/profiling/index.mdx
@@ -49,14 +49,15 @@ Sentry.init({
   profilesSampleRate: 1.0,
 });
 
+// Profiling happens automatically after setting it up with `Sentry.init()`.
+// However, specific elements can be manually profiled with `startSpan`.
 Sentry.startSpan(
   {
     op: "rootSpan",
     name: "My root span",
   },
   () => {
-    // Any code executed inside this callback
-    // will now be automatically profiled.
+    // Any code in this callback will be profiled.
   }
 );
 ```
--- a/docs/platforms/javascript/guides/azure-functions/profiling/index.mdx
+++ b/docs/platforms/javascript/guides/azure-functions/profiling/index.mdx
@@ -49,14 +49,15 @@ Sentry.init({
   profilesSampleRate: 1.0,
 });
 
+// Profiling happens automatically after setting it up with `Sentry.init()`.
+// However, specific elements can be manually profiled with `startSpan`.
 Sentry.startSpan(
   {
     op: "rootSpan",
     name: "My root span",
   },
   () => {
-    // Any code executed inside this callback
-    // will now be automatically profiled.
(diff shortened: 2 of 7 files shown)


docs/platforms/javascript/guides/aws-lambda/profiling/index.mdx
```diff
@@ -49,14 +49,15 @@ Sentry.init({
   profilesSampleRate: 1.0,
 });
 
+// Profiling happens automatically after setting it up with `Sentry.init()`.
+// However, specific elements can be manually profiled with `startSpan`.
```
mydea, line 53:
```suggestion
// Profiling happens automatically after setting it up with `Sentry.init()`.
// Any spans that are captured in your application will have profiling data attached to them
```

maybe like this? 🤔

s1gr1d: I wanted to explain the `startSpan` part a little bit more. What do you think of this addition?

```md
// Profiling happens automatically after setting it up with `Sentry.init()`. 
// Any spans that are captured in your application will have profiling data attached to them.
// Spans can also be manually captured with `startSpan` like below.
```

vivianyentran: I like that addition! How's this?

```suggestion
// Profiling happens automatically after setting it up with `Sentry.init()`. 
// All spans captured in your application will have profiling data attached to them.
// You can also manually capture spans with `startSpan`, as shown below:
```

docs/platforms/javascript/guides/aws-lambda/profiling/index.mdx
```diff
@@ -49,14 +49,15 @@ Sentry.init({
   profilesSampleRate: 1.0,
 });
 
+// Profiling happens automatically after setting it up with `Sentry.init()`.
```
AbhiPrasad, line 52: I think this is a little confusing for me because profiling requires the existence of transactions/spans, so it only starts automatically if `Sentry.init` creates transactions via automatic instrumentation. Could we make that more clear?

docs/platforms/javascript/guides/aws-lambda/profiling/index.mdx
mydea: sounds very good to me!

docs/platforms/javascript/guides/aws-lambda/profiling/index.mdx
s1gr1d: Do you have a suggestion of how to write this a little bit more understandable?

AbhiPrasad: I think we can keep this line if we use the suggestion from @vivianyentran in https://github.com/getsentry/sentry-docs/pull/9783#discussion_r1576628367 because of the `All spans captured in your application will have profiling data attached to them` line!

mydea (approved):

PR #9787, opened 2024-04-23
feat(native): advanced usage section to explain OS error handling

supervacuus:
This PR introduces a new section, "Advanced usage," which allows us to go into much more detail when describing how operating systems behave and how this affects users of the Native SDK.

This section should house topics that our users repeatedly encounter, which we cannot hide behind a nice facade and which runtime developers do not abstract away from our responsibility (as is the case for most other SDKs).

While this section will soon see additions like

- [WER module usage ](https://github.com/getsentry/sentry-native/issues/875#issuecomment-1689564967)
- Debugging & resolving `UnhandledExceptionFilter` overwrites
- [Stackoverflow handling on Windows](https://github.com/getsentry/sentry-native/issues/977)
- [Trade-offs in determining which backend to use](https://github.com/getsentry/sentry-native/issues/908#issuecomment-1830031091)

The current driver for the section and content of this PR is the use of POSIX signal handling on Linux and Android and the issues users need to be aware of: https://github.com/getsentry/sentry-native/issues/960.

@kahest, can I ask you for a first review of whether this level of detail makes sense? It is really hard to reduce this further without linking every other word. I tried to balance this and removed a lot of detail already. I adapted it also to the future content in expected tone and level of detail.

### DIFF
--- a/docs/platforms/native/advanced-usage/index.mdx
+++ b/docs/platforms/native/advanced-usage/index.mdx
@@ -3,12 +3,10 @@ title: Advanced Usage
 description: "Learn how the Native SDK backends handle errors on their supported platforms and how that affects your usage."
 sidebar_order: 6000
 ---
-This section details recurring topics where the knowledge beyond what the SDK boundary can abstract away from its users no longer holds.
+This section discusses recurring topics that go beyond what the SDK can abstract away from users. 
 
-These will include operating system specifics regarding error handling or how some platforms deviate from established standards.
+While not a complete reference, this section covers topics like how different operating systems handle errors and where some platforms don't follow usual standards. Here, we also offer solutions to help users deal with specific platform quirks. 
 
-This section is not a conclusive reference documentation but will cover solutions that a user should understand to adapt to particular platform behavior.
-
-While we try to keep the content light, due to the in-depth nature of the topics following the text might require more effort than the other parts of the documentation.
+Due to the in-depth and nuanced nature of these topics, users may need a more advanced level of expertise. 
 
 <PageGrid />
--- a/docs/platforms/native/advanced-usage/index.mdx
+++ b/docs/platforms/native/advanced-usage/index.mdx
@@ -10,3 +10,5 @@ These will include operating system specifics regarding error handling or how so
 This section is not a conclusive reference documentation but will cover solutions that a user should understand to adapt to particular platform behavior.
 
 While we try to keep the content light, due to the in-depth nature of the topics following the text might require more effort than the other parts of the documentation.
+
+<PageGrid />
--- a/docs/platforms/native/advanced-usage/index.mdx
+++ b/docs/platforms/native/advanced-usage/index.mdx
@@ -1,0 +1,12 @@
+---
+title: Advanced Usage
+description: "Learn how the Native SDK backends handle errors on their supported platforms and how that affects your usage."
+sidebar_order: 6000
+---
+This section details recurring topics where the knowledge beyond what the SDK boundary can abstract away from its users no longer holds.
+
+These will include operating system specifics regarding error handling or how some platforms deviate from established standards.
+
+This section is not a conclusive reference documentation but will cover solutions that a user should understand to adapt to particular platform behavior.
+
+While we try to keep the content light, due to the in-depth nature of the topics following the text might require more effort than the other parts of the documentation.
--- a/docs/platforms/native/advanced-usage/signal-handling/index.mdx
+++ b/docs/platforms/native/advanced-usage/signal-handling/index.mdx
@@ -14,7 +14,7 @@ The topic of signal handling is vast, and this documentation can only guide you 
 * We recommend chapters 20-22 of Michael Kerrisk's [The Linux Programming Interface](https://man7.org/tlpi/) if you want a book-length treatment.
  
 
-## Signals of interest
+## Signals of Interest
 
 ## Signal Handling Conflicts
 
--- a/docs/platforms/native/advanced-usage/signal-handling/index.mdx
+++ b/docs/platforms/native/advanced-usage/signal-handling/index.mdx
@@ -31,7 +31,7 @@ In the context of the Native SDK, we are only interested in signals that will re
 * `SIGABRT`: Raised when `abort()` was called. This is the only signal handled on Windows (and only by the `crashpad` backend) because calling `abort()` bypasses Structured Exception Handling ([`SEH`](https://learn.microsoft.com/en-us/windows/win32/debug/structured-exception-handling)).
 * `SIGFPE`: An erroneous arithmetic operation occurred in the CPU.
 * `SIGILL`: The process attempted to execute an illegal (often privileged) processor instruction.
-* `SIGTRAP`: This is typically used to inform debuggers when execution hits a breakpoint instruction (`int 3`, `brk`, etc.), which instrumentation tools and sanitizers sometimes also issue.
+* `SIGTRAP`: This is typically used to inform debuggers when execution hits a breakpoint instruction (`int 3`, `brk`, etc.). Instrumentation tools and sanitizers also sometimes issue breakpoints.
 
 If you use the `crashpad` backend, in addition to the above, on Linux, you will also get crash reports for:
 
--- a/docs/platforms/native/advanced-usage/signal-handling/index.mdx
+++ b/docs/platforms/native/advanced-usage/signal-handling/index.mdx
@@ -10,7 +10,6 @@ A (POSIX) signal is a way for an operating system kernel to inform an applicatio
 The topic of signal handling is vast, and this documentation can only guide you in the context of the Native SDK. If you are interested in more general documentation on the topic, we can recommend the following:
 
 * The [main Wikipedia article](https://en.wikipedia.org/wiki/Signal_(IPC)) provides an overview and links to many signal details.
-* [The signal handling chapter](https://beej.us/guide/bgc/html/split/signal-handling.html) of "Beej's Guide to C Programming" provides a concise but practical introduction.
 * The [glibc-docs](https://www.gnu.org/software/libc/manual/html_node/Signal-Handling.html) and [the man pages](https://man7.org/linux/man-pages/man7/signal.7.html) offer reference details for the signal vocabulary.
 * We recommend chapters 20-22 of Michael Kerrisk's [The Linux Programming Interface](https://man7.org/tlpi/) if you want a book-length treatment.
  
--- a/docs/platforms/native/advanced-usage/signal-handling/index.mdx
+++ b/docs/platforms/native/advanced-usage/signal-handling/index.mdx
@@ -7,7 +7,7 @@ The Native SDK uses signal handling on all supported platforms. On Linux (and An
 
 A (POSIX) signal is a way for an operating system kernel to inform an application about an exceptional event and provide a mechanism for the application to react to that event. The operating system delivers this information by interrupting the current execution of the affected  application with a "handler" function that the application registered before.
 
-The topic of signal handling is vast, and this documentation can only guide you in the context of the Native SDK. If you are interested in more general documentation on the topic, we can recommend the following:
+This document provides guidance on signal handling in the Native SDK. If you'd like to learn more about signal handling in general, we recommend the following resources:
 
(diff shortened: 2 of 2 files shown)


kahest:
@supervacuus thank you, this is great! IMO the LoD is perfect 💯 
cc @lizokm  @stephanie-anderson this is the ["Advanced Usage"](https://sentry-docs-git-fork-supervacuus-feat-advancedusagesecti-3458ae.sentry.dev/platforms/native/advanced-usage/) section and the first "article" that I kinda announced a while ago - wdyt of the menu structure, title, etc?

docs/platforms/native/advanced-usage/signal-handling/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+* `SIGXCPU`: the process exceeded a previously specified CPU [rlimit](https://pubs.opengroup.org/onlinepubs/9699919799/functions/getrlimit.html).
+* `SIGXFSZ`: similar to `SIGXCPU`, raised when a process grows a file beyond a maximum allowed size ([`RLIMIT_FSIZE`](https://man7.org/linux/man-pages/man2/getrlimit.2.html)). 
+
+If your application currently installs handlers for any of the above signals, that can result in a potential conflict. Expressly
+
+* refrain from registering a conflicting signal handler in your application.
+* if you have to, register this handler after `sentry_init()` and let it invoke the previously installed handler (ours) after it finished or you won't get a report.
+* if you or one of your dependencies register a handler for the above signals _before_ you call `sentry_init()`, we will reinstall the previous handler as soon as you call `sentry_close()`. Only the `inproc` backend invokes the previously installed handler at the end of its handler.
+* if you want to act on any crash and not consider specific signals, [register an `on_crash()` hook](https://docs.sentry.io/platforms/native/configuration/filtering/#using-on_crash).
```
vivianyentran, line 42:
```suggestion
## Signal Handling Conflicts

Sentry's Native SDK comes with its own set of signal handlers. There may be conflicts if your application currently installs handlers for any [signals of interest](#signals-of-interest) (listed below) that can result in a crash. To avoid this, do the following:

* Refrain from registering a conflicting signal handler in your application.
* If you have to, register this handler after `sentry_init()` and let it invoke the previously installed handler (ours) after it finishes or you won't get a report.
* If you or one of your dependencies register a handler for the signals listed below _before_ you call `sentry_init()`, we will reinstall the previous handler as soon as you call `sentry_close()`. Only the `inproc` backend invokes the previously installed handler at the end of its handler.
* If you want to act on any crash and not consider specific signals, [register an `on_crash()` hook](https://docs.sentry.io/platforms/native/configuration/filtering/#using-on_crash).

### Signals of Interest

In the context of the Native SDK, we are only interested in signals that will result in the application's unexpected termination (crashes). These typically include:

* `SIGSEGV`: The process tried to access an invalid virtual memory "segment".
* `SIGBUS`: The process tried to access an invalid physical address or a VM page failed to "page-in".
* `SIGABRT`: Raised when `abort()` was called. This is the only signal handled on Windows (and only by the `crashpad` backend) because calling `abort()` bypasses Structured Exception Handling ([`SEH`](https://learn.microsoft.com/en-us/windows/win32/debug/structured-exception-handling)).
* `SIGFPE`: An erroneous arithmetic operation that occurred in the CPU.
* `SIGILL`: The process attempted to execute an illegal (often privileged) processor instruction.
* `SIGTRAP`: Typically used to inform debuggers when execution hits a breakpoint instruction (`int 3`, `brk`, etc.), which instrumentation tools and sanitizers sometimes also issue.

If you use the `crashpad` backend, in addition to the above, on Linux, you will also get crash reports for:

* `SIGSYS`: The kernel received a system call with an invalid argument.
* `SIGEMT`: An emulator trap occurred.
* `SIGXCPU`: The process exceeded a previously specified CPU [rlimit](https://pubs.opengroup.org/onlinepubs/9699919799/functions/getrlimit.html).
* `SIGXFSZ`: Similar to `SIGXCPU`, raised when a process grows a file beyond a maximum allowed size ([`RLIMIT_FSIZE`](https://man7.org/linux/man-pages/man2/getrlimit.2.html)). 
```

I think it makes more sense to put the useful info (how to avoid conflicts) up top so it's easier for users to skim. Otherwise, it's easy to miss after a long reference list. I'd also create a new subsection for signals of interest for easy reference.

vivianyentran, line 42: I think referencing "previous handler" gets confusing. Can we simplify the wording by saying "your handler" and "Sentry's handler"? That would make it much clearer which handler we're referencing.

docs/platforms/native/advanced-usage/signal-handling/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+## What to consider when writing `on_crash` hooks
+
+The callback that you register as an `on_crash` hook (and `before_send` if you didn't turn it off for crashes) runs in the context of a signal handler.
+
+Since signal handlers can interrupt the running process at any time (and even each other), take  special care when writing them. 
+
+Our ["develop" docs provide guidelines for writing signal handlers](https://develop.sentry.dev/sdk/signal-handlers/), which also apply to the hooks if you run on Linux or Android. Please take them seriously.
+
+## Stack size in the signal handler
```
vivianyentran, line 52:
```suggestion
## Stack Size in the Signal Handler
```

docs/platforms/native/advanced-usage/signal-handling/index.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+In addition to the above guidelines, you must consider the signal handler's stack usage. Because a stack overflow (where the process stack is exhausted) can be one reason to end up in a signal handler, we typically reserve a separate signal handler stack (`sigaltstack()`) so that we can continue to run the handler when it gets executed.
+
+This stack is 64KiB on Linux. On Android, it is 16KiB on 32-bit and 32KiB on 64-bit devices.
```
vivianyentran, line 56:
```suggestion
This stack is 64KiB on Linux. On Android, it is 16KiB on 32-bit devices and 32KiB on 64-bit devices.
```

vivianyentran (changes requested):
Did a first pass at the signal handling section and left some suggestions. Didn't have time to review the "advanced usage" page. Will take another look tomorrow!

docs/platforms/native/advanced-usage/signal-handling/index.mdx
supervacuus: Yeah, that makes a lot of sense. Thanks. Fixed it here: https://github.com/getsentry/sentry-docs/pull/9787/commits/fb22952c2f2d4e20122bbcf3243703fc958eba00

docs/platforms/native/advanced-usage/signal-handling/index.mdx
```diff
@@ -0,0 +1,62 @@
+---
+title: Signal Handling
+description: "Learn about error handling common to all POSIX platforms."
+sidebar_order: 1000
+---
+The Native SDK uses signal handling on all supported platforms. On Linux (and Android), it is the primary mechanism to detect errors, while on macOS and Windows, signal handlers detect specific errors that other mechanisms do not detect or in lightweight backends that don't support the lower-level error handling mechanisms (for instance `inproc` on macOS).
+
+A (POSIX) signal is a way for an operating system kernel to inform an application about an exceptional event and provide a mechanism for the application to react to that event. The operating system delivers this information by interrupting the current execution of the affected  application with a "handler" function that the application registered before.
```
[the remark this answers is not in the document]
supervacuus: I can't entirely agree here because "exception" is an overloaded term; it either is too generic or has a technical meaning that refers to something very different from signals. 

I am okay with using different wording than "exceptional event", but "exception" is confusing given what most technical readers have in mind when reading that term. What about the following:

> A (POSIX) signal is a way for an operating system kernel to interrupt the execution of an application to deliver a message of urgency, often with the intent to terminate the application. The affected application can react to that interruption with a "handler" function it registered before.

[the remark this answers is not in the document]
kahest: IMO we can keep "exceptional event" because it was used intentional as a more generic term

docs/platforms/native/advanced-usage/index.mdx
```diff
@@ -0,0 +1,14 @@
...
+description: "Learn how the Native SDK backends handle errors on their supported platforms and how that affects your usage."
+sidebar_order: 6000
+---
+This section details recurring topics where the knowledge beyond what the SDK boundary can abstract away from its users no longer holds.
+
+These will include operating system specifics regarding error handling or how some platforms deviate from established standards.
+
+This section is not a conclusive reference documentation but will cover solutions that a user should understand to adapt to particular platform behavior.
+
+While we try to keep the content light, due to the in-depth nature of the topics following the text might require more effort than the other parts of the documentation.
```
vivianyentran, line 12:
```suggestion
This section discusses recurring topics that go beyond what the SDK can abstract away from users. 

While not a complete reference, this section covers topics like how different operating systems handle errors and where some platforms don't follow usual standards. Here, we also offer solutions to help users deal with specific platform quirks. 

Due to the in-depth and nuanced nature of these topics, users may need a more advanced level of expertise. 
```

docs/platforms/native/advanced-usage/signal-handling/index.mdx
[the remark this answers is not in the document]
vivianyentran: ah, makes sense. I'm ok with leaving it as "exceptional event" then.

docs/platforms/native/advanced-usage/signal-handling/index.mdx
```diff
@@ -0,0 +1,62 @@
+---
+title: Signal Handling
+description: "Learn about error handling common to all POSIX platforms."
+sidebar_order: 1000
+---
+The Native SDK uses signal handling on all supported platforms. On Linux (and Android), it is the primary mechanism to detect errors, while on macOS and Windows, signal handlers detect specific errors that other mechanisms do not detect or in lightweight backends that don't support the lower-level error handling mechanisms (for instance `inproc` on macOS).
```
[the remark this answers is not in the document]
vivianyentran: I like the first option!

supervacuus:
I adapted the text with all agreed-upon suggestions. Thanks a lot, @vivianyentran and @kahest!

kahest (approved):
Thanks for this, looks really good now 🚀

docs/platforms/native/advanced-usage/signal-handling/index.mdx
```diff
@@ -0,0 +1,64 @@
...
+* `SIGXCPU`: The process exceeded a previously specified CPU [rlimit](https://pubs.opengroup.org/onlinepubs/9699919799/functions/getrlimit.html).
+* `SIGXFSZ`: Similar to `SIGXCPU`, raised when a process grows a file beyond a maximum allowed size ([`RLIMIT_FSIZE`](https://man7.org/linux/man-pages/man2/getrlimit.2.html)). 
+
+## What to Consider When Writing `on_crash` Hooks
+
+The callback that you register as an `on_crash` hook (and `before_send` if you didn't turn it off for crashes) runs in the context of a signal handler.
+
+Since signal handlers can interrupt the running process at any time (and even each other), take  special care when writing them. 
+
+Our ["develop" docs provide guidelines for writing signal handlers](https://develop.sentry.dev/sdk/signal-handlers/), which also apply to the hooks if you run on Linux or Android. Please take them seriously.
```
vivianyentran, line 52:
```suggestion
Our developer documentation provides [guidelines for writing signal handlers](https://develop.sentry.dev/sdk/signal-handlers/). If you run on Linux or Android, these guidelines also apply to hooks. Signal handlers are often difficult to work with, so following these guidelines is important if you want to get good crash reports.
```
Instead of telling them to take it seriously, let's give them a reason to 🙂

vivianyentran (approved):
Added a couple more wording suggestions, but this is looking great now and feels much easier to read!

docs/platforms/native/advanced-usage/signal-handling/index.mdx
supervacuus: Okay, let's make it really impactful then: this is not about getting good or better crash reports but rather about causing crashes or other severe faults (like deadlocks) in the handler itself, which can prevent sending reports or programs from terminating.

PR #9796, opened 2024-04-24
feat: Add JavaScript SDK migration docs to Sentry docs site

AbhiPrasad:
This is a start to revamped migration docs for our Sentry JavaScript SDK.

These still link to the repo readme, but contain more accessible migration guides for our users. As an example, I've added a relatively fleshed out set of instructions for migrating a `Next.js` app.

If this general approach is agree'd upon, I'm going to work toward porting the rest of the migration docs into fitting this format, and then adding docs for Node, Browser, React, and Express. With those changed (and Next.js being finished), we then incrementally tackle the other more detailed migration guides whenever we want! (we could even merge them in after v8 releases).

Top level migration docs:

image

Troubleshooting (maybe should be named FAQ?) at the bottom:

image

OpenTelemetry docs! 

image

### DIFF
--- a/docs/platforms/javascript/common/migration/v6-to-v7.mdx
+++ b/docs/platforms/javascript/common/migration/v6-to-v7.mdx
@@ -1,6 +1,6 @@
 ---
-title: Migrate from Sentry JavaScript SDK 6.x to 7.x
-sidebar_order: 8998
+title: Migrate from 6.x to 7.x
+sidebar_order: 8997
 description: "Learn about migrating from Sentry JavaScript SDK 6.x to 7.x"
 ---
 
@@ -12,13 +12,13 @@ should consider when upgrading.
 
 **TL;DR** If you only use basic features of Sentry, or you simply copy & pasted the setup examples from our docs, here's what changed for you:
 
-- If you installed additional Sentry packages, such as`@sentry/tracing` alongside your Sentry SDK (e.g. `@sentry/react` or `@sentry/node`), make sure to upgrade all of them to version 7.
+- If you installed additional Sentry packages, such as `@sentry/tracing` alongside your Sentry SDK (e.g. `@sentry/react` or `@sentry/node`), make sure to upgrade all of them to version 7.
 - Our CDN bundles are now ES6 - you will need to reconfigure your script tags if you want to
   keep supporting ES5 and IE11 on the new SDK version.
 - Distributed CommonJS files will be ES6. Use a transpiler if you need to support old node versions.
-- We bumped the TypeScript version we generate our types with to 3.8.3. Please check if your TypeScript projects using TypeScript version 3.7 or lower still compile. Otherwise, upgrade your TypeScript version.
+- We bumped the TypeScript version we generate our types with to `3.8.3`. Please check if your TypeScript projects using TypeScript version `3.7` or lower still compile. Otherwise, upgrade your TypeScript version.
 - `whitelistUrls` and `blacklistUrls` have been renamed to `allowUrls` and `denyUrls` in the `Sentry.init()` options.
 - The `UserAgent` integration is now called `HttpContext`.
 - If you are using Performance Monitoring and with tracing enabled, you might have to make adjustments to your server's CORS settings.
 
-Please see our detailed migration guide for more information on migrating from `6.x` to `7.x`: [link](https://github.com/getsentry/sentry-javascript/blob/develop/docs/migration/v6-to-v7.md)
+Please see our [detailed migration guide](https://github.com/getsentry/sentry-javascript/blob/develop/docs/migration/v6-to-v7.md) for more information on migrating from `6.x` to `7.x`.
--- a/docs/platforms/javascript/common/migration/v6-to-v7.mdx
+++ b/docs/platforms/javascript/common/migration/v6-to-v7.mdx
@@ -1,0 +1,24 @@
+---
+title: Migrate from Sentry JavaScript SDK 6.x to 7.x
+sidebar_order: 8998
+description: "Learn about migrating from Sentry JavaScript SDK 6.x to 7.x"
+---
+
+The v7 version of the JavaScript SDK requires a self-hosted version of Sentry `20.6.0` or higher.
+
+The main goal of version 7 is to reduce bundle size. This version is breaking because we removed deprecated APIs,
+upgraded our build tooling, and restructured npm package contents. Below we will outline all the breaking changes you
+should consider when upgrading.
+
+**TL;DR** If you only use basic features of Sentry, or you simply copy & pasted the setup examples from our docs, here's what changed for you:
+
+- If you installed additional Sentry packages, such as`@sentry/tracing` alongside your Sentry SDK (e.g. `@sentry/react` or `@sentry/node`), make sure to upgrade all of them to version 7.
+- Our CDN bundles are now ES6 - you will need to reconfigure your script tags if you want to
+  keep supporting ES5 and IE11 on the new SDK version.
+- Distributed CommonJS files will be ES6. Use a transpiler if you need to support old node versions.
+- We bumped the TypeScript version we generate our types with to 3.8.3. Please check if your TypeScript projects using TypeScript version 3.7 or lower still compile. Otherwise, upgrade your TypeScript version.
+- `whitelistUrls` and `blacklistUrls` have been renamed to `allowUrls` and `denyUrls` in the `Sentry.init()` options.
+- The `UserAgent` integration is now called `HttpContext`.
+- If you are using Performance Monitoring and with tracing enabled, you might have to make adjustments to your server's CORS settings.
+
+Please see our detailed migration guide for more information on migrating from `6.x` to `7.x`: [link](https://github.com/getsentry/sentry-javascript/blob/develop/docs/migration/v6-to-v7.md)
--- a/docs/platforms/javascript/common/migration/v6-to-v7.mdx
+++ b/docs/platforms/javascript/common/migration/v6-to-v7.mdx
@@ -1,6 +1,6 @@
 ---
 title: Migrate from 6.x to 7.x
-sidebar_order: 8997
+sidebar_order: 8920
 description: "Learn about migrating from Sentry JavaScript SDK 6.x to 7.x"
 ---
 
--- a/docs/platforms/javascript/common/migration/v7-to-v8/index.mdx
+++ b/docs/platforms/javascript/common/migration/v7-to-v8/index.mdx
@@ -1,0 +1,79 @@
+---
+title: Migrate from 7.x to 8.x
+sidebar_order: 8999
+description: "Learn about migrating from Sentry JavaScript SDK 7.x to 8.x"
+---
+
+The main goal of version 8 is to improve our performance monitoring APIs, integrations API, and ESM support. This
+version is breaking because we removed deprecated APIs, restructured npm package contents, and introduced new
+dependencies on OpenTelemetry.
+
+## Migration Codemod
+
+Before updating to `8.x` of the SDK, we recommend upgrading to the latest version of `7.x`. To fix all deprecations on `7.x`, you can use the `@sentry/migr8` codemod to automatically update your SDK usage. `@sentry/migr8` requires Node 18+.
+
+```bash
+npx @sentry/migr8@latest
+```
+
+This will let you select which updates to run, and automatically update your code. Make sure to still review all code changes!
+
+## Migration Guide
+
+Please see our [detailed migration guide](https://github.com/getsentry/sentry-javascript/blob/develop/MIGRATION.md#upgrading-from-7x-to-8x) for more information on migrating from `7.x` to `8.x`.
+
+{/* Make sure to update the PlatformSection below when new fully fleshed out migration guides are added */}
+
+<PlatformSection supported={["javascript.nextjs"]}>
+
+## Upgrading to `8.x`
+
+<PlatformContent includePath="migration/javascript-v8/intro" />
+
+We recommend you read through all the [Important Changes](#important-changes) as they affect all Next.js SDK users. The [Other Changes](#other-changes) linked below only affect users who have very custom instrumentation. There is also a [Troubleshooting](#troubleshooting) section for common issues.
+
+## Important Changes
+
+<PlatformContent includePath="migration/javascript-v8/important-changes" />
+
+## Other Changes
+
+<PlatformContent includePath="migration/javascript-v8/other-changes" />
+
+### Removal of `@sentry/replay` package
+
+The `@sentry/replay` package is no longer required. Instead you can import the relevant methods directly from your SDK. In addition to this, the integration is now functional instead of class based.
+
+```JavaScript diff
+-import { Replay } from '@sentry/replay';
+-
+ Sentry.init({
+   dsn: '___PUBLIC_DSN___',
+   integrations: [
+-    new Replay(),
++    new Sentry.replayIntegration(),
+   ],
+ });
+```
+
+<PlatformContent includePath="migration/javascript-v8/integrations-package-removal" />
+
+### Removal of `@sentry/hub`
+
+`@sentry/hub` has been removed and will no longer be published. All of the @sentry/hub exports have moved to `@sentry/core` or other related SDKs.
+
+<PlatformContent includePath="migration/javascript-v8/tracing-package-removal" />
+
+## Troubleshooting
+
+<PlatformContent includePath="migration/javascript-v8/troubleshooting" />
+
+</PlatformSection>
+
+<PlatformSection notSupported={["javascript.nextjs"]}>
+
+{/* TODO: Remove the generic summary */}
+
+<PlatformContent includePath="migration/javascript-v8/summary" />
+
+</PlatformSection>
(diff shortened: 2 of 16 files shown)


docs/platforms/javascript/common/migration/v7-to-v8/index.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+<PlatformContent includePath="migration/javascript-v8/troubleshooting" />
+
+</PlatformSection>
+
+<PlatformSection notSupported={["javascript.nextjs"]}>
+
+{/* TODO: Remove the generic summary */}
+
+<PlatformContent includePath="migration/javascript-v8/summary" />
```
AbhiPrasad, line 77: For SDKs we have not created detailed migration docs for, we reference a summary in `platform-includes/migration/javascript-v8/summary/javascript.mdx`. This means we can incrementally convert SDKs to have more detailed migration docs based on priority.

Ideally by the end, no SDK uses these generic docs and each of them have structures similar to Next.js docs above

docs/platforms/javascript/common/migration/v7-to-v8/v8-opentelemetry.mdx
```diff
@@ -0,0 +1,93 @@
+---
+title: OpenTelemetry Support
```
AbhiPrasad, line 2: Eventually we should move this under `javascript/guides/node/performance/instrumentation/opentelemetry/` and similar, but I think this is a good place for this to live for now while v8 is in a beta state.

docs/platforms/javascript/common/migration/v7-to-v8/v8-new-performance-api.mdx
```diff
@@ -0,0 +1,253 @@
+---
+title: New Performance Monitoring APIs
```
AbhiPrasad, line 2: Eventually this should live under `platforms/javascript/guides/node/performance/instrumentation/custom-instrumentation/`, but this is a good spot to throw all of this in until we finally release v8.

docs/platforms/javascript/common/migration/v6-to-v7.mdx
```diff
@@ -0,0 +1,24 @@
+---
+title: Migrate from 6.x to 7.x
+sidebar_order: 8997
```
mydea, line 3: Can we invert the order so the most recent migration (e.g. 7. to 8.) is on top?

docs/platforms/javascript/common/migration/v7-to-v8/index.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+{/* Make sure to update the PlatformSection below when new fully fleshed out migration guides are added */}
+
+<PlatformSection supported={["javascript.nextjs"]}>
+
+## Upgrading to `8.x`
+
+<PlatformContent includePath="migration/javascript-v8/intro" />
+
+We recommend you read through all the [Important Changes](#important-changes) as they affect all Next.js SDK users. The [Other Changes](#other-changes) linked below only affect users who have very custom instrumentation. There is also a [Troubleshooting](#troubleshooting) section for common issues.
```
mydea, line 33: Not sure if we want/should do this in this PR or in a follow up, but we should make sure to only show the relevant stuff for the current guide (e.g. replay can be hidden for node guides, profiling-node for non-node guides etc).

AbhiPrasad: Yes! I moved replay into browser specific other changes

lforst (approved):
Forgot to press approve. The only thing I would add is maybe an <Alert /> on top of the getting started docs about v8

docs/platforms/javascript/common/migration/v7-to-v8/index.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+## Migration Codemod
+
+Before updating to `8.x` of the SDK, we recommend upgrading to the latest version of `7.x`. To fix all deprecations on `7.x`, you can use the `@sentry/migr8` codemod to automatically update your SDK usage. `@sentry/migr8` requires Node 18+.
+
+```bash
+npx @sentry/migr8@latest
+```
+
+This will let you select which updates to run, and automatically update your code. Make sure to still review all code changes!
```
Lms24, line 19:
```suggestion
Our migration tool will let you select which updates to run, and automatically update your code. In some cases, we cannot automatically change code for you. These will be marked with a `TODO(sentry)` instead. Make sure to review all code changes after running `@sentry/migr8`!
```

platform-includes/migration/javascript-v8/summary/javascript.mdx
```diff
@@ -0,0 +1,48 @@
...
+  - Opera 50
+  - Samsung Internet 8.2
+
+For IE11 support please transpile your code to ES5 using babel or similar and add required polyfills.
+
+- Integrations are now functional instead of class based. For example, the import for session replay uses `Sentry.replayIntegration` instead of `new Sentry.Replay`.
+
+- The SDK's Performance API has been revamped. More details can be found in documentation about the [new Performance Monitoring APIs](./v8-new-performance-api/)
+
+- Initializing the Node SDK has been vastly simplified. You no longer need to use `autoDiscoverNodePerformanceMonitoringIntegrations` or use `Sentry.Handlers`, as these are now automatically included by the SDK without configuration. To make this work, you now have to ensure `Sentry.init` is initialized before any other code is imported and used.
```
Lms24, line 22: This point, along with the code snippet below, currently also shows up in browser guides. I think we should scope it to node guides only.

Lms24, line 22: l: I'd make this bold as it's crucial but easy to miss:

```suggestion
- Initializing the Node SDK has been vastly simplified. You no longer need to use `autoDiscoverNodePerformanceMonitoringIntegrations` or use `Sentry.Handlers`, as these are now automatically included by the SDK without configuration. To make this work, you now have to ensure `Sentry.init` is initialized **before any other code is imported and used**.
```

docs/platforms/javascript/common/migration/v7-to-v8/v8-new-performance-api.mdx
```diff
@@ -0,0 +1,253 @@
...
+
+There are three key APIs available to start spans:
+
+- `startSpan()`
+- `startSpanManual()`
+- `startInactiveSpan()`
+
+All three span APIs take `StartSpanOptions` as a first argument, which has the following shape:
+
+```ts
```
Lms24, line 30: l/nit (feel free to ignore): hmm I could have sworn this worked better before but `ts`/`js` renders to 

image

Should we change to "JavaScript"/"TypeScript" anyway for the meantime?

Lms24 (approved):
Nice work! This looks a lot better than the repo-only migration guide file.

docs/platforms/javascript/common/migration/v7-to-v8/v8-new-performance-api.mdx
```diff
@@ -0,0 +1,253 @@
+---
+title: New Performance Monitoring APIs
+sidebar_order: 8997
+description: "Learn about new Performance Monitoring APIs in Sentry JavaScript SDK 8.x"
+---
+
+The SDK `8.x` release introduces new APIs for performance monitoring. These APIs are designed to provide more control over how performance data is collected and reported to Sentry.
+
+These APIs are also available in `7.x` so you can incrementally moveover to the new APIs before upgrading to `8.x`.
```
s1gr1d, line 9:
```suggestion
These APIs are also available in `7.x` so you can incrementally move over to the new APIs before upgrading to `8.x`.
```

docs/platforms/javascript/common/migration/v7-to-v8/v8-new-performance-api.mdx
```diff
@@ -0,0 +1,253 @@
...
+
+Previously, there were two key APIs for adding manual performance instrumentation to your applications:
+
+- `startTransaction()`
+- `span.startChild()`
+
+This showed the underlying data model that Sentry was originally based on, which is that there is a root Transaction which can have a nested tree of Spans.
```
s1gr1d, line 16:
```suggestion
Those APIs were based on the original data model of Sentry. This model was based on a root Transaction that could contain a nested tree of Spans.
```

docs/platforms/javascript/common/migration/v7-to-v8/v8-new-performance-api.mdx
```diff
@@ -0,0 +1,253 @@
...
+
+In the new model, transactions are conceptually gone. Instead, you will always operate on spans, no matter where in the tree you are. Note that in the background, spans may still be grouped into a transaction for the Sentry UI. However, this happens transparently, and from an SDK perspective, all you have to think about are spans.
```
s1gr1d, line 18: In the new model, transactions are conceptually gone. Now, operations are always performed on spans, regardless of their position in the tree. Note that spans may still be grouped into a transaction for the Sentry UI. However, this happens in the background and, from the SDK's point of view, all you need to be concerned about are spans.

docs/platforms/javascript/common/migration/v7-to-v8/v8-new-performance-api.mdx
```diff
@@ -0,0 +1,253 @@
...
+
+Instead of manually starting & ending transactions and spans, the new model does not differentiate between these two. Instead, you always use the same APIs to start a new span, and it will automatically either create a new Root Span (which is just a regular span, only that it has no parent, and is thus conceptually roughly similar to a transaction) or a Child Span for whatever is the currently active span.
```
s1gr1d, line 20:
```suggestion
The new model does not differentiate between a transaction or a span when manually starting or ending them. Instead, you always use the same APIs to start a new span. This will automatically create either a new Root Span (a regular span without a parent, conceptually similar to a transaction) or a Child Span, depending on what is the currently active span.
```

docs/platforms/javascript/common/migration/v7-to-v8/v8-new-performance-api.mdx
```diff
@@ -0,0 +1,253 @@
...
+  attributes?: SpanAttributes;
+  op?: string;
+  scope?: Scope;
+  forceTransaction?: boolean;
+}
+```
+
+### `startSpan()`
+
+This is the most common API that should be used in most circumstances. It will start a new span, make it the active span
```
s1gr1d, line 43:
```suggestion
This is the most common API that should be used in most circumstances. It will start a new span, making it the active span
```

docs/platforms/javascript/common/migration/v7-to-v8/v8-new-performance-api.mdx
```diff
@@ -0,0 +1,253 @@
...
+for the duration of a given callback, and automatically end it when the callback ends. You can use it like this:
```
s1gr1d, line 44:
```suggestion
for the duration of a given callback, and automatically ends it when the callback ends. You can use it like this:
```

docs/platforms/javascript/common/migration/v7-to-v8/v8-new-performance-api.mdx
```diff
@@ -0,0 +1,253 @@
...
+
+Since `startSpan()` will make the created span the active span, any automatic or manual instrumentation that creates
+spans inside of the callback will attach new spans as children of the span we just started.
+
+Note that if an error is thrown inside of the callback, the span status will automatically be set to be errored.
+
+### `startSpanManual()`
+
+This is a variation of `startSpan()` with the only change that it does not automatically end the span when the callback
+ends, but you have to call `span.end()` yourself:
```
s1gr1d, line 86: The start of the sentence can also be something else. I just thought it reads easier when having shorter sentences.

```suggestion
This is a variation of `startSpan()` with the only difference that it does not automatically end the span when the callback
ends. When using this, you have to call `span.end()` yourself:
```

docs/platforms/javascript/common/migration/v7-to-v8/v8-new-performance-api.mdx
```diff
@@ -0,0 +1,253 @@
...
+| `getDynamicSamplingContext` | `getDynamicSamplingContextFromSpan(span)`        |
+
+### Attributes vs. Data vs. Tags vs. Context
+
+In the old model, you had the concepts of **Data**, **Tags** and **Context** which could be used for different things.
+However, this has two main downsides: One, it is not always clear which of these should be used when. And two, not all
+of these are displayed the same way for transactions or spans.
+
+Because of this, in the new model, there are only **Attributes** to be set on spans anymore. Broadly speaking, they map
+to what Data used to be.
```
s1gr1d, line 209:
```suggestion
Previously, the model had the concepts of **Data**, **Tags**, and **Context**, which could be used for different things. However, this had two main downsides: Firstly, it was not always clear which concept should be used in a given situation. Secondly, these concepts were not displayed the same way for transactions or spans.  

To address these issues, the new model only has **Attributes** that can be set on spans. Broadly speaking, they map to what was formerly known as Data.
```

s1gr1d (commented):
Only halfway through, finishing it up later :D

docs/platforms/javascript/common/migration/v7-to-v8/v8-new-performance-api.mdx
```diff
@@ -0,0 +1,253 @@
...
+anywhere in detail, the shape of this context will change in v8. Going forward, this will mostly receive the attributes
+of the span, as well as some other relevant data of the span. Some properties we used to (sometimes) pass there, like
+`req` for node-based SDKs or `location` for browser tracing, will not be passed anymore.
+
+### No more `undefined` spans
+
+In v7, the performance APIs `startSpan()` / `startInactiveSpan()` / `startSpanManual()` would receive an `undefined`
+span if tracing was disabled or the span was not sampled.
+
+In v8, aligning with OpenTelemetry, these will _always_ return a span - _but_ the span may eb a Noop-Span, meaning a
```
s1gr1d, line 239:
```suggestion
In v8, aligning with OpenTelemetry, these will _always_ return a span - _but_ the span may be a Noop-Span, meaning a
```

includes/migration/javascript-v8/browser-other-changes.mdx
```diff
@@ -0,0 +1,19 @@
+### Removal of `@sentry/replay` package
+
+The `@sentry/replay` package is no longer required. Instead you can import the relevant methods directly from your SDK. In addition to this, the integration is now functional instead of class based.
```
s1gr1d, line 3:
```suggestion
The `@sentry/replay` package is no longer required. Instead you can import the relevant methods directly from your SDK. In addition to this, the integration is now functional instead of class-based.
```

platform-includes/migration/javascript-v8/integrations-package-removal/javascript.nextjs.mdx
```diff
@@ -0,0 +1,20 @@
+### Removal of `@sentry/integrations` package
```
s1gr1d, line 1: As far as I can see this only exists for Next.js in this PR, but this also applies to other packages 🤔

s1gr1d (commented):
Thanks Abhi for creating those docs  🙌🏻

docs/platforms/javascript/common/migration/v7-to-v8/index.mdx
```diff
@@ -0,0 +1,79 @@
...
+description: "Learn about migrating from Sentry JavaScript SDK 7.x to 8.x"
+---
+
+The main goal of version 8 is to improve our performance monitoring APIs, integrations API, and ESM support. This
+version is breaking because we removed deprecated APIs, restructured npm package contents, and introduced new
+dependencies on OpenTelemetry.
+
+## Migration Codemod
+
+Before updating to `8.x` of the SDK, we recommend upgrading to the latest version of `7.x`. To fix all deprecations on `7.x`, you can use the `@sentry/migr8` codemod to automatically update your SDK usage. `@sentry/migr8` requires Node 18+.
```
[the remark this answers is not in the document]
mydea: yeah, should probably be "fix most deprecations" or something like this? 😅

docs/platforms/javascript/common/migration/v7-to-v8/v8-new-performance-api.mdx
AbhiPrasad:
```suggestion
In the new model, transactions are conceptually gone. Now, operations are always performed on spans, regardless of their position in the tree. Note that spans may still be grouped into a transaction for the Sentry UI. However, this happens in the background and, from the SDK's point of view, all you need to be concerned about are spans.
```

platform-includes/migration/javascript-v8/integrations-package-removal/javascript.nextjs.mdx
AbhiPrasad: Yes this is because this is part of the next.js specific docs! The rest of the sdks rely on the simpler migration docs, but we'll move them over to the better one's one by one.

platform-includes/migration/javascript-v8/summary/javascript.mdx
AbhiPrasad: This is a good idea, but I'm going to do it in a follow up.

PR #9812, opened 2024-04-25
docs(integrations): Add environment information to integrations

s1gr1d:
## Description of changes

To also show Node.js integrations in meta-frameworks like Next.js or SvelteKit.

### DIFF
--- a/docs/platforms/javascript/common/configuration/integrations/console.mdx
+++ b/docs/platforms/javascript/common/configuration/integrations/console.mdx
@@ -18,7 +18,7 @@ supported:
 ---
 
 <Alert level="info">
-  This integration only works inside a Node.js environment.
+  This integration only works inside server environments (Node.js, Bun, Deno).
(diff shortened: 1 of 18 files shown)


docs/platforms/javascript/common/configuration/integrations/console.mdx
```diff
@@ -11,8 +11,16 @@ supported:
...
   - javascript.koa
   - javascript.electron
+  - javascript.nextjs
+  - javascript.sveltekit
+  - javascript.remix
+  - javascript.astro
 ---
 
+<Alert level="info">
+  This integration only works inside a Node.js environment.
```
AbhiPrasad, line 21: So it works for bun and deno as well, we should make that clear. What do you think about something like?

```suggestion
  This integration only works inside server environments (Node.js, Bun, Deno).
```

vivianyentran (approved):

PR #9832, opened 2024-04-26
docs(relocation): Apply comments from #9816

azaslavsky:


lizokm (approved):
This looks good! Thanks for taking the time :)

PR #9842, opened 2024-04-27
feat: `docs/**` content change hot reload

### LINKED ISSUE #9038: Markdown file changes don't trigger hot reloads
Reported by @mydea on Slack 🙌

a-hariti:
# Authoring quality of life improvement

Next.js seems to be blind to content changes under `/docs/**/` because content files are outside the module dependency graph

There's no obvious way in the docs to let Next.js know about the changes directly

This is workaround establishes a Web Socket connection with a backend file watcher and triggers a hot reload as needed.

This solution should add ~0 js to the bundle sizse.

Closes #9038

### DIFF
--- a/src/hotReloadWatcher.mjs
+++ b/src/hotReloadWatcher.mjs
@@ -1,0 +1,53 @@
+import path from 'path';
+import {watch} from 'node:fs/promises';
+import {WebSocketServer} from 'ws';
+
+const watchedConent = new Set(['.mdx', '.md', '.png', '.jpg', '.jpeg', '.gif', '.svg']);
+
+export const throttle = (fn, delay) => {
+  let last = 0;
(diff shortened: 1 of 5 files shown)


stephanie-anderson (approved):
omg, thanks so much!! 🙏

src/hotReloadWatcher.mjs
```diff
@@ -0,0 +1,53 @@
+import path from 'path';
+import {watch} from 'node:fs/promises';
+import {WebSocketServer} from 'ws';
+
+const watchedConent = new Set(['.mdx', '.md', '.png', '.jpg', '.jpeg', '.gif', '.svg']);
```
stephanie-anderson, line 5: `watchedConent` has a typo :)

PR #9906, opened 2024-05-03
feat(v8): Update node docs for v8

mydea:
This is a draft until v8 is released!

PR #9907, opened 2024-05-03
feat(js): Update scope docs for v8

mydea:
I split this out of https://github.com/getsentry/sentry-docs/pull/9906 because I think we can merge this independently, and this is much more content heavy than the other PR.

### DIFF
--- a/docs/platforms/javascript/common/enriching-events/scopes/index.mdx
+++ b/docs/platforms/javascript/common/enriching-events/scopes/index.mdx
@@ -8,7 +8,7 @@ information from the current scope. SDKs will typically automatically manage the
 for you in the framework integrations and you don't need to think about them. However,
 you should know what a scope is and how you can use it for your advantage.
 
-## What's a Scope
+## What's a Scope?
 
 The scope will hold useful information that should be sent along with the
 event. For instance [contexts](../context/) or
@@ -17,7 +17,7 @@ the scope. When a scope is forked, it inherits all data from the parent scope.
 
 The default SDK integrations will fork scopes intelligently. For
 instance web framework integrations will fork scopes around your
-routes or controllers.
+routes or request handlers.
 
 ## How Scopes Work
 
--- a/docs/platforms/javascript/common/enriching-events/scopes/index.mdx
+++ b/docs/platforms/javascript/common/enriching-events/scopes/index.mdx
@@ -1,5 +1,5 @@
 ---
-title: Scopes and Hubs
+title: Scopes
 description: "SDKs will typically automatically manage the scopes for you in the framework integrations. Learn what a scope is and how you can use it to your advantage."
 ---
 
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/common/enriching-events/scopes/index.mdx
```diff
@@ -8,35 +8,97 @@ information from the current scope. SDKs will typically automatically manage the
 for you in the framework integrations and you don't need to think about them. However,
 you should know what a scope is and how you can use it for your advantage.
 
-## What's a Scope, What's a Hub
-
-You can think of the hub as the central point that our SDKs use to route an
-event to Sentry. When you call `init()` a hub is created and a client and a
-blank scope are created on it. That hub is then associated with the current
-thread and will internally hold a stack of scopes.
+## What's a Scope
```
Lms24, line 11:
```suggestion
## What's a Scope?
```

docs/platforms/javascript/common/enriching-events/scopes/index.mdx
```diff
@@ -8,35 +8,97 @@ information from the current scope. SDKs will typically automatically manage the
...
 
 The scope will hold useful information that should be sent along with the
 event. For instance [contexts](../context/) or
 [breadcrumbs](../breadcrumbs/) are stored on
-the scope. When a scope is pushed, it inherits all data from the parent scope
-and when it pops all modifications are reverted.
+the scope. When a scope is forked, it inherits all data from the parent scope.
```
Lms24, line 16: l: I'm a bit on the fence if we should publicly speak about "forking" a scope. It's one of these terms that not all devs will understand immediately. Though to be honest, just saying "cloning" is probably not descriptive enough and something like "create scopes" doesn't imply that we inherit data... (especially in the next paragraph). Not sure, feel free to leave it as-is; this is an expert section anyway

docs/platforms/javascript/common/enriching-events/scopes/index.mdx
```diff
@@ -8,35 +8,97 @@ information from the current scope. SDKs will typically automatically manage the
...
 
-The default SDK integrations will push and pop scopes intelligently. For
-instance web framework integrations will create and destroy scopes around your
+The default SDK integrations will fork scopes intelligently. For
+instance web framework integrations will fork scopes around your
 routes or controllers.
```
Lms24, line 20: l: Maybe a bit closer to the truth, wdyt? 🤔 

```suggestion
instance web framework integrations will fork scopes around your
routes or request handlers.
```

Lms24 (approved):

docs/platforms/javascript/common/enriching-events/scopes/index.mdx
mydea: yeah, I know what you mean, I was also not 100% sure there... 🤔

docs/platforms/javascript/common/enriching-events/scopes/index.mdx
```diff
@@ -8,35 +8,97 @@ information from the current scope. SDKs will typically automatically manage the
...
-and when it pops all modifications are reverted.
+the scope. When a scope is forked, it inherits all data from the parent scope.
+
+The default SDK integrations will fork scopes intelligently. For
+instance web framework integrations will fork scopes around your
+routes or request handlers.
+
+## How Scopes Work
+
+Scopes are basically a stack of data that is attached to events. When an event is captured, the SDK will merge the data from the current scope into the event. This allows you to attach data to events that is relevant to the context in which the event was captured.
```
vivianyentran, line 24:
```suggestion
Scopes are basically a stack of data that is attached to events. When an event is captured, the SDK will merge the data from the active scopes into the event. This allows you to attach data to events that is relevant to the context in which the event was captured.
```
Hope I'm not misunderstanding, but it sounds like data from all active scopes (global, isolation, current) is applied to the event, not just the current scope.

docs/platforms/javascript/common/enriching-events/scopes/index.mdx
```diff
@@ -8,35 +8,97 @@ information from the current scope. SDKs will typically automatically manage the
...
+  Sentry.captureException(new Error("my error"));
+});
+// this event will not have the tag:
+Sentry.captureException(new Error("my other error"));
+```
 
-When you call a global function such as <PlatformIdentifier name="capture-event" /> internally Sentry
-discovers the current hub and asks it to capture an event. Internally the hub will
-then merge the event with the topmost scope's data.
+You can access the curent scope via `Sentry.getCurrentScope()`, but usually you should use `Sentry.withScope()` to interact with local scopes instead.
```
vivianyentran, line 76:
```suggestion
You can access the current scope via `Sentry.getCurrentScope()`, but usually you should use `Sentry.withScope()` to interact with local scopes instead.
```

docs/platforms/javascript/common/enriching-events/scopes/index.mdx
```diff
@@ -8,35 +8,97 @@ information from the current scope. SDKs will typically automatically manage the
...
+
+## How Scope Data is Applied to Events
+
+Before an event (e.g. an error or transaction) is sent to Sentry, the currently active scopes are applied to it.
```
vivianyentran, line 80:
```suggestion
Before an event (like an error or transaction) is sent to Sentry, the currently active scopes are applied to it.
```

docs/platforms/javascript/common/enriching-events/scopes/index.mdx
```diff
@@ -8,35 +8,97 @@ information from the current scope. SDKs will typically automatically manage the
...
+
+The global scope is applied first, followed by the isolation scope, and finally the current scope. This means that any data set on the current scope will take precedence over data set on the isolation or global scope:
```
vivianyentran, line 82:
```suggestion
The global scope is applied first, followed by the isolation scope, and finally the current scope. This means that any data set on the current scope will take precedence over data set on the isolation and global scopes:
```

docs/platforms/javascript/common/enriching-events/scopes/index.mdx
```diff
@@ -8,35 +8,97 @@ information from the current scope. SDKs will typically automatically manage the
...
+  the browser the isolation scope is effectively global.
+</PlatformCategorySection>
 
-The default SDK integrations will push and pop scopes intelligently. For
-instance web framework integrations will create and destroy scopes around your
-routes or controllers.
+### Current Scope
 
-## How the Scope and Hub Work
+The current scope is the local scope that is currently active. In contrast to the isolation scope, which is usually only forked rarely, e.g. per request, the current scope may be forked more frequently and under the hood. It can be used to store data that should only be applied to specific events. In most cases, you should not access this scope directly, but use `Sentry.withScope` to create a new scope that is only active for a specific part of your code:
```
vivianyentran, line 62:
```suggestion
The current scope is the local scope that is currently active. Unlike the rarely-forked isolation scope, the current scope may be forked more frequently and under the hood. It can be used to store data that should only be applied to specific events. In most cases, you should not access this scope directly, but use `Sentry.withScope` to create a new scope that is only active for a specific part of your code:
```

vivianyentran (approved):
Added some copyedits

PR #9920, opened 2024-05-04
feat(js): Remove metrics configuration for v8

timfish:
For v8 of the JavaScript SDKs, metrics features are supported without any additional configuration.


vivianyentran (approved):
I think we should also update the version support note (<PlatformContent includePath="metrics/version-support-note" />) to clarify that in v8 and up, metrics works out of the box with no additional configuration required. 

We should also think about how to preserve info like the configuration setup for previous versions. There will probably be users who want to keep using v7 but still want to set up metrics.

timfish:
Needs workx

PR #9932, opened 2024-05-07
add instruction on node --require

andreiborza:


docs/platforms/javascript/guides/node/index.mdx
```diff
@@ -37,16 +37,28 @@ Sentry should be initialized as early in your app as possible. It is essential t
 
 Once this is done, Sentry's Node SDK captures unhandled exceptions as well as tracing data for your application.
 
+Therefore, we recommend to create a file named `instrumentation.js` that imports and initializes Sentry.
```
s1gr1d, line 40: Nitpick regarding line 36 and 46 in this file: Can you change the dash to an [em dash](https://www.thepunctuationguide.com/em-dash.html)?

platform-includes/getting-started-run-alternative/javascript.node.mdx
```diff
@@ -0,0 +1,17 @@
+<SignInNote />
+
+```javascript {tabTitle:CommonJS}
+require('./instrumentation.js')
+
```
lforst, line 5: I would argue that we should not honor this escape hatch an entire code block.

Considerations:
- People might do both --require AND the import
- It might be confusing
- We _really_ do not want people to do that unless absolutely necessary and a dedicated code block might be promoting it.

platform-includes/getting-started-config/javascript.node.mdx
```diff
@@ -11,10 +11,6 @@ Sentry.init({
   // We recommend adjusting this value in production
   tracesSampleRate: 1.0,
 });
-
-// http and other libraries are only instrumented
-// if required _after_ Sentry has been initialized
-const http = require("http");
 ```
 
 ```javascript {tabTitle:ESM}
```
lforst, line 16: Adding a filename here now would be nice!

andreiborza: Fixed

platform-includes/getting-started-run-alternative/javascript.node.mdx
mydea: I have a different opinion: We _definitely_ should to show this, because there will def. be people out there who do not have the (easy) capability to adjust the code their app is run in production. Lots of devs do not have direct access to their deployment infrastructure etc, making it _much_ harder for them to get started with using sentry. We can call out why the other solution is preferable and lead with that, but should still show how you can get it to run otherwise. There is really nothing wrong with the instrument-on-top approach for require.js.

mydea: actually, we need to keep this anyhow and add a note that this is _definitely_ needed for node <18 - no other way to run stuff there, I think?

mydea: > --require is supported since v1.6.0

Ah, sweet! Good to know 🚀 

> We were just thinking that people will skip over reading the leading text and just copy paste code snippets around, ending up with both --require and a top level import.

You need to read this anyhow, as the text instructs you to create another file etc anyhow - this is not as easy as copy-pasting a snippet into your app.js anyhow, I'd argue. If you don't read this, it'll not work anyhow.

I'd just like to argue that while _for us_ it is fine and easy to "just add a CLI flag when running the node app", there are _loads_ of devs out there for whom this is not _as_ trivial, and (as stated before) even more that simply can't do this themselves. Why should we not document how they can still use Sentry even if they can't easily adjust this themselves? I don't really see the upside of "hiding" this away? It's not even that problematic if this is double-initialized, it should work just fine even if somebody does that. And we could even guard against this, if we actually need it.

andreiborza: Yea you're right. I can definitely see some orgs disallowing passing flags to node, or like you said, people not even having the ability to change how things are run.

Will keep it, but maybe give it its own heading — as is, it's easy to not see that this is an alternative way of running it, wdyt?

docs/platforms/javascript/guides/node/index.mdx
```diff
@@ -33,20 +33,32 @@ yarn add @sentry/node
 
 ## Configure
 
-Sentry should be initialized as early in your app as possible. It is essential that you call `Sentry.init` before you require any other modules in your application - otherwise, auto-instrumentation of these modules will _not_ work.
+Sentry should be initialized as early in your app as possible. It is essential that you call `Sentry.init` before you require any other modules in your application — otherwise, auto-instrumentation of these modules will _not_ work.
 
 Once this is done, Sentry's Node SDK captures unhandled exceptions as well as tracing data for your application.
 
+Therefore, we recommend to create a file named `instrument.js` that imports and initializes Sentry.
```
vivianyentran, line 40:
```suggestion
We recommend creating a file named `instrument.js` that imports and initializes Sentry.
```

vivianyentran (approved):

Lms24 (commented):
Can we also update the SDK initialization section in the migration guide? https://docs.sentry.io/platforms/javascript/guides/node/migration/v7-to-v8/#updated-sdk-initialization

(feel free to do this in a follow-up PR, don't want to block this any longer than necessary)

s1gr1d (commented):
Thanks for adding this 🙌🏻 
Do not forget to also add this to koa, aws-lambda, azure-functions, connect, express and fastify.

docs/platforms/javascript/guides/node/index.mdx
```diff
@@ -33,20 +33,34 @@ yarn add @sentry/node
 
 ## Configure
 
-Sentry should be initialized as early in your app as possible. It is essential that you call `Sentry.init` before you require any other modules in your application - otherwise, auto-instrumentation of these modules will _not_ work.
+Sentry should be initialized as early in your app as possible. It is essential that you call `Sentry.init` before you require any other modules in your application — otherwise, auto-instrumentation of these modules will _not_ work.
```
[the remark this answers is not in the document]
andreiborza: Thanks, fixed

docs/platforms/javascript/guides/node/index.mdx
```diff
@@ -33,20 +33,34 @@ yarn add @sentry/node
...
 Once this is done, Sentry's Node SDK captures unhandled exceptions as well as tracing data for your application.
 
+We recommend creating a file named `instrument.js` that imports and initializes Sentry.
+
 <PlatformContent includePath="getting-started-config" />
 
 If you set a `tracesSampleRate`, performance instrumentation will automatically be enabled for you. See <PlatformLink to="/performance/instrumentation/automatic-instrumentation">Automatic Instrumentation</PlatformLink> to learn about all the things that the SDK automatically instruments for you.
 
-You can also manually capture performance data - see <PlatformLink to="/performance/instrumentation/custom-instrumentation">Custom Instrumentation</PlatformLink> for details.
+You can also manually capture performance data — see <PlatformLink to="/performance/instrumentation/custom-instrumentation">Custom Instrumentation</PlatformLink> for details.
```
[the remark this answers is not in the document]
andreiborza: Fixed

docs/platforms/javascript/guides/node/run/index.mdx
```diff
@@ -0,0 +1,28 @@
+---
+title: Run Methods
```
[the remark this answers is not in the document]
andreiborza: Yep, renamed it.

docs/platforms/javascript/guides/node/run/run-as-toplevel-import.mdx
```diff
@@ -0,0 +1,8 @@
+---
+title: Run as toplevel import
```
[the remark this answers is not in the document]
andreiborza: Thank you :) fixed

s1gr1d (approved):

andreiborza:
> Thanks for adding this 🙌🏻 Do not forget to also add this to koa, aws-lambda, azure-functions, connect, express and fastify.

Added the others except `aws-lambda` and `azure-functions` because I'm unsure if we should advocate an extra file there as we discussed in the office the other day.

If you could have another look please :)

PR #9936, opened 2024-05-07
fix: Changelog Images

HazAT:
next/image doesn't work with the redirects we do and static images we serve from Google
image


a-hariti:
maybe adding some `// eslint-disable-next-line @next/next/no-img-element` makes sense here to get rid of the eslint warnings? @HazAT

PR #9967, opened 2024-05-08
chore(api): Created Nav Link for API Docs

iamrajjoshi:
Our API Docs were a little hidden and had no visibility from the home page. I added the Navlink to match how our docs were before:
image

image

I don't think we should group our API docs with Product docs as they serve two different purposes.


isabellaenriquez (approved):
🙏🏽  but wait for docs team to approve as well

PR #10002, opened 2024-05-13
feat(node): Add fallback guide

s1gr1d:
Adds possibility for fallback guide which is useful for e.g. node-based libraries such as express. The `PlatformContent` component will not fallback to the javascript snippets but the node snippets (as node is just a guide in the JS platform). Currently, this is solved by using symlinks to the node code snippets to keep the snippets the same as node.


mydea (approved):
I like it! 🚀

stephanie-anderson (approved):
Niiiiice!! Thank you!

PR #10017, opened 2024-05-14
Don't log for failed code block formatting

lforst:
To clean up vercel build logs


andreiborza (approved):
LGTM

a-hariti:
I think we can make it more usable by adding dev mode only logs with the actual file name that has the error
wyt @lforst ?

lforst:
@a-hariti I think you can make it out pretty well already what is causing the errors. The thing is not every code block should be formatted and this is just noise.

PR #10020, opened 2024-05-14
Fix order of init and imports for react onboarding

lforst:
Ref: https://github.com/getsentry/sentry-javascript/issues/12018#issuecomment-2110194883


platform-includes/getting-started-config/javascript.react.mdx
```diff
@@ -35,6 +38,10 @@ Sentry.init({
   replaysOnErrorSampleRate: 1.0,
 });
 
+// Make sure Sentry.init() is called before importing any of your other code
+import App from "./App";
+import { createRoot } from "react-dom/client";
```
Slessi, line 43: Aren't all imports are evalulated first, regardless where they appear in the file? Thought you would need to make a separate file for this, and import it first.

lforst: It depends a bit on what the bundlers will do, but I think you are right. Vite seems to hoist the import statements up. I will update. Thanks!

platform-includes/getting-started-config/javascript.react.mdx
```diff
@@ -1,39 +1,53 @@
+It is important to initialize Sentry as early as possible in your application. We recommend putting the Sentry initialization code into its own file and including that file as the first import in your application-entrypoint:
```
vivianyentran, line 1:
```suggestion
It is important to initialize Sentry as early as possible in your application. We recommend putting the Sentry initialization code into its own file and including that file as the first import in your application entry point:
```

vivianyentran (approved):

mydea (approved):

PR #10035, opened 2024-05-15
fix typo

rodolfoBee:
fix sourceMap option for vite plugin: https://vitejs.dev/config/build-options.html#build-sourcemap

it is saying `sourcemaps`, while the correct option in the Vite documentation (and other sentry pages) is `sourcemap`


vivianyentran (approved):
Thanks for fixing!

PR #10071, opened 2024-05-17
docs(lambda): Add AWS Lambda layer for v7 node sdk

andreiborza:
Ideally we would want to have another dropdown for v7 that automatically spits out the latest ARN for v7, but this is a change that needs to be carefully evaluated in `craft` and `sentry-release-registry`. For now we'll have to live with a hard-coded ARN for v7.

ref: https://github.com/getsentry/sentry-javascript/issues/12098


Lms24 (approved):
Thanks! It's a good solution for now!

PR #10072, opened 2024-05-17
feat(Unity): Auto Performance

bitsandfoxes:
Because it's missing.


docs/platforms/unity/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -0,0 +1,25 @@
...
+title: Automatic Instrumentation
+description: "Learn what transactions are captured after tracing is enabled."
+sidebar_order: 20
+---
+
+## Automatically Captured Transactions
+
+The Unity SDK automatically creates transactions for the game's startup and scene loading. Additionally, if opted-in, it generates spans for all `Awake` calls during these operations.
+
+You can manage these settings through the Editor Config Window or <Link to="/platforms/unity/configuration/options/#programmatic-configuration">programmatically</Link>.
```
vivianyentran, line 11:
```suggestion
The Unity SDK automatically creates transactions for the game's startup and scene loading. 

If you want the SDK to also generate spans for all `Awake` calls during startup and scene operations, you can enable the following settings through the Editor Config Window or <Link to="/platforms/unity/configuration/options/#programmatic-configuration">programmatically</Link>.
```

Clarify if the settings below are to generate spans for Awake calls or just for scene loading and startup. I assume it's for awake calls since you said startup & scene loading transactions are automatically captured, but it might be confusing.

docs/platforms/unity/performance/instrumentation/automatic-instrumentation.mdx
```diff
@@ -0,0 +1,25 @@
...
+
+```csharp
+public override void Configure(SentryUnityOptions options)
+{
+    options.AutoSceneLoadTraces = true;
+    options.AutoStartupTraces = true;
+}
+```
+
+Once automatic instrumentation is configured, the Unity SDK will generate transactions as shown in the image below.
```
vivianyentran, line 23:
```suggestion
After you've configured your instrumentation, the Unity SDK will generate traces as shown in the image below.
```
Maybe with the am3 plans, we should start moving away from a transaction-based model.

vivianyentran (approved):
Added some suggestions, but good catch that this was missing!

Issue #10041, opened 2024-05-15
Azure Functions Documentation


kerenkhatiwada:
### Core or SDK?

Platform/SDK

### Which part? Which one?

https://docs.sentry.io/platforms/dotnet/guides/azure-functions-worker/#configure

### Description

Reproted by user:"Not really knowing what I'm doing I tried to add Sentry to my basic Azure Functions app. The documentation suggested I use ConfigureFunctionsWorkerDefaults but there was already a call on ConfigureFunctionsWebApplication so I added the recommended the ConfigureFunctionsWorkerDefaults call before it. This is a bad idea, it compiles and all the functions appear to be callable and return "OK" results but with no data! This is because they are not actually called! Once you figure it out the fix is easy, just use ConfigureFunctionsWebApplication INSTEAD of ConfigureFunctionsWorkerDefaults. So I don't have a problem, just wanted to let someone know."

### Suggested Solution

Clarify in the documentation, that users can choose between ConfigureFunctionsWorkerDefaults and ConfigureFunctionsWebApplication, depending on their specific app configuration.

david-maw:
I  reported this, and as I said the workaround, once I figured it out, was easy. If the incorrect code I chose (call `ConfigureFunctionsWorkerDefaults `then `ConfigureFunctionsWebApplication`) had failed to compile, or run with an error, I'd have probably figured it out earlier, but because it seemed to work, but actually didn't, diagnosis took more work. As far as I can tell in the failed configuration my function code was never invoked, the infrastructure just returned status code OK for all calls with no attached data and my caller just quietly did the wrong thing.

All I would have needed is a hint to use either `ConfigureFunctionsWorkerDefaults` or  `ConfigureFunctionsWebApplication`, whichever was already there.

bitsandfoxes:
Thanks for letting us know. I'll update the docs.

david-maw:
You're welcome, I'm relatively new to Azure Functions but as far as I can see Microsoft now seem to prefer to use the newer `ConfigureFunctionsWebApplication`  function in order to enable some ASP.NET functionality in Azure Functions but still support `ConfigureFunctionsWorkerDefaults`, presumably for existing code. In particular,  `ConfigureFunctionsWebApplication` is what they emit if you create a default Azure Functions project in Visual Studio (which is how I stumbled over this).

bitsandfoxes:
[...]

PR #10089, opened 2024-05-21
fix(.NET): Update Azure Functions

bitsandfoxes:
Fixes https://github.com/getsentry/sentry-docs/issues/10041

If you (as a user) are using the ASP.NET integration use the `ConfigureFunctionsWebApplication` call instead of `ConfigureFunctionsWorkerDefaults`. It's an either/or situation, but not both.

### DIFF
--- a/docs/platforms/dotnet/guides/azure-functions-worker/index.mdx
+++ b/docs/platforms/dotnet/guides/azure-functions-worker/index.mdx
@@ -23,7 +23,7 @@ This package extends [Sentry.Extensions.Logging](/platforms/dotnet/guides/extens
 
 ## Configure
 
-Sentry integration with Azure Functions is done by calling `.UseSentry()` and specifying the options. E.g.:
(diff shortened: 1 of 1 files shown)


docs/platforms/dotnet/guides/azure-functions-worker/index.mdx
```diff
@@ -23,10 +23,16 @@ This package extends [Sentry.Extensions.Logging](/platforms/dotnet/guides/extens
 
 ## Configure
 
-Sentry integration with Azure Functions (when using the Isolated Worker Process execution mode) is done by calling `.UseSentry()` and specifying the options. E.g.:
+Sentry integration with Azure Functions is done by calling `.UseSentry()` and specifying the options. E.g.:
```
lizokm, line 26:
```suggestion
Sentry integration with Azure Functions is done by calling `.UseSentry()` and specifying the options, for example:
```

lizokm (approved):
Looks good, thanks for adding!

PR #10092, opened 2024-05-21
fix(express): Ensure we document error handler positioning

mydea:
Fixes https://github.com/getsentry/sentry-javascript/issues/12113

### DIFF
--- a/platform-includes/getting-started-add-top-level-import/javascript.express.mdx
+++ b/platform-includes/getting-started-add-top-level-import/javascript.express.mdx
@@ -9,7 +9,7 @@ const app = express();
 // Add your routes, etc.
 
 // Add this after all routes,
-// but before any and other middlewares are defined
+// but before any and other error-handling middlewares are defined
 Sentry.setupExpressErrorHandler(app);
 
 app.listen(3000);
@@ -26,7 +26,7 @@ const app = express();
 // Add your routes, etc.
 
 // Add this after all routes,
-// but before any and other middlewares are defined
+// but before any and other error-handling middlewares are defined
 Sentry.setupExpressErrorHandler(app);
 
 app.listen(3000);
--- a/platform-includes/getting-started-add-top-level-import/javascript.express.mdx
+++ b/platform-includes/getting-started-add-top-level-import/javascript.express.mdx
@@ -8,7 +8,8 @@ const app = express();
 
 // Add your routes, etc.
 
-// Add this after all routes and other middlewares are defined
+// Add this after all routes,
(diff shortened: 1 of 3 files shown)


platform-includes/getting-started-add-top-level-import/javascript.express.mdx
```diff
@@ -8,7 +8,8 @@ const app = express();
 
 // Add your routes, etc.
 
-// Add this after all routes and other middlewares are defined
+// Add this after all routes,
+// but before any and other middlewares are defined
```
Lms24, line 12: m: I think "middlewares" is a bit misleading as I'm fairly sure people also define middleware before they register their routes, no?

```suggestion
// but before any error handlers are defined
```

vivianyentran: Would "error-handling middleware" work?

mydea: I'll go with this 👍

AbhiPrasad (approved):
Because I just got bit with this lol

pckilgore:
How would you recommend instrumenting more complicated router setups given the "before error middleware, but after routes" constraint?

For example:

```js
const app = express()
app.use(parserMiddlewares)
app.use(authorizationMiddleware) 

const apiRouter = Router()         // serves JSON for AJAX
apiRouter.all("/resource")
....
// < --- Position 1
apiRouter.use(errorMiddleware).     // serves JSON error responses.

app.use("/api", apiRouter) 

app.use(siteRouter)                 // serves HTML templates
// < --- Position 1
app.use(errorMiddleware)            // serves text/html error responses
```

Call `Sentry.setupExpressErrorHandler(app);` twice, once in each position?

mydea:
Hey,

in this case, you need to add `setupExpressErrorHandler(app)` before each error middleware:

```js
const app = express()
app.use(parserMiddlewares)
app.use(authorizationMiddleware) 

const apiRouter = Router()         // serves JSON for AJAX
apiRouter.all("/resource")
....
// < --- Position 1
Sentry.setupExpressErrorHandler(apiRouter);
apiRouter.use(jsonErrorMiddleware).     // serves JSON error responses.

app.use("/api", apiRouter) 

app.use(siteRouter)                 // serves HTML templates
// < --- Position 1
Sentry.setupExpressErrorHandler(app);
app.use(htmlErrorMiddleware)            // serves text/html error responses
```

I think this should work! Really we just need to get the error before it is handled by some other middleware 😅

PR #10093, opened 2024-05-21
feat(node): Add docs for `NODE_OPTIONS` esm config

mydea:
This was still missing from our docs.

### DIFF
--- a/docs/platforms/javascript/common/install/esm.mdx
+++ b/docs/platforms/javascript/common/install/esm.mdx
@@ -40,7 +40,7 @@ Adjust the Node.js call for your application to use the [--import](https://nodej
 node --import ./instrument.mjs app.mjs
 ```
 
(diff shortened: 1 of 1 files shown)


andreiborza (approved):

docs/platforms/javascript/common/install/esm.mdx
```diff
@@ -40,4 +40,10 @@ Adjust the Node.js call for your application to use the [--import](https://nodej
 node --import ./instrument.mjs app.mjs
 ```
 
+Sometimes, you cannot directly pass the `--import` flag to the Node.js binary. In such cases, you can use the `NODE_OPTIONS` environment variable to pass it, like this:
```
lforst, line 43:
```suggestion
If it is not possible for you to pass the `--import` flag to the Node.js binary, you can alternatively use the `NODE_OPTIONS` environment variable as follows:
```

lforst (approved):

Lms24 (approved):

PR #10095, opened 2024-05-21
feat(js): Improve JS custom instrumentation docs

mydea:
This rewrites the JS (node & browser) custom performance instrumentation docs to have much more detail than before.

### DIFF
--- a/docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
+++ b/docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
@@ -32,6 +32,40 @@ A key constraint for active spans is that they can only be made active inside of
 
 In places where you are not able to wrap executing code in a callback (e.g. when working with hooks or similar) you have to work with inactive spans, and can combine this with [withActiveSpan](#with-active-span) to manually associate child spans with the correct parent span.
 
+<PlatformCategorySection supported={['browser']}>
+
+## Span Hierarchy in the Browser
+
+In browser environments, it is not possible to reliably keep track of the active span across asynchronous boundaries. This means that if multiple asynchronous operations are started in parallel, it is not possible to determine which span is the parent of which child span. Imagine the following example:
+
+```javascript
+Sentry.startSpan({ name: "span 1" }, async () => {
+  await fetch("https://example.com/1");
+  await fetch("https://example.com/2");
+  await fetch("https://example.com/3");
+});
+
+Sentry.startSpan({ name: "span 2" }, async () => {
+  await fetch("https://example.com/4");
+  await fetch("https://example.com/5");
+  await fetch("https://example.com/6");
+});
+```
+
+In the browser, there would be no way to know that `span 1` is only active inside of its callback, while `span 2` is active in the other callback. Because of this, in realitiy _all_ fetch spans would become children of `span 2`. This is misleading and confusing, which is why by default in the browser, **all spans will become children of the root span** (which is usually the pageload or navigation span). This means that you will always have a flat hierarchy of spans.
+
+This is a tradeoff that we have made to ensure that the data that is captured is accurate and reliable. If you need to capture a more complex hierarchy of spans, you can opt-out of this behavior by setting `parentSpanIsAlwaysRootSpan: false`:
+
+```javascript
+Sentry.init({
+  parentSpanIsAlwaysRootSpan: false,
+});
+```
+
+This will revert to use the full hierarchy behavior, where spans are children of the currently active span. However, this may lead to incorrect data in the case of multiple parallel asynchronous operations - it is up to you to ensure there are not multiple parallel asynchronous operations that start spans in this case.
+
+</PlatformCategorySection>
+
 ## Span Starting Options
 
 The following options can be used for all span starting functions:
--- a/docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
+++ b/docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
@@ -36,7 +36,9 @@ In places where you are not able to wrap executing code in a callback (e.g. when
 
 ## Span Hierarchy in the Browser
 
-In browser environments, it is not possible to reliably keep track of the active span across asynchronous boundaries. This means that if multiple asynchronous operations are started in parallel, it is not possible to determine which span is the parent of which child span. Imagine the following example:
+In browser environments, spans are by default collected in a flat hierarchy where every span is the direct child of the root span. You can opt into a more fine grained hierarchy but there are trade-offs. 
+
+The key reason for keeping a flat hierarchy is because in browsers it's impossible to reliably keep track of the active span across asynchronous boundaries. This means that if multiple asynchronous operations are started in parallel, it is not possible to determine which span is the parent of which child span. Imagine the following example:
 
 ```javascript
 Sentry.startSpan({ name: "span 1" }, async () => {
@@ -52,7 +54,7 @@ Sentry.startSpan({ name: "span 2" }, async () => {
 });
 ```
 
-In the browser, there would be no way to know that `span 1` is only active inside of its callback, while `span 2` is active in the other callback. Because of this, in realitiy _all_ fetch spans would become children of `span 2`. This is misleading and confusing, which is why by default in the browser, **all spans will become children of the root span** (which is usually the pageload or navigation span). This means that you will always have a flat hierarchy of spans.
+In the browser, there would be no way to know that `span 1` is only active inside of its callback, while `span 2` is active in the other callback. Because of this, in realitiy, _all_ fetch spans would become children of `span 2`. This is misleading and confusing, which is why by default in the browser, **all spans will become children of the root span** (which is usually the pageload or navigation span). This means that you will always have a flat hierarchy of spans.
 
 This is a tradeoff that we have made to ensure that the data that is captured is accurate and reliable. If you need to capture a more complex hierarchy of spans, you can opt-out of this behavior by setting `parentSpanIsAlwaysRootSpan: false`:
 
@@ -62,7 +64,7 @@ Sentry.init({
 });
 ```
 
-This will revert to use the full hierarchy behavior, where spans are children of the currently active span. However, this may lead to incorrect data in the case of multiple parallel asynchronous operations - it is up to you to ensure there are not multiple parallel asynchronous operations that start spans in this case.
+This will revert to use the full hierarchy behavior, where spans are children of the currently active span. However, this may lead to incorrect data in the case of multiple parallel asynchronous operations - it is up to you to ensure there are no multiple parallel asynchronous operations that start spans in this case.
 
 </PlatformCategorySection>
 
@@ -89,13 +91,13 @@ For most scenarios, we recommend to start active spans with `Sentry.startSpan()`
 
 ## Starting an Active Span with Manual End (`startSpanManual`)
 
-Sometimes, you do not want the span to be ended automatically when the callback is done. In this case, you can use `Sentry.startSpanManual()`. This will start a new span that is active in the provided callback, but will not automatically end the span when the callback is done. You have to manually end the span by calling `span.end()`.
+Sometimes, you do not want the span to be ended automatically when the callback is done. In this case, you can use `Sentry.startSpanManual()`. This will start a new span that is active in the provided callback, but will not be automatically ended when the callback is done. You have to manually end the span by calling `span.end()`.
 
 <PlatformContent includePath="performance/start-span-manual" />
 
 ## Starting Inactive Spans (`startInactiveSpan`)
 
-To add spans that aren't active, you can create independent spans. This is useful when you have work that is grouped together under a single parent span, but is independent from the current active span. However, in most cases you'll want to create and use the [startSpan](#starting-an-active-span-startspan) API from above.
+To add spans that aren't active, you can create independent spans. This is useful when you have work that is grouped together under a single parent span, but is independent from the currently active span. However, in most cases you'll want to create and use the [startSpan](#starting-an-active-span-startspan) API from above.
 
 <PlatformContent includePath="performance/start-inactive-span" />
 
--- a/docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
+++ b/docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
@@ -16,21 +16,119 @@ To get started, import the SDK.
 
 <PlatformContent includePath="enriching-events/import" />
 
-<PlatformContent includePath="performance/span-api-version" />
+There are three key APIs for creating spans:
 
-## Start Span
+- [startSpan](#starting-an-active-span-startspan): Creates a new span that is active, and which is automatically ended.
+- [startSpanManual](#starting-an-active-span-with-manual-end-startspanmanual): Creates a new span that is active, which has to be ended manually.
+- [startInactiveSpan](#starting-inactive-spans-startinactivespan): Creates a new span that is inactive, which has to be ended manually.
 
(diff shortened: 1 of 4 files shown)


docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -16,21 +16,119 @@ To get started, import the SDK.
 
 <PlatformContent includePath="enriching-events/import" />
 
-<PlatformContent includePath="performance/span-api-version" />
+There are three key APIs for creating spans:
```
Lms24, line 19: l: Not sure if we call these APIs anywhere else 🤔 

```suggestion
There are three key functions for creating spans:
```

docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -16,21 +16,119 @@ To get started, import the SDK.
...
+
+In contrast, **inactive spans** will never have children automatically associated with them. This is useful if you do not care about capturing child activity.
+
+A key constraint for active spans is that they can only be made active inside of a callback. This constraint exists because otherwise it becomes impossible to associate spans with the correct parent span when working with asynchronous code.
+
+In places where you are not able to wrap executing code in a callback (e.g. when working with hooks or similar) you have to work with inactive spans, and can combine this with [withActiveSpan](#with-active-span) to manually associate child spans with the correct parent span.
+
+## Span Starting Options
+
+The following options can be used for all span starting methods:
```
Lms24, line 37: l: I don't particularly care if we go with API/function/method but we should remain consistent

```suggestion
The following options can be used for all span starting functions:
```

docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -16,21 +16,119 @@ To get started, import the SDK.
 
 <PlatformContent includePath="enriching-events/import" />
 
-<PlatformContent includePath="performance/span-api-version" />
+There are three key APIs for creating spans:
 
-## Start Span
+- [startSpan](#starting-an-active-span-startspan): Creates a new span that is active, and which is automatically ended.
```
Lms24, line 21: l/q: Given we recommend `startSpan` below - should we already hint this here?

```suggestion
- [startSpan](#starting-an-active-span-startspan): Creates a new span that is active, and which is automatically ended. You'll likely want to use this function.
```

docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -16,21 +16,119 @@ To get started, import the SDK.
...
-## Start Inactive Spans
+## Starting an Active Span with Manual End (`startSpanManual`)
+
+Sometimes, you do not want the span to be ended automatically when the callback is done. In this case, you can use `Sentry.startSpanManual()`. This will start a new span that is active in the provided callback, but will not automatically end the span when the callback is done. You have to manually end the span by calling `span.end()`.
+
+<PlatformContent includePath="performance/start-span-manual" />
+
+## Starting Inactive Spans (`startInactiveSpan`)
 
 To add spans that aren't active, you can create independent spans. This is useful for when you have work that is grouped together under a single parent span, but is independent from the current active span. However, in most cases you'll want to create and use the start span API from above.
```
Lms24, line 64: l:

```suggestion
To add spans that aren't active, you can create independent spans. This is useful when you have work that is grouped together under a single parent span, but is independent from the current active span. However, in most cases you'll want to create and use the [startSpan](#starting-an-active-span-startspan) API from above.
```

Lms24 (approved):
Looks good to me! It's more information than before and a lot more cleaner what "active" implies. Just had a couple of nits - feel free to ignore any of them

docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -16,21 +16,153 @@ To get started, import the SDK.
...
+
+This is a tradeoff that we have made to ensure that the data that is captured is accurate and reliable. If you need to capture a more complex hierarchy of spans, you can opt-out of this behavior by setting `parentSpanIsAlwaysRootSpan: false`:
+
+```javascript
+Sentry.init({
+  parentSpanIsAlwaysRootSpan: false,
+});
+```
+
+This will revert to use the full hierarchy behavior, where spans are children of the currently active span. However, this may lead to incorrect data in the case of multiple parallel asynchronous operations - it is up to you to ensure there are not multiple parallel asynchronous operations that start spans in this case.
```
andreiborza, line 65:
```suggestion
This will revert to use the full hierarchy behavior, where spans are children of the currently active span. However, this may lead to incorrect data in the case of multiple parallel asynchronous operations - it is up to you to ensure there are no multiple parallel asynchronous operations that start spans in this case.
```

docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -16,21 +16,153 @@ To get started, import the SDK.
...
+});
+
+Sentry.startSpan({ name: "span 2" }, async () => {
+  await fetch("https://example.com/4");
+  await fetch("https://example.com/5");
+  await fetch("https://example.com/6");
+});
+```
+
+In the browser, there would be no way to know that `span 1` is only active inside of its callback, while `span 2` is active in the other callback. Because of this, in realitiy _all_ fetch spans would become children of `span 2`. This is misleading and confusing, which is why by default in the browser, **all spans will become children of the root span** (which is usually the pageload or navigation span). This means that you will always have a flat hierarchy of spans.
```
andreiborza, line 55:
```suggestion
In the browser, there would be no way to know that `span 1` is only active inside of its callback, while `span 2` is active in the other callback. Because of this, in realitiy, _all_ fetch spans would become children of `span 2`. This is misleading and confusing, which is why by default in the browser, **all spans will become children of the root span** (which is usually the pageload or navigation span). This means that you will always have a flat hierarchy of spans.
```

docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -16,21 +16,153 @@ To get started, import the SDK.
...
+| Option             | Type                        | Description                                                         |
+| ------------------ | --------------------------- | ------------------------------------------------------------------- |
+| `name`             | `string`                    | The name of the span.                                               |
+| `op`               | `string`                    | The operation of the span.                                          |
+| `startTime`        | `number`                    | The start time of the span.                                         |
+| `attributes`       | `Record<string, Primitive>` | Attributes to attach to the span.                                   |
+| `onlyIfParent`     | `boolean`                   | If true, ignore the span if there is no active parent span.         |
+| `forceTransaction` | `boolean`                   | If true, ensure this span shows up as transaction in the Sentry UI. |
+
+Only `name` is required, all other options are optional.
```
andreiborza, line 82: l: I think this would be better encoded in the table as another column for each param.

docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -16,21 +16,153 @@ To get started, import the SDK.
...
+## Starting an Active Span (`startSpan`)
+
+For most scenarios, we recommend to start active spans with `Sentry.startSpan()`. This will start a new span that is active in the provided callback, and will automatically end the span when the callback is done. The callback can be synchronuous, or asynchronous (a promise). In the case of an asynchronous callback, the span will be ended when the promise is resolved or rejected. If the provided callback errors or rejects, the span will be marked as failed.
 
 <PlatformContent includePath="performance/start-span" />
 
-## Start Inactive Spans
+## Starting an Active Span with Manual End (`startSpanManual`)
+
+Sometimes, you do not want the span to be ended automatically when the callback is done. In this case, you can use `Sentry.startSpanManual()`. This will start a new span that is active in the provided callback, but will not automatically end the span when the callback is done. You have to manually end the span by calling `span.end()`.
```
andreiborza, line 92: l: how about having an example here instead of just `Sometimes`?

question: what happens if user's never call `span.end()`?

andreiborza, line 92:
```suggestion
Sometimes, you do not want the span to be ended automatically when the callback is done. In this case, you can use `Sentry.startSpanManual()`. This will start a new span that is active in the provided callback, but will not be automatically ended when the callback is done. You have to manually end the span by calling `span.end()`.
```

docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -16,21 +16,153 @@ To get started, import the SDK.
...
+
+<PlatformContent includePath="performance/start-span-manual" />
 
-To add spans that aren't active, you can create independent spans. This is useful for when you have work that is grouped together under a single parent span, but is independent from the current active span. However, in most cases you'll want to create and use the start span API from above.
+## Starting Inactive Spans (`startInactiveSpan`)
+
+To add spans that aren't active, you can create independent spans. This is useful when you have work that is grouped together under a single parent span, but is independent from the current active span. However, in most cases you'll want to create and use the [startSpan](#starting-an-active-span-startspan) API from above.
```
andreiborza, line 98:
```suggestion
To add spans that aren't active, you can create independent spans. This is useful when you have work that is grouped together under a single parent span, but is independent from the currently active span. However, in most cases you'll want to create and use the [startSpan](#starting-an-active-span-startspan) API from above.
```

andreiborza (approved):

docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -16,21 +16,153 @@ To get started, import the SDK.
...
+
+A key constraint for active spans is that they can only be made active inside of a callback. This constraint exists because otherwise it becomes impossible to associate spans with the correct parent span when working with asynchronous code.
+
+In places where you are not able to wrap executing code in a callback (e.g. when working with hooks or similar) you have to work with inactive spans, and can combine this with [withActiveSpan](#with-active-span) to manually associate child spans with the correct parent span.
+
+<PlatformCategorySection supported={['browser']}>
+
+## Span Hierarchy in the Browser
+
+In browser environments, it is not possible to reliably keep track of the active span across asynchronous boundaries. This means that if multiple asynchronous operations are started in parallel, it is not possible to determine which span is the parent of which child span. Imagine the following example:
```
Lms24, line 39: l/m: I'd suggest addinga TLDR-like intro sentence to this section. Right now, we start with the problem and tell folks about the consequence  4 paragraphs in.

I was thinking of something along these lines. Feel free to adjust as you see fit.

```suggestion
## Span Hierarchy in the Browser

In browser environments, spans are by default collected in a flat hierarchy where every span is the direct child of the root span. You can opt into a more fine grained hierarchy but there are trade-offs. 

The key reason for keeping a flat hierarchy is because in browsers it's impossible to reliably keep track of the active span across asynchronous boundaries. This means that if multiple asynchronous operations are started in parallel, it is not possible to determine which span is the parent of which child span. Imagine the following example:
```

docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
mydea: this makes the table become a bit unwieldy, as we don't have a lot of horizontal space 😬

docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
mydea: there is an example below, I guess (with the middleware)?

--> if end is never called, the span is never sent (and also any child spans are never sent)

docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
andreiborza: Ah, no worries then.

andreiborza (approved):

docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -16,21 +16,155 @@ To get started, import the SDK.
...
+
+A key constraint for active spans is that they can only be made active inside of a callback. This constraint exists because otherwise it becomes impossible to associate spans with the correct parent span when working with asynchronous code.
+
+In places where you are not able to wrap executing code in a callback (e.g. when working with hooks or similar) you have to work with inactive spans, and can combine this with [withActiveSpan](#with-active-span) to manually associate child spans with the correct parent span.
+
+<PlatformCategorySection supported={['browser']}>
+
+## Span Hierarchy in the Browser
+
+In browser environments, spans are by default collected in a flat hierarchy where every span is the direct child of the root span. You can opt into a more fine grained hierarchy but there are trade-offs.
```
AbhiPrasad, line 39:
```suggestion
In browser environments, spans are by default collected in a flat hierarchy where every span is the direct child of the root span. You can opt into a more fine-grained hierarchy but there are trade-offs. 
```

docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -16,21 +16,155 @@ To get started, import the SDK.
...
+});
+
+Sentry.startSpan({ name: "span 2" }, async () => {
+  await fetch("https://example.com/4");
+  await fetch("https://example.com/5");
+  await fetch("https://example.com/6");
+});
+```
+
+In the browser, there would be no way to know that `span 1` is only active inside of its callback, while `span 2` is active in the other callback. Because of this, in realitiy, _all_ fetch spans would become children of `span 2`. This is misleading and confusing, which is why by default in the browser, **all spans will become children of the root span** (which is usually the pageload or navigation span). This means that you will always have a flat hierarchy of spans.
```
AbhiPrasad, line 57:
```suggestion
In the browser, there would be no way to know that `span 1` is only active inside of its callback, while `span 2` is active in the other callback. Because of this, in reality, _all_ fetch spans would become children of `span 2`. This is misleading and confusing, which is why by default in the browser, **all spans will become children of the root span** (which is usually the pageload or navigation span). This means that you will always have a flat hierarchy of spans.
```

docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation.mdx
```diff
@@ -16,21 +16,155 @@ To get started, import the SDK.
...
+| `startTime`        | `number`                    | The start time of the span.                                         |
+| `attributes`       | `Record<string, Primitive>` | Attributes to attach to the span.                                   |
+| `onlyIfParent`     | `boolean`                   | If true, ignore the span if there is no active parent span.         |
+| `forceTransaction` | `boolean`                   | If true, ensure this span shows up as transaction in the Sentry UI. |
+
+Only `name` is required, all other options are optional.
+
+## Starting an Active Span (`startSpan`)
+
+For most scenarios, we recommend to start active spans with `Sentry.startSpan()`. This will start a new span that is active in the provided callback, and will automatically end the span when the callback is done. The callback can be synchronuous, or asynchronous (a promise). In the case of an asynchronous callback, the span will be ended when the promise is resolved or rejected. If the provided callback errors or rejects, the span will be marked as failed.
```
AbhiPrasad, line 88:
```suggestion
For most scenarios, we recommend to start active spans with `Sentry.startSpan()`. This will start a new span that is active in the provided callback, and will automatically end the span when the callback is done. The callback can be synchronous, or asynchronous (a promise). In the case of an asynchronous callback, the span will be ended when the promise is resolved or rejected. If the provided callback errors or rejects, the span will be marked as failed.
```

PR #10123, opened 2024-05-23
ref(javascript): Streamline queue instrumentation page

Lms24:
- Only show page on server-side guides
- remove imports from `@sentry/core` - we should never tell people to import from it
  as it's only a transitive dependency from their SDK. Here, we can make use of the 
  `Sentry.` namespace import that people are familiar with from other docs. This also
  saves us the hassle of having to customize the imported package for each guide :)
- just minor formatting (prettier) and code snippet fixes

### DIFF
--- a/docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation/queues-module.mdx
+++ b/docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation/queues-module.mdx
@@ -87,7 +87,7 @@ Use `Sentry.continueTrace` to connect your consumer spans to their associated pr
 
 <Note>
 
-Your `queue.process` span must exist as a child within a parent span in order to be recognized as a consumer span. If you don't already have a parent for the consumer span, you can wrap the consumer span with a parent span by using `Sentry.startSpan`.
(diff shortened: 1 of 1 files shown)


andreiborza (approved):

docs/platforms/javascript/common/performance/instrumentation/custom-instrumentation/queues-module.mdx
```diff
@@ -24,56 +41,57 @@ Your `queue.publish` span must exist as a child within a parent span in order to
...
+| `messaging.message.body.size`        | number | Size of the message body in bytes                                   |
+| `messaging.message.retry.count `     | number | The number of times a message was attempted to be processed         |
 | `messaging.message.receive.latency ` | number | The time in milliseconds that a message awaited processing in queue |
 
 Use `Sentry.continueTrace` to connect your consumer spans to their associated producer spans, and `setStatus` to mark the trace of your message as success or failed.
 
-
 <Note>
 
 Your `queue.process` span must exist as a child within a parent span in order to be recognized as a consumer span. If you don't already have a parent consumer span, you can start a new one using `Sentry.startSpan`.
```
mydea, line 90:
```suggestion
Your `queue.process` span must exist as a child within a parent span in order to be recognized as a consumer span. If you don't already have a parent for the consumer span, you can wrap the consumer span with a parent span by using `Sentry.startSpan`.
```

PR #10130, opened 2024-05-23
feat: onboarding options on getting started pages

a-hariti:
Usage docs [here](https://sentry-docs-git-onboarding-options.sentry.dev/contributing/pages/components/#onboarding-options)


docs/platforms/php/index.mdx
```diff
@@ -25,28 +25,73 @@ Sentry captures data by using an SDK within your application’s runtime. These
 
 Install the SDK using [Composer](https://getcomposer.org/).
 
+
+<OnboardingOptionsButtons
```
stephanie-anderson, line 29: Would avoid duplicate plurals in the component name:
`OnboardingOptionsButtons` -> `OnboardingOptionButtons`

docs/platforms/php/index.mdx
```diff
@@ -25,28 +25,73 @@ Sentry captures data by using an SDK within your application’s runtime. These
...
+      id: 'error-monitoring',
+      name: 'Error Monitoring',
+      disabled: true,
+      tooltip: "Let's admit it, we all have errors.",
+    },
+    {
+      id: 'performance',
+      name: 'Performance Monitoring',
+      tooltip:
+        'Automatic performance issue detection across services and context on who is impacted,' +
```
stephanie-anderson, line 41: The tooltips / names will repeat on every page we use the component. Can we move that into the component itself, so tooltips etc must only be maintained in a single place? Also stuff like the dependency is always the same across SDKs.

docs/platforms/php/index.mdx
```diff
@@ -25,28 +25,73 @@ Sentry captures data by using an SDK within your application’s runtime. These
...
-After you’ve completed setting up a project in Sentry, Sentry will give you a value which we call a DSN or Data Source Name. It looks a lot like a standard URL, but it’s just a representation of the configuration required by the Sentry SDKs. It consists of a few pieces, including the protocol, public key, the server address, and the project identifier.
+## Configure
 
 To capture all errors, even the one during the startup of your application, you should initialize the Sentry PHP SDK as soon as possible.
 
 <SignInNote />
 
-```php
-\Sentry\init(['dsn' => '___PUBLIC_DSN___' ]);
+```php {onboardingOptions: {performance: '3-4', profiling: '5-6'}}
```
stephanie-anderson, line 77: this is so neat 😍

docs/platforms/php/index.mdx
```diff
@@ -25,28 +25,73 @@ Sentry captures data by using an SDK within your application’s runtime. These
...
+  'dsn' => '___PUBLIC_DSN___' ,
+  // Specify a fixed sample rate
+  'traces_sample_rate' => 1.0,
+  // Set a sampling rate for profiling - this is relative to traces_sample_rate
+  'profiles_sample_rate' => 1.0,
+]);
 ```
 
-## Usage
+# Verify
```
stephanie-anderson, line 87: Headline was made a h1 by accident?

docs/platforms/go/guides/gin/index.mdx
```diff
@@ -9,6 +9,16 @@ For a quick reference, there is a [complete example](https://github.com/getsentr
 
 ## Install
 
+<OnboardingOptionButtons
+  options={[
+    {
+      id: 'error-monitoring',
```
stephanie-anderson, line 15: Can we also simplify the 'error-monitoring' option? It will always be part (but passing 'error-monitoring' in the list is still worth it for explicity), but it will never be able to enable the checkbox. So configuring it in every instance of this component feels unnecessary

stephanie-anderson: Something like
```
<OnboardingOptionButtons
    'error-monitoring',
    'performance',
    'profiling',
  ]}
/>
```

a-hariti: hardcode it as disabled?

stephanie-anderson (approved):
Yay!!

PR #10134, opened 2024-05-23
fix(.NET): Release creation

bitsandfoxes:
Updated how the SDK will look up and create a release.

### DIFF
--- a/platform-includes/set-release/dotnet.mdx
+++ b/platform-includes/set-release/dotnet.mdx
@@ -8,8 +8,8 @@ options.Release = "my-project-name@2.3.12";
 options.Release <- "my-project-name@2.3.12"
 ```
 
-The SDK attempts to locate the release to add to events sent to Sentry.
+The SDK attempts to locate the release to attach to events sent to Sentry.
 
-The SDK will first check for a version set on the environment via `SENTRY_RELEASE` and use it as-is.
+The SDK will first check if there's a version set on the environment via `SENTRY_RELEASE` and use it as-is.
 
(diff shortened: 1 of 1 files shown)


platform-includes/set-release/dotnet.mdx
```diff
@@ -10,7 +10,6 @@ options.Release <- "my-project-name@2.3.12"
 
 The SDK attempts to locate the release to add to events sent to Sentry.
 
-The SDK will first look at the [entry assembly's](<https://msdn.microsoft.com/en-us/library/system.reflection.assembly.getentryassembly(v=vs.110).aspx>) `AssemblyInformationalVersionAttribute`, which accepts a string as
-value and is often used to set a GIT commit hash.
+The SDK will first check for a version set on the environment via `SENTRY_RELEASE` and use it as-is.
 
-If that returns null, it'll look at the default `AssemblyVersionAttribute` which accepts the numeric version number.
+If no version can be found it falls back to look at the [entry assembly's](<https://msdn.microsoft.com/en-us/library/system.reflection.assembly.getentryassembly(v=vs.110).aspx>) `AssemblyInformationalVersionAttribute`, which accepts a string as value and is often used to set a GIT commit hash. If that returns null, it'll look at the default `AssemblyVersionAttribute` which accepts the numeric version number. The resulting release will be in the format `assembly-name@version`.
```
vivianyentran, line 15:
```suggestion
If no version is found, the SDK will look at the [entry assembly's](<https://msdn.microsoft.com/en-us/library/system.reflection.assembly.getentryassembly(v=vs.110).aspx>) `AssemblyInformationalVersionAttribute`, which accepts a string value and is often used to set a GIT commit hash. If that returns null, then the SDK will look at the default `AssemblyVersionAttribute`, which accepts the numeric version number. The resulting release will be in the format `<assembly-name>@<version-number>`.
```

vivianyentran (approved):

PR #10146, opened 2024-05-24
feat(javascript): Document new `startNewTrace` API

Lms24:
See https://github.com/getsentry/sentry-javascript/pull/12138 for details.

### DIFF
--- a/docs/platforms/javascript/common/distributed-tracing/index.mdx
+++ b/docs/platforms/javascript/common/distributed-tracing/index.mdx
@@ -33,7 +33,7 @@ If you run any JavaScript applications in your distributed system, make sure tha
 
 <PlatformCategorySection supported={["browser"]}>
 
-In the browser, the SDK automatically starts a new trace in the following situation:
+In the browser, the SDK automatically starts a new trace in the following situations:
 
 - On page load: Whenever the page is (re-)loaded, a new trace is started. At the same time, a `pageload` span is created (see <PlatformLink to="/performance/instrumentation/automatic-instrumentation">Performance Monitoring</PlatformLink>). Once this span ends, the trace remains until the next navigation or page load. In case the server serving the initial page already started a trace and sent the necessary [HTML tags](./custom-instrumentation/#extract-tracing-information-from-html-meta-tags) to the browser, the SDK will continue this trace instead of starting a new one.
 - On navigation: Whenever a user navigates (for example in a single-page application), a new trace is started. At the same time, a `navigation` span is created (see <PlatformLink to="/performance/instrumentation/automatic-instrumentation">Performance Monitoring</PlatformLink>). Once this span ends, the trace remains until the next navigation or page load.
@@ -46,8 +46,8 @@ In both cases, this means that if you start spans after the automatic `pageload`
 
 Server-side SDKs handle traces automatically on a per-request basis. This means that SDKs will:
 
-- continue an existing trace if the incoming request contains a trace header
-- start a new trace if the incoming request does not contain a trace header. This trace stays active until the response is sent.
+- Continue an existing trace if the incoming request contains a trace header.
+- Start a new trace if the incoming request does not contain a trace header. This trace stays active until the response is sent.
 
 </PlatformCategorySection>
 
--- a/docs/platforms/javascript/common/distributed-tracing/index.mdx
+++ b/docs/platforms/javascript/common/distributed-tracing/index.mdx
@@ -35,8 +35,8 @@ If you run any JavaScript applications in your distributed system, make sure tha
 
 In the browser, the SDK automatically starts a new trace in the following situation:
 
-- On page load: Whenever the page is (re-)loaded, a new trace is started. Simultaneously, a `pageload` span is created (see <PlatformLink to="/performance/instrumentation/automatic-instrumentation">Performance Monitoring</PlatformLink>). Once this span ends, the trace remains until the next navigation or page load. In case the server serving the initial page already started a trace and sent the necessary [HTML tags](./custom-instrumentation/#extract-tracing-information-from-html-meta-tags) to the browser, the SDK will continue this trace instead of starting a new one.
-- On navigation: Whenever a user navigates (for example in a single-page application), a new trace is started. Simultaneously, a `navigation` span is created (see <PlatformLink to="/performance/instrumentation/automatic-instrumentation">Performance Monitoring</PlatformLink>). Once this span ends, the trace remains until the next navigation or page load.
+- On page load: Whenever the page is (re-)loaded, a new trace is started. At the same time, a `pageload` span is created (see <PlatformLink to="/performance/instrumentation/automatic-instrumentation">Performance Monitoring</PlatformLink>). Once this span ends, the trace remains until the next navigation or page load. In case the server serving the initial page already started a trace and sent the necessary [HTML tags](./custom-instrumentation/#extract-tracing-information-from-html-meta-tags) to the browser, the SDK will continue this trace instead of starting a new one.
+- On navigation: Whenever a user navigates (for example in a single-page application), a new trace is started. At the same time, a `navigation` span is created (see <PlatformLink to="/performance/instrumentation/automatic-instrumentation">Performance Monitoring</PlatformLink>). Once this span ends, the trace remains until the next navigation or page load.
 
 In both cases, this means that if you start spans after the automatic `pageload` and `navigation` spans ended, they will still be part of the same trace. This makes it easy to connect what happened before and after your span.
 
@@ -44,7 +44,7 @@ In both cases, this means that if you start spans after the automatic `pageload`
 
 <PlatformCategorySection supported={["server"]}>
 
-Server-side SDKs handle traces automatically on a per-request basis. This means that SDKs will
+Server-side SDKs handle traces automatically on a per-request basis. This means that SDKs will:
 
 - continue an existing trace if the incoming request contains a trace header
 - start a new trace if the incoming request does not contain a trace header. This trace stays active until the response is sent.
--- a/docs/platforms/javascript/common/distributed-tracing/index.mdx
+++ b/docs/platforms/javascript/common/distributed-tracing/index.mdx
@@ -29,6 +29,30 @@ What happens in the background is that Sentry uses reads and further propagates 
 
 If you run any JavaScript applications in your distributed system, make sure that those two headers are added to your CORS allowlist and won't be blocked or stripped by your proxy servers, gateways, or firewalls.
 
+## Trace Duration
+
+<PlatformCategorySection supported={["browser"]}>
+
+In the browser, the SDK automatically starts a new trace in the following situation:
+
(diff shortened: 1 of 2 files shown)


docs/platforms/javascript/common/distributed-tracing/index.mdx
```diff
@@ -29,6 +29,30 @@ What happens in the background is that Sentry uses reads and further propagates
 
 If you run any JavaScript applications in your distributed system, make sure that those two headers are added to your CORS allowlist and won't be blocked or stripped by your proxy servers, gateways, or firewalls.
 
+## Trace Duration
+
+<PlatformCategorySection supported={["browser"]}>
+
+In the browser, the SDK automatically starts a new trace in the following situation:
+
+- On page load: Whenever the page is (re-)loaded, a new trace is started. Simultaneously, a `pageload` span is created (see <PlatformLink to="/performance/instrumentation/automatic-instrumentation">Performance Monitoring</PlatformLink>). Once this span ends, the trace remains until the next navigation or page load. In case the server serving the initial page already started a trace and sent the necessary [HTML tags](./custom-instrumentation/#extract-tracing-information-from-html-meta-tags) to the browser, the SDK will continue this trace instead of starting a new one.
```
mydea, line 38:
```suggestion
- On page load: Whenever the page is (re-)loaded, a new trace is started. At the same time, a `pageload` span is created (see <PlatformLink to="/performance/instrumentation/automatic-instrumentation">Performance Monitoring</PlatformLink>). Once this span ends, the trace remains until the next navigation or page load. In case the server serving the initial page already started a trace and sent the necessary [HTML tags](./custom-instrumentation/#extract-tracing-information-from-html-meta-tags) to the browser, the SDK will continue this trace instead of starting a new one.
```

docs/platforms/javascript/common/distributed-tracing/index.mdx
```diff
@@ -29,6 +29,30 @@ What happens in the background is that Sentry uses reads and further propagates
...
+- On navigation: Whenever a user navigates (for example in a single-page application), a new trace is started. Simultaneously, a `navigation` span is created (see <PlatformLink to="/performance/instrumentation/automatic-instrumentation">Performance Monitoring</PlatformLink>). Once this span ends, the trace remains until the next navigation or page load.
```
mydea, line 39:
```suggestion
- On navigation: Whenever a user navigates (for example in a single-page application), a new trace is started. At the same time, a `navigation` span is created (see <PlatformLink to="/performance/instrumentation/automatic-instrumentation">Performance Monitoring</PlatformLink>). Once this span ends, the trace remains until the next navigation or page load.
```

super-l nit :D

docs/platforms/javascript/common/distributed-tracing/index.mdx
```diff
@@ -29,6 +29,30 @@ What happens in the background is that Sentry uses reads and further propagates
...
+
+In both cases, this means that if you start spans after the automatic `pageload` and `navigation` spans ended, they will still be part of the same trace. This makes it easy to connect what happened before and after your span.
+
+</PlatformCategorySection>
+
+<PlatformCategorySection supported={["server"]}>
```
mydea, line 45: m: Should we instead make this `notSupported={['browser']}`? There is also serverless which I think is not covered by either of these 🤔

docs/platforms/javascript/common/distributed-tracing/index.mdx
```diff
@@ -29,6 +29,30 @@ What happens in the background is that Sentry uses reads and further propagates
...
+
+Server-side SDKs handle traces automatically on a per-request basis. This means that SDKs will
```
mydea, line 47:
```suggestion
Server-side SDKs handle traces automatically on a per-request basis. This means that SDKs will:
```

platform-includes/distributed-tracing/custom-instrumentation/javascript.mdx
```diff
@@ -173,6 +173,28 @@ In this example, tracing information is propagated to the project running at `ht
 
 The two services are now connected with your custom distributed tracing implementation.
 
+### Starting a New Trace
+
+In case the SDK's [default behavior](../#trace-duration) for the trace duration does not fir your use case, you can manually start a new trace that will no longer be connected to the current (distributed) trace.
```
mydea, line 178:
```suggestion
In case the SDK's [default behavior](../#trace-duration) for the trace duration does not fit your needs, you can manually start a new trace that will no longer be connected to the current (distributed) trace.
```

mydea (approved):
very nice!

docs/platforms/javascript/common/distributed-tracing/index.mdx
Lms24: Oh I had no idea that serverless isn't covered by "server". Good to know! 
Yup, will change it to notSupported then.

Lms24: Oh, so turns out, `<PlatformCategorySection>` doesn't have a `notSupported` prop. So for now I just added `serverless` here. Tbh, this feels like a huge footgun to me. Historically we always assumed serverless===server in docs. I'm afraid we have some places where we should include serverless but don't exactly for this reason. I'll see if there's a simple improvement we can make here...

docs/platforms/javascript/common/distributed-tracing/index.mdx
```diff
@@ -29,6 +29,30 @@ What happens in the background is that Sentry uses reads and further propagates
 
 If you run any JavaScript applications in your distributed system, make sure that those two headers are added to your CORS allowlist and won't be blocked or stripped by your proxy servers, gateways, or firewalls.
 
+## Trace Duration
+
+<PlatformCategorySection supported={["browser"]}>
+
+In the browser, the SDK automatically starts a new trace in the following situation:
```
vivianyentran, line 36:
```suggestion
In the browser, the SDK automatically starts a new trace in the following situations:
```

docs/platforms/javascript/common/distributed-tracing/index.mdx
```diff
@@ -29,6 +29,30 @@ What happens in the background is that Sentry uses reads and further propagates
...
+
+- On page load: Whenever the page is (re-)loaded, a new trace is started. At the same time, a `pageload` span is created (see <PlatformLink to="/performance/instrumentation/automatic-instrumentation">Performance Monitoring</PlatformLink>). Once this span ends, the trace remains until the next navigation or page load. In case the server serving the initial page already started a trace and sent the necessary [HTML tags](./custom-instrumentation/#extract-tracing-information-from-html-meta-tags) to the browser, the SDK will continue this trace instead of starting a new one.
```
vivianyentran, line 38: In the performance monitoring link, it says that a new transaction (not trace) is started on page load and navigation. We should update that page as well to refer to trace not transaction?

docs/platforms/javascript/common/distributed-tracing/index.mdx
```diff
@@ -29,6 +29,30 @@ What happens in the background is that Sentry uses reads and further propagates
...
+- On navigation: Whenever a user navigates (for example in a single-page application), a new trace is started. At the same time, a `navigation` span is created (see <PlatformLink to="/performance/instrumentation/automatic-instrumentation">Performance Monitoring</PlatformLink>). Once this span ends, the trace remains until the next navigation or page load.
```
vivianyentran, line 39: I'm not sure what "navigation" here means and what differentiates it from a page load. Is this referring to scrolling within the same page and the URL changes? Otherwise, when you navigate within a site to another page, isn't that a page load?

docs/platforms/javascript/common/distributed-tracing/index.mdx
```diff
@@ -29,6 +29,30 @@ What happens in the background is that Sentry uses reads and further propagates
...
+
+In both cases, this means that if you start spans after the automatic `pageload` and `navigation` spans ended, they will still be part of the same trace. This makes it easy to connect what happened before and after your span.
```
vivianyentran, line 41: I think what's missing for me is a description of what causes a trace to end.

docs/platforms/javascript/common/distributed-tracing/index.mdx
```diff
@@ -29,6 +29,30 @@ What happens in the background is that Sentry uses reads and further propagates
...
+
+</PlatformCategorySection>
+
+<PlatformCategorySection supported={["server", "serverless"]}>
+
+Server-side SDKs handle traces automatically on a per-request basis. This means that SDKs will:
+
+- continue an existing trace if the incoming request contains a trace header
+- start a new trace if the incoming request does not contain a trace header. This trace stays active until the response is sent.
```
vivianyentran, line 50:
```suggestion
- Continue an existing trace if the incoming request contains a trace header.
- Start a new trace if the incoming request does not contain a trace header. This trace stays active until the response is sent.
```

platform-includes/distributed-tracing/custom-instrumentation/javascript.mdx
```diff
@@ -173,6 +173,30 @@ In this example, tracing information is propagated to the project running at `ht
...
+      { op: "ui.interaction.click", name: "fetch click" },
+      async () => {
+        await fetch("http://example.com");
+      }
+    );
+  });
+});
+```
+
+Once the callback ended, the SDK will continue the previous trace (if available).
```
vivianyentran, line 198:
```suggestion
Once the callback ends, the SDK will continue the previous trace (if available).
```

docs/platforms/javascript/common/distributed-tracing/index.mdx
Lms24: In a way both is true but the wording is not optimal. We want to phase out mentioning "transaction" at all because our SDK no longer exposes transaction in a public API. Likewise, the product is also moving more towards spans. So instead of a pageload transaction, the SDK creates a pageload _span_.

Additionally, the SDK also startes a _trace_ to which the span is associated to. I hope that makes sense

docs/platforms/javascript/common/distributed-tracing/index.mdx
Lms24: This is a bit tricky indeed. We're generally differentiating like so:

- page load: The browser actually loads a new html page. For example when you first go to a website or when you reload it. And yes, you're right, in some cases (depending on the chosen technology), clicking on a link to a different page within the same site/domain will also cause the browser to load a completely new page. So this is conceptually also a "navigation" but technically a pageload. You can generally call this type of web application Multi-Page Applications. 
- navigation: The URL of the browser changes but without the browser reloading the entire page. This is what a lot of Single Page Applications, like a React app, do. They have "one" HTML page but its content and the associated URL path change as the user navigates through the application. So for SPAs, loading the first time is a pageload and every further navigation is what we refer to here as a navigation.

docs/platforms/javascript/common/distributed-tracing/index.mdx
Lms24: From the paragraphs above: "the trace remains until the next navigation or page load"
So IOW, the span only ends when users reload the page or navigate to another part of the page. Does this make sense?

Issue #10159, opened 2024-05-27
Allow negative check for `<OnboardingOption>`


mydea:
You can use something like this to show a block only when an option is enabled:

```jsx
<OnboardingOption optionId="profiling">
  show when profiling is enabled
</OnboardingOption>
```

However, this does not allow to do some kind of "else" logic. For example, I want to be able to show either this snippet:

```bash
npm install --save @sentry/node
```

Or, if profiling is enabled, show this snippet:

```bash
npm install --save @sentry/node @sentry/profiling-node
```

I don't think this is doable today? (or if it is, happy for pointers).

## How we could do this

We could allow to pass `hideForOptionsId='profiling'`, which we could use to build else blocks manually?

a-hariti:
What about:

```bash {"onboardingOptions": {"performance": "1"}}
npm install --save @sentry/node \
  @sentry/profiling-node
```
but this comes with the issue of the trailing `\`

@mydea

mydea:
This is not working, it is removing everything for me 😅 and even so I wouldn't want to have a line break in there either way!

a-hariti:
I'm thinking:

```jsx
<OnboardingOption optionId="profiling" hideForThisOption>
  hide when profiling is enabled
</OnboardingOption>
```
wyt @mydea ?
should this be hidden by default in the DOM?

mydea:
I _think_ considering how this works for other things we have, I'd prefer an inverse option (e.g. think we have `<PlatformSection notSupported=....`, but no strong feelings.

I think hiding it by default makes sense!

PR #10164, opened 2024-05-27
feat: Update node getting started pages

mydea:
This updates the getting started page for all node docs to use the new product selector.

I actually extracted the page out into an include because they all (fastify, express, node, ...) share the same content, so this is much easier to update.

Once this is merged: https://github.com/getsentry/sentry-docs/issues/10159 we can streamline the top section for the install command a bit more.


lforst (approved):
Seems good to me. Node looks pretty easy, browser is more spicy. Crazy how much duplication we had.

PR #10171, opened 2024-05-28
feat(node): Add docs for late init

mydea:
This adds a new alternative installation method to late-init Sentry.

### DIFF
--- a/docs/platforms/javascript/common/install/late-initializtion.mdx
+++ b/docs/platforms/javascript/common/install/late-initializtion.mdx
@@ -1,0 +1,131 @@
+---
+title: Late Initialization (ESM or CJS)
+sidebar_order: 12
+description: "Learn about running Sentry in an ESM or CJS application, in scenarios where you cannot run init early."
+supported:
+  - javascript.node
+  - javascript.connect
+  - javascript.express
+  - javascript.fastify
+  - javascript.hapi
+  - javascript.koa
+---
+
+<Alert>
+  Are you unsure if you should use this installation method? Review our
+  [installation methods](../).
+</Alert>
+
+In order for auto-instrumentation to work, it is generally required to run `Sentry.init()` as early as possible, before anything else is imported in your application.
+
+However, in some cases this may not be possible to do - for example, if you are fetching your DSN from an external source. In this case, you can use the `@sentry/node/preload` hook to ensure modules are wrapped early, which allows you to call `Sentry.init()` later at a time of your choosing.
+
+<Note>
+  We recommend to only use this method if strictly necessary. In most cases, it
+  is better to find a way to run `Sentry.init()` early in your application, in
+  order to ensure that no error can go unreported.
+</Note>
+
+## Late Initialization with CommonJS (CJS)
+
+In your CJS application, use the `@sentry/node/preload` hook with `--require` to ensure modules are wrapped early:
+
+```bash
+# Note: This is only available for Node v18.19.0 onwards.
+node --require @sentry/node/preload app.js
+```
+
+Then, in your application you can call `Sentry.init()` at a later point:
+
+```javascript {filename: main.js}
+const startApp = require("./app");
+const fetchDsn = require("./utils/fetchDsn");
+const Sentry = require("@sentry/node");
+
+startApp();
+
+const dsn = fetchDsn();
+Sentry.init({
+  dsn,
+
+  // Add Performance Monitoring by setting tracesSampleRate
+  // We recommend adjusting this value in production
+  tracesSampleRate: 1.0,
+});
+
+// From now on, Sentry is initialized,
+// but the app is still auto-instrumented
+```
+
+## Late Initialization with ESM
+
+In your CJS application, use the `@sentry/node/preload` hook with `--import` to ensure modules are wrapped early:
+
+```bash
+# Note: This is only available for Node v18.19.0 onwards.
+node --import @sentry/node/preload app.,js
+```
+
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/common/install/late-initializtion.mdx
```diff
@@ -0,0 +1,131 @@
...
+---
+
+<Alert>
+  Are you unsure if you should use this installation method? Review our
+  [installation methods](../).
+</Alert>
+
+In order for auto-instrumentation to work, it is generally required to run `Sentry.init()` as early as possible, before anything else is imported in your application.
+
+However, in some cases this may not be possible to do - for example, if you are fetching your DSN from an external source. In this case, you can use the `@sentry/node/preload` hook to ensure modules are wrapped early, which allows you to call `Sentry.init()` later at a time of your choosing.
```
andreiborza, line 21:
```suggestion
However, in some cases this may not be possible to do—for example, if you are fetching your DSN from an external source. In this case, you can use the `@sentry/node/preload` hook to ensure modules are wrapped early, which allows you to call `Sentry.init()` later at a time of your choosing.
```

docs/platforms/javascript/common/install/late-initializtion.mdx
```diff
@@ -0,0 +1,131 @@
...
+  is better to find a way to run `Sentry.init()` early in your application, in
+  order to ensure that no error can go unreported.
+</Note>
+
+## Late Initialization with CommonJS (CJS)
+
+In your CJS application, use the `@sentry/node/preload` hook with `--require` to ensure modules are wrapped early:
+
+```bash
+# Note: This is only available for Node v18.19.0 onwards.
```
andreiborza, line 34: q: Is that true for CJS?

docs/platforms/javascript/common/install/late-initializtion.mdx
```diff
@@ -0,0 +1,131 @@
...
+  tracesSampleRate: 1.0,
+});
+
+// From now on, Sentry is initialized,
+// but the app is still auto-instrumented
+```
+
+## What does Preloading mean?
+
+Integrations that are preloaded ensure that the necessary modules are wrapped early, before they can be imported by your application. At this point, the modules are wrapped, but will not do anything - nothing will be emitted or captured from them.
```
andreiborza, line 93:
```suggestion
Integrations that are preloaded ensure that the necessary modules are wrapped early, before they can be imported by your application. At this point, the modules are wrapped, but will not do anything—nothing will be emitted or captured from them.
```

docs/platforms/javascript/common/install/late-initializtion.mdx
mydea: Actually I don't know, good question - here it says 1.x, so probably it's fine xD https://nodejs.org/api/cli.html#-r---require-module

So I'll remove this here!

mydea, line 34:
```suggestion
```

docs/platforms/javascript/common/install/late-initializtion.mdx
mydea: I guess this is an em-dash, which is nice, but should we not still have spaces around this? Seems (visually) a bit weird to me without the spaces...?

andreiborza: No space around emdashes. I was taught my Sigrid :)
https://www.thepunctuationguide.com/em-dash.html#spaces

mydea: I don't like it :D But I accept it ;) 

It does say there though:

> Spaces with the em dash
The em dash is typically used without spaces on either side, and that is the style used in this guide. **Most newspapers, however, set the em dash off with a single space on each side.**

but 🤷 no strong feelings ;)

andreiborza: tbh I've also always added spaces. If the em-dash is visibly large enough then I don't mind without spaces but sometimes they are very similar (e.g. here on github).

fwiw on the actual docs page it loo
Screenshot 2024-05-28 at 10 56 05
ks a bit better:

andreiborza: I think we should pick one style and go with it (currently we have it without spaces).

andreiborza (approved):

AbhiPrasad (approved):

AbhiPrasad (commented):
Oh forgot - we should add a version note! So people know they need to use `8.5.0` for this!

docs/platforms/javascript/common/install/late-initializtion.mdx
```diff
@@ -0,0 +1,130 @@
...
+
+In order for auto-instrumentation to work, it is generally required to run `Sentry.init()` as early as possible, before anything else is imported in your application.
+
+However, in some cases this may not be possible to do—for example, if you are fetching your DSN from an external source. In this case, you can use the `@sentry/node/preload` hook to ensure modules are wrapped early, which allows you to call `Sentry.init()` later at a time of your choosing.
+
+<Note>
+  We recommend to only use this method if strictly necessary. In most cases, it
+  is better to find a way to run `Sentry.init()` early in your application, in
+  order to ensure that no error can go unreported.
+</Note>
```
mydea, line 27:
```suggestion
</Note>

This initialization method is available starting in version **8.5.0**.
```

mydea: @AbhiPrasad adding this note!

PR #10181, opened 2024-05-28
add caches walkthrough arcade

DominikB2014:
Adds cache module arcade to caches documentation and queue module arcade.

I added it in a location consistent to the other modules, for examples requests shows the arcade just after the quick blurb on how to use the module
https://docs.sentry.io/product/performance/requests/


edwardgou-sentry (approved):
Thanks!

Issue #10141, opened 2024-05-24
Sentry Distributed Tracing - App Architecture Diagram with its Sentry Trace


thinkocapo:
### Core or SDK?

Platform/SDK

### Which part? Which one?

all sdk's

### Description

<img width="1617" alt="image" src="https://github.com/getsentry/sentry-docs/assets/8920574/ed92a3f3-53f9-4763-9615-fe1b97e6fb30">
<img width="1098" alt="image" src="https://github.com/getsentry/sentry-docs/assets/8920574/466d38dc-337c-4f51-a094-44f348cff00d">

**Most Important Thing We Need - High Level**
- App Architecture Diagram
- Sentry Trace (UI) that shows the corresponding Spans and Sentry SDK code needed for everything connected in the App Architecture Diagram

**We need**
- Distinction of Web vs Non-Web, written out in words before the diagram.
- Vertically oriented flow of frontend-to-backend, instead of our horizontal left-to-right. see screenshots of the other vendors.
- Sentry Trace that shows all of the apps+services in the diagram, all the http requests, any non-http requests (whether custom instrumented or not, that's a different question)
- Highlight apps+services in grey (like in Lightstep's) which are not covered by auto-instrumentation (http propagation) so that way it's light-vs-day clear to a user what to expect.
- Mention that custom instrumentation is needed for the non-http and put a hyperlink to a code example
- Mention that "no agent necessary" and consider writing the benefits of our approach and the drawbacks of needing agents. Emphasize focus on developers, our vision.
- Mention `sentry-trace` header is used
- We don't need to worry about, "But the trace view UI will become outdated" because currently we don't show one whatsoever and this makes it super hard for a newcomer to sentry to understand what to expect

**Problems with Sentry Diagram**
- "Browser" and "Backend" how about "Web App" or "Ecommerce App Forntend" and "Backend API" or "Springboot Server" something more realistic? AuthService|BillingService|OrderProcessing would be great.
- More apps+servers. Enterprise Customers have hundreds of things, why are we only showing 3.
- Caches&Queues (and the Sentry Trace View to show the right spans for it) are not shown
- "Database Server" so is that an app you install an sdk in? I never have users+customers talk to me about "database server" or does that mean the VM/machine that a database is on? Or an app with its own REST API and its sole purpose is to talk to the database? I thought we'd only want to show apps+services with sdk's installed and then you see transaction.spans representing CRUD operations over to a database. I prefer we avoid confusion that you install a sdk inside a database
[...]

Issue #8287, opened 2023-10-18
Open Telemetry support should mention Collector and Exporters and include a Diagram


thinkocapo:
### Core or SDK?

Platform/SDK

### Which part? Which one?

all sdk's that have OTEL integration

### Problem
I explain Sentry's approach with Span Processors on calls, I show the code, our documentation. The customer's response sounds as if they completely ignored what I said and still don't understand. Maybe their mind is super wrapped around a different approach, or they don't have a roadmap defined for their own Otel use case. Maybe they have a poor understanding of Otel in general. Or my explanation is super bad.

When there are multiple variant techniques for how a technology (Otel) can be implemented, and it's unknown the currently implemented approach and roadmap of a customer, then it's bad to rely on a verbal conversation alone without having a Visual to set the stage for what the possibilities are. It helps both sides understand what's currently in place versus what the possibilities are. It demonstrates better Thought Leadership too on the subject.

Additionally, we're not always live on a phone call to have such a discussion and hand-hold them through all this. So a visual would be key here.

### Description
Asks
1. Could we get a diagram in SDK docs? Or, in a general product docs area? a diagram like the one in https://github.com/getsentry/sentry/discussions/40712
2. Could we get the words Collector or Exporter added somewhere in our SDK docs? 
3. Can we describe if we are tail-based sampling or not and why? I get asked a lot.
4. [docs.sentry.io/product/performance/opentelemetry](https://docs.sentry.io/product/performance/opentelemetry/)

**Why:** Currently there is 0 mention of Collectors and Exporters in SDK docs. Oftentimes I show them our SDK docs, which to me is pretty clear, I explain it on a call. They still start asking me questions about Collectors and Exporters. If we're not a Collector or Exporter, _then say it_ in docs. If we're something else, _then say it_.

This explanation is clear to me, but it's toug to grasp without a nice visual that sets the stage indicating other Otel backends, exporters, collectors, and the Sentry backend. While all that might be _implied_, I think people have a hard time visualizing all this. https://docs.sentry.io/platforms/python/performance/instrumentation/opentelemetry/#opentelemetry-and-sentry

### Suggested Solution

Diagram and the words Collector + Exporter in every single SDK's documentation, or in an sdk-agnostic area like "docs.sentry.io/product/performance/otel" would be nice, currently it does not exist. Or add a link in each SDK's documentation, pointing over to the general product doc for OTEL (which does not exist yet).

1. Could we have a diagram with Sentry colors and UI elements? 
[...]

PR #10192, opened 2024-05-29
Create opentelemetry.mdx

smeubank:
Add early product docs on OTel

building off of this comment:

https://github.com/getsentry/sentry/discussions/55023#discussioncomment-9571661

also potentially this discussion

- Motivation for OTel support Sentry
- What is OTel support at Sentry?
  - what won't we support

### DIFF
--- a/docs/concepts/key-terms/tracing/opentelemetry.mdx
+++ b/docs/concepts/key-terms/tracing/opentelemetry.mdx
@@ -1,0 +1,35 @@
+---
+title: "OpenTelemetry"
+sidebar_order: 71
+description: "Learn more about how Sentry works with OpenTelemetry sourced data."
+---
+
+## What is OpenTelemetry
+
+## What does Sentry support
+
+1. Performance Powered by OpenTelemetry: 🔋
+   - This is the topic of this discussion, and has officially been launched with our latest major release of our JS SDK: [v8 Changelog](https://github.com/getsentry/sentry-javascript/releases/tag/8.0.0)
+   - In simple terms, this is the Sentry Node SDK, and our Node.js tracing and performance monitoring features are now largely handled by OpenTelemetry "under the hood" as a dependency of our SDK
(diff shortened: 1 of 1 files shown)


thinkocapo:
**Ask1:**
From a convo with an engineer:
> we need a coherent story that we tell the outside world about what exactly we are doing with otel
> 
> we’ll never be fully otel compatible at the end of the day. if this is the case we still need to explain to people both internally and externally what parts we use and why.
> 
> but can we summarize a clean sentence like ‘we are using OTEL for leveraging their instrumentation ecosystem’
> 
> like if we can say for now we are only interested in the extra instrumentation, it makes things clearer in every engineer’s head

From the OriginalPost:

Upon reading this, it should be clear **why** we are doing what we are doing, why it supports solving the problems we seek to solve, and why it is an advantage.
> Motivation for OTel support Sentry

Great, the sounds like the **what**.
> What is OTel support at Sentry?

Yes. Bonus if you can explain "why" we don't support it. Otherwise, explaining the **why** for what we do support, should be okay. Can always respond with that answer when asked, "Why don't you have a exporter/collector?"
> what won't we support

**Ask2:**
A diagram would still be fantastic. Doesn't have to show Exporter/Collector, especially if that's not what we do. It could show only what we do. OTEL SDK, Sentry SDK, Sentry.io backend, YourApp.
"Open Telemetry support should mention Collector and Exporters and include a Diagram #8287"

**Ask3:**
Bonus: Reminds me of this tracing documentation request. Super helpful to show what's traced from an example system, as well as the resulting Sentry data. Do this...but for when OTEL is being used. Take for example Sentry's demo app for Otel (telescopes app) in [Introducing OpenTelemetry Support: Take Action on Your Observability Data](https://blog.sentry.io/introducing-opentelemetry-support/) it shows the resulting data.
"Sentry Distributed Tracing - App Architecture Diagram with its Sentry Trace #10141"

PR #10194, opened 2024-05-29
ref: Use default prettier config for formatting code blocks instead of repository config

lforst:
We do not want "our" (as in, the docs developers) prettier config to be what dictates the content of code blocks, but rather we want to stick to the default prettier config as much as possible because that is what people are used to.


a-hariti (approved):
LGTM 🚀

PR #10241, opened 2024-06-03
feat(js): Add new onboarding for all JavaScript guides

mydea:
Except for base browser JS, which still uses the loader-based onboarding for now.

This changes quite a few things, but hopefully for the better!

I did not touch cordova, but this should be fine for now...

Meta SDKs are also not touched by this because they use the wizard.

Two other changes that were necessary to make this work:

1. I added `notSupported` to `<PlatformCategorySection>`, in order to properly be able to skip meta frameworks
2. I updated bun to count as "server" SDK, not both server + browser

### DIFF
--- a/docs/platforms/javascript/common/index.mdx
+++ b/docs/platforms/javascript/common/index.mdx
@@ -41,10 +41,29 @@ If you prefer to follow video instructions, see [How to Install the Sentry Next.
 
 </PlatformSection>
 
-
 <PlatformSection notSupported={["javascript.deno"]}>
 
 ## Install
+
+<PlatformCategorySection notSupported={["server"]}>
+  <PlatformSection notSupported={["javascript", 'javascript.cordova']}>
+    <OnboardingOptionButtons
+      options={["error-monitoring", "performance", "session-replay"]}
+    />
+  </PlatformSection>
+</PlatformCategorySection>
+
+<PlatformCategorySection notSupported={["browser"]}>
+  <PlatformSection notSupported={["javascript.bun"]}>
+    <OnboardingOptionButtons
+      options={["error-monitoring", "performance", "profiling"]}
+    />
+  </PlatformSection>
+
+  <PlatformSection supported={["javascript.bun"]}>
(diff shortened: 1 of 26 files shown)


docs/platforms/javascript/common/index.mdx
```diff
@@ -37,11 +37,30 @@ If you prefer to follow video instructions, see [How to Install the Sentry Next.
 
 </PlatformSection>
 
-
 <PlatformSection notSupported={["javascript.deno"]}>
 
 ## Install
 
+<PlatformCategorySection notSupported={["server"]}>
```
mydea, line 44: This is all a bit complicated, but it makes sense:

1. We want to show this only for browser-only SDKs, not for any SDK that _also_ has a browser aspect
2. We also don't show it for cordova and plain JavaScript

docs/platforms/javascript/common/index.mdx
```diff
@@ -37,11 +37,30 @@ If you prefer to follow video instructions, see [How to Install the Sentry Next.
...
+  <PlatformSection notSupported={["javascript", 'javascript.cordova']}>
+    <OnboardingOptionButtons
+      options={["error-monitoring", "performance", "session-replay"]}
+    />
+  </PlatformSection>
+</PlatformCategorySection>
+
+<PlatformCategorySection notSupported={["browser"]}>
```
mydea, line 52: Most server guides to not actually render this page, but still, to be safe...

docs/platforms/javascript/common/performance/index.mdx
```diff
@@ -17,13 +17,22 @@ If you’re adopting Performance in a high-throughput environment, we recommend
 
 </PlatformCategorySection>
 
+<PlatformContent supported={['javascript']}>
```
mydea, line 20: For non-base browser JS, this really should only say "Tracing is enabled by default", which has no value. We used to show something like "If you use the loader, ..." here for all SDKs, which was not really correct.

AbhiPrasad (approved):

s1gr1d (approved):

PR #10246, opened 2024-06-03
feat(tracing): Update spans section

wmak:
- Updating spans section in the tracing page for some changes coming up
- Given this is user facing documentation, removing the `(optional)` text since that doesn't make sense given the public api for spans is through `sentry_sdk.start_span()`
- Linking to searchable properties (that doesn't exist yet so this PR will need to wait for that) from here if that's what the user is looking for


lizokm (approved):
LGTM, thanks for adding.

PR #10247, opened 2024-06-03
feat(performance): adds note about webvitals samples

edwardgou-sentry:
Adds a note to clarify web vitals samples list in page overview:
image

### DIFF
--- a/docs/product/performance/web-vitals/index.mdx
+++ b/docs/product/performance/web-vitals/index.mdx
@@ -60,6 +60,10 @@ The **Page Overview** page displays a "Page Loads" chart in the right sidebar of
 
 ### Samples List
 
+<Note>
+The samples list only shows page loads with web vitals and performance score. If a page load is missing a required web vital, it will not be shown in the samples list.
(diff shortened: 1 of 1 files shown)


docs/product/performance/web-vitals/index.mdx
```diff
@@ -60,6 +60,10 @@ The **Page Overview** page displays a "Page Loads" chart in the right sidebar of
 
 ### Samples List
 
+<Note>
+The samples list only shows page loads with web vitals and performance score. If a page load is missing a required web vital, it will not be shown in the samples list.
```
zoesyc, line 64: Can we add reference to the Browser Support section to help them understand why a pageload may not capture web vitals?

zoesyc (changes requested):

docs/product/performance/web-vitals/index.mdx
edwardgou-sentry: Updated!

zoesyc (approved):

PR #10253, opened 2024-06-04
feat(angular): Remove compatibility note in replay/feedback install

mydea:
IMHO this is a bit redundant, I think it's enough to have this on the getting started page...?


Lms24 (approved):
LGTM! I'd argue we don't even need to show the npm commands and SDK setup on this page but just link to getting started instead and show how to add the replay-specific init options here. But that requires a bigger discussion, I'm afraid 😅

stephanie-anderson (approved):

PR #10306, opened 2024-06-06
docs(replayIntegration): Add snippets without top level await

s1gr1d:
## DESCRIBE YOUR PR
Adding snippets for setups without top level await support. Also added filename arguments to the Next.js replayIntegration docs.
ref https://github.com/getsentry/sentry-javascript/issues/12364 

## IS YOUR CHANGE URGENT?  

## PRE-MERGE CHECKLIST

### DIFF
--- a/platform-includes/session-replay/setup/javascript.angular.mdx
+++ b/platform-includes/session-replay/setup/javascript.angular.mdx
@@ -61,3 +61,21 @@ Sentry.init({
 const { replayIntegration } = await import("@sentry/angular");
 Sentry.addIntegration(replayIntegration());
 ```
+
+If your setup does not support top level await, you can create a function and immediately call it (or write an IIFE).
+
+```javascript {tabTitle:Function Definition}
+async function loadReplay() {
+  const { replayIntegration } = await import('@sentry/angular');
+  Sentry.addIntegration(replayIntegration());
+}
+
+loadReplay();
+```
+
+```javascript {tabTitle:IIFE}
+(async () => {
+  const { replayIntegration } = await import('@sentry/angular');
+  Sentry.addIntegration(replayIntegration());
+})();
+```
--- a/platform-includes/session-replay/setup/javascript.angular.mdx
+++ b/platform-includes/session-replay/setup/javascript.angular.mdx
@@ -58,24 +58,7 @@ Sentry.init({
 });
 
 // Sometime later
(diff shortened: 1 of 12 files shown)


platform-includes/session-replay/setup/javascript.angular.mdx
```diff
@@ -61,3 +61,21 @@ Sentry.init({
 const { replayIntegration } = await import("@sentry/angular");
 Sentry.addIntegration(replayIntegration());
 ```
+
+If your setup does not support top level await, you can create a function and immediately call it (or write an IIFE).
```
lforst, line 65: I recommend we do one of the two. Personally I would vote just the `async function` one because some people, especially beginners don't know the IIFE syntax.

Actually now that I think of it. Why can't we just recommend this by default instead:

```ts
// Sometime later
import("@sentry/angular").then((lazyLodedSentry) => {
  Sentry.addIntegration(lazyLodedSentry.replayIntegration());
});
```

s1gr1d: Yes, that's even nicer! I tried it and it works just as well

s1gr1d: Hm...or does this cause problems as a not-at-the-top import?

lforst: > Hm...or does this cause problems as a not-at-the-top import?

what do you mean exactly?

s1gr1d: And about the IIFE: I added it in the second tab, so it's not the default. But people might actually choose the IIFE over something different (e.g. a person in [this issue](https://github.com/getsentry/sentry-javascript/issues/12364#issuecomment-2152471782)) and junior devs can learn something new :)  I can link the [mdn glossary](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) though.

s1gr1d: > > Hm...or does this cause problems as a not-at-the-top import?
> 
> what do you mean exactly?

Ah no, it is a dynamic import anyway. The things I wrote are referring to the non-dynamic `import`, which does not apply here. Wrong thinking.

Lms24 (approved):

lizokm (approved):
LGTM!

PR #10333, opened 2024-06-10
fix(apple): Use id<SentrySpan> in performance obj-c code snippets

krystofwoldrich:
I've noticed our Apple docs mention Span protocol, but that didn't work for me. Should it be `SentrySpan`?


philipphofmann (approved):
Indeed, it should be `SentrySpan`. I guess hardly anybody uses ObjC anymore 😄. LGTM

PR #10382, opened 2024-06-12
fix(profiling) wrong integration casing

JonasBa:
The diffs are sad, but the casing used before was `Sentry.BrowserProfilingIntegration` whereas it should be `Sentry.browserProfilingIntegration`


Zylphrex (approved):

vivianyentran (approved):

stephanie-anderson (changes requested):
Hey @JonasBa - the contents in `public/` must not be changed (and will be overwritten anyways). Please update the corresponding .mdx files in `docs/`, `includes/` or `platform-includes/`

stephanie-anderson:
> Hey @JonasBa - the contents in `public/` must not be changed (and will be overwritten anyways). Please update the corresponding .mdx files in `docs/`, `includes/` or `platform-includes/`

Sorry, found the [reason](https://github.com/getsentry/sentry-docs/blob/master/public/_platforms/_README.md) for this. Will follow up with the Telemetry Experience team if this is still needed.

JonasBa:
@stephanie-anderson we still need this because the docs on the sentry dashboard are still using this. 

I found no other reference to these files in any mdx docs, so I assume they are the correct source and need to update them, else the in-app onboarding is broken.

stephanie-anderson (approved):

PR #10472, opened 2024-06-24
Make `allowUrls` and `denyUrls` more precise

lforst:
## DESCRIBE YOUR PR

This PR makes the docs around `allowUrls` and `denyUrls` more precise. There is a subtle but very important difference between where an error was *created*, where it has been *thrown*, and where it has been *captured*. These option concern themselves with where an error has been *created*.

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

### DIFF
--- a/includes/platforms/configuration/options/allow-urls.mdx
+++ b/includes/platforms/configuration/options/allow-urls.mdx
@@ -1,6 +1,6 @@
 An array of strings or regex patterns that match the URLs of scripts where errors have been created.
-Only errors that have been created on these URLs will be sent sent to Sentry.
-If you use this option, only Errors whose top stack frame file URL contains (string) or matches (regex) at least one entry in the array will be sent.
+Only errors that have been created on these URLs will be sent to Sentry.
+If you use this option, errors will only be sent when the top stack frame file URL contains (string) or matches (regex) at least one entry in the `allowUrls` array.
 
 For example, if you add `'foo.com'` to the array, errors created on `https://bar.com/myfile/foo.com` will be captured because URL will be matched with "contains" logic and the last segment of the URL contains `foo.com`.
 
--- a/includes/platforms/configuration/options/allow-urls.mdx
+++ b/includes/platforms/configuration/options/allow-urls.mdx
@@ -1,1 +1,15 @@
-A list of strings or regex patterns that match error URLs which should exclusively be sent to Sentry. If you use this option, only errors whose entire file URL contains (string) or matches (regex) at least one entry in the list will be sent. As a result, if you add `'foo.com'` to it, it will also match on `https://bar.com/myfile/foo.com`. Keep in mind that this only applies for captured exceptions, not raw message events. By default, all errors are sent.
+An array of strings or regex patterns that match the URLs of scripts where errors have been created.
+Only errors that have been created on these URLs will be sent sent to Sentry.
+If you use this option, only Errors whose top stack frame file URL contains (string) or matches (regex) at least one entry in the array will be sent.
+
+For example, if you add `'foo.com'` to the array, errors created on `https://bar.com/myfile/foo.com` will be captured because URL will be matched with "contains" logic and the last segment of the URL contains `foo.com`.
+
+This matching logic applies for captured exceptions, not raw message events. By default, all errors are sent.
+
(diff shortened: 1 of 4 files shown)


includes/platforms/configuration/options/allow-urls.mdx
```diff
@@ -1 +1,15 @@
-A list of strings or regex patterns that match error URLs which should exclusively be sent to Sentry. If you use this option, only errors whose entire file URL contains (string) or matches (regex) at least one entry in the list will be sent. As a result, if you add `'foo.com'` to it, it will also match on `https://bar.com/myfile/foo.com`. Keep in mind that this only applies for captured exceptions, not raw message events. By default, all errors are sent.
+An array of strings or regex patterns that match the URLs of scripts where errors have been created.
+Only errors that have been created on these URLs will be sent sent to Sentry.
+If you use this option, only Errors whose top stack frame file URL contains (string) or matches (regex) at least one entry in the array will be sent.
+
```
lizokm, line 4:
```suggestion
An array of strings or regex patterns that match the URLs of scripts where errors have been created.
Only errors that have been created on these URLs will be sent to Sentry.
If you use this option, at least one entry in the array will be sent if an error's top stack frame file URL contains (string) or matches (regex).

```

includes/platforms/configuration/options/allow-urls.mdx
```diff
@@ -1 +1,15 @@
...
+For example, if you add `'foo.com'` to the array, errors created on `https://bar.com/myfile/foo.com` will be captured because URL will be matched with "contains" logic and the last segment of the URL contains `foo.com`.
+
+This matching logic applies for captured exceptions, not raw message events. By default, all errors are sent.
+
+If your scripts are loaded from `cdn.example.com` and your site is `example.com`, you can set `allowUrls` to the follwing to exclusively capture errors being created in scripts in these locations:
```
lizokm, line 9:
```suggestion
For example, if you add `'foo.com'` to the array, any errors created on `https://bar.com/myfile/foo.com` will be captured because the URL will be matched with "contains" logic (and the last segment of the URL contains `foo.com`).

This matching logic applies to captured exceptions not raw message events. By default, all errors are sent.

If your scripts are loaded from `cdn.example.com` and your site is `example.com`, you can set `allowUrls` to the following to only capture errors being created in scripts in these locations:
```

includes/platforms/configuration/options/deny-urls.mdx
```diff
@@ -1 +1,5 @@
-A list of strings or regex patterns that match error URLs that should not be sent to Sentry. Errors whose entire file URL contains (string) or matches (regex) at least one entry in the list will not be sent. As a result, if you add `'foo.com'` to the list, it will also match on `https://bar.com/myfile/foo.com`. Keep in mind that this only applies for captured exceptions, not raw message events. By default, all errors are sent.
+An array of strings or regex patterns that match the URLs of scripts where errors have been created.
+Only errors that have been created on these URLs will not be sent sent to Sentry.
+If you use this option, Errors whose top stack frame file URL contains (string) or matches (regex) at least one entry in the array will not be sent.
+
+This matching logic applies for captured exceptions, not raw message events. By default, all errors are sent.
```
lizokm, line 5:
```suggestion
An array of strings or regex patterns that match the URLs of scripts where errors have been created. Errors that have been created on these URLs won't be sent to Sentry.
If you use this option, at least one entry in the array in which an error's top stack frame file URL contains (string) or matches (regex) won't be sent.

This matching logic applies to captured exceptions not raw message events. By default, all errors are sent.
```

lizokm (approved):
Thanks for adding this. I have a few small tweaks, but otherwise this looks great!

includes/platforms/configuration/options/allow-urls.mdx
lforst: I think I know what you are getting at but the suggestion makes the content logically wrong. I took another stab: 9275103e631d1f295b24fc4dc2579d5f4abb2800

includes/platforms/configuration/options/deny-urls.mdx
lforst: Also took another stab here: b711447a0d9e72b20df090acf99cefbc063b5465

Issue #10491, opened 2024-06-25
Unable to bump Next.js to the newest version


lizokm:
### URL

Unable to update b/c we are hitting the max lambda size which is 300mb with 440mb

### Additional Info

Also affects this: https://github.com/getsentry/sentry-docs/pull/9987

PR #10497, opened 2024-06-26
deps: Bump next and related dependencies

lforst:
When bumping next, Vercel's NFT logic seems to have changed and we were pulling in weird additional files like `.git/objects` and webpack caches. We identified that this was due to `mdx.ts` using `process.cwd()` and path concatenations which was confusing NFT, making it think pretty much everything in the root of the repo was needed to run the app. This caused us to be over the 300 MB limit for AWS lambdas.

Entails:
- All kinds of related patch and minor bumps to Next.js and MDX
- Minor QoL improvement in vercel.json
- A more statically analyzable `src/mdx.ts` file

Fixes https://github.com/getsentry/sentry-docs/issues/10491


lforst:
From 180MB

Screenshot 2024-06-26 at 12 52 29

to 15MB

Screenshot 2024-06-26 at 12 52 38

package.json
```diff
@@ -24,7 +24,8 @@
     "lint:prettier:fix": "prettier --write \"./{src,app,scripts}/**/*.{md,mdx,ts,tsx,js,jsx,mjs}\"",
     "lint:fix": "yarn run lint:prettier:fix && yarn run lint:eslint:fix",
     "sidecar": "yarn spotlight-sidecar",
-    "test": "jest"
+    "test": "jest",
+    "build-trace": "npx next@canary internal turbo-trace-server .next/trace"
```
chargome, line 28: Should we keep that?

chargome (approved):
🚀

chargome:
Fixes #10491

src/mdx.ts
```diff
@@ -96,8 +98,8 @@ async function getDocsFrontMatterUncached(): Promise<FrontMatter[]> {
   return frontMatter;
 }
 
-export function getAllFilesFrontMatter(folder: string = 'docs') {
```
a-hariti, line 99: this `folder` shouldn't be hardcoded as we need to pass it as a parameter for developer docs for example later on

lforst: Right now it doesn't need to be. We can add it back if needed. We shouldn't add code for the mere possibility of needing it.

a-hariti: I just mentioned it since this is deleting something that was already there, but I see what you mean

oioki (approved):

PR #10508, opened 2024-06-26
demo of code block padding

szokeasaurusrex:


szokeasaurusrex:
Adding the WIP label to prevent closing (this PR itself is not WIP, but helps demonstrate behavior which should be fixed)

PR #10511, opened 2024-06-26
deps: Bump `next` and MDX

lforst:
A less aggressive version of https://github.com/getsentry/sentry-docs/pull/10497 bumping Next.js and the related MDX dependency.

We still need to remove the non/statically analysible path concatenation.

Bumping next caused an error to be thrown in the changelog about caching dynamic `header()` calls which is not allowed - makes sense I think and even seems like a security vulnerability tbh.


chargome (approved):

src/mdx.ts
```diff
@@ -96,8 +96,8 @@ async function getDocsFrontMatterUncached(): Promise<FrontMatter[]> {
   return frontMatter;
 }
 
-export function getAllFilesFrontMatter(folder: string = 'docs') {
-  const docsPath = path.join(root, folder);
```
lforst, line 100:
```
function getAllFilesFrontMatter(folder: 'docs' | 'develop') {
	let docsPath;
	if (folder === 'docs') {
	  docsPath = path.join(root, 'docs');
	} else {
	  docsPath = path.join(root, 'develop');
	}
```

@a-hariti If we need to parameterize this function in the future we can do the above so that NFT continues to be able to analyze what is required for this file.

PR #10530, opened 2024-06-27
fix: Fix custom otel docs with correct imports

mydea:
Forgot about something there, small adjustments to the otel docs.


lforst (approved):

mydea:
You are right, I opened an issue for this, we'll fix this in the migration docs as well!

PR #10533, opened 2024-06-27
Ignore the whole /changelog before Algolia index build

a-hariti:
I noticed the Algolia index script is failing lately after adding a `_admin/create` under `/changelog`

this entails deleting the whole `/changelog` directory before the build


lizokm (approved):

lizokm (approved):
Thanks for catching this!

PR #10534, opened 2024-06-27
fix focus hijacking on mobile search input

### LINKED ISSUE #10471: Search input breaks on mobile
[Feedback from Slack]

The search input completely breaks on mobile.
Why are we hijacking keypresses?
I type s and it selects sandbox rather than letting me enter a query

a-hariti:
closes #10471


stephanie-anderson (approved):
Thank you for fixing this, @a-hariti!

PR #10539, opened 2024-06-27
fix(insights): Clarify Requests Insights span data requirements

gggritso:
- Add linebreak after arcade
- Link out to exact supported platforms. There are only _two_, so using the redirect was a bad experience
- Link to the developer documentation which has full explanation for manual instrumentation


vivianyentran (approved):
Thanks for keeping these docs up to date!

gggritso:
@vivianyentran thanks for the suggestions! 👍🏻

PR #10605, opened 2024-07-08
Add React Component Tracking and Error Boundary to React Native docs pages

krystofwoldrich:
This PR adapts React browser pages for React Native:

- https://docs.sentry.io/platforms/javascript/guides/react/features/component-monitoring/
- https://docs.sentry.io/platforms/javascript/guides/react/features/error-boundary/

This was done in preparation for a new page for https://docs.sentry.io/platforms/javascript/guides/react/features/component-names/ which for React Native differes in significant way due to configuration thru Metro.


docs/platforms/react-native/integrations/error-boundary.mdx
```diff
@@ -0,0 +1,194 @@
...
+
+`beforeCapture` (Function)
+
+A function that gets called before an error is sent to Sentry, allowing for extra tags or context to be added to the error.
+
+## Examples
+
+#### Setting a Fallback Function (Render Props)
+
+Below is an example where a fallback prop, using the [render props approach](https://react.dev/reference/react/cloneElement#passing-data-with-a-render-prop), is used to display a fallback UI on error. The fallback UI returns to a standard component state when reset using the `resetError()` API provided by the component through render props.
```
lizokm, line 109:
```suggestion
Below, is an example where a fallback prop, using the [render props approach](https://react.dev/reference/react/cloneElement#passing-data-with-a-render-prop), is used to display a fallback UI on error. The fallback UI returns to a standard component state when reset using the `resetError()` API, provided by the component through render props.
```

docs/platforms/react-native/integrations/error-boundary.mdx
```diff
@@ -0,0 +1,194 @@
...
+    );
+  }
+}
+
+export default MyComponent;
+```
+
+#### Using multiple error boundaries
+
+When using multiple error boundaries, we recommend using `beforeCapture` to set tags/context so that you can tell which error boundary the error occurred from. In the example below, we attach tags to errors based on what route they rendered in.
```
lizokm, line 164:
```suggestion
When using multiple error boundaries, we recommend using `beforeCapture` to set tags/context so that you can tell which error boundary the error occurred from. In the example below, we attach tags to errors based on what route they rendered in:
```

lizokm (approved):
Thanks so much for adding this, it looks great! I've left a few wording suggestions.

docs/platforms/react-native/integrations/component-monitoring.mdx
```diff
@@ -0,0 +1,77 @@
...
+export default Sentry.withProfiler(App, { name: "CustomAppName" });
+```
+
+`name` (string)
+
+The name of the component being profiled. By default, the name is taken from the component `displayName` property or the component `name` property.
+
+`includeRender` (boolean)
+
+If a `ui.react.render` span should be created by the Profiler. Set as true by default.
```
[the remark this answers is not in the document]
krystofwoldrich: Thanks for asking, yes, it make sence.

PR #10626, opened 2024-07-09
add warning to project transfer

souredoutlook:
Project Transfer has many limitations - we should add this warning and enumerate the other limitations

@leedongwei can we please get a full list of project transfer limitations and add them to this PR?

### DIFF
--- a/docs/organization/membership/index.mdx
+++ b/docs/organization/membership/index.mdx
@@ -127,6 +127,12 @@ If "Open Membership" is toggled off, team access is on an invite-only basis. Use
 
 ## Transfer a Project
 
+<Note>
+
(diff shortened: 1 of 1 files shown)


docs/organization/membership/index.mdx
```diff
@@ -127,6 +127,12 @@ If "Open Membership" is toggled off, team access is on an invite-only basis. Use
 
 ## Transfer a Project
 
+<Note>
+
+When transferring projects some project data such as releases, and session data are not transferred. Please also note that Organization Settings such as team assignments, and code-mappings will need to be applied to the project again.
```
lizokm, line 132:
```suggestion
Some project data such as releases and session data won't be transferred during project transfer. In addition, Organization Settings such as team assignments, and code-mappings will need to be specified for the project again. 
```

lizokm (approved):
Thanks for adding this!

PR #10644, opened 2024-07-11
PHP updates (Caches, Queues, Requests)

cleptric:


HazAT (approved):

kayw-geek (commented):
I noticed the `SENTRY_TRACES_SAMPLE_RATE ` has no effect, because the `couldHavePerformanceTracingEnabled()` method has not been called.[BaseServiceProvider.php](https://github.com/getsentry/sentry-laravel/blob/f987314955bb08d96520c4e02943542540d8f3db/src/Sentry/Laravel/BaseServiceProvider.php#L61-L61)

PR #10647, opened 2024-07-11
fix: Add docs for v7 loader integrations

mydea:
Fixes https://github.com/getsentry/sentry-javascript/issues/12758

I updated the docs for using pluggable integrations with the loader to the v8 syntax, however the loader is not yet even fully out with v8 support, which leads to confusion and problems.

This PR ensures we have docs for both v7 and v8 of the loader, so this should work both ways.

### DIFF
--- a/platform-includes/configuration/capture-console/javascript.mdx
+++ b/platform-includes/configuration/capture-console/javascript.mdx
@@ -15,9 +15,8 @@ Sentry.init({
   window.sentryOnLoad = function () {
     Sentry.init({});
 
-    const integration = Sentry.captureConsoleIntegration?.();
-    if (integration) {
-      Sentry.addIntegration(integration);
+    if (Sentry.captureConsoleIntegration) {
+      Sentry.addIntegration(Sentry.captureConsoleIntegration());
     }
   };
 </script>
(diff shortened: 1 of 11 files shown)


platform-includes/configuration/capture-console/javascript.mdx
```diff
@@ -9,7 +9,31 @@ Sentry.init({
...
+```html {tabTitle: Loader (v7)}
+<script>
+  // Configure sentryOnLoad before adding the Loader Script
+  window.sentryOnLoad = function () {
+    Sentry.init({});
+
+    const integration = Sentry.captureConsoleIntegration?.();
+    if (integration) {
+      Sentry.addIntegration(integration);
+    }
```
lforst, line 21: Optional chaining is not that widely supported yet I believe 

```suggestion
    if (Sentry.captureConsoleIntegration) {
      Sentry.addIntegration(Sentry.captureConsoleIntegration());
    }
```

lforst (approved):

platform-includes/configuration/capture-console/javascript.mdx
mydea: It's a good point, I rewrote all like this!

PR #10666, opened 2024-07-11
chore(codeowners): Add owners-api to `resolveOpenAPI.ts`

iamrajjoshi:
Docs Team gets a bunch of PRs such as https://github.com/getsentry/sentry-docs/pull/10657. 

We should always approve, but for a sanity check lets make sure owners-api approve these prs.


lizokm (approved):
Thanks for updating!

PR #10673, opened 2024-07-12
fix dev docs search results priority;

a-hariti:
please rename `DEVELOPER_DOCS=1` to `NEXT_PUBLIC_DEVELOPER_DOCS=1` before the merge

both on Github secrets and on Vercel

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -1,7 +1,7 @@
 import type {NextRequest} from 'next/server';
 import {NextResponse} from 'next/server';
 
-// This env var is set in next.config.js based on the `DEVELOPER_DOCS` env var at build time
+// This env var is set in next.config.js based on the `NEXT_PUBLIC_DEVELOPER_DOCS` env var at build time
 // a workaround edge middleware not having access to env vars
 const isDeveloperDocs = process.env.DEVELOPER_DOCS_;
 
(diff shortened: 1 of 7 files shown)


chargome (changes requested):
@a-hariti Thank you! We should also update the env var in our algolia github action.

a-hariti:
Yes, that's what I said on the PR description

chargome:
I meant the actual workflow: `algolia-index.yml`

chargome (approved):

chargome:
Note: I have added the env var in vercel, still need to confirm the github secret update by @stephanie-anderson

src/middleware.ts
```diff
@@ -1,7 +1,7 @@
 import type {NextRequest} from 'next/server';
 import {NextResponse} from 'next/server';
 
-// This env var is set in next.config.js based on the `DEVELOPER_DOCS` env var at build time
+// This env var is set in next.config.js based on the `NEXT_PUBLIC_DEVELOPER_DOCS` env var at build time
 // a workaround edge middleware not having access to env vars
 const isDeveloperDocs = process.env.DEVELOPER_DOCS_;
```
stephanie-anderson, line 6: @a-hariti is the underscore intentional?

stephanie-anderson: And should it also be renamed?

chargome: @stephanie-anderson this is intentional since this env var is being set in `next.config.js`

a-hariti: yes, it is

That's a workaround to avoid conflict with the actual env var without an `_`

PR #10714, opened 2024-07-15
update repos info on developer docs

a-hariti:
update [this](https://develop.sentry.dev/development/docs/#repositories)


develop-docs/development/docs/index.mdx
```diff
@@ -13,9 +13,7 @@ Sentry's documentation lives in numerous repositories, so what we're covering he
 Key places which contain documentation are:
 
 - [getsentry/sentry-docs](https://github.com/getsentry/sentry-docs) - [docs.sentry.io](https://docs.sentry.io/)
-- [getsentry/develop](https://github.com/getsentry/develop) - these docs!
-
-Our goal is to eventually unify the platform used for docs, which may result in us merging the above repositories.
+- [getsentry/develop](https://github.com/getsentry/sentry-docs/tree/master/develop-docs) - these docs!
```
chargome, line 16: Might be confusing for users, should we communicate that we have only one docs repo with two content dirs respectively?

PR #10729, opened 2024-07-16
feat(js): Add docs for creating a child span

mydea:
This adds docs for the `parentSpan` option we added somewhat recently to the SDK, which should make it easier to understand. It also adds a dedicated section to custom instrumentation docs to explain this, as it is a rather common scenario.

Ref https://github.com/getsentry/sentry-javascript/pull/12928

### DIFF
--- a/docs/platforms/javascript/common/tracing/instrumentation/custom-instrumentation/index.mdx
+++ b/docs/platforms/javascript/common/tracing/instrumentation/custom-instrumentation/index.mdx
@@ -72,14 +72,15 @@ This will revert to use the full hierarchy behavior, where spans are children of
 
 The following options can be used for all span starting functions:
 
-| Option             | Type                        | Description                                                         |
-| ------------------ | --------------------------- | ------------------------------------------------------------------- |
-| `name`             | `string`                    | The name of the span.                                               |
-| `op`               | `string`                    | The operation of the span.                                          |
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/common/tracing/instrumentation/custom-instrumentation/index.mdx
```diff
@@ -101,6 +102,20 @@ To add spans that aren't active, you can create independent spans. This is usefu
 
 <PlatformContent includePath="performance/start-inactive-span" />
 
+## Starting Spans as Children of a Specific Span
```
Lms24, line 105: Are we guarding this paragraph for node guides? AFAIK this won't work in browser SDK where we have a flat span hierarchy, right?

Lms24 (approved):

docs/platforms/javascript/common/tracing/instrumentation/custom-instrumentation/index.mdx
mydea: hmm, good point, I mean it will kind of work, if the span is a root span for example. I would leave this as is for now, we can maybe clarify this even further if needed!

Lms24: ok sounds good to me!

PR #10740, opened 2024-07-16
replace mdx imports with their actual content on dev docs

a-hariti:
this is not ideal, it leads to some content duplication, but should fix some failing imports for now


chargome (approved):

a-hariti:
wyt @stephanie-anderson ?

PR #10745, opened 2024-07-16
fix: follow non platform redirects on platform-redirect

a-hariti:
when no platforms applies to the `next` query param, try to see if it's a valid url on it's own


chargome:
I don't quite understand the value of this function, since the native `redirect` will basically just do the same?

a-hariti:
I think you got a point @chargome, just redirect and let the app take care of it

damn, I just spent 30mins on a no op 😂

chargome:
But the recursive func was a nice idea! 

Closing for now

PR #10753, opened 2024-07-16
fix: Use /orgredirect/ instead of linking to sentry org

ryan953:
Test plan:
1. Login to an org that is not sentry.sentry.io
   Example: demo.sentry.io
3. Visit the links, you should be redirected to `demo.sentry.io/**`


lizokm (approved):
Thanks for updating this, Ryan!

PR #10759, opened 2024-07-17
feat(js): Improve `tracesSampler` docs for JS

mydea:
Related to https://github.com/getsentry/sentry-javascript/pull/12945

This updates the JS docs around sampling with `tracesSampler`. First, we had two redundant sections about the sampling conrtext - I condended this into one section.

Then, I also explained further that the `name` may not be the final name, because this may be confusing, and also added an example (which I think is the most common one people may run into).


Lms24 (approved):
Nice!

PR #10771, opened 2024-07-18
fix(javascript): Add entry to migration guide about forcing a sampling decision

Lms24:
Our migration guide didn't mention the removal of `sampled: boolean` at all previously. This PR adds an entry how to migrate forcing a sampling decision. 
Since we already mention `tracesSampler` in the docs and how to return a definitely positive sampling decision from it, I think we don't need an extra "Forcing Sampling Decision" guide in the regular docs.

### DIFF
--- a/docs/platforms/javascript/common/migration/v7-to-v8/v8-new-performance-api.mdx
+++ b/docs/platforms/javascript/common/migration/v7-to-v8/v8-new-performance-api.mdx
@@ -252,3 +252,29 @@ Sentry.startSpan((span: Span) => {
   span.setAttribute("attr", 1);
 });
 ```
+
+### Forcing a sampling decision
+
+Previously in v7, you could force a positive or negative sampling decision when calling `startTransaction` by setting the `sampled` option.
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/common/migration/v7-to-v8/v8-new-performance-api.mdx
```diff
@@ -252,3 +252,29 @@ Sentry.startSpan((span: Span) => {
...
+Sentry.startTransition({op: 'function.myFunction', sampled: true});
+
+// v8
+// 1. define a tracesSampler
+Sentry.init({
+  tracesSampler: (samplingContext) => {
+    if (samplingContext.op === 'function.myFunction') {
+      return 1;
+    }
+    return 0.1;
```
andreiborza, line 274: The text above talks about `1` or `0`, but there's a value in-between used here. Is that on purpose? I'd probably rewrite the paragraph above if so.

Lms24: I added an example for a positive (1) and negative (0) decision, as well as a comment to show that the 0.1 is used as a default sample rate

andreiborza (approved):

Issue #10747, opened 2024-07-16
404s within develop docs


a-hariti:
External links with 404 on dev docs

http://localhost:3000/frontend
- [x] [`Flux`](https://facebook.github.io/flux/) `(https://facebook.github.io/flux/)`
- [x] [`fixtures/js-stubs/`](https://github.com/getsentry/sentry/tree/master/fixtures/js-stubs) `(https://github.com/getsentry/sentry/tree/master/fixtures/js-stubs)`

http://localhost:3000/integrations/azuredevops
- [x] [`documentation on installing and configuring the Azure DevOps integration`](https://docs.sentry.io/product/integrations/azure-devops/) `(https://docs.sentry.io/product/integrations/azure-devops/)`

http://localhost:3000/integrations/msteams
- [x] [`documentation on using the Microsoft Teams integration`](https://docs.sentry.io/product/integrations/msteams/) `(https://docs.sentry.io/product/integrations/msteams/)`

http://localhost:3000/sdk/overview
- [x] [`/unreal/`](https://docs.sentry.io/platforms/native/ue4/) `(https://docs.sentry.io/platforms/native/ue4/)`

http://localhost:3000/sdk/event-payloads
- [x] [`source code defining the protocol as understood by the server`](https://github.com/getsentry/relay/tree/master/relay-general/src/protocol) `(https://github.com/getsentry/relay/tree/master/relay-general/src/protocol)`

http://localhost:3000/sdk/event-payloads/contexts
- [x] [`Native SDK repository`](https://github.com/getsentry/sentry-native/blob/master/tests/fixtures/os_releases/distribution_names.txt) `(https://github.com/getsentry/sentry-native/blob/master/tests/fixtures/os_releases/distribution_names.txt)`

http://localhost:3000/sdk/performance
- [x] [`@sentry/tracing (https://github.com/getsentry/sentry-javascript/tree/master/packages/tracing)`](JavaScript) `(JavaScript)`

http://localhost:3000/sdk/performance/span-operations
- [x] [`Operations Breakdown`](https://docs.sentry.io/product/sentry-basics/tracing/event-detail/#operations-breakdown) `(https://docs.sentry.io/product/sentry-basics/tracing/event-detail/#operations-breakdown)`

http://localhost:3000/application/dynamic-sampling/architecture
[...]

PR #10772, opened 2024-07-18
fix(develop-docs): Fix remaining 404s

chargome:
relates to #10747

note: remaining links on `/application/dynamic-sampling/architecture` will be updated in #10773 since the information is outdated


lizokm (approved):
Thank you!

PR #10776, opened 2024-07-18
Update reverse-proxy.mdx chapter nginx

londi:
## Description

Extended nginx reverse proxy description to avoid csrf errors on envelope requests.  This adjustment helps upcoming sentry selfhosted user to not run into the same problem as me.

Issue: https://github.com/getsentry/self-hosted/issues/3205

## IS YOUR CHANGE URGENT?  

Help us prioritize incoming PRs by letting us know when the change needs to go live.
- [ ] Urgent deadline (GA date, etc.): 
- [ ] Other deadline: 
- [x] None: Not urgent, can wait up to 1 week+

## LEGAL BOILERPLATE

Look, I get it. The entity doing business as "Sentry" was incorporated in the State of Delaware in 2015 as Functional Software, Inc. and is gonna need some rights from me in order to utilize my contributions in this here PR. So here's the deal: I retain all rights, title and interest in and to my contributions, and by keeping this boilerplate intact I confirm that Sentry can use, modify, copy, and redistribute my contributions, under Sentry's choice of terms.

## EXTRA RESOURCES

- [Sentry Docs contributor guide](https://docs.sentry.io/contributing/)

PR #10779, opened 2024-07-19
self-hosted: link to updated nginx configuration

joshuarli:
Supersedes https://github.com/getsentry/sentry-docs/pull/10776.

### DIFF
--- a/develop-docs/self-hosted/reverse-proxy.mdx
+++ b/develop-docs/self-hosted/reverse-proxy.mdx
@@ -42,53 +42,7 @@ Endpoint for health checks is available on `/_health/` endpoint using HTTP proto
 
 We recommend installing NGINX since that's what we are using on [sentry.io](https://sentry.io/).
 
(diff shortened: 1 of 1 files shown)


develop-docs/self-hosted/reverse-proxy.mdx
```diff
@@ -42,53 +42,7 @@ Endpoint for health checks is available on `/_health/` endpoint using HTTP proto
 
 We recommend installing NGINX since that's what we are using on [sentry.io](https://sentry.io/).
```
joshuarli, line 43: Actually uhh, just became aware we provide nginx in the default install, so probably have to change this wording too.

joshuarli: The configuration I removed is for an nginx deployment outside of the self-hosted docker compose environment, should probably keep that...

londi (approved):

PR #10784, opened 2024-07-19
docs(nestjs): Update nestjs sdk docs with new setup

nicohrubec:
Should be merged after the new nestjs setup is released, but prepared this in the meantime for review.

Setup still WIP and will be introduced with [this PR](https://github.com/getsentry/sentry-javascript/pull/12920)

### DIFF
--- a/platform-includes/getting-started-use/javascript.nestjs.mdx
+++ b/platform-includes/getting-started-use/javascript.nestjs.mdx
@@ -35,7 +35,7 @@ import { AppService } from './app.service';
 export class AppModule {}
 ```
 
-By default, exceptions with status code 4xx are not sent to Sentry. If you still want to capture these exceptions, you can do that manually with `Sentry.captureException()`:
+By default, exceptions with status code 4xx are not sent to Sentry. If you still want to capture these exceptions, you can do so manually with `Sentry.captureException()`:
 
 ```javascript {11}
 import { ArgumentsHost, BadRequestException, Catch } from '@nestjs/common';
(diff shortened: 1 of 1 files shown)


platform-includes/getting-started-use/javascript.nestjs.mdx
```diff
@@ -1,24 +1,33 @@
...
+<Note>The SentryIntegrationModule needs to be the first module registered in your application.</Note>
+
+```javascript {filename: app.module.ts} {7}
+import { Module } from '@nestjs/common';
+import { SentryIntegrationModule } from '@sentry/nestjs/setup';
+import { AppController } from './app.controller';
+import { AppService } from './app.service';
+
+@Module({
+  imports: [SentryIntegrationModule.forRoot()],
```
chargome, line 28:
```suggestion
  imports: [
      SentryIntegrationModule.forRoot(),
      // ...other modules
  ],
```

chargome (approved):

lforst:
I would like for us to document the behaviour if people have existing exception filters set up, and how they can opt their exception filters into reporting Sentry errors.

nicohrubec:
@lforst In which specific case? If people throw an exception from a custom exception filter that should work out of the box.

lforst:
@nicohrubec For example, if users registered a global exception filter that catches one type of exceptions (meaning those exceptions would not get captured by the SDK automatically), I would like us to document how they can still report those exceptions. i.e. by explaining that they would need to add a captureException to their custom exception filter or similar.

nicohrubec:
@lforst Sure makes sense, will add an additional example in the Use section

platform-includes/getting-started-use/javascript.nestjs.mdx
```diff
@@ -1,24 +1,57 @@
...
+      SentryModule.forRoot(),
+      // ...other modules
+  ],
+  controllers: [AppController],
+  providers: [AppService],
+})
+export class AppModule {}
+```
+
+By default, exceptions with status code 4xx are not sent to Sentry. If you still want to capture these exceptions, you can do that manually with `Sentry.captureException()`:
```
lforst, line 38:
```suggestion
By default, exceptions that are caught by your own exception filters are not reported to Sentry. If you still want to capture these exceptions, you can do so manually with `Sentry.captureException()`:
```

nicohrubec: @lforst I don't think that is accurate, because if you rethrow a 500 in an exception filter then that will still be reported. Only if you catch an exception and then throw a different exception the original exception won't be reported. So left the 4xx text but simplified the example, should be fine for users to get the gist

PR #10785, opened 2024-07-19
feat: Automatic preview cleanups

### LINKED ISSUE #10695: Stale preview deployments
### Problem Statement

We are running into issues of bandwidth usage by stale preview deployments.

### Solution Brainstorm

~~We should provide our own deployment pipeline by using [the vercel CLI via gh actions](https://vercel.com/guides/how-can-i-use-github-actions-with-vercel).~~

We will probably solve this by periodically running a script to remove old preview deployments.

chargome:
- Adds a script to cleanup stale preview deployments (mainly ported from #6592)
- Adds a gh workflow that calls this script every day for user docs and developer docs

closes #10695

### DIFF
--- a/.github/workflows/cleanup-preview-deployments.yml
+++ b/.github/workflows/cleanup-preview-deployments.yml
@@ -1,0 +1,34 @@
+name: Delete Stale Vercel Preview Deployments
+
+on:
+  # Run the workflow daily at 2:00 AM UTC
+  schedule:
+    - cron: '0 2 * * *'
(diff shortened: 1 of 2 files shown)


chargome:
Some notes:
- due to vercel rate limiting the first run took hours locally and deleted almost 10k stale preview deployments, but all consecutive runs should be finished fairly quickly
- we need to update/add the env vars used in the gh workflow before merging this

a-hariti (approved):
🚀

PR #10787, opened 2024-07-19
chore(docs): Deleting Threads Integration Docs

iamrajjoshi:
This company was bought out by another company, so this integration no longer exists. We should delete the docs for it.


schew2381 (approved):
lgtm

PR #10796, opened 2024-07-22
update docs to include is in

nhsiehgit:
Adds new `IS IN` filter condition

### DIFF
--- a/docs/product/alerts/create-alerts/issue-alert-config.mdx
+++ b/docs/product/alerts/create-alerts/issue-alert-config.mdx
@@ -59,8 +59,8 @@ Sentry checks "if conditions", or filters, after “When” conditions are met, 
 - The issue has happened at least \{X\} times.
 - The issue is assigned to \{no one/a team/a member\}.
 - The event is from the latest release.
-- The event's \{attribute\} \{matches\} \{value\}. Match types: equals, does not equal, starts with, ends with, contains, does not contain, is set, or is not set, is in, or is not in.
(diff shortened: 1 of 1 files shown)


docs/product/alerts/create-alerts/issue-alert-config.mdx
```diff
@@ -59,8 +59,8 @@ Sentry checks "if conditions", or filters, after “When” conditions are met,
 - The issue has happened at least \{X\} times.
 - The issue is assigned to \{no one/a team/a member\}.
 - The event is from the latest release.
-- The event's \{attribute\} \{matches\} \{value\}. Match types: equals, does not equal, starts with, ends with, contains, does not contain, is set, or is not set.
-- The event's \{tag\} \{matches\} \{value\}. Match types: equals, does not equal, starts with, ends with, contains, does not contain, is set, or is not set.
+- The event's \{attribute\} \{matches\} \{value\}. Match types: equals, does not equal, starts with, ends with, contains, does not contain, is set, or is not set, is in, or is not in.
+- The event's \{tag\} \{matches\} \{value\}. Match types: equals, does not equal, starts with, ends with, contains, does not contain, is set, or is not set, is in, or is not in.
```
cathteng, line 63:
```suggestion
- The event's \{attribute\} \{matches\} \{value\}. Match types: equals, does not equal, starts with, ends with, contains, does not contain, is set, is not set, is in, or is not in.
- The event's \{tag\} \{matches\} \{value\}. Match types: equals, does not equal, starts with, ends with, contains, does not contain, is set, is not set, is in, or is not in.
```

cathteng (approved):

Issue #10765, opened 2024-07-17
Maven dependency blocks contain poorly formatted `version` tags


liamjones:
### Steps to Reproduce

View a Java page including a Maven dependency block such as https://docs.sentry.io/platforms/java/guides/spring-boot/ or https://docs.sentry.io/platforms/java/guides/log4j2/

Try to copy/paste the block into a pom.xml (either manually or via the copy to clipboard button)

### Expected Result

```xml
<dependency>
    <groupId>io.sentry</groupId>
    <artifactId>sentry-spring-boot-starter</artifactId>
    <version>7.12.0</version>
</dependency>

```

### Actual Result

```xml
<dependency>
    <groupId>io.sentry</groupId>
    <artifactId>sentry-spring-boot-starter</artifactId>
    <version
  >7.12.0</version>
</dependency>
```

### Additional Info

The tag isn't strictly broken, just ugly. 😄

PR #10806, opened 2024-07-23
fix: Exclude code snippets with injects from formatting

chargome:
Skip codesnippets that use a `@inject` value from formatting as they lead to unintended linebreaks

fixes [#10765](https://github.com/getsentry/sentry-docs/issues/10765)

### DIFF
--- a/src/remark-format-code.js
+++ b/src/remark-format-code.js
@@ -11,8 +11,14 @@ export default function remarkFormatCodeBlocks() {
       // skip code blocks with diff meta as they might have
       // broken syntax due to + and - characters
       // or with `onboardingOptions` as they need to have predictable line numbers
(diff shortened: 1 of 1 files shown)


src/remark-format-code.js
```diff
@@ -11,8 +11,14 @@ export default function remarkFormatCodeBlocks() {
       // skip code blocks with diff meta as they might have
       // broken syntax due to + and - characters
       // or with `onboardingOptions` as they need to have predictable line numbers
+      // or @inject statements as they ofeten lead to unnecessary line breaks
```
andreiborza, line 14:
```suggestion
      // or `@inject` statements as they ofeten lead to unnecessary line breaks
```

andreiborza (approved):

PR #10818, opened 2024-07-24
ref(changelog): Duplicate changelog into workspace package

lforst:
This is the first step of separating the changelog from the docs.

This PR simply duplicates the changelog with all of its dependencies into a workspace package, so that it is completely separated from the logic of the docs.

Ref https://github.com/getsentry/sentry-docs/issues/10819


package.json
```diff
@@ -10,6 +10,10 @@
   "bugs": {
     "url": "https://github.com/getsentry/sentry-docs/issues"
   },
+  "private": true,
+  "workspaces": [
+    "./packages/*"
+  ],
```
chargome, line 16: l: For our target architecture we could think of moving changelog, user-docs and dev-docs to `./apps/*` and all shared code (UI, tsconfig, eslint, tailwind,...) to `./packages`

chargome (approved):
LGTM !

PR #10837, opened 2024-07-25
ref(react): Improve React Router instrumentation docs

Lms24:
While creating a small react router test app, I tried to apply our RR6 routing instrumentation and ran into every possible pitfall along the way. First, I imported the hooks from the wrong packages, then I missed that I need to apply `wrapCreateBrowserRouter` and ultimately (not my proudest moment lol) missed calling `sentryCreateBrowserRouter`.

Because IMO it's really easy to fall into at least one of these problems, I thought, our code snippet highlighting feature would work well here because the setup is (compared to other routing instrumentations) quite involved. So this PR adds some highlighting to the various RR version code snippets and slightly streamlines wording.

Example:
image

Note: All snippet highlights assume that `Sentry.init` was already set up beforehand which I think is a fair assumption, given this is a guide on a sub-page.

### DIFF
--- a/docs/platforms/javascript/guides/react/features/react-router.mdx
+++ b/docs/platforms/javascript/guides/react/features/react-router.mdx
@@ -23,7 +23,7 @@ _(Available in version 7 and above)_
 
 To use React Router v6 with Sentry:
 
-Add `Sentry.reactRouterV6BrowserTracingIntegration` instead of the regular `Sentry.browserTracingIntegration` and provide the required react functions:
+Add `Sentry.reactRouterV6BrowserTracingIntegration` instead of the regular `Sentry.browserTracingIntegration` and provide the required React hooks and router functions:
 
     - `useEffect` hook from `react`
     - `useLocation` and `useNavigationType` hooks from `react-router-dom`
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/guides/react/features/react-router.mdx
```diff
@@ -23,15 +23,15 @@ _(Available in version 7 and above)_
 
 To use React Router v6 with Sentry:
 
-Add `Sentry.reactRouterV6BrowserTracingIntegration` instead of the regular `Sentry.browserTracingIntegration` and provide the functions it needs to enable performance tracing:
+Add `Sentry.reactRouterV6BrowserTracingIntegration` instead of the regular `Sentry.browserTracingIntegration` and provide the required react functions:
```
s1gr1d, line 26:
```suggestion
Add `Sentry.reactRouterV6BrowserTracingIntegration` instead of the regular `Sentry.browserTracingIntegration` and provide the required React hooks and router functions:
```

L: Everything with a `use` is called a hook in React, so I don't know if people find it weird if we just call it "function" (technically it is one though).

s1gr1d (approved):

docs/platforms/javascript/guides/react/features/react-router.mdx
Lms24: sounds good, thx!

Issue #10844, opened 2024-07-25
Fix python attachments docs for 2.x


sl0thentr0py:
### Core or SDK?

Platform/SDK

### Which part? Which one?

Python SDK

### Description

Attachment docs are incorrect after the new scope apis.
`add_to_transactions` parameter is also undocumented.

See discord convo

![Image](https://github.com/user-attachments/assets/e79d7c04-55a4-4dda-863d-5bf04838eb78)

![Image](https://github.com/user-attachments/assets/88c6c7c1-d347-47cd-9018-8944697aaadd)

PR #10851, opened 2024-07-25
docs(python): Update attachment docs for 2.x

szokeasaurusrex:
## DESCRIBE YOUR PR
Update attachment docs to reflect new API added in Python SDK 2.0.0, and introduce other general improvements to the page.

Depends on https://github.com/getsentry/sentry-python/pull/3342, which adds parameter documentation to the API Docs.

Fixes: #10844
Related: getsentry/sentry-python#3340

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## PRE-MERGE CHECKLIST

### DIFF
--- a/platform-includes/enriching-events/add-attachment/python.mdx
+++ b/platform-includes/enriching-events/add-attachment/python.mdx
@@ -1,4 +1,4 @@
-Because attachments live on the [`Scope`](/platforms/python/enriching-events/scopes/), the SDK sends all attachments on a given `Scope` with every non-transaction event captured within the `Scope`. It's also possible to send an attachment with all events, including transactions by setting the `add_to_transactions` option to `True`.
+Because attachments live on the <PlatformLink to="/enriching-events/scopes/">Scope</PlatformLink>, the SDK sends all attachments on a given `Scope` with every non-transaction event captured within the `Scope`. It's also possible to send an attachment with all events, including transactions by setting the `add_to_transactions` option to `True`.
 
 ```python
 # Add an attachment
--- a/platform-includes/enriching-events/add-attachment/python.mdx
+++ b/platform-includes/enriching-events/add-attachment/python.mdx
@@ -1,4 +1,4 @@
-Attachments live on the [`Scope`](/platforms/python/enriching-events/scopes/). The SDK sends all attachments on a given `Scope` with every non-transaction event captured within the `Scope`. Optionally, an attachment can be sent with all events, including transactions, when the attachment is added with the `add_to_transactions` option set to `True`.
+Because attachments live on the [`Scope`](/platforms/python/enriching-events/scopes/), the SDK sends all attachments on a given `Scope` with every non-transaction event captured within the `Scope`. It's also possible to send an attachment with all events, including transactions by setting the `add_to_transactions` option to `True`.
 
 ```python
 # Add an attachment
--- a/platform-includes/enriching-events/add-attachment/python.mdx
+++ b/platform-includes/enriching-events/add-attachment/python.mdx
@@ -1,28 +1,27 @@
-Attachments live on the `Scope` and will be sent with all events.
+Attachments live on the [`Scope`](/platforms/python/enriching-events/scopes/). The SDK sends all attachments on a given `Scope` with every non-transaction event captured within the `Scope`. Optionally, an attachment can be sent with all events, including transactions, when the attachment is added with the `add_to_transactions` option set to `True`.
 
 ```python
 # Add an attachment
-from sentry_sdk import configure_scope
+from sentry_sdk import Scope
 
-with configure_scope() as scope:
-    scope.add_attachment(bytes=b"Hello World!", filename="attachment.txt")
-    scope.add_attachment(path="/path/to/attachment/file.txt")
+# Scope.get_current_scope() or Scope.get_global_scope() can also be used.
+# Read about the different scopes on our Scope docs page.
+scope = Scope.get_isolation_scope()
+
+# Add attachment titled "attachment.txt" with content "Hello World!"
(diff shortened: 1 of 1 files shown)


platform-includes/enriching-events/add-attachment/python.mdx
```diff
@@ -1,28 +1,22 @@
...
-    scope.add_attachment(bytes=b"Hello World!", filename="attachment.txt")
-    scope.add_attachment(path="/path/to/attachment/file.txt")
-```
-
-An attachment has the following fields:
-
-`path`
+# Scope.get_isolation_scope() or Scope.get_global_scope() can also be used.
+# Read about the different scopes on our Scope docs page.
+scope = Scope.get_current_scope()
```
sl0thentr0py, line 9: I'd actually make this `get_isolation_scope` since for most cases attachments need to be scoped (like tags etc) to the isolation scope

sl0thentr0py: further, maybe we should also just expose `add_attachment` that adds it to isolation scope on the top level api now

platform-includes/enriching-events/add-attachment/python.mdx
```diff
@@ -1,28 +1,22 @@
...
-
-The filename is required if using `bytes` and will be displayed in [sentry.io](https://sentry.io).
+# This attachment will also be sent with transactions
+scope.add_attachment(path="/other/attachment.txt", add_to_transactions=True)
+```
 
-`content_type`
 
-The type of content stored in this attachment. Any [MIME type](https://www.iana.org/assignments/media-types/media-types.xhtml) may be used; the default is `application/octet-stream`.
+Our [API Docs](https://getsentry.github.io/sentry-python/apidocs.html#sentry_sdk.Scope.add_attachment) describe the `add_attachment` method's parameters in detail.
```
sl0thentr0py, line 22: we've in the past kept these docs isolated from apidocs (no other references to apidocs in `sentry-docs`) and not really advertized them so I wouldn't start now.
Can you keep the original parameter descriptions here and add `add_to_transactions`? You can keep the examples you added above, they're useful.

sl0thentr0py (changes requested):

platform-includes/enriching-events/add-attachment/python.mdx
szokeasaurusrex: > we've in the past kept these docs isolated from apidocs (no other references to apidocs in sentry-docs) and not really advertized them so I wouldn't start now.

Huh, I did not know that. There is actually [one other reference I know of in the apidocs](https://docs.sentry.io/platforms/python/tracing/instrumentation/custom-instrumentation/#add-a-transaction) (which I added since I did not know that we did not want to advertise this).

Why do we not want to link to the API Docs? Imo it is redundant to have the same information (in this case, the parameters) documented in multiple places, and it increases maintenance burden, since any changes need to be documented in two places. 

And if we never really advertise that the API Docs exist, why do we even maintain them? I think we should either reference those docs or delete them.

sl0thentr0py: because nobody maintained the apidocs and they are wayyy outdated.

antonpirker: I showed them some love some time ago. But stopped because wanted to wait until the Hub is gone to really revamp them (and add versioned docs) 
I still think having versioned api docs is the way to go, 
but it is also true, that in the current state they are not useful at all

sl0thentr0py (approved):
ty!

szokeasaurusrex:
Hello, could we please get a @getsentry/docs review here? Thanks in advance.

platform-includes/enriching-events/add-attachment/python.mdx
```diff
@@ -1,28 +1,27 @@
-Attachments live on the `Scope` and will be sent with all events.
+Attachments live on the [`Scope`](/platforms/python/enriching-events/scopes/). The SDK sends all attachments on a given `Scope` with every non-transaction event captured within the `Scope`. Optionally, an attachment can be sent with all events, including transactions, when the attachment is added with the `add_to_transactions` option set to `True`.
```
lizokm, line 1:
```suggestion
Because attachments live on the [`Scope`](/platforms/python/enriching-events/scopes/), the SDK sends all attachments on a given `Scope` with every non-transaction event captured within the `Scope`. It's also possible to send an attachment with all events, including transactions by setting the `add_to_transactions` option to `True`.
```

lizokm (approved):
A small wording suggestion, otherwise this looks great. Thanks for updating!

platform-includes/enriching-events/add-attachment/python.mdx
```diff
@@ -1,28 +1,27 @@
-Attachments live on the `Scope` and will be sent with all events.
+Because attachments live on the [`Scope`](/platforms/python/enriching-events/scopes/), the SDK sends all attachments on a given `Scope` with every non-transaction event captured within the `Scope`. It's also possible to send an attachment with all events, including transactions by setting the `add_to_transactions` option to `True`.
```
lforst, line 1: you very likely want a `<PlatformLink>` here instead of the plain md link

szokeasaurusrex: It's not necessary for Python because we don't have guides for different platforms, but probably a good idea to do this

antonpirker: made platformlink. takes the same time as commenting :-)

szokeasaurusrex: Thanks @antonpirker 🥔

PR #10854, opened 2024-07-25
fix(.NET): Removed deprecated `EnableTracing`

bitsandfoxes:
The .NET SDK has deprecated `EnableTracing` from the options in favor of relying on `TracesSampleRate` [here](https://github.com/getsentry/sentry-dotnet/pull/3381).


lizokm (approved):
Thanks for updating!

PR #10855, opened 2024-07-25
fix(.NET): Added `BeforeSendTransaction` to common options

bitsandfoxes:
Since it was missing. And it's common.

### DIFF
--- a/docs/platforms/dotnet/common/configuration/options.mdx
+++ b/docs/platforms/dotnet/common/configuration/options.mdx
@@ -228,9 +228,9 @@ The callback typically gets a second argument (called a "hint") which contains t
 
 <ConfigKey name="before-send-transaction">
 
-This function is called with an SDK-specific transaction object, and can return a modified transaction object, or `null` to skip reporting the transaction. This can be used, for instance, for manual PII stripping before sending.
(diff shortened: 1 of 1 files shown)


docs/platforms/dotnet/common/configuration/options.mdx
```diff
@@ -226,6 +226,14 @@ The callback typically gets a second argument (called a "hint") which contains t
 
 </ConfigKey>
 
+<ConfigKey name="before-send-transaction">
+
+This function is called with an SDK-specific transaction object, and can return a modified transaction object, or `null` to skip reporting the transaction. This can be used, for instance, for manual PII stripping before sending.
+
+By the time <PlatformIdentifier name="before-send-transaction" /> is executed, all scope data has already been applied to the event. Further modification of the scope won't have any effect.
```
lizokm, line 233:
```suggestion

This function is called with an SDK-specific transaction object, and can return a modified transaction object, or `null` to skip reporting the transaction. This can be used, for instance, for manual PII-stripping before sending.

By the time <PlatformIdentifier name="before-send-transaction" /> is executed, all scope data has already been applied to the event and further modification of the scope won't have any effect.
```

lizokm (approved):
Thanks for adding this!

PR #10876, opened 2024-07-26
deps(changelog): Bump Next.js to canary and Sentry SDK to latest

lforst:
Bumps Next.js to a canary version so that we can test out newer Next.js features that are useful for instrumentation.

All the changes that look unrelated in this PR are actually things that started throwing errors in dev in the new version.


apps/changelog/.env.development
chargome, line 1: shouldn't this be .env.example? 🤔

lforst: I purposefully added this. You probably already know this but `.env.development` will get picked up if you run `next dev` which I hope will save people time and confusion when they try to run the dev server for the changelog for the first time.

apps/changelog/package.json
```diff
@@ -21,15 +21,14 @@
     "@radix-ui/react-icons": "^1.3.0",
     "@radix-ui/react-toolbar": "^1.0.4",
     "@radix-ui/themes": "^2.0.3",
-    "@sentry/nextjs": "^8.8.0",
-    "@spotlightjs/spotlight": "^2.1.1",
```
chargome, line 25: q: Why did you remove spotlight?

chargome (approved):

apps/changelog/package.json
lforst: It was one more thing I needed to get working with the new Next.js version. I will add it back in a follow-up. But good that you bring it up!

PR #10877, opened 2024-07-26
chore(flagpole): Remove Instructions to add flag to `flagpole.flagpole_only_features`

iamrajjoshi:
Because all flags are Flagpole only features and we don't have to add them to `flagpole.flagpole_only_features`, we should delete this from the docs.

### DIFF
--- a/develop-docs/application/feature-flags/flagpole.mdx
+++ b/develop-docs/application/feature-flags/flagpole.mdx
@@ -160,7 +160,7 @@ To see the latest context properties available, check out the repo-specific cont
 
 ## Rolling out a new Flagpole feature
 
-Creating a new Flagpole Feature is currently a 3 step process:
+Creating a new Flagpole Feature is currently a 2 step process:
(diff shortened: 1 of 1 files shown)


develop-docs/application/feature-flags/flagpole.mdx
```diff
@@ -176,8 +176,6 @@ features.add(
 feature.<feature_scope>:<feature_name>
 ```
 
-3. Add the feature name to the `flagpole.flagpole_only_features` list in sentry-options-automator's default [`app.yaml`](https://github.com/getsentry/sentry-options-automator/blob/483737d45dbc68253e926c3f860b5ae33111697b/options/default/app.yaml#L219) file, omitting the `feature.` prefix.
```
GabeVillalobos, line 179: We need to update lines 163 and the flagpole feature testing piece below ~183-215 mentioning testing with the flagpole_only_features stuff, otherwise this looks good!

GabeVillalobos (approved):

PR #10890, opened 2024-07-29
build: Add build command for dev docs on vercel

lforst:
We're doing this so that we can gradually migrate away from calling `rm -r app/changelog` in vercel.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "dev:developer-docs": "yarn enforce-redirects && NEXT_PUBLIC_DEVELOPER_DOCS=1 yarn dev",
     "build:developer-docs": "yarn enforce-redirects && git submodule init && git submodule update && NEXT_PUBLIC_DEVELOPER_DOCS=1 yarn build",
     "build": "yarn enforce-redirects && prisma generate && next build",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -19,6 +19,7 @@
     "dev:developer-docs": "yarn enforce-redirects && NEXT_PUBLIC_DEVELOPER_DOCS=1 yarn dev",
     "build:developer-docs": "yarn enforce-redirects && git submodule init && git submodule update && NEXT_PUBLIC_DEVELOPER_DOCS=1 yarn build",
     "build": "yarn enforce-redirects && prisma generate && next build",
+    "vercel:build:developer-docs": "rm -r app/changelog && yarn enforce-redirects && git submodule init && git submodule update && NEXT_PUBLIC_DEVELOPER_DOCS=1 yarn build",
```
chargome, line 22: we are already running `rm -r app/changelog && yarn build:developer-docs` as a custom vercel build command

lforst: That's the reason I am adding this! 😄 I cannot remove the `rm -r app/changelog` from Vercel to fix CI for [this PR](https://github.com/getsentry/sentry-docs/pull/10889), because I would break all other deploys. If we have these commands in code we can do atomic changes.

chargome: Got it 😂

chargome (approved):

PR #10894, opened 2024-07-29
feat(changelog): Add back spotlight

lforst:
Adds back spotlight to the changelog.

### DIFF
--- a/apps/changelog/next.config.mjs
+++ b/apps/changelog/next.config.mjs
@@ -1,4 +1,5 @@
 import {withSentryConfig} from '@sentry/nextjs';
+import WebpackHookPlugin from 'webpack-hook-plugin';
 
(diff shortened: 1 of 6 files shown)


chargome (approved):

a-hariti (approved):
LGTM

qq: I had done a bundle analysis a while back and noticed that Spotlight ended up in the production bundle

do you think this way of initialization fixes it @lforst ?

lforst:
> qq: I had done a bundle analysis a while back and noticed that Spotlight ended up in the production bundle

I think as long as stuff is guarded with `process.env.NODE_ENV === 'production'` things should tree shake out in prod. Note that I didn't change anything from before here!

a-hariti:
It didn't shake out for some reason, will take a closer look later 👍

Issue #9857, opened 2024-04-29
Inaccessible headings on basic config page


paulGeoghegan:
### Steps to Reproduce

1. open https://docs.sentry.io/platforms/javascript/guides/sveltekit/configuration/options/
2. Turn on NVDA or another screen reader,
3. Try to read the individual configuration options.

### Expected Result

dsn etc

### Actual Result

common options permalink

### Additional Info

Pretty much the <ConfigKey name=""> elements are not accessible the part inside the name attribute should be exposed to screen readers but it is not.

PR #10912, opened 2024-07-30
fix common options permalink aria label

a-hariti:
fixes [#9857](https://github.com/getsentry/sentry-docs/issues/9857)


lizokm (approved):
Woo hoo! Thanks for fixing!

PR #10924, opened 2024-07-31
add only supported in python note

rodolfoBee:


antonpirker (approved):

lizokm (approved):
Thanks for adding this!

PR #10942, opened 2024-07-31
chore(feedback): Update info on self hosted flags and support

aliu39:
Update after verifying https://github.com/getsentry/self-hosted/pull/3193 works. Even after that's merged, this helps people who are on outdated branches

### DIFF
--- a/develop-docs/application/feedback-architecture.mdx
+++ b/develop-docs/application/feedback-architecture.mdx
@@ -4,10 +4,12 @@ title: User Feedback Architecture
 
 **The goal of this doc is to give engineers an in-depth understanding of User Feedback's backend.**
 It will:
+
 1. describe the relevant ingestion pipelines, data models, and functions.
 2. explain the difference between “feedback”, “user reports”, and “crash reports”, and why we built and need to support each.
 
 ## Creation sources
+
 When broken down, there are **5** ways to create feedback in our system 😵‍💫.
 (But 4 of them, related to user reports, are quite similar!) A good reference is the
 `FeedbackCreationSource(Enum)` in [create_feedback.py](https://github.com/getsentry/sentry/blob/2b642e149c79b251e1c2f4339fc73d656347d74e/src/sentry/feedback/usecases/create_feedback.py#L33-L33).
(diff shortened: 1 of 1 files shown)


develop-docs/application/feedback-architecture.mdx
```diff
@@ -186,12 +199,23 @@ filters, **skipping emails if:**
...
+- `organizations:user-feedback-ui`
+- `organizations:user-feedback-replay-clip`
+
+Auto-registered flags for issue platform (you won't be able to grep these in the sentry repo, but you need to enable them too):
+
+- `organizations:feedback-ingest`
+- `organizations:feedback-visible`
+- `organizations:feedback-post-process-group`
+
+---
```
aldy505, line 216: For this, you should also add that by version 24.8.0 (probably), user feedback is available on feature-complete version of sentry. Since we have "errors-only" and "feature-complete", so we won't encounter people mistakenly trying to enable user feedback while they're on errors-only version.

develop-docs/application/feedback-architecture.mdx
```diff
@@ -186,12 +199,25 @@ filters, **skipping emails if:**
...
+- `organizations:user-feedback-ui`
+- `organizations:user-feedback-replay-clip`
+
+Auto-registered flags for issue platform (you won't be able to grep these in the sentry repo, but you need to enable them too):
+
+- `organizations:feedback-ingest`
+- `organizations:feedback-visible`
+- `organizations:feedback-post-process-group`
+
+User feedback will be available on a feature-complete version of Sentry by v24.8.0.
```
c298lee, line 216: The set up docs mention 24.4.2 for full functionality of UF, and 24.5.1 for attachments to be sent in the same envelope as feedback

aldy505: Uhh.. Wouldn't this caused misunderstandings? Since I believe on early days, the required feature flags for enabling User Feedback is quite different than what's currently being set?

Also I don't think `organizations:user-feedback-replay-clip` is available on 24.4.2.

develop-docs/application/feedback-architecture.mdx
```diff
@@ -199,15 +199,27 @@ filters, **skipping emails if:**
...
+Auto-registered flags for issue platform (you won't be able to grep these in the sentry repo, but you need to enable them too):
+
+- `organizations:feedback-ingest`
 - `organizations:feedback-visible`
-- `organizations:user-feedback-ingest`
+- `organizations:feedback-post-process-group`
+
+You need to have v24.4.2 or higher in order to use the full functionality of UF
+(ex: screenshots). Lower versions may have limited functionality. Also note
+feedback is only available in feature-complete Sentry, _not_ errors-only.
```
c298lee, line 218:
```suggestion
You need to have v24.4.2 or higher in order to use the full functionality of User Feedback
such as feedback with screenshots. Lower versions may have limited functionality. Also note
feedback is only available in feature-complete Sentry, _not_ errors-only.
```

develop-docs/application/feedback-architecture.mdx
```diff
@@ -199,15 +199,27 @@ filters, **skipping emails if:**
...
+
+---
 
-To disable auto spam filters, set these to false:
+Auto/AI spam filters are currently unsupported. Make sure `organizations:user-feedback-spam-filter-ingest` is false.
```
c298lee, line 222: I don't think we want to mention AI with our spam filtering
```suggestion
Auto spam filters are currently unsupported. Make sure `organizations:user-feedback-spam-filter-ingest` is false.
```

c298lee (approved):

develop-docs/application/feedback-architecture.mdx
aliu39: I think it's fine @aldy505 , if it helps I'll add a note in the feature flag section that the flags apply for 24.7.1+.

Issue #10592, opened 2024-07-05
Search delivers results in unexpected order


stephanie-anderson:
This was reported via [Slack Message](https://sentry.slack.com/archives/C8H0BQRDJ/p1720165203423089?thread_ts=1720165203.423089&cid=C8H0BQRDJ):

> We should _really_ fix the search in the docs... :sweat_smile: e.g. being in the node docs and searching for `startNewTrace` it only shows the result in the electron docs? that makes no sense at all, the same content exists in the node docs too :see_no_evil:

Let's make sure search results are context sensitive, e.g.
* when in user docs deliver results from user docs (don't return any from develop)
* when in develop docs deliver results from develop docs (don't return any from user-facing docs)
* when on a guide site, return results, that match this guide
* when on a platform site, return results, that match this platform, but not their guides, or other platforms or guides of those

On that note, we should also restrict access to modifying the indices

mydea:
More examples, if needed - e.g. search like this:

![Image](https://github.com/user-attachments/assets/54dd5d2e-1046-46dd-bbb3-54dd86888a14)

Gives me not a single result that I would expect/want based on the place I am searching from.

chargome:
note: modifying the indices should not be an issue, they will be overwritten with each new algolia workflow run

chargome:
Closing this as we now have context sensitive search activated. Let's create separate issues for any other search-related improvements.

PR #10948, opened 2024-08-01
feat: Improve algolia search results

chargome:
- Applies custom search settings from `sentry-global-search`
- Tags algolia records with their respective sdk and framework names for improving path bias (when searching from react guide, results will prioritise records within the guide)
- Updates search settings for algolia client to include framework name for improving search results

relates to [#10592](https://github.com/getsentry/sentry-docs/issues/10592)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -95,3 +95,7 @@ public/page-data
 tsconfig.tsbuildinfo
 
 public/mdx-images/*
+
+# yalc
+.yalc
+yalc.lock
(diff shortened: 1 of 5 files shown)


a-hariti:
Does it address this example from the mentioned issue by @mydea ?

Opentelementry search while on node platform

a-hariti (commented):
I like this idea!

chargome:
@a-hariti yes it does improve the search for that case. Generally speaking, the path bias makes search much more usable this way, but there is probably still room for improvement. Let's try to gather cases and improve incrementally I would say

a-hariti:
is this preview using test indices with tagged records?

chargome:
No the preview is still running on the old index. You can only test that out with a local version of sentry-global-search

a-hariti:
I understand, so you can only test the ranking locally?

chargome:
Yes, or through the dashboard on algolia

lforst (approved):

PR #10953, opened 2024-08-01
chore(billing): Revive legacy docs on managing performance units quota

dashed:
For customers still on pre-AM3 plans, we still link to legacy documentation from the Sentry SaaS web app, and we want these pages to not bork leaving these customers hanging.

Here, I am reviving the legacy docs on managing performance units quota. But this page will be hidden and not linked from anywhere else in the docs website.

Associated pull request to link to this page is in https://github.com/getsentry/sentry/pull/75477


isabellaenriquez (approved):

isabellaenriquez:
I think the offending url needs to be /platform-redirect/?next=/tracing/instrumentation/custom-instrumentation/

PR #10963, opened 2024-08-02
docs(development/testing) Docs about temporary test models

vgrozdanic:
Added a section about temporary models which can be used when developer needs to test a behaviour of a feature that is not tied to a specific model (e.g. custom lookup, or custom field).

I have struggled with this, and I wish I had a docs where I could have looked this up, when I needed it.

### DIFF
--- a/develop-docs/development/testing.mdx
+++ b/develop-docs/development/testing.mdx
@@ -139,7 +139,7 @@ developers yelling at you later for slowing down CI.
 
 If you have a use case where you need to test some behavior of a model (e.g. custom field, or custom lookup), those kind of test shouldn't be tied to the existent models since they can change at any time.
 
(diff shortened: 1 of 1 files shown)


develop-docs/development/testing.mdx
```diff
@@ -135,6 +135,23 @@ If the test runner detects that you used the Django `TestCase` class but you did
 end up needing it, it will yell at you. This protects you against other
 developers yelling at you later for slowing down CI.
 
+### Database tests on temporary models
```
shellmayr, line 138: We could add a link to an example test file here to use as an example, e.g. https://github.com/getsentry/sentry/blob/774af7dc287a1f9199ce855953c9e9184d8bb9c5/tests/sentry/db/models/fields/test_jsonfield.py

shellmayr (approved):

develop-docs/development/testing.mdx
vgrozdanic: Good suggestion, I have added the link to the example

Issue #10510, opened 2024-06-26
Diff code blocks have different padding from normal code blocks


szokeasaurusrex:
### Steps to Reproduce

1. Create a normal code block on a docs page, for example:
<pre>
```python
def hello_world():
    print("hello_world")

hello_world()
print("done")
```
</pre>

2. Create a diff code block on a docs page, for example:
<pre>
```python diff
def hello_world():
-     print("hello_world")
+     print("Hello, world!")

hello_world()
- print("done")
+ print("Done!")
 ```
</pre>

### Expected Result

Both code blocks have the same padding

### Actual Result

The blocks have different padding. [This example](https://sentry-docs-git-szokeasaurusrex-code-block-padding-demo.sentry.dev/product/code-block-example/), which is the rendered preview of #10508, demonstrates the problem.

**Comment by a-hariti** [comment]
this code block:

```python diff
 def hello_world():
+    print("hello_world")
-    print("Hello, world!")
```
produces:
<img width="735" alt="Screenshot 2024-06-28 at 12 30 37" src="https://github.com/getsentry/sentry-docs/assets/17055517/db28488d-11da-416c-9a43-630eaae6b195">
which is inline with a non diff code block

normal code blocks
<img width="732" alt="Screenshot 2024-06-28 at 12 33 02" src="https://github.com/getsentry/sentry-docs/assets/17055517/2effb240-0906-4902-b1ed-6db7f2842e52">

PR #10996, opened 2024-08-06
fix: Add padding to diff code blocks

chargome:
fixes [#10510](https://github.com/getsentry/sentry-docs/issues/10510)

### DIFF
--- a/src/components/codeBlock/code-blocks.module.scss
+++ b/src/components/codeBlock/code-blocks.module.scss
@@ -22,8 +22,11 @@
   }
 
   pre[class^='language-']:has(:global(.code-diff)) {
-    padding-left: 0;
     padding-right: 0;
+
+    & :global(.token.operator) {
+      padding-right: 12px;
+    }
   }
 
   pre[class^='language-']:has(:global(.highlight-line)) {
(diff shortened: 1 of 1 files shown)


src/components/codeBlock/code-blocks.module.scss
```diff
@@ -73,12 +76,16 @@
   :global(.diff-inserted),
   :global(.inserted) {
     background-color: rgba(16, 185, 129, 0.2);
+    margin-left: -12px;
+    width: calc(100% + 12px);
   }
 
   /* Set deleted line (-) color */
```
szokeasaurusrex, line 83: Maybe update this comment (and the comment for the above changes) to reflect that we are also setting the padding here

szokeasaurusrex (commented):
I will take a look at the preview once it is ready. Since I am not super experienced with CSS, it would be good if someone with more experience could take a look at the code changes

chargome:
Alignment works the same way, but the padding is now added automatically. So if a diff block currently contains manual indentations, the padding will be added on top. There are ~100 diff blocks in the project, I can check if they need to be updated.

lforst (approved):

ezio-melotti:
The selector is a bit too generic and adds padding after other operators in the code examples.  I'm trying to fix that in:
* #11052

PR #11011, opened 2024-08-06
Clean up platforms page

lizokm:
Clean up language and delete SDKs that are no longer being maintained.


chargome (approved):
Nice update, thanks!

PR #11019, opened 2024-08-07
ref: Refactor changelog list page

lforst:
"cleans up" the changelog list page. This is all a goal to remove hydration errors.


apps/changelog/src/client/components/date.tsx
```diff
@@ -1,11 +1,10 @@
 const formatDate = (date: string | Date) => {
-  const options: Intl.DateTimeFormatOptions = {
+  const now = new Date(date).toLocaleDateString('en-EN', {
     year: 'numeric',
     month: 'long',
     day: 'numeric',
     timeZone: 'UTC'
-  };
-  const now = new Date(date).toLocaleDateString('en-EN', options);
+  });
```
chargome, line 7: l: we could reuse this formatting, or maybe just use day.js?

chargome (approved):

apps/changelog/src/client/components/date.tsx
lforst: I think this is slimmer for now!

PR #11020, opened 2024-08-07
docs(replay): Update docs on public replay api methods

chargome:
relates to [#12810](https://github.com/getsentry/sentry-javascript/issues/12810)

### DIFF
--- a/docs/platforms/javascript/common/session-replay/understanding-sessions.mdx
+++ b/docs/platforms/javascript/common/session-replay/understanding-sessions.mdx
@@ -85,7 +85,7 @@ replay.start();
 replay.startBuffering();
 ```
 
-This can be used either if both `replaysSessionSampleRate` and `replaysOnErrorSampleRate` are `0` and thus no session has been started automatically, or if you previously stopped a session and want to start a new one (see below). If a session is currently running, `start()` and `startBuffering()` are safe and have no effect. They will [log a debug message](/platforms/javascript/configuration/options/#debug) in this case.
+If both `replaysSessionSampleRate` and `replaysOnErrorSampleRate` are `0`, then you'll need to manually start a session replay, as shown above. This is also useful if you previously stopped a session and want to start a new one (see below). If a session is currently running, `start()` and `startBuffering()` are safe and have no effect. They will [log a debug message](/platforms/javascript/configuration/options/#debug) in this case.
 
 ## Manually Stopping Replay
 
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/common/session-replay/understanding-sessions.mdx
```diff
@@ -107,7 +107,7 @@ await replay.flush();
 
 In `session` mode, this will upload any pending recording data to Sentry. In `buffer` mode, this will upload any pending recording data to Sentry and then continue recording, the same as when an error is sampled with `replaysOnErrorSampleRate`.
 
-Note that it's safe to call `flush()` at any time, even if Session Replay is stopped, in which case, it will do nothing.
+Calling `flush()` while Session Replay is stopped will start a new session recording.
```
aliu39, line 110:
```suggestion
It's safe to call `flush()` at any time. Calling `flush()` while Session Replay is stopped will start a new session recording.
```
Just some nits!

aliu39 (approved):

docs/platforms/javascript/common/session-replay/understanding-sessions.mdx
```diff
@@ -85,7 +85,7 @@ replay.start();
 replay.startBuffering();
 ```
 
-This can be used either if both `replaysSessionSampleRate` and `replaysOnErrorSampleRate` are `0` and thus no session has been started automatically, or if you previously stopped a session and want to start a new one (see below). `start()` and `startBuffering()` will throw an error if a session is currently running.
+This can be used either if both `replaysSessionSampleRate` and `replaysOnErrorSampleRate` are `0` and thus no session has been started automatically, or if you previously stopped a session and want to start a new one (see below). If a session is currently running, `start()` and `startBuffering()` are safe and have no effect. They will [log a debug message](/platforms/javascript/configuration/options/#debug) in this case.
```
vivianyentran, line 88:
```suggestion
If both `replaysSessionSampleRate` and `replaysOnErrorSampleRate` are `0`, then you'll need to manually start a session replay, as shown above. This is also useful if you previously stopped a session and want to start a new one (see below). If a session is currently running, `start()` and `startBuffering()` are safe and have no effect. They will [log a debug message](/platforms/javascript/configuration/options/#debug) in this case.
```

vivianyentran (approved):
Thanks for updating! Added a suggestion

mydea (approved):

PR #11040, opened 2024-08-09
fix: Fix too large build output

lforst:
I have the theory that the .git folder is sometimes included in our lambdas, causing the vercel build to fail because the lambdas are too large.


chargome (approved):

lforst:
Idk if this will actually do anything but let's try

Issue #10719, opened 2024-07-15
Syntax highlight doesn't work when open on new tab


bruno-garcia:
### Steps to Reproduce

If I open links on another tab, syntax highlight doesn't work:

![image](https://github.com/user-attachments/assets/7fe6bbb9-1359-4843-9f4a-8fe6ccc9b450)

If I just navigate away (or if I refresh the new tab with broken highlight), it works:

![image](https://github.com/user-attachments/assets/f82ec60b-565b-4081-b52e-8c63d418feb7)

### Expected Result

highlighting

### Actual Result

no highlighting

a-hariti:
I have noticed this happens "sometimes", can you repro consistently @bruno-garcia ?

chargome:
couldn't reproduce this with the latest changes, feel free to re-open!

PR #11046, opened 2024-08-09
fix: Fix code block hydration errors

### LINKED ISSUE #10574: Hydration errors on dev server
### Steps to Reproduce

- Start local dev server according to instructions
- Navigate to any /platforms route
- Refresh

### Expected Result

Should refresh without hydration errors.

### Actual Result

Screenshot 2024-07-03 at 11 37 28

### Additional Info

_No response_

chargome:
- Remove unused code
- Stop using @next/mdx
- Load prismjs dynamically in `ApiExamples` component

fixes #10574 
might fix errors for switching code tabs (#10719)


src/mdxOptions.js
lforst, line 1: Thanks for removing this. I think I missed this file in the changelog migration.

src/components/apiExamples/apiExamples.tsx
```diff
@@ -25,29 +29,39 @@ function Example({api, selectedTabView, selectedResponse}: ExampleProps) {
...
+          className={jsonCodeBlockStyles}
           dangerouslySetInnerHTML={{
-            __html: Prism.highlight(
-              JSON.stringify(api.responses[selectedResponse].content?.schema, null, 2),
-              Prism.languages.json,
-              'json'
+            __html: JSON.stringify(
+              api.responses[selectedResponse].content?.schema,
+              null,
+              2
```
lforst, line 64: Is there any way uncontrolled input can land here?

src/components/apiExamples/apiExamples.tsx
```diff
@@ -25,29 +29,39 @@ function Example({api, selectedTabView, selectedResponse}: ExampleProps) {
     exampleJson = api.responses[selectedResponse].content?.example;
   }
 
+  // load prism dynamically for these codeblocks,
+  // otherwise the highlighting applies globally
+  useEffect(() => {
```
lforst, line 34: Will this cause layout shift or flashing because it paints some time after the initial page load? 🤔

next.config.js
```diff
@@ -33,15 +31,7 @@ const nextConfig = {
   redirects,
 };
 
-const withMDX = createMDX({
```
lforst, line 36: I was always wondering why we had this... thanks for removing.

app/layout.tsx
```diff
@@ -28,7 +28,7 @@ export const metadata: Metadata = {
 
 export default function RootLayout({children}: {children: React.ReactNode}) {
   return (
-    <html lang="en" suppressHydrationWarning>
```
lforst, line 31: 😎

lforst (approved):
Generally, I wouldn't mind if we merged in this state. However, can we leave a few extra comments explaining wtf is going on with the lazy loading stuff?

src/components/apiExamples/apiExamples.tsx
chargome: good point, but no, this is just the filtered response of our openAPI call

src/components/apiExamples/apiExamples.tsx
chargome: Yes, but it did that before as well. We could also think about not dynamically loading prismjs at all and instead use what we already have (rehypePrismPlus). This would streamline the style of the API JSON examples with our other codeblocks. This would require more work though, we can do it in another pr if needed.

lforst:
Feel free to merge!

Issue #11070, opened 2024-08-13
Hydration errors


chargome:
### Steps to Reproduce

After getting of codeblocks hydration errors in #11046, we're still facing hydration errors in other parts of the platform.

- [x] Errors caused by rendered images (decedents of `<p>`)
- [x] Errors caused by ~~sidebar~~ codetabs (probably localstorage related)

### Expected Result

No hydration errors present on Sentry.

### Actual Result

Hydration errors present on Sentry.

PR #11072, opened 2024-08-13
fix: Get rid of remaining hydration errors

chargome:
fixes [#11070](https://github.com/getsentry/sentry-docs/issues/11070)

### DIFF
--- a/src/components/codeContext.tsx
+++ b/src/components/codeContext.tsx
@@ -296,9 +296,12 @@ const getLocallyStoredSelections = (): SelectedCodeTabs => {
 export function CodeContextProvider({children}: {children: React.ReactNode}) {
   const [codeKeywords, setCodeKeywords] = useState(cachedCodeKeywords ?? DEFAULTS);
   const [isLoading, setIsLoading] = useState<boolean>(cachedCodeKeywords ? false : true);
-  const [storedCodeSelection, setStoredCodeSelection] = useState<SelectedCodeTabs>(() =>
-    getLocallyStoredSelections()
-  );
+  const [storedCodeSelection, setStoredCodeSelection] = useState<SelectedCodeTabs>({});
+
+  // populate state using localstorage
+  useEffect(() => {
+    setStoredCodeSelection(getLocallyStoredSelections());
+  }, []);
 
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


a-hariti (approved):

src/components/codeContext.tsx
```diff
@@ -277,10 +284,21 @@ export async function createOrgAuthToken({
...
+  }
+  return {};
+};
+
 export function CodeContextProvider({children}: {children: React.ReactNode}) {
   const [codeKeywords, setCodeKeywords] = useState(cachedCodeKeywords ?? DEFAULTS);
-
   const [isLoading, setIsLoading] = useState<boolean>(cachedCodeKeywords ? false : true);
+  const [storedCodeSelection, setStoredCodeSelection] = useState<SelectedCodeTabs>(() =>
+    getLocallyStoredSelections()
```
lforst, line 300: l/m: As discussed: I would not make the initial state dependent on localStorage as it invites hydration errors in the future.

lforst (approved):
nice!

src/components/codeContext.tsx
chargome: updated it 👍

src/components/codeContext.tsx
```diff
@@ -277,10 +284,24 @@ export async function createOrgAuthToken({
...
 export function CodeContextProvider({children}: {children: React.ReactNode}) {
   const [codeKeywords, setCodeKeywords] = useState(cachedCodeKeywords ?? DEFAULTS);
-
   const [isLoading, setIsLoading] = useState<boolean>(cachedCodeKeywords ? false : true);
+  const [storedCodeSelection, setStoredCodeSelection] = useState<SelectedCodeTabs>({});
+
+  // populate state using localstorage
+  useEffect(() => {
+    setStoredCodeSelection(getLocallyStoredSelections());
+  }, []);
```
lforst, line 304: with this I think we can get rid of the isLocalStorageAvailable check right?

chargome: I would just leave it tbh

lforst: sounds good

PR #11075, opened 2024-08-13
fix split code blocks for code blocks without specified language

### LINKED ISSUE #11066: Code-blocks are detached from tabs when no language is specified
### Steps to Reproduce

Create a code block with no language and tabtitles:

```
``` {tabTitle: npm}
NODE_OPTIONS='--import=./src/instrument.server.mjs npm start
...
```

### Expected Result

The code-block should not be detached from the tabs.

Image

### Actual Result

The `<pre>` code-block has top-margin when no language is set.

Image

### Additional Info

_No response_

a-hariti:
fixes #11066

### DIFF
--- a/src/components/codeBlock/code-blocks.module.scss
+++ b/src/components/codeBlock/code-blocks.module.scss
@@ -7,6 +7,8 @@
     background: #251f3d;
     border: 1px solid #40364a;
     border-radius: 0;
+    margin-top: 0;
+    margin-bottom: 0;
   }
 
   code[class*='language-'],


src/components/codeBlock/code-blocks.module.scss
```diff
@@ -7,6 +7,8 @@
     background: #251f3d;
     border: 1px solid #40364a;
     border-radius: 0;
+    margin-top: 0;
+    margin-bottom: 0;
```
andreiborza, line 11: q: Do we have code-blocks that don't have a header? Are they still okay?

I think a less-intrusive solution is to specifically target code-blocks that have a header.

andreiborza (approved):

PR #11091, opened 2024-08-14
chore(ruby): Update common options to match other platforms

sl0thentr0py:
This makes it look better and now finally has a sidebar.

### DIFF
--- a/docs/platforms/ruby/common/configuration/options.mdx
+++ b/docs/platforms/ruby/common/configuration/options.mdx
@@ -159,9 +159,9 @@ Sentry automatically sets the current environment from the environment variables
 
 <ConfigKey name="enabled-environments">
 
-By default, events will be sent to Sentry in all environments. If you do not wish to send events in an environment, we suggest you unset the `SENTRY_DSN` variable in that environment.
+By default, events will be sent to Sentry in all environments. If you don't want to send events in a specific environment, you can unset the `SENTRY_DSN` variable in that environment.
 
-Alternately, you can configure Sentry to run only in certain environments by configuring the `enabled_environments` list. For example, to only run Sentry in production:
+You can also set up Sentry to only run in certain environments by configuring the `enabled_environments` list. For example, to only run Sentry in production:
 
 ```ruby
 config.enabled_environments = %w[production]
--- a/docs/platforms/ruby/common/configuration/options.mdx
+++ b/docs/platforms/ruby/common/configuration/options.mdx
@@ -110,7 +110,7 @@ config.send_modules = false # if you don't want to send all the dependency info
 
 <ConfigKey name="include_local_variables">
 
-Whether to capture local variables from the raised exception's frame. Default is `false`. This was called `capture_exception_frame_locals` in older versions.
+Whether to capture local variables from the raised exceptions frame. The default is `false`. (In older versions, this was called `capture_exception_frame_locals`.)
 
 </ConfigKey>
 
--- a/docs/platforms/ruby/common/configuration/options.mdx
+++ b/docs/platforms/ruby/common/configuration/options.mdx
@@ -129,8 +129,8 @@ config.max_breadcrumbs = 30
 Sentry supports different breadcrumbs loggers in the Ruby SDK:
 
 - `:sentry_logger` - A general breadcrumbs logger for all Ruby applications.
-- `:http_logger` - It captures requests made with the standard `net/http` library.
-- `:redis_logger` - It captures breadcrumbs from redis operations.
+- `:http_logger` - Captures requests made with the standard `net/http` library.
+- `:redis_logger` - Captures breadcrumbs from redis operations.
 - `:active_support_logger` - Built on top of [ActiveSupport instrumentation](https://guides.rubyonrails.org/active_support_instrumentation.html) and provides many Rails-specific information.
 - `:monotonic_active_support_logger` - Similar to `:active_support_logger` but breadcrumbs will have monotonic time values. Only available with Rails 6.1+.
 
--- a/docs/platforms/ruby/common/configuration/options.mdx
+++ b/docs/platforms/ruby/common/configuration/options.mdx
@@ -23,8 +23,8 @@ Learn more about [DSN utilization](/product/sentry-basics/dsn-explainer/#dsn-uti
 
 <ConfigKey name="debug">
 
-Turns debug mode on or off. When enabled, SDK errors will be logged with backtrace.
-Please use `config.logger.level` for more output; this is merely for attaching backtraces to the messages.
+This turns debug mode on or off. When enabled, SDK errors will be logged with backtrace.
+If you want more output, use `config.logger.level`. `debug` only works for attaching backtraces to the messages.
 
 </ConfigKey>
 
--- a/docs/platforms/ruby/common/configuration/options.mdx
+++ b/docs/platforms/ruby/common/configuration/options.mdx
@@ -71,6 +71,10 @@ This lets you attach diagnostic client reports about dropped events to an existi
 
 ```ruby
 config.send_client_reports = false
+
+<Note>
+This information isn't currently visible to users, but we're planning on adding it to the user-facing UI in the near future.
+</Note>
 ```
 
 </ConfigKey>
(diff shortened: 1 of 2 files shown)


docs/platforms/ruby/common/configuration/options.mdx
```diff
@@ -77,193 +141,191 @@ config.breadcrumbs_logger = [:active_support_logger]
...
+<PlatformContent includePath="performance/traces-sample-rate" />
 
-`release`
+</ConfigKey>
 
-Track the version of your application in Sentry.
+<ConfigKey name="traces-sampler">
 
-We guess the release intelligently in the following order of preference:
+A lambda or proc responsible for determining the percentage chance a given transaction will be sent to Sentry. It will automatically be passed information about the transaction and the context in which it's being created, and must return a number between `0` (0% chance of being sent) and `1` (100% chance of being sent). Can also be used for filtering transactions, by returning 0 for those that are unwanted. Either this or <PlatformIdentifier name="traces-sample-rate" /> must be defined to enable tracing.
```
lizokm, line 228:
```suggestion
A lambda or proc that's responsible for determining the chance that a given transaction has of being sent to Sentry (from 0-100%). It will automatically be passed information about the transaction and the context in which it's being created, and must return a number between `0` (0% chance of being sent) and `1` (100% chance of being sent). 

It can also be used for filtering transactions, by returning 0 for those that are of no interest. Either this or <PlatformIdentifier name="traces-sample-rate" /> must be defined to enable tracing.
```

docs/platforms/ruby/common/configuration/options.mdx
```diff
@@ -77,193 +141,191 @@ config.breadcrumbs_logger = [:active_support_logger]
...
+<PlatformContent includePath="performance/traces-sampler-as-sampler" />
 
-<PlatformContent includePath="set-release" />
+</ConfigKey>
 
-`sample_rate`
+<ConfigKey name="enable-backpressure-handling">
 
-The sampling factor to apply to events. A value of 0.00 will deny sending any events, and a value of 1.00 will send 100% of events.
+A boolean (default false) that controls whether a new monitor thread will be spawned to perform health checks on the SDK. If the system is unhealthy, the SDK will keep halving the `traces_sample_rate` set by you in 10 second intervals until recovery. This downsampling helps ensure that the system stays stable and reduces SDK overhead under high load.
```
lizokm, line 236:
```suggestion
A boolean (the default is false) that controls whether a new monitor thread will be spawned to perform health checks on the SDK. If the system is unhealthy, the SDK will keep halving the `traces_sample_rate` set by you in 10 second intervals until recovery. This downsampling helps ensure that the system stays stable and reduces SDK overhead under high load.
```

docs/platforms/ruby/common/configuration/options.mdx
```diff
@@ -77,193 +141,191 @@ config.breadcrumbs_logger = [:active_support_logger]
...
-`send_client_reports`
+<ConfigKey name="trace-propagation-targets">
 
-Attach diagnostic client reports about dropped events to an existing envelope max once every 30s. Default is `true`.
+An optional property that controls which downstream services receive tracing data, in the form of a `sentry-trace` and a `baggage` header attached to any outgoing HTTP requests.
 
-This information will not be visible to users at the moment, but we're planning to add this information to user-facing UI.
+The option may contain an array of strings or regex against which the URLs of outgoing requests are matched.
+If one of the entries in the list matches the URL of an outgoing request, trace headers will be attached to that request.
+String entries do not have to be full matches, meaning the URL of a request is matched when it _contains_ a string provided through the option.
```
lizokm, line 246:
```suggestion
String entries don't have to be full matches, (meaning the URL of a request is matched when it _contains_ a string provided through the option).
```

docs/platforms/ruby/common/configuration/options.mdx
```diff
@@ -77,193 +141,191 @@ config.breadcrumbs_logger = [:active_support_logger]
...
 
-If you **do not** want to send this data, you can opt-out by setting
+By default, trace headers are attached to every outgoing request from the instrumented client.
+
+</ConfigKey>
+
+<ConfigKey name="propagate-traces">
+
+By default, Sentry injects `sentry-trace` and `baggage` headers to outgoing requests made with `Net::HTTP` to connect traces between services. You can disable this behavior with
```
lizokm, line 254:
```suggestion
By default, Sentry injects `sentry-trace` and `baggage` headers to outgoing requests made with `Net::HTTP` to connect traces between services. You can disable this behavior with:
```

docs/platforms/ruby/common/configuration/options.mdx
```diff
@@ -77,193 +141,191 @@ config.breadcrumbs_logger = [:active_support_logger]
...
+</ConfigKey>
 
-When its value is `false` (default), sensitive information like
+## Hooks
 
-- user ip
-- user cookie
-- request body
-- query string in the url
+These options can be used to hook the SDK in various ways to customize the reporting of events.
```
lizokm, line 264:
```suggestion
The below options can be used to hook the SDK in various ways and customize how events are being reported.
```

docs/platforms/ruby/common/configuration/options.mdx
```diff
@@ -77,193 +141,191 @@ config.breadcrumbs_logger = [:active_support_logger]
...
 
-will not be sent to Sentry.
+<ConfigKey name="before-send">
 
-You can re-enable it by setting:
+Provide a lambda or proc that is called with an SDK-specific message or error event object, and can return a modified event object, or `nil` to skip reporting the event. This can be used, for instance, for manual PII stripping before sending.
```
lizokm, line 268:
```suggestion
Provides a lambda or proc that's called with an SDK-specific message or error event object, and can return a modified event object, or `nil` to skip reporting the event. This can be used, for instance, for manual PII stripping before sending.
```

docs/platforms/ruby/common/configuration/options.mdx
```diff
@@ -77,193 +141,191 @@ config.breadcrumbs_logger = [:active_support_logger]
...
-A boolean to decide whether to send modules (dependencies) information to Sentry (default is `true`).
+</ConfigKey>
 
-```ruby
-config.send_modules = false # if you don't want to send all the dependency info
-```
+<ConfigKey name="before-send-transaction">
 
-`skip_rake_integration`
+Provide a lambda or proc that is called with an SDK-specific transaction event object, and can return a modified transaction event object, or `nil` to skip reporting the event. One way this might be used is for manual PII stripping before sending.
```
lizokm, line 278:
```suggestion
Provides a lambda or proc that's called with an SDK-specific transaction event object, and can return a modified transaction event object, or `nil` to skip reporting the event. One way this might be used is for manual PII stripping before sending.
```

docs/platforms/ruby/common/configuration/options.mdx
```diff
@@ -77,193 +141,191 @@ config.breadcrumbs_logger = [:active_support_logger]
...
-Determine whether to ignore exceptions caused by rake integrations. Default is `false`.
+<PlatformContent includePath="configuration/before-send-transaction/" />
+
+</ConfigKey>
 
-`trusted_proxies`
+<ConfigKey name="backtrace-cleanup-callback">
 
-These trusted proxies will be skipped when the SDK computing the user's ip address. `sentry-rails` will automatically inject the value of `Rails.application.config.action_dispatch.trusted_proxie` to this option.
+If you want to clean up exceptions' backtrace before it's sent to Sentry, you can specify a callback with `backtrace_cleanup_callback` to do that. For example:
```
lizokm, line 286:
```suggestion
If you want to clean up the backtrace of an exception before it's sent to Sentry, you can specify a callback with `backtrace_cleanup_callback`, for example:
```

docs/platforms/ruby/common/configuration/options.mdx
```diff
@@ -1,74 +1,138 @@
 ---
 title: Basic Options
 sidebar_order: 1
-description: "Learn more about how to configure the SDK."
+description: "Learn more about how to configure the SDK. These options are set when the SDK is first initialized, passed to the init function as an object."
```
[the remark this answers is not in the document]
sl0thentr0py: this copy is the same across other languages, can you fix that separately for all of them?

docs/platforms/ruby/common/configuration/options.mdx
sl0thentr0py: this isn't correct because the option doesn't do the providing, the user does

PR #11104, opened 2024-08-15
Fix a property name typo from `release` to `dist`

abd-almofleh:
While reading the docs on source maps [Associating dist with Artifact Bundle](https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/typescript/#associating-dist-with-artifact-bundle), I found a typo in one of the examples where the `release` property exist twice but the second time it should be `dist`.

Image

## IS YOUR CHANGE URGENT?  
Not urgent, can wait up to 1 week+


vivianyentran (approved):
Thanks for fixing this! 😄

PR #11112, opened 2024-08-16
fix: add favicon to app dir

chargome:
should add favicon to production deployment


andreiborza (approved):
Nice find!

Let's also remove the icons from`/public` and the code that switches favicons when in production mode, otherwise it's confusing.

Also please record that fact in the commit message.

chargome:
note: will do the cleanup in a follow-up PR, if the changes work as expected on sentry.io/changelog

PR #11113, opened 2024-08-16
fix(changelog): use absolute url favicon

chargome:
- reverts  #11112
- updates favicon to use an absolute url that points to the alternative prod domain (changelog.sentry.dev), otherwise accessing the changelog via sentry.io/changelog leads to the favicon not being found (since it is looking for it on sentry.io)
- adds a csp entry for accessing images on changelog.sentry.dev


apps/changelog/src/app/layout.tsx
```diff
@@ -16,7 +16,9 @@ export const metadata: Metadata = {
   title: 'Home',
   icons: {
     icon:
-      process.env.NODE_ENV === 'production' ? '/favicon.ico' : '/favicon_localhost.png',
+      process.env.NODE_ENV === 'production'
+        ? 'https://changelog.sentry.dev/favicon.ico'
```
andreiborza, line 20: 😬

let's add a comment on why this is needed

andreiborza (approved):

andreiborza (approved):

PR #11118, opened 2024-08-16
Added custom screen size to account for header links

JLuse:
Essentially extended a custom screen size in the tailwind.config.js, then added style to the new header link.


lizokm (approved):
Great fix!

lizokm:
@chargome Here's a fix for the issue Rahul mentioned

PR #11119, opened 2024-08-16
set highlight to plaintext

mdtro:
The highlighting defaults to _Bash_ when not defined. This causes IP addresses to have some odd coloring.

Before:
image

Instead, I'm explicitly setting them to `plaintext` to disable the highlighting.

After:
image


vivianyentran (approved):
Thanks for fixing this!

PR #11132, opened 2024-08-23
feat: Giscus comments based on GH discussions

a-hariti:
enable users to login with GH account and leave comments on docs,

which will show up under the repo discussions and on the docs page ofc.

Setup guide: https://giscus.app: basically install the Giscus app and enable Discussions

Currently I have it setup to my [personal fork](https://github.com/a-hariti/sentry-docs) for the preview

TODO:
- [ ] Point to actual repo
- [ ] Adjust to dark mode when dark mode PR is merged
- [ ] is the pathname -> discussion mapping enough given the dev docs/user docs situation?

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "@emotion/core": "^11.0.0",
     "@emotion/react": "^11.11.0",
     "@emotion/styled": "^11.0.0",
+    "@giscus/react": "^3.0.0",
     "@google-cloud/storage": "^7.7.0",
     "@mdx-js/loader": "^3.0.0",
     "@mdx-js/react": "^3.0.0",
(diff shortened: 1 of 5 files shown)


Jeffreyhung (changes requested):
@a-hariti we have a commitment to all of our public facing websites that we do not use any sort of tracking technology to track users and does not drop cookies other than the one we listed [here](https://sentry.io/cookiebounty/#essential-cookies). 
Can you provide some insights on if Giscus will track users or drop any cookies?

a-hariti:
Does that include iframes @Jeffreyhung ?

Jeffreyhung:
@a-hariti iframe is a common way that drops 3rd party cookies, but it doesn't necessary do so

a-hariti:
It does drop these cookies, the question is does the no cookie policy cover iframes or not?

cookies from giscus.app

a-hariti:
Cancelling this due to cookie policy

PR #11133, opened 2024-08-23
docs: fix integration usage to prevent error

charpeni:
Since the import above uses `import * as Sentry...`, let's ensure `browserTracingIntegration` is used from `Sentry. browserTracingIntegration`.


Lms24 (approved):
Thanks!

PR #11135, opened 2024-08-23
self-hosted: minimum system requirements

joshuarli:
current stuff is outdated, felt it would be best to just link to our readme since that's more updated in tandem with `install/_min-requirements.sh    `

### DIFF
--- a/develop-docs/self-hosted/index.mdx
+++ b/develop-docs/self-hosted/index.mdx
@@ -23,20 +23,9 @@ git checkout ${VERSION}
 sudo ./install.sh
 ```
 
-## Recommended system resource
+## Required Minimum System Resources
 
-Depending on your usage, required system resource to run Sentry may varies. Minimum system specification for running Sentry looks like this:
-
-- 2 CPU cores
-- 4 GB RAM
-
-Having a machine that have lower system specification than that will fail the installation script.
-
-Current recommended system specification for running Sentry is:
-
-- 4 CPU cores
-- 16 GB RAM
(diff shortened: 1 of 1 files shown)


aldy505 (changes requested):
Hey don't do this. We shouldnt be referring to the SH's README anymore. What we should do is the opposite: from the README, we refer to the docs site. So there'll be more centralized docs about SH.

vivianyentran (approved):

hubertdeng123:
@aldy505 The docs website is linked in self-hosted README already. To be honest, I would agree and say that self-hosted README should be mostly removed besides basic installation instructions in favor of the docs website given that it's mostly duplicate information. Any thoughts here @BYK ?

BYK:
kill-it-with-fire-kill

aldy505:
> @aldy505 The docs website is linked in self-hosted README already. To be honest, I would agree and say that self-hosted README should be mostly removed besides basic installation instructions in favor of the docs website given that it's mostly duplicate information. Any thoughts here @BYK ?

@hubertdeng123 I prefer an even drastic approach, the readme should only contains "version", a small description about what SH is, and "for everything about this, see the website". 😆

joshuarli:
Sounds good, I'll follow up with moving the self-hosted README over here.

aldy505:
> Sounds good, I'll follow up with moving the self-hosted README over here.

@joshuarli already did here https://github.com/getsentry/sentry-docs/pull/10883

PR #11151, opened 2024-08-27
feat: auto include SignInNote

### LINKED ISSUE #7741: `SignInNote` should be rendered via `CodeBlock`
Instead of having to specify `<SignInNode />` everywhere, we should detect code blocks that are using user-specific variables and render a nice indication that they need to sign in

a-hariti:
closes #7741

an LGTM basically means things look like they used to 😄


chargome (approved):
Nice refactor 👍 

I was just wondering why develop-docs was skipped on this one since we had component changes?

a-hariti:
yeah, not sure why it was skipped :\

PR #11185, opened 2024-08-29
docs(self-hosted) Remove mention of the date of the live workshop

vgrozdanic:
In the self-hosted docs we mention the date of the workshop as if it will happen in the future, but since it has already happened, we should remove the date details.


vivianyentran (approved):
Thanks for fixing this!

PR #11199, opened 2024-09-02
Improve <Expandable />

### LINKED ISSUE #11183: [Expandable Content] Expandable Content Doesn't Work With Anchor Tags/Isn't Stateful
### Steps to Reproduce

1. Navigate directly to an anchor on a page with expandable content: https://docs.sentry.io/platforms/javascript/troubleshooting/#using-the-javascript-proxy-api
2. Observe that the anchor fails to work as expected because the content is now visible.

### Expected Result

The expandable content is shown when the location of the browser includes the relevant anchor tag

### Actual Result

The expandable content is not shown

### Additional Info

IMO this expandable content should be stateful and/or the pageload should hook into the location and toggle the expansion of the relevant section if the anchor is in the URL.

NOTE: Expandable content like this has another UX downside which involves not being able to link to the most visually prominent part of the expandable content (the section title). I think a hack would be to include an anchor immediately inside each expandable section however I also believe that like all other headings on this site it should change the location to include an anchor if you click on the heading.

a-hariti:
- use native `<details />` element
- <kbd>Ctrl</kbd><kbd>F</kbd> expands corresponding matches
- add missing `permalink` flags

Closes #11183

### DIFF
--- a/docs/platforms/javascript/common/troubleshooting/index.mdx
+++ b/docs/platforms/javascript/common/troubleshooting/index.mdx
@@ -6,7 +6,7 @@ notSupported: ["javascript.capacitor"]
 sidebar_order: 9000
 ---
 
-<Expandable title="Updating to a new Sentry SDK version">
+<Expandable permalink title="Updating to a new Sentry SDK version">
 
 If you update your Sentry SDK to a new major version, you might encounter breaking changes that need some adaption on your end.
 Check out our [migration guide](https://github.com/getsentry/sentry-javascript/blob/master/MIGRATION.md) to learn everything you need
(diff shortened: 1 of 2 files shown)


docs/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -266,7 +266,7 @@ curl https://browser.sentry-cdn.com/5.20.1/bundle.min.js -o sentry.browser.5.20.
 
 The last option is to use `Proxy` guard, which will make sure that your code won't break, even when you call our SDK, which was blocked. `Proxy` is supported by all browser except Internet Explorer, though there are no extensions in this browser. Also if `Proxy` is not in any of your user's browser, it will be silently skipped, so you don't have to worry about it breaking anything.
 
-Place this snippet immediately **above** the `<Expandable>` tag containing our CDN bundle. The snippet in a readable format presents like this:
+Place permalink this snippet immediately **above** the `<Expandable>` tag containing our CDN bundle. The snippet in a readable format presents like this:
```
stephanie-anderson, line 269: I don't know if this change makes sense 😅

stephanie-anderson (approved):

Issue #11219, opened 2024-09-04
Page titles use the wrong variable


stephanie-anderson:
### Steps to Reproduce

I notices our platform docs must use the wrong frontmatter information when it comes to generating a pages' title.

### Expected Result

Should be "PHP"
![Image](https://github.com/user-attachments/assets/751aff31-10cc-491a-86c9-f05940fa14f0)

Should be ".NET"
![Image](https://github.com/user-attachments/assets/13d2508a-bb29-497e-ba42-3fddf90fabc8)

Should be "React Native"
![Image](https://github.com/user-attachments/assets/289d0a88-ef32-432a-8e2d-6c86b9669dfb)

a-hariti:
is this related to the change you made to platform capitalisation @chargome ?

chargome:
Shouldn't be, might be, I'll check

chargome:
@stephanie-anderson  Ok this has been this way ever since, we want to get rid of the "Sentry for ..."?

stephanie-anderson:
I don't think that's related - see React Native. I think it's simply using the wrong data from frontmatter (should use the `title`)

![Image](https://github.com/user-attachments/assets/ca283002-9f34-4fdb-8e56-0d1e49a9bbd0)

stephanie-anderson:
@chargome no, this should stay as it is

`<title>.NET | Sentry for Dotnet</title>`

should become

`<title>.NET | Sentry for .NET</title>`

stephanie-anderson:
I had a look at the PR @chargome - I think a sustainable solution would be changing what we pass into this function:
https://github.com/getsentry/sentry-docs/blob/a61c058a5d94860eecd98825d6fddfeedb16f5d3/app/%5B%5B...path%5D%5D/page.tsx#L145

where it shouldn't be the `name` but the `title`.

Otherwise we will have to put all platforms in those conditions, which does not make sense since we have the correct information available anyways :)

chargome:
Oh lol you are right I didn't think about that, I'll refactor it

stephanie-anderson:
Thank you! :D

PR #11223, opened 2024-09-04
refactor: Use frontmatter title for page titles

chargome:
closes [#11219](https://github.com/getsentry/sentry-docs/issues/11219) (again)

### DIFF
--- a/app/[[...path]]/page.tsx
+++ b/app/[[...path]]/page.tsx
@@ -19,7 +19,6 @@ import {isDeveloperDocs} from 'sentry-docs/isDeveloperDocs';
 import {getDevDocsFrontMatter, getDocsFrontMatter, getFileBySlug} from 'sentry-docs/mdx';
 import {mdxComponents} from 'sentry-docs/mdxComponents';
 import {setServerContext} from 'sentry-docs/serverContext';
-import {formatGuideOrPlatformTitle} from 'sentry-docs/utils';
 
 export async function generateStaticParams() {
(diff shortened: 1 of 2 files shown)


app/[[...path]]/page.tsx
```diff
@@ -141,9 +140,7 @@ export async function generateMetadata({params}: MetadataProps): Promise<Metadat
       const guideOrPlatform = getCurrentPlatformOrGuide(rootNode, params.path);
       title =
         pageNode.frontmatter.title +
-        (guideOrPlatform
-          ? ` | Sentry for ${formatGuideOrPlatformTitle(guideOrPlatform.name)}`
-          : '');
+        (guideOrPlatform ? ` | Sentry for ${pageNode.frontmatter.title}` : '');
```
Comment (stephanie-anderson) [inline app/[[...path]]/page.tsx:143]: You need to get that info from `guideOrPlatform`, otherwise it will always take the current page's title

See:
https://sentry-docs-git-cg-refactor-page-titles.sentry.dev/platforms/javascript/install/

Screenshot 2024-09-04 at 15 14 15

Screenshot 2024-09-04 at 15 14 33

stephanie-anderson (changes requested):

app/[[...path]]/page.tsx
Comment (chargome) [reply app/[[...path]]/page.tsx]: 🫠

stephanie-anderson (approved):
Yay!!! Thank you, @chargome !

PR #11226, opened 2024-09-04
fix: Update dsn placeholders

chargome:
fixes all DSN placeholders that have missing underscores


cleptric (approved):

cleptric:
This is almost worth adding a lint for, happens quite often 😬

bruno-garcia:
> what about we make reduce the `_`s to two?
> 
> obviously more than one but no doubt about 3

Having made this mistake myself years ago, I +1 on this

chargome:
merging this for now to get the fix out

Issue #11094, opened 2024-08-14
Optimize vercel lambda sizes


chargome:
### Problem Statement

We're sporadically running into `Error: A Serverless Function has exceeded the unzipped maximum size of 250 MB` when deploying on vercel.

### Solution Brainstorm

Inspect actual vercel output using `vercel build --debug` and see if we can optimise our lambda functions by updating `outputFileTracingExcludes` in our `next.config.js`.

PR #11227, opened 2024-09-04
fix: Optimize lambda output

chargome:
closes [#11094](https://github.com/getsentry/sentry-docs/issues/11094)

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -3,6 +3,34 @@ const {redirects} = require('./redirects.js');
 const {codecovWebpackPlugin} = require('@codecov/webpack-plugin');
 const {withSentryConfig} = require('@sentry/nextjs');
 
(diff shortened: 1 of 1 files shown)


a-hariti (approved):

next.config.js
```diff
@@ -3,6 +3,34 @@ const {redirects} = require('./redirects.js');
...
+        'docs/product/**/*',
+        'docs/concepts/**/*',
+        'docs/api/**/*',
+        'docs/pricing/**/*',
+        'docs/account/**/*',
+        '**/*.gif',
+        'public/mdx-images/**/*',
+        '*.pdf',
+      ],
+      '\\[\\[\\.\\.\\.path\\]\\]': [
```
lforst, line 26: Beautiful

lforst (approved):
Generally looks good, in general though this seems more like a bandaid solution to me and I'd love for us in the future to change our code in a way that all of these things are pulled in in a reasonable way.

PR #11246, opened 2024-09-05
feat: add Definition list support

a-hariti:
Add a remark plugin to handle definition lists
 
adapted from https://github.com/Symbitic/remark-plugins/blob/master/packages/remark-deflist/src/index.ts

Not a perfect implementation as esbuild chokes on html/jsx syntax within definition lists

It works as long as the definition contains plain markdown syntax
 
Line breaks are transformed to `<br />` elements on purpose as a workaround


chargome:
@a-hariti we have removed all remaining definition lists in https://github.com/getsentry/sentry-docs/pull/11238 🤔 ?

chargome (approved):

PR #11250, opened 2024-09-05
replay mobile: ios faq

bruno-garcia:


brustolin (approved):

docs/platforms/apple/guides/ios/session-replay/index.mdx
```diff
@@ -95,3 +95,11 @@ Errors that happen on the page while a replay is running will be linked to the r
...
 - The replay was deleted by a member of your org.
 - There were network errors and the replay wasn't saved.
+
+### FAQ
+
+Q: Doesn't it work with SwiftUI?
+A: Yes. Session Replay works with UIKit as well as SwiftUI.
+
+Q: What's the lowest version of iOS supported?
+A: Session Replay doesn't require a more modern iOS version to work. Recording happens even on the lowest version supported by the Sentry SDK. Today that's iOS 12.
```
lizokm, line 105:
```suggestion
### FAQ

Q: Does Session Replay work with SwiftUI?
A: Yes. It works with both UIKit and SwiftUI.

Q: What's the lowest version of iOS supported?
A: Session Replay recording happens even on the lowest version supported by the Sentry SDK. (Today that's iOS 12.)
```
I'm not sure why, but in the preview, it looks like the line breaks between Q/A aren't showing up. :(

lizokm (approved):
Thanks for adding!

Issue #9284, opened 2024-02-29
Usage of `captureMessage` unclear


johanbook:
### Steps to Reproduce

On page https://docs.sentry.io/platforms/javascript/usage/ the API of `captureMessage` is documented, but however, I cannot find the docs of what the method actually does and where these messages show up in Sentry.

### Expected Result

There to be docs on what method does.

### Actual Result

There is no docs on it (as far as I can find)

lforst:
@lizokm This is fair feedback but it is the same for all SDKs. Can we track this more globally?

lizokm:
@lforst  Yup! I'll add this to my backlog.

PR #11258, opened 2024-09-06
feat: Add docs about what `captureMessage` does

mydea:
This PR updates the SDK usage docs for `captureMessage` to make it clear that those also show up in your issue stream.

Not sure if that is the best wording, but IMHO it should at least give a basic understanding, I guess...?

Closes https://github.com/getsentry/sentry-docs/issues/9284


lforst (approved):

docs/platforms/android/usage/index.mdx
```diff
@@ -29,4 +29,6 @@ While capturing an event, you can also record the breadcrumbs that lead up to th
 
 Another common operation is to capture a bare message. A message is textual information that should be sent to Sentry. Typically, our SDKs don't automatically capture messages, but you can capture them manually.
 
+Messages also show up in your issue stream.
```
vivianyentran, line 32: I think this could be clearer. Where in the issue stream does it show up? Is it accurate to say something like:

 ```suggestion
When captured, messages are used as the issue name in the issue stream. Messages are also displayed under "Message" in the Issue Details page. To search and filter issues by messages, use the [`message`](/concepts/search/searchable-properties/issues/#message) property. 
```

Comment (mydea) [reply docs/platforms/android/usage/index.mdx:32]: Hmm, I don't know, this feels like more details than we have for the capture exception case above? 🤔 

IMHO this:

>  Messages are also displayed under "Message" in the Issue Details page.

sounds a bit redundant to me (we also don't explain that exceptions show up under exception etc, do we?) 🤔 

What about something like this instead, somewhere in the middle:

```suggestion
When captured, messages show up in your issue stream and use the message as the issue name.
```

vivianyentran: What about making it even simpler and just saying this:
```suggestion
When captured, messages show up as the issue name.
```

mydea: I think this misses the key point I wanted to add here (based on the linked issue) that is, what happens with messages at all? 🤔 

What about:

```suggestion
Messages show up as issues on your issue stream, with their message as the issue name.
```

Issue #9524, opened 2024-03-20
Unclear what access token permissions are required for uploading source maps


tjhorner:
### Core or SDK?

Core Sentry product

### Which part? Which one?

Source map uploading

### Description

On [this page](https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/webpack/) it is unclear what permissions a token needs to be given for source maps uploads to work. Ideally I would want to give the token the least permissions possible.

### Suggested Solution

n/a

tjhorner:
Ah I see there is the concept of an org-associated auth token. I now see it's linked with the text "manually create an Auth Token" but I had assumed it would just take me to the user auth token page that I was already at. Maybe being more clear about the difference between an org auth token vs a user auth token would be beneficial.

mydea:
Hey, we'll clarify this more on that page that you should set up an org auth token!

PR #11259, opened 2024-09-06
feat: Clarify that org auth tokens should be used for bundler plugins

mydea:
We do link to org auth tokens further down, but it does not hurt to be explicit there.

Closes https://github.com/getsentry/sentry-docs/issues/9524

### DIFF
--- a/docs/platforms/javascript/common/sourcemaps/uploading/angular-nx.mdx
+++ b/docs/platforms/javascript/common/sourcemaps/uploading/angular-nx.mdx
@@ -54,6 +54,9 @@ If you are on an older version and you want to upload source maps we recommend u
 </Note>
 
 To upload source maps you have to configure an [Organization Auth Token](https://sentry.io/orgredirect/organizations/:orgslug/settings/auth-tokens/).
+
+Alternatively, you can also use a [User Auth Token](https://sentry.io/orgredirect/organizations/:orgslug/settings/account/api/auth-tokens/), with the "Project: Read & Write" and "Release: Admin" permissions.
+
 Auth tokens can be passed to the plugin explicitly with the `authToken` option, with a `SENTRY_AUTH_TOKEN` environment variable, or with an `.env.sentry-build-plugin` file in the working directory when building your project.
 We recommend you add the auth token to your CI/CD environment as an environment variable.
 
--- a/docs/platforms/javascript/common/sourcemaps/uploading/angular-nx.mdx
+++ b/docs/platforms/javascript/common/sourcemaps/uploading/angular-nx.mdx
@@ -53,7 +53,7 @@ If you are on an older version and you want to upload source maps we recommend u
 
 </Note>
 
-To upload source maps you have to configure an auth token.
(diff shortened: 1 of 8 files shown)


docs/platforms/javascript/common/sourcemaps/uploading/angular-nx.mdx
```diff
@@ -53,7 +53,7 @@ If you are on an older version and you want to upload source maps we recommend u
 
 </Note>
 
-To upload source maps you have to configure an auth token.
+To upload source maps you have to configure an [Organization Auth Token](https://sentry.io/orgredirect/organizations/:orgslug/settings/auth-tokens/).
```
Lms24, line 56: what about this? (makes it sound a little bit less strict)
```suggestion
To upload source maps, configure an [Organization Auth Token](https://sentry.io/orgredirect/organizations/:orgslug/settings/auth-tokens/).
```

Lms24 (commented):
Hmm this kinda implies that the user-based tokens don't let you upload source maps. Which last I checked they do, no? No problem for new users but if existing users try to debug their source maps setup and they see this, they might get lead to a wrong path 🤔  Maybe we should adjust the wording a bit (see comment)

docs/platforms/javascript/common/sourcemaps/uploading/angular-nx.mdx
mydea: hmm, not 100% sure, I'll defer this to @getsentry/docs folks for the best wording :D

lforst: I am with @Lms24. You don't *have* to use an org auth token.

vivianyentran: Sorry I'm confused. It sounds like this is a requirement, but also like it's not? Can you clarify what the org auth token is used for? Can users still upload source maps manually without the org auth token?

mydea: So you can either use an org auth token (which is the recommend way and what we generally propose now everywhere), which is easy to set up in that there are no config options.
Or you can also use a personal auth token, which has to be configured correctly.

What if we go with something like this then:

> To upload source maps you have to configure an Organization Auth Token.
> 
> Alternatively, you can also use a Personal Auth Token with the X, Y and Z scopes (TBD: what are these exactly again now?)

vivianyentran: I like that suggestion 👍 

If we want to recommend using org auth tokens as opposed to personal auth tokens, we can also say

> To upload source maps, you should first configure an Organization Auth Token.

lforst: > what are these exactly again now?

- Project read & write
- Release admin

(I think)

mydea: What about `Organization`? 🤔 
> Manage Organizations, resolve IDs, retrieve Repositories and Commits

but I'll adjust it, we can still change it if something is missing!

Lms24 (approved):

PR #11294, opened 2024-09-09
update regex list

rodolfoBee:
updating list of words used in the scrubbing regex: https://github.com/getsentry/relay/blob/master/relay-pii/src/regexes.rs#L286-L290


docs/security-legal-pii/scrubbing/server-side-scrubbing/index.mdx
```diff
@@ -16,19 +16,17 @@ With it enabled, Sentry will scrub the following:
...
-  - secret
-  - passwd
-  - api_key
-  - apikey
-  - access_token
-  - auth
-  - credentials
-  - mysql_pwd
-  - stripetoken
-  - card[number]
```
jjbayer, line 28: Do we not scrub `card[number]` anymore?

PR #11296, opened 2024-09-09
Remove mention of sentinel and prefix actions

armenzg:
These two actions are being removed in https://github.com/getsentry/sentry/pull/76994 since they were never shipped to GA.


vivianyentran (approved):
Thanks for updating!

PR #11300, opened 2024-09-10
feat(headings): Copy link to clipboard on click

ArthurKnaus:
Copy the heading link to the clipboard on click.
We got the feedback from multiple people that they expected this behaviour.

Screenshot 2024-09-10 at 08 29 31

closes https://github.com/getsentry/sentry/issues/77226

### DIFF
--- a/src/components/smartLink.tsx
+++ b/src/components/smartLink.tsx
@@ -1,3 +1,6 @@
+'use client';
+
+import {useCallback} from 'react';
 import Link from 'next/link';
 
 import {ExternalLink} from './externalLink';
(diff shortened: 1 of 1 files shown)


ArthurKnaus:
The current implementation would not have been my first choice, but I did not manage to configure our autolink plugin in a way to achieve this.
maybe @mjq or @a-hariti know a better way to do this?

vgrozdanic (approved):

src/components/smartLink.tsx
```diff
@@ -1,3 +1,6 @@
+'use client';
```
priscilawebdev, line 1: why do we need `'use client';`?

src/components/smartLink.tsx
```diff
@@ -1,3 +1,6 @@
...
+
+import {useCallback} from 'react';
```
vgrozdanic, line 3: I am not very familiar with NextJS, but maybe if we don't use `useCallback` we could still have server side rendering of this component?

src/components/smartLink.tsx
ArthurKnaus: In NextJs components are server components per default and therefore are unable to implement any interactivity like `onClick` callbacks. The `'use client'` directive tells React to "make them interactive" once the html is rendered in the browser.

src/components/smartLink.tsx
ArthurKnaus: No the click handler is the problem.
See https://github.com/getsentry/sentry-docs/pull/11300#discussion_r1753216612

priscilawebdev (approved):
code looks good to me! 🙌

PR #11312, opened 2024-09-10
fix(angular): Ensure sourcemaps links have correct text color

Lms24:
Just noticed that if you make a link bold, the text color changes, making the link no longer appear purple. Might be a one-off though because we rarely make lists like this afaict

EDIT: Looks like any bold link is no longer coloured purple. Should we fix this? I guess there's rarely a reason to make a link bold so maybe it's also just fine if we leave it as-is

### DIFF
--- a/platform-includes/sourcemaps/overview/javascript.angular.mdx
+++ b/platform-includes/sourcemaps/overview/javascript.angular.mdx
@@ -36,18 +36,21 @@ To generate source maps, you need to add the `sourceMap` option to your `angular
 ```
 
 <Alert>
-Generating sourcemaps may expose them to the public, potentially causing your source code to be leaked. You can prevent this by configuring your server to deny access to `.js.map` files, or by deleting the sourcemaps before deploying your application.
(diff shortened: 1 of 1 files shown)


platform-includes/sourcemaps/overview/javascript.angular.mdx
```diff
@@ -36,18 +36,21 @@ To generate source maps, you need to add the `sourceMap` option to your `angular
 ```
 
 <Alert>
-Generating sourcemaps may expose them to the public, potentially causing your source code to be leaked. You can prevent this by configuring your server to deny access to `.js.map` files, or by deleting the sourcemaps before deploying your application.
+  Generating sourcemaps may expose them to the public, potentially causing your
+  source code to be leaked. You can prevent this by configuring your server to
+  deny access to `.js.map` files, or by deleting the sourcemaps before deploying
+  your application.
```
Lms24, line 42: auto-formatting(?)

chargome (approved):
Weird behaviour on the bold links 🤔 I'll open an issue for that, thanks!

PR #11354, opened 2024-09-13
fix(js): Remove duplicated "latest SDK" prerequisite

Lms24:
Removes duplicated and somewhat confusing entry that the "Latest JavaScript SDK" was a prerequesite in the getting started guides (second red square in screenshot). For now we have the alert that people should upgrade to v8 SDK. We probably can remove that too, once we have versioned docs or after a couple of months. 

image


lizokm (approved):
Looks good to me!

PR #11373, opened 2024-09-15
fix: platform-redirect

chargome:
fixes broken platform-redirect


lizokm (approved):
Thanks so much for fixing!

PR #11376, opened 2024-09-16
docs(profiling): add new cont profiling page docs

JonasBa:
## DESCRIBE YOUR PR

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

## LEGAL BOILERPLATE

## EXTRA RESOURCES

### DIFF
--- a/docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
+++ b/docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
@@ -1,7 +1,7 @@
 ---
 title: Transaction vs Continuous Profiling
 sidebar_order: 140
-description: "Continuous and transaction based profiling are two different modes with which you can profile the runtime of your application. This document serves as an explanation as to why we did it and what it enables."
+description: "Learn about the differences between continuous and transaction-based profiling."
 ---
 
 We've released a new profiling mode called **continuous profiling**. Read on to learn about the differences between transaction-based and continuous profiling mode. 
--- a/docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
+++ b/docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
@@ -36,11 +36,11 @@ We recommend that you call the `startProfiling` method right after the Sentry SD
 
 ## Enabling transaction or continuous profiling mode
 
-Unfortunately, it is currently not possible to use both profiling modes at the same time, as they are mutually exlusive. Since the mode you intend on using depend on the SDK initialization arguments, it means that the profiling mode will have to be selected at the time the Sentry SDK is initialized.
+Unfortunately, at this time it's not possible to use both profiling modes at the same time since they're mutually exclusive. Since the mode you intend to use is dependent on the SDK initialization arguments, profiling mode will have to be selected when the Sentry SDK is first initialized.
 
-To enable continuous profiling mode, you should ensure that neither profileSampleRate or profilesSampler are set as the values to the Sentry.Init call. If either of those value are set, the SDK will default to transaction based profiling. When the SDK is configured for continuous profiling, the top level calls to start profiles will enable calls to the profiler. If the SDK is configured for transaction based profiling, then these calls will void and not trigger the profiler.
+To enable continuous profiling, you have to make sure that neither `profileSampleRate` nor `profilesSampler` are set as the values of the `Sentry.Init` call. If either of those values are set, the SDK will default to transaction-based profiling. When the SDK is configured for continuous profiling, the top level calls to start profiles will enable calls to the profiler. If the SDK is configured for transaction-based profiling, these calls will void and not trigger the profiler.
 
-Example of enabling continuous profiling in NodeJS
+Here's an example of how to enable continuous profiling in NodeJS:
 
 ```javascript
 Sentry.Init({
--- a/docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
+++ b/docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
@@ -68,7 +68,7 @@ const transaction = Sentry.startTransaction();
 transaction.finish();
 ```
 
-Note that while the profiling mode cannot be changed at runtime, it is fine for different projects or applications to use different profiling modes, or to switch modes.
+Note, that while the profiling mode can't be changed at runtime, it's fine for different projects or applications to use different profiling modes, or to switch modes.
 
 ## Differences in product experience
 
--- a/docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
+++ b/docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
@@ -34,7 +34,7 @@ Transaction-based profiling was opaque from the SDK side, with the SDK being in 
 
 We recommend that you call the `startProfiling` method right after the Sentry SDK is initialized so that you gain visibility at the earliest point in your application lifecycle. From then on, the profiler will keep collecting profiles and sending chunks to Sentry until `stopProfiling` is called.
 
-## Enabling transaction or continuous profiling mode
+## Choosing Between Transaction and Continuous Profiling Mode
 
 Unfortunately, at this time it's not possible to use both profiling modes at the same time since they're mutually exclusive. Since the mode you intend to use is dependent on the SDK initialization arguments, profiling mode will have to be selected when the Sentry SDK is first initialized.
 
--- a/docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
+++ b/docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
@@ -1,8 +1,46 @@
-
 ---
-title: "Transaction vs Continuous Profiling"
-sidebar_order: 120
-description: "Continuous and transaction based profiling are two different modes with which you can profile the runtime of your application. They approach profiling from a different points of view with different tradeoffs. The documentation here serves as an explanation of those tradeoffs."
+title: Transaction vs Continuous Profiling
+sidebar_order: 140
+description: "Continuous and transaction based profiling are two different modes with which you can profile the runtime of your application. This document serves as an explanation as to why we did it and what it enables."
 ---
 
-Nothing to see here
+<Note>
+We have released a new profiling mode called continuous profiling mode with fundamental differences from the previous transaction based profiling mode. This document serves as an explanation as to why we did so along with a couple of things that are helpful to know.
+</Note>
+
+## A bit of context first
+
+Transaction based profiling was the first profiling mode supported at Sentry, it  simply means that any code executed between Sentry.startTransaction and transaction.finish is subject to being profiled. In this mode, all profiles are attached to transactions (and are even sent as part of the same envelope). This has the benefit of automatically profiling parts of the application that are actually instrumented while requiring no extra effort. This approach however has drawback that continuous profiling aims to address, one of the most obvious ones being that profiles cannot exceed 30 second durations.
+
+In theory, transactions can be infinitely long as their duration does not impact payload size as start and end are represented by two timestamps. Profiles however are different, each stack sample that is collected by the profiler increases the payload size, which is why all Sentry SDKs had enforced a max profile duration of 30s. This limitation helps us safeguard from large payload that could otherwise harm your application performance. This unfortunately comes at the expense of limiting profiling capabilities as it means profiling is now no longer able to profile long running tasks like machine learning pipelines or build workflows. This feedback was the major driving force why we've introduced continuous profiling.
+
+The cap on max profile duration is however not the only drawback of transaction based profiling, another limitation of it is that the profiling data you collect, will only ever be as good as the instrumentation you have. In other words, if there are parts of your application that you did not instrument, the chance of you finding out that they are slowing down your application are near zero.
+
+To address these limitations, we have worked on a different profiling mode, one that does not impose constraints on profile durations and can surface parts of your application that might be slowing down your application even though you may not have instrumented them.
+
+## Continuous Profiling mode
+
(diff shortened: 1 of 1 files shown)


docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
```diff
@@ -0,0 +1,74 @@
+---
+title: Transaction vs Continuous Profiling
+sidebar_order: 140
+description: "Continuous and transaction based profiling are two different modes with which you can profile the runtime of your application. This document serves as an explanation as to why we did it and what it enables."
+---
+
+<Note>
+We have released a new profiling mode called continuous profiling mode with fundamental differences from the previous transaction based profiling mode. This document serves as an explanation as to why we did so along with a couple of things that are helpful to know.
+</Note>
```
lizokm, line 9:
```suggestion
We've released a new profiling mode called **continuous profiling**. Read on to learn about the differences between transaction-based and continuous profiling mode. 

```
I don't think this needs to be a note. It can just be the intro.

docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
```diff
@@ -0,0 +1,74 @@
+---
+title: Transaction vs Continuous Profiling
+sidebar_order: 140
+description: "Continuous and transaction based profiling are two different modes with which you can profile the runtime of your application. This document serves as an explanation as to why we did it and what it enables."
```
lizokm, line 4:
```suggestion
description: "Learn about the differences between continuous and transaction-based profiling."
```

docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
```diff
@@ -0,0 +1,74 @@
...
+---
+
+<Note>
+We have released a new profiling mode called continuous profiling mode with fundamental differences from the previous transaction based profiling mode. This document serves as an explanation as to why we did so along with a couple of things that are helpful to know.
+</Note>
+
+## A bit of historical context first
```
lizokm, line 11:
```suggestion
## Historical Context
```

docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+The cap on max profile duration is however not the only drawback of transaction based profiling, another limitation of it is that the profiling data you collect, will only ever be as good as the instrumentation you have. In other words, if there are parts of your application that you did not instrument, the chance of you finding out that they are slowing down your application are near zero.
+
+To address these limitations, we have worked on a different profiling mode, one that does not impose constraints on profile durations and can surface parts of your application that might be slowing down your application even though you may not have instrumented them.
```
lizokm, line 19:
```suggestion
Transaction-based profiling was the first profiling mode supported by Sentry. It made it so that any code executed between `Sentry.startTransaction` and `transaction.finish` could be profiled. In this mode, all profiles were attached to transactions and sent as part of the same envelope. 

This had the benefit of automatically profiling parts of the application that were actually instrumented while requiring no extra effort. This approach had drawbacks that continuous profiling aims to address - one of the most obvious ones being that profiles couldn't exceed 30-second durations.

In theory, transactions can be infinitely long since their duration doesn't impact payload size because start and end are represented by two timestamps. But profiles are different. Each stack sample that is collected by the profiler increases the payload size, which is why all Sentry SDKs had enforced a max profile duration of 30s. This limitation helps safeguard against large payloads that could harm your application performance. 

This unfortunately, comes at the expense of limiting profiling capabilities, making it so that profiling isn't able to profile long-running tasks such as machine learning pipelines or build workflows. This feedback was the driving force behind continuous profiling.

The cap on max profile duration isn't the only drawback of transaction-based profiling. Another limitation is that the profiling data you collect will only ever be as good as the instrumentation you have. In other words, if there are parts of your application that you didn't instrument, the chance of you finding out that they're slowing down your application are near zero.

To address these limitations, we created a different profiling mode, one that doesn't impose constraints on profile durations and can surface parts of your application that might be slowing down your application even if you haven't instrumented them.
```

docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+## Continuous profiling mode
+
+In countinuous profiling mode, the profiler runs continuously (no pun intended) and regularly flushes what we call profile "profile chunks" to the server, this enables us to extend profile durations and continuously profile your application.
+
+Continuous profiling mode is suitable for profiling long running workflows or processes that you want full visibility into, while the transaction based profiling is largely suitable for workflows where you want to limit profiling to only a subset of your application.
```
lizokm, line 25:
```suggestion
## Continuous Profiling Mode

In continuous profiling mode, the profiler runs continuously (no pun intended) and regularly flushes what we call "profile chunks" to the server. This enables us to extend profile durations and continuously profile your application.

Continuous profiling mode is capable of profiling long-running workflows or processes that you want full visibility into, while transaction-based profiling is intended for workflows where you want to limit profiling to only a subset of your application.
```

docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+
+## SDK differences
+
+Transaction based profiling was opaque from the SDK point of view with the SDK being in full control of when the profiler woulds start and stop based on the transactions it would generate. In continuous profiling mode, this is no longer true. Developers can now control when the profiler is started or stopped via new top level SDK methods. The exact method naming varies, but most SDKs that support continuous profiling now expose a top level Sentry.profiler that exposes a startProfiling and stopProfiling method (please see SDK docs for exact definitions).
+
+It is advisable that you call the startProfiling method right after the Sentry SDK is initialized so that you gain visibility at the earliest point in your application lifecycle. From then on, the profiler will keep collecting profiles and sending chunks to Sentry until stopProfiling is called.
```
lizokm, line 32:
```suggestion
## SDK Differences

Transaction-based profiling was opaque from the SDK side, with the SDK being in full control of when the profiler would start and stop based on the transactions it generated. In continuous profiling mode, this is no longer true. Developers can now control when the profiler is started or stopped via new top-level SDK methods. The exact method-naming varies, but most SDKs that support continuous profiling now expose a top level `Sentry.profiler` that exposes a `startProfiling` and `stopProfiling` method. (Please see the SDK docs for exact definitions.)

We recommend that you call the `startProfiling` method right after the Sentry SDK is initialized so that you gain visibility at the earliest point in your application lifecycle. From then on, the profiler will keep collecting profiles and sending chunks to Sentry until `stopProfiling` is called.
```

docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+## Enabling transaction or continuous profiling mode
```
lizokm, line 34:
```suggestion
## Choosing Between Transaction and Continuous Profiling Mode
```

docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+Unfortunately, it is currently not possible to use both profiling modes at the same time, as they are mutually exlusive. Since the mode you intend on using depend on the SDK initialization arguments, it means that the profiling mode will have to be selected at the time the Sentry SDK is initialized.
+
+To enable continuous profiling mode, you should ensure that neither profileSampleRate or profilesSampler are set as the values to the Sentry.Init call. If either of those value are set, the SDK will default to transaction based profiling. When the SDK is configured for continuous profiling, the top level calls to start profiles will enable calls to the profiler. If the SDK is configured for transaction based profiling, then these calls will void and not trigger the profiler.
+
+Example of enabling continuous profiling in NodeJS
```
lizokm, line 40:
```suggestion
Unfortunately, at this time it's not possible to use both profiling modes at the same time since they're mutually exclusive. Since the mode you intend to use is dependent on the SDK initialization arguments, profiling mode will have to be selected when the Sentry SDK is first initialized.

To enable continuous profiling, you have to make sure that neither `profileSampleRate` nor `profilesSampler` are set as the values of the `Sentry.Init` call. If either of those values are set, the SDK will default to transaction-based profiling. When the SDK is configured for continuous profiling, the top level calls to start profiles will enable calls to the profiler. If the SDK is configured for transaction-based profiling, these calls will void and not trigger the profiler.

Here's an example of how to enable continuous profiling in NodeJS:
```

docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
```diff
@@ -0,0 +1,74 @@
...
+  integrations: [nodeProfilingIntegration()]
+})
+
+Sentry.profiler.startProfiling();
+```
+
+If you wish to keep using transaction based profiling, then the options remain the same and you should either set profilesSampleRate or profilesSampler option on the SDK.
+
+
+Example of enabling transaction based profiling in NodeJS
```
lizokm, line 54:
```suggestion
If you want to keep using transaction-based profiling, then the options are the same. You can set either the `profilesSampleRate` or the `profilesSampler` option on the SDK.


Here's an example of enabling transaction-based profiling in NodeJS:
```

docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
```diff
@@ -0,0 +1,74 @@
...
+  profileSampleRate: 0.1 // profiles 10% of transactions
+  integrations: [nodeProfilingIntegration()]
+})
+
+const transaction = Sentry.startTransaction();
+//code executed between these two calls is subject to profiling
+transaction.finish();
+```
+
+Note that while the profiling mode cannot be changed at runtime, it is fine for different projects or applications to use different profiling modes, or to switch modes.
```
lizokm, line 68:
```suggestion
Note, that while the profiling mode can't be changed at runtime, it's fine for different projects or applications to use different profiling modes, or to switch modes.
```

docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+## Differences in product experience
+
+The major difference in product experience when using continuous profiling is that you will be able to visualize a flamegraph for your entire application, which means you can now take a step back from the previous transaction based view and look at your application's runtime as a whole. This makes it easier to for you to prioritize the functions that are slowing down your entire application and not just one particular transaction.
+
+The product experience otherwise remains largely the same with entrypoints into profiling being supported from various parts of the performance product.
```
lizokm, line 74:
```suggestion

## Differences When Using Sentry

The major difference between continuous profiling and transaction-based profiling when using Sentry, is that with continuous profiling you'll be able to visualize a flamegraph for your entire application. This means, that you'll be able to take a step back from the previous transaction-based view and look at your application's runtime as a whole, which makes it easier to prioritize the functions that are slowing down your entire application and not just one particular transaction.

Otherwise, what you'll see in Sentry will be mostly the same. Entry points into profiling will continue to be supported by various parts of the performance product.
```

lizokm (approved):
Made some edits, but otherwise this looks great. Thanks for adding!

docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
```diff
@@ -0,0 +1,77 @@
...
+
+
+## Historical Context
+
+Transaction-based profiling was the first profiling mode supported by Sentry. It made it so that any code executed between `Sentry.startTransaction` and `transaction.finish` could be profiled. In this mode, all profiles were attached to transactions and sent as part of the same envelope. 
+
+This had the benefit of automatically profiling parts of the application that were actually instrumented while requiring no extra effort. This approach had drawbacks that continuous profiling aims to address - one of the most obvious ones being that profiles couldn't exceed 30-second durations.
+
+In theory, transactions can be infinitely long since their duration doesn't impact payload size because start and end are represented by two timestamps. But profiles are different. Each stack sample that is collected by the profiler increases the payload size, which is why all Sentry SDKs had enforced a max profile duration of 30s. This limitation helps safeguard against large payloads that could harm your application performance.
```
indragiek, line 16: Could be more explicit and say that large profiles would mean a lot of data being buffered in memory, which could result in high memory usage for the application.

docs/product/explore/profiling/transaction-vs-continuous-profiling.mdx
```diff
@@ -0,0 +1,77 @@
...
+transaction.finish();
+```
+
+Note, that while the profiling mode can't be changed at runtime, it's fine for different projects or applications to use different profiling modes, or to switch modes.
+
+## Differences When Using Sentry
+
+The major difference between continuous profiling and transaction-based profiling when using Sentry, is that with continuous profiling you'll be able to visualize a flamegraph for your entire application. This means, that you'll be able to take a step back from the previous transaction-based view and look at your application's runtime as a whole, which makes it easier to prioritize the functions that are slowing down your entire application and not just one particular transaction.
+
+Otherwise, what you'll see in Sentry will be mostly the same. Entry points into profiling will continue to be supported by various parts of the performance product.
```
indragiek, line 77: > Entry points into profiling will continue to be supported by various parts of the performance product.

It's unclear to me what this means

JonasBa: Yeah, lets remove this, folks shouldt have to care about it

JonasBa: Removed this entire line at the end

PR #11378, opened 2024-09-17
feat: Add docs about request isolation in Node SDK

mydea:
This adds docs for how to use `withIsolationScope`.

Additionally, it also adds a new section (for server SDKs) under "enriching events" describing how process isolation works. We lack docs for this as of now, and this is something users may def. fall into.

It would be ideal if we could somehow auto-fork this for middlewares, but in the meanwhile...

See https://github.com/getsentry/sentry-javascript/discussions/12191#discussioncomment-10667098 for how this affects users.

### DIFF
--- a/docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
+++ b/docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
@@ -21,7 +21,7 @@ notSupported:
 
 In server-side environments, the <PlatformLink to='/enriching-events/scopes'>isolation scope</PlatformLink> is automatically forked around request boundaries. This means that each request will have its own isolation scope, and data set on the isolation scope will only apply to events captured during that request. This is done automatically by the SDK.
 
-However, the request isolation happens when the request itself is processed. This means that if you e.g. have a middleware where you want to set Sentry data (e.g. `Sentry.setUser()` in an auth middleware), you have to manually fork the isolation scope with `Sentry.withIsolationScope()` - see [Using withIsolationScope](../scopes/#using-withisolationscope).
+However, the request isolation happens when the request callback itself is being executed. This means that if you e.g. have a middleware where you want to set Sentry data (e.g. `Sentry.setUser()` in an auth middleware), you have to manually fork the isolation scope with `Sentry.withIsolationScope()` - see [Using withIsolationScope](../scopes/#using-withisolationscope).
 
 This is also necessary if you want to isolate a non-request process, e.g. a background job.
 
--- a/docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
+++ b/docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
@@ -21,27 +21,16 @@ notSupported:
 
 In server-side environments, the <PlatformLink to='/enriching-events/scopes'>isolation scope</PlatformLink> is automatically forked around request boundaries. This means that each request will have its own isolation scope, and data set on the isolation scope will only apply to events captured during that request. This is done automatically by the SDK.
 
-However, the request isolation happens when the request callback itself is being executed. This means that if you e.g. have a middleware where you want to set Sentry data (e.g. `Sentry.setUser()` in an auth middleware), you have to manually fork the isolation scope with `Sentry.withIsolationScope()` - see [Using withIsolationScope](../scopes/#using-withisolationscope).
+However, tehre are also other cases where you may want to have isolation, for example in background jobs or when you want to isolate a specific part of your code. In these cases, you can use `Sentry.withIsolationScope()` to create a new isolation scope that is valid inside of the callback you pass to it - see [Using withIsolationScope](../scopes/#using-withisolationscope).
 
-This is also necessary if you want to isolate a non-request process, e.g. a background job.
-
-The following example shows how you can use `withIsolationScope` to attach a user and a tag in an auth middleware:
+The following example shows how you can use `withIsolationScope` to attach data for a specific job run:
 
 ```javascript
-const auth = (req, res, next) => {
-  Sentry.withIsolationScope(() => {
-    const authUser = findUserForHeader(req.headers["authorization"]);
-    if (!authUser) {
-      Sentry.setTag("Authenticated", false);
-      Sentry.setUser(null);
-      next(new Error("Authentication Error"));
-    } else {
-      Sentry.setTag("Authenticated", true);
-      Sentry.setUser(authUser);
-      next();
-    }
+async function job(jobId) {
+  return Sentry.withIsolationScope(async () => {
+    // Only valid for events in this callback
+    Sentry.setTag("jobId", jobId);
+    await doSomething();
   });
-};
+}
 ```
-
-This way, the user & tag will only be attached to events captured during the request that passed the auth middleware.
--- a/docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
+++ b/docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
@@ -21,7 +21,7 @@ notSupported:
 
 In server-side environments, the <PlatformLink to='/enriching-events/scopes'>isolation scope</PlatformLink> is automatically forked around request boundaries. This means that each request will have its own isolation scope, and data set on the isolation scope will only apply to events captured during that request. This is done automatically by the SDK.
 
-However, tehre are also other cases where you may want to have isolation, for example in background jobs or when you want to isolate a specific part of your code. In these cases, you can use `Sentry.withIsolationScope()` to create a new isolation scope that is valid inside of the callback you pass to it - see [Using withIsolationScope](../scopes/#using-withisolationscope).
+However, there are also other cases where you may want to have isolation, for example in background jobs or when you want to isolate a specific part of your code. In these cases, you can use `Sentry.withIsolationScope()` to create a new isolation scope that is valid inside of the callback you pass to it - see [Using withIsolationScope](../scopes/#using-withisolationscope).
 
 The following example shows how you can use `withIsolationScope` to attach data for a specific job run:
 
--- a/docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
+++ b/docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
@@ -1,0 +1,47 @@
+---
+title: Process Isolation
+description: "Learn more about how process isolation (or request isolation) works in the Sentry SDK."
+supported:
+  - javascript.nextjs
+  - javascript.node
+  - javascript.connect
+  - javascript.express
+  - javascript.fastify
+  - javascript.hapi
+  - javascript.koa
+  - javascript.nestjs
+  - javascript.nuxt
+  - javascript.solidstart
+  - javascript.sveltekit
+  - javascript.astro
+  - javascript.remix
+notSupported:
+  - javascript
+---
+
+In server-side environments, the <PlatformLink to='/enriching-events/scopes'>isolation scope</PlatformLink> is automatically forked around request boundaries. This means that each request will have its own isolation scope, and data set on the isolation scope will only apply to events captured during that request. This is done automatically by the SDK.
+
+However, the request isolation happens when the request itself is processed. This means that if you e.g. have a middleware where you want to set Sentry data (e.g. `Sentry.setUser()` in an auth middleware), you have to manually fork the isolation scope with `Sentry.withIsolationScope()` - see [Using withIsolationScope](../scopes/#using-withisolationscope).
+
+This is also necessary if you want to isolate a non-request process, e.g. a background job.
+
+The following example shows how you can use `withIsolationScope` to attach a user and a tag in an auth middleware:
+
+```javascript
+const auth = (req, res, next) => {
+  Sentry.withIsolationScope(() => {
+    const authUser = findUserForHeader(req.headers["authorization"]);
+    if (!authUser) {
+      Sentry.setTag("Authenticated", false);
+      Sentry.setUser(null);
+      next(new Error("Authentication Error"));
+    } else {
+      Sentry.setTag("Authenticated", true);
+      Sentry.setUser(authUser);
+      next();
+    }
+  });
+};
+```
+
+This way, the user & tag will only be attached to events captured during the request that passed the auth middleware.
--- a/docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
+++ b/docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
@@ -19,11 +19,11 @@ notSupported:
   - javascript
 ---
 
-In server-side environments, the <PlatformLink to='/enriching-events/scopes'>isolation scope</PlatformLink> is automatically forked around request boundaries. This means that each request will have its own isolation scope, and data set on the isolation scope will only apply to events captured during that request. This is done automatically by the SDK.
+In server-side environments, the <PlatformLink to='/enriching-events/scopes'>isolation scope</PlatformLink> automatically forks around request boundaries. This is done automatically by the SDK. As a result, each request has its own isolation scope, and data set on the isolation scope only applies to events captured during that request. 
 
-However, there are also other cases where you may want to have isolation, for example in background jobs or when you want to isolate a specific part of your code. In these cases, you can use `Sentry.withIsolationScope()` to create a new isolation scope that is valid inside of the callback you pass to it - see [Using withIsolationScope](../scopes/#using-withisolationscope).
+However, there are also other times when you may want to have isolation, for example, in background jobs or when you want to isolate a specific part of your code. In these cases, you can use `Sentry.withIsolationScope()` to create a new isolation scope that's valid inside of the callback you pass to it. Learn more about using [withIsolationScope](../scopes/#using-withisolationscope).
 
-The following example shows how you can use `withIsolationScope` to attach data for a specific job run:
+The following example shows how you can use `withIsolationScope` to attach data to a specific job run:
 
 ```javascript
 async function job(jobId) {
--- a/docs/platforms/javascript/common/enriching-events/scopes/index.mdx
+++ b/docs/platforms/javascript/common/enriching-events/scopes/index.mdx
@@ -143,3 +143,11 @@ In the following example we use <PlatformIdentifier name="with-scope" /> to atta
 The scope inside the `withScope()` callback is only valid inside of the callback. Once the callback ends, the scope will be removed and no longer applied. The inner scope is only applied to events that are captured inside of the callback. `withScope()` will clone (or fork) the current scope, so that the current scope is not modified. This allows you to
 more easily isolate pieces of context information to specific locations in your code or
 even call <PlatformIdentifier name="clear" /> to briefly remove all context information.
+
+<PlatformCategorySection supported={['server']}>
+## Using `withIsolationScope`
+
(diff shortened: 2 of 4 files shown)


docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
```diff
@@ -0,0 +1,47 @@
+---
+title: Process Isolation
```
Lms24, line 2: m: I'm not sure if "Process" isolation is a good term for the heading. We worry about isolating certain jobs/tasks/requests _within_ one process. 

I'd rather call this page "Request Isolation" or more generally "Task Isolation". WDYT?

docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
```diff
@@ -0,0 +1,47 @@
...
+  - javascript.sveltekit
+  - javascript.astro
+  - javascript.remix
+notSupported:
+  - javascript
+---
+
+In server-side environments, the <PlatformLink to='/enriching-events/scopes'>isolation scope</PlatformLink> is automatically forked around request boundaries. This means that each request will have its own isolation scope, and data set on the isolation scope will only apply to events captured during that request. This is done automatically by the SDK.
+
+However, the request isolation happens when the request itself is processed. This means that if you e.g. have a middleware where you want to set Sentry data (e.g. `Sentry.setUser()` in an auth middleware), you have to manually fork the isolation scope with `Sentry.withIsolationScope()` - see [Using withIsolationScope](../scopes/#using-withisolationscope).
```
Lms24, line 24: not actionable for docs but
that's a massive limitation actually 😬 I was under the impression that the OTEL http instrumentation would fork the isolation scope automatically before any express middleware is evaluated. Is that not the case?

docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
```diff
@@ -0,0 +1,47 @@
...
+    } else {
+      Sentry.setTag("Authenticated", true);
+      Sentry.setUser(authUser);
+      next();
+    }
+  });
+};
+```
+
+This way, the user & tag will only be attached to events captured during the request that passed the auth middleware.
```
Lms24, line 47: l: I'd move the sentence about bg jobs after the user/middleware example for consistency

```suggestion
The following example shows how you can use `withIsolationScope` to attach a user and a tag in an auth middleware:

```javascript
const auth = (req, res, next) => {
  Sentry.withIsolationScope(() => {
    const authUser = findUserForHeader(req.headers["authorization"]);
    if (!authUser) {
      Sentry.setTag("Authenticated", false);
      Sentry.setUser(null);
      next(new Error("Authentication Error"));
    } else {
      Sentry.setTag("Authenticated", true);
      Sentry.setUser(authUser);
      next();
    }
  });
};
```

This way, the user & tag will only be attached to events captured during the request that passed the auth middleware.

Manually calling `withIsolationScope` is also necessary if you want to isolate a non-request task, like background jobs or queue task.
```

**Review by Lms24 — COMMENTED** [review]
I'm a bit concerned about the middleware thing tbh. I always thought the http instrumentation would create the root span and fork the isolation scope before any middleware is evaluated. After all, we have `middleware` child spans inside an `http.server` span 🤔 

Docs-wise, I just had a couple of suggestions but good idea to more add more content for the isolation scope!

File: docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx (hunk 3 above)
Comment (Lms24) [reply docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx:47]: argh the nested ``` lines destroy the diff. I just moved the bg job sentence to the bottom and slightly adjusted it.

**Comment by sbriceland** [comment]
Thanks for taking the time to improve the docs! I just wanted to add my two cents in case it may help some folks down the road. 

From the [current documentation](https://docs.sentry.io/platforms/javascript/enriching-events/scopes/#isolation-scope):
image

This portion of the existing docs is what lead me to believe that I could use the `Sentry.setXXX` methods such that when a `Sentry.captureXXX` is called it would auto-magically have the appropriate scoped data.

While this is an accurate description of Isolation Scope behavior, I believe it would be helpful to call out  **when** you are inside of an isolation scope and a reference to your updated docs here. This way developers have a heads up that Isolation Scope isn't guaranteed e.g. when inside of middleware, unless the isolation scope is manually forked/setup.

**Comment by sbriceland** [comment]
> It would be ideal if we could somehow auto-fork this for middlewares, but in the meanwhile...

On that point, either the `@sentry/node` lib could include a helper method like the following. Or since it's simple enough, the docs could let people know something like this should work for Express

```ts
import * as Sentry from '@sentry/node';
import type { RequestHandler } from 'express';

export const setupExpressIsolationScopeMiddleware: RequestHandler = (_req, _res, next) => {
    Sentry.withIsolationScope(() => {
        next();
    });
};
```

Set up isolation scope pre request for Express middleware:

```ts
const app: Application = express();

// before routes are added to app
app.use(setupExpressIsolationScopeMiddleware);
```

docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
nwalters512: In my app, this _is_ the case with the "standard" OTel instrumentation packages. I can't speak to Sentry's internals, but it's very very strange that this doesn't just work out of the box.

Regardless of that: the language "when the request is processed" is, to me, imprecise and maybe even inaccurate. I consider middleware execution to very much be part of "request processing" - after all, I have middleware parsing request bodies, performing authorization checks, redirecting authenticated requests to the login page, and so on. My app has many requests that are completely handled by middleware!

docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
mydea: Yeah, I was also not 100% sure about this. My first instinct was also to go with `Request Isolation` but this does not really cover e.g. isolation background jobs 🤔 but maybe request isolation is still more understandable 🤔

docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
mydea: so in otel you do not have this concept of scopes and capturing data on them for a request etc., so this problem does not really exist there, as far as I can tell.

We have the problem that the way that the auto-isolation works is actually forking the isolation scope after a http.server span is emitted. And this is of course only emitted once the request is actually "done" (or at least started), not when the middlewares run. I agree that this kind of sucks and we should see if we find a way to solve this better. I'll adjust the wording for this for now!

docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
mydea: hmm, I do not want to bury this at the very bottom because it is actually a very important aspect that we don't really cover anywhere 🤔 like, if you have a background job or similar you will _definitely always_ have to do this manually 🤔

docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
mydea: OK, I did some more digging - actually this is not correct! Middlewares _are_ actually correctly isolated in express, so I will remove this section again (the general content is still good). 

@nwalters512 the problem in your app is not a fundamental problem with middlewares, but it happens because you added http instrumentation yourself, which overwrites ours (and this is what handles the auto isolation). See https://docs.sentry.io/platforms/javascript/guides/node/opentelemetry/custom-setup/#custom-http-instrumentation for details on this.  Sentry relies on the http instrumentation for a bunch of things. It would be nice to find a way to warn about this, though 🤔

mydea:
After some more digging, I figured out that actually middlewares do work as expected, the problem was a different one - see https://github.com/getsentry/sentry-docs/pull/11378#discussion_r1764605970.

However, this PR is still good to have because this content was missing, which was an oversight anyhow!

Lms24 (approved):
Thanks for making the changes! LGTM now!

docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
sbriceland: Hmmm.. if we can't create an instance of of OTEL http instrumentation, how are we supposed to go about registering it with our existing OTEL setup?

You're suggesting:

```ts
// do not import/create a new HttpInstrumentation instance
// import { HttpInstrumentation } from '@opentelemetry/instrumentation-http';
const sentryHttpIntegration = Sentry.httpIntegration();

const sentryClient = Sentry.init({
    dsn: process.env.SENTRY_DSN,
    integrations: (defaultIntegration) =>
        defaultIntegration.map((integration) => (integration.name === 'Http' ? sentryHttpIntegration : integration)),
    release,
    environment,
    skipOpenTelemetrySetup: true,
    tracesSampleRate: 1.0,
});
```

if that's the case what we're unable to [register the instrumentation](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-sdk-trace-node/README.md#instrumentation-configuration)

mydea: You don't need to add the http instrumentation yourself at all - just do not run `registerInstrumentations({ instrumentation: [new HttpInstrumentation()] })`, this happens automatically under the hood. Does that lead to problems/does not work as expected for you, with your custom setup? 🤔 We could expose our HttpInstrumentation and allow you to add this yourself, I guess, if that's necessary!

sbriceland: I updated my demo sentry + OTEL setup as follows:

```ts
import * as Sentry from '@sentry/node';
import { SentryPropagator, SentrySampler } from '@sentry/opentelemetry';
// import { registerInstrumentations } from '@opentelemetry/instrumentation';
// import { ExpressInstrumentation } from '@opentelemetry/instrumentation-express';
// import { HttpInstrumentation } from '@opentelemetry/instrumentation-http';
import { BatchSpanProcessor } from '@opentelemetry/sdk-trace-base';
import { NodeTracerProvider } from '@opentelemetry/sdk-trace-node';

const release = 'sentry-scope-testing';
const environment = 'local';

const sentryClient = Sentry.init({
    dsn: process.env.SENTRY_DSN,
    beforeSend: (event, hint, ...args) => {
        const { contexts, exception, extra, tags, message, user } = event;
        console.dir(
            {
                whoami: 'sentry:beforeSend',
                event: { contexts, exception, extra, tags, message, user },
                hint,
                args,
            },
            { depth: null }
        );
        return event;
    },
    release,
    environment,
    skipOpenTelemetrySetup: true,
    tracesSampleRate: 1.0,
});

const provider = new NodeTracerProvider({
    sampler: new SentrySampler(sentryClient!),
});

provider.addSpanProcessor(
    // mock processor to avoid polluting console output
    new BatchSpanProcessor({
        export: (_spans, cb) => cb({ code: 0 }),
        shutdown: () => Promise.resolve(),
        forceFlush: () => Promise.resolve(),
    })
);

provider.register({
    propagator: new SentryPropagator(),
    contextManager: new Sentry.SentryContextManager(),
});

// registerInstrumentations({
//     instrumentations: [new ExpressInstrumentation(), new HttpInstrumentation()],
// });

Sentry.validateOpenTelemetrySetup();

export const initializedSentry = Sentry;
```

Then for the express middleware, i removed forking of the isolation scope to see if Middleware is correctly isolated in express:

```ts
const auth: RequestHandler = (req, res, next) => {
    const authUser = Users.find((u) => u.id === req.headers['authorization']);
    // Sentry.withIsolationScope(() => {
    if (!authUser) {
        Sentry.setTag('Authenticated', false);
        Sentry.setTag('UserID', null);
        Sentry.setUser(null);
        next(new Error('Authentication Error'));
    } else {
        Sentry.setTag('Authenticated', true);
        Sentry.setTag('UserID', authUser.id);
        Sentry.setUser(authUser);
        res.locals.authUser = authUser;
        next();
    }
    // });
};
```

However, my output reverted to showing the issue:

```js
{
  whoami: 'sentry:beforeSend',
  event: { ... },
    exception: { ... },
    extra: {
      expectedUser: { id: '4', email: 'foo4@example.com', name: 'foo example4' }
    },
    tags: { Authenticated: true, UserID: '1' },
    message: undefined,
    user: { id: '1', email: 'foo@example.com', name: 'foo example' }
  },
  hint: { ... },
  args: []
}
```

sbriceland: Furthermore, switching my "mock exporter" to the `ConsoleSpanExporter` does not print an OTEL spans

```ts
provider.addSpanProcessor(
    // mock processor to avoid polluting console output
    new BatchSpanProcessor(new ConsoleSpanExporter())
    // new BatchSpanProcessor({
    //     export: (_spans, cb) => cb({ code: 0 }),
    //     shutdown: () => Promise.resolve(),
    //     forceFlush: () => Promise.resolve(),
    // })
);
```

mydea: I opened a PR here: https://github.com/sbriceland/sentry-debug/pull/1 where we can discuss this more!

AbhiPrasad (approved):

docs/platforms/javascript/common/enriching-events/scopes/index.mdx
```diff
@@ -143,3 +143,11 @@ In the following example we use <PlatformIdentifier name="with-scope" /> to atta
...
 more easily isolate pieces of context information to specific locations in your code or
 even call <PlatformIdentifier name="clear" /> to briefly remove all context information.
+
+<PlatformCategorySection supported={['server']}>
+## Using `withIsolationScope`
+
+`withIsolationScope` works fundamentally the same as `withScope`, but it will fork the isolation scope instead of the current scope. Generally, the isolation scope is meant to be forked less frequently than the current scope, and in most cases the SDK will handle this automatically for you. But in cases where you e.g. want to capture SDK events in a middleware (which happens before the request is processed and thus before the SDKs automatic handling), or if you want to isolate a non-request process (e.g. a background job), you can use `withIsolationScope` to create a new isolation scope that is only active for the duration of the callback:
+
+<PlatformContent includePath="enriching-events/scopes/with-isolation-scope" />
+</PlatformCategorySection>
```
lizokm, line 153:
```suggestion
`withIsolationScope` works fundamentally the same as `withScope` with the exception that it forks the isolation scope instead of the current scope. Generally, the isolation scope is meant to be forked less frequently than the current scope, and in most cases the SDK will handle this automatically for you. 

But in cases where you for example, want to capture SDK events in a middleware (which happens before the request is processed and thus before the SDK can automatically handle it), or if you want to isolate a non-request process (for example, a background job), you can use `withIsolationScope` to create a new isolation scope that's only active for the duration of the callback by instrumenting the below:

<PlatformContent includePath="enriching-events/scopes/with-isolation-scope" />
</PlatformCategorySection>
```

docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
lizokm: I vote for "Request Isolation". I think it's most clear and I think that it's ok that it doesn't cover everything. "Data isolation" is super broad and can mean just about anything. Same thing with "Process isolation".

lizokm (approved):
Made a few comments. Looks good!

docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
```diff
@@ -0,0 +1,36 @@
...
+  - javascript.remix
+notSupported:
+  - javascript
+---
+
+In server-side environments, the <PlatformLink to='/enriching-events/scopes'>isolation scope</PlatformLink> is automatically forked around request boundaries. This means that each request will have its own isolation scope, and data set on the isolation scope will only apply to events captured during that request. This is done automatically by the SDK.
+
+However, there are also other cases where you may want to have isolation, for example in background jobs or when you want to isolate a specific part of your code. In these cases, you can use `Sentry.withIsolationScope()` to create a new isolation scope that is valid inside of the callback you pass to it - see [Using withIsolationScope](../scopes/#using-withisolationscope).
+
+The following example shows how you can use `withIsolationScope` to attach data for a specific job run:
```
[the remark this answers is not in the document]
mydea: Maybe slightly different:

```suggestion
However, there are also other times when you may want to have isolation, for example, in background jobs or when you want to isolate a specific part of your code. In these cases, you can use `Sentry.withIsolationScope()` to create a new isolation scope that's valid inside of the callback you pass to it. Learn more about using [withIsolationScope](../scopes/#using-withisolationscope).

The following example shows how you can use `withIsolationScope` to attach data to a specific job run:
```

docs/platforms/javascript/common/enriching-events/process-isolation/index.mdx
mydea: will rename it!

docs/platforms/javascript/common/enriching-events/scopes/index.mdx
mydea: I removed the part about middlewares because it turns out it is not actually correct. rewrote it slightly!

mydea:
FYI I updated this and incorporated the feedback, will merge it soon.

docs/platforms/javascript/common/enriching-events/scopes/index.mdx
```diff
@@ -143,3 +143,13 @@ In the following example we use <PlatformIdentifier name="with-scope" /> to atta
 The scope inside the `withScope()` callback is only valid inside of the callback. Once the callback ends, the scope will be removed and no longer applied. The inner scope is only applied to events that are captured inside of the callback. `withScope()` will clone (or fork) the current scope, so that the current scope is not modified. This allows you to
 more easily isolate pieces of context information to specific locations in your code or
 even call <PlatformIdentifier name="clear" /> to briefly remove all context information.
+
+<PlatformCategorySection supported={['server']}>
+## Using `withIsolationScope`
+
+`withIsolationScope` works fundamentally the same as `withScope`, but it will fork the isolation scope instead of the current scope. Generally, the isolation scope is meant to be forked less frequently than the current scope, and in most cases the SDK will handle this automatically for you.
```
BYK, line 150: For a person who is not very familiar with Sentry SDK, this sentence is probably too dense for me. I know what request isolation and isolating scopes are but what exactly is "forking a scope". What is the current scope and how it is determined? Why is is meant to be forked less frequently and how is it relevant to me or this conversation? What are those "most cases" where the SDK handles for me. At least what _kind_ of cases are meant to be automatically handled?

mydea: we (try to) explain these things in the page above this section (see https://docs.sentry.io/platforms/javascript/guides/node/enriching-events/scopes/). I do agree that all of this is not super easy to understand, feel free to PR changes to wording to this!

BYK: > I do agree that all of this is not super easy to understand, feel free to PR changes to wording to this!

That's the thing! I would love to but I 100% don't understand what's going on. The questions I raised in my first comment are questions I don't know the answers to 😅

elliotdickison:
I am initializing Sentry for Express as instructed in the setup guide (making sure that Sentry is initialized before express is imported), do not have any custom instrumentation, and am setting tags on the isolation scope per the instructions here. Despite all that tags are still not set per-request. This is a massive problem for us. Any ideas?

Issue #11128, opened 2024-08-21
Missing documentation about the SENTRY_PROFILER_BINARY_DIR and SENTRY_PROFILER_BINARY_PATH env variable


Jimmy89:
### Core or SDK?

Platform/SDK

### Which part? Which one?

@sentry/profiling-node

### Description

I was debugging errors from my webpack output (packing node_modules as well) for our lambda, as I received a missing `SENTRY_FILE.node` error.

After some debugging, I found out that the @sentry/profiling-node package dynamically lookup a compiled .node file, where the path can be set with SENTRY_PROFILER_BINARY_DIR or SENTRY_PROFILER_BINARY_PATH. However, I did not find any documentation within Sentry that these env variables exist.

### Suggested Solution

Document these two variables.

JonasBa:
@Jimmy89 We intentionally did not document this as we do not really want to encourage the usage - in most cases this should never be used. Is this something that you need in your project?

Jimmy89:
@JonasBa yes, as we used webpack to minimize our build (a copy of our node_modules and code exceeded the 50mb zip file limit), we used webpack to minimize our build (including the node_modules artifacts). We got errors from the sentry package which stated that the binaries could not be found within the build path location (running within the lambda environment). Therefore, I started digging into the code to find a way to patch the lookup path for the binary. Resulting into finding these env variables, including the `sentry-prune-profiler-binaries`. By using both (env and pruner) I managed to create a very minimal and working package in our lambda, without monkey-patching any code.

We used the [lambda layer](https://docs.sentry.io/platforms/javascript/guides/aws-lambda/install/), which should also have the binaries, but with the latest major version (v8) of Sentry EcmaScript code (which we use) are assigned to the NPM modules. In combination with a webpack build you must use those Sentry env variables. Therefore, I think it's good to document these, as it will help others.

andreiborza:
Hi @Jimmy89, we have https://github.com/getsentry/sentry-javascript/issues/13049 in our backlog aiming to create a dedicated layer for Lambda, but didn't get to it yet.

This might still be nice to document tho @JonasBa wdyt?

JonasBa:
@andreiborza yes, I'm adding the doc today. I planned on adding it yesterday, but ended up prioritizing other work.

JonasBa:
Added the documentation in https://github.com/getsentry/sentry-docs/pull/11382

PR #11382, opened 2024-09-17
ref: document all profiler runtime flags

JonasBa:
## DESCRIBE YOUR PR
Documents profiler runtime flags, fixes https://github.com/getsentry/sentry-docs/issues/11128

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

### DIFF
--- a/docs/platforms/javascript/guides/aws-lambda/profiling/index.mdx
+++ b/docs/platforms/javascript/guides/aws-lambda/profiling/index.mdx
@@ -73,7 +73,7 @@ Under the hood, the Sentry profiler uses V8's [CpuProfiler](https://v8docs.nodes
 
 ## Runtime Flags
 
-There are three runtime flags that you can set which the behavior of the profiler. Two of the flags relate to how the SDK resolves the profiler binaries and another flag which alters how the underlying profiler is initialized by v8.
+There are three runtime flags that you can set which control the behavior of the profiler. Two of the flags relate to how the SDK resolves the profiler binaries and another flag which alters how the underlying profiler is initialized by v8.
 
 These flags are for advanced used cases and setting them is not required in majority of the cases.
 
--- a/docs/platforms/javascript/guides/aws-lambda/profiling/index.mdx
+++ b/docs/platforms/javascript/guides/aws-lambda/profiling/index.mdx
@@ -79,7 +79,7 @@ These flags are for advanced used cases and setting them is not required in majo
 
 - SENTRY_PROFILER_BINARY_PATH
 
-    This flag sets the profiler binary path and bypasses arch, platform and libc checks. It is useful if you want to override which binary is required at runtime, which can be useful in some build configurations.
+    This flag sets the profiler binary path and bypasses arch, platform and libc checks. It can be useful in some build configurations if you want to override which binary is required at runtime.
 - SENTRY_PROFILER_BINARY_DIR
 
   Acts similarly as the flag above, however this flag only specifies the directory where the binaries are located and defers to the runtime to resolve the correct binary depending on the arch, platform and libc version.
--- a/docs/platforms/javascript/guides/aws-lambda/profiling/index.mdx
+++ b/docs/platforms/javascript/guides/aws-lambda/profiling/index.mdx
@@ -73,16 +73,30 @@ Under the hood, the Sentry profiler uses V8's [CpuProfiler](https://v8docs.nodes
 
 ## Runtime Flags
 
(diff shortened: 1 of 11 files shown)


docs/platforms/javascript/guides/aws-lambda/profiling/index.mdx
```diff
@@ -73,17 +73,31 @@ Under the hood, the Sentry profiler uses V8's [CpuProfiler](https://v8docs.nodes
 
 ## Runtime Flags
 
-The default mode of the v8 CpuProfiler is [kEagerLogging](https://v8docs.nodesource.com/node-18.2/d2/dc3/namespacev8.html#a874b4921ddee43bef58d8538e3149374) which enables the profiler even when no profiles are active—this is good because it makes calls to `startProfiling` fast with the tradeoff of constant CPU overhead. This behavior can be controlled via the `SENTRY_PROFILER_LOGGING_MODE` environment variable with values of `eager|lazy`. If you opt to use the lazy logging mode, calls to `startProfiling` may be slow. (Depending on environment and node version, it can be in the order of a few hundred ms.)
+There are three runtime flags that you can set which the behavior of the profiler. Two of the flags relate to how the SDK resolves the profiler binaries and another flag which alters how the underlying profiler is initialized by v8.
```
andreiborza, line 76:
```suggestion
There are three runtime flags that you can set which control the behavior of the profiler. Two of the flags relate to how the SDK resolves the profiler binaries and another flag which alters how the underlying profiler is initialized by v8.
```
missing word here, `control` or `change` or something?

andreiborza (commented):
m: Are these docs ever different between platforms? Wonder if we could just lift them up to `common` and avoid the duplication.

docs/platforms/javascript/guides/aws-lambda/profiling/index.mdx
```diff
@@ -73,17 +73,31 @@ Under the hood, the Sentry profiler uses V8's [CpuProfiler](https://v8docs.nodes
...
 
-Here's an example of starting a server with lazy logging mode:
+These flags are for advanced used cases and setting them is not required in majority of the cases.
```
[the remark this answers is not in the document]
JonasBa: Yup, good idea

includes/profiling-node-runtime-flags.mdx
```diff
@@ -0,0 +1,28 @@
...
+  The default mode of the v8 CpuProfiler is [kEagerLogging](https://v8docs.nodesource.com/node-18.2/d2/dc3/namespacev8.html#a874b4921ddee43bef58d8538e3149374) which enables the profiler even when no profiles are active—this is good because it makes calls to `startProfiling` fast with the tradeoff of constant CPU overhead. This behavior can be controlled via the `SENTRY_PROFILER_LOGGING_MODE` environment variable with values of `eager|lazy`. If you opt to use the lazy logging mode, calls to `startProfiling` may be slow. (Depending on environment and node version, it can be in the order of a few hundred ms.)
+
+  Here's an example of starting a server with lazy logging mode:
+
+  ```bash
+  # Run profiler in lazy mode
+  SENTRY_PROFILER_LOGGING_MODE=lazy node server.js
+  ```
+
+We recommend you have your own CPU resource-monitoring in place, because the actual resource use could be environment-dependent.
```
lizokm, line 28:
```suggestion
- SENTRY_PROFILER_BINARY_PATH

This flag sets the profiler binary path and bypasses arch, platform, and libc checks. It can be useful in some build configurations if you want to override which binary is required at runtime.
    
- SENTRY_PROFILER_BINARY_DIR

Acts similarly to the flag above, however, this flag only specifies the directory where the binaries are located and defers to the runtime to resolve the correct binary depending on the arch, platform, and libc version.

- SENTRY_PROFILER_LOGGING_MODE

The default mode of the v8 CpuProfiler is [kEagerLogging](https://v8docs.nodesource.com/node-18.2/d2/dc3/namespacev8.html#a874b4921ddee43bef58d8538e3149374), which enables the profiler even when no profiles are active—this is good because it makes calls to `startProfiling` fast with the tradeoff of constant CPU overhead. This behavior can be controlled via the `SENTRY_PROFILER_LOGGING_MODE` environment variable with values of `eager|lazy`. If you opt to use the lazy-logging mode, calls to `startProfiling` may be slow. (Depending on environment and node version, it can be in the order of a few hundred ms.)

Here's an example of starting a server with lazy-logging mode:

  ```bash
  # Run profiler in lazy mode
  SENTRY_PROFILER_LOGGING_MODE=lazy node server.js
  ```

We recommend that you have your own CPU resource-monitoring in place, because the actual resource use could be environment-dependent.
```

lizokm (approved):
Left a few suggestions, but otherwise looks great from my end.

PR #11398, opened 2024-09-19
ref(stats): Add project abuse limit explanation

priscilawebdev:
closes https://github.com/getsentry/sentry/issues/77770


priscilawebdev:
This was clarified internally and is an outcome we want to expose as internal - as it is already.

PR #11413, opened 2024-09-23
fix(docs): Fix missing python configure SDK code snippet

vgrozdanic:
In our SDK docs, for most of the platforms we were missing init SDK code snippet for python in **Configure** section (e.g. [AIOHTTP](https://docs.sentry.io/platforms/python/integrations/aiohttp/)).

This is because we were missing `python.mdx` as a platfrom in our `platform-includes/getting-started-config/` directory.

The code snippet here is the same as for the default Python SDK configuration.


sentrivana (approved):
Thanks for spotting this and for the fix! LGTM.

PR #11431, opened 2024-09-25
fix(browser): Update supported browsers due to `globalThis` usage

timfish:
Before the v8 release we decided to use `globalThis` to simplify access the global object across all platforms and contexts.

`globalThis` is not ES2018 but we didn't consider this when the minimum supported browser versions were documented.


chargome:
@timfish can you rebase on master, we had an issue with the vercel deployments 🥲

timfish:
Done!

chargome (approved):

coolguyzone (approved):
Looks good 👍

PR #11441, opened 2024-09-26
feat(develop): Add Tracing without Performance documentation

Lms24:
This PR adds a develop docs page to our SDK->Telemetry->Tracing section for "Tracing without Performance" (TwP). TWP has been around for some time and most (all?) SDKs support it today. However, we didn't specify the behaviour in develop docs yet. This page specifies the expected high-level behaviour on continuing and propagating traces in TwP as well as attaching trace data to events and envelopes. 

This specification is based on the previously existing internal [Notion doc](https://www.notion.so/sentry/Tracing-without-performance-efab307eb7f64e71a04f09dc72722530) as well as some discussions we recently had around TwP and its implications. 

[Page Preview](https://develop-docs-git-lms-feat-twp.sentry.dev/sdk/telemetry/traces/tracing-without-performance/)

### DIFF
--- a/develop-docs/sdk/telemetry/traces/tracing-without-performance.mdx
+++ b/develop-docs/sdk/telemetry/traces/tracing-without-performance.mdx
@@ -76,7 +76,7 @@ Note, that for incoming requests with trace headers, SDKs should still continue 
 SDKs implementing TwP mode must adhere to the behavior described in this section.
 Everything else (e.g. how to store the trace data and specific tracing options) is considered an implementation detail and can be implemented as needed.
 
-### Continuing incoming traces
+### Continuing Incoming Traces
 
 SDKs in TwP mode must continue incoming traces (e.g. from incoming HTTP requests) and attach the trace data of that trace to any events that are created during the lifetime of this trace.
 Furthermore, they must propagate the trace as usual to any outgoing requests.
--- a/develop-docs/sdk/telemetry/traces/tracing-without-performance.mdx
+++ b/develop-docs/sdk/telemetry/traces/tracing-without-performance.mdx
@@ -17,7 +17,7 @@ This means that SDKs in this mode always:
 - attach the `trace` envelope header to Sentry envelopes, populated from the dynamic sampling context
 - propagate trace data (`sentry-trace`, `baggage`) via the usual channels (e.g. HTTP headers, `<meta> HTML tags`, etc), with the correct [sampling decision](#sampling-decision)
 
-If users [explicitly opt out](#completely-disable-tracing) of all tracing capabilities, TwP mode shall no longer be active.
+If users [explicitly opt out](#completely-disable-tracing) of all tracing capabilities, TwP mode will no longer be active.
 
 ## Configuration
 
--- a/develop-docs/sdk/telemetry/traces/tracing-without-performance.mdx
+++ b/develop-docs/sdk/telemetry/traces/tracing-without-performance.mdx
@@ -107,8 +107,8 @@ Traces in TwP mode should have the same duration as regular traces.
 For example, a TwP trace for a backend server should generally last for the duration of one individual request.
 This usually corresponds with the lifetime of an isolation scope (or current scope created within the isolation scope).
 
-SDKs in TwP mode must store trace data in a way that it can be attached to events and propagated to outgoing requests.
-The exact storage mechanism is up to the SDK implementation but it is recommended to use the same mechanism as for regular traces.
+SDKs in TwP mode must store trace data in a way that makes it possible for them to be attached to events and propagated to outgoing requests.
+The exact storage mechanism is up to the SDK implementation, but we recommended using the same mechanism as for regular traces.
 Usually this means storing the data on the scope in a field called `propagationContext` as [recommended here](/sdk/telemetry/traces/distributed-tracing/).
 
 On a related note, the `propagationContext` should be populated with a random `traceId` and `spanId` if no incoming trace is present.
--- a/develop-docs/sdk/telemetry/traces/tracing-without-performance.mdx
+++ b/develop-docs/sdk/telemetry/traces/tracing-without-performance.mdx
@@ -111,7 +111,7 @@ SDKs in TwP mode must store trace data in a way that it can be attached to event
 The exact storage mechanism is up to the SDK implementation but it is recommended to use the same mechanism as for regular traces.
 Usually this means storing the data on the scope in a field called `propagationContext` as [recommended here](/sdk/telemetry/traces/distributed-tracing/).
 
-Related, the `propagationContext` should be populated with a random `traceId` and `spanId` if no incoming trace is present.
+On a related note, the `propagationContext` should be populated with a random `traceId` and `spanId` if no incoming trace is present.
 This—in combination with the `sentry-trace` header specification requiring a `spanId`—has an important implication on the Sentry product:
 A non-existing `spanId` will be propagated along with the trace and attached to events.
 While not ideal, we accept this limitation as the Sentry product can and should handle non-existing (parent) spans anyway.
--- a/develop-docs/sdk/telemetry/traces/tracing-without-performance.mdx
+++ b/develop-docs/sdk/telemetry/traces/tracing-without-performance.mdx
@@ -88,7 +88,7 @@ If an SDK in TwP mode doesn't receive an incoming trace, it should start a new t
 In this case, the new trace is not sampled (as in, there is no sampling decision, neither positive nor negative).
 Instead, the sampling decision is _deferred_ to the next downstream SDK.
 
-This means that
+This means that:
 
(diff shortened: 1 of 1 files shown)


develop-docs/sdk/telemetry/traces/tracing-without-performance.mdx
```diff
@@ -0,0 +1,134 @@
...
+This usually corresponds with the lifetime of an isolation scope (or current scope created within the isolation scope).
+
+SDKs in TwP mode must store trace data in a way that it can be attached to events and propagated to outgoing requests.
+The exact storage mechanism is up to the SDK implementation but it is recommended to use the same mechanism as for regular traces.
+Usually this means storing the data on the scope in a field called `propagationContext` as [recommended here](/sdk/telemetry/traces/distributed-tracing/).
+
+Related, the `propagationContext` should be populated with a random `traceId` and `spanId` if no incoming trace is present.
+This—in combination with the `sentry-trace` header specification requiring a `spanId`—has an important implication on the Sentry product:
+A non-existing `spanId` will be propagated along with the trace and attached to events.
+While not ideal, we accept this limitation as the Sentry product can and should handle non-existing (parent) spans anyway.
```
lizokm, line 117:
```suggestion
On a related note, the `propagationContext` should be populated with a random `traceId` and `spanId` if no incoming trace is present.
This—in combination with the `sentry-trace` header specification requiring a `spanId`—has an important implication on the Sentry product:
A non-existing `spanId` will be propagated along with the trace and attached to events.
While not ideal, we accept this limitation as the Sentry product can and should handle non-existing (parent) spans anyway.
```

develop-docs/sdk/telemetry/traces/tracing-without-performance.mdx
```diff
@@ -0,0 +1,134 @@
...
+Given that no span is actually available in TwP mode, the DSC will not contain any keys related to spans (`transaction`, `sample_rate` or `sampled`).
+
+In SDKs adapting OpenTelemetry's tracing capabilities ([POTel](/sdk/hub_and_scope_refactoring/#f-use-otel-for-performance-instrumentation-potel)), the TwP trace data could also be stored in a non-recording span.
+For Example, the Node SDK also starts non-recording Otel spans in TwP mode and takes the trace data from them. In case no span is started (e.g. a node script without request handling), it falls back to the propagation context on the scope.
+Note that in the case of using the non-recording span, the span is also not sampled, meaning the sampling decision must still be deferred [when starting a new Trace](#starting-a-new-trace).
+
+## Historical Context
+
+TwP mode was introduced after SDKs were already capable of sending spans and transactions. Before TwP, SDKs would only handle traces if span sending was enabled and otherwise not attach trace data to any events.
+The primary motivation for TwP was to ensure that errors across application or service boundaries are still connected if they occurred in the same trace.
```
lizokm, line 129:
```suggestion
TwP mode was introduced after SDKs were already capable of sending spans and transactions. Before TwP, SDKs would only handle traces if span-sending was enabled. Otherwise, no trace data would be attached to any events.
The primary motivation for TwP was to ensure that errors across application or service boundaries were still connected if they occurred in the same trace.
```

develop-docs/sdk/telemetry/traces/tracing-without-performance.mdx
```diff
@@ -0,0 +1,134 @@
...
+
+The name "Tracing without Performance" was chosen because at the time of introduction, we associated spans purely with performance monitoring.
+Today, we associate spans more towards "Tracing" in general, and only (some of) the data from spans with Performance Monitoring.
+This is why the name from today's perspective is a bit misleading.
+As a mental model, think of TwP as "Tracing without Spans".
```
lizokm, line 134:
```suggestion

The name "Tracing without Performance" was chosen because at the time of introduction, we associated spans purely with performance monitoring.
Today, we associate spans with "Tracing" in general, and only (some of) the data from spans with Performance Monitoring.
This is why the name from today's perspective is a bit misleading.
As a mental model, think of TwP as "Tracing without Spans".
```

lizokm (approved):
Thanks for writing these!!

develop-docs/sdk/telemetry/traces/tracing-without-performance.mdx
```diff
@@ -0,0 +1,134 @@
...
+TwP mode can be disabled by either "upgrading" to Tracing _with_ Performance (see below) or by entirely opting out of trace propagation.
+
+<Alert>
+
+Note that some SDKs like the JavaScript Browser SDK require additional integrations to enable tracing all together.
+This is fine as long as there's a good reason why additional configuration is required (like keeping bundle size in check in JS).
+
+</Alert>
+
+### Upgrading to Tracing With Performance
```
cleptric, line 42: We should not call it like this, maybe just enable tracing or smth.

develop-docs/sdk/telemetry/traces/tracing-without-performance.mdx
```diff
@@ -0,0 +1,134 @@
...
+### Completely Disabling Tracing
+
+To completely opt out of any distributed tracing capabilities, users can pass an empty array (or language-equivalent parameter) to the `tracePropagationTargets` option. This will prevent the SDK from propagating trace information any further.
+
+```js
+Sentry.init({
+  dsn: "___PUBLIC_DSN___",
+  tracePropagationTargets: [],
+});
+```
```
cleptric, line 70: Might want to call out `tracesSampleRate: null`, too?

Lms24: What should `tracesSampleRate: null` do? At least in JS this isn't a valid value. If we set `tracesSampleRate: undefined` TwP is still active.

cleptric: So `null/nil/undefined` all has the same behaviour. It does no longer take trace parent headers with `-1` into account at all.

Lms24: Sorry but I'm a bit confused:

This section is about disabling tracing without performance, i.e. no longer propagating a trace. As far as I can tell, this is only the case when setting `tracePropagationTargets: []`. 

> It does no longer take trace parent headers with -1 into account at all

What does this mean specifically? If an incoming SDK receives a header with `-1`, and its `tracesSampleRate` is `null|undefined|nil`, should the SDK _not_ continue the positive sampling decision? Maybe I'm missing something obvious but I currently don't see this behaviour in the SDK code. 

To further add to the confusion, we would like to ensure that in the future, setting `tracesSampleRate: undefined` behaves in the same way as not setting `tracesSampleRate` at all. It's a bit intransparent to users to understand that there is difference while in both cases querying `options.tracesSamplerate` would return `undefined`.

cleptric: If this entire page only focuses on TwP, then the section header (Completely Disabling Tracing) confuses me 😬 

> What does this mean specifically? If an incoming SDK receives a header with -1, and its tracesSampleRate is null|undefined|nil, should the SDK not continue the positive sampling decision?

Yes, this is how it should work today.

We'll properly fix this when we re-work all SDKs towards only spans and the new sampling. `tracesSampleRate` will only be a float between `0.0` and `1.0`.

Lms24: > then the section header (Completely Disabling Tracing) confuses me

Fair but I think it's important to specify how you opt out of TwP. There are two options:
1. Enable regular tracing (with spans) - i.e. specify `tracesSampleRate: number` or `tracesSampler`. 
2. Completely disabling tracing. Should we rename the section? I was also thinking about "Disabling Tracing without Performance" but this could also entail 1... Open for better suggestions! 

> Yes, this is how it should work today.

Okay, so I think we're talking about 2 things:

- Assuming no other tracing options (`tracesSampler`/`enableTracing`) are set: TwP should be enabled if `tracesSampleRate` is not set at all or if `tracesSampleRate` is set to `undefined/null/nil`. 
I think that makes sense to me and is what we want to achieve in v9 of the JS SDK anyway. Today, there's a behavioural difference between setting `tracesSampleRate: undefined` and not setting it at all. But this shouldn't affect the develop spec. 

- Separately to this, we should document that SDKs in TwP mode, that receive incoming traces with `-1` (and `-0` as well?), should **not** forward the sampling decision, meaning, propagate `sentry-trace` with a deferred sampling decision. This should go under another section though because it isn't connected to Disabling TwP.

Lms24: As discussed in the call yesterday:

- Removed the "completely disable tracing" section in favour of "disable trace propagation"
- Noted explicitly that at this time, SDKs will always continue incoming traces in TwP mode. The only way to opt out of this would be to manually start a new trace (but this is admittedly a niche edge case, given that users would probably only do this if they'd also like to send spans). 
- Realized I already added a "historical context" section at the bottom but I now added a link on the top to it. Slightly reworked the historical context section though.

cleptric (approved):
Nice, thank you!

PR #11459, opened 2024-09-30
Update Go Gin

smeubank:
Install link went to a 404, needed to indicate a branch


cleptric (changes requested):
This is not a link, this is the install command. Did you mean to change something different?

PR #11462, opened 2024-09-30
Add troubleshooting section to React Native Annotate Components page

krystofwoldrich:
## DESCRIBE YOUR PR
Based on message from customer on Discord adding the troubleshooting steps to the docs as well.

https://discord.com/channels/621778831602221064/1286059289252986940/1290353709678858371

This steps applies on all the SDK version with this feature.

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 


docs/platforms/react-native/integrations/component-names.mdx
```diff
@@ -71,6 +71,10 @@ Here's what the resulting node would look like if your bundler had applied the p
 
 The Sentry browser SDK will pick off the value from these `data` attributes and collect them when your components are interacted with.
 
+## Troubleshooting
+
+- Use `export SENTRY_LOG_LEVEL=debug` to enable debug logs. When generating the JS Bundle, you should see logs about the Sentry Metro Plugin saving and loading the default Babel transformer.
```
kahest, line 76: is there any more guidance we can give in case the logs don't show the expected lies?

lizokm, line 76:
```suggestion
## Troubleshooting

Use `export SENTRY_LOG_LEVEL=debug` to enable debug logs. When generating the JS Bundle, you should see logs about the Sentry Metro Plugin saving and loading the default Babel transformer.
```
Since there's only thing here for now, it doesn't really make sense to make it bulleted. Also, would it be better to call this section "Enable Debug Logs"? I'm probably missing some context, but it's not clear to me what would have to go wrong in order for "enabling debug logs" to be the solution. Maybe it would be good to add a scenario in which it would help?

lizokm (commented):
I have a logistical question about the troubleshooting solution. @krystofwoldrich let me know if it would be helpful to chat for a few minutes :)

docs/platforms/react-native/integrations/component-names.mdx
krystofwoldrich: Thank you, yes, I've added more guidance.

krystofwoldrich: Thank you make sense, I've added more guidance. So I think now the bullet points make sense. But I'm okay leaving it as a paragraph.

docs/platforms/react-native/integrations/component-names.mdx
```diff
@@ -71,6 +71,13 @@ Here's what the resulting node would look like if your bundler had applied the p
 
 The Sentry browser SDK will pick off the value from these `data` attributes and collect them when your components are interacted with.
 
+## Troubleshooting
+
+- If component names are missing in the reported events, check if the annotations are present in the JS bundle generated by Metro. You can text search in the bundle for the `data-sentry-component` key.
+- To add the annotations `config.transformer.babelTransformerPath` is required to be set in the Metro config before calling `withSentryConfig`.
+- Function `getSentryExpoConfig` returns the configuration with the Sentry Babel Transformer set, do not overwrite `config.transformer.babelTransformerPath`.
+- Use `export SENTRY_LOG_LEVEL=debug` to enable debug logs and observe if Sentry detected the default transformer and if Sentry Babel Transformer was executed.
+
```
lizokm, line 80:
```suggestion
## Troubleshooting

### Missing Component Names 

If component names are missing from reported events, it could be because you don't have annotations. To troubleshoot this:

1. Check to see if the JS bundle generated by Metro has annotations by text searching for the `data-sentry-component` key in the bundle.
2. If annotations are missing, add them by setting `config.transformer.babelTransformerPath` in the Metro config before calling `withSentryConfig`.
3. Since the `getSentryExpoConfig` function returns the configuration with the Sentry Babel Transformer set, make sure not to overwrite `config.transformer.babelTransformerPath`.
4. Use `export SENTRY_LOG_LEVEL=debug` to enable debug logs so you can make sure that Sentry detected the default transformer and that the Sentry Babel Transformer was executed.

```
Hey there, I don't have enough understanding of the problem to know if this section makes sense as four steps. If it doesn't, let's set up some time to talk. You can walk me through this and we could decide how to frame it together :)

lizokm (approved):
I've made a suggestion, but I'm not sure it's the right one. If it's not, let's set up some time to chat through this.

docs/platforms/react-native/integrations/component-names.mdx
krystofwoldrich:
```suggestion
## Troubleshooting

### Missing Component Names 

If component names are missing from reported events, it could be because the annotations we not added to the JS bundle. To troubleshoot this:

1. Check to see if the JS bundle generated by Metro has annotations by text searching for the `data-sentry-component` key in the bundle.
2. If annotations are missing and you are using React Native without frameworks, check that `config.transformer.babelTransformerPath` is set before passing it to `withSentryConfig(config)`. If you are Expo user using `const config = getSentryExpoConfig(__dirname)` make sure the `config.transformer.babelTransformerPath` set by the `getSentryExpoConfig` is not overwritten.
4. Use `export SENTRY_LOG_LEVEL=debug` to enable debug logs to verify that Sentry detected the default transformer and that the Sentry Babel Transformer was executed.
```

The structure looks good, I just made some adjustments to keep the original meaning in. Let me know if this is make sense. I'm happy to chat about it.

PR #11480, opened 2024-10-03
fix: getSentryExpoConfig should receive __driname not config

krystofwoldrich:
Copy paste typo.


antonis (approved):
LGTM 🚀 

Note: if I'm not mistaken there is a [similar typo in RN manual setup](https://github.com/getsentry/sentry-docs/blob/master/docs/platforms/react-native/manual-setup/metro.mdx?plain=1#L37).

krystofwoldrich:
@antonis  Thanks, yes it's the same typo. I added the fix of that to this PR as well.

PR #11485, opened 2024-10-03
update CTAs

lizokm:
Update CTAs to match marketing site.


coolguyzone (approved):
Tested in various device widths and mobile, looks good 👍

PR #11495, opened 2024-10-06
onboarding: dotnet list package

bruno-garcia:
onboarding is incorrectly dictating only `Sentry.Profiling` when the profiling checkbox is chcked. It should add it to the existing `Sentry` package.

### DIFF
--- a/docs/platforms/dotnet/index.mdx
+++ b/docs/platforms/dotnet/index.mdx
@@ -53,14 +53,17 @@ Install the **NuGet** package to add the Sentry dependency:
   </Note>
 
   ```shell {tabTitle:.NET Core CLI}
+  dotnet add package Sentry -v {{@inject packages.version('sentry.dotnet') }}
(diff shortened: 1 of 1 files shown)


docs/platforms/dotnet/index.mdx
```diff
@@ -75,7 +78,6 @@ SentrySdk.Init(options =>
 {
     options.Dsn = "https://eb18e953812b41c3aeb042e666fd3b5c@o447951.ingest.us.sentry.io/5428537";
     options.Debug = true;
-    options.AutoSessionTracking = true;
```
bruno-garcia, line 78: Also removed this since it can't be added on a blank dotnet instruction. Otherwise this would be on by default.

On mobile for example it's already on by default. On aspnet it takes no effect since no server mode was added

bruno-garcia:
image

Looks OK now.

Please just merge after (if) approving

bitsandfoxes (approved):

PR #6404, opened 2023-03-02
feat(angular): Add Angular CLI + Webpack Plugin sourcemaps upload guide

Lms24:
This PR adds a new source maps upload guide to our Angular SDK docs: Users can make use of the Angular CLI and the Sentry Webpack plugin to upload source maps. This requires a custom Angular builder. Changes:

* Add a tutorial for Angular CLI + Webpack plugin
* Add Angular-specific source maps overview and source maps uploading index page
* Let the `PageGrid` component accept an `exclude` prop to exclude certain entries from being displayed
  * Otherwise we show source maps upload option twice, once in the "recommended" section and once in the "other options" section of the uploading index page. 

closes #6326

PR #11510, opened 2024-10-08
fix(components): Re-add `exclude` functionality in `PageGrid`

Lms24:
This PR re-adds the functionality of the `exclude` prop in the `PageGrid` component. I introduced this option in https://github.com/getsentry/sentry-docs/pull/6404 because I wanted to "feature" certain source maps upload guides for Angular but still display all other guides via `PageGrid`. Looks like this funcitonality got lost in one of the major refactorings of our docs repo. So this PR re-adds the filter functionality.

(Noticed while reviewing #11494)

### DIFF
--- a/docs/contributing/pages/components.mdx
+++ b/docs/contributing/pages/components.mdx
@@ -41,9 +41,7 @@ See also the [Note component](#note).
 
 Render an expandable section.
 
-<Expandable title="This is a title">
-This is some content
-</Expandable>
+<Expandable title="This is a title">This is some content</Expandable>
 
 ```markdown {tabTitle:Example}
 <Expandable title="This is a title">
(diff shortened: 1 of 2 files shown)


docs/contributing/pages/components.mdx
```diff
@@ -193,6 +192,7 @@ Attributes:
 
 - `header` (string) - optional header value to include, rendered as an H2
 - `nextPages` (boolean) - only render pages which come next based on sidebar ordering
+- `exclude` (string[]) - an array of pages to exclude from the grid. Specify the file name of the page, for example, `"index"` for `index.mdx`.
```
Lms24, line 195: hmm looks like this file wasn't formatted previously. I only added this line, rest of the file is just auto-format.

chargome (approved):

PR #11521, opened 2024-10-09
fix(platform): table hydration errors

chargome:
Incorrectly nested `<tr>` elements as direct children of `<table>` cause hydration errors as `<tbody>` elements get injected by the browser.


ryan953 (approved):
I know 100"% that react will inject missing tags like this. 

Lmk how the experience is to find each cause and get a PR out like this wrt the hydration issue in sentry!

coolguyzone (approved):
Tables look fine in preview. Adding the `<tbody>` tags should help with accessibility as well as the rendering errors 👍

Issue #11513, opened 2024-10-08
Add workflow for enforcing version-conventions


chargome:
Add a github workflow for enforcing versioned docs conventions.

relates to https://github.com/getsentry/sentry-docs/pull/11290#discussion_r1791529611

PR #11290, opened 2024-09-09
Versioned Docs

chargome:
closes [#3881](https://github.com/getsentry/sentry-docs/issues/3881)

This PR introduces versioned documentation for docs.sentry.io

## Features
- [x] Versioned files
- [x] Versioned platform includes
- [x] Warning banner 
- [x] Update metadata for versioned files
- [x] Check for versioning in developer docs
- [x] Version index files
- [x] Version top level guides
- [x] ~~Make redirects links~~
- [x] ~~(nice to have) redirect non-existing versioned pages on the server if a unversioned variant exists~~
- [ ] (OPTIONAL) Exclude versioned files from static build

## 🐛 Known bugs
- [x] Version selector does not update on soft navigation 
- [x] Versions are displayed on sidebar
- [x] Versioned pages do not include other versions
- [x] Original page should be highlighted in sidebar when versioned page is selected
- [x] Wrong color in version selector in dark mode

## Cleanup
- [x] Remove versioned demo pages 

NOTE: In the current state, versioned platform-redirects can only be displayed if their containing root page is also versioned

## NOTE FOR REVIEWERS:
Maybe read the contributing entry that is included in this pr first.

You can inspect a versioned file by browsing these paths on the preview url:
/platforms/javascript/guides/nextjs/manual-setup/ (scoped to Next.js)
/platforms/javascript/guides/nextjs/usage/ (scoped to JavaScript)
/platforms/javascript/guides/express/ (scoped to Express - this one also uses versioned platform includes)

PR #11522, opened 2024-10-09
ci: Add workflow for checking version conventions

chargome:
- Adds a GH workflow for enforcing file naming conventions for versioned files.
- Enables running vitest for CI

related to #11290 
fixes https://github.com/getsentry/sentry-docs/issues/11513

### DIFF
--- a/.github/workflows/enforce-version-convention.yml
+++ b/.github/workflows/enforce-version-convention.yml
@@ -1,0 +1,22 @@
+name: Enforce Version Conventions
+
+on:
+  push:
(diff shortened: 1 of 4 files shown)


Lms24 (approved):
Nice, thanks!

Should we already add a convention (+check) for future versions? Something like `next`/`preview`? We can also cross that bridge when we get to it. Just a though!

chargome:
Yeah I'd rather wait for that check, maybe this won't be used anyway 😄

PR #11528, opened 2024-10-09
meta: Develop docs issue template

szokeasaurusrex:
Add an issue template for [develop docs](https://develop.sentry.dev).


.github/ISSUE_TEMPLATE/issue-content-develop.yml
stephanie-anderson, line 1: The issue templates will later be listed ordered by their name. If you rename the file to `issue-content-03-develop.yml` it will become clear that you explicitly want to list it as 3rd option. (it will work anyways I guess, since numbers usually are listed first, but this way we would stick to the naming convention as well)

See https://github.com/getsentry/sentry-docs/blob/master/.github/ISSUE_TEMPLATE/

.github/ISSUE_TEMPLATE/issue-content-develop.yml
```diff
@@ -0,0 +1,38 @@
+name: "📝 Developer Documentation"
```
stephanie-anderson, line 1: And I think we need to consider the ambiguity of this title, as many of our users will think this applies to them.

What about:

```suggestion
name: "📝 Documentation on develop.sentry.dev"
```

szokeasaurusrex:
@stephanie-anderson addressed your feedback

stephanie-anderson (approved):

PR #11535, opened 2024-10-09
fix: Improve dark theme detection on codeKeywords (Dsn selector)

### LINKED ISSUE #11419: DSN Selection is unreadable when using dark mode
### Steps to Reproduce

1. Use dark mode
2. Click a DSN selector (I did so on this page: https://docs.sentry.io/platforms/native/guides/minidumps/)

### Expected Result

Dropdown should be readable

### Actual Result

Colors are off:

Image

### Additional Info

_No response_

a-hariti:
fixes #11419


a-hariti:
@chargome

TLDR; the issue seems to be fixed

the preview doesn't show the issue/solution as it only shows one project (example project)

if you want to verify, I suggest checkout the branch locally and edit the `src/components/codeContext.tsx` file and add another project (I copy pasted the existing one and changed the `PROJECT_ID` and the `title`

```javascript
export const DEFAULTS: CodeKeywords = {
  PROJECT: [
    // ...
  ]
}
```

chargome (approved):

PR #11536, opened 2024-10-09
fix: increase heading scroll margin under onboarding options

### LINKED ISSUE #11379: Anchor sections are hidden underneath onboarding options
### Problem Statement

When linking to specific sections on pages with `OnboardingOption`, the section ends up underneath the `OnboardingOption`.

**Expected Result**
Image

**Actual Result**
Image

### Solution Brainstorm

Use [scroll-margin-top](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-margin-top) on pages with onboarding options.

a-hariti:
fixes #11379


chargome (approved):

coolguyzone:
Hey @a-hariti , I verified that this is fixed on desktop, however on mobile it looks like the 2rem isn't enough and it still gets cut off. Maybe add a media query for this case?

Screenshot 2024-10-16 at 1 58 51 PM

PR #11547, opened 2024-10-11
fix: Update sorting order for guides on platform filter

chargome:
Based on feedback that the `node` guide is not immediately visible for filtering SDKs for `node`.

## Before:
Screenshot 2024-10-11 at 11 57 57

## After:
Screenshot 2024-10-11 at 11 57 41


lizokm (approved):
This looks good @chargome, thanks for updating! Is there a way to also change this for when you're filtering for platforms in the docs? The same issue is happening there...

chargome:
@lizokm I added it also to the platform selector in the sidebar (I hope that's what you meant 😅 )

PR #11563, opened 2024-10-14
feat(relay): Add document describing span and transaction quotas

Dav1dde:
## DESCRIBE YOUR PR

Description of how span and transaction outcomes are (to be) handled within Relay.

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

### DIFF
--- a/develop-docs/relay/transaction-span-ratelimits.mdx
+++ b/develop-docs/relay/transaction-span-ratelimits.mdx
@@ -1,0 +1,230 @@
+---
+title: Transaction and Span Rate Limiting
+---
+
+<Alert level="warning" title="Important">
+This document describes a desired state, it is not yet fully implemented in Relay.
+</Alert>
+
+
+Relay enforces quotas defined in Sentry and propagates them as rate limits. In most cases,
+this is a simple one-to-one mapping of data category to envelope item.
+
+The transaction and span rate limits are more complicated. Spans and Transactions both
+have a total and an additional indexed category. They are also closely related, a transaction is a container for spans.
+A dropped transaction results in dropped spans.
+
+The following document describes how transaction and span quotas interact with each other within Relay.
+
+<Alert level="info" title="Important">
+This section describes how Relay has to interpret quotas, SDKs and other consumers
+should read [SDK Development / Rate Limiting](/sdk/rate-limiting/).
+</Alert>
+
+
+Related Documentation:
+- [The Indexed Outcome Category](/application/dynamic-sampling/outcomes/)
+
+
+
+## Enforcement
+
+The following rules hold true:
+- A quota for the transaction category is equivalent to a quota for the spans category.
+- A quota for the spans category is equivalent to a quota for the transaction category.
+- When a transaction is dropped, all the contained spans should be reported in outcomes.
+- If either transactions or spans are rate limited, clients should receive a limit for both categories.
+- Indexed quotas only affect the payload for their respective category.
+
+These rules can be visualized using this table:
+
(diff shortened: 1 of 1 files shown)


develop-docs/relay/transaction-span-ratelimits.mdx
```diff
@@ -0,0 +1,230 @@
...
+
+An envelope containing a transaction with 3 child spans generates 4 outcomes for rate limited spans in the
+`spans_indexed` category. 1 count for the generated segment span from the transaction and 3 counts for the
+contained spans. The transaction itself will still be ingested.
+
+**Ingestion**:
+
+| Transaction Payload | Transaction Metrics | Span Payload | Span Metrics |
+| ------------------- | ------------------- | ------------ | ------------ |
+| ✅                  | ✅                  | ❌           | ✅           |
```
iambriccardo, line 120: Question: the reason we have span metrics here is that we emit the metric of the transaction that was ingested and treat it as a span metric besides a transaction metric?

Dav1dde: It goes back to the table defined above, for an indexed quota we **only** drop the payload but keep normal processing going (also produce metrics).

This in isolation (or current AM3 without sampling) may seem a bit weird, but in combination with dynamic sampling it makes more sense. Consider the case dynamic sampling has sampled a payload (we ingest metrics + payload), but then afterwards the abuse quota for indexed payloads forces us to filter the payload, we still want to ingest the metric and not drop it. Then there is no difference whether something is sampled or not, otherwise everything that is not sampled is ingested as metrics, but the few that are sampled are completely dropped.

I also try to explain this a bit in `Should I ever use an indexed category in a Quota?` when you should actually use and indexed quota and the answer is pretty much never, so this example here is a very rare case (or should be).

iambriccardo: Ah nvm, I misread it was not `span` but `indexed_span`. Thanks!

develop-docs/relay/transaction-span-ratelimits.mdx
```diff
@@ -0,0 +1,230 @@
...
+- Indexed quotas only affect the payload for their respective category.
+
+These rules can be visualized using this table:
+
+|                         | Transaction Payload | Transaction Metric | Span Payload | Span Metric |
+| ----------------------- | ------------------- | ------------------ | ------------ | ----------- |
+| **Transaction**         | ❌                  | ❌                 | ❌           | ❌          |
+| **Transaction Indexed** | ❌                  | ✅                 | ✅           | ✅          |
+| **Span**                | ❌                  | ❌                 | ❌           | ❌          |
+| **Span Indexed**        | ✅                  | ✅                 | ❌           | ✅          |
```
jjbayer, line 46: Could you define what ❌ | ✅ mean here?

develop-docs/relay/transaction-span-ratelimits.mdx
```diff
@@ -0,0 +1,230 @@
...
+`spans_indexed` category. 1 count for the generated segment span from the transaction and 3 counts for the
+contained spans. The transaction itself will still be ingested.
+
+**Ingestion**:
+
+| Transaction Payload | Transaction Metrics | Span Payload | Span Metrics |
+| ------------------- | ------------------- | ------------ | ------------ |
+| ✅                  | ✅                  | ❌           | ✅           |
+
+**Outcomes**:
```
jjbayer, line 122:
```suggestion
**Negative outcomes**:
```

Issue #11568, opened 2024-10-15
Lines in diff code block are shifted to the left


Lms24:
### Steps to Reproduce

1. Go to https://docs.sentry.io/platforms/javascript/guides/cloudflare/frameworks/sveltekit/
2. Observe removed/added lines after the two unchanged lines being shifted to the left

### Expected Result

The changed lines should vertically align with the unchanged lines and not be shifted to the left

### Actual Result

![Image](https://github.com/user-attachments/assets/1904e3ae-afcf-45c7-979f-62037f82ce30)

### Additional Info

I'll open a PR with a potential fix in a minute but I saw that https://github.com/getsentry/sentry-docs/pull/10996 basically introduced the CSS that's now causing the issue. So I'm wondering if there's a deeper problem here.

Lms24:
turns out this was caused by the specific code block language selection rather than a style bug. Closing

PR #11569, opened 2024-10-15
fix: Ensure diff code block lines are formatted correctly

Lms24:
This PR removes the negative margin from the changed lines in the diff view to fix the vertical alignment of the changed and unchanged lines I noticed in https://github.com/getsentry/sentry-docs/issues/11568. 

I'm not sure if this is the best way to fix this so please feel free to close the PR in favour of a better solution! Especially because it seems like we added the negative margin specifically to fix formatting previously in https://github.com/getsentry/sentry-docs/pull/10996/files.

potentially fixes https://github.com/getsentry/sentry-docs/issues/11568

before:
image

after: 
image

### DIFF
--- a/src/components/codeBlock/code-blocks.module.scss
+++ b/src/components/codeBlock/code-blocks.module.scss
@@ -79,22 +79,16 @@
 
   /* Diff highlighting, classes provided by rehype-prism-plus */
   /* Set inserted line (+) color */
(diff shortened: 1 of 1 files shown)


Lms24:
turns out this was caused by the specific code block language selection rather than a style bug. Closing this PR and will open another one in a sec

PR #11570, opened 2024-10-15
fix(cloudflare): Improve SvelteKit framework guide code block

Lms24:
This PR improves the SvelteKit Cloudflare guide code block a bit by

- fixing formatting of the diff block (TIL, the code block language needs to be set to `<language> diff`)
- re-adding the `handleError` hook just to point out that this remains unchanged from the default setup. (@AbhiPrasad Just to sanity-check: It _does_ remain unchanged, right?)

### DIFF
--- a/docs/platforms/javascript/guides/cloudflare/frameworks/sveltekit.mdx
+++ b/docs/platforms/javascript/guides/cloudflare/frameworks/sveltekit.mdx
@@ -25,9 +25,9 @@ compatibility_flags = ["nodejs_compat"]
 # compatibility_flags = ["nodejs_als"]
 ```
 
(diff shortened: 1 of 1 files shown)


AbhiPrasad (approved):
Yes the error handler should remain unchanged - thanks for the fix!

coolguyzone:
Hi @Lms24, thanks for the update!  I noticed that the example code here is for a file that could be a `.ts` or `.js` file. The code example provided is in TypeScript, should we also provide the alternative in JavaScript? Or do you think the TS example will suffice?

Lms24:
@coolguyzone the code for JS is identical (minus any type syntax but in this example there's none). I think for the moment, the TS example should suffice and we can spare us the duplication.

PR #11589, opened 2024-10-16
fix(integration): Adds a redirect from old launchdarkly doc link to new path

edwardgou-sentry:
The `/organization/integrations/launchdarkly/` path no longer works and should be redirected to `/organization/integrations/feature-flag/launchdarkly/`


stephanie-anderson (approved):
looks good!

PR #11628, opened 2024-10-21
fix(docs): Fix markdown bold formatting

MichaelSun48:
https://develop.sentry.dev/application/api/public/#faqs

image


lizokm (approved):
Thanks for the fix!

PR #11637, opened 2024-10-23
feat(apple): Update cocoa session replay FAQ

brustolin:
Adding an extra FAQ item for cocoa session replay.

### DIFF
--- a/docs/platforms/apple/guides/ios/session-replay/index.mdx
+++ b/docs/platforms/apple/guides/ios/session-replay/index.mdx
@@ -103,5 +103,8 @@ Errors that happen on the page while a replay is running will be linked to the r
 Q: Does Session Replay work with SwiftUI?
 A: Yes. It works with both UIKit and SwiftUI.
 
+Q: Why are parts of my replay not masked?
+A: Text views, input views, images, video players and webviews are all masked by default. Images with bundled assets, aren't masked because the likelihood of these assets containing PII is low. If you encounter a view that should be masked by default, consider opening a [GitHub issue](https://github.com/getsentry/sentry-cocoa/issues).
+
 Q: What's the lowest version of iOS supported?
-A: Session Replay recording happens even on the lowest version supported by the Sentry SDK. (Today that's iOS 12.)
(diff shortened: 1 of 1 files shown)


docs/platforms/apple/guides/ios/session-replay/index.mdx
```diff
@@ -103,5 +103,8 @@ Errors that happen on the page while a replay is running will be linked to the r
 Q: Does Session Replay work with SwiftUI?
 A: Yes. It works with both UIKit and SwiftUI.
 
+Q: Why are parts of my replay not masked?
+A: Text views, input views, images, video players and webviews are all masked by default. Images with bundled assets, aren't masked because the likelihood of these assets containing PII is low. If you encounter a view that should be masked by default, consider opening a [GitHub issue](https://github.com/getsentry/sentry-cocoa/issues).
```
coolguyzone, line 107:
```suggestion

A: Text views, input views, images, video players and webviews are all masked by default. Images with bundled assets aren't masked because the likelihood of these assets containing PII is low. If you encounter a view that should be masked by default, consider opening a [GitHub issue](https://github.com/getsentry/sentry-cocoa/issues).
```

docs/platforms/apple/guides/ios/session-replay/index.mdx
```diff
@@ -103,5 +103,8 @@ Errors that happen on the page while a replay is running will be linked to the r
...
+
 Q: What's the lowest version of iOS supported?
-A: Session Replay recording happens even on the lowest version supported by the Sentry SDK. (Today that's iOS 12.)
+A: Session Replay recording happens even on the lowest version supported by the Sentry SDK, which is aligend with appstore support.
```
coolguyzone, line 110:
```suggestion

A: Session Replay recording happens even on the lowest version supported by the Sentry SDK, which is aligend with appstore support.
```

coolguyzone (approved):
Thanks for updating! The Q and A were showing up on the same line so I added some line breaks, everything else looks good.

Issue #11572, opened 2024-10-15
Show links to `previous` and `next` pages


stephanie-anderson:
### Description

Add buttons for easier navigation through docs at the bottom of each page (for `previous` and `next` pages), similar to this one here:

![Image](https://github.com/user-attachments/assets/57455de9-821b-4055-b623-d7a5d5736abe)

Source: https://docusaurus.io/docs/static-assets

### Requirements

1. General:

- Despite the screenshot below, wording should be `Previous` and `Next`
- Buttons should render the page's title

2. Automatically generate `previous` and `next` buttons:
- The docs index page (`/`) does neither have `previous` nor `next` buttons
- The first child of a parent node should render the parent as its `previous` page
- The last child of a parent node should render the next sibling of its parent as `next` page
- If that parent does not have a next sibling, no `next` page link should be rendered

![Image](https://github.com/user-attachments/assets/1c526870-a62c-4e55-989b-fb7e61b963c9)

- If the page is a direct child of the index page (`/`), previous should point to the index page (`/`)

![Image](https://github.com/user-attachments/assets/f587b894-c32b-444c-9e0d-535fe5a23e10)

3. It should be possible to manually overwrite what `previous` and `next` should be for a given page. For example, for some JavaScript SDKs we have 3 different ways for users to install the SDK. We don't want to guide users from one install method to the others, but instead might guide them next to some configuration pages.

chargome:
@stephanie-anderson do we want this on dev docs?

PR #11639, opened 2024-10-23
feat(platform): Navigation buttons

chargome:
closes https://github.com/getsentry/sentry-docs/issues/11572

### DIFF
--- a/src/components/docPage/index.tsx
+++ b/src/components/docPage/index.tsx
@@ -85,7 +85,7 @@ export function DocPage({
                 <CodeContextProvider>{children}</CodeContextProvider>
               </div>
 
-              <div className="grid grid-cols-2 gap-4 not-prose mt-8">
+              <div className="grid grid-cols-2 gap-4 not-prose mt-16">
                 <div className="col-span-1">
                   {previousPage && <PaginationNav node={previousPage} title="Previous" />}
                 </div>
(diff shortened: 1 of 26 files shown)


src/components/docPage/index.tsx
```diff
@@ -78,6 +84,16 @@ export function DocPage({
               <div id="main">
                 <CodeContextProvider>{children}</CodeContextProvider>
               </div>
+
+              <div className="grid grid-cols-2 gap-4 not-prose mt-8">
```
stephanie-anderson, line 88: just a small nit - would use `mt-16` here to allow for more space between content and navigation

stephanie-anderson:
Haven't had the chance to check - I think this is not up to date with master. Have we checked if hidden pages are properly excluded from that navigation? I can think of the metrics page in develop docs as a good example: 

Screenshot 2024-10-23 at 17 53 19

chargome:
@stephanie-anderson this is up to date with master, metrics is not excluded – or maybe I don't understand the question 🤔

stephanie-anderson:
Oh, this page was never hidden in the first place 🙈 I was referring to whether we checked if the frontmatter config `sidebar_hidden` would be properly respected (as in, it would never be a previous or next page suggestion)

src/docTree.ts
```diff
@@ -146,6 +146,108 @@ export function nodeForPath(node: DocNode, path: string | string[]): DocNode | u
...
+
+  const index = siblings.indexOf(node);
+  if (index > 0) {
+    return siblings[index - 1];
+  }
+
+  return undefined;
+};
+
+const sortSiblingsByOrder = (a: DocNode, b: DocNode) =>
```
s1gr1d, line 242:
```suggestion
const sortBySidebarOrder = (a: DocNode, b: DocNode) =>
```

s1gr1d (commented):
Super cool addition to the docs!

a-hariti (approved):

stephanie-anderson (approved):

PR #11644, opened 2024-10-24
Updates to tracing walkthrough

whitep4nth3r:
Hi @lizokm 

I made a few edits to the tracing walkthrough for your consideration.

I also added some placeholders e.g. `_ADD IMAGE HERE OF SINGLE TRACE PAGE_` where I think it would be useful to add some images to illustrate the point, just in case people haven't signed up to Sentry yet and aren't able to view anything in the deep-linked pages.


lizokm:
@whitep4nth3r thank you so much for updating this. I will add the images as you suggested!

whitep4nth3r:
Just adding a comment here so it doesn't auto close.

lizokm:
@whitep4nth3r you can actually close this PR. I've added all your suggestions to the live page :) https://docs.sentry.io/product/tracing/

Issue #10517, opened 2024-06-26
Platform redirect page could be more specific about where you're heading next


cstavitsky:
### Problem Statement

Platform redirect page ([example for sampling](https://docs.sentry.io/platform-redirect/?next=/configuration/sampling/)) shows generic text at top:
> ## Platform Specific Content
> ...list of SDK-specific links

If I land on this page, it's not clear that the redirect page will lead me to docs about Sampling.

### Solution Brainstorm

Instead, I would like it to read something like this, so I know where I'm about to head after the redirect.

> # Platform Specific Content
> ### Sampling
> #### Learn how to configure the volume of error and transaction events sent to Sentry.
> ...list of SDK-specific links

realkosty:
@stephanie-anderson
> Unfortunately, there is no guarantee that all pages across platforms will have the same title and/or description:
IMO that's fine. We can pick the most common title, a concatenation of all titles, or even just the first one - would be way better than nothing. The titles should be fairly similar anyway, right?

I personally lean towards putting it in the big header instead. As a Sentry user not very familiar with our docs platform I would be confused by huge "Platform Specific Content" and assume I opened the wrong link or looking at the wrong tab. But your proposed solution would be definitely a big improvement too. Or what Chris suggested above.

realkosty:
@lizokm what do you think?

cstavitsky:
> As a Sentry user not very familiar with our docs platform I would be confused by huge "Platform Specific Content" and assume I opened the wrong link or looking at the wrong tab.

How about replacing `Platform Specific Content` with something that feels more actionable? For example `Continue To Docs For Your Platform:`

> I think what we could do is show a preview of title and description of all platforms that have that content. Unfortunately, there is no guarantee that all pages across platforms will have the same title and/or description

> IMO that's fine. We can pick the most common title, a concatenation of all titles, or even just the first one - would be way better than nothing. The titles should be fairly similar anyway, right?

Yes, either of these would help. Anything that will give the user an expectation of what exactly they're about to navigate to!

lizokm:
I think it would be best to put the information in the header/ and maybe elaborate a little more in the text just under the header. Then we could put the "The page you are looking for is customized...." text in a <Note></Note> so that it stands out :)

lizokm:
@chargome this is what we were talking about in our meeting on Monday. I'd love to make this happen.

PR #11646, opened 2024-10-24
feat(platform): Display requested title in platform redirect

chargome:
closes https://github.com/getsentry/sentry-docs/issues/10517

With this change we display the title and description of the requested page (if it exists).

## before 
Screenshot 2024-10-24 at 13 02 44

## after
Screenshot 2024-10-24 at 12 58 33

The risk here is that different platforms might have different titles and descriptions for the same page, and in this PR we just take the first one we find (which is usually .NET I think). I tested out some pages and it seemed to work fine every time.

### DIFF
--- a/app/platform-redirect/page.tsx
+++ b/app/platform-redirect/page.tsx
@@ -1,5 +1,6 @@
 import {redirect} from 'next/navigation';
 
+import {Alert} from 'sentry-docs/components/alert';
 import {DocPage} from 'sentry-docs/components/docPage';
 import {PlatformIcon} from 'sentry-docs/components/platformIcon';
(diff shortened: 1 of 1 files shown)


stephanie-anderson:
Oh, how fast! Thanks for implementing this. I initially thought we would add the page titles and descriptions in the list of platforms (instead of .NET etc), but this is also an option. Would let @lizokm decide on what she would prefer

chargome:
@stephanie-anderson yeah I just went with lizas most recent comment in the connected issue

stephanie-anderson (approved):

cstavitsky (approved):
Nice!!! Excited to see this 🎉

PR #11660, opened 2024-10-25
Suggested edits for tracing tutorial

whitep4nth3r:
All suggested edits have been made.

I have one last bit of feedback for the very end of the tutorial.

When I clicked into traces, there was one trace that had both icons, as described in the tutorial.

I got the error on the front end:

image

And in the log:

image

However, the _error_ was not present in the trace in my trace list.

image

Going to the issues screen, clicking into the generated issue, and clicking on the full trace ID, however, got me to the correct trace:

image

image

It's weird because I checked the sample rates and they were all set to 100%.

I just wonder if it's a better happy path if we direct users to the issue screen, and then ask them to click the full trace ID, so we can guarantee the error will be shown in the trace.


coolguyzone (approved):
Great suggestions! Thanks!

coolguyzone:
Thanks @whitep4nth3r , I agree with and merged in your suggestions. That's an interesting discrepancy you noticed in the trace view/vs issues view, I haven't run into it myself, but I will walk through it a couple of more times to see if I can reproduce.

PR #11662, opened 2024-10-25
develop: UF remove language from snippet

bruno-garcia:
image

Currently shows like this ^

We should go with `JavaScript` or if pseudo code maybe don't specify it at all (this PR attempts the latter)


aliu39:
Could we put "Pseudo-code" at the top where language used to be?

aliu39 (approved):

bruno-garcia:
thanks Andrew!

PR #11668, opened 2024-10-28
feat(api): troubleshooting tests for publishing [WIP]

isabellaenriquez:


develop-docs/backend/api/public.mdx
```diff
@@ -410,6 +410,77 @@ will be triggered to make your changes automatically go live:
    makes and merges a PR in `sentry-docs`, which kicks off a deploy via Vercel to
    https://docs.sentry.io/api/.
 
+## Troubleshooting API Docs Tests
```
sentaur-athena, line 413: suggestion: we also get a lot of questions from people that see the diff test fail and think it's a test failure. Would be good to add that here too and say it's supposed to fail to show you the difference between new and old schema.

sentaur-athena (approved):

develop-docs/backend/api/public.mdx
isabellaenriquez: good idea!

PR #11690, opened 2024-10-30
feat: convert mermaid code blocks to SVGs client side

### LINKED ISSUE #11561: Add MermaidJS plugin
### Problem Statement

Graphs and diagrams in the docs (be it application or develop docs) are right now created in another tool, and then the image is linked in our docs. 

This leads to outdated graphs everywhere. 

### Solution Brainstorm

It would be great if we could create/change [Mermaid](https://mermaid.js.org/) graphs directly in the .mdx files. 

So it is possible to add something like this directly in the docs (develop and other docs) and the graph is rendered on the page:

````markdown
```mermaid
graph TD;
    A-->B;
    A-->C;
    B-->D;
    C-->D;
```
````

The default styles of the graphs should also be updated, because right now mermaid graphs are not readable that good in dark mode. See:
- https://develop.sentry.dev/application/architecture/
- https://develop.sentry.dev/backend/cross-region-replication/

### PR DESCRIPTION
**Pull request by a-hariti**
Converts `mermaid` code blocks to svg using `mermaid.js` using a minimal diff

this library comes with the limitation of working client side only (the same way it's handled on Github Readmes for example),

some solutions for node.js come with a whole browser environment to take a screenshot (clunky solution)

on the flip side, client side conversion causes some flickering on the pages with mermaid diagrams,

but this seems to be the most pragmatic way to do this

**Note:** to preview this, go to the developer docs preview under `/application/architecture/`

TODO:
- [ ] port other mermaid diagrams source code to this PR

closes #11561

### DIFF
--- a/src/components/docPage/index.tsx
+++ b/src/components/docPage/index.tsx
@@ -1,5 +1,4 @@
 import {ReactNode} from 'react';
-import Script from 'next/script';
 
 import {getCurrentGuide, getCurrentPlatform, nodeForPath} from 'sentry-docs/docTree';
 import {serverContext} from 'sentry-docs/serverContext';
@@ -14,6 +13,7 @@ import {Breadcrumbs} from '../breadcrumbs';
 import {CodeContextProvider} from '../codeContext';
 import {GitHubCTA} from '../githubCTA';
 import {Header} from '../header';
+import Mermaid from '../mermaid';
 import {PaginationNav} from '../paginationNav';
 import {PlatformSdkDetail} from '../platformSdkDetail';
 import {Sidebar} from '../sidebar';
@@ -54,9 +54,6 @@ export function DocPage({
   const unversionedPath = getUnversionedPath(path, false);
 
   const leafNode = nodeForPath(rootNode, unversionedPath);
-
-  // a hack to show syntax highlighting on editors
-  const javascript = String.raw;
 
   return (
     <div className="tw-app">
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/components/docPage/index.tsx [hunk 1]
```diff
@@ -108,6 +112,34 @@ export function DocPage({
...
+      {/* can't use useEffect here (server component) */}
+      <Script
+        type="module"
+        id="mermaid-script"
+        dangerouslySetInnerHTML={{
+          __html: javascript`
+              const mermaidBlocks = document.querySelectorAll('.language-mermaid');
+              if (mermaidBlocks.length >= 0) {
+                const {default: mermaid} = await import(
+                  'https://cdn.jsdelivr.net/npm/mermaid@11/dist/mermaid.esm.min.mjs'
```
chargome, line 124: this breaks our docs CSP, you'll need to add this script url to our vercel.json

src/components/docPage/index.tsx
```diff
@@ -108,6 +112,34 @@ export function DocPage({
           )}
         </main>
       </section>
+      {/* can't use useEffect here (server component) */}
```
chargome, line 115: Could we move that into a separate client component? I think this would also cause hydration errors

src/components/docPage/index.tsx
```diff
@@ -54,6 +55,9 @@ export function DocPage({
 
   const leafNode = nodeForPath(rootNode, unversionedPath);
 
+  // a hack to show syntax highlighting on editors
+  const javascript = String.raw;
```
chargome, line 59: That's a bit confusing, could we just inline this with a comment where it is used?

chargome (changes requested):

src/components/docPage/index.tsx
a-hariti: we can't as we are already insde `return`, but a client component will obviate the need for this

src/components/docPage/index.tsx
a-hariti: better yet add it as a dependency 👍

src/components/mermaid.tsx
```diff
@@ -0,0 +1,29 @@
...
+        document.querySelectorAll<HTMLDivElement>('.language-mermaid');
+      if (mermaidBlocks.length === 0) {
+        return;
+      }
+      const {default: mermaid} = await import('mermaid');
+      mermaid.initialize({startOnLoad: false});
+      mermaidBlocks.forEach(block => {
+        // get rid of code highlighting
+        const code = block.textContent ?? '';
+        block.innerHTML = code;
```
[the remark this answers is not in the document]
a-hariti: @chargome 👀

chargome (approved):
Thanks for adding this! @antonpirker will be happy 😄

antonpirker:
This is sooo cool @a-hariti !

PR #11692, opened 2024-10-31
docs(js): Add troubleshooting section on terser

chargome:
Adds a troubleshooting section for build-errors when using a minifier like terser.

documents https://github.com/getsentry/sentry-javascript/issues/14091


Lms24 (approved):
Nice!

PR #11698, opened 2024-10-31
chore(platform): Bump search insights + catch analytics call

chargome:
Currently on every click on search result we make a call to algolia search insights.
If this fails for some reason we don't want the error to bubble up or log an error to the console, so we just catch it and report it to Sentry.


chargome:
closing in favour of https://github.com/getsentry/sentry-docs/pull/11700

Issue #11699, opened 2024-10-31
Index anchor links in algolia


chargome:
### Problem Statement

It seems we currently do not index anchor links in Algolia (e.g. `/platforms/javascript/configuration/options/#ignore-errors`).
This is generally not a big issue since the content will still move the user to the desired page where the content lives on.

This does become an issue (or leads to confusion) though when a user searches for content and is already on the correct page.

### Solution Brainstorm

- [X] As a first step we could simply close the search result box when the user clicks on a result for a page that is the current one.

- [ ] As a second step we could think of indexing anchor links and adding them to the search results, so we can make use of `.scrollIntoView()` and provide a better UX.

PR #11700, opened 2024-10-31
fix(platform): Automatically scroll to search hits on same page

chargome:
This pr checks if a search result is on the current page and if so, automatically scrolls the result into the view.

related to https://github.com/getsentry/sentry-docs/issues/11699


a-hariti (approved):
Nice!

PR #11701, opened 2024-10-31
Develop sdks

stephanie-anderson:


develop-docs/sdk/philosophy.mdx
```diff
@@ -1,9 +1,9 @@
 ---
 title: Philosophy
+description: This document sets some general guidance for how to approach SDK development at Sentry. It should aid both internal and external developers to understand what motivations go into the design of the SDKs and why we're deciding in certain ways.
```
chargome, line 3:
```suggestion
description: This document provides general guidance for SDK development at Sentry. It aims to help both internal and external developers understand the motivations behind SDK design decisions and our decision-making process.
```

I know this is not about content but the description didn't read well

src/components/docPage/type.scss
```diff
@@ -68,7 +72,7 @@
   hgroup h2 {
     font-weight: 300;
     font-size: 1.3rem;
-    color: var(--foreground-secondary);
+    color: var(--desatPurple2);
```
chargome, line 75: was there a special reason for this?

chargome (approved):
Looks better and makes sense 👍

develop-docs/sdk/philosophy.mdx
stephanie-anderson: Thanks for the suggestion! I'll add it ^^

src/components/docPage/type.scss
stephanie-anderson: Yes, it was never done by our design team, and I think the grey that we had before simply was off. Personally, I think it looks much cleaner now

## before
Screenshot 2024-10-31 at 16 59 02
Screenshot 2024-10-31 at 16 58 58


## after

Screenshot 2024-10-31 at 16 58 44
Screenshot 2024-10-31 at 16 58 39


what do you think?

chargome: I think the color in light mode is a bit strong but that's fine 👍

PR #11715, opened 2024-11-02
API examples syntax highlighting

### LINKED ISSUE #11093: Unify API pages code snippets
### Problem Statement

Currently api pages use their own syntax highlighting using prismjs. 

### Solution Brainstorm

We want to streamline the appearance of code blocks across docs and use the same styling on api pages.

a-hariti:
The intent was to unify the syntax highlighting logic for API examples and other docs,
but API examples have specific requirements that prevent us from reusing existing logic

code blocks are usually handled though `CodeTabs`, which adds tabs and manages the corresponding code blocks, but we need more tabs (api responses for example)

The curl bash snippet was possible to handle using a simple `CodeBlock`

I tried to use the same syntax highlighting libraries (using refractor) server side, and gave the API examples code blocks the same look and behavior as elsewhere

verify [here](https://sentry-docs-git-api-syntax-highlighting.sentry.dev/api/alerts/retrieve-a-metric-alert-rule-for-an-organization/)

closes #11093

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,6 @@
     "parse-numeric-range": "^1.3.0",
     "platformicons": "^6.0.3",
     "prism-sentry": "^1.0.2",
-    "prismjs": "^1.27.0",
     "query-string": "^6.13.1",
     "react": "^18",
     "react-dom": "^18",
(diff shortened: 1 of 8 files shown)


chargome (approved):
Thanks! Can you leave a brief explanation why we the global syntax highlighting does not work?

PR #11732, opened 2024-11-05
feat(apple): Add enableTracingWhenCrashing

philipphofmann:
## DESCRIBE YOUR PR

Add docs for the new option enableTracingWhenCrashing.

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

### DIFF
--- a/docs/platforms/apple/common/configuration/options.mdx
+++ b/docs/platforms/apple/common/configuration/options.mdx
@@ -272,11 +272,12 @@ If <PlatformIdentifier name="trace-propagation-targets" /> is not provided, trac
 
 </ConfigKey>
 
-<ConfigKey name="enable-tracing-when-crashing">
+<ConfigKey name="enable-persisting-traces-when-crashing">
+
 
 <Note>
 
-This feature is experimental, may have bugs and is new in [sentry-cocoa version 8.40.0](https://github.com/getsentry/sentry-cocoa/blob/main/CHANGELOG.md#8400).
+This feature is experimental, may have bugs and is new in [sentry-cocoa version 8.41.0](https://github.com/getsentry/sentry-cocoa/blob/main/CHANGELOG.md#8410).
 
 </Note>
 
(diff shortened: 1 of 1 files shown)


docs/platforms/apple/common/configuration/options.mdx
```diff
@@ -271,3 +271,15 @@ String entries do not have to be full matches, meaning the URL of a request is m
 If <PlatformIdentifier name="trace-propagation-targets" /> is not provided, trace data is attached to every outgoing request from the instrumented client.
 
 </ConfigKey>
+
+<ConfigKey name="enable-tracing-when-crashes">
```
cleptric, line 275: This is a very odd name in light of us removing `enableTracing` from the SDKs.

philipphofmann: Thanks for pointing that out. Naming is hard 😅. It's basically finishing transactions bounded to the scope when an app crashes. The code is here https://github.com/getsentry/sentry-cocoa/pull/4504. 

I also considered `enableCrashTracing` and `enableTracingCrashes` but that's closer to tracing actual crashes, which the SDK already does. And I have a typo there. It should be `enableTracingWhenCrashing`. Is it still odd?

romtsn: @philipphofmann do we need a flag for this actually? That's the default behavior on android. Does it introduce any unwanted side-effects?

philipphofmann: > Does it introduce any unwanted side-effects?

Maybe yes. Other signal handlers might be effected, and the app can get stuck after writing the crash report to disk. So my plan is to make this the default behaviour, but for now it should be an experimental feature.

> Ok, so this will only work if tracing is actually enabled, and will then persist the transaction to disc?

Yes, indeed.

brustolin (approved):

docs/platforms/apple/common/configuration/options.mdx
cleptric: Ok, but then it does not enable tracing. So I would suggest we call it something less confusing, like `persistTracesOnDiscIfCrashed`.

philipphofmann: We prefix everything with enable in the Cocoa SDK, so I would like to keep that pattern. What about `enablePersistingTracesWhenCrashing`, @cleptric?

cleptric: Also ok

coolguyzone (approved):
LGTM! Thanks for the addition!

PR #11743, opened 2024-11-06
feat: Rename docsbot button to "Ask AI"

chargome:
We'd like to measure if a label containing **AI** has any influence on how often the button is clicked by our users.


stephanie-anderson (approved):
Can we also add an icon to that?

chargome:
Sure, I tried a magic wand icon but that didn't look good I think, any preferences?

chargome:
Screenshot 2024-11-06 at 12 11 15

PR #11754, opened 2024-11-07
fix mermaid charts in dark mode

### LINKED ISSUE #11733: Toggling light/dark mode breaks mermaid diagrams
### Steps to Reproduce

- Go to https://develop.sentry.dev/application/architecture/
- Toggle light/dark mode
- The diagram is now gone and some css is printed.

### Expected Result

The mermaid diagram should render with right theme set. 

### Actual Result

The mermaid component breaks and just prints some CSS.

### Additional Info

Thats the code where we set the mermaid theme:

https://github.com/getsentry/sentry-docs/blob/master/src/components/mermaid.tsx#L29

a-hariti:
closes  #11733

### DIFF
--- a/src/components/mermaid.tsx
+++ b/src/components/mermaid.tsx
@@ -30,9 +30,8 @@ export default function Mermaid() {
       if (mermaidBlocks.length === 0) {
         return;
       }
-      // @ts-ignore
       const {default: mermaid} = await import('mermaid');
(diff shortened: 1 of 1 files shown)


antonpirker (approved):
Cant say much about the code (because my JS/TS is very rusty) but I have tested it and it works!

PR #11761, opened 2024-11-07
feat: Banner pathname regex

### LINKED ISSUE #11532: Promo Banners
### Problem Statement

Since the docs are great for boosting our SEO traffic, I was wondering if we could extend the promo banner beyond the homepage.

Currently, it's managed through `src/components/banner` with no real option on which page we would like to target, and based on the codebase, it's only published on the docs homepage (though please let me know if I'm missing something). 

### Solution Brainstorm

We would like the following functionality:

- Have banners appear on any page in docs and have that selectively controlled through `src/components/banner/index.tsx`
- Have the ability to have different banners appear on different pages

As an example of general functionality, it can be found on `static-sites` (`packages/shared/components/Banner`).

a-hariti:
Adds support for multiple banners along with Regex pathname matching and expiration dates 

closes  #11532

**TODO**
- [x] move live examples for preview purposes to comments documenting the feature

### DIFF
--- a/docs/contributing/pages/banners.mdx
+++ b/docs/contributing/pages/banners.mdx
@@ -1,0 +1,72 @@
+---
+title: Banners
+noindex: true
+sidebar_order: 80
+---
+
+You can add arbitrary banners to the top of a page by adding adding an entry to the `BANNERS` array on
+the `banner/index.tsx` file. The `BANNERS` array is an array of objects with the following properties:
+
+```typescript {filename:banner/index.tsx}
+type BannerType = {
+  /** This is an array of strings or RegExps to feed into new RegExp() */
+  appearsOn: (string | RegExp)[];
+  /** The label for the call to action button */
+  linkText: string;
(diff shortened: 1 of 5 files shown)


chargome (changes requested):
Nice approach! Just some UI issues I noticed when looking at the banner example in /platforms/javascript/:

- padding seems too small
- font size is slightly larger than on the banner on welcome page
- dark mode should not change the font color if the background stays the same
- banner is not responding well on mobile

stephanie-anderson:
I was about to write that we need to check if this approach (regex) is suitable for the target audience - but it seems this is what they already do in static-pages :) Please make sure to document it in the contribution guide (would not rely on the comments in code)

a-hariti:
I think the home page banner is the one that needs to be obvious, other will probably be set by devs

and `static-sites` is mostly managed by marketing people :)

chargome (approved):

PR #11769, opened 2024-11-08
fix Ureal engine DSN placeholder

### LINKED ISSUE #11556: DSN template regression: shows `__PUBLIC_DSN__`
### Steps to Reproduce

On both Unity and Unreal docs we have `__PUBLIC_DSN__` which are not getting replaced by a demo URL or a user DSN if logged in.
In the case of Unity it has prefix/suffix `___`, as it did Unreal, which I tried fixing by removing the extra `_` but didn't help.

https://docs.sentry.io/platforms/unreal/#configure

https://docs.sentry.io/platforms/unity/#configure

### Expected Result

A URL

### Actual Result

a template

### Additional Info

_No response_

a-hariti:
fixes #11556

### DIFF
--- a/docs/platforms/unreal/index.mdx
+++ b/docs/platforms/unreal/index.mdx
@@ -78,8 +78,8 @@ PublicDependencyModuleNames.AddRange(new string[] { ..., "Sentry" });
 The minimum configuration required is the [DSN](/product/sentry-basics/dsn-explainer/) of your project:
 
 
-```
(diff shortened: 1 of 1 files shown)


stephanie-anderson (approved):

bruno-garcia:
image

Doesn't look fixed in the vercel preview: https://sentry-docs-git-fix-ureal-dsn-placeholder.sentry.dev/platforms/unreal/

PR #11791, opened 2024-11-12
feat(dynamic-sampling): organization settings

constantinius:
Add section under organization settings to explain the benefits of dynamic sampling and its relevant settings

Closes https://github.com/getsentry/projects/issues/214 and https://github.com/getsentry/projects/issues/213


lizokm (approved):
Thanks so much for starting on this! I know it's still in draft. Let me know when you'd like me to take another look. Also, take a look at our current DS docs to see if there's anything we can copy over. (https://docs.sentry.io/product/performance/retention-priorities/)

jan-auer (commented):
Some ideas for information that could be added:
- In the introduction, explain what "dynamic sampling" truly is. As in, what about it is dynamic and what it tries to achieve. 
  - How does DS fit with sampling in the SDKs and when do I use which?
  - What is the default (100% = no sampling)
  - How does DS re-compute rates? Note that here we shouldn't give any strict guarantees (e.g. "every 10 minutes") but mention that there's a short sliding window so that DS can adjust to seasonality like on weekends and over night traffic patterns.
- AM2 and AM3 have different capabilities. The sample rates are chosen automatically on AM2, while only on AM3 do you have the new settings screen and control over rates. We should make this clear. @lizokm, how do we usually talk about the plan generations?
- I'd structure this document into multiple larger blocks (not necessarily sections): 
  - Organization-wide settings, including the sampling mode, project preview, sample rates
  - Project-wide settings, which for now are just the priorities
  - Connection to other product surface, esp subscription settings and stats
- We can expand the description of the modes, particularly
  - How does DS work in each of the modes (i.e. what's the same and what's different)
  - How do I think about configuring each of the modes?
  - Especially in manual mode, describe the org-level assist. That's where we could use an arcade. 
  - What happens when I change into this mode?
- On project preview, what do the estimated rates mean, what's the difference between the two time frames, and how do they relate to what DS actually does? 
- Retention priorities are now called "[Dynamic] Sampling Priorities", so let's rename them here, too.

docs/organization/dynamic-sampling/index.mdx
```diff
@@ -1,18 +1,54 @@
...
+description: "Learn how to make your billing quota go farther with Sentry's Dynamic Sampling."
 ---
 
+## Overview
+
+High-volume projects that generate many events can quickly consume a large portion of your billing quota, leaving less room for other projects and decreasing visibility across them.
+
+With high-volume projects, you'll often see many similar events, which can be redundant and less helpful for debugging. But with Dynamic Sampling, you can maximize your quota by adjusting sample rates based on each project’s event volume. The sample rate is a percentage from 0% (no events passed) to 100% (all events passed), that can be tailored to each project’s needs.
+
+When you use Dynamic Sampling, projects in your organization are sampled based on their recent event volumes. High-volume projects are sampled at lower rates while lower-volume projects are sampled at higher rates or not sampled at all, based on your recent event data. This makes it possible for each project to retain a meaningful set of key events, improving visibility into all projects regardless of their size.
```
jan-auer, line 13: - For someone not too familiar with the matter, does "event volume" make sense in this context? For me knowing the internals this sounds wrong (we don't classify spans as events) but maybe for a user this is fine. 
- "not sampled" means that everything is dropped. I think you mean to say "sampled at higher rates up to 100%". Also, in this sentence, we don't sample based on event data, but based on event volume.

docs/organization/dynamic-sampling/index.mdx
```diff
@@ -1,18 +1,54 @@
...
+title: Dynamic Sampling
+sidebar_order: 30
+description: "Learn how to make your billing quota go farther with Sentry's Dynamic Sampling."
 ---
 
+## Overview
+
+High-volume projects that generate many events can quickly consume a large portion of your billing quota, leaving less room for other projects and decreasing visibility across them.
+
+With high-volume projects, you'll often see many similar events, which can be redundant and less helpful for debugging. But with Dynamic Sampling, you can maximize your quota by adjusting sample rates based on each project’s event volume. The sample rate is a percentage from 0% (no events passed) to 100% (all events passed), that can be tailored to each project’s needs.
```
jan-auer, line 11: This might be misleading. Without going into the details, there's always going to be a quota for how much is sent to Sentry by the SDKs. Dynamic Sampling has no effect on that, so you cannot just "maximize your quota". Sent spans is always going to be the smaller price component, so it makes sense to move more of the sampling to Sentry. The benefit of that is that Dynamic Sampling is smarter and adapts to changes, so you generally will get higher quality samples than with the random sampling in SDKs. We can further call out some examples from the priorities, like automatically boosting latest releases or boosting small endpoints.

docs/organization/dynamic-sampling/index.mdx
```diff
@@ -1,18 +1,54 @@
...
 
-It's important to note that even when we begin to only store the most valuable data based on retention priorities, the [performance metrics](/product/performance/metrics/) you see are still based on **all** the events you send to Sentry. This ensures that you get a full and complete view of your application's health.
+## Differences to Sampling in the Sentry SDKs
+
+Dynamic Sampling is different from the sampling that you can configure in the Sentry SDKs. The SDK sampling rate is applied before events are sent to Sentry, while Dynamic Sampling is applied after events are received by Sentry. This means that Dynamic Sampling can be used to adjust the sample rate for events that have already been sent to Sentry and only take into account the volume of events that have been received by Sentry.
+
+Also in contrast to the Dynamic Sampling, the SDK sample rate is never dynamically adjusted.
```
jan-auer, line 39: Strictly speaking, "never" isn't true :) `traces_sampler` allows customers to write their own code for dynamically adjusting sample rates. I think the true differences can rather be summarized as:
- Sampling in SDKs affects how much is captured and sent to Sentry. For high throughput scenarios, this may be necessary to reduce instrumentation overheads.
- To change sample rates in SDKs, one has to update the `traces_sample_rate` or `traces_sampler` callback, which usually requires a new deployment or release.
- With dynamic sampling, one can change sample rates without touching the application or SDK. This affects the amount of data _stored_ at Sentry, however, it will not have an effect on how much is sent from the SDKs.
- Dynamic sampling continuously monitors incoming traffic patterns and adjusts fine-grained sample rates within the projects to prioritize more important data. See the sampling priorities for examples.

docs/organization/dynamic-sampling/index.mdx
```diff
@@ -1,18 +1,54 @@
...
+
+## Dynamic Sampling Priorities
+
+<Note>
+
+Dynamic Sampling priorities are only available on the latest version of our plans, starting from 1M reserved transactions. Anyone in your org with owner or billing-level permissions can update your plan by going to Settings > Subscription, and clicking the "Manage Subscription" button.
```
jan-auer, line 45: This will need to be updated. DS priorities are available on AM2 _and_ AM3 (right now this is still correct, as AM2 is latest).

docs/organization/dynamic-sampling/index.mdx
```diff
@@ -1,18 +1,54 @@
...
+    
+    * **`manual`**: If you want to set the exact sample rate for each project in your organization yourself, choose manual mode. Note, that in this mode, there is no automatic recalibration.
+
+
+[]: # (TODO: Rename AM3 with the actual user-facing name.)
+
 <Note>
 
-Retention priorities are only available on the latest version of our plans, starting from 1M reserved transactions. Anyone in your org with owner or billing-level permissions can update your plan by going to Settings > Subscription, and clicking the "Manage Subscription" button.
+Please note that these Dynamic Sampling modes are only available for customers on the AM3 platform.
```
jan-auer, line 30: Should we call out that on AM2 there's also DS, the sample rate is chosen by the system based on transaction volume (1M+), and you still have controls for the sampling priorities?

constantinius:
Hey @lizokm!

Together with @jan-auer we did work out more details of Dynamic Sampling. Also we moved retention policies in the same document. Could you give this another proofread please?

docs/organization/dynamic-sampling/index.mdx
```diff
@@ -0,0 +1,219 @@
...
+
+Switching back to automatic mode from manual mode will start the recalibration process immediatly and resets the sample rates of each project in the organization to the calculated values. There is no way to restore the manually configured project sample rates after the switch.
+
+### Manual mode
+
+In the "manual" mode, you can configure a fixed sample rate for each project in your organization. This value will specify how much of the incoming traffic should be stored for the respective project. The sample rate for each project can be set between 0% and 100% and it defaults to 100% and are not adjusted by Dynamic Sampling.
+
+When the sample rates of each project is edited, an estimated overall organization sample rate is calculated as a weighted average across all projects. This value is always an estimation and changes according to volume within projects.
+
+If you want to change the sample rates for all projects at once, you an edit the overall organition sample rate. This will in turn adjust the sample rates of all projects to achieve the estimate rate. The maximum value is 100% and other projects will be adjusted accordingly to achieve the estimate rate.
```
jan-auer, line 79:
```suggestion
If you want to change the sample rates for all projects at once, you an edit the overall organization sample rate. This will proportionally adjust the sample rates of all projects to achieve the estimate rate. The maximum value is 100% and other projects will be adjusted accordingly to achieve the estimate rate. You can make final changes to the updated sample rates before saving all projects at once.
```

docs/organization/dynamic-sampling/index.mdx
```diff
@@ -0,0 +1,219 @@
...
+
+In order to apply the changes, you need to click the "Apply changes" button, otherwise, the changes will not be saved.
+
+As with automatic mode [Dynamic Sampling Priorities](#dynamic-sampling-priorities) apply within the project, which will boost certain spans and deprioritize others.
```
jan-auer, line 83:
```suggestion
Like in automatic mode, [Dynamic Sampling Priorities](#dynamic-sampling-priorities) apply within each project, which boosts certain spans and deprioritizes others.
```

jan-auer (commented):
Preliminary review up to "Differences to Sampling in the Sentry SDKs".

lizokm:
I'll take another look at this first thing Monday morning.

lizokm (approved):
I've made a bunch of edits. Please take a look when you get a chance and let me know if you want to chat about anything :) Thanks so much for putting together this massive update!

docs/organization/dynamic-sampling/index.mdx
```diff
@@ -0,0 +1,223 @@
...
+
+## Configuration
+
+In this section you'll learn how to use Dynamic Sampling in your organization.
+
+![The dynamic sampling mode and target sample rate setting.](./img/mode-and-target-sample-rate.png)
+
+### Modes
+
+If you prefer not to set sample rates manually, Sentry can do it for you. You just have to set a global sample rate for your organization and Sentry will then automatically balance the sample rates across projects, giving priority to factors like low-volume endpoints and new releases. If you want more fine-grained control, you can customize data retention by setting individual sample rates for specific projects.
```
matejminar, line 48: Here I miss an introductory sentence, something like "we offer two modes: simple mode and advanced mode".

matejminar (approved):
Do we also need to update the arcade guide at the bottom of this page?
CleanShot 2025-01-28 at 14 45 21

docs/organization/dynamic-sampling/index.mdx
```diff
@@ -0,0 +1,219 @@
...
+![The dynamic sampling mode and target sample rate setting.](./img/project-sample-rate-preview.png)
+
+Because project rates continuously change, we recommend that you choose the 24-hour preview to get a more accurate view of your current sample rates for real-time analysis. We recommend the 30-day preview for seeing average sample rates in your billing cycle for long-term planning.
+
+<Alert>
+  If you deactivate advanced mode, the recalibration process will
+  start immediatly and the per-project sample rates will be reset to the
+  calculated values. There's no way to restore the manually configured project
+  sample rates once you've deactivated advanced mode.
+</Alert>
```
jan-auer, line 71: I would move this into the "Advanced Mode" section below a paragraph that describes that you can deactivate advanced mode and go back to configuring the organization-wide target rate.

PR #11804, opened 2024-11-12
docs: fix queue custom instrumentation code

adrienbrault:
Hey, when following the [doc](https://docs.sentry.io/platforms/python/tracing/instrumentation/custom-instrumentation/queues-module/) I encountered the following error:
```
Span.__init__() got an unexpected keyword argument 'name'
```

It works with `description` instead.


adrienbrault:
@getsantry don't close :P

sfanahata:
@adrienbrault - I recently built a test app while working on updating this documentation, and I used the `name` parameter without a problem. In our documentation, we note: 

`Changed in 2.15.0
The parameter name in start_span() used to be called description. In version 2.15.0 description was deprecated and from 2.15.0 on, only name should be used. description will be removed in 3.0.0.`

Is it possible you're running an older version of the SDK?

PR #11819, opened 2024-11-14
feat(apple): Add onboarding source context Option

### LINKED ISSUE #11810: Highlight better the effects of the "include sources" option
### SDK

Apple SDK

### Description

Looking at [this page](https://docs.sentry.io/platforms/apple/guides/ios/dsym/), the "include sources" option is added in the code snippets. There is a description for what this option does on the text above, but a customer believes it does not sufficiently describe its effects, namely, that it will show readable source code in the stack traces.

The accidental inclusion of this option can lead to legal issues on the customer's side.

### Suggested Solution

Describe more clearly, with a warning, that this this option will show readable source code in the event's stack traces. Additionally, include a comment that indicates this in the code snippets so people are aware when copying the code snippets without reading the text around them.

Alternatively, make the "include sources" option opt-in instead of opt-out.

brustolin:
Added source-coxtext as onboarding component option in order to manipulate the code snippet related to upload debug symbols.

closes #11810

### DIFF
--- a/platform-includes/debug-symbols-apple/_default.mdx
+++ b/platform-includes/debug-symbols-apple/_default.mdx
@@ -1,5 +1,11 @@
 Sentry requires dSYMs (debug information files) to symbolicate your stack traces. The symbolication process unscrambles the stack traces to reveal the function, file names, and line numbers of the crash.
 
+You can also upload your code for source context. This feature allows Sentry to display snippets of your code next to the event stack traces.
+
+<OnboardingOptionButtons
+  options={[{ id:"source-context", checked: false }]}
+/> 
+
 Every solution requires a **Sentry Auth Token**. You can create tokens on the [Organization Auth Tokens](https://sentry.io/orgredirect/organizations/:orgslug/settings/auth-tokens/) settings page.
 
 To view uploaded dSYMs in your project, select an existing project from the "Projects" page, then go to **Settings > Debug Files**. You can upload dSYMs using:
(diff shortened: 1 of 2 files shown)


platform-includes/debug-symbols-apple/_default.mdx
```diff
@@ -1,5 +1,11 @@
 Sentry requires dSYMs (debug information files) to symbolicate your stack traces. The symbolication process unscrambles the stack traces to reveal the function, file names, and line numbers of the crash.
 
+You can also upload your code for source context. This feature allows Sentry to display snippets of your code next to the event stack traces.
+
+<OnboardingOptionButtons
+  options={[{ id:"source-context", checked: false }]}
+/>
```
philipphofmann, line 7: `h`: I think this should be directly above the code snippets. The experience is quite confusing to me now as it's I have no clue what the checkbox is doing, cause I don't see the code snippets.

CleanShot 2024-11-14 at 16 10 11@2x

philipphofmann (commented):
Thanks for doing this, but the current approach is quite confusing to me right now.

platform-includes/debug-symbols-apple/_default.mdx
philipphofmann: @brustolin, I think it would be better to somehow make the button non sticky. When I navigate down to the SentryFastlane plugin, I now see the button but it misses explanation.
CleanShot 2024-11-18 at 15 56 35@2x

brustolin: But It affects all the code snippets in the page. And since the entire page is about uploading debug information, I think this makes sense. 

Its like this in all other pages that this component is used.

philipphofmann (approved):
LGTM

PR #11827, opened 2024-11-14
feat(flags): add browser JS integration docs for Launchdarkly

aliu39:
Documents the integration added in https://github.com/getsentry/sentry-javascript/pull/14207. Equivalent for server-side Python is https://github.com/getsentry/sentry-docs/pull/11675

### DIFF
--- a/docs/platforms/javascript/common/configuration/integrations/launchdarkly.mdx
+++ b/docs/platforms/javascript/common/configuration/integrations/launchdarkly.mdx
@@ -1,6 +1,6 @@
 ---
 title: LaunchDarkly
-description: "Attaches recent LD feature flag evaluations to error event context."
+description: "Attaches recent LaunchDarkly feature flag evaluations to error event context."
 notSupported:
   - javascript.aws-lambda
   - javascript.azure-functions
@@ -29,7 +29,7 @@ This integration only works inside a browser environment.
 
 _Import name: `Sentry.launchDarklyIntegration` and `Sentry.buildLaunchDarklyFlagUsedHandler`_
 
-The [LaunchDarkly](https://launchdarkly.com/) integration tracks feature flag evaluations produced by the LaunchDarkly SDK. These evaluations are held in memory and, in the event an error occurs, sent to Sentry for review and analysis.
+The [LaunchDarkly](https://launchdarkly.com/) integration tracks feature flag evaluations produced by the LaunchDarkly SDK. These evaluations are held in memory and, in the event an error occurs, sent to Sentry for review and analysis. **At the moment, we only support boolean flag evaluations.**
 
 ```JavaScript
 import * as Sentry from '@sentry/browser';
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/common/configuration/integrations/launchdarkly.mdx
```diff
@@ -0,0 +1,54 @@
...
+  - javascript.deno
+  - javascript.electron
+  - javascript.express
+  - javascript.fastify
+  - javascript.gcp-functions
+  - javascript.hapi
+  - javascript.koa
+  - javascript.nestjs
+  - javascript.nodejs
+  - javascript.wasm
```
aliu39, line 21: These are the non-frontend frameworks

michellewzhang (approved):

docs/platforms/javascript/common/configuration/integrations/launchdarkly.mdx
bruno-garcia: feature flags can be used on BE too. Does LD for example have a packge for browser and another for node?

aliu39: LD has server-side node, yes, but the implementation + installation will look different, so we focus on client-side for now. See https://github.com/getsentry/sentry-javascript/pull/14207#issuecomment-2471504674

aliu39: Tldr; the hook used to listen for evaluations doesn't exist for node. If it does, it works differently. This is also the reason we can't support specific frontend frameworks, only the generic launchdarkly-js-client-sdk

docs/platforms/javascript/common/configuration/integrations/launchdarkly.mdx
```diff
@@ -0,0 +1,54 @@
+---
+title: LaunchDarkly
+description: "Attaches recent LD feature flag evaluations to error event context."
```
billyvg, line 3: Probably better to avoid acronyms in docs

```suggestion
description: "Attaches recent LaunchDarkly feature flag evaluations to error event context."
```

docs/platforms/javascript/common/configuration/integrations/launchdarkly.mdx
```diff
@@ -0,0 +1,54 @@
...
+import * as LaunchDarkly from 'launchdarkly-js-client-sdk';
+
+Sentry.init({integrations: [Sentry.launchDarklyIntegration()]});
+
+const ldClient = LaunchDarkly.initialize(
+    'my-client-ID',
+    {kind: 'user', key: 'my-user-context-key'},
+    {inspectors: [Sentry.buildLaunchDarklyFlagUsedHandler()]}
+);
+const flagVal = ldClient.variation('my-flag', false); // evaluates a flag
```
billyvg, line 45: Are LD flags always booleans? (just wondering if we should clarify we only support boolean values)

aliu39: They're not. And yeah I think we should! cc @cmanallen maybe we can clarify this for Python too

cmanallen: 👍

Issue #11870, opened 2024-11-19
Redirects failing on prod


chargome:
### URL

Seems like none of our redirects are being handled currently when the app is deployed on vercel.

chargome:
This was an internal issue on vercel, redirects started working again.

PR #11873, opened 2024-11-19
fix(404): Update internal references to sdk/data-handling

chargome:
surfaced because of https://github.com/getsentry/sentry-docs/issues/11870


coolguyzone (approved):
Verified all the links, good to go!

PR #11839, opened 2024-11-15
feat(flags): Add browser JS integration docs for Launchdarkly

cmanallen:
Re-adds LaunchDarkly integration documentation.

Reverts getsentry/sentry-docs#11838

PR #11879, opened 2024-11-19
docs(flags): add index flag docs for python

michellewzhang:
create a index page for feature flags under the python platform (currently we only have specific instructions under each of the integrations, e.g. https://docs.sentry.io/platforms/python/integrations/openfeature/)

url will be: https://docs.sentry.io/platforms/python/feature-flags/

[figjam ref](https://www.figma.com/board/ZFPJSbpTYyjMzkp0qqWAgp/Feature-Flag-Docs-Map?node-id=1-596&node-type=shape_with_text&t=gMvrvhNpF7u073HY-0) mapping out the docs

localhost_3000_platforms_python_feature-flags_ (1)

### DIFF
--- a/docs/platforms/python/feature-flags/index.mdx
+++ b/docs/platforms/python/feature-flags/index.mdx
@@ -22,6 +22,6 @@ Please read the note below to ensure that you also complete one additional step.
 
 <Alert level="warning" title="Note">
 
-In order to take full advantage of the feature flag capabilities Sentry offers, there is an additional setup step needed, which is setting up your integration-specific webhook. This is needed to enable **feature flag change tracking**, so that your integration may communicate feature flag changes to Sentry. Learn how to set this up by [reading the docs](/product/explore/feature-flags/). 
+In order to take full advantage of the feature flag capabilities Sentry offers, there is an additional setup step needed, which is setting up your integration-specific webhook. This is needed to enable **feature flag change tracking**, so that your integration may communicate feature flag changes to Sentry. Learn how to set this up by [reading the docs](/product/explore/feature-flags/#set-up-your-integration-specific-webhook). 
 
 </Alert>
--- a/docs/platforms/python/feature-flags/index.mdx
+++ b/docs/platforms/python/feature-flags/index.mdx
@@ -1,0 +1,11 @@
+---
+title: Set Up Feature Flags
+sidebar_order: 5200
+description: "Learn how to enable Feature Flags in your app if it is not already set up."
+---
+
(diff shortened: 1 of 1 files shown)


docs/platforms/python/feature-flags/index.mdx
```diff
@@ -0,0 +1,11 @@
+---
+title: Set Up Feature Flags
+sidebar_order: 5200
+description: "Learn how to enable Feature Flags in your app if it is not already set up."
```
ryan953, line 4: I wanna point out early and often that we should be clear what kind of feature-flag debugging experience this is talking about. "enable Feature Flags" could easily also be interpreted as "i want to setup open feature for the first time" which is not quite the same.

This is the SDK setup... so we should say something specific like 'feature flag evaluation tracking' or something like that, and if there's a similar page 'feature flag change/definition setup (by provider)' page we could link to that from here.

ryan953:
something to coordinate merging along with:
https://github.com/getsentry/sentry-docs/pull/11839
https://github.com/getsentry/sentry-docs/pull/11837
https://github.com/getsentry/sentry-docs/pull/11749

Maybe all these need an 'EA' or 'beta' banner along the top then we could start merging now?

docs/platforms/python/feature-flags/index.mdx
michellewzhang: i don't think we have a PR yet that instructs how to do the feature flag change/definition setup, we should probably make a ticket for that

michellewzhang:
> something to coordinate merging along with: #11839 #11837 #11749
> 
> Maybe all these need an 'EA' or 'beta' banner along the top then we could start merging now?

SCR-20241120-ktoz
added a beta label

coolguyzone (approved):
Thanks for adding! LGTM 🫡

ryan953 (approved):

michellewzhang:
TODO: we should link to the change tracking docs from this index page too.

jas-kas (approved):
Same feedback as what's on https://github.com/getsentry/sentry-docs/pull/11971 :)

cmanallen:
@michellewzhang Simplifying the language here to focus on short exposition and explanation of enabling the feature set.  I'm doing the same for the Javascript PR and more shift more of the exposition burden onto the consolidated product docs page.

ryan953:
Looks like this got pulled into the big docs PR here: https://github.com/getsentry/sentry-docs/pull/12005/files#diff-92bc49d4e80dcc25621cfea3389eea5cb6d268e5cdd9d7674a87e18933a8e7c7

closing this so we can focus there.

PR #11885, opened 2024-11-20
feat(profiling): Add explanation for lockstep sampling

BYK:
We had this odd sampling frequency value in our docs which looks like a typo but it is not. This PR adds a note box to reassure the readers with some explanation and a reference link to why.

### DIFF
--- a/develop-docs/sdk/telemetry/profiles/sample-format-v1.mdx
+++ b/develop-docs/sdk/telemetry/profiles/sample-format-v1.mdx
@@ -295,6 +295,18 @@ stack ID.
 You should collect samples at a frequency of 101Hz, or roughly once every 10
 milliseconds.
 
+<Alert title="Note" level="info">
+The 101Hz number above is not a typo. It is intentionally slightly off from 100Hz to avoid a condition
+named "lockstep sampling" where the profiling samples occur at the same frequency as a loop in the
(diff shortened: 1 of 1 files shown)


develop-docs/sdk/telemetry/profiles/sample-format-v1.mdx
```diff
@@ -295,6 +295,18 @@ stack ID.
 You should collect samples at a frequency of 101Hz, or roughly once every 10
 milliseconds.
 
+<Alert title="Note" level="info">
+The 101Hz number above is not a typo. It is intentionally slightly off from 100Hz to avoid a condition
+named "lockstep sampling" where the profiling samples occur at the same frequency as a loop in the
+application. Ideally, the samples should be much more frequent than any cycles in the application, or
+at random intervals, so that the chance it occurs in any particular operation is proportional to the
+amount of time that operation takes. But this is often not feasible, so the next best thing is to use
+a sampling rate that doesn't coincide with the likely frequency of program cycles.
```
armcknight, line 304: A note on primality:
```suggestion
a sampling rate that doesn't coincide with the likely frequency of program cycles. 

We also chose 101 for its primality, whereas 1 below 100–99–is evenly divisible by several 
smaller numbers, which could lead to similar lockstep behavior.
```

viglia (approved):

PR #11886, opened 2024-11-20
feat(dynamic-sampling): adapt docs to new dynamic sampling logic

shellmayr:
Update the documentation for dynamic sampling to reflect the new status quo:
- Depending on the feature flag, dynamic sampling operates on transactions or spans
- Instead of using `transaction` for the event description, we use `events` to encompass both transactions and spans.
- Add a description of manual and automatic mode to set target sample rates


develop-docs/application-architecture/dynamic-sampling/fidelity-and-biases.mdx
```diff
@@ -13,19 +13,20 @@ A sample rate is a number in the interval `[0.0, 1.0]` that will determine the l
 
 ## The Concept of Fidelity
 
-At the core of Dynamic Sampling there is the concept of **fidelity**, which translates to an overall **target sample rate** that should be applied across all transactions of an organization.
+At the core of Dynamic Sampling there is the concept of **fidelity**, which translates to an overall **target sample rate** that should be applied across all events of an organization.
```
jan-auer, line 16: Today, it applies to some of our tracing data but not all. DS applies to spans and transaction events, but not to error  and message events from error monitoring. What further complicates this is that technically errors are a part of the trace, and as such the trace definition you deliver below is absolutely correct.

Note that also in many parts of the code base "Event" refers to a base type for the different kind of data collected through error monitoring (error, security, message), whereas spans have a different root type. Transaction events used to share that same base type for historical reasons, but over time were considered something else until we started considering their deprecation.

develop-docs/application-architecture/dynamic-sampling/fidelity-and-biases.mdx
```diff
@@ -13,19 +13,20 @@ A sample rate is a number in the interval `[0.0, 1.0]` that will determine the l
...
-
-_The `get_sampling_tier_for_volume`, like the `get_blend_sample_rate` function (defined [here](https://github.com/getsentry/sentry/blob/f3a2220ccd3a2118a1255a4c96a9ec2010dab0d8/src/sentry/quotas/base.py#L466)), is a function that must be overridden by the user to customize the behavior of Dynamic Sampling._
+### Target Sample Rate Adjustment: Automatic Mode and Manual Mode
+There are two available modes to govern the target sample rates for Dynamic Sampling: Automatic Mode and Manual Mode.
+- **Automatic mode** dynamically manages the target sample rate for each project based on the target sample rate for the organization, prioritizing lower volume projects to increase visibility.
+- **Manual mode** allows the user to set static target sample rates on a per-project basis that serve as the baseline sample rate before applying the dynamic biases outlined below. Target sample rates are not adjusted by the system.
```
jan-auer, line 21: Should we mention the in-code names here in contrast to the marketing term?

develop-docs/application-architecture/dynamic-sampling/fidelity-and-biases.mdx
```diff
@@ -71,30 +72,19 @@ An example of how the UI looks is shown in the following screenshot (the content
...
-    "*healthy*",
-    "*live*",
-    "*ready*",
-    "*heartbeat*",
-    "*/health",
-    "*/healthz",
-    # ...
-]
-```
+This bias is used to prioritize traces that are coming from a new release. The goal is to increase the sample rate in the time window that occurs between the creation of a release and its adoption by users. _The identification of a new release is done in the `event_manager` defined [here](https://github.com/getsentry/sentry/blob/master/src/sentry/event_manager.py#L937-L937)._
```
jan-auer, line 79: The line numbers are likely to move frequently. What would you prefer to make this a stable reference?
- Pin a current revision
- Reference some name from code
- Move the code to a more stable place and reference the file

develop-docs/application-architecture/dynamic-sampling/fidelity-and-biases.mdx
```diff
@@ -115,34 +105,40 @@ The list of development environments is available [here](https://github.com/gets
...
+This bias is used to prioritize low-volume transactions that can be drowned out by high-volume transactions. The goal is to rebalance sample rates of the individual transactions so that low-volume transactions are more likely to have representative samples. The bias is of type trace, which means that the transaction considered for rebalancing will be the root transaction of the trace.
 
-The latest release bias uses a decaying rule to interpolate between a starting sample rate and an ending sample rate over a time window that is statically defined for each platform (the list of time to adoptions is define [here](https://github.com/getsentry/sentry/blob/master/src/sentry/dynamic_sampling/rules/helpers/time_to_adoptions.py#L26-L26). For example, Android has a bigger time window than Javascript because on average Android apps take more time to get adopted by users.
+In order to rebalance transactions, the system computes the counts of the transactions for each project and runs an algorithm that, given the sample rate of the organization and the counts of each transaction, computes a new sample rate for each transaction assuming an ideal distribution of the counts.
```
jan-auer, line 112: In a recent conversation, you brought up that it is surprising how the algorithm tries to achieve an _equal_ sample count for each "class" (as it is called in code). This could be considered an implementation detail, but I think it's worth writing this down as the desired behavior of the system at this time.

develop-docs/application-architecture/dynamic-sampling/the-big-picture.mdx
```diff
@@ -6,29 +6,37 @@ sidebar_order: 1
 
 ![Sequencing](./images/sequencing.png)
 
+
+<Alert title="💡 Note" level="info">
+
+Dynamic Sampling currently operates on either spans or transactions, based on the feature flag `dynamic-sampling-spans`. The logic between the two event types is similar, so most of this documentation is kept at a generic level and important differences are pointed out explicitly.
```
jan-auer, line 12: The feature flag will be tied to what is the primary stats and billing metric according to the organization's plan. The default going forward is spans.

develop-docs/application-architecture/dynamic-sampling/the-big-picture.mdx
```diff
@@ -6,29 +6,37 @@ sidebar_order: 1
...
+
+</Alert>
+
+
 ## Sequencing
 
 Dynamic Sampling occurs at the edge of our ingestion pipeline, precisely in [Relay](https://github.com/getsentry/relay).
 
-When transaction events arrive, in a simplified model, they go through the following steps (some of which won't apply if you self-host Sentry):
+When events arrive, in a simplified model, they go through the following steps (some of which won't apply if you self-host Sentry):
```
jan-auer, line 21: Reading through it again, all of these apply in a self-hosted environment by now. We can remove this remark. 

```suggestion
When events arrive, in a simplified model, they go through the following steps:
```

develop-docs/application-architecture/dynamic-sampling/the-big-picture.mdx
```diff
@@ -37,49 +45,49 @@ The ingestion pipeline has two kinds of rate limits that  behave differently com
...
+Clients have their own [traces sample rate](https://docs.sentry.io/platforms/javascript/tracing/#configure). The client sample rate is a number in the range `[0.0, 1.0]` (from 0% to 100%) that controls **how many events arrive at Sentry**. While documentation will generally suggest a sample rate of `1.0`, for some use cases it might be better to reduce it.
 
-Dynamic Sampling further reduces how many transactions get stored internally. **While many-to-most graphs and numbers in Sentry are based on total transactions**, accessing spans and tags requires stored transactions. The sample rates apply on top of each other.
+Dynamic Sampling further reduces how many events get stored internally. **While most graphs and numbers in Sentry are based on total events**, accessing spans and tags requires stored events. The sample rates apply on top of each other.
```
jan-auer, line 68:
```suggestion
Dynamic Sampling further reduces how many events get stored internally. **While most graphs and numbers in Sentry are based on metrics**, accessing spans and tags requires stored events. The sample rates apply on top of each other.
```

Should we already foreshadow extrapolation? In that case, everything is based on stored.

jan-auer (commented):
Is there a place where you're planning to describe the newly introduced organization and project options? I also noticed that the biases don't describe how their behavior is different in the two modes. The most notable differences are:

- Boost low volume projects is disabled in manual mode
- Boost low volume transactions uses the sample rate from the project option
- All other biases compute factors that are independent of the sample rate
- Recalibration switches to project level

develop-docs/application-architecture/dynamic-sampling/fidelity-and-biases.mdx
shellmayr: Yeah good point - I was a bit worried about the redundancy, but I suppose pointing out the two namings is a valid way of adding context for anyone who will work on this.

develop-docs/application-architecture/dynamic-sampling/the-big-picture.mdx
shellmayr: I would prefer to not foreshadow - let's change that when it's rolled out, otherwise it may be confusing to understand where the docs can be trusted, and where they can't.

develop-docs/application-architecture/dynamic-sampling/fidelity-and-biases.mdx
shellmayr: Pinned it - after I had to search for it because it had indeed moved a lot 😅

jan-auer (approved):

PR #11897, opened 2024-11-21
Update docs on how to delete tables and columns

wedamija:
This uses the new operations from
https://github.com/getsentry/sentry/pull/81063
https://github.com/getsentry/sentry/pull/81098

### DIFF
--- a/develop-docs/api-server/application-domains/database-migrations/index.mdx
+++ b/develop-docs/api-server/application-domains/database-migrations/index.mdx
@@ -147,10 +147,10 @@ To avoid this, follow these steps:
 
 - (Optional, but ideal) Make a PR to remove all uses of the column in the codebase in a separate PR. This mostly helps with code cleanliness. This should be merged ahead of the migration prs, but we don't need to worry about whether it is deployed first.
 - Make another PR that:
- - Checks if the column is either not nullable, or doesn't have a db_default set. If either of these is true, then make it nullable via `null=True`.
- - If the column is a foreign key, remove the database level foreign key constraint it by setting `db_constraint=False`.
- - Remove the column and in the generated migration use `SafeRemoveField(..., deletion_action=DeletionAction.MOVE_TO_PENDING)` to replace `RemoveField(...)`. This only marks the state for the column as removed.
- - Combine these migrations together to save making multiple deploys
+    - Checks if the column is either not nullable, or doesn't have a db_default set. If either of these is true, then make it nullable via `null=True`.
+    - If the column is a foreign key, remove the database level foreign key constraint it by setting `db_constraint=False`.
+    - Remove the column and in the generated migration use `SafeRemoveField(..., deletion_action=DeletionAction.MOVE_TO_PENDING)` to replace `RemoveField(...)`. This only marks the state for the column as removed.
+    - Combine these migrations together to save making multiple deploys
 - Deploy. It's important that all previous prs are in production before we remove the actual column from the table.
 - Make a PR that create a new migration that has the same `SafeRemoveField` operation as before, but set `deletion_action=DeletionAction.DELETE` instead. This deletes the actual column from the table in Postgres.
 - Deploy
--- a/develop-docs/api-server/application-domains/database-migrations/index.mdx
+++ b/develop-docs/api-server/application-domains/database-migrations/index.mdx
@@ -153,7 +153,7 @@ To avoid this, follow these steps:
     - Combine these migrations together to save making multiple deploys
 - Deploy your migration changes. It's important that all previous pull requests are in production before we remove the actual column from the table.
 - Make a pull request that create a new migration that has the same `SafeRemoveField` operation as before, but set `deletion_action=DeletionAction.DELETE` instead. This deletes the actual column from the table in Postgres.
-- Deploy
+- Deploy the drop column migration.
 
 Here's an example of removing the `project` column from this model. It is both a foreign key and not null:
 
--- a/develop-docs/api-server/application-domains/database-migrations/index.mdx
+++ b/develop-docs/api-server/application-domains/database-migrations/index.mdx
@@ -251,8 +251,8 @@ Extra care is needed here if the table is referenced as a foreign key in other t
 - Make another pull request to:
     - Remove any database level foreign key constraints from this table to other tables by setting `db_constraint=False` on the columns.
     - Remove the model and in the generated migration use `SafeDeleteModel(..., deletion_action=DeletionAction.MOVE_TO_PENDING)` to replace `DeleteModel(...)`. This only marks the state for the model as removed.
-- Deploy. It's important that all previous prs are in production before we remove the actual table.
-- Make a PR that create a new migration that has the same `SafeDeleteModel` operation as before, but set `deletion_action=DeletionAction.DELETE` instead. This deletes the actual table from Postgres.
+- Deploy. It's important that all previous pull requests are in production before we remove the actual table.
+- Make a pull request that create a new migration that has the same `SafeDeleteModel` operation as before, but set `deletion_action=DeletionAction.DELETE` instead. This deletes the actual table from Postgres.
 - Deploy
 
 Here's an example of removing this model:
--- a/develop-docs/api-server/application-domains/database-migrations/index.mdx
+++ b/develop-docs/api-server/application-domains/database-migrations/index.mdx
@@ -252,7 +252,7 @@ Extra care is needed here if the table is referenced as a foreign key in other t
     - Remove any database level foreign key constraints from this table to other tables by setting `db_constraint=False` on the columns.
     - Remove the model and in the generated migration use `SafeDeleteModel(..., deletion_action=DeletionAction.MOVE_TO_PENDING)` to replace `DeleteModel(...)`. This only marks the state for the model as removed.
 - Deploy. It's important that all previous pull requests are in production before we remove the actual table.
-- Make a pull request that create a new migration that has the same `SafeDeleteModel` operation as before, but set `deletion_action=DeletionAction.DELETE` instead. This deletes the actual table from Postgres.
+- Make a pull request that creates a new migration that has the same `SafeDeleteModel` operation as before, but set `deletion_action=DeletionAction.DELETE` instead. This deletes the actual table from Postgres.
 - Deploy
 
 Here's an example of removing this model:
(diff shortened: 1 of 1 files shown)


develop-docs/api-server/application-domains/database-migrations/index.mdx
```diff
@@ -145,144 +145,202 @@ This is complicated due to our deploy process. When we deploy, we run migrations
...
-- Remove the column from the model, but in the migration make sure we only mark the state as removed.
-- Deploy.
-- Finally, create a migration that deletes the column.
+- (Optional, but ideal) Make a PR to remove all uses of the column in the codebase in a separate PR. This mostly helps with code cleanliness. This should be merged ahead of the migration prs, but we don't need to worry about whether it is deployed first.
+- Make another PR that:
+    - Checks if the column is either not nullable, or doesn't have a db_default set. If either of these is true, then make it nullable via `null=True`.
+    - If the column is a foreign key, remove the database level foreign key constraint it by setting `db_constraint=False`.
+    - Remove the column and in the generated migration use `SafeRemoveField(..., deletion_action=DeletionAction.MOVE_TO_PENDING)` to replace `RemoveField(...)`. This only marks the state for the column as removed.
+    - Combine these migrations together to save making multiple deploys
+- Deploy. It's important that all previous prs are in production before we remove the actual column from the table.
```
evanpurkhiser, line 154: Should we call this out even more explicitly, maybe a `<Note>` type thing?

markstory, line 154:
```suggestion
- Deploy your migration changes. It's important that all previous pull requests are in production before we remove the actual column from the table.
```

develop-docs/api-server/application-domains/database-migrations/index.mdx
```diff
@@ -145,144 +145,202 @@ This is complicated due to our deploy process. When we deploy, we run migrations
...
+- Make a PR that create a new migration that has the same `SafeRemoveField` operation as before, but set `deletion_action=DeletionAction.DELETE` instead. This deletes the actual column from the table in Postgres.
```
markstory, line 155:
```suggestion
- Make a pull request that create a new migration that has the same `SafeRemoveField` operation as before, but set `deletion_action=DeletionAction.DELETE` instead. This deletes the actual column from the table in Postgres.
```

develop-docs/api-server/application-domains/database-migrations/index.mdx
```diff
@@ -145,144 +145,202 @@ This is complicated due to our deploy process. When we deploy, we run migrations
...
+- Deploy
```
markstory, line 156:
```suggestion
- Deploy the drop column migration.
```

develop-docs/api-server/application-domains/database-migrations/index.mdx
```diff
@@ -145,144 +145,202 @@ This is complicated due to our deploy process. When we deploy, we run migrations
...
+operations = [
+    SafeRemoveField(model_name="testmodel", name="project", deletion_action=DeletionAction.DELETE),
 ]
 ```
 
+Then we deploy this and we're done. So to recap the steps here:
+ - Remove all references to the column in the code in a separate PR and merge. Doesn't matter if this deploys before the next step or not.
+ - If the column has an fk constraint them remove it. If it's not null and has no db_default then mark it as nullable. Then delete the column using SafeRemoveField(..., deletion_action=DeletionAction.MOVE_TO_PENDING). These operations can be in the same migration to save time.
+ - Deploy all previous before continuing.
+ - Remove the column from the table in from Postgres using SafeRemoveField(..., deletion_action=DeletionAction.DELETE),
```
markstory, line 244:
```suggestion
 - Remove all references to the column in the code in a separate pull request and merge. Doesn't matter if this deploys before the next step or not.
 - If the column has an foreign key constraint them remove it. If it's not null and has no db_default then mark it as nullable. Then delete the column using `SafeRemoveField(..., deletion_action=DeletionAction.MOVE_TO_PENDING)`. These operations can be in the same migration to save time.
 - Deploy all previous before continuing.
 - Remove the column from the table in from Postgres using `SafeRemoveField(..., deletion_action=DeletionAction.DELETE),`
```

develop-docs/api-server/application-domains/database-migrations/index.mdx
```diff
@@ -145,144 +145,202 @@ This is complicated due to our deploy process. When we deploy, we run migrations
...
 
 Extra care is needed here if the table is referenced as a foreign key in other tables. In that case, first remove the foreign key columns in the other tables, then come back to this step.
 
-- Remove any database level foreign key constraints from this table to other tables by setting `db_constraint=False` on the columns.
-- Deploy
-- Remove the model and all references from the sentry codebase. Make sure that the migration only marks the state as removed.
-- Deploy.
-- Create a migrations that deletes the table.
+- (Optional, but ideal) Make a PR to remove all uses of the model in the codebase in a separate PR. This mostly helps with code cleanliness. This should be merged ahead of the migration prs, but we don't need to worry about whether it is deployed first
+- Make another PR to
```
markstory, line 251:
```suggestion
- (Optional, but ideal) Make a pull request to remove all uses of the model in the codebase in a separate pull request. This mostly helps with code cleanliness. This should be merged ahead of the migration pull requests, but we don't need to worry about whether it is deployed first.
- Make another pull request to:
```

develop-docs/api-server/application-domains/database-migrations/index.mdx
```diff
@@ -145,144 +145,202 @@ This is complicated due to our deploy process. When we deploy, we run migrations
...
+    - Remove any database level foreign key constraints from this table to other tables by setting `db_constraint=False` on the columns.
+    - Remove the model and in the generated migration use `SafeDeleteModel(..., deletion_action=DeletionAction.MOVE_TO_PENDING)` to replace `DeleteModel(...)`. This only marks the state for the model as removed.
+- Deploy. It's important that all previous prs are in production before we remove the actual table.
+- Make a PR that create a new migration that has the same `SafeDeleteModel` operation as before, but set `deletion_action=DeletionAction.DELETE` instead. This deletes the actual table from Postgres.
```
markstory, line 255:
```suggestion
- Deploy. It's important that all previous pull requests are in production before we remove the actual table.
- Make a pull request that create a new migration that has the same `SafeDeleteModel` operation as before, but set `deletion_action=DeletionAction.DELETE` instead. This deletes the actual table from Postgres.
```

develop-docs/api-server/application-domains/database-migrations/index.mdx
```diff
@@ -145,144 +145,202 @@ This is complicated due to our deploy process. When we deploy, we run migrations
...
+    SafeDeleteModel(name="TestModel", deletion_action=DeletionAction.DELETE),
 ]
 ```
 
-Then we deploy this and we're done.
+Then we deploy this and we're done. So to recap the steps here:
+ - Remove all references to the model in the code in a separate PR and merge. Doesn't matter if this deploys before the next step or not.
+ - Remove any fk constraints and delete the model using SafeDeleteModel(..., deletion_action=DeletionAction.MOVE_TO_PENDING). These operations can be in the same migration to save time.
+ - Deploy all previous before continuing.
+ - Remove the table from Postgres using SafeDeleteModel(..., deletion_action=DeletionAction.DELETE),
```
markstory, line 343:
```suggestion
 - Remove all references to the model in the code in a separate pull request and merge. Doesn't matter if this deploys before the next step or not.
 - Remove any foreign key constraints and delete the model using `SafeDeleteModel(..., deletion_action=DeletionAction.MOVE_TO_PENDING)`. These operations can be in the same migration to save time.
 - Deploy all previous before continuing.
 - Remove the table from Postgres using `SafeDeleteModel(..., deletion_action=DeletionAction.DELETE),`
```

develop-docs/api-server/application-domains/database-migrations/index.mdx
wedamija: I'm not sure if it's necessary, people should just be following these steps in order, and we've had the docs like this for a while

markstory (approved):

develop-docs/api-server/application-domains/database-migrations/index.mdx
```diff
@@ -145,144 +145,202 @@ This is complicated due to our deploy process. When we deploy, we run migrations
 
 To avoid this, follow these steps:
 
-- Mark the column as nullable if it isn't, and create a migration. (ex. `BoundedIntegerField(null=True)`)
-- Deploy.
-- Remove the column from the model, but in the migration make sure we only mark the state as removed.
-- Deploy.
-- Finally, create a migration that deletes the column.
+- (Optional, but ideal) Make a PR to remove all uses of the column in the codebase in a separate PR. This mostly helps with code cleanliness. This should be merged ahead of the migration prs, but we don't need to worry about whether it is deployed first.
```
evanpurkhiser, line 148: I'm slightly confused about the optionality of this, is it just optional that it's a separate PR. It kind of makes it sound like we don't need to remove all the uses of the column, we definitely do.

I actually wouldn't mind if this just was the perscribed flow and remove the note about optionality.

```suggestion
- Make a PR to remove all uses of the column in the codebase in a separate PR. This mostly helps with code cleanliness. This should be merged ahead of the migration prs, but we don't need to worry about whether it is deployed first.
```

develop-docs/api-server/application-domains/database-migrations/index.mdx
```diff
@@ -145,144 +145,202 @@ This is complicated due to our deploy process. When we deploy, we run migrations
...
+- Make another PR that:
+    - Checks if the column is either not nullable, or doesn't have a db_default set. If either of these is true, then make it nullable via `null=True`.
```
evanpurkhiser, line 150:
```suggestion
    - Checks if the column is either not nullable, or doesn't have a `db_default` set. If either of these is true, then make it nullable via `null=True`.
```

develop-docs/api-server/application-domains/database-migrations/index.mdx
```diff
@@ -145,144 +145,202 @@ This is complicated due to our deploy process. When we deploy, we run migrations
...
+operations = [
+    migrations.RemoveField(model_name="testmodel", name="project"),
+]
+```
+
+Deleting this way is unsafe, so we want to replace this with `SafeDeleteModel` instead. So this becomes
+
+```python
+operations = [
+    SafeRemoveField(model_name="testmodel", name="project", deletion_action=DeletionAction.MOVE_TO_PENDING),
```
evanpurkhiser, line 208: Do you have to import `SafeRemoveField`? Should we monkey-patch it onto the `migrations` module?

develop-docs/api-server/application-domains/database-migrations/index.mdx
```diff
@@ -145,144 +145,202 @@ This is complicated due to our deploy process. When we deploy, we run migrations
...
+]
+```
+
+Using `SafeRemoveField` allows us to remove all the state related to the column, but defer deleting it until a later migration. So now, we can combine the migration to remove the constraints and delete the column state together like so
```
evanpurkhiser, line 212: It's a little confusing here that the next codeblock we're saying "That second migration you just made, throw that one away, we can combine them into the first"

We do need to describe both, since in the case where there's no alter needed for removing null or FKs, we just want the safe remove, but I feel like we should tweak this a bit to make it more clear that that's whats happening here

wedamija: Maybe it'd help to just post the two migrations necessary first, then write the steps to get to that state after?

develop-docs/api-server/application-domains/database-migrations/index.mdx
wedamija: Hmm, monkeypatching it in feels a bit weird... Usually I just quick import things using my IDE, so I'm not sure it's good to do even more monkeying here?

evanpurkhiser: Yeah sounds good. Let’s at least add the import to the code example though

develop-docs/api-server/application-domains/database-migrations/index.mdx
evanpurkhiser: I think that would make sense

wedamija: @evanpurkhiser I updated the docs for deleting a model. Let me know if it's an improvement, and then I'll update the column docs too

develop-docs/api-server/application-domains/database-migrations/index.mdx
wedamija: Yeah I think I agree. The process is flexible for me, because maybe there are very few uses and you want to just do those in the first pr, but probably better to just be prescriptive

PR #11915, opened 2024-11-22
docs(js): Add warning for usage of beforeSendSpan

chargome:
documents https://github.com/getsentry/sentry-javascript/issues/14335

### DIFF
--- a/includes/platforms/configuration/options/warn-before-send-span.mdx
+++ b/includes/platforms/configuration/options/warn-before-send-span.mdx
@@ -1,0 +1,5 @@
+{/* TODO(v9): Remove this alert in js sdk v9 */}
+
+<Alert level="warning">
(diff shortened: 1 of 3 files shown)


includes/platforms/configuration/options/warn-before-send-span.mdx
```diff
@@ -0,0 +1,5 @@
+{/* TODO(v9): Remove this alert in js sdk v9 */}
+
+<Alert level="warning">
+  In the Sentry JavaScript SDK v9.0.0, the ability to drop spans via `beforeSendSpan` will be removed. The callback will only allow modifying span attributes.
```
lforst, line 4:
```suggestion
  In the Sentry JavaScript SDK v9.0.0, the ability to drop spans via `beforeSendSpan` by returning `null` will be removed. The callback will only allow modifying span attributes.
```

no strong feeling tho

lforst (approved):

PR #11939, opened 2024-11-25
docs(sdks): New Span API

cleptric:


develop-docs/sdk/telemetry/spans/span-api.mdx
```diff
@@ -0,0 +1,61 @@
+---
+title: Span API
+---
+
+<Alert level="info">
+  This document uses key words such as "MUST", "SHOULD", and "MAY" as defined in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) to indicate requirement levels.
```
mydea, line 6: We should mention here that this only applies to non-OTEL SDKs.

Generally, I still feel a bit conflicted about designing a new Span API that is not compatible out of the box and by design with OTEL-based SDKs, which are the most important SDKs AFAIK (JS & Python, though not sure if python can possibly do these changes anyhow).

stephanie-anderson: As soon as we're clear with the Performance teams roadmap, I'd like to revamp the API discussion and make sure each team proposes, what they would think would make most sense. I think this puts us in a better situation to actually discuss different solutions and understand the reasoning behind them.

Lms24:
> _ No description provided. _

Lms24:
Updated this PR with the new API design discussed upfront with @cleptric and @Litarnus. The spec is minimal on purpose and hands a lot of additional API design decision to individual SDK authors to ensure APIs work with the platform.  We can of course refine and iterate on the minim specification further.

Lms24 (approved):
We can iterate on missing points (just noting them down from our last convo):

- we need an option to set a fully remote span as the parent span aka
- we need something like `sampled: boolean` to force a sampling decision

PR #11947, opened 2024-11-25
devenv: internal troubleshooting

joshuarli:
adds a link to internal troubleshooting doc + updates the obsoleted `make install-py-dev` with `devenv sync`


bc-sentry (approved):
LGTM

PR #11953, opened 2024-11-26
docs(astro): Update getting started page

chargome:
documents https://github.com/getsentry/sentry-javascript/issues/14274

- Documents manual configuration files as default
- Adds onboarding options for replay, tracing and profiling
- Removes manual setup subpage
- Adds a redirect from manual setup to getting started to avoid dead links from external pages


platform-includes/getting-started-config/javascript.astro.mdx
```diff
@@ -1,4 +1,11 @@
-Get started by adding your DSN to your Astro config file (`astro.config.mjs`):
+Configuration should happen as early as possible in your application's lifecycle.
```
lforst, line 1: @Lms24 please confirm but I don't think this particular statement makes sense for Astro because we already put the init code in an early spot for the user automatically.

platform-includes/getting-started-config/javascript.astro.mdx
```diff
@@ -1,4 +1,11 @@
...
+
+To set up the Sentry SDK, register the Sentry integration and initialize the SDK for client and server in the root directory of your project:
+
+- Client-Side: `sentry.client.config.ts` containing a `Sentry.init` call
+- Server-Side: `sentry.server.config.ts` containing a `Sentry.init` call
```
lforst, line 6: Can we somehow co-locate the code blocks with this paragraph? I would also err towards making the `defineConfig` the very first thing in this guide.

platform-includes/getting-started-config/javascript.astro.mdx
```diff
@@ -18,4 +24,57 @@ export default defineConfig({
 });
 ```
 
-Once you've added your `dsn`, the SDK will automatically capture and send errors and performance events to Sentry.
+<Alert level="warning">
+  Passing runtime-specific configuration options to the Sentry integration will be deprecated in future versions.
+  We recommend passing your configuration directly to the respective `Sentry.init()` calls in `sentry.client.config.js` and `sentry.server.config.js` instead.
```
lforst, line 29: I would be specific in saying we will no longer allow the  `dsn`, `environment`, `sampleRate`, `tracesSampleRate`, `replaysSessionSampleRate`, `replaysOnErrorSampleRate` options.

lforst (commented):
Generally works for me! I'll let @Lms24 also give it a pass though.

platform-includes/getting-started-config/javascript.astro.mdx
chargome: we can also deprecate `release` since it is not used if there is a config file

platform-includes/getting-started-config/javascript.astro.mdx
chargome: like not having separate headings for integration, client and server?

platform-includes/getting-started-config/javascript.astro.mdx
lforst: Release is actually conventionally fine as a build-time config because it also affects source-map upload.

platform-includes/getting-started-config/javascript.astro.mdx
lforst: Yeah, I don't have hard opinions on how we format it but it felt a bit weird that there is the build-time config in between this paragraph and the related headings.

platform-includes/getting-started-config/javascript.astro.mdx
chargome: Conventionally yes but I think it can be misleading since this option will be disregarded in the astro config – or maybe I'm missing something

lforst: Oh that's right - we're not forwarding it. Interesting. Yes then we can deprecate it.

platform-includes/getting-started-next-steps/javascript.astro.mdx
Lms24, line 1: l/no action required: I find it a bit unfortunate that the advanced config section is also on the Getting Started page. We might want to consider moving this somewhere else. Perviously, "Manual Setup" gave us the option to write all of this down but still have a short and concise "Getting Started" page. I think this might be worth to revisit later, potentially after the rework on the "Getting Started" pages (contrator?) of other SDKs is in a good state 🤔

Lms24 (commented):
Thanks for taking this on, Charly! General thoughts:

- I'm okay with removing manual setup and merging it with Getting Started as long as we don't have a wizard. Which in all likelihood we won't have for Astro. 
- I'd like us to continue using Astro's CLI to add the Astro integration because it's just what users are used to do. Unless there's a _good_ reason not to. Definitely open to hearing your thoughts on this though!

platform-includes/getting-started-install/javascript.astro.mdx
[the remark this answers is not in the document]
chargome: good point, I'll update that

platform-includes/getting-started-next-steps/javascript.astro.mdx
chargome: yeah was not sure where to put this tbh, I guess just adding an advanced config sub page also would not make sense?

platform-includes/getting-started-install/javascript.astro.mdx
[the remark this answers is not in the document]
chargome: the only thing that bugged me here is that we have another code snippet for adding profiling

platform-includes/getting-started-config/javascript.astro.mdx
```diff
@@ -18,4 +19,57 @@ export default defineConfig({
 });
 ```
 
-Once you've added your `dsn`, the SDK will automatically capture and send errors and performance events to Sentry.
+<Alert level="warning">
+  Passing runtime-specific configuration options (`dsn`, `release`, `environment`, `sampleRate`, `tracesSampleRate`, `replaysSessionSampleRate`, `replaysOnErrorSampleRate`) to the Sentry integration will be deprecated in future versions.
+  We recommend passing your configuration directly to the respective `Sentry.init()` calls in `sentry.client.config.js` and `sentry.server.config.js` instead.
+</Alert>
+
+### Client-side Setup
```
coolguyzone, line 27:
```suggestion
### Client-Side Setup
```

platform-includes/getting-started-config/javascript.astro.mdx
```diff
@@ -18,4 +19,57 @@ export default defineConfig({
...
+  // in development and sample at a lower rate in production
+  replaysSessionSampleRate: 0.1,
+
+  // If the entire session is not sampled, use the below sample rate to sample
+  // sessions when an error occurs.
+  replaysOnErrorSampleRate: 1.0,
+});
+```
+
+### Server-side Setup
```
coolguyzone, line 54:
```suggestion
### Server-Side Setup
```

platform-includes/getting-started-next-steps/javascript.astro.mdx
```diff
@@ -0,0 +1,89 @@
+## Advanced Configuration Options
+
+### Changing config files location
```
coolguyzone, line 3:
```suggestion
### Changing Config Files Location
```

platform-includes/getting-started-next-steps/javascript.astro.mdx
```diff
@@ -0,0 +1,89 @@
...
+  Sentry.handleRequest()
+  // other middleware handlers
+);
+```
+
+If you have multiple request handlers, make sure to add the Sentry middleware as the first handler in the sequence.
+
+### Customize Server Instrumentation
+
+Sentry's Astro middleware allows control over what additional data should be added to the recorded request spans.
```
coolguyzone, line 52:
```suggestion
Sentry's Astro middleware gives you control over what additional data should be added to the recorded request spans.
```

coolguyzone (approved):
Thanks for the updates! Added a couple suggestions, looks good to me! 🫡🦃

platform-includes/getting-started-next-steps/javascript.astro.mdx
chargome: @Lms24 merging this so we get this out for now, we can do a follow up!

Lms24: sorry, for not reviewing this in time! I'm fine with leaving it as-is for the moment

PR #11990, opened 2024-11-28
docs(react): Add React Router v7 docs

chargome:
This adds documentation for RR7 and also declutters the docs for RR in general by splitting up the versions into different files.

documents https://github.com/getsentry/sentry-javascript/issues/14518


lforst:
Or maybe even a completely separate page for v7 where we outline that the framework mode is not yet supported.

chargome:
Yeah I tried with a different page at first but imo it makes no sense since the content is basically identical. I'll update the codetabs and add an alert that we do not yet support rr7 framework

lforst:
> Yeah I tried with a different page at first but imo it makes no sense since the content is basically identical. I'll update the codetabs and add an alert that we do not yet support rr7 framework

It generally doesn't matter if content is identical. People see the version they use, click on it, and will never notice that v6 and v7 are the same. If anything it takes away complexity when reading the docs.

lforst (approved):

PR #12018, opened 2024-12-02
feat(rollback): Remove Rollback docs page

malwilley:
Sorry for the whiplash, but we decided that it would make more sense for the page to live in our help center instead of docs. Added here: https://github.com/getsentry/sentry-docs/pull/11982


coolguyzone (approved):
Thanks for the PR @malwilley , it looks like Google hasn't indexed this yet so I'm comfortable removing the page without any redirects. Let's merge before the crawler sees it! 🫡

PR #12023, opened 2024-12-03
ref(js): Add section about version alignment to Micro Frontends documentation

Lms24:
## DESCRIBE YOUR PR

Adds a section about the importance of having SDKs aligned to the same version in MFE setups plus an optional workaround for when this is not feasible. I opted to only link to the workaround posted in GH for the moment because we're still awaiting confirmation that this works and it requires some deep SDK behaviour intervention from users. We might revisit this and provide a better abstraction for it on demand.

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

## EXTRA RESOURCES

### DIFF
--- a/docs/platforms/javascript/common/best-practices/micro-frontends.mdx
+++ b/docs/platforms/javascript/common/best-practices/micro-frontends.mdx
@@ -21,11 +21,14 @@ keywords:
     "module metadata",
   ]
 ---
+
 <Alert>
-To ensure the Sentry JavaScript SDK works with your micro frontends, make sure all micro frontends that import from a `@sentry/*` package are using the same version of the Sentry SDK.
+  To ensure the Sentry JavaScript SDK works with your micro frontends, make sure
+  all micro frontends that import from a `@sentry/*` package are [using the same
+  version of the Sentry SDK](#sdk-version-alignment).
 </Alert>
 
-If your app uses micro frontends, it’s very useful to be able to track which one an error is coming from. To do this with Sentry, you can create either an automatic or a manual setup where you send events to separate Sentry projects representing each of your micro frontends. This makes it easier to see what’s going wrong and where, helping you track issues and fix them faster, especially in complex frontend architectures.
+If your app uses micro frontends, it's very useful to be able to track which one an error is coming from. To do this with Sentry, you can create either an automatic or a manual setup where you send events to separate Sentry projects representing each of your micro frontends. This makes it easier to see what's going wrong and where, helping you track issues and fix them faster, especially in complex frontend architectures.
 
 Below you'll find setup instructions for both an automatic and a manual way to route errors to different Sentry projects.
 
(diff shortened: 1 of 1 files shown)


smeubank (approved):

lizokm (approved):
Thanks so much for updating this! I added another PR with additional updates addressing the user feedback I got: https://github.com/getsentry/sentry-docs/pull/12040

docs/platforms/javascript/common/best-practices/micro-frontends.mdx
```diff
@@ -192,3 +192,17 @@ captureException(new Error("oh no!"), (scope) => {
   return scope;
 });
 ```
+
+## SDK Version Alignment
+
+Starting with version 8, if you have multiple Sentry JavaScript SDKs on the same page, they will [only interact with each other](https://github.com/getsentry/sentry-javascript/pull/12206) if they're all using the same version. This prevents unwanted cross-SDK interactions, for example, from browser extensions or 3rd party scripts.
+
+However, for micro frontends, where you _want_ SDK interaction between multiple SDKs, you'll need to ensure that all SDKs are using the same version.
```
davemyersworld, line 200: Perhaps something like: 

However, the micro frontend integration depends upon multiple configurations interacting with the same Sentry instance. This means that all microfrontends and the host _must_ use the same exact version of Sentry.

docs/platforms/javascript/common/best-practices/micro-frontends.mdx
```diff
@@ -192,3 +192,17 @@ captureException(new Error("oh no!"), (scope) => {
...
+
+Version alignment is important when you initialize the SDK in a host or skeleton application, but use Sentry SDK calls in micro frontend child applications. For example, if your individual micro frontends call `Sentry.captureException`.
```
davemyersworld, line 202: Might be worth adding something like "Module federation is one way to achieve version alignment."

docs/platforms/javascript/common/best-practices/micro-frontends.mdx
```diff
@@ -192,3 +192,17 @@ captureException(new Error("oh no!"), (scope) => {
   return scope;
 });
 ```
+
+## SDK Version Alignment
+
+Starting with version 8, if you have multiple Sentry JavaScript SDKs on the same page, they will [only interact with each other](https://github.com/getsentry/sentry-javascript/pull/12206) if they're all using the same version. This prevents unwanted cross-SDK interactions, for example, from browser extensions or 3rd party scripts.
```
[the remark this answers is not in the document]
Lms24: Yup, we can be more specific. I changed it to 8.7.0.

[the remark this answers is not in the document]
Lms24: I reworded this section a bit and focussed on incompatibility problems. 

And yes, you're right: The version key does not guarantee isolation between extensions and host apps. We also started blocking `Senty.init` calls in browser extension environments as we recommend extension users to follow a more [isolated Sentry client setup](https://docs.sentry.io/platforms/javascript/best-practices/shared-environments/).

docs/platforms/javascript/common/best-practices/micro-frontends.mdx
Lms24: > However, the micro frontend integration depends upon multiple configurations interacting with the same Sentry instance.

I don't think it strictly depends on same versions. Some MFE<>Sentry users just use Sentry in the host app and not in the MFEs at all. For them, the "multiple SDKs" part doesn't apply. However, I'll reword this because it's a bit misleading, I agree.

docs/platforms/javascript/common/best-practices/micro-frontends.mdx
Lms24: I'm by far no module federation expert but to me it module federation is one of many ways how to architect an MFE setup. How would module federation guarantee version alignment?

davemyersworld: The aspect of Module Federation that I was thinking about was the dependency sharing. Perhaps it could be said more generically. 

But the idea is: if the `@sentry/browser` dependency is _shared_ (amongst the host and MFEs), then it will necessarily be the same version. (Of course, if the MFE doesn't make use of that dependency and chooses to use its own version, that's another story).

Issue #11999, opened 2024-11-29
Docs-generated `authToken`s return 401 Unauthorized for generating releases


whitep4nth3r:
### Steps to Reproduce

1. Generate `authToken` via docs page (e.g. https://docs.sentry.io/platforms/javascript/guides/nuxt/manual-setup/#source-maps-upload)
2. Attempt to use generated `authToken` in CI/CD for e.g. generating sourcemap uploads and releases
3. Observe Sentry returns a 401 when attempting to use generated `authToken` for releases (sourcemaps upload is successful)

This might be intentional, because on the docs, the specified use case for the generated `authToken` is for sourcemaps upload. However, the Nuxt Sentry SDK attempted to automatically create a release during CI/CD build (without any further configuration from me) and this `authToken` did not have the permissions to send a POST request to `POST https://sentry.io/api/0/projects/the-claw/pantherworld/releases/`.

When generating an `authToken` via the Sentry app for the same organisation and project, this `authToken` had permissions for both sourcemap uploads AND posting a new release.

### Expected Result

`authTokens` generated on the docs pages have the correct permissions required for releases as well as sourcemap uploads.

### Actual Result

```
11:59:54 AM:   DEBUG   2024-11-29 11:59:54.865075967 +00:00 response status: 401
11:59:54 AM:   DEBUG   2024-11-29 11:59:54.865084898 +00:00 body: {"detail":"Invalid org token"}
11:59:54 AM:   error: API request failed
11:59:54 AM:   Caused by:
11:59:54 AM:   sentry reported an error: Invalid org token (http status: 401)
```

![Image](https://github.com/user-attachments/assets/1d07cdfd-d1ba-4e1f-90c8-9259e2604be1)
![Image](https://github.com/user-attachments/assets/3e5869a7-ebbe-4382-b1fa-787ef1dd0de6)

whitep4nth3r:
It looks like what happened is that I double clicked to copy the token, which didn't select the whole token.

I wonder how many other people have experienced this, so I wonder if there is a way to make this more obvious. Maybe we wrap the code in the code block so it's more obvious, it's also better for accessibility.

It's an old blog post but I wrote about it here: https://whitep4nth3r.com/blog/how-to-make-your-code-blocks-accessible-on-your-website/#elements-that-have-scrollable-content-should-be-accessible-by-keyboard

PR #12042, opened 2024-12-04
feat(platform): Add CI scope to auth token creation

chargome:
This PR adds the `org:ci` scope to the `createOrgAuthToken` request that is used for creating auth tokens within docs. With this the token should be both eligible for creating releases and uploading source maps.

(potentially) closes https://github.com/getsentry/sentry-docs/issues/11999


chargome:
The scope seems to be applied automatically anyway. closing

PR #12047, opened 2024-12-04
fix(platform): Sanitise next in platform-redirect

chargome:
Closes an open redirect for platform-redirects by adding better sanitising on the passed next param


geoffg-sentry (approved):
Excellent, much appreciated!

PR #12054, opened 2024-12-05
Clarify NestJS error filters

lforst:
## DESCRIBE YOUR PR

Hopefully makes it a bit more clear how specific error filters need to be transformed so that they will capture errors.

closes https://github.com/getsentry/sentry-javascript/issues/14580

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

### DIFF
--- a/platform-includes/getting-started-use/javascript.nestjs.mdx
+++ b/platform-includes/getting-started-use/javascript.nestjs.mdx
@@ -49,7 +49,6 @@ export class YourCatchAllExceptionFilter implements ExceptionFilter {
 }
 ```
 
-
 {/* TODO(v9): Remove this note */}
 _Note that `@SentryExceptionCaptured()` was called `@WithSentry` in SDK versions `8.38.0` and prior._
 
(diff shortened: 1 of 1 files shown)


platform-includes/getting-started-use/javascript.nestjs.mdx
```diff
@@ -94,3 +94,6 @@ export class ExampleExceptionFilter extends BaseExceptionFilter {
   }
 }
 ```
+
+By default, only unhandled exceptions that are not caught by an error filter are reported to Sentry.
+`HttpException`s also are not captured by default because they mostly act as control flow vehicles.
```
Lms24, line 99: l: let's maybe mention the derivatives of HttpException as there a quite a few
```suggestion
`HttpException`s (including [derivatives](https://docs.nestjs.com/exception-filters#built-in-http-exceptions)) are also not captured by default because they mostly act as control flow vehicles.
```

platform-includes/getting-started-use/javascript.nestjs.mdx
```diff
@@ -49,7 +49,6 @@ export class YourCatchAllExceptionFilter implements ExceptionFilter {
 }
 ```
 
-
```
Lms24, line 52: l: Can we also mention the last point, about HttpExceptions not being caught, below the code block just above this line? I think it's valuable information since users might expect otherwise in a catch-all filter?

Lms24 (approved):
Thanks for clarifying this! Reads way better now than before.

platform-includes/getting-started-use/javascript.nestjs.mdx
```diff
@@ -73,12 +72,13 @@ import { SentryGlobalFilter } from "@sentry/nestjs/setup";
...
 export class AppModule {}
 ```
+
 {/* TODO(v9): Remove this note. */}
 
 **Note:** If you have a NestJS + GraphQL application and you are using the `@sentry/nestjs` SDK version `8.38.0` or earlier, replace the `SentryGlobalFilter` with the `SentryGlobalGenericFilter`.
 In SDK versions `8.39.0` and above, the `SentryGlobalGenericFilter` is deprecated because the `SentryGlobalFilter` will handle GraphQL contexts automatically.
 
-By default, exceptions with status code 4xx are not sent to Sentry. If you still want to capture these exceptions, you can do so manually with `Sentry.captureException()`:
+If you have error filters for specific types of exceptions (for example `@Catch(HttpException)`, or any other `@Catch(...)` with arguments) and you want to capture errors caught by these filters, capture the errors in the `catch()` handler with `Sentry.captureException()`:
```
chargome, line 81: In this case they would also need to remove our `SentryExceptionCaptured` decorator, otherwise unexpected errors get captured twice right?

lforst: This is a tricky one to think through. In general, our dedupe integration will drop the second capture call, as long as the two capture calls happen consecutively (no capture call on another error in between).

I am not sure if we should tell people to remove their `SentryExceptionCaptured` decorators though. The two things serve different purposes. One is more fine grained the other more blanket.

platform-includes/getting-started-use/javascript.nestjs.mdx
Lms24: LGTM as you have it now

PR #12078, opened 2024-12-10
ref(profiling) fix method name

JonasBa:
Method name was wrong as identified [here](https://github.com/getsentry/sentry-javascript/issues/14561#issuecomment-2522295906), the correct naming is startProfiler and stopProfiler


lizokm (approved):
Thanks for updating!

PR #12079, opened 2024-12-10
feat: Analytics scroll depth

### LINKED ISSUE #12061: Add Scroll Depth tracking on Plausible
### Problem Statement

Matt Henderson and I have added Scroll Depth tracking to the marketing site and blog on Plausible, and Liza would like this expanded to docs.

### Solution Brainstorm

Here are the [code](https://github.com/getsentry/static-sites/pull/2840/files) on the blog and some [additional insights](https://github.com/plausible/analytics/discussions/121). We currently have it setup with a custom property readProgress connected to the goal Read Progress (both are setup in Docs Plausible to start collecting data).

a-hariti:
closes #12061


app/layout.tsx
```diff
@@ -3,7 +3,7 @@ import './globals.css';
 import {Theme} from '@radix-ui/themes';
 import type {Metadata} from 'next';
 import {Rubik} from 'next/font/google';
-import Script from 'next/script';
+import PlausibleProvider from 'next-plausible';
```
chargome, line 6: What is the advantage of having this instead of the script?

src/components/track-reader-depth.tsx
```diff
@@ -0,0 +1,65 @@
...
+        reachedMilestones.add(milestone);
+        sendProgressToPlausible(milestone);
+      }
+    };
+
+    // if the page is not scrollable, we don't need to track anything
+    if (document.documentElement.scrollHeight - window.innerHeight === 0) {
+      return () => {};
+    }
+    const debouncedTrackProgress = debounce(trackProgress, 20);
```
chargome, line 56: As discussed offline: this prevents events from being sent when a user very quickly scrolls to the bottom and then leaves the page - which might not count as a read anyway

chargome (approved):

app/layout.tsx
a-hariti: send events dynamically through a hook, nice TS and local DX experience

app/layout.tsx
```diff
@@ -31,6 +31,9 @@ export const metadata: Metadata = {
 export default function RootLayout({children}: {children: React.ReactNode}) {
   return (
     <html lang="en" suppressHydrationWarning>
+      <head>
+        <PlausibleProvider domain="docs.sentry.io,rollup.sentry.io" />
```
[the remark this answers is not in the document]
a-hariti: Probably, we should ask @lizokm

src/components/track-reader-depth.tsx
a-hariti: yes, it doesn't happen if you scroll at any reasonable reading speed

app/layout.tsx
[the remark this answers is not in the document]
elijames-codecov: It doesn't look like we have Plausible installed on dev docs (if it is, it's not hooked up to the company account).

src/components/track-reader-depth.tsx
elijames-codecov: Have you considered using the IntersectionObserver API for this?

a-hariti: I think the current approach is more precise

the marketing site / blog use the intersection observer and report 100% when the last heading is reached

a useful heuristic, but not very precise

a-hariti: please keep in mind the measurement on the marketing sites is very rough (based on headers instead of the actual page scroll)

But I can do it @elijames-codecov

a-hariti: turns out I had bug in my `debounce` logic, now it works as intended

it's simple, precise and performant

maybe we should port it to the marketing site @elijames-codecov once we make sure it works well?

src/components/track-reader-depth.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const plausible = usePlausible<{[EVENT]: EVENT_PROPS}>();
+
+  const sendProgressToPlausible = (progress: Milestone) => {
+    // TODO: remove this after PR review
+    const args = [
+      EVENT,
+      {props: {readProgress: progress, page: document.title}},
+    ] as const;
+    plausible(...args);
+    console.log('plausible event', ...args);
```
elijames-codecov, line 22: We probably want to remove this.

src/components/track-reader-depth.tsx
elijames-codecov: @a-hariti - Glad you got the debounce working! Since we're looking for insights as to which sections the users scroll into (to inform if there's seeing certain content and make content and some layout decisions), each section having a header, IntersectionObserver actually works well for our needs without added complexity and expensive operations...which is why we will continue to use it.

a-hariti:
Are you available to make sure things work as expected after the merge @elijames-codecov? (I don't have access to Analytics)

a-hariti:
@stephanie-anderson 👀

chargome:
@a-hariti can we merge this one? We have some upcoming tasks for plausible and would like to make use of that integration

a-hariti:
Yes, it just needs someone to make sure events are sent properly

chargome:
I can check that

PR #12085, opened 2024-12-10
fix: Make DSN snippet clickable

chargome:
ref https://github.com/getsentry/sentry-docs/issues/11556


coolguyzone (approved):
Works as expected, changes look good 👍

Issue #8471, opened 2023-11-10
Add attachments types supporting preview


krystofwoldrich:
### Core or SDK?

Core Sentry product

### Which part? Which one?

Events Attachments

### Description

Currently, the docs don't describe mime types supported by the preview feature. 

### Suggested Solution

List attachment types that are supported by the preview feature in Sentry UI (Event Detail Page).

https://github.com/getsentry/sentry/blob/923e301e5d23bb259255f7469622afd1419bad91/static/app/components/events/eventAttachments.tsx#L38

PR #12118, opened 2024-12-12
ref(docs): add info about mimetype in docs

michellewzhang:
add specific info in the docs about which MIME types are supported / renderable for event attachments.

based on `previewAttachmentTypes.tsx`:

https://github.com/getsentry/sentry/blob/bf167f9145882b76bc1f38eb4b7f62bedf6d9b71/static/app/components/events/attachmentViewers/previewAttachmentTypes.tsx#L7-L32

closes https://github.com/getsentry/sentry-docs/issues/8471


coolguyzone (approved):
Thanks for the update! Looks good.

Issue #12138, opened 2024-12-16
Unclear that the JavaScript loader does not support User Feedback


whitep4nth3r:
### SDK

Android SDK

### Description

I'm using the JavaScript loader on my website.

Whenever I'm browsing the top-level docs of JavaScript, I am expecting the instructions to apply to the loader.

On [this page](https://docs.sentry.io/platforms/javascript/user-feedback/#installation) regarding User Feedback, there are instructions for both the SDKs and the "CDN bundle". Given I am using the loader, imported via a script tag, I assumed I was using the CDN bundle, especially since the src URL for the loader contains the string `cdn`.

I made a lot of notes on the information architecture of the docs regarding the JavaScript SDKs and loader, which I am going to dump here.

Related issue: https://github.com/getsentry/sentry-docs/issues/11705

Current state:

- The documentation for the “JavaScript loader” lives under the top-level heading of JavaScript: https://docs.sentry.io/platforms/javascript/)](https://docs.sentry.io/platforms/javascript/
- JavaScript-based frameworks and libraries (such as Next.js, Vue, React) are nested under “JavaScript”.
    - Whilst you can definitely use the JavaScript loader for any front-end project, it is recommended to use the official framework/library SDKs.
- Sometimes, it is unclear that the “JavaScript” top-level documentation references the vanilla JavaScript SDK usage. It becomes more clear when looking at the code samples that reference using the loader script with <script> tags.
    - Granted, there is a callout at the top of the JavaScript top-level page, but developers will often miss this by trying to find the code they need to solve their problem as quickly as possible

### Suggestions:

- Nest the “JavaScript” (loader/vanilla SDK) under “browser”.
    - This introduces one more click for the developer needing these docs, but makes what developers are looking at clearer.
- On the base “JavaScript” route, display all available SDKs, similar to how the SDK selection is made on the docs home page

### Problems this may cause:

- Other languages are set up in the same way as JavaScript, such as Ruby (i.e. the top-level page for Ruby shows instructions for setting up the SDK for raw Ruby, and has a callout for if you’re using a Ruby framework) — this makes sense for Ruby
    - Note that PHP does *not* have this callout, despite Sentry providing SDKs for specific frameworks (why is this?)
- Rearchitecting the information for JavaScript introduces an inconsistency in how information is presented across different SDKs
    - **HOWEVER, JavaScript is an outlier given that developers are building both server and client applications with this language.**

### Secondary suggestion based on both of the above:

[...]

PR #12140, opened 2024-12-16
fix(js): Fix docs for using feedback with the Loader

mydea:
Closes https://github.com/getsentry/sentry-docs/issues/12138

We did not really document well how to use user feedback with the loader. So I expanded this section in the docs to properly explain how it can be used.

### DIFF
--- a/platform-includes/user-feedback/install/javascript.mdx
+++ b/platform-includes/user-feedback/install/javascript.mdx
@@ -10,7 +10,9 @@ window.sentryOnLoad = function () {
 
   Sentry.lazyLoadIntegration("feedbackIntegration")
     .then((feedbackIntegration) => {
-      Sentry.addIntegration(feedbackIntegration());
+      Sentry.addIntegration(feedbackIntegration({
+      	// User Feedback configuration options
+      }));
     })
     .catch(() => {
       // this can happen if e.g. a network error occurs,
--- a/platform-includes/user-feedback/install/javascript.mdx
+++ b/platform-includes/user-feedback/install/javascript.mdx
@@ -21,7 +21,7 @@ window.sentryOnLoad = function () {
 };
 ```
 
-If you're using CDN bundles instead of NPM packages, you need to load the User Feedback integration CDN bundle in addition to your browser bundle:
+If you're using CDN bundles instead of NPM packages, you need to use a respective bundle that includes user feedback:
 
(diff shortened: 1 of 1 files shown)


lforst (approved):

platform-includes/user-feedback/install/javascript.mdx
```diff
@@ -1,17 +1,26 @@
...
+window.sentryOnLoad = function () {
+  Sentry.init({
+    // add other configuration here
+  });
 
-```bash {tabTitle:pnpm}
-pnpm add @sentry/browser
+  Sentry.lazyLoadIntegration("feedbackIntegration")
+    .then((feedbackIntegration) => {
+      Sentry.addIntegration(feedbackIntegration());
```
whitep4nth3r, line 13:
```suggestion
      Sentry.addIntegration(feedbackIntegration({
      	// configuration options
      }));
```

whitep4nth3r: I just feel like this makes it clearer where to add config, as I had to double check this and wasn't confident

whitep4nth3r (approved):

platform-includes/user-feedback/install/javascript.mdx
```diff
@@ -1,17 +1,28 @@
...
+      }));
+    })
+    .catch(() => {
+      // this can happen if e.g. a network error occurs,
+      // in this case User Feedback will not be enabled
+    });
+};
 ```
 
+If you're using CDN bundles instead of NPM packages, you need to load the User Feedback integration CDN bundle in addition to your browser bundle:
```
ryan953, line 24: "in addition" seems like it should go away. People should pick one CDN script with the combination of integrations that they want to have.

mydea: yeah, good point, we call this out in the snippet anyhow!

mydea, line 24:
```suggestion
If you're using CDN bundles instead of NPM packages, you need to use a respective bundle that includes user feedback:
```

mydea: @ryan953 like this?

coolguyzone (approved):
Thanks for the update! LGTM

billyvg (approved):

PR #12160, opened 2024-12-17
chore: add redirect from api-server to backend

shellmayr:
[After we reverted the name change from api-server to backend](https://github.com/getsentry/sentry-docs/pull/12141), links on google, etc. don't work for the time being. This PR adds a redirect from api-server to backend to restore functionality.


antonpirker (approved):
thanks for fixing this!

chargome (approved):

PR #12169, opened 2024-12-17
feat(develop): Clarify adding constraints

isabellaenriquez:
## DESCRIBE YOUR PR
Updates develop docs Database Migrations page to clarify the caveats for adding any type of constraint (as opposed to just not null).

### Adding Columns
image

###  Adding Not Null to Columns -> Adding Constraints to Columns (Including Not Null)
image

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## PRE-MERGE CHECKLIST

### DIFF
--- a/develop-docs/backend/application-domains/database-migrations/index.mdx
+++ b/develop-docs/backend/application-domains/database-migrations/index.mdx
@@ -440,8 +440,11 @@ With postgres 14, columns can be added to tables of all sizes as deploy time
 migrations if you follow the guidelines on default values & allowing nulls. When
 creating new columns they should either be:
 
-- Not null with a default. https://develop.sentry.dev/database-migrations/#adding-not-null-to-columns
+- Not null with a default
 - Created as nullable. If no default value can be set on the column, then it's best just to make it nullable.
+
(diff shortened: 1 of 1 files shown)


vbro (approved):
LGTM
wait for owners-migrations before merging

vbro:
FYI: https://www.postgresql.org/docs/current/sql-altertable.html#SQL-ALTERTABLE-NOTES

```
Scanning a large table to verify a new foreign key or check constraint can take a long time, and other updates to the table are locked out until the ALTER TABLE ADD CONSTRAINT command is committed ...
```

isabellaenriquez:
@vbro thanks, this is covered in the section already:
> One approach is to create the constraint as not valid. Then we validate it afterwards. We still need to scan the whole table to validate, but we only need to hold a SHARE UPDATE EXCLUSIVE lock, which only blocks other ALTER TABLE commands, but will allow reads/writes to continue. This works well, but has a slight performance penalty of 0.5-1%. After Postgres 12 we can extend this method to add a real NOT NULL constraint.

vgrozdanic (approved):
looks good, thanks for updating the docs 🙌

wedamija (commented):
Thanks for adding this - I'm wondering if we should considering just banning the positive integer fields as well. Either that, or modify them to validate the constraint afterwards without the statement timeout

isabellaenriquez:
@wedamija We use them a lot on billing side for currency fields, so I'd lean towards the latter solution

PR #12189, opened 2024-12-19
docs(react): Update warning on react router library docs

chargome:
- Updates the warning on order of calling react router related functions.
- Updates a reference to a legacy function


coolguyzone (approved):
Looks good!

PR #12194, opened 2024-12-20
feat(js): Update automatic instrumentation docs

mydea:
This PR updates the browser JS "Automatic Instrumentation" docs:

1. The `heartbeatTimeout` option does not exist anymore, I replaced it with the new `childSpanTimeout`. This has been replaced in v8.
2. We used "transaction" everywhere still, I rewrote this to use the term span instead.

Closes https://github.com/getsentry/sentry-javascript/issues/14805


Lms24 (approved):

coolguyzone (approved):
Thanks for the update!

PR #12215, opened 2024-12-26
ref(flags): move min version requirements from index to integration docs

aliu39:
As we add and release more integrations, the versioning will be different for each one, so we should not specify it in the list of integrations.

### DIFF
--- a/docs/platforms/javascript/common/configuration/integrations/openfeature.mdx
+++ b/docs/platforms/javascript/common/configuration/integrations/openfeature.mdx
@@ -25,11 +25,12 @@ notSupported:
 
 <Alert level="info">
 
-This integration only works inside a browser environment. It is only available from a package-based install (e.g. `npm` or `yarn`), version 8.43.0 or higher.
+This integration only works inside a browser environment. It is only available from a package-based install (e.g. `npm` or `yarn`).
 
 </Alert>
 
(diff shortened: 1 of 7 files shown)


docs/platforms/javascript/common/configuration/integrations/openfeature.mdx
```diff
@@ -29,7 +29,8 @@ This integration only works inside a browser environment. It is only available f
 
 </Alert>
 
-The [OpenFeature](https://openfeature.dev/) integration tracks feature flag evaluations produced by the OpenFeature SDK. These evaluations are held in memory, and in the event an error occurs, sent to Sentry for review and analysis. **At the moment, we only support boolean flag evaluations.**
+The [OpenFeature](https://openfeature.dev/) integration tracks feature flag evaluations produced by the OpenFeature SDK. These evaluations are held in memory, and in the event an error occurs, sent to Sentry for review and analysis. **At the moment, we only support boolean flag evaluations.** This integration is available in
+Sentry SDK **versions 8.43.0 or higher.**
```
ryan953, line 33: Plural 'versions'?

ryan953 (approved):

docs/platforms/javascript/common/configuration/integrations/openfeature.mdx
aliu39: thought it flowed a little better, but I'm ok either way

PR #12235, opened 2025-01-02
🔧 chore: cleanup slack & msteams dev docs

iamrajjoshi:
cleaned up the instructions for slack and msteams installation for development

### DIFF
--- a/develop-docs/integrations/discord/index.mdx
+++ b/develop-docs/integrations/discord/index.mdx
@@ -28,7 +28,7 @@ discord.client-secret: "<client secret>"
 discord.bot-token: "<bot token>"
 ```
 
-**NOTE**: If you're doing local Sentry development, you should use a tool like Ngrok and use `getsentry` repo when starting your Sentry server. Otherwise, installing the Discord bot won't work, as it will not be able to communicate with your local Sentry instance.
+**NOTE**: If you're doing local Sentry development, you should use a tool like Ngrok to expose your local Sentry instance to the internet. Otherwise, the OAuth2 Redirect URI will not work.
 
 After you update the <code>config.yml</code>, restart your Sentry server to continue the setup process.
 
--- a/develop-docs/integrations/discord/index.mdx
+++ b/develop-docs/integrations/discord/index.mdx
@@ -28,7 +28,7 @@ discord.client-secret: "<client secret>"
 discord.bot-token: "<bot token>"
 ```
 
-**NOTE**: If you're doing local Sentry development, you should use a tool like Ngrok and add `system.url-prefix: "https://<your_subdomain>.ngrok.io"` to your `~/.sentry/config.yml`. Otherwise, installing the Discord bot won't work, as it will not be able to communicate with your local Sentry instance.
+**NOTE**: If you're doing local Sentry development, you should use a tool like Ngrok and use `getsentry` repo when starting your Sentry server. Otherwise, installing the Discord bot won't work, as it will not be able to communicate with your local Sentry instance.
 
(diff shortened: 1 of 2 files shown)


develop-docs/integrations/discord/index.mdx
```diff
@@ -28,7 +28,7 @@ discord.client-secret: "<client secret>"
 discord.bot-token: "<bot token>"
 ```
 
-**NOTE**: If you're doing local Sentry development, you should use a tool like Ngrok and add `system.url-prefix: "https://<your_subdomain>.ngrok.io"` to your `~/.sentry/config.yml`. Otherwise, installing the Discord bot won't work, as it will not be able to communicate with your local Sentry instance.
+**NOTE**: If you're doing local Sentry development, you should use a tool like Ngrok and use `getsentry` repo when starting your Sentry server. Otherwise, installing the Discord bot won't work, as it will not be able to communicate with your local Sentry instance.
```
cathteng, line 31: we still need to use ngrok right? does it matter if you use getsentry or sentry if you use ngrok?

develop-docs/integrations/slack/index.mdx
```diff
@@ -21,11 +21,13 @@ slack.client-secret: <client secret>
 slack.signing-secret: <signing secret>
 ```
 
-**NOTE**: If you're doing local Sentry development, you should use a tool like Ngrok and add `system.url-prefix: "https://<your_subdomain>.ngrok.io"` to your `~/.sentry/config.yml`, otherwise, installing the Slack APP will complain because the redirect URI points to `localhost:8000`. Also, install the Slack integration from the ngrok subdomain or it will fail to install.
+If you a Sentry employee, you can set the `slack.client-id` and `slack.client-secret` in `devlocal.py` inside `getsentry` instead.
```
cathteng, line 24: is devlocal.py only a sentry employee thing or can anyone add it and use it for local dev?

cathteng (commented):
questions

develop-docs/integrations/slack/index.mdx
iamrajjoshi: its a sentry employee thing. i had to think about this for a while, whether i should mention it here, but i felt it was ok because we explain the difference between `sentry` and `getsentry` 

https://develop.sentry.dev/application-architecture/sentry-vs-getsentry/

develop-docs/integrations/discord/index.mdx
iamrajjoshi: the `getsentry` repo is where the Ngrok config is more actively maintained and will have everything preconfigured and setup so u don't need to modify stuff like url-prefix which has been janky in the past. i have linked the  ngrok docs i wrote under the diff here with more details

develop-docs/integrations/slack/index.mdx
coolguyzone, line 24:
```suggestion
If you're a Sentry employee, you can set the `slack.client-id` and `slack.client-secret` in `devlocal.py` inside `getsentry` instead.
```

coolguyzone (approved):
Thanks for updating!

develop-docs/integrations/discord/index.mdx
iamrajjoshi: the `getsentry` repo is where the Ngrok config is more actively maintained and will have everything preconfigured and setup so u don't need to modify stuff like url-prefix which has been janky in the past.

iamrajjoshi: i updated this since we fixed the sentry ngrok presets

ameliahsu (approved):

PR #12268, opened 2025-01-07
Document how to attach data attributes onto a span/transaction in PHP

cleptric:
https://github.com/getsentry/projects/issues/524


coolguyzone (approved):
Thanks for the add, looks good!

PR #12269, opened 2025-01-07
mobile: react native options

bruno-garcia:
The call is out of context, so customers were not sure where to put this, or the fact that it should update the current call they have under `integrations`.


krystofwoldrich (approved):
Thank you, that's a good point.

PR #12270, opened 2025-01-07
docs(flags): update generic docs

michellewzhang:
renames self-serve to generic

PR #12274, opened 2025-01-07
ref(flags): rename self-serve in generic ffs index

aliu39:
follow-up on https://github.com/getsentry/sentry-docs/pull/12270, missed a spot


michellewzhang (approved):
THANKS!

PR #12284, opened 2025-01-08
fix(changelog): Fix jumpy search

lforst:


chargome (approved):
nice, never noticed this before

PR #12285, opened 2025-01-09
Remove backup testing notes

wedamija:
After https://github.com/getsentry/sentry/pull/83140 and https://github.com/getsentry/sentry/pull/83143 these no longer apply


coolguyzone (approved):
Thanks for the fix 🛠️

PR #12309, opened 2025-01-13
Add React SDK's `wrapCreateMemoryRouterVX` docs.

onurtemizkan:
Documents: https://github.com/getsentry/sentry-javascript/pull/14854

### DIFF
--- a/docs/platforms/javascript/guides/react/features/react-router/v6.mdx
+++ b/docs/platforms/javascript/guides/react/features/react-router/v6.mdx
@@ -5,7 +5,8 @@ sidebar_order: 20
 ---
 
 <Alert level="info">
-- React Router v6 support is included in the `@sentry/react` package since version `7`.
+  - React Router v6 support is included in the `@sentry/react` package since
+  version `7`.
 </Alert>
 
 Update your `Sentry.browserTracingIntegration` to `Sentry.reactRouterV6BrowserTracingIntegration` and provide the required React hooks and router functions:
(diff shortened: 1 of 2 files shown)


docs/platforms/javascript/guides/react/features/react-router/v6.mdx
```diff
@@ -19,18 +19,21 @@ Update your `Sentry.browserTracingIntegration` to `Sentry.reactRouterV6BrowserTr
...
+### Usage with `createBrowserRouter` or `createMemoryRouter`
 
-<Alert level="info">
-Available in `@sentry/react` version `7.21.0` and above.
-</Alert>
+If you choose to create your router instance with [`createBrowserRouter`](https://reactrouter.com/en/main/routers/create-browser-router) or [`createMemoryRouter`](https://reactrouter.com/en/main/routers/create-memory-router), you can use `Sentry.wrapCreateBrowserRouterV6` or `Sentry.wrapCreateMemoryRouterV6` to wrap it with the instrumentation:
+
+<Alert level="warning" title="Note">
+
+`wrapCreateMemoryRouterV6` is available from `@sentry/react` version `8.49.0`. For versions prior to `8.49.0`, you can use `wrapCreateBrowserRouterV6` with `createMemoryRouter`.
```
lforst, line 32: > you can use `wrapCreateBrowserRouterV6` with `createMemoryRouter`.

is that true actually?

onurtemizkan: Yes, we were recommending that prior to this update. (Without the support of `initialEntries` though)

lforst: Ah right, for some reason I thought it was completely broken. I would maybe rephrase this so it is more specific as to what to do:

```suggestion
`wrapCreateMemoryRouterV6` was introduced in SDK version 8.49.0. Prior to that version, we suggested using `wrapCreateBrowserRouterV6` with `createMemoryRouter`. If you are currently using `wrapCreateBrowserRouterV6` to wrap `createMemoryRouter`, it is recommended that you use `wrapCreateMemoryRouterV6` instead.
```

Reads horribly though 😂

lforst (approved):

Issue #12115, opened 2024-12-12
🔗 404 Error


elijames-codecov:
### URL

docs.sentry.io/product/error-monitoring/stacktrace/

a-hariti:
any idea what this should point to now @elijames-codecov ?

also #12114

Issue #12114, opened 2024-12-12
🔗 404 Error


elijames-codecov:
### URL

docs.sentry.io/product/error-monitoring/user-impact/

PR #12311, opened 2025-01-13
fix(404): Add issue details related redirects 

chargome:
closes https://github.com/getsentry/sentry-docs/issues/12115
closes https://github.com/getsentry/sentry-docs/issues/12114


stephanie-anderson (commented):
I think we should also add a catch-all redirect for `/product/error-monitoring/*` underneath the specific ones, so that we don't have any more 404s + we can always add specific redirects if need be

a-hariti (approved):

PR #12345, opened 2025-01-14
ref(flags): update python supported vers and install snippets

aliu39:
Update python integration docs to follow the pattern of [clickhouse-driver](https://docs.sentry.io/platforms/python/integrations/clickhouse-driver/). Versions are based on `tox.ini` and extras_require dict in `setup.py` of python repo. Using the extra (e.g. `[launchdarkly]`) will automatically install the required dependencies.

LD after:
Screenshot 2025-01-14 at 3 57 46 PM

Before:
Screenshot 2025-01-14 at 3 58 17 PM


coolguyzone (approved):
Looks good, thanks for the update.

PR #12347, opened 2025-01-15
Update develop docs data handling section for new APIs

mydea:
Noticed this was using outdated API names.

### DIFF
--- a/develop-docs/sdk/expected-features/data-handling.mdx
+++ b/develop-docs/sdk/expected-features/data-handling.mdx
@@ -69,6 +69,10 @@ This helps Relay to know what kind of data it receives and this helps with scrub
   });
   ```
 
+  ```python
+  span.set_data("http.query", query)
+  ```
+
   Additionally all semantic conventions of OpenTelementry for http spans should be set in the `span.data` if applicable:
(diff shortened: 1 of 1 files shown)


develop-docs/sdk/expected-features/data-handling.mdx
```diff
@@ -63,7 +63,7 @@ This helps Relay to know what kind of data it receives and this helps with scrub
   If query strings are present in the URL or fragments (Browser SDKs only) are present in the URI, both are set into the data attribute of the span.
 
   ```js
-  span.setData({
+  span.setAttributes({
```
cleptric, line 66: Only JS exposes `setAttributes`.

mydea: sad, but JS does not expose `setData` anymore 😅 so if we use a JS example here it should probably be attributes?

cleptric: let’s add both then

mydea: I added a JS and python section here now?

cleptric (approved):
Thanks!

PR #12352, opened 2025-01-15
ref(sveltekit): Remove nonce approach in favor of hash for optional CSP config

Lms24:
The SvelteKit SDK injects a small inline `<script>` into the server-side rendered HTML responses of SvelteKit applications. Due to concerns raised with setting a nonce to except this script from being blocked, this PR removes the nonce documentation in favor of a better, hash-based CSP exception approach.


coolguyzone (approved):
Thanks for the updates, LGTM!

PR #12375, opened 2025-01-16
Clean up Symfony options

cleptric:
Follow up on https://github.com/getsentry/sentry-symfony/issues/907


coolguyzone (approved):
Thanks for the fix 🛠️

Issue #12241, opened 2025-01-03
Add custom events


chargome:
### Problem Statement

We currently do not have any custom metrics running on docs. e.g. the number of clicks on our ask ai button.

### Solution Brainstorm

Investigate if we can easily make use of https://plausible.io/docs/custom-event-goals

PR #12380, opened 2025-01-17
feat(platform): Add reusable hook for custom plausible events

chargome:
Adds `usePlausibleEvent` hook which will be used in upcoming tracking tasks

closes https://github.com/getsentry/sentry-docs/issues/12241


a-hariti (approved):
✨

PR #12400, opened 2025-01-20
feat: send content feedback plausible events

### LINKED ISSUE #11363: Ask for content feedback
### Problem Statement

I really like GitHubs' docs where they ask for feedback regarding the content of the page:

Image

### Solution Brainstorm

Take inspiration, maybe we can also integrate User Feedback for that?

a-hariti:
closes #11363


chargome:
Should we maybe move this somewhere into the "Help improve this content section below the navigation buttons?"

a-hariti:
I thought about it, but I think it's more useful to be immediately next to the content, also it's very compact

src/components/docFeedback/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+type Props = {
+  pathname: string;
+};
+
+export function DocFeedback({pathname}: Props) {
+  const {emit} = usePlausibleEvent();
+  const [showFeedbackModal, setShowFeedbackModal] = useState(false);
+  const [feedbackSubmitted, setFeedbackSubmitted] = useState(false);
+
+  const handleFeedback = (helpful: boolean) => {
```
chargome, line 18: We'll need to do something in the UI whenever this is `true` !

src/components/docFeedback/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+        </button>
+      </div>
+
+      <Modal
+        isOpen={showFeedbackModal}
+        onClose={() => {
+          setShowFeedbackModal(false);
+          setFeedbackSubmitted(false);
+        }}
+        title="Help us improve"
```
chargome, line 69: Wdyt about not showing this heading after the form got submitted?

src/components/docFeedback/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+    try {
+      Sentry.captureFeedback(
+        {message: comments},
+        {captureContext: {tags: {page: pathname}}}
+      );
+      setFeedbackSubmitted(true);
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error('Failed to submit feedback:', error);
```
chargome, line 39: Let's handle this error case in the UI somehow and instead of logging this error call `Sentry.captureException(error)`

chargome (commented):
Generally looks good to me, left some comments!

I'll also need to set up the custom plausible event before merging this.

Let's also get feedback from @stephanie-anderson on this one

src/components/docFeedback/index.tsx
a-hariti: explode confetti? 🎊

src/components/docFeedback/index.tsx
a-hariti: or something like "✅ thanks for your feedback"

stephanie-anderson:
General feedback from the preview so far:

## User Feedback
We should definitely provide feedback on each interactive element, as soon as users interact with it. So for example, when someone presses "Yes" or "No", we should let the user know that the system actually did something (which is currently not the case for "Yes"). 

There are other examples out there that work like this preview ([stripe](https://docs.stripe.com/no-code/invoices#automatic-reconciliation), [mintlify](https://mintlify.com/docs/quickstart#publishing)), but they too are lacking a feedback component.

## Placing
We should think about the proper placing of this element again. Was this a conscious decision to put it above the page navigation buttons on the bottom? I'm torn, but I do like how [Next.js is handling it](https://nextjs.org/docs/app/getting-started/data-fetching-and-streaming#api-reference). What I like about it is that this visual element is easily distinguishable from the rest, it's subtle but still eye-catching. And bonus: it offers a range of options.

## Getting Feedback for "Yes" as well
Again something that is really nice about the Next.js docs, they always ask for feedback, regardless of the user's selection. What I also like about it is the subtlety of the textarea. It's not a modal, it's tiny. I think this could positively influence how willingly users are to providing a feedback message. Modal feels like a form they have to fill out. This little box though seems "manageable" and doesn't interrupt the users flow (of setting up Sentry for their app) so much.

Screenshot 2025-01-24 at 13 14 30

## The form is only submittable once
We should make that visible to the user, by no longer showing the voting options. Instead we could let the users know that their feedback has been submitted, and remove the buttons.

## Styles
If we stick with the modal, we need to address the contrast issues for the primary button:

Screenshot 2025-01-24 at 13 07 18

## Text-heaviness
We need to shorten the wording on the modal (if it stays). "Help us improve", "What could we improve" etc is redundant, we should try to make it as sharp and short as possible. Same goes for the confirmation step of the modal.

src/components/docFeedback/index.tsx
chargome: that might be a bit too much 😂

stephanie-anderson:
@a-hariti yes, I'd say let's collect additional feedback on all of the options we offer (even if that's just these two at the moment)

a-hariti:
what's the best way to capture the positive/negative flag along with the user comment?
one of these?
```js
Sentry.setExtra('battery', 0.7);
Sentry.setTag('user_mode', 'admin');
```
@chargome @stephanie-anderson

stephanie-anderson:
Wow, I like the changes you've done, @a-hariti ! 

Additional feedback now is, that I would increase the spacing between the next/previous nav and the feedback a little bit

Screenshot 2025-01-24 at 14 48 34

We could also test with 
* making the question bold(er)
* aligning the border radius to the other visual elements, like the next/previous nav buttons
* aligning the font size of question and options
* maybe remove too much horizontal padding between options, and options/question
* I also played around with emojis, for me personally, the thumbs up/down make most sense (and are the easiest to identify with a smaller font size)

Screenshot 2025-01-24 at 15 01 03

Screenshot 2025-01-24 at 15 04 55

🙏

stephanie-anderson:
> what's the best way to capture the positive/negative flag along with the user comment? one of these?
> 
> ```js
> Sentry.setExtra('battery', 0.7);
> Sentry.setTag('user_mode', 'admin');
> ```
> 
> @chargome @stephanie-anderson

@a-hariti I would go for `setTag`, cause tags are able to be searched for, whereas context (previously `setExtra`) is not

PR #12418, opened 2025-01-22
feat: Drop `success` and `danger` Alert levels

mydea:
We do not use success at all, and danger only in few places, where we can also use `warning` instead.

### DIFF
--- a/docs/contributing/pages/components.mdx
+++ b/docs/contributing/pages/components.mdx
@@ -25,7 +25,7 @@ This is an alert
 Attributes:
 
 - `title` (string)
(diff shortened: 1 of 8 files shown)


chargome:
I would actually keep these tbh, or was this a requirement to drop these with the new design?

chargome (approved):

chargome:
@mydea Just saw that @inventarSarah was making use of "success" in https://github.com/getsentry/sentry-docs/pull/12291, would you mind reverting? 😭

mydea:
I will re-add success here then: https://github.com/getsentry/sentry-docs/pull/12422!

PR #12429, opened 2025-01-22
feat(electron): Make passing framework config more obvious

timfish:
Before it was not immediately obvious where to pass the react config:
- Ref https://github.com/getsentry/sentry-electron/issues/1055#issuecomment-2607314258


coolguyzone:
@AbhiPrasad Please take a look at this one

AbhiPrasad (approved):

PR #12436, opened 2025-01-23
fix(proguard): Add type field for debug images doc

tobias-wilfert:
While looking at the docs I saw that the `type` field for proguard is missing, this PR adds that field to fix the docs.


stefanosiano (approved):
nice catch!
Leaving the relay part as a reference -> https://github.com/getsentry/relay/blob/7e53a2aef13637dcc14bde125cc938c22a94ba38/relay-event-schema/src/protocol/debugmeta.rs#L467-L492

coolguyzone (approved):

PR #12457, opened 2025-01-24
feat: Update `Expandable` style

mydea:
For this, I extracted the `Alert` component stuff into a more generic `Callout` component, which the `Alert` extends.

The `Expandable` can now use the same basis, ensuring we have the same style etc, and just pass some more stuff into it.

For this, the `Callout` component also allows to pass `id` and `titleOnClick` attributes which we need for the Expandable.

## Before

Screenshot 2025-01-24 at 11 19 02
Screenshot 2025-01-24 at 11 19 07

## After

Screenshot 2025-01-24 at 11 19 41

## Permalinks

I made sure that permalinks continue to work. I tweaked the behavior a bit, now it does not scroll into view when you open a permalink expandable, this should be a bit more subtle now. You can try the behavior on any Troubleshoot page in the JS repo, for example.

### DIFF
--- a/src/components/callout/index.tsx
+++ b/src/components/callout/index.tsx
@@ -1,0 +1,25 @@
+import {ForwardRefExoticComponent, ReactNode} from 'react';
+
+// explicitly not usig CSS modules here
+// because there's some prerendered content that depends on these exact class names
+import './styles.scss';
+
+type CalloutProps = {
+  Icon: ForwardRefExoticComponent<any>;
+  children?: ReactNode;
+  level?: 'info' | 'warning' | 'success';
+  role?: string;
+  title?: string;
+};
+
+export function Callout({title, children, level = 'info', Icon, role}: CalloutProps) {
+  return (
+    <div className={`callout ${'callout-' + level}`} role={role}>
+      <Icon className="callout-icon" />
+      <div className="callout-content">
+        {title && <h5 className="callout-header">{title}</h5>}
+        <div className="callout-body content-flush-bottom">{children}</div>
+      </div>
+    </div>
+  );
+}
--- a/src/components/callout/index.tsx
+++ b/src/components/callout/index.tsx
@@ -1,4 +1,4 @@
-import {ForwardRefExoticComponent, ReactNode} from 'react';
+import {ForwardRefExoticComponent, MouseEventHandler, ReactNode} from 'react';
 
 // explicitly not usig CSS modules here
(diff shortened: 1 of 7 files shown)


src/components/callout/styles.scss
```diff
@@ -28,30 +28,46 @@
...
-  color: var(--alert-highlight-color);
+  color: var(--callout-highlight-color);
+
+  &[role="button"] {
+    cursor: pointer;
+  }
+}
+
+.callout-content {
+  min-width: 0;
```
mydea, line 56: Without this, e.g. code boxes can lead to exceeding the width of the content. this should "fix" this.

src/components/callout/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+        onClick={onClick}
+        role={onClick ? 'button' : undefined}
+      >
+        {title}
+      </h5>
+    );
+  }
+
+  // We want to avoid actually triggering the link
+  const wrappedOnClick = onClick
```
chargome, line 40: would move this into a `useCallback`

src/components/expandable.tsx
```diff
@@ -1,86 +1,74 @@
...
-    </h2>
-  ) : (
-    title
-  );
+export function Expandable({title, level, children, permalink}: Props) {
+  const id = permalink ? slugify(title) : undefined;
 
-export function Expandable({title, children, permalink}: Props) {
-  const [isExpanded, setIsExpanded] = useState(false);
+  const defaultIsExpanded = id && window.location.hash === `#${id}`;
```
chargome, line 25: we'll need to move this into an effect to avoid hydration errors

chargome (changes requested):

src/components/callout/index.tsx
mydea: I changed it, better like this?

src/components/expandable.tsx
mydea: moved this around!

chargome (approved):
Looks neat! 💅

stephanie-anderson:
Thanks for taking care of this, @mydea! Could you have a look at the case, when there is special formatting involved, like in [this case](https://sentry-docs-git-fn-expandable.sentry.dev/platforms/javascript/troubleshooting/#missing-module-diagnosticschannel): 

Screenshot 2025-01-24 at 13 26 07


So that its style aligns more with the paragraph styles?

Screenshot 2025-01-24 at 13 27 34

PR #12460, opened 2025-01-24
fix(platform): Enable scrolling via keyboard buttons

chargome:
Enables scrolling with keyboard arrow keys, by nesting the `useListKeyboardNavigate` within search results

### DIFF
--- a/src/components/search/index.tsx
+++ b/src/components/search/index.tsx
@@ -9,18 +9,18 @@ import {
   SentryGlobalSearch,
   standardSDKSlug,
 } from '@sentry-internal/global-search';
-import DOMPurify from 'dompurify';
-import Link from 'next/link';
-import {usePathname, useRouter} from 'next/navigation';
+import {usePathname} from 'next/navigation';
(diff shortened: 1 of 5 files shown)


src/components/search/index.tsx
```diff
@@ -181,9 +181,10 @@ export function Search({path, autoFocus, searchPlatforms = [], showChatBot}: Pro
     []
   );
 
-  const {focused} = useKeyboardNavigate({
+  const {focused} = useListKeyboardNavigate({
     list: flatHits,
     onSelect: hit => router.push(relativizeUrl(hit.url)),
+    disableEventListeners: !inputFocus,
```
andreiborza, line 187: m: I think the condition should be the length of `flatHits > 0` if we are going with a flag approach, but to be honest, I think the best approach is to extract the search results into their own component that uses the hook, no condition and only show the results if we do have hits.

chargome: I'll move it into a component 👍

PR #12467, opened 2025-01-27
ref: Remove `<Note>` component in favor of using `<Alert>`

mydea:
Removes the `<Note>` component in favor of just using `<Alert>`. They have already been exactly the same, so this is just redundant.


inventarSarah:
@mydea should we also remove the Note section in Contributing to Docs at the same time? (https://docs.sentry.io/contributing/pages/components/#note)

mydea:
That should be part of this PR!

inventarSarah:
> That should be part of this PR!

Yes, you are 100% right -- I missed that!

Looks good to me 👍

inventarSarah (approved):

stephanie-anderson (approved):

PR #12475, opened 2025-01-27
feat(js): Update custom trace propagation docs

mydea:
This was brought up by @whitep4nth3r - today, our docs on custom instrumentation are a bit confusing:

1. We have two pages called "Custom Instrumentation", once nested below "Set Up Tracing" and once below "Set Up Tracing > Distributed Tracing"
2. The pages are pretty generic and show a bunch of information that is hard to parse for relevancy.

This PR updates this in a few ways:

1. The pages are renamed to "Custom Span Instrumentation" and "Custom Trace Propagation", respectively, to make it easier in e.g. search to identify what they are about.
2. The custom trace propagation page is updated:
  - We show specifically correct data now. Meaning, we do not show how trace propagation is set up in other SDKs, but specifically if in your SDK it is set up automatically. This includes updating phrasing to indicate that in all node SDKs, distributed tracing is enabled by default.
  - Explain _what_ is auto-instrumented - namely, only outgoing http requests.
  - Updates an example code snippet to use `Sentry.getTraceData()` which is actually the preferred way to manually instrument stuff today.
  - Restructure & split up stuff a bit to hopefully make it a bit easier to follow the content structure - this was quite heavily nested and split into platform category sections, making this hard to parse. Hopefully it is easier to understand now.

### DIFF
--- a/platform-includes/distributed-tracing/custom-instrumentation/server/javascript.mdx
+++ b/platform-includes/distributed-tracing/custom-instrumentation/server/javascript.mdx
@@ -28,10 +28,7 @@ Here's an example of how to extract and store incoming tracing information using
 const http = require("http");
 
 http.createServer((request, response) => {
-  const sentryTraceHeaders = request.headers["sentry-trace"];
-  const sentryTrace = Array.isArray(sentryTraceHeaders)
-    ? sentryTraceHeaders.join(",")
-    : sentryTraceHeaders;
+  const sentryTrace = request.headers["sentry-trace"];
   const baggage = request.headers["baggage"];
 
   Sentry.continueTrace({ sentryTrace, baggage }, () => {


platform-includes/distributed-tracing/custom-instrumentation/javascript.mdx
```diff
@@ -1,174 +1,15 @@
...
-    // do something
-  }
-);
-```
-
-In this example, we create a new root span that is attached to the trace specified in the `sentry-trace` and `baggage` HTML `<meta>` tags.
+{/* The first section should only apply for browser-only platforms, not for Meta-Frameworks - so we include browser but then exclude server/serverless platforms */}
 
+<PlatformCategorySection supported={["browser"]}>
+  <PlatformCategorySection notSupported={["server", "serverless"]}>
```
chargome, line 6: `<PlatformCategorySection supported={["browser"]} notSupported={["server", "serverless"]}>`

does not work?

chargome (approved):

platform-includes/distributed-tracing/custom-instrumentation/server/javascript.mdx
```diff
@@ -0,0 +1,100 @@
+Trace propagation is set up automatically in this SDK. By default, outgoing HTTP requests will automatically be instrumented for you.
+For other cases where you may want to continue traces (e.g. in a job queue or when working with web sockets), you can manually extract and inject tracing information.
```
coolguyzone, line 2:
```suggestion
For other cases where you may want to continue traces (for example, in a job queue or when working with web sockets), you can manually extract and inject tracing information.
```

platform-includes/distributed-tracing/custom-instrumentation/server/javascript.mdx
```diff
@@ -0,0 +1,100 @@
...
+        // Your API code...
+      }
+    );
+  });
+});
+```
+
+In this example, we create a new transaction that is attached to the trace specified in the `sentry-trace` and `baggage` headers.
+
+### Injecting Tracing Information into HTML
```
coolguyzone, line 61:
```suggestion
### Injecting Tracing Information Into HTML
```

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+  // If this is 0, tracing is disabled
+  // but distributed tracing is not
+  tracesSampleRate: 0,
+});
+```
+
+### Continuing a Trace Manually
+
+By default, the `browserTracingIntegration` will automatically continue a trace found in a `<meta>` tags - see <PlatformLink to="/tracing/trace-propagation/#automatic-trace-propagation">Automatic Trace Propagation</PlatformLink> for details.
```
coolguyzone, line 25:
```suggestion
By default, the `browserTracingIntegration` will automatically continue a trace found in a `<meta>` tag - see <PlatformLink to="/tracing/trace-propagation/#automatic-trace-propagation">Automatic Trace Propagation</PlatformLink> for details.
```

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
```diff
@@ -0,0 +1,98 @@
...
+If you want to continue a different trace, for example because you cannot propagate the trace through meta tags but through some different mechanism, you can do this as follows:
```
coolguyzone, line 26:
```suggestion
If you want to continue a different trace, for example because you cannot propagate the trace through meta tags but through some different mechanism, you can do this:
```

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
```diff
@@ -0,0 +1,98 @@
...
+Sentry.startBrowserTracingPageLoadSpan(
+  client,
+  {
+    name: window.location.pathname,
+  },
+  { sentryTrace, baggage }
+);
+```
+
+## Configuring Distributed Tracing without `browserTracingIntegration`
```
coolguyzone, line 52:
```suggestion
## Configuring Distributed Tracing Without `browserTracingIntegration`
```

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
```diff
@@ -0,0 +1,98 @@
...
+If you don't want to use the `browserTracingIntegration` integration, you can manually extract and inject tracing data in your application to connect multiple systems. For this, you must:
+
+- Extract and store incoming tracing information from HTML `<meta>` tags when loading the page.
+- Inject tracing information to any outgoing requests.
+
+To learn more about distributed tracing, see our <PlatformLink to="/tracing/trace-propagation/">Distributed Tracing</PlatformLink> docs.
+
+### Extract Tracing Information From HTML `meta` Tags
+
+If you have a server that renders your application's HTML (SSR) and is also running a Sentry SDK, you can connect your backend to your backend via tracing.
```
coolguyzone, line 63:
```suggestion
If you have a server that renders your application's HTML (server-side rendering) and is also running a Sentry SDK, you can connect your backend to your backend via tracing.
```

platform-includes/distributed-tracing/custom-instrumentation/javascript.mdx
mydea: ohh that's so much better! thanks!

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
```diff
@@ -0,0 +1,98 @@
+
+Distributed tracing will be set up automatically if you've <PlatformLink to="/tracing/">Set Up Tracing</PlatformLink>.
+
+If you have not enabled tracing, or you need to propagate traces in other places than outgoing HTTP requests, you can manually set up your application for distributed tracing to work.
```
whitep4nth3r, line 4: I'm sorry I have no idea what this sentence is trying to communicate.

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+## Enabling Automatic Distributed Tracing
+
+To enable distributed tracing for your frontend, add `browserTracingIntegration` to your `Sentry.init()` options as described in the <PlatformLink to="/tracing/instrumentation/automatic-instrumentation/">Automatic Instrumentation</PlatformLink> docs.
+
+If you want to use distributed tracing but not tracing, set the `tracesSampleRate` option to `0`.
```
whitep4nth3r, line 10: `If you want to use distributed tracing but not tracing`- this is so confusing. We need to explain this in more detail. Even the SDK option doesn't make much sense, because you've set something to 0 yet it is kind of working in a way because of the integration... it's hurting my head.

platform-includes/distributed-tracing/custom-instrumentation/server/javascript.mdx
```diff
@@ -0,0 +1,100 @@
+Trace propagation is set up automatically in this SDK. By default, outgoing HTTP requests will automatically be instrumented for you.
+For other cases where you may want to continue traces (for example, in a job queue or when working with web sockets), you can manually extract and inject tracing information.
+
+If you want to use automated distributed tracing but not tracing, set the `tracesSampleRate` option to `0`.
```
whitep4nth3r, line 4: I know this is the same as the other page where I called out this comment, but this needs to be explained better here, too.

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
```diff
@@ -0,0 +1,98 @@
+
+Distributed tracing will be set up automatically if you've <PlatformLink to="/tracing/">Set Up Tracing</PlatformLink>.
```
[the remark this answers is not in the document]
mydea: I think the idea here was to match the nav label, but I agree it does not really make sense - will change it!

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
mydea: I rewrote it a bit, is it clearer now?

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
mydea: Rewrote this a bit, is it clearer now?

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+When tracing is enabled, the Sentry SDK will not only create spans for pageloads and transactions but also propagate trace information to other systems. This allows you to connect multiple systems and see how they interact with each other.
+
+If you have not enabled tracing, or you need to propagate traces in places other than outgoing HTTP requests (e.g. when working with websockers), you can manually set up your application for distributed tracing to work.
+
+## Enabling Automatic Distributed Tracing
+
+To enable distributed tracing for your frontend, add `browserTracingIntegration` to your `Sentry.init()` options as described in the <PlatformLink to="/tracing/instrumentation/automatic-instrumentation/">Automatic Instrumentation</PlatformLink> docs.
+
+If you want to use distributed tracing without creating spans, set the `tracesSampleRate` option to `0`.
```
whitep4nth3r, line 12: Why would I want to do this? Maybe I completely misunderstand the whole concept of tracing... but how can you trace something without creating spans? How does it look in Sentry?

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
whitep4nth3r: It's clearer, yes, but I still have more questions :D added a comment

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
whitep4nth3r: So I'm trying to understand this logic.

IF I have not enabled tracing (so I have not set tracesSampleRate)
OR if I need to propagate traces in other places that are NOT HTTP requests (e.g. websocket events, I guess?)
THEN I can do some manual setup to make distributed tracing work

BUT

Why would I want to manually setup distributed tracing, if I have not enabled tracing???
Why would I not enable tracing?
How does distributed tracing work without creating spans?

I'm sorry I must sound so stupid but I think it would be useful to provide some examples here because apart from the OR clause above, I don't know why you would consider the first IF?

Maybe this stuff is covered elsewhere on the docs so maybe it is worth linking to but it just seems so overly complex and counterintuitive to me.

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
```diff
@@ -1,13 +1,15 @@
 
-Distributed tracing will be set up automatically if you've <PlatformLink to="/tracing/">Set Up Tracing</PlatformLink>.
+Distributed tracing will be set up automatically if you've <PlatformLink to="/tracing/">set up tracing</PlatformLink>.
 
-If you have not enabled tracing, or you need to propagate traces in other places than outgoing HTTP requests, you can manually set up your application for distributed tracing to work.
+When tracing is enabled, the Sentry SDK will not only create spans for pageloads and transactions but also propagate trace information to other systems. This allows you to connect multiple systems and see how they interact with each other.
+
+If you have not enabled tracing, or you need to propagate traces in places other than outgoing HTTP requests (e.g. when working with websockers), you can manually set up your application for distributed tracing to work.
```
whitep4nth3r, line 6: So I'm trying to understand this logic.

IF I have not enabled tracing (so I have not set tracesSampleRate)
OR if I need to propagate traces in other places that are NOT HTTP requests (e.g. websocket events, I guess?)
THEN I can do some manual setup to make distributed tracing work

BUT

Why would I want to manually setup distributed tracing, if I have NOT enabled tracing???
Why would I not enable tracing? Is this because I don't want to propagate trace information to other systems?
How does distributed tracing work without creating spans?

Maybe this stuff is covered elsewhere on the docs so maybe it is worth linking to but it just seems so overly complex and counterintuitive to me. Or maybe the information in the two small paragraphs needs to be combined into one to provide more concrete use cases of each situation.

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
mydea: We have this page that tries to explain the concept of distributed tracing: https://docs.sentry.io/platforms/javascript/guides/react/tracing/trace-propagation/#what-is-distributed-tracing - is that helpful? It's a bit hard to describe/explain this here again without exceeding the scope of it considerably 😬 I can try to put an expandable in there!

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
mydea: the only reason to not enable tracing here is to save cost - I do not want to pay for spans. I am not sure how exactly we want to point this out here. Generally we _want_ people to just enable tracing.

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
whitep4nth3r: Nope sorry, that page doesn't answer why I would want to NOT create spans and still see distributed traces in Sentry. I don't understand.

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
whitep4nth3r: Ah ha, I figured this was the case. But the point still stands. How does distributed tracing work without spans if everything is based on spans? It just seems impossible. And maybe we don't even need to go into it, so maybe we don't even call this out, and just talk about tracing in other places that are not HTTP-based?

mydea: So I adjusted this some more, after talking with others on the team!

1. I added an expandable that explains what distributed tracing is, for context.
2. I removed the `tracesSampleRate: 0` mentions as that is not even necessary anymore. this should hopefully improve clarity a bit...

whitep4nth3r: This is much clearer now, thank you for taking the time to do this!

whitep4nth3r (approved):

platform-includes/distributed-tracing/custom-instrumentation/browser/javascript.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+### Extract Tracing Information From HTML `meta` Tags
+
+If you have a server that renders your application's HTML (server-side rendering) and is also running a Sentry SDK, you can connect your backend to your backend via tracing.
+
+To do this, have your server render HTML `<meta>` tags with the Sentry trace information. In your frontend, extract that tracing information when the page is loading and use it to create new transactions connected to that incoming backend trace.
+
+Some Sentry backend SDKs provide a built-in way to inject these `<meta>` tags into rendered HTML. For example:
+
+- [Python](/platforms/python/distributed-tracing/custom-instrumentation/#inject-tracing-information-into-rendered-html)
```
Lms24, line 61: l/m: I'd add the link to Node docs as well here
```suggestion
- [Node](/platforms/javascript/guides/node/tracing/trace-propagation/custom-instrumentation/#injecting-tracing-information-into-html)
- [Python](/platforms/python/distributed-tracing/custom-instrumentation/#inject-tracing-information-into-rendered-html)
```

platform-includes/distributed-tracing/custom-instrumentation/server/javascript.mdx
```diff
@@ -0,0 +1,92 @@
...
+- You also can pick up tracing information from environment variables.
+
+Here's an example of how to extract and store incoming tracing information using `continueTrace()`:
+
+```javascript
+const http = require("http");
+
+http.createServer((request, response) => {
+  const sentryTraceHeaders = request.headers["sentry-trace"];
+  const sentryTrace = Array.isArray(sentryTraceHeaders)
```
Lms24, line 32: m: I don't think joining multiple sentry-trace headers by `,` is [valid](https://github.com/getsentry/sentry-javascript/blob/1d797ea3ee4653ccea20cb217195a79766689491/packages/core/src/utils-hoist/tracing.ts/#L8-L14). I guess this could happen for _reasons_ but is it worth pointing this out in the docs? If at all, we should probably just take one? Or any chance this got mixed up with `baggage`? baggage arrays could happen I guess if multiple instrumentation libraries add baggage headers 🤔

Lms24 (commented):
This is a nice improvement, thanks a lot for tackling it! Two things we should address before merging.

Lms24 (approved):
thanks!

PR #12476, opened 2025-01-27
ref: Drop remaining usage of `<Note>`

mydea:
Hmm, it seems https://github.com/getsentry/sentry-docs/pull/12467 did not catch everything. Probably some search-replace stuff. Now, it should be fully gone, and also removes the actual component code itself!


coolguyzone (approved):
My scroll finger is gonna need a vacation after this one 💪  Thanks for fixing!

PR #12477, opened 2025-01-27
search: respect the stored platform on non platform pages

a-hariti:
currently, search relies on the page pathname to prioritize sdks/platform search results

but when the user navigates to a non platform page while they have some platform previously selected on the sidebar,
search ignores that platform

this PR makes search respect that selection


chargome (commented):
Could you update the description on what this PR does exactly?

chargome (approved):

PR #12502, opened 2025-01-29
feat(js): Add SDK Options page

mydea:
After talking some about this with @lforst and @Lms24, and triggered by talks with @inventarSarah, I decided to make a POC of a new SDK Options page for the JS SDKs.

This was borne out of the following thoughts:

* Today, we have a page `Configuration > Basic Options`, which is a bit weird. First, it is pretty generic (originally applying to all SDKs, so containing a lot of generic language). Second, it is not fully up to date. And lastly, it is not really complete - some (many) options are described but not all. 
* It can be hard to actually find the available options.
* It is hard for SDK maintainers to know where to document things.
* The location/naming of this as "Configuration > Basic Options" is not super intutitive.

To improve on this, this PR introduces (for JS only, for now) a new top-level "SDK Options" nav item. This is a single page that contains _all_ top-level SDK options that can be passed to `Sentry.init()`. There are a few goals here:

1. Make it complete - this should include all top-level options we want to document. Including replay, profiling, ... It does not literally have to be complete, we can choose to omit options we do not want to document.
2. Make it consistent - ensure all options have a consistent formatting, making it easy to read & author new options.
3. Add an index that can make it easier to scan this. Ideally, auto-generated.
4. Call out that certain options only apply to client/server SDKs

For this, I wrote two new components, a `<SdkOption>` component, and a `<SdkOptionOverview>` which just auto-generates an index for this. You can put headings and text between those for sectioning and explanation.

I also removed the old Basic options page as well as the transports page and folded that into the new page, adding redirects.

Some screenshots showing the different things:

Auto-generated Overview:
image
 
Option with env-var configuration:
image
Note that the `env var` entry will not be shown in environments where this is not auto-picked up.

Option with default value:

image

If in a meta-framework, options that are only available in one part or the other will be called out:
image

We can add more things to the SDK options if needed. The authoring experience with this should be pretty straightforward and make it easy to keep this up-to-date, hopefully.

## Follow up

If this generally finds agreement, we can merge this. We should then make some passes over the actual options and make sure we add ones that have been forgotten, and/or improve docs on them where applicable. In some places, we have docs for options in more specific places, which should be folded into this page where possible.

When this is done, we would like to also add a top-level `SDK APIs` page, which does the same thing but for APIs, e.g. `Sentry.captureException()`, `Sentry.setTag()` etc. We can discuss this more separately. 

The idea is that with these two pages, we'll be able to consolidate a good amount of the pages we currently have in various levels of nesting, and only keep additional pages for places where more explanation/examples are needed.


src/components/platformCategorySection.tsx
```diff
@@ -1,25 +1,26 @@
 import {getCurrentPlatformOrGuide} from 'sentry-docs/docTree';
 import {serverContext} from 'sentry-docs/serverContext';
-import {Platform, PlatformGuide} from 'sentry-docs/types';
+import {Platform, PlatformCategory, PlatformGuide} from 'sentry-docs/types';
 
 type Props = {
   children?: React.ReactNode;
   noGuides?: boolean;
-  notSupported?: string[];
+  notSupported?: `${PlatformCategory}`[];
```
mydea, line 8: Actually I wonder if we can remove that enum and just use string literals there overall, could not find a place where this was really used that way, but not sure... cc @chargome

docs/platforms/javascript/common/options.mdx
```diff
@@ -0,0 +1,454 @@
...
+
+<SdkOption name="includeLocalVariables" type='boolean' defaultValue='false'  categorySupported={['server', 'serverless']}>
+
+Set this option to `true` to add stack local variables to stack traces.
+
+For more advanced configuration options, see the documentation on the <PlatformLink to="/configuration/integrations/localvariables">Local Variables integration options</PlatformLink>.
+
+</SdkOption>
+
+<SdkOption name="integrations" type='Array<Integration>' defaultValue='[]'>
```
Lms24, line 194:
```suggestion
<SdkOption name="integrations" type='Array<Integration> | (integrations: Array<Integration>) => Array<Integration>' defaultValue='[]'>
```

docs/platforms/javascript/common/options.mdx
```diff
@@ -0,0 +1,454 @@
...
+<PlatformContent includePath="configuration/config-intro" />
+
+## Available Options
+
+<SdkOptionsOverview />
+
+## Core Options
+
+<SdkOption name='dsn' type='string' envVar='SENTRY_DSN'>
+The _DSN_ tells the SDK where to send the events. If this value is not provided, the SDK will try to read it from the `SENTRY_DSN` environment variable. If that variable also does not exist, the SDK will just not send any events.
```
Lms24, line 17: since we talked about env vars in the morning: We should change this description for client vs server

docs/platforms/javascript/common/options.mdx
```diff
@@ -0,0 +1,454 @@
...
+
+In runtimes without a process environment (such as the browser) that fallback does not apply.
+
+Learn more about [DSN utilization](/product/sentry-basics/dsn-explainer/#dsn-utilization).
+
+</SdkOption>
+
+<SdkOption name="debug" type='boolean' defaultValue='false'>
+
+Turns debug mode on or off. If debug is enabled SDK will attempt to print out useful debugging information if something goes wrong with sending the event. The default is always `false`. It's generally not recommended to turn it on in production, though turning `debug` mode on will not cause any safety concerns.
```
Lms24, line 27: safety sounds weird here 🤔 either we say "security" or write something like it won't expose any critical information? Or just omit it?

```suggestion
Turns debug mode on or off. If debug is enabled SDK will attempt to print out useful debugging information if something goes wrong with sending the event. The default is always `false`. It's generally not recommended to turn it on in production, though turning `debug` mode on will not cause any security concerns.
```

docs/platforms/javascript/common/options.mdx
```diff
@@ -0,0 +1,454 @@
...
+
+<SdkOption name="maxBreadcrumbs" type='number' defaultValue="100">
+
+This variable controls the total amount of breadcrumbs that should be captured. You should be aware that Sentry has a [maximum payload size](https://develop.sentry.dev/sdk/data-model/envelopes/#size-limits) and any events exceeding that payload size will be dropped.
+
+</SdkOption>
+
+<SdkOption name="attachStacktrace" type='boolean' defaultValue='false'>
+
+When enabled, stack traces are automatically attached to all messages logged. Stack traces are always attached to exceptions; however, when this option is set, stack traces are also sent with messages. This option, for instance, means that stack traces appear next to all log messages.
```
Lms24, line 84: l: maybe this makes it a bit clearer?

```suggestion
When enabled, stack traces are automatically attached to all messages logged. Stack traces are always attached to exceptions; however, when this option is set, stack traces are also sent with messages. This option, for instance, means that stack traces appear next to all messages captured with `Sentry.captureMessage()`.
```

docs/platforms/javascript/common/options.mdx
```diff
@@ -0,0 +1,454 @@
...
+- `name`: The name of the span as it was started.
+- `attributes`: The initial attributes of the span.
+  <PlatformCategorySection supported={["server", "serverless"]}>
+    - `normalizedRequest`: Request information of the currently active HTTP
+    server request, if applicable.
+  </PlatformCategorySection>
+
+</SdkOption>
+
+<SdkOption name="tracePropagationTargets" type='Array<string | RegExp>'>
```
Lms24, line 335: I'd also add a default value entry into the table. Which needs to be different on browser and server.

docs/platforms/javascript/common/options.mdx
```diff
@@ -0,0 +1,454 @@
...
+    return scope;
+  },
+});
+```
+
+</SdkOption>
+
+<SdkOption name="maxValueLength" type='number' defaultValue='250'>
+
+Maximum number of characters a single value can have before it will be truncated.
```
Lms24, line 158: l: probably a bit clearer? (feel free to disregard)

```suggestion
Maximum number of characters a every string property on events sent to Sentry can have before it will be truncated.
```

docs/platforms/javascript/common/options.mdx
```diff
@@ -0,0 +1,454 @@
...
+See <PlatformLink to="/configuration/integrations/#modifying-default-integrations">integration docs</PlatformLink> to see how you can modify the default integrations.
+
+</SdkOption>
+
+<SdkOption name="transport" type="(transportOptions: TransportOptions) => Transport">
+
+The JavaScript SDK uses a transport to send events to Sentry. On modern browsers, most transports use the browsers' fetch API to send events. Transports will drop an event if it fails to send due to a lack of connection.
+
+<PlatformCategorySection supported={["browser"]}>
+  In the browser, a fetch-based transport is used by default.
```
Lms24, line 215: l:
```suggestion
  In the browser, a `fetch`-based transport is used by default.
```

docs/platforms/javascript/common/options.mdx
```diff
@@ -0,0 +1,454 @@
...
+  `caCerts`: A path or list of paths to a CA certificate, or a buffer of CA
+  certificates. - `httpModule`: A custom HTTP module to use for requests.
+  Defaults to the the native `http` and `https` modules. - `keepAlive`:
+  Determines whether to keep the socket alive between requests. Defaults to
+  `false`.
+</PlatformCategorySection>
+
+<PlatformCategorySection supported={["browser", "desktop"]}>
+  - `headers`: An object containing headers to be sent with every request. Used
+  by the SDK's fetch and XHR transports. - `fetchOptions`: An object containing
```
Lms24, line 240: l:
```suggestion
  by the SDK's `fetch` and XHR transports. - `fetchOptions`: An object containing
```

Lms24, line 240: are we intentionally not putting the individual transport options into their own lines?

Lms24 (commented):
I really like this rewrite! IMHO, this is much cleaner than the previous "Basic Options" page, both in description as well as UI. Love the table with the type and default value information. 
I read through the options (haven't done this in a while) and had some minor suggestions. Once this is out of draft state, I'm happy to ✅ !

src/components/sdkOption.tsx
```diff
@@ -0,0 +1,135 @@
...
+              <path
+                d="M9.199 13.599a5.99 5.99 0 0 0 3.949 2.345 5.987 5.987 0 0 0 5.105-1.702l2.995-2.994a5.992 5.992 0 0 0 1.695-4.285 5.976 5.976 0 0 0-1.831-4.211 5.99 5.99 0 0 0-6.431-1.242 6.003 6.003 0 0 0-1.905 1.24l-1.731 1.721a.999.999 0 1 0 1.41 1.418l1.709-1.699a3.985 3.985 0 0 1 2.761-1.123 3.975 3.975 0 0 1 2.799 1.122 3.997 3.997 0 0 1 .111 5.644l-3.005 3.006a3.982 3.982 0 0 1-3.395 1.126 3.987 3.987 0 0 1-2.632-1.563A1 1 0 0 0 9.201 13.6zm5.602-3.198a5.99 5.99 0 0 0-3.949-2.345 5.987 5.987 0 0 0-5.105 1.702l-2.995 2.994a5.992 5.992 0 0 0-1.695 4.285 5.976 5.976 0 0 0 1.831 4.211 5.99 5.99 0 0 0 6.431 1.242 6.003 6.003 0 0 0 1.905-1.24l1.723-1.723a.999.999 0 1 0-1.414-1.414L9.836 19.81a3.985 3.985 0 0 1-2.761 1.123 3.975 3.975 0 0 1-2.799-1.122 3.997 3.997 0 0 1-.111-5.644l3.005-3.006a3.982 3.982 0 0 1 3.395-1.126 3.987 3.987 0 0 1 2.632 1.563 1 1 0 0 0 1.602-1.198z"
+                fill="currentColor"
+              />
+            </svg>
+            {name}
+          </a>
+        </h3>
+
+        <table className="w-auto">
```
chargome, line 56: WDYT about wrapping this table with a border radius? (same value as the alert)

src/components/platformCategorySection.tsx
chargome: Yeah getting rid of enums per se I think is a good idea, also could't spot anything where this would be actually used

src/components/sdkOption.tsx
```diff
@@ -0,0 +1,135 @@
...
+                      <code>{envVar}</code>
+                    </td>
+                  </tr>
+                )}
+              </PlatformSection>
+            </PlatformCategorySection>
+
+            {showBrowserOnly && (
+              <tr>
+                <th>Only available in</th>
```
chargome, line 89:
```suggestion
                <th>Only available on</th>
```

src/components/sdkOption.tsx
```diff
@@ -0,0 +1,135 @@
...
+                <td>Client</td>
+              </tr>
+            )}
+
+            {showServerLikeOnly && (
+              <tr>
+                <th>Only available in</th>
```
chargome, line 96:
```suggestion
                <th>Only available on</th>
```

Not 100% sure though lol

src/middleware.ts
```diff
@@ -925,6 +925,22 @@ const USER_DOCS_REDIRECTS: Redirect[] = [
...
+    to: '/platforms/javascript/guides/:guide/options/',
+  },
+  {
+    from: '/platforms/javascript/configuration/environments/',
+    to: '/platforms/javascript/options/#environment',
+  },
+  {
+    from: '/platforms/javascript/guides/:guide/configuration/environments/',
+    to: '/platforms/javascript/guides/:guide/options/#environment',
+  },
```
chargome, line 943: these need to go into `redirects.js` – middleware only works for static redirects!

src/components/sdkOptionOverview.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const [treeItems, setTreeItems] = useState<TreeItem[]>([]);
+
+  // gather the sdk option items on mount
+  useEffect(() => {
+    if (typeof document === 'undefined') {
+      return;
+    }
+    const main = document.getElementById('main');
+    if (!main) {
+      throw new Error('#main element not found');
```
chargome, line 28: Would not throw this one in a client component, any way to handle this in the ui?

platform-includes/configuration/config-intro/javascript.mdx
```diff
@@ -1,10 +1,8 @@
 Options are passed to the `init()` function as object:
```
chargome, line 1:
```suggestion
Options are passed to the `init()` function as an object:
```

Not your change but I think it reads better

src/components/sdkOptionOverview.tsx
```diff
@@ -0,0 +1,92 @@
...
+};
+
+type TreeNode = {
+  element: HTMLElement;
+  id: string;
+  level: number;
+  name: string;
+};
+
+export function SdkOptionsOverview() {
```
chargome, line 18: nit: rename either filename or component to have the same same

src/components/sdkOptionOverview.tsx
```diff
@@ -0,0 +1,92 @@
...
+    setTreeItems(_tocItems.filter(item => item.children.length > 0));
+  }, []);
+
+  return (
+    <ul>
+      {treeItems.map(item => {
+        return (
+          <li key={item.id}>
+            <a href={`#${item.id}`}>{item.name}</a>
+            <ul>
```
chargome, line 78: Wondering if we should move that into a collapsible so the list does not take too much space here?

src/components/sdkOptionOverview.tsx
```diff
@@ -0,0 +1,92 @@
...
+          id: node.id,
+          name: node.name,
+          children: [],
+        });
+      }
+    }
+
+    // Remove groups without children
+    setTreeItems(_tocItems.filter(item => item.children.length > 0));
+  }, []);
```
chargome, line 70: Just generally wondering: Doing this on render will cause some layout shift and pending UI state, would there be a way to get these options on the server during build time?

chargome (commented):
Nice change!

inventarSarah:
That looks much clearer!
How are these options currently sorted? Would it make sense to sort them alphabetically to make them easier to find?

docs/platforms/javascript/common/options.mdx
mydea: maybe we should just remove the last sentence here, not sure how relevant this is actually..?

src/components/sdkOptionOverview.tsx
mydea: I just copied this from the table of contents component xD We can probably also just not throw here at all and simply handle this as "no table of contents"!

src/components/sdkOptionOverview.tsx
mydea: IMHO the list is pretty nice, it's the only place where you can really scan options, we do not have another good place for this as of now?

mydea:
> That looks much clearer! How are these options currently sorted? Would it make sense to sort them alphabetically to make them easier to find?

I've thought about this too some, IMHO it makes sense to have them manually sorted - right now it is just the order that you put them into the mdx file! Once we have general agreement on the direction etc. I would take some time to reorder these better, for now I mostly just cobbled it together in the order it already was in basic options.

src/components/sdkOptionOverview.tsx
chargome: Let's keep it then, just found it to be a bit long but no hard opinion.

src/components/sdkOption.tsx
chargome: Hmm tricky, the `prose` class also makes this harder. Looks better with 1px imo and we need to adapt it to dark mode – I can do that tomorrow if you want!

mydea: some more hacking around, figured out a way to make 1px borders work!

docs/platforms/javascript/common/options.mdx
mydea: yeah this is a bit tricky because 1. it is hard to implement this nicely (we can, if we want to) and especially 2. the default value here is not an actual value but computed (e.g. based on the current host), which is why I opted to rather explain the default in this case?

docs/platforms/javascript/common/options.mdx
mydea: no! auto-format messed this up, good catch!

mydea:
I think this is ready for review now, I tried to address the things that have already been brought up. The options list itself for sure can use some more iteration and work, but IMHO we can do that step by step then too.

src/components/sdkOption/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+  defaultValue,
+  envVar,
+  categorySupported = [],
+}: Props) {
+  // This is a literal copypaste of the HTML Gatsby outputs for regular
+  // Markdown headings because we can't figure out how to make Gatsby
+  // render component content like regular markdown/MDX content. We tried
+  // MDXRenderer but that one needs "compiled MDX" which we are unable to
+  // obtain. We might get one if we inserted a graphql node somewhere but
+  // who knows at this point.
```
chargome, line 33:
```suggestion
```

coolguyzone (approved):
Thanks @mydea! I don't see any issues here, however since a lot of these changes are on the more technical side, please also make sure a docs engineer approves as well before merging.

src/components/platformCategorySection.tsx
mydea: I actually extracted this out a bit here: https://github.com/getsentry/sentry-docs/pull/12558

chargome (approved):
Thanks for the update, let's ship it and maybe check if can render the ToC during build time in a follow up PR 👍

docs/platforms/javascript/common/options.mdx
Lms24: idea (logaf-l) should we then add something liks "see description" for the default value field?

Lms24 (approved):

stephanie-anderson (approved):

mydea:
after talking with @stephanie-anderson , for now we leave it under configuration > options!

PR #12571, opened 2025-02-04
Fix GitHub link

RoSk0:


aldy505 (approved):

aldy505:
cc @hubertdeng123 @BYK

develop-docs/self-hosted/index.mdx
```diff
@@ -217,6 +217,6 @@ Keep in mind that all this setup uses single-nodes for all services, including K
...
+Sentry is [Fair Source](https://fair.io/) under the [FSL (Functional Source License)](https://fsl.software/). Note that Fair Source is not under the [OSI](https://opensource.org/) umbrella, it becomes literally open source under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0) after 2 years, and is practically open source before that except when you are a Sentry competitor. This license means that users can use Sentry and deploy it anywhere (even inside an enterprise ecosystem), but users are prohibited to sell deployed self-hosted Sentry as any kind of offering (SaaS or any other business model), and users are prohibited to be a direct competitor of Sentry by using Sentry's code that is FSL-licensed. The source code of Sentry and its surrounding service & libraries for self-hosted deployment are [available on GitHub](https://github.com/search?q=topic%3Atag-production+org%3Agetsentry&type=Repositories).
```
BYK, line 220: Good catch, however I don't think that the new link is correct either. I think this is a lot closer to what was intended in the original text: https://github.com/search?q=org%3Agetsentry+content%3A%22Functional+Source+License%2C+Version+1.1%2C+Apache+2.0+Future+License%22&type=code

Wdyt @aldy505?

BYK (changes requested):

develop-docs/self-hosted/index.mdx
aldy505: @BYK your link doesn't work either 😂

image

BYK: Ugh, must be something tied to the state of the account then. Maybe we should just drop the link at this point?

sfanahata: Shall we drop this link and merge this PR?

aldy505: > Shall we drop this link and merge this PR? 

Yes

sfanahata:
```suggestion
Sentry is [Fair Source](https://fair.io/) under the [FSL (Functional Source License)](https://fsl.software/). Note that Fair Source is not under the [OSI](https://opensource.org/) umbrella, it becomes literally open source under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0) after 2 years, and is practically open source before that except when you are a Sentry competitor. This license means that users can use Sentry and deploy it anywhere (even inside an enterprise ecosystem), but users are prohibited to sell deployed self-hosted Sentry as any kind of offering (SaaS or any other business model), and users are prohibited to be a direct competitor of Sentry by using Sentry's code that is FSL-licensed. The source code of Sentry and its surrounding service & libraries for self-hosted deployment are available on GitHub.
```

sfanahata (approved):
Good to go.

sfanahata:
@BYK any concerns about merging?

BYK (approved):
Sorry for the delay, looks good now!

PR #12585, opened 2025-02-05
fix(.NET): Removed `enable-tracing` from options

bitsandfoxes:
The .NET SDK removed the option `enable tracing` [here](https://github.com/getsentry/sentry-dotnet/pull/3569) that has been part of the `5.0.0` release.


stephanie-anderson (approved):
thank you @bitsandfoxes !

PR #12592, opened 2025-02-05
docs: Clarify `sample_rand` specs

szokeasaurusrex:
Clarify the docs on how SDKs should set and propagate the `sample_rand` value. [Rendered preview](https://develop-docs-git-szokeasaurusrex-samplerand-clarification.sentry.dev/sdk/telemetry/traces/#propagated-random-value).

### DIFF
--- a/develop-docs/sdk/telemetry/traces/index.mdx
+++ b/develop-docs/sdk/telemetry/traces/index.mdx
@@ -231,16 +231,16 @@ To improve the likelihood of capturing complete traces when backend services use
 SDKs must set and propagate the `sample_rand` as follows:
 
 1. Whenever a new trace is started, including in the Tracing without Performance case, the SDK generates a `sample_rand` value for the **entire trace**.
-   1. SDKs must generate the `sample_rand` by pseudo-randomly sampling a floating-point number from the uniform distribution over the range `[0.0, 1.0)` (including `0.0`, excluding `1.0`).
-   1. The exact method for pseudo-random sampling is implementation-defined, and may therefore vary between SDK implementations.
+   1. SDKs must generate the `sample_rand` by randomly picking a floating-point number in the range `[0.0, 1.0)` (including `0.0`, excluding `1.0`).
    1. We _recommend_ that SDKs generate the random number deterministically by using the trace ID to seed the random number generator. If multiple threads share the same random instance, seed the random number generator atomically, so that seeds do not get shared accross threads.
-1. The `sample_rand` is part of the Dynamic Sampling Context (DSC). Therefore, SDKs must propagate the `sample_rand` to downstream SDKs by adding it to the `baggage` header.
+   1. The exact method for (pseudo-)random sampling is implementation-defined, and may therefore vary between SDK implementations.
+1. The `sample_rand` is part of the Dynamic Sampling Context (DSC), which is propagated to downstream SDKs.
    1. In particular, SDKs must propagate `sample_rand` even when the sampling decision is deferred, e.g. in the Tracing without Performance case.
-   1. The `sample_rand` value must be sent in the baggage header's `sentry-sample_rand` field. The value must be expressed in decimal notation (e.g. `0.1234`).
-1. On incoming traces, SDKs must use the incoming `sentry-sample_rand` value. They also must propagate the incoming `sample_rand` value to any downstream SDKs.
+   1. The `sample_rand` value must be expressed in decimal notation (e.g. `0.1234`) when propagating to downstream SDKs.
+1. On incoming traces, SDKs must use the incoming `sample_rand` value. They also must propagate the incoming `sample_rand` value to any downstream SDKs.
    1. Prior to using the incoming `sample_rand` value, SDKs must validate that the value is a floating-point number within the range `[0.0, 1.0)`. If the value is invaild, the SDK should ignore the incoming `sample_rand` value and generate a new one according to the rules described below.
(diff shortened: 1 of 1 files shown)


develop-docs/sdk/telemetry/traces/index.mdx
```diff
@@ -228,16 +228,22 @@ A transaction's sampling decision should be passed to all of its children, inclu
...
-   1. If `sample_rate` (inside `baggage`) and the sampling decision (trailing `-1` or `-0` from the `sentry-trace` header) are present in the incoming trace, create `sample_rand` so that when compared to the incoming `sample_rate` it would lead to the same sampling decision that is in the `sentry-trace` header. This means, for a decision of `True` generate a random number in half-open range `[0, rate)` and for a decision of `False` generate a random number in range `[rate, 1)`.
-   1. If the sampling decision is missing on the incoming trace, generate a random number in range of `[0, 1)` (including 0.0, excluding 1.0), like for a new trace.
+SDKs must set and propagate the `sample_rand` as follows:
+
+1. Whenever a new trace is started, including in the Tracing without Performance case, the SDK generates a `sample_rand` value for the **entire trace**.
+   1. SDKs must generate the `sample_rand` by pseudo-randomly sampling a floating-point number from the uniform distribution over the range `[0.0, 1.0)` (including `0.0`, excluding `1.0`).
```
jan-auer, line 234: I think it would be easier to say that we're generating a pseudorandom number, rather than sampling one, to not confuse this with the actual sampling that occurs afterwards. However, is it even necessary to call out "pseudo-random" here if in the last sub-bullet it recommends a deterministic number? An implementation doesn't have to follow that recommendation and we actually don't care if the number is pseudorandom or true random.

Same for the "uniform distribution". This is correct and I think this is good to add for correctness, but I wonder if this just adds complexity to the spec given that a uniform distribution is implied when one says "generate a random number".

develop-docs/sdk/telemetry/traces/index.mdx
```diff
@@ -228,16 +228,22 @@ A transaction's sampling decision should be passed to all of its children, inclu
...
+   1. The exact method for pseudo-random sampling is implementation-defined, and may therefore vary between SDK implementations.
+   1. We _recommend_ that SDKs generate the random number deterministically by using the trace ID to seed the random number generator. If multiple threads share the same random instance, seed the random number generator atomically, so that seeds do not get shared accross threads.
+1. The `sample_rand` is part of the Dynamic Sampling Context (DSC). Therefore, SDKs must propagate the `sample_rand` to downstream SDKs by adding it to the `baggage` header.
```
jan-auer, line 237: Would love some input from SDK folks here, since the previous spec already mentioned `baggage`:

I would not call out baggage explicitly. There's already gotta be a place where we define how the DSC is supposed to be propagated and we shouldn't need to duplicate this mechanism here. However, if it helps to mention `baggage` for clarity, I'm open to keep it.

Same for the sub-bullet on how to propagate sample_rand in baggage. Shouldn't we have central documentation for this on the [DSC page](https://develop.sentry.dev/sdk/telemetry/traces/dynamic-sampling-context/)?

lforst (approved):

jan-auer (commented):
Initial review pass, I will look into this further tomorrow. The changes look correct to me, see below for comments on what I think does not need to be specified. I'll check tomorrow for a way to specify this in a less complex way.

develop-docs/sdk/telemetry/traces/index.mdx
lforst: +1 - it read overly and a bit unnecessarily clever. "generate a random number in the range ..." feels good enough

szokeasaurusrex:
@jan-auer, made some changes in response to your feedback. What do you think?

develop-docs/sdk/telemetry/traces/index.mdx
```diff
@@ -228,16 +228,22 @@ A transaction's sampling decision should be passed to all of its children, inclu
...
+   1. The `sample_rand` value must be expressed in decimal notation (e.g. `0.1234`) when propagating to downstream SDKs.
+1. On incoming traces, SDKs must use the incoming `sample_rand` value. They also must propagate the incoming `sample_rand` value to any downstream SDKs.
+   1. Prior to using the incoming `sample_rand` value, SDKs must validate that the value is a floating-point number within the range `[0.0, 1.0)`. If the value is invaild, the SDK should ignore the incoming `sample_rand` value and generate a new one according to the rules described below.
+1. If `sample_rand` is missing or invalid on an incoming trace, the SDK generates a new `sample_rand` value using the following **special rules**. The SDK must then use the new `sample_rand` value and propagate it to any downstream SDKs.
+   1. If `sample_rate` **and** the sampling decision are present in the incoming trace, generate a new `sample_rand` value by pseudo-randomly selecting a floating-point number from the appropriate range:
+      1. If the trace is sampled, select from the range `[0.0, sample_rate)`.
+      1. If the trace is not sampled, select from the range `[sample_rate, 1.0)`.
```
szokeasaurusrex, line 245: What do we do when a user in the head SDK forces a sampling decision for a specific trace?

szokeasaurusrex: e.g. sample rate is 0.5, but the user creates a transaction with `sentry_sdk.start_transaction(sampled=True)`. Do we even bother to generate a `sample_rand` in this case?

philipphofmann:
@szokeasaurusrex and @jan-auer can we merge this?

PR #12600, opened 2025-02-06
docs(js): Update build options for Next.js

chargome:
This updates `/platforms/javascript/guides/nextjs/configuration/build/` to include all build options from `withSentryConfig` – the data basically mirrors our ts interface.


chargome:
@inventarSarah fyi this pr adds the options we talked about

PR #12605, opened 2025-02-06
feat: Add APIs page for JavaScript

mydea:
This adds a new top-level page to the JavaScript docs called "APIs", which is a list of all the top-level APIs that users may need.

This is a manually curated list, and not supposed to be fully exhaustive - but it should contain everything a user may want to call. These are all the `Sentry.xxx` functions, basically.

For this, some new components are added to handle formatting of this, as well as code-based syntax highlighting. For now only typescript is supported (to reduce bundle size?) but we can add more if needed. Platforms can also get a `language` defined now which controls how stuff is formatted.

As of now, parameters are formatted in a pseudo-typescript format. Not sure if that works for all languages, but we can do something platform specific in the future if needed.

There may be adjustments to be done to the APIs, and I most likely have forgotten one thing or another, but we can always tweak this further where needed.

Example API references:
- https://getsentry.github.io/sentry-python/api.html
- https://pkg.go.dev/github.com/getsentry/sentry-go


philipphofmann:
I think it would be better to change the parameters to use the full width and have the object description below them. Now we have plenty of unused space and the object parameters could use the full width instead.

CleanShot 2025-02-07 at 10 19 30@2x

Instead of
| | |
|--------|--------|
| breadcrumb | description |
| hint | description | 

Do this
| | 
|--------|
| breadcrumb  |
| description |
| hint |
| description | 

How does that sound?

timfish:
Oooh, nice!

Don't forget dark mode:

image

mydea:
> I think it would be better to change the parameters to use the full width and have the object description below them. Now we have plenty of unused space and the object parameters could use the full width instead.

Yeah, I think that makes sense. I agree it is a bit wasteful as-is. This is really an early draft so def. needs some more love :D Also re: dark mode, this def. needs to be done, true!

mydea:
I have re-vamped the design completely now:

https://github.com/user-attachments/assets/3a429755-773d-4850-a1cb-9f70282676a5

(it should also work in dark mode now)

docs/platforms/javascript/common/enriching-events/scopes/index.mdx
```diff
@@ -77,7 +77,7 @@ Sentry.withScope((scope) => {
 Sentry.captureException(new Error("my other error"));
 ```
 
-You can access the current scope via `Sentry.getCurrentScope()`, but usually you should use `Sentry.withScope()` to interact with local scopes instead.
+You can access the current scope via `Sentry.getCurrentScope()`, but in most cases you should not use this API, but instead use `withScope` to generate and access a local scope. There are no guarantees about the consistency of `getCurrentScope` across different parts of your application, as scope forking may happen under the hood at various points.
```
mydea, line 80: @s1gr1d sneaked this change in here!

docs/platforms/javascript/common/apis.mdx
lforst, line 1: I think the styling could still be improved a bit. Something about this feels misaligned or empty (?):

Screenshot 2025-03-28 at 12 00 17

Like the title of the expandable not having any separator to the content and the heading within the content being larger than the title of the expandable.

We can do that in a follow-up PR tho.

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,941 @@
...
+
+<SdkApi
+  name="getFeedback"
+  signature="function getFeedback(): ReturnType<feedbackIntegration> | undefined"
+>
+  Get the feedback integration, if it has been added. This can be used to access
+  the feedback integration in a type-safe way.
+</SdkApi>
+
+<SdkApi
```
lforst, line 766: Something's busted here. It doesn't render past sendFeedback.

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,941 @@
...
+    ],
+  },
+},
+]}
+>
+Wraps a callback with a cron monitor check in. The check in will be sent to Sentry when the callback finishes.
+</SdkApi>
+</PlatformCategorySection>
+```
+````
```
lforst, line 941: This looks like odd syntax.

src/components/sdkApi.tsx
```diff
@@ -0,0 +1,158 @@
...
+type Props = {
+  name: string;
+  parameters: ParameterDef[];
+  signature: string;
+  categorySupported?: PlatformCategory[];
+  children?: React.ReactNode;
+  language?: string;
+};
+
+export function SdkApi({
```
lforst, line 34: I think it is super duper critical that we make the content of the API searchable (cmd+f) even if the expandibles are collapsed. Otherwise I would maybe argue it is better to expand them by default.

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,941 @@
+---
+title: APIs
```
lforst, line 2: Another random thing I spotted: The right sidebar is so large that it goes beyond the screen, however when the "current heading" updates while scrolling the highlighted heading doesn't scroll into view.

src/components/sdkApi.tsx
```diff
@@ -0,0 +1,158 @@
...
+        <Expandable title="Parameters">
+          <div className="space-y-3">
+            {parameters.map(param => (
+              <ApiParameterDef key={param.name} language={lang} {...param} />
+            ))}
+          </div>
+        </Expandable>
+      ) : null}
+
+      {children}
```
lforst, line 67: We need to play with the margins of the paragraphs of text that comes after the code/expandible. It needs to be closer I think. And after the text we need more whitespace.

lforst (commented):
Quick random pass with things I noticed. Absolutely love this change though ❤️

docs/platforms/javascript/common/apis.mdx
mydea: yeah, I agree this can be prettier/nicer for sure!

docs/platforms/javascript/common/apis.mdx
mydea: huh, pushed a fix hopefully!

docs/platforms/javascript/common/apis.mdx
s1gr1d: Maybe it's because all margins have a different width 🤔 
Created a comment about this for the "examples" (but it's the same in "Parameters"): https://github.com/getsentry/sentry-docs/pull/12605#discussion_r2018656467

src/components/sdkApi.tsx
```diff
@@ -0,0 +1,158 @@
...
+  language,
+}: ParameterDef & {language: string}) {
+  return (
+    <div className="space-y-1">
+      <div className="font-bold m-0">
+        {name}
+        {required ? <span className="text-red">*</span> : null}
+      </div>
+      <div className="space-y-1">
+        <div>
```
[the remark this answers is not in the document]
mydea: let me try how good that works - it's a bit tricky with the `<pre>` tag but maybe it's good enough here :)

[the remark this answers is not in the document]
mydea: I think it works fine, updating it 🙏

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,937 @@
...
+The started span will automatically be ended when the callback returns, and will thus measure the duration of the callback. The callback cann also be an async function.
+
+  <Expandable title='Examples'>
+  ```javascript
+  // Synchronous example
+ Sentry.startSpan({ name: 'my-span' }, (span) => {
+    measureThis();
+  });
+
+// Asynchronous example
```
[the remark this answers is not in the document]
mydea: this is probably outside of the scope of this PR then, as this is just the regular `<Expandable>` usage - imho the margins here are _generally_ fine 🤔 maybe it just looks a bit weird because there is no text, I'll add a sentence above the code block, maybe that helps!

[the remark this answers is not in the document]
mydea: hmm, I'd say that does not really make it much better xD
image

src/components/sdkApi.tsx
mydea: this is sadly pretty hard, because the text below "Parameters" is automatically converted to a <p>, which has `margin-top: 1.25em` - so no matter what margin-bottom we set for the expandable, there will always be at least this space there 😬 I do agree it would be nicer to have the spacing be a bit smaller there...

About the space after the text: Do you really feel that is not enough? Feels pretty "spacy" to me below the text 🤔 

image

docs/platforms/javascript/common/apis.mdx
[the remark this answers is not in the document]
s1gr1d: It's okay to keep it like this in this PR :D
But it makes sense to check the margins in a follow-up.

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,942 @@
...
+<SdkApi name="getClient" signature="function getClient(): Client | undefined">
+  Returns the currently active client.
+</SdkApi>
+
+<SdkApi
+  name="setCurrentClient"
+  signature="function setCurrentClient(client: Client): void"
+>
+  Make the given client the current client. You do not need this if you use
+  `init()`, this is only necessary if you are manually setting up a client.˝
```
Lms24, line 40:
```suggestion
  `init()`, this is only necessary if you are manually setting up a client.
```

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,942 @@
...
+  Flushes all pending events.
+</SdkApi>
+
+<SdkApi name="isEnabled" signature="function isEnabled(): boolean">
+  Returns true if the SDK is initialized & enabled.
+</SdkApi>
+
+<SdkApi
+  name="close"
+  signature="function flush(timeout?: number): Promise<boolean>"
```
Lms24, line 72:
```suggestion
  signature="function close(timeout?: number): Promise<boolean>"
```

Lms24: Also, should we put this directly under `flush` since the two are related?

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,942 @@
...
+      type: "(event: Event, hint: EventHint) => Event | null | Promise<Event | null>",
+    },
+  ]}
+>
+  Adds an event processor to the SDK. An event processor receives every event
+  before it is sent to Sentry. It can either mutate the event (and return it) or
+  return `null` to discard the event. Event processors can also return a
+  promise, but it is recommended to use this only when necessary as it slows
+  down event processing.
+</SdkApi>
```
Lms24, line 101: IIRC we discussed removing some smaller dedicated pages in favour of this API page. I'm not sure if we're still going to do this. In case of `addEventProcessor`, should we link to the [event processors](https://docs.sentry.io/platforms/javascript/enriching-events/event-processors/) page?

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,942 @@
...
+
+<SdkApi
+  name="addIntegration"
+  signature="function addIntegration(integration: Integration): void"
+>
+  Adds an integration to the SDK. This can be used to conditionally add
+  integrations after `Sentry.init()` has been called. Note that it is
+  recommended to pass integrations to `init` instead of calling this method,
+  where possible.
+</SdkApi>
```
Lms24, line 111: same q here - should we link to the [Integrations](https://docs.sentry.io/platforms/javascript/configuration/integrations/) page?

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,942 @@
...
+```javascript
+Sentry.lazyLoadIntegration("replayIntegration")
+  .then((integration) => {
+    Sentry.addIntegration(integration);
+  })
+  .catch((error) => {
+    // Make sure to handle errors here!
+    // This rejects e.g. if the CDN bundle cannot be loaded
+  });
+```
```
Lms24, line 125: same q here

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,942 @@
...
+If you use a bundler, using e.g. `const { replayIntegration } = await import('@sentry/browser')` is recommended instead.
+
+</SdkApi>
+
+## Scopes
+
+<SdkApi name="getCurrentScope" signature="function getCurrentScope(): Scope">
+  Returns the current scope. See{" "}
+  <PlatformLink to="/enriching-events/scopes/">Scopes</PlatformLink> for more
+  information.
```
Lms24, line 136: l: I think for this section we can link to the Scopes page directly below the heading, wdyt? If so, we can remove the individual links or we target the links to the specific sections on the Scopes page.

```suggestion
## Scopes

See{" "}
  <PlatformLink to="/enriching-events/scopes/">Scopes</PlatformLink> for more
  information.

<SdkApi name="getCurrentScope" signature="function getCurrentScope(): Scope">
  Returns the current scope. 
```

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,942 @@
...
+  signature="function getIsolationScope(): Scope"
+>
+  Returns the current isolation scope.
+</SdkApi>
+
+<SdkApi
+  name="withIsolationScope"
+  signature="function withIsolationScope(callback: (scope: Scope) => void): void"
+>
+  Forks the current isolation scope and calls the callback with the forked
```
Lms24, line 162: super-l: Not sure if "forking" is as widely understood as "cloning" but happy to leave this up to you

```suggestion
  Clones the current isolation scope and calls the callback with the cloned
```

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,942 @@
...
+  scope. See <PlatformLink to="/enriching-events/scopes/">Scopes</PlatformLink>{" "}
```
Lms24, line 163: l: If we go with my suggestion above, I'd remove the link here or link to the isolation scope section

```suggestion
  scope.
```

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,942 @@
...
+  for more information.
+</SdkApi>
+
+<SdkApi name="getGlobalScope" signature="function getGlobalScope(): Scope">
+  Returns the global scope. See{" "}
+  <PlatformLink to="/enriching-events/scopes/">Scopes</PlatformLink> for more
+  information.
```
Lms24, line 170: l: same here OR alternatively link directly to the global scope section on the scope page?
```suggestion
  Returns the global scope.
```

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,942 @@
...
+  signature={`function captureException(
+  exception: unknown,
+  captureContext?: CaptureContext
+): EventId`}
+  parameters={[
+    {
+      name: "exception",
+      required: true,
+      type: "unknown",
+      description: "The exception to capture. This is usually an Error object.",
```
Lms24, line 186: super-l: Maybe rephrase this a bit to explain `unknown`?
```suggestion
      description: "The exception to capture. For best results, pass an `Error` object but it accepts any kind of value.",
```

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,942 @@
...
+`CaptureContext`, you can also pass a `SeverityLevel` as second argument, e.g.
+`"error"` or `"warning"`.
+
+Messages show up as issues on your issue stream, with the message as the issue name.
+
+</SdkApi>
+
+## Enriching Events
+
+<SdkApi
```
Lms24, line 299: l: should we link to Enriching Events here? Similarly to my comment in scopes, we can also just link to specific sections in the respective methods

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,942 @@
...
+  Get the currently active span.
+</SdkApi>
+
+<SdkApi name="getRootSpan" signature="function getRootSpan(span: Span): Span">
+  Get the root span of a span.
+</SdkApi>
+
+## Sessions
+
+<SdkApi name="startSession" signature="function startSession(): void">
```
Lms24, line 661: m: I would add a sentence about the session and release health here to avoid confusion with what users would usually think a session is.

Lms24 (approved):
I like the page as it gives us a great overview of the most important public SDK API! Also thanks for iterating on the design!

I had some comments about linking to other pages from the respecitive sections in this doc. I think we should do this for almost every section (where it makes sense).

mydea:
I opened a PR to make the `<Expandable>` component, which is also used by the parameters here, more accessible and improve the styling with code blocks etc inside: https://github.com/getsentry/sentry-docs/pull/13168

docs/platforms/javascript/common/apis.mdx
inventarSarah, line 1: General question: will this page have a higher relevance for search? For example, when I look for "addIntegration", it would be great to have this page listed in the results somewhere near the top

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,942 @@
...
+  { name: "callback", type: "(span: Span) => T", required: true },
+]}>
+  Starts a new span, that is active in the provided callback.
+  This span will be a child of the currently active span, if there is one.
+
+Any spans created inside of the callback will be children of this span.
+
+The started span will automatically be ended when the callback returns, and will thus measure the duration of the callback. The callback cann also be an async function.
+
+  <Expandable title='Examples'>
```
inventarSarah, line 449: I like how you've used the Expandable component throughout the page – it makes the site look very neat! The consistency in use is also great and helps users find their way around 👍

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,942 @@
...
+
+This page shows all available to-level APIs of the SDK. You can use them like this:
+
+```javascript
+import * as Sentry from "@sentry/browser";
+
+Sentry.setTag("tag", "value");
+```
+
+## Available Options
```
inventarSarah, line 16: Should this be "Available APIs"?

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,942 @@
+---
+title: APIs
+description: "Learn more about APIs of the SDK."
+customCanonicalTag: "/platforms/javascript/apis/"
+sidebar_order: 3
+---
+
+This page shows all available to-level APIs of the SDK. You can use them like this:
```
inventarSarah, line 8: We could add a bit more information here at the top (e.g., very briefly explain the purpose of these APIs).
It could also be good to use the word "function" to make it clear that these are SDK APIs and not REST APIs (something like "... the primary functions provided by the SDK for interacting with ...").

Let me know if I can support you with this :)

docs/platforms/javascript/common/apis.mdx
```diff
@@ -0,0 +1,942 @@
...
+
+```javascript
+import * as Sentry from "@sentry/browser";
+
+Sentry.setTag("tag", "value");
+```
+
+## Available Options
+
+<TableOfContents ignoreIds={["available-options"]} />
```
inventarSarah, line 18: Just a note, probably not a topic right now:
When you open the page you see two ToC's - one in the center and one to the right.
I can imagine that users gravitate towards the ToC in the center to find the API they're looking for (it is more eye-catching than the ToC in the sidebar). So, it is definitely valuable and needs to be here!
Still, I wish we could toggle the ToC (visible by default); especially considering that the list will become longer rather than shorter in the future 😅

docs/platforms/javascript/common/apis.mdx
inventarSarah: Question: How do you currently decide which function to add an example to?

I'm just wondering if it would make sense to add more? I understand that some will not really need one, though

inventarSarah (approved):
Love the page! 
I believe many users will find this overview page helpful for accessing basic information quickly without reading through extensive documentation. I will use this page a loooot!

docs/platforms/javascript/common/apis.mdx
mydea: good point, will add a bit more context here!

docs/platforms/javascript/common/apis.mdx
mydea: yes 😅

docs/platforms/javascript/common/apis.mdx
mydea: hmm, I get what you mean. What would you propose (for a follow up):

1. allow to toggle the TOC in the main page?
2. Do not show the TOC in the sidebar?
3. Something else...?

docs/platforms/javascript/common/apis.mdx
mydea: I think for this case, we can/should eventually remove the event processors page 🤔 But I'll port some more of it's content in here!

docs/platforms/javascript/common/apis.mdx
mydea: this I would like to mostly get rid of eventually - most of the `enrichting events" pages are really just describing a single API or so!

docs/platforms/javascript/common/apis.mdx
mydea: really very much a gut-feeling thing, where I feel the API is more complex/harder to understand without an example 😅 We can def. add more where we feel like it!

docs/platforms/javascript/common/apis.mdx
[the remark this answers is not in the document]
mydea: for what it's worth, this should be a bit better now with the changes to the expandable!

docs/platforms/javascript/common/apis.mdx
mydea: lol, this shows: we do not have any page in docs about sessions, I think? 😬 This is the closest we have, I guess it's ok... http://localhost:3000/platforms/javascript/guides/react/configuration/releases/#sessions

docs/platforms/javascript/common/apis.mdx
Lms24: also fine with me!

docs/platforms/javascript/common/apis.mdx
Lms24: sounds good!

docs/platforms/javascript/common/apis.mdx
inventarSarah: My proposition:
3. allow to toggle the second level of list elements in the main area (default: hidden)
 
  Core APIs >
  Capturing Events >
  Enriching Events v
      setTag
      setTags
      
I understand if this is too extra :D

PR #12615, opened 2025-02-06
fix: smarter sidebar scrolling behavior

### LINKED ISSUE #10978: Auto-scroll to active element in nav sidebar
### Problem Statement

Today, when you click on an element in the nav sidebar that is below the fold, you get to the new page, but loose focus in the sidebar. This is a bit confusing/disorienting. See:

https://github.com/user-attachments/assets/c0507f33-190e-4119-9765-72933535492f

### Solution Brainstorm

We should either keep scroll position in the sidebar, or at least scroll down to the active element on pageload.

a-hariti:
The current behavior of the sidebar is is jarring when dealing with items below the fold

this PR makes it smarter and behave as expected

closes #10978

### DIFF
--- a/src/components/focus-active-link.tsx
+++ b/src/components/focus-active-link.tsx
@@ -1,23 +1,54 @@
 'use client';
 import {useEffect} from 'react';
 
+import {debounce} from 'sentry-docs/utils';
+
 type Props = {
   activeLinkSelector: string;
 };
 
 /** Make sure the active link is visible in the sidebar */
 export function ScrollActiveLink({activeLinkSelector}: Props) {
-  const isVsibible = (link: Element) => {
-    const rect = link.getBoundingClientRect();
-    return rect.top >= 0 && rect.bottom <= window.innerHeight;
-  };
+  useEffect(() => {
+    const sidebar = document.querySelector('[data-sidebar-link]')?.closest('aside');
+    if (!sidebar) {
+      return;
+    }
+    sidebar.addEventListener('click', e => {
+      const target = e.target as HTMLElement;
+      if (target.hasAttribute('data-sidebar-link')) {
+        const top = target.getBoundingClientRect().top;
+        sessionStorage.setItem('sidebar-link-poisition', top.toString());
+      }
+    });
+    // track active link position on scroll as well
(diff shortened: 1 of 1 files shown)


src/components/focus-active-link.tsx
```diff
@@ -1,23 +1,54 @@
...
+  useEffect(() => {
+    const sidebar = document.querySelector('[data-sidebar-link]')?.closest('aside');
+    if (!sidebar) {
+      return;
+    }
+    sidebar.addEventListener('click', e => {
+      const target = e.target as HTMLElement;
+      if (target.hasAttribute('data-sidebar-link')) {
+        const top = target.getBoundingClientRect().top;
+        sessionStorage.setItem('sidebar-link-poisition', top.toString());
```
chargome, line 21: can we use local storage here? Session storage has been causing gdpr problems lately

src/components/focus-active-link.tsx
```diff
@@ -1,23 +1,54 @@
...
-  const isVsibible = (link: Element) => {
-    const rect = link.getBoundingClientRect();
-    return rect.top >= 0 && rect.bottom <= window.innerHeight;
-  };
+  useEffect(() => {
+    const sidebar = document.querySelector('[data-sidebar-link]')?.closest('aside');
+    if (!sidebar) {
+      return;
+    }
+    sidebar.addEventListener('click', e => {
```
chargome, line 17: Please remove all event listeners in the useEffect cleanup function!

chargome (changes requested):
Thanks for adding this, works great! Left some feedback comments

src/components/focus-active-link.tsx
a-hariti: I think it's the perfect level of memory for this task (remember the position of the active link as long as the user is on the website)

why is it causing GDPR problems?

chargome: Our problem was related to the usage of session storage using our SDKs. But I agree with you, let me just get that confirmed.

chargome: good to go!

chargome (approved):
🚀

PR #12640, opened 2025-02-10
feat: deep link search results

### LINKED ISSUE #11699: Index anchor links in algolia

a-hariti:
Deep link to search result destinations when possible

Closes #11699


chargome (approved):
Nice!

We just bumped a dependency for @sentry/nextjs please make sure the lock file is synced correctly

PR #12660, opened 2025-02-11
docs(self-hosted): monitor self-hosted instance

aldy505:
## DESCRIBE YOUR PR

We don't have any good article or explanation on how to monitor self-hosted Sentry.

"Monitoring self-hosted Sentry" means 3 things:
1. Monitor the actual services made by Sentry (sentry, snuba, symbolicator, relay) by using Statsd
2. Monitor the dependencies (Kafka, Postgres, etc) by using Prometheus exporters, or similar things
3. Monitor the host machine or Docker daemon

cc @getsetnry/dev-infra

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

## LEGAL BOILERPLATE

## EXTRA RESOURCES

### DIFF
--- a/develop-docs/self-hosted/experimental/monitoring.mdx
+++ b/develop-docs/self-hosted/experimental/monitoring.mdx
@@ -1,0 +1,182 @@
+---
+title: Self-Hosted Monitoring
+sidebar_title: Monitoring
+sidebar_order: 10
+---
+
+<Alert title="Important" level="warning">
+    These are community-contributed docs. Sentry does not officially provide support for self-hosted configurations beyond the default install.
+</Alert>
+
+This page is considered experimental because everyone will have different setup and requirements for their monitoring system. It is also best to use your existing monitoring system, and try to integrate Sentry's with it, instead of spinning up a new one.
+
+Most containers have Statsd client that you can ingest to your monitoring system. If you have a native Statsd server instance, you can directly use it. If you don't, you might want to add some kind of converter that converts the ingested Statsd format into your own. For example, if you are using Prometheus, you can use [prometheus-statsd-exporter](https://github.com/prometheus/statsd_exporter) to bridge the gap.
+
+If you are using DataDog as your monitoring system, it is recommended to use [DogStatsD](https://docs.datadoghq.com/developers/dogstatsd/?tab=hostagent) as the collector.
+
+Recommended native Statsd server are [the original Statsd server made by Etsy](https://github.com/statsd/statsd) and some others on the [server re-implementations](https://github.com/statsd/statsd/blob/master/docs/server_implementations.md).
+
+Other than service monitoring, you should also monitor your own host instance by using [Prometheus Node Exporter](https://github.com/prometheus/node_exporter), and your Docker daemon by using [built-in Prometheus exporter](https://docs.docker.com/engine/daemon/prometheus/) or similar tools. Sentry does not provides any alerts if your host instance has low available remaining RAM or low remaining disk space.
+
+<Alert title="Note">
+    After changing configuration files, re-run the <code>./install.sh</code> script, to rebuild and restart the containers. See the <Link to="/self-hosted/#configuration">configuration section</Link> for more information.
+</Alert>
+
+## Sentry-related configurations
+
+### Sentry
+
+You can configure Sentry to send metrics to Statsd server by configuring your `sentry/sentry.conf.py` file:
+
+```python
+SENTRY_METRICS_BACKEND = 'sentry.metrics.statsd.StatsdMetricsBackend'
+SENTRY_METRICS_OPTIONS: dict[str, Any] = {
+    'host': 'statsd.yourcompany.com',
+    'port': 8125,
+}
+# SENTRY_METRICS_SAMPLE_RATE = 1.0   # Adjust this to your needs, default is 1.0
+# SENTRY_METRICS_PREFIX = "sentry."  # Adjust this to your needs, default is "sentry."
+```
+
+### Snuba
+
+You can configure Snuba to send metrics to Statsd server by configuring your `docker-compose.yml` file:
+
+```yaml
+# The rest of your Docker Compose file
+
+x-snuba-defaults: &snuba_defaults
+  <<: *restart_policy
+  depends_on:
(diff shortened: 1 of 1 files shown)


develop-docs/self-hosted/experimental/monitoring.mdx
```diff
@@ -0,0 +1,182 @@
...
+    ports:
+      - "127.0.0.1:9121:9121"
+    environment:
+      REDIS_ADDR: "redis:6379"
+  prometheus-kafka-exporter:
+    <<: *restart_policy
+    image: danielqsj/kafka-exporter
+    ports:
+      - "127.0.0.1:9308:9308"
+    command: ["--kafka.server=kafka:9092"]
```
aldy505, line 129: I was a bit hesitant to use this, since most people recommend this: https://github.com/prometheus/jmx_exporter

Personally I don't use both, since I use Redpanda, and they already got built-in Prometheus metrics https://docs.redpanda.com/current/manage/monitoring/

develop-docs/self-hosted/experimental/monitoring.mdx
```diff
@@ -0,0 +1,182 @@
...
+        <port>9363</port>
+        <metrics>true</metrics>
+        <events>true</events>
+        <asynchronous_metrics>true</asynchronous_metrics>
+        <errors>true</errors>
+    </prometheus>
+</clickhouse>
+```
+
+Up to this point, all metrics are exposed only on `localhost`. If you don't want your Sentry installation to expose any other ports, you can remove every `ports` directive on the `docker-compose.yml` file and route everything through the `nginx` container, by modifying your `nginx.conf` file to the following:
```
aldy505, line 154: There is also one reason why I don't provide monitoring the Nginx container, because I don't think that the `stub_status` module is not included on the Nginx's Alpine Docker image: https://nginx.org/en/docs/http/ngx_http_stub_status_module.html

aldy505:
Ping @bc-sentry @hubertdeng123 @BYK

aldy505:
@bc-sentry @hubertdeng123 @BYK another ping

develop-docs/self-hosted/experimental/monitoring.mdx
```diff
@@ -0,0 +1,184 @@
...
+title: Self-Hosted Monitoring
+sidebar_title: Monitoring
+sidebar_order: 10
+---
+
+<Alert title="Important" level="warning">
+    These are community-contributed docs. Sentry does not officially provide support for self-hosted configurations beyond the default install.
+</Alert>
+
+This page is considered experimental because everyone will have different setup and requirements for their monitoring system. It is also best to use your existing monitoring system, and try to integrate Sentry's with it, instead of spinning up a new one.
```
BYK, line 11:
```suggestion
This page is considered experimental because everyone will have different setup and requirements for their monitoring system. It is also best to use your existing monitoring system, and try to integrate Sentry with it, instead of spinning up a new one.
```

BYK (changes requested):
This is pretty good but I have two issues with it:

1. Looks like there's a preference of using Prometheus for most things but then since Sentry uses `statsd` clients by default it keeps referring to those. I think the article should take a clear stance and voice: go all in with Prometheus, or just mention it once and be done with it. I think I'd lean towards all-in Prometheus but I'll leave the final decision to you.
2. The code examples for monitoring are hard to navigate. I think we should just move these over to their respective files, commented out with a link to this article. WDYT?

develop-docs/self-hosted/experimental/monitoring.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+Most containers have Statsd client that you can ingest to your monitoring system. If you have a native Statsd server instance, you can directly use it. If you don't, you might want to add some kind of converter that converts the ingested Statsd format into your own. For example, if you are using Prometheus, you can use [prometheus-statsd-exporter](https://github.com/prometheus/statsd_exporter) to bridge the gap.
+
+If you are using DataDog as your monitoring system, it is recommended to use [DogStatsD](https://docs.datadoghq.com/developers/dogstatsd/?tab=hostagent) as the collector.
```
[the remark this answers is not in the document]
aldy505: Dunno, I never pay for those :P

aldy505:
> 1. Looks like there's a preference of using Prometheus for most things but then since Sentry uses `statsd` clients by default it keeps referring to those. I think the article should take a clear stance and voice: go all in with Prometheus, or just mention it once and be done with it. I think I'd lean towards all-in Prometheus but I'll leave the final decision to you.

I'm not sure either, as I don't think most of the people who deploy self-hosted Sentry are devops person. I think it'd be best if we provide a rough idea on how they can monitor anything other than what Sentry made (meaning monitoring Kafka, Postgres, etc). But if they have gone this far, I guess they should already know on how to monitor those.

> 2. The code examples for monitoring are hard to navigate. I think we should just move these over to their respective files, commented out with a link to this article. WDYT?

I would prefer to have it this way. I think this is a part that doesn't easily get outdated anyway.

BYK:
>I would prefer to have it this way. I think this is a part that doesn't easily get outdated anyway.

It's not about getting outdated but putting the contextually relevant things close together. If these live on docs, then one need to go back and forth between the docs and the config files. If they are in config files with clear references, one can simply read the docs and then switch to configs and code completely to implement it without backreferencing.

Makes sense?

BYK:
>I'm not sure either, as I don't think most of the people who deploy self-hosted Sentry are devops person. I think it'd be best if we provide a rough idea on how they can monitor anything other than what Sentry made (meaning monitoring Kafka, Postgres, etc). But if they have gone this far, I guess they should already know on how to monitor those.

Bu we already take sides so either make it fully neutral (which is very hard) or just pick your battle and expect other people to understand how to adapt it to their needs. A half-baked or middle-of-the-road doc like this is more confusing. At least you'll be internally consistent and easier to transform if you just pick a side.

aldy505:
Okay let's just provide the docs to monitor Sentry-stuff using statsd. Let's ignore everything else and put a disclaimer.

BYK (approved):

PR #12667, opened 2025-02-11
downrank API search results

### LINKED ISSUE #11333: Search: adapt API page ranking
### Problem Statement

Based on this feedback from Slack:
>Is there a way for us to mark the Product explanation pages as "most important"?
If I search on certain topics I first have to go through (in this case) 6 rows of API hits instead of the Spike protection general docs I'm looking for.
I should not be looking into API docs until I have an edge case tbh

### Solution Brainstorm

Maybe we should generally rank api pages lower - up to discuss

a-hariti:
closes #11333


src/components/search/index.tsx
```diff
@@ -172,18 +172,42 @@ export function Search({
         return;
       }
 
-      const queryResults = await search.query(
```
a-hariti, line 175: @chargome 👀

chargome (commented):
The search parts looks good but I don't get why all these other changes are in this pr

a-hariti:
> The search parts looks good but I don't get why all these other changes are in this pr

I did a `git merge master` to avoid a build error, the PR changed one file only

PR #12670, opened 2025-02-11
ci: Don't use deprecated volta action

lforst:
Fixes failures on main


lforst:
Doing this in https://github.com/getsentry/sentry-docs/pull/12666

PR #12679, opened 2025-02-11
Updating images for legacy pricing for uptime

codyde:
Updating images for legacy pricing page to include uptime monitors. Adding uptime monitoring content link.


coolguyzone (approved):
Looks good! FYI You'll need to rebase from master to resolve the failing Vercel builds here.

PR #12683, opened 2025-02-12
Adding unsupported hot reload

rodolfoBee:
As concluded in  https://github.com/getsentry/sentry-dotnet/issues/3889 Hot Reload isn't supported with static libraries. Adding a note to the troubleshooting page.

Closes getsentry/sentry-dotnet#3959

### DIFF
--- a/platform-includes/troubleshooting/dotnet.mdx
+++ b/platform-includes/troubleshooting/dotnet.mdx
@@ -50,3 +50,11 @@ using SentrySession = Sentry.Session;
 ```
 
 Then `SentrySession` can be used instead of `Sentry.Session`.
+
+### VS Hot Reload isn't compatible with static assemblies
+
+Attempting to build a .NET MAUI iOS app will result in an error if you use Visual Studio Hot Reload when developing on Windows with a remote mac:
+
+> The xcframework sentry.bindings.cocoa\5.0.1\lib\net8.0-ios17.0\Sentry.Bindings.Cocoa.resources.zip has an incorrect or unknown format and cannot be processed.
+
+This happens because Hot Reload does not support static iOS libraries or frameworks containing static libraries, as detailed in the [Microsoft documentation](https://learn.microsoft.com/en-us/dotnet/maui/ios/hot-restart?view=net-maui-9.0#limitations).
--- a/platform-includes/troubleshooting/dotnet.mdx
+++ b/platform-includes/troubleshooting/dotnet.mdx
@@ -51,9 +51,9 @@ using SentrySession = Sentry.Session;
 
 Then `SentrySession` can be used instead of `Sentry.Session`.
 
-### VS Hot Restart isn't compatible with static assemblies
+### Visual Studio Hot Restart Incompatibility
 
-Attempting to build a .NET MAUI iOS app will result in an error if you use Visual Studio Hot Restart when developing on Windows with a remote mac:
+Attempting to build a .NET MAUI iOS app will result in an error if you use Visual Studio Hot Restart when developing on Windows with a remote Mac:
 
 > The xcframework sentry.bindings.cocoa\5.0.1\lib\net8.0-ios17.0\Sentry.Bindings.Cocoa.resources.zip has an incorrect or unknown format and cannot be processed.
 
--- a/platform-includes/troubleshooting/dotnet.mdx
+++ b/platform-includes/troubleshooting/dotnet.mdx
@@ -51,10 +51,10 @@ using SentrySession = Sentry.Session;
 
 Then `SentrySession` can be used instead of `Sentry.Session`.
 
-### VS Hot Reload isn't compatible with static assemblies
+### VS Hot Restart isn't compatible with static assemblies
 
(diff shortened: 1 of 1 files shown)


rodolfoBee:
@phunkeler thank you for the correction :D

platform-includes/troubleshooting/dotnet.mdx
```diff
@@ -50,3 +50,11 @@ using SentrySession = Sentry.Session;
 ```
 
 Then `SentrySession` can be used instead of `Sentry.Session`.
+
+### VS Hot Restart isn't compatible with static assemblies
```
phunkeler, line 54:
```diff
- ### VS Hot Restart isn't compatible with static assemblies
+ ### Visual Studio Hot Restart Limitation
```

```diff
- ### VS Hot Restart isn't compatible with static assemblies
+ ### Visual Studio Hot Restart Incompatibility
```

A bit more concise.

platform-includes/troubleshooting/dotnet.mdx
```diff
@@ -50,3 +50,11 @@ using SentrySession = Sentry.Session;
...
+
+Attempting to build a .NET MAUI iOS app will result in an error if you use Visual Studio Hot Restart when developing on Windows with a remote mac:
```
bruno-garcia, line 56:
```suggestion
Attempting to build a .NET MAUI iOS app will result in an error if you use Visual Studio Hot Restart when developing on Windows with a remote Mac:
```

platform-includes/troubleshooting/dotnet.mdx
bruno-garcia, line 54:
```suggestion
### Visual Studio Hot Restart Incompatibility
```

platform-includes/troubleshooting/dotnet.mdx
```diff
@@ -50,3 +50,11 @@ using SentrySession = Sentry.Session;
...
 
 Then `SentrySession` can be used instead of `Sentry.Session`.
+
+### Visual Studio Hot Restart Incompatibility
+
+Attempting to build a .NET MAUI iOS app will result in an error if you use Visual Studio Hot Restart when developing on Windows with a remote Mac:
+
+> The xcframework sentry.bindings.cocoa\5.0.1\lib\net8.0-ios17.0\Sentry.Bindings.Cocoa.resources.zip has an incorrect or unknown format and cannot be processed.
+
+This happens because Hot Restart does not support static iOS libraries or frameworks containing static libraries, as detailed in the [Microsoft documentation](https://learn.microsoft.com/en-us/dotnet/maui/ios/hot-restart?view=net-maui-9.0#limitations).
```
bruno-garcia, line 60: @aritchie curious, nothing we can do about this? 

People need to uninstall Sentry to use Hot Restart?

platform-includes/troubleshooting/dotnet.mdx
```diff
@@ -50,3 +50,11 @@ using SentrySession = Sentry.Session;
 ```
 
 Then `SentrySession` can be used instead of `Sentry.Session`.
+
+### Visual Studio Hot Restart Incompatibility
+
+Attempting to build a .NET MAUI iOS app will result in an error if you use Visual Studio Hot Restart when developing on Windows with a remote Mac:
```
phunkeler, line 56: I don't mean to split-hairs here but Hot Restart does not require a "remote Mac", so I'd suggest removing that part. Per the [MS doc](https://learn.microsoft.com/en-us/dotnet/maui/ios/hot-restart?view=net-maui-9.0): 

> .NET Multi-platform App UI (.NET MAUI) hot restart enables you to quickly deploy a .NET MAUI app to a 64-bit local iOS device, from Visual Studio 2022, without requiring a Mac build host

bruno-garcia: I think in these cases it's better to error on the side of 'over communicating' or being redundant because not everyone really knows these terms well (the diff between Hot Restart and Hot Reload) so while having the right term is important, expanding on the conditions where the issue maninfest is fine imo

platform-includes/troubleshooting/dotnet.mdx
aritchie: @bruno-garcia We may be able to detect hotrestart in a prebuild step, but I'm not sure.  I'll look into it.  I wish the tooling gave better error messages in these cases

coolguyzone (approved):

platform-includes/troubleshooting/dotnet.mdx
bruno-garcia: Just seems like we at least need to say then: "To use Hot Restart you'll need to remove the Sentry SDK" if that's the solution. But that's quite sub optimal obviously

aritchie: There is likely some form of condition that can be applied to the package reference and if/def constant you can use in code to remove UseSentry.  I'll ask around.

Thoughts

```xml
<PackageReference Include="Sentry.Maui" Condition="$(HOT_RESTART) == 'false'" />
```

```csharp
#if !HOT_RESTART
.UseSentry(...)
#endif
```

PR #12737, opened 2025-02-16
feat(electron): Update child process details

timfish:
Much of the docs around native crashes in child processes was outdated for the Electron versions we support. I've replaced it with a table detailing what child processes are supported on each platform.

image


Lms24 (approved):
Thanks!

PR #12739, opened 2025-02-17
feat(aws-lambda): Use versioned docs for v7 and v8 lambda layers

andreiborza:


chargome (approved):
Thanks for adding, if you want you can opt-out of indexing the versioned pages with `noindex: true` in the frontmatter.

PR #12745, opened 2025-02-17
fix: index index community sdks

### LINKED ISSUE #12639: Add keywords for community SDKs to /platforms
### Problem Statement

Page: https://docs.sentry.io/platforms/

We just discovered that the community SDKs can not be found via search:
Image

### Solution Brainstorm

Can we add keywords or something so that a potential search for e.g. "Hono" would lead to that page?

a-hariti:
I changed the list of community sdks into a `<ul>`

I ran the indexing locally, it seems to pick them up

closes  #12639


stephanie-anderson (approved):
uiiiiiii!!! 🙏

PR #12793, opened 2025-02-20
fix(rust): fix logo for actix-web

lcian:
## DESCRIBE YOUR PR
Fixes the logo to the correct one for `actix-web`, which was previously showing up with the Rust logo instead because `rust-actix` didn't match the platform name.

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

### DIFF
--- a/src/components/platformIcon.tsx
+++ b/src/components/platformIcon.tsx
@@ -997,7 +997,7 @@ export const PLATFORM_TO_ICON = {
   'ruby-sidekiq': 'sidekiq',
   'ruby-sinatra': 'sinatra',
   rust: 'rust',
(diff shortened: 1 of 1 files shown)


adinauer (approved):
Fix LGTM

`l` side note:
Currently looks like this:
Screenshot 2025-02-21 at 07 33 47

Should we use a different version of the icon?

lcian:
> Fix LGTM
> 
> `l` side note: Currently looks like this: Screenshot 2025-02-21 at 07 33 47
> 
> Should we use a different version of the icon?

Good catch, looks like the homepage https://docs.sentry.io/ uses different logos than the platform page.

PR #12797, opened 2025-02-21
feat(alerts): Add info about custom metric alerts limitation

priscilawebdev:
Preview:


docs/product/alerts/alert-types.mdx
```diff
@@ -36,9 +36,12 @@ The **Alert Details** page also includes a list of issues that triggered the ale
...
+Metric alerts monitor macro-level metrics for both error and transaction events. A metric takes a set of events and computes an aggregate value using a function, such as `count()` or `avg()`, applied to the event properties over a period of time. When you create a metric alert, you can filter events by attributes and <PlatformLink to="/enriching-events/tags/">tags</PlatformLink>, which is particularly useful for aggregating across events that aren't grouped into single issues. Sentry has the following limits on metric alerts:
+
+- **Organization**: A maximum of **1000 metric alerts** is allowed per organization.
+- **Custom Metric**: A maximum of **50 custom metric alerts** is allowed per project.
```
ale-cota, line 44: @priscilawebdev I don't think it's correct to say that the custom metric alerts limit is 50. I think these on-demand alerts can be created by picking any alert type under Performance or Custom (because they're based on the Transactions dataset). So for example it doesn't apply to metric alerts on Errors or Sessions data.

And within those types of alerts, they become on-demand when you use advanced filters, so I would suggest to maybe add the into about the limit here instead: https://docs.sentry.io/product/alerts/create-alerts/metric-alert-config/#advanced-filters-for-transactions

Also, this only applies to AM2 plans 🙈 I realise that on the latest version (AM3 plans) customers don't run into this. So maybe we can update the copy to "A maximum of **50 alerts** with advanced filters is allowed per project on certain legacy plans."

Screenshot 2025-02-21 at 11 58 35

@obostjancic, can you maybe confirm that what I said above is all true? 🙏

priscilawebdev: that makes total sense @ale-cota thank you! 🙏 Because I could not find anything mentioning 'on-demand' alerts, I assumed that it was 'custom metrics'.

obostjancic: @ale-cota yep, exactly!

obostjancic (approved):

PR #12819, opened 2025-02-24
docs(replay): Clarify beforeErrorSampling

chargome:
Clarifies that beforeErrorSampling is only triggered in `buffer` mode.

closes https://github.com/getsentry/sentry-javascript/issues/15470


coolguyzone (approved):
🛳️

PR #12720, opened 2025-02-13
ref(insights): move performance docs under insights

DominikB2014:
Work for https://github.com/getsentry/sentry/issues/84021

We are removing the performance link from the sidebar, all performance pages will fall under insights. 

This PR isn't to be merged for a few days until the performance link is removed, but we should have doc changes ready.

This PR updates the docs to reflect the layout of the product.
image

1. Most of `/product/performance` is moved to `/product/insights/overview`
2. The exception to the above is the `setup`, `retention priorities` and `performance overhead` as these pages are more generic to insights
3. Adds redirect for old perf pages -> insights
4. Updates links for old perf pages to use insights links
5. Some features are getting removed like web vitals in the transaction summary, information about those features have been removed
6. The old perf page is moved under getting started with sentry
7. Remove setup page
8. Product/performance redirects to getting started/performance

PR #12865, opened 2025-02-27
fix(insights): Fix broken getting-started link

priscilawebdev:
In a new organization, when I navigate to performance monitoring or the mobile/frontend/backend products within insights, I encounter the following empty state:

Screenshot 2025-02-27 at 07 35 46

By clicking on the 'Set Up' button I am redirected to

image

This happens because, in the PR, we moved all the performance docs to insights, but the "Getting Started" documentation was [deleted](https://github.com/getsentry/sentry-docs/pull/12720/files#diff-c96d9efe4a56b03fc4212c0a817f04ccb2997fa35b7026d4376f4e6e87df43a6). I attempted to find similar documentation, but we don't have anything available. Therefore, I am restoring it, now under insights, fixing the issue but we'll need to update the copies accordingly cc @DominikB2014

### DIFF
--- a/docs/product/insights/getting-started.mdx
+++ b/docs/product/insights/getting-started.mdx
@@ -1,0 +1,106 @@
+---
+title: Set Up
+sidebar_order: 1
+description: "Get started with Sentry's Performance Monitoring, which allows you to see macro-level metrics to micro-level spans, cross-reference transactions with related issues, and customize queries."
+---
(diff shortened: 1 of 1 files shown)


constantinius (approved):
LGTM
Are there any links to this document we may need to update? Probably not, as it was deleted earlier

PR #12913, opened 2025-03-04
fix(platform): Update onboarding options on tab switch

chargome:
- Sync selected onboarding options to `codeContext`
- Export reusable function from onboarding options to update code snippets
- In code tabs, trigger the function on tabSwitch based on the globally synced options

closes https://github.com/getsentry/sentry-docs/issues/12726


stephanie-anderson (approved):
Thank you, @chargome !

kahest:
@chargome on the [iOS docs](https://docs.sentry.io/platforms/apple/guides/ios/) for example, go to the `Objective-C` tab - it will initially show 2 superfluous empty lines which should be hidden:
Screenshot 2025-03-07 at 15 50 24

Toggling e.g. `Tracing` on/off shows the correct state:
Screenshot 2025-03-07 at 15 55 09

 After refreshing the page though, the original behaviour shows:
Screenshot 2025-03-07 at 15 55 19

It's not a huge problem, but it seems likely to have the same root cause and would be nice to fix in this PR if possible.

chargome:
@kahest I think this is already solved with this pr: https://sentry-docs-git-cg-fix-onboarding-options-java.sentry.dev/platforms/apple/guides/ios/

PR #12938, opened 2025-03-06
Clarify JumpCloud SSO actions

InterstellarStella:
Added a specification on the last step for setting up JumpCloud SSO. This should avoid confusion on which attributes to use.


coolguyzone (approved):
Thanks for the clarification! LGTM

PR #13006, opened 2025-03-14
feat(java): enable OTEL by default in onboarding options

lcian:
## DESCRIBE YOUR PR
The OTEL option will now be checked by default on the getting started guides for Java backend platforms.
OTEL is fully integrated with the SDK by now and improves scopes and trace propagation.
It can be enabled independently from tracing, which we still leave unchecked by default.

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

### DIFF
--- a/platform-includes/getting-started-install/java.jul.mdx
+++ b/platform-includes/getting-started-install/java.jul.mdx
@@ -1,5 +1,5 @@
 <OnboardingOptionButtons
-  options={["error-monitoring", "performance", "opentelemetry"]}
+  options={["error-monitoring", "performance", {id: "opentelemetry", checked: true}]}
 />
 
 ```xml {tabTitle:Maven}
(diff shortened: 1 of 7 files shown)


priscilawebdev:
Recently, we decided to have only errors selected by default in our SaaS product ([see](https://github.com/getsentry/sentry/pull/86395)) and it seems that the docs reflect that. Are we sure we want to have OpenTelemetry checked?

adinauer:
@priscilawebdev yeah, we won't enable Performance by default but would prefer to have OpenTelemetry on by default as it adds support for lots of frameworks. It's not just for creating spans but also `Scopes` (used to be hub) propagation as well as distributed tracing.

priscilawebdev:
@adinauer I understand. I just saw that in the onboarding, the Open Telemetry is checked by default, but there we use a different component.

image

adinauer:
Yes, we want it enabled by default everywhere.

PR #13011, opened 2025-03-17
feat(js): Stop using `provider.addSpanProcessor()`

mydea:
This fixes some things pointed out in https://github.com/getsentry/sentry-javascript/pull/15518, which are no longer compatible with OTEL v2. We already have a replacement for this in place.


andreiborza (approved):
Nice that this already works with our setup!

PR #13014, opened 2025-03-17
chore(apple): add SentrySDK.startProfiler to configure

philprime:
## DESCRIBE YOUR PR

The current configuration docs display an example usage of the `SentrySDK.startProfiler()` and `SentrySDK.stopProfiler()` and mentions that the call of stop is optional, as the profiler will automatically stop when going to the background.

Now if a user calls `SentrySDK.startProfiler()` in the `didFinishLaunching` hook, which is only called once per app start, the profiler would stop when the app goes to the first going to the background and not start again.

I would like to propose/discuss if the `SentrySDK.startProfiler()` should be used in [`applicationDidBecomeActive(_:)`](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/applicationdidbecomeactive(_:)) too.

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

cc @jas-kas

### DIFF
--- a/docs/platforms/apple/common/index.mdx
+++ b/docs/platforms/apple/common/index.mdx
@@ -119,6 +119,13 @@ func application(_ application: UIApplication,
 
     return true
 }
+
+// Start continuous profiling when the app becomes active again.
+//
+// This is optional, and you can start and stop the profiler as needed.
+func applicationDidBecomeActive(_ application: UIApplication) {
+    SentrySDK.startProfiler()
+}
 ```
 
 ```objc {tabTitle:Objective-C} {"onboardingOptions": {"performance": "12-15", "profiling": "17-27"}}
--- a/docs/platforms/apple/common/index.mdx
+++ b/docs/platforms/apple/common/index.mdx
@@ -87,7 +87,7 @@ To capture all errors, initialize the SDK as soon as possible, such as in your `
 
 <PlatformSection notSupported={["apple.tvos", "apple.watchos", "apple.visionos"]}>
 
-```swift {tabTitle:Swift} {"onboardingOptions": {"performance": "13-16", "profiling": "18-28"}}
+```swift {tabTitle:Swift} {"onboardingOptions": {"performance": "13-16", "profiling": "18-28,32-38"}}
 import Sentry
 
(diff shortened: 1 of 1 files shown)


philprime:
@armcknight it would be great if you could add some insights if these changes makes sense

docs/platforms/apple/common/index.mdx
```diff
@@ -119,6 +119,13 @@ func application(_ application: UIApplication,
 
     return true
 }
+
+// Start continuous profiling when the app becomes active again.
+//
+// This is optional, and you can start and stop the profiler as needed.
+func applicationDidBecomeActive(_ application: UIApplication) {
+    SentrySDK.startProfiler()
+}
```
philipphofmann, line 128: I'm not sure if this doesn't add too much noise. We already mention `SentrySDK.startProfiler()` above and tell the user that they have to manually start and stop the profiler.

coolguyzone (approved):

philprime:
I actually just noticed that the [Apple Documentation](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/applicationdidbecomeactive(_:)#Discussion) states the following:

> If you’re using scenes (see [Scenes](https://developer.apple.com/documentation/uikit/scenes)), UIKit will not call this method. Use [sceneDidBecomeActive(_:)](https://developer.apple.com/documentation/uikit/uiscenedelegate/scenedidbecomeactive(_:)) instead to restart any tasks or refresh your app’s user interface. UIKit posts a [didBecomeActiveNotification](https://developer.apple.com/documentation/uikit/uiapplication/didbecomeactivenotification) regardless of whether your app uses scenes.

AFAIK scenes are the default for creating new Xcode projects, so maybe we should also mention `sceneDidBecomeActive` for the SceneDelegate

kahest (changes requested):
@philprime this is probably superseded by https://github.com/getsentry/sentry-docs/pull/13197 - in general I think we're moving away from showing `start/stopProfiler` on this page completely

philprime:
Closing in favor of #13197

PR #13026, opened 2025-03-17
fix(docs): clarify database migrations

JoshFerge:
Was just talking with @MichaelSun48 and I believe this is worded in a slightly confusing way. If the column you're looking to drop is not nullable, you should make it nullable first before removing references. lmk if this makes sense @wedamija

### DIFF
--- a/develop-docs/backend/application-domains/database-migrations/index.mdx
+++ b/develop-docs/backend/application-domains/database-migrations/index.mdx
@@ -135,7 +135,7 @@ This is complicated due to our deploy process. When we deploy, we run migrations
 To avoid this, follow these steps:
 
 - First, if the column is either not nullable, or doesn't have a `db_default` set, then make a PR to make it nullable via `null=True`.
-- Then, make a PR to remove all uses of the column in the codebase in a separate PR. This mostly helps with code cleanliness. This should be merged ahead of the next migration PRs, but we don't need to worry about whether it is deployed first.
+- Then, remove all uses of the column in the codebase in a separate PR; this mostly helps with code cleanliness. This should be merged ahead of the next migration PRs, but we don't need to worry about whether it is deployed first.
 - Make another PR that:
     - If the column is a foreign key, remove the database level foreign key constraint it by setting `db_constraint=False`.
     - Remove the column and in the generated migration use `SafeRemoveField(..., deletion_action=DeletionAction.MOVE_TO_PENDING)` to replace `RemoveField(...)`. This only marks the state for the column as removed.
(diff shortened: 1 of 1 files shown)


develop-docs/backend/application-domains/database-migrations/index.mdx
```diff
@@ -134,7 +134,8 @@ This is complicated due to our deploy process. When we deploy, we run migrations
 
 To avoid this, follow these steps:
 
-- Make a PR to remove all uses of the column in the codebase in a separate PR. This mostly helps with code cleanliness. This should be merged ahead of the migration prs, but we don't need to worry about whether it is deployed first.
+- First, if the column is either not nullable, or doesn't have a `db_default` set. If either of these is true, then make it nullable via `null=True`.
```
MichaelSun48, line 137:
```suggestion
- First, if the column is either not nullable, or doesn't have a `db_default` set, then make a PR to make it nullable via `null=True`.
```

MichaelSun48 (commented):
Just one clarifying suggestion, otherwise lgtm — definitely more clear than before. I'll let @wedamija stamp this in case we're missing something here

develop-docs/backend/application-domains/database-migrations/index.mdx
```diff
@@ -134,7 +134,8 @@ This is complicated due to our deploy process. When we deploy, we run migrations
 
 To avoid this, follow these steps:
 
-- Make a PR to remove all uses of the column in the codebase in a separate PR. This mostly helps with code cleanliness. This should be merged ahead of the migration prs, but we don't need to worry about whether it is deployed first.
+- First, if the column is either not nullable, or doesn't have a `db_default` set, then make a PR to make it nullable via `null=True`.
+- Then, make a PR to remove all uses of the column in the codebase in a separate PR. This mostly helps with code cleanliness. This should be merged ahead of the next migration PRs, but we don't need to worry about whether it is deployed first.
```
coolguyzone, line 138:
```suggestion
- Then, remove all uses of the column in the codebase in a separate PR; this mostly helps with code cleanliness. This should be merged ahead of the next migration PRs, but we don't need to worry about whether it is deployed first.
```

develop-docs/backend/application-domains/database-migrations/index.mdx
```diff
@@ -134,7 +134,8 @@ This is complicated due to our deploy process. When we deploy, we run migrations
 
 To avoid this, follow these steps:
 
-- Make a PR to remove all uses of the column in the codebase in a separate PR. This mostly helps with code cleanliness. This should be merged ahead of the migration prs, but we don't need to worry about whether it is deployed first.
+- First, if the column is either not nullable, or doesn't have a `db_default` set, then make a PR to make it nullable via `null=True`.
```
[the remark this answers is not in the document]
wedamija: Right, that makes sense to me!

wedamija (approved):

PR #13030, opened 2025-03-18
Dart and Flutter Integrations Shared in platform docs correctly

smeubank:
- creates a subdirectory in includes for dart and flutter integrations
- from here these can be easily re-used by Dart and/or Flutter Docs
- the Dart integrations functionally remain the same
- Flutter now has all Dart integrations as well
- Updated the index page for Flutter integrations to also have a table, to describe each integration and what it does, auto, errors, performance
  - this is similar to what we see in [JS](https://docs.sentry.io/platforms/javascript/configuration/integrations/) and [Python](https://docs.sentry.io/platforms/python/integrations/) Docs
  - we could do more, icons like Python for example, or more generic information about config and setup like in JS, but it can be saved for another PR, when we agree on the table and check marks

Flutter:

image

Flutter Integrations:

image

Dart:

image


smeubank:
@buenaflor if we can agree on a version of the table for the integrations page, and the check marks, i think we can ship otherwise it's not a ton of content change. Just some ABC sorting and doc maintenance QoL stuff

kahest:
huge improvement, thanks @smeubank 🙏 
I'd keep the table as-is, not all of the integrations are for specific libraries that hat an associated icon, and I really like how concise the table is now

one tiny nit (feel free to disregard) - can we align on the title, i.e. if we append "integration", "instrumentation", or nothing? currently we have a mix. @buenaflor wdyt?

kahest (approved):

buenaflor:
I personally like the style JS goes for where the titles of the integration in the table have the same name as in code

I'll follow up here on which boxes to tick, I can manually edit the PR as well

smeubank:
> can we align on the title

nits welcome :D and I am fine with whichever decision y'all make we can always adapt later as well

> I'll follow up here on which boxes to tick, I can manually edit the PR as well

🙏  thank ya kindle

buenaflor (approved):
I changed the table names without the integration/instrumentation append

lgtm, we can iterate on this later on as well

PR #13032, opened 2025-03-18
feat(develop): Add clarification for capturing internal exceptions in SDKs

sl0thentr0py:


develop-docs/sdk/overview.mdx
```diff
@@ -328,6 +330,19 @@ Retrying too often may cause further rate limiting or blocking by the Sentry
...
 </Alert>
 
+### SDK Errors
+
+We try our best to make our SDKs error-free. We are an exception monitoring service after all
+and throwing our own exceptions is awkward. However, if our SDKs do throw exceptions, we have to make
+sure we swallow them gracefully and emit an error level log describing the failure.
+
+As a **design principle**, we never capture Sentry events for exceptions happening within our SDKs.
+This includes user-defined callbacks and hooks such as `before_send`, `traces_sampler` and so on.
```
szokeasaurusrex, line 340: wording nit

```suggestion
As a **design principle**, we never capture Sentry events for exceptions happening within our SDKs, including within user-defined callbacks and hooks such as `before_send` or `traces_sampler`.
```

sentrivana (approved):
Thanks for adding!

develop-docs/sdk/overview.mdx
```diff
@@ -328,6 +330,19 @@ Retrying too often may cause further rate limiting or blocking by the Sentry
...
+
+The main justification for this is that we are already in an event capturing flow where the scope has been applied
+and capturing another event at that point in the lifecycle would lead to undefined behavior. In the worst case,
+we could create a busy loop of creating and sending events repeatedly till the system crashes.
```
szokeasaurusrex, line 344: I think as written, this paragraph might be slightly unclear (especially due to the use of the word "this" in the first sentence; "this" can often be vague and imo should generally be avoided and technical writing), and perhaps a bit too technical for an overview page. I think it is enough to say that we could have undefined behavior and to mention the infinite loop example.

```suggestion
The reason we avoid capturing SDK exceptions is that capturing these exceptions could cause undefined behavior.
For example, we could easily end up in an infinite loop where trying to capture an exception triggers that same exception again.
```

szokeasaurusrex (approved):
I think the wording could be improved, but looks good in general, so I'll approve.

Please address my inline comments before merging.

develop-docs/sdk/overview.mdx
```diff
@@ -328,6 +330,18 @@ Retrying too often may cause further rate limiting or blocking by the Sentry
...
+
+We try our best to make our SDKs error-free. We are an exception monitoring service after all
+and throwing our own exceptions is awkward. However, if our SDKs do throw exceptions, we have to make
+sure we swallow them gracefully and emit an error level log describing the failure.
+
+As a **design principle**, we never capture Sentry events for exceptions happening within our SDKs, including within user-defined callbacks and hooks such as `before_send` or `traces_sampler`.
+
+The reason we avoid capturing internal SDK exceptions is that we are already in an event capturing flow
+where the scope has been applied and capturing another event at that point in the lifecycle
+would lead to undefined behavior. In the worst case, we could create a busy loop of creating and sending events repeatedly till the system crashes.
```
szokeasaurusrex, line 343: [nit] I think "till" is a bit informal for docs (and maybe difficult for non-native English speakers).

```suggestion
would lead to undefined behavior. In the worst case, we could create a busy loop of creating and sending events repeatedly until the system crashes.
```

develop-docs/sdk/overview.mdx
szokeasaurusrex: Sounds good!

PR #13039, opened 2025-03-18
fix(platform): sidebar deep links 404s

### LINKED ISSUE #12882: 404 error when switching between guides
### Steps to Reproduce

1. navigate to a page that is only available for some of multiple guides on the same platform - e.g. https://docs.sentry.io/platforms/apple/guides/ios/profiling/
2. switch to a different guide for the same platform, which does not have this page - open the platform/guide switcher, and select e.g. `visionOS`

### Expected Result

fall back to getting started page of the selected guide

### Actual Result

404 error is shown

### Additional Info

in the example given, the page is available for the platform (`apple`) and two guides (`apple.ios`, `apple.macos`), and not available for all other guides. this is implemented by marking the guides as `supported`/`notSupported` in the [common index page](https://github.com/getsentry/sentry-docs/blob/a2c773c80ba7b025c9117800fead35884a018fad/docs/platforms/apple/common/profiling/index.mdx?plain=1#L5). it might work differently if the respective page is only present in the supporting guides

a-hariti:
closes #12882

### DIFF
--- a/src/components/sidebar/index.tsx
+++ b/src/components/sidebar/index.tsx
@@ -60,9 +60,10 @@ export async function Sidebar({path, versions}: SidebarProps) {
 
         return {
           ...platform,
-          url: platformPageForCurrentPath
-            ? '/' + platformPageForCurrentPath.path + '/'
-            : platform.url,
+          url:
+            platformPageForCurrentPath && !platformPageForCurrentPath.missing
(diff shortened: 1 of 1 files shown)


coolguyzone (approved):
Code change looks good, works correctly on preview branch. LGTM!

PR #13041, opened 2025-03-19
feat(platform): preview project name on top of DSN

### LINKED ISSUE #13015: Add preview for org / project name for DSN
### Problem Statement

During an "audit"/testing of the Cocoa SDK onboarding flow I got redirected from the [sentry-wizard](https://github.com/getsentry/sentry-wizard) to the configuration documentation and I copied the configuration snippet by clicking the copy button in the top right corner.

Image

I did not click on the DSN, therefore it did see which org/project I chose and when I started sending events via the Cocoa SDK the events did not show up, **because they DSN was for another project**. If it somehow displayed the org/project I might have noticed earlier that the DSN was for the wrong one.

### Solution Brainstorm

One idea would be displaying the selected org / project with the DSN.

I drafted a simple visual example by editing the `keywordSelector.tsx`, replacing the `span` with this code snippet:

```tsx
        <span
          style={{
            // We set inline-grid only when animating the keyword so they
            // correctly overlap during animations, but this must be removed
            // after so copy-paste correctly works.
            display: isAnimating ? 'inline-grid' : undefined,
            position: 'relative',
          }}
        >
          <AnimatePresence initial={false}>
            <Keyword
              onAnimationStart={() => setIsAnimating(true)}
              onAnimationComplete={() => setIsAnimating(false)}
              key={currentSelectionIdx}
            >
              {currentSelection[keyword]}
            </Keyword>
          </AnimatePresence>
          {currentSelection?.title && (
            <div
              style={{
                position: 'absolute',
                top: '-20px',
                left: '50%',
                transform: 'translateX(-50%)',
                fontSize: '12px',
                backgroundColor: isDarkMode ? '#333' : '#f5f5f5',
                color: isDarkMode ? '#fff' : '#333',
                padding: '2px 6px',
                borderRadius: '3px',
                pointerEvents: 'none',
                whiteSpace: 'nowrap',
                opacity: 0.9,
                userSelect: 'none',
              }}
            >
              {currentSelection.title}
            </div>
          )}
        </span>
```

This looked something like this:

Image

Another idea would be changing the tooltip/alert/message/toast of the copy button to display what exactly has been copied:

Image

This might no help if a user selects the code by cursor and copies it as text.

cc @jas-kas @kahest

a-hariti:
also fixes a "copied" feedback z-index issue

closes #13015

### DIFF
--- a/src/components/codeBlock/code-blocks.module.scss
+++ b/src/components/codeBlock/code-blocks.module.scss
@@ -152,4 +152,5 @@
   border: none;
   color: var(--white);
   transition: opacity 150ms;
+  z-index: 10000;
 }
--- a/src/components/codeKeywords/keywordSelector.tsx
+++ b/src/components/codeKeywords/keywordSelector.tsx
@@ -24,6 +24,7 @@ import {
   Selections,
 } from './styles.css';
 import {dropdownPopperOptions} from './utils';
+import {relative} from 'node:path/posix';
 
 type KeywordSelectorProps = {
   group: string;
@@ -137,6 +138,7 @@ export function KeywordSelector({keyword, group, index}: KeywordSelectorProps) {
             // correctly overlap during animations, but this must be removed
             // after so copy-paste correctly works.
             display: isAnimating ? 'inline-grid' : undefined,
+            position: 'relative',
           }}
         >
           <AnimatePresence initial={false}>
@@ -148,6 +150,28 @@ export function KeywordSelector({keyword, group, index}: KeywordSelectorProps) {
               {currentSelection[keyword]}
             </Keyword>
           </AnimatePresence>
+          {!isOpen && currentSelection?.title && (
+            <div
+              className='no-copy'
+              style={{
+                position: 'absolute',
+                top: '-24px',
+                left: '50%',
+                transform: 'translateX(-50%)',
+                fontSize: '12px',
+                backgroundColor: '#333',
+                color: '#fff',
+                padding: '2px 6px',
+                borderRadius: '3px',
+                pointerEvents: 'none',
+                whiteSpace: 'nowrap',
+                opacity: 0.9,
+                userSelect: 'none',
+              }}
+            >
+              {currentSelection.title}
+            </div>
+          )}
         </span>
       </KeywordDropdown>
       {isMounted &&
--- a/src/components/codeBlock/index.tsx
+++ b/src/components/codeBlock/index.tsx
@@ -14,6 +14,31 @@ export interface CodeBlockProps {
   title?: string;
 }
 
+/**
+ *
+ * Copy `element`'s text children as long as long as they are not `.no-copy`
+ */
+function getCopiableText(element: HTMLDivElement) {
+  let text = '';
+  const walker = document.createTreeWalker(element, NodeFilter.SHOW_TEXT, {
+    acceptNode: function (node) {
+      // Skip if parent has .no-copy class
+      if (node.parentElement?.classList.contains('no-copy')) {
+        return NodeFilter.FILTER_REJECT;
+      }
+      return NodeFilter.FILTER_ACCEPT;
+    },
+  });
+
+  let node: Node | null;
+  // eslint-disable-next-line no-cond-assign
+  while ((node = walker.nextNode())) {
+    text += node.textContent;
+  }
+
+  return text.trim();
+}
+
 export function CodeBlock({filename, language, children}: CodeBlockProps) {
   const [showCopied, setShowCopied] = useState(false);
   const codeRef = useRef<HTMLDivElement>(null);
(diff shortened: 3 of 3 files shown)


philprime:
@a-hariti Thanks for the quick implementation!

How should we deal with the project preview overlaying code?
I don't think there is any good solution because the preview will most likely always be in the way. 

One idea that popped to my mind would be reducing the opacity when the cursor is on top of the preview, so that it is possible to view what's below the code, but not sure about that.

https://sentry-docs-git-dsn-preview.sentry.dev/platforms/apple/guides/ios/

image
image

a-hariti:
The ideal solution might be to track the user's gaze with AI and move the popup out of their eyesight 😂

a-hariti:
I came up with this hack to add some margin to lines with clickable "keywords", it doesn't discriminate between DSN's or other things, but the worst case scenario is having an apparent extra line of whitespace, so it should do @philprime 

image

src/components/codeKeywords/styles.css.ts
```diff
@@ -4,6 +4,22 @@ import {ArrowDown} from 'react-feather';
 import styled from '@emotion/styled';
 import {motion} from 'framer-motion';
 
+export const ProejectPreview = styled('div')`
```
philprime, line 7:
```suggestion
export const ProjectPreview = styled('div')`
```

src/components/codeKeywords/keywordSelector.tsx
```diff
@@ -148,6 +150,11 @@ export function KeywordSelector({keyword, group, index}: KeywordSelectorProps) {
               {currentSelection[keyword]}
             </Keyword>
           </AnimatePresence>
+          {!isOpen && currentSelection?.title && (
+            <ProejectPreview className="no-copy">
+              {currentSelection.title}
+            </ProejectPreview>
```
philprime, line 156:
```suggestion
            <ProjectPreview className="no-copy">
              {currentSelection.title}
            </ProjectPreview>
```

src/components/codeKeywords/keywordSelector.tsx
```diff
@@ -21,6 +21,7 @@ import {
   KeywordIndicator,
   KeywordSearchInput,
   PositionWrapper,
+  ProejectPreview,
```
philprime, line 24:
```suggestion
  ProjectPreview,
```

philprime (commented):
Found a typo, added suggestions.

As my background is more mobile UI than web UI development, it would be great if @chargome could do a proper review on the implementation.

chargome:
I think we shuold only show this for dsn, not for orgs and projects:

Screenshot 2025-03-24 at 10 47 11

philprime:
@a-hariti Am I missing something or is the preview for DSNs not in there anymore?

a-hariti:
You didn't miss anything @philprime, I messed up 🫣

please take another look :)

philprime (commented):
Code looks fine but it does not seem to pick up all usage cases:

https://sentry-docs-git-dsn-preview.sentry.dev/platforms/javascript/guides/node/
Screenshot 2025-04-07 at 09 36 46

https://sentry-docs-git-dsn-preview.sentry.dev/platforms/apple/guides/ios/
Screenshot 2025-04-07 at 09 37 08

https://sentry-docs-git-dsn-preview.sentry.dev/platforms/dart/guides/flutter/
Screenshot 2025-04-07 at 09 37 21

philprime (commented):
Can we make the tooltip not to be picked up from selection?

I selected the text to copy it (instead of using the copy button) and it copied like this:

```objc
@import Sentry;
(BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
    [SentrySDK startWithConfigureOptions:^(SentryOptions *options) {
        options.dsn = @"https://examplePublicKey@o0.ingest.sentry.io/0example-org / example-project";
        options.debug = YES; // Enabled debug when first installing is always helpful
        // Adds IP for users.
        // For more information, visit: https://docs.sentry.io/platforms/apple/data-management/data-collected/
        options.sendDefaultPii = YES
    }];
    return YES;
}
```

It seems to append the `example-org / example-project` to the DSN, while visually it is not selected.

This will cause UX issues with users who prefer to select code to copy it.

Screenshot 2025-04-08 at 13 02 39

a-hariti:
Good catch @philprime, it required some gymnastics, but I believe it's good to go now :)

chargome (commented):
@a-hariti could we only display this when the user is logged in? Not really useful when the example project is shown I think 🤔

sfanahata:
@a-hariti - what's the status of this one?

philprime:
This pull request seems to be abandoned by @a-hariti.

@chargome who would be the right person to ping to pick this up?

chargome:
I can take a look yes, otherwise @sergical is the man to ping now!

src/components/codeBlock/code-blocks.module.scss
```diff
@@ -150,4 +150,5 @@
   border: none;
   color: var(--white);
   transition: opacity 150ms;
+  z-index: 10000;
```
chargome, line 153: Not sure if this is really necessary, but should not interfere with anything else I think

chargome (approved):
lgtm
@sergical just leaving this open for you to take a look – feel free to merge in after!

PR #13068, opened 2025-03-20
Clarify message truncation

cleptric:


develop-docs/sdk/data-model/event-payloads/message.mdx
```diff
@@ -14,14 +14,14 @@ help to group similar messages into the same issue.
 : **Required**. The fully formatted message. If missing, Sentry will try to
   interpolate the message.
 
-  It must not exceed 8192 characters. Longer messages will be truncated. Sentry
+  A limit of 8192 characters is exposed in Relay and longer messages will be truncated. Sentry
```
philipphofmann, line 17: I think it would be great to clearly state that we recommend letting Relay truncate the message cause it gives us more flexibility.
 ```suggestion
  Relay truncates messages longer than 8192 characters. SDKs should not enforce the limit to allow greater flexibility for the truncation. Sentry
```

philipphofmann (approved):
LGTM, and I think we can improve the wording a bit.

PR #13072, opened 2025-03-20
fix: Avoid showing top-level platform in search

mydea:
If you search for something, e.g. "Next.js", we do not need to show the link to the top-level "JavaScript" entry in the platforms list.

### Old

Screenshot 2025-03-20 at 16 56 33

### New

Screenshot 2025-03-20 at 16 56 19

Note that if you search for the top-level term, it is still there - but only if the top-level entry actually matches the search term:

Screenshot 2025-03-20 at 16 56 24


a-hariti (approved):
Nice!

chargome (approved):
Thanks for all these docs prs! 🚀

PR #13087, opened 2025-03-21
chore(platform): Add auth token to config

chargome:
Uses SENTRY_WEBPACK_PLUGIN_AUTH_TOKEN as auth token for source maps uploads.


chargome:
we're actually renaming the auth token so it gets picked up automatically

PR #13090, opened 2025-03-21
Add info that associating proguard files with release is optional

buenaflor:
Currently the docs for associating proguard files with releases is not clear enough that it is optional and is not required.

Related: https://github.com/getsentry/sentry/issues/82821


docs/cli/dif.mdx
```diff
@@ -226,10 +226,16 @@ sentry-cli upload-proguard \
...
 ```
 
-Additionally, you can associate the proguard mapping file to a specific release:
+Additionally, you can associate the ProGuard mapping file to a specific release:
 
-```bash
-# Additionally create an association to the release my.app.id@1.0.0+1
+<Alert>
+
+This step is optional but can help you identify to which release a mapping file belongs to when multiple files exist for the same app.
```
coolguyzone, line 233:
```suggestion
This step is optional but can help you identify which release a mapping file belongs to when multiple files exist for the same app.
```

coolguyzone (approved):
LGTM!

docs/cli/dif.mdx
kahest: maybe "This step is not required for deobfuscation to work, but can help..." to be more specific about what "optional" means

kahest (approved):
small wording suggestions, otherwise lgtm, thanks!

docs/cli/dif.mdx
buenaflor: sounds good 👍

buenaflor:
So far I can't reproduce it, tested with [spotlight](https://spotlightjs.com/about/) to ingest the sent event data

First app start:

image

---
Back button, then reopen app:

image

So this is clearly not the same data unless I missed something important

PR #13108, opened 2025-03-24
feat(react): Document `Sentry.captureReactException`

AbhiPrasad:
Documents https://github.com/getsentry/sentry-javascript/pull/15746 released in `9.8.0` of the Sentry React SDK.

`captureReactException` is a method exposed that users can use if they don't want to use the Sentry Error Boundary component, but instead their own custom Error Boundary component.

### DIFF
--- a/docs/platforms/javascript/guides/react/features/error-boundary.mdx
+++ b/docs/platforms/javascript/guides/react/features/error-boundary.mdx
@@ -199,7 +199,7 @@ function App({ props }) {
 export default App;
 ```
 
-## Manual Error Capturing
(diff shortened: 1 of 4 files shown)


docs/platforms/javascript/guides/react/features/error-boundary.mdx
```diff
@@ -199,6 +199,24 @@ function App({ props }) {
 export default App;
 ```
 
+## Manual Error Capturing
```
coolguyzone, line 202:
```suggestion
## Manually Capturing Errors
```

PR #13109, opened 2025-03-24
feat(sentry): Store previous trace information in `sessionStorage`

Lms24:
Let's give this a try in our docs project to see how well it works. Happy to close if anyone has concerns regarding sessionStorage usage but we already use Replay in docs, so we already write and read from it. With this change, we can link to a previous trace even if users hard-navigate or hard-reload the page.


chargome (approved):
🚀

s1gr1d (approved):

PR #13110, opened 2025-03-24
ref(rust): add a more direct link to tracing crate integration

lcian:
## DESCRIBE YOUR PR
People using Rust expect to find docs about our integration with the `tracing` crate and might not know that they need to go to "automatic instrumentation" to find it.
Additionally, they already know what our "tracing" features do if they are familiar with that crate.
Let's add an alert so people can jump directly to that.


coolguyzone (approved):
Thanks for adding!

PR #13128, opened 2025-03-25
ref(feedback): add additional searchable props 

aliu39:
Adding fields that are missing here but suggested in feedback list searchbar (see https://github.com/getsentry/sentry/blob/master/static/app/utils/fields/index.ts#L2383-L2414)


aliu39:
Delaying this as we should test the validity of some of these searches, as well as the suggestions

aliu39:
Closing until this ticket's addressed https://github.com/getsentry/sentry/issues/88766

Issue #9625, opened 2024-04-08
Document how to setup source context in Rust


szokeasaurusrex:
### Core or SDK?

Platform/SDK

### Which part? Which one?

Rust SDK

### Description

How to correctly upload Rust debug files to obtain source context for errors in Sentry

### Suggested Solution

Add a page to document how to configure the `Cargo.toml` file to obtain debug files in the correct format, and explain how to use the Sentry CLI to upload debug files with sources.

Specifically, the `Cargo.toml` needs to contain the following:

```toml
# For source context on errors emitted from development builds
[profile.dev]
split-debuginfo = "packed"

# For source context on errors emitted from release builds
[profile.release]
split-debuginfo = "packed"
```

After configuring the `Cargo.toml` per the above, users need to build the code with `cargo build` (or `cargo build --release`, as desired). 

Once compiled, the debug files and source context can be uploaded with:

```sh
sentry-cli debug-files upload --include-sources /path/to/project
```

Roshan-R:
Thanks! This fixed the issue for me. It should definitely be added to the documentation

lcian:
Let's keep it open as the documented process only works for Linux and not other platforms.

PR #13151, opened 2025-03-28
feat(rust): document how to set up source context

lcian:
## DESCRIBE YOUR PR
Closes https://github.com/getsentry/sentry-docs/issues/9625

This is the process we use for relay.
It works on Linux and macOS, not sure yet how to do this on Windows but we can add it in the future.

### DIFF
--- a/docs/platforms/rust/common/source-context/index.mdx
+++ b/docs/platforms/rust/common/source-context/index.mdx
@@ -6,7 +6,7 @@ description: "Learn about showing your source code as part of stack traces."
 
 ## Prerequisites
 
-[Install](cli/installation/) and [Configure](cli/configuration/) Sentry CLI by providing an auth token, org and project.
+[Install](/cli/installation/) and [Configure](/cli/configuration/) Sentry CLI by providing an auth token, org and project.
 
 Then, enable full debug information for your release build:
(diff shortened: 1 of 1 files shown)


docs/platforms/rust/common/source-context/index.mdx
```diff
@@ -0,0 +1,40 @@
...
+Use Sentry CLI to create a source bundle:
+
+```bash
+sentry-cli difutil bundle-sources target/release/app.d
+mv target/release/app.src.zip ./app.src.zip
+```
+
+Finally, use Sentry CLI to upload both debug information files and sources to Sentry:
+```bash
+sentry-cli upload-dif ./app-debug.zip ./app.src.zip
```
Swatinem, line 39: It is not really necessary to zip the debug file.
You should be able to `debug-files upload --include-sources target/release/app.d` instead.

The `upload-dif` command is a legacy alias for the "more modern" `debug-files upload`. `--include-sources` can be used to directly upload sources without having to bundle them manually beforehand.

lcian: Thank you, updated!

Swatinem (approved):

PR #13153, opened 2025-03-28
fix(js): Fix hierarchy and examples for sample rate configuration

mydea:
This PR improves the "Tracing > Configure Sampling" page in two main ways:

1. It introduces proper headings for the two main sections - this way, you can link there, and it makes more sense structurally IMHO.
2. It fixes the `tracesSampler` examples - you should _always_ use `inheritOrSampleWith` for the default sample rate, otherwise you break propagation. We need to make sure to use this consistently in examples!

### DIFF
--- a/docs/platforms/javascript/common/tracing/configure-sampling/index.mdx
+++ b/docs/platforms/javascript/common/tracing/configure-sampling/index.mdx
@@ -10,14 +10,14 @@ Sentry's tracing functionality helps you monitor application performance by capt
 
 The JavaScript SDK provides two main options for controlling the sampling rate:
 
-1. Uniform Sample Rate (`tracesSampleRate`)
+### Uniform Sample Rate (`tracesSampleRate`)
 This option sets a fixed percentage of transactions to be captured:
 
 <PlatformContent includePath="/tracing/sample-rate" />
 
 With `tracesSampleRate` set to `0.25`, approximately 25% of transactions will be recorded and sent to Sentry. This provides an even cross-section of transactions regardless of where in your app they occur.
 
-2. Sampling Function (`tracesSampler`)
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/common/tracing/configure-sampling/index.mdx
```diff
@@ -10,14 +10,14 @@ Sentry's tracing functionality helps you monitor application performance by capt
...
-1. Uniform Sample Rate (`tracesSampleRate`)
+### Uniform Sample Rate (`tracesSampleRate`)
 This option sets a fixed percentage of transactions to be captured:
 
 <PlatformContent includePath="/tracing/sample-rate" />
 
 With `tracesSampleRate` set to `0.25`, approximately 25% of transactions will be recorded and sent to Sentry. This provides an even cross-section of transactions regardless of where in your app they occur.
 
-2. Sampling Function (`tracesSampler`)
+### Sampling Function (`tracesSampler`)
```
codyde, line 20: One thing im wondering, based on the feedback from the other Python PR, is if we should actually be changing the language of this section to suggest you can configure "A" traceSampler function; instead of implying this is a function we provide. 

The way it was written originally def implies that its a function we give - which isnt the case if I understand right - but that might also be an unnecessary detail. What do you think?

mydea: Yeah, that may make sense 🤔 IMHO we can still merge this PR and tweak the exact language we use in a follow up?

codyde: Yup - im good with that.

codyde (approved):

PR #13154, opened 2025-03-28
feat(js): Streamline links for explanations about tracing sampling

mydea:
Since we have a dedicated page about this, we should link there instead of to the generic options, which have less information than other places.

Generally, I wonder if it really makes sense to have the "Configure Sample Rate" page under "Set Up Tracing" as well as the "Sampling" page under configuration, which have almost-but-not-quite the same information as of now 🤔 IMHO one of these should stay and the other should go (personally I would vote for configuration/sampling and linking there from the tracing docs)

### DIFF
--- a/docs/platforms/javascript/common/tracing/index.mdx
+++ b/docs/platforms/javascript/common/tracing/index.mdx
@@ -4,7 +4,7 @@ description: "Learn how to enable tracing in your app."
 sidebar_order: 4000
 ---
 
-With [tracing](/product/insights/overview/), Sentry automatically tracks your software performance across your application services, measuring metrics like throughput and latency, and displaying the impact of errors across multiple systems. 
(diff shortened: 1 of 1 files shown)


codyde:
> Since we have a dedicated page about this, we should link there instead of to the generic options, which have less information than other places.
> 
> Generally, I wonder if it really makes sense to have the "Configure Sample Rate" page under "Set Up Tracing" as well as the "Sampling" page under configuration, which have almost-but-not-quite the same information as of now 🤔 IMHO one of these should stay and the other should go (personally I would vote for configuration/sampling and linking there from the tracing docs)

Yeah - this ones a tough one. Dhrumil had asked (and I agreed) to have a dedicated page for it within the tracing section because we def expect people to need to sample out their traces (high volume transactions and all), but to your point - it's valuable info across the board for other sample areas too. 

Its something we could leave in place and track, to see how much traffic hits and where user volume lands?

Either way - PR looks good; thanks for helping clean this up a bit. Approving for my side.

codyde (approved):
LGTM

PR #13168, opened 2025-03-31
feat: Update `Expandable` to use native `<details>` element

mydea:
The `<details>` element is a native HTML element that handles expanding/collapsing out of the box. This is more accessible, and has some additional important benefits for us:

1. It should be covered by search etc. as this is statically rendered now and just "hidden" via CSS.
2. The browser automatically searches in it when using CMD+F - it will auto-open respective details that contain the searched content.


src/components/expandable/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+//
+// <Callout
+//       level={level}
+//       title={title}
+//       Icon={isExpanded ? ChevronDownIcon : ChevronRightIcon}
+//       id={id}
+//       titleOnClick={toggleIsExpanded}
+//     >
+//       {isExpanded ? children : undefined}
+//     </Callout>
```
lforst, line 115: dead commented code

lforst (approved):
<3 love me some cmd-f

Lms24 (approved):

src/components/codeBlock/code-blocks.module.scss
```diff
@@ -19,7 +19,7 @@
     font-size: 0.85rem;
     border: 0;
     border-radius: 4px;
-    margin: 0 0 1rem;
+    margin: 0;
```
coolguyzone, line 22: Just curious, why are we getting rid of the bottom-margin here?

coolguyzone (approved):

src/components/codeBlock/code-blocks.module.scss
mydea: sorry, should have explained: This was a bit hard to work with, as this element is nested in a bunch of other divs, making it impossible to remove this, basically. Also, as far as I've seen this is always in _another_ div that _also_ has margin, so this was redundant - I scanned a bunch of pages and it did not seem to affect the styling, so hopefully this is good 🤞 
Without this change, having a code block in an expandable (see example in the component page) looked weird because there was always a margin on bottom.

PR #13169, opened 2025-03-31
Add tabs to reflect the new binaries available

smeubank:
https://github.com/getsentry/sentry/pull/85861

This PR only updates android. The prio of order of tabs could be changed, and this could be applied to flutter and iOS install in docs. PRs welcome:)

image


coolguyzone (approved):
Looks good from a content perspective! Also confirmed that the brew, npx, and OSX commands work as expected.

Issue #13159, opened 2025-03-31
Upgrade TableOfContents component to work with different structures


inventarSarah:
### Problem Statement

The TableOfContents component seems to be tailor-made for our SDK options pages (eg [here](https://docs.sentry.io/platforms/elixir/configuration/options/#core-options)).

We wanted to use this component on a different page ([React Router](https://docs.sentry.io/platforms/javascript/guides/react/features/react-router/v7/)) but since the structure is different, the component does not render as expected.

I did some testing, and found that:

\## heading 2
\### heading 3
\### heading 3

displays as:
- heading 2
  - heading 3 
  - heading 3

-> this is what we want

\## heading 2
\### heading 3
\### heading 3
\## heading 2
\## heading 2

displays as:
- heading 2
  - heading 3 
  - heading 3
 
-> it is not rendering the h2's that have no children

\## heading 2
\## heading 2
\## heading 2

displays as:
-> it does not render anything

\### heading 3
\### heading 3
\### heading 3

displays as:
- heading 3
  -  heading 3
  - heading 3
 
-> it renders a hierarchy even though there should be none

### Solution Brainstorm

I can imagine this component would be useful on different pages, so it would be great to make it work with different structures (see examples above) ❤
If we're doing this, we should also add a description to the component list in "Contributing to the Docs"

PR #13171, opened 2025-03-31
fix: Improve `<TableOfContents>` to allow empty sections

mydea:
Not sure why I even added this, but this seems unnecessary!

Closes https://github.com/getsentry/sentry-docs/issues/13159


inventarSarah (approved):
Awesome, thanks for the quick update -- I'll use the component in #13138 as soon as this is merged 🎉

PR #13174, opened 2025-03-31
fix(develop): clarify trace context's client_sample_rate

mjq:
## DESCRIBE YOUR PR
When reading the Trace Context docs, I assumed that `client_sample_rate` (like all the other fields) was expected to be set by the client. Instead, this field is generated in Relay. If a client sets this field, it will be ignored and overwritten. Added a note to the docs for this field to clarify how it is meant to be used.

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

### DIFF
--- a/develop-docs/sdk/data-model/event-payloads/contexts.mdx
+++ b/develop-docs/sdk/data-model/event-payloads/contexts.mdx
@@ -682,7 +682,7 @@ Additional information that allows Sentry to connect multiple transactions, span
 
 `client_sample_rate`
 
-: _Optional_. The client-side sample rate.
+: _Optional_. The client-side sample rate. This field will be populated by relay. Incoming values are ignored.
 
 - Example: `0.1`
 
--- a/develop-docs/sdk/data-model/event-payloads/contexts.mdx
+++ b/develop-docs/sdk/data-model/event-payloads/contexts.mdx
@@ -682,7 +682,7 @@ Additional information that allows Sentry to connect multiple transactions, span
 
 `client_sample_rate`
 
-: _Optional_. The client-side sample rate. This field will be populated by relay. Incoming values are ignored.
+: _Optional_. The client-side sample rate. This field will be populated by relay; incoming values are ignored.
 
 - Example: `0.1`
 


develop-docs/sdk/data-model/event-payloads/contexts.mdx
```diff
@@ -682,7 +682,7 @@ Additional information that allows Sentry to connect multiple transactions, span
 
 `client_sample_rate`
 
-: _Optional_. The client-side sample rate.
+: _Optional_. The client-side sample rate. This field will be populated by relay. Incoming values are ignored.
```
Dav1dde, line 685: I don't think this is fully correct, we have a code path which re-creates the DSC when the parent project is not in the same org or missing.

Maybe @jan-auer can also chime in?

mjq: Yes, this was the subject of @jan-auer's recent changes. The DSC generation happens [here](https://github.com/getsentry/relay/blob/6cd49932a662696eddfd022e69c9e7c4723313b4/relay-server/src/services/processor/dynamic_sampling.rs#L47-L77). But the sample rate is taken from the incoming DSC, never from the trace context's `client_sample_rate` - that field is never read, only [set](https://github.com/getsentry/relay/blob/6cd49932a662696eddfd022e69c9e7c4723313b4/relay-server/src/services/processor.rs#L1493-L1496).

Hopefully Jan can confirm.

coolguyzone, line 685:
```suggestion
: _Optional_. The client-side sample rate. This field will be populated by relay; incoming values are ignored.
```

coolguyzone (approved):

jan-auer (approved):

develop-docs/sdk/data-model/event-payloads/contexts.mdx
jan-auer: This is correct, it happens here: https://github.com/getsentry/relay/blob/cbf72c2930ad01ed0d780e36968088cd6a051878/relay-event-normalization/src/event.rs#L354-L356

cleptric:
@mjq reverted your change in https://github.com/getsentry/sentry-docs/pull/13325 and removed all none SDK context from the docs, as this is just a source of confusion about what the SDKs emit.

mjq:
@cleptric Makes sense. Thanks! 👍

PR #13175, opened 2025-03-31
JavaScript Tracing  Section Corrections

codyde:
## DESCRIBE YOUR PR
Cleaning up issues across the larger tracing improvements we made previously to improve technical accuracy, simplify examples, and provide a better user flow through. 

* Significant work on sampling examples 
* Removing a queue example from the tracing samples 
* Improving span metrics examples 
* Various clarity updates

## IS YOUR CHANGE URGENT?  

- [x] Urgent deadline (GA date, etc.): No immediate deadline but wanting to improve the overall flow.


docs/platforms/javascript/common/tracing/configure-sampling/index.mdx
```diff
@@ -11,15 +11,15 @@ Sentry's tracing functionality helps you monitor application performance by capt
 The JavaScript SDK provides two main options for controlling the sampling rate:
 
 1. Uniform Sample Rate (`tracesSampleRate`)
-This option sets a fixed percentage of transactions to be captured:
+`tracesSampleRate` is floating point value 0.0 and 1.0, which controls the probability that a transaction will be sampled.
 
 <PlatformContent includePath="/tracing/sample-rate" />
 
-With `tracesSampleRate` set to `0.25`, approximately 25% of transactions will be recorded and sent to Sentry. This provides an even cross-section of transactions regardless of where in your app they occur.
+With `tracesSampleRate` set to `0.25`, each transaction in your application is randomly sampled with a probability of `0.25`, so you can expect that one in every four transactions will be sent to Sentry.
```
mydea, line 18:
```suggestion
With `tracesSampleRate` set to `0.25`, each transaction in your application is randomly sampled with a probability of 25%, so you can expect that one in every four transactions will be sent to Sentry.
```

maybe a bit easier to understand?

docs/platforms/javascript/common/tracing/configure-sampling/index.mdx
```diff
@@ -96,98 +119,16 @@ tracesSampler: (samplingContext) => {
...
-
-**Note:** The `inheritOrSampleWith` helper was introduced in version 9 of the SDK. For earlier versions, you can implement similar logic manually using the `parentSampled` property.
-
 ## Sampling Decision Precedence
 
 When multiple sampling mechanisms could apply, Sentry follows this order of precedence:
 
-- If `tracesSampler` is defined, its decision is used (can consider parent sampling)
-- If no `tracesSampler` but parent sampling is available, parent decision is used
+- If a sampling decision is passed to `startSpan`, that decision is used
```
mydea, line 126:
```suggestion
```

this is not correct anymore for JS, you cannot pass a sampling decision like this!

docs/platforms/javascript/common/tracing/configure-sampling/index.mdx
```diff
@@ -96,98 +119,16 @@ tracesSampler: (samplingContext) => {
...
-// Add to your custom request (example using WebSocket)
-webSocket.send(JSON.stringify({
-  message: "Your data here",
-  metadata: {
-    sentryTrace: sentryTraceHeader,
-    baggage: sentryBaggageHeader
-  }
-}));
-```
+- If none of the above are set, no transactions are sampled. This is equivalent to setting `tracesSampleRate` to `0.0`.
```
mydea, line 130: hmm this is kind of true but misleading - it is _not_ the same to omit this setting vs. setting it to `0`. The former will completely disable tracing, not propagating any tracing decision. The latter will ensure no spans are created but also that no downstream spans will be sampled because they will inherit a "no" sampling decision.

docs/platforms/javascript/common/tracing/configure-sampling/index.mdx
codyde: So originally i had gone with percentage also, by in the Python PR that @sfanahata was working on, @szokeasaurusrex mentioned that a percentage was wrong to use. I def agree that percentage translates better to users.

szokeasaurusrex: 25% is not wrong to use, because a sample rate of 0.25 means that there is a 25% probability of sampling each transaction. 

My comment on the Python PR was referring to calling the traces sample rate the "percentage" of transactions which will be captured. Calling this rate a percentage is wrong, because it is specified as a fractional value between zero and one, not as a percentage. E.g. it is invalid to set the sample rate to 25; if you want to capture 25% of transactions you have to set the rate to 0.25. 

Sorry for the confusion. Does the distinction make sense now?

docs/platforms/javascript/common/tracing/configure-sampling/index.mdx
codyde: I adjusted this to add that additional clarity! Thank you!

coolguyzone (approved):
🫡

Issue #13073, opened 2025-03-20
Table of Contents shows headings hidden by OnboardingOptions component


inventarSarah:
### Steps to Reproduce

1.  Go to our Node.js Quick Start guide: [https://docs.sentry.io/platforms/javascript/guides/node/](https://docs.sentry.io/platforms/javascript/guides/node/)
2.  Make sure **not** to select any onboarding options!
3.  Look at the main content area under "Step 4: Verify your setup" -> there is no "Tracing" subheading
4.  Look at the table of contents (TOC) on the right sidebar under "Step 4: Verify your setup" -> the "Tracing" subheading is visible

### Expected Result

If a heading is hidden by the `OnboardingOptions` component, it should also be hidden from the table of contents.

### Actual Result

Tracing is visible in the Toc:

![Image](https://github.com/user-attachments/assets/5263daa8-ef14-4071-bb04-e133b3dac819)

### Additional Info

In the example described above, we used the `OnboardingOptions` component to hide/show content

PR #13201, opened 2025-04-02
fix(platform): Ensure hidden sections are not show in sidebar TOC

mydea:
This updates the sidebar table of content (TOC) component to not show hidden headings:

https://github.com/user-attachments/assets/1d06f6ce-f44d-42b9-bceb-c0a9feaffc25

For this, we check `element.checkVisibility()` which is a _relatively_ new [API](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility#browser_compatibility).

Additionally, we register a mutation observer to check if the visible headings change, and if they do we update the sidebar.

Closes https://github.com/getsentry/sentry-docs/issues/13073

### DIFF
--- a/src/components/sidebarTableOfContents/index.tsx
+++ b/src/components/sidebarTableOfContents/index.tsx
@@ -71,6 +71,36 @@ function buildTocTree(toc: TocItem[]): TocItem[] {
   return items;
 }
 
+function getTocItems(main: HTMLElement) {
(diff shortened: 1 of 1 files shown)


inventarSarah (approved):
looks great 🎉 
(tested it on the node.js quick start guide and it works as expected 👌)

PR #13203, opened 2025-04-02
feat(develop): Add `page` context

mydea:
This adds a new context to develop docs:

The new `page` context should be used to add information about the current page that an event originated on. It should be used for client SDKs instead of putting this information into `event.request`.

Relay will need to extract data from this (if available) instead of `event.request`.


develop-docs/sdk/data-model/event-payloads/contexts.mdx
```diff
@@ -778,6 +778,42 @@ envelope endpoint.
...
+
+: **Required.** The URL of the page that the event occurred on.
+
+- Example: `https://sentry.io/issues?page=1`
+
+`user_agent`
+
+: _Optional_. The user agent of the page that the event occurred on.
+
+- Example: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/123.0.0.0 Safari/537.36`
```
cleptric, line 795: How can a "page" have a user agent?

cleptric:
Overall not a big fan. A page does not exist on mobile, and the naming could be better.

develop-docs/sdk/data-model/event-payloads/contexts.mdx
mydea: hmm, how do we phrase this? We need to send the user agent somehow, as the server infers the os/browser stuff from it 🤔 

We could make it `page_request`, would that make more sense?

mydea:
> A page does not exist on mobile, and the naming could be better.

An `app` also does not exist on web but we still have an `app` context 😅 We basically want something similar but for the web.

cleptric:
Well, you wrote client SDKs. If this is just for Web FE, it's a different story.

develop-docs/sdk/data-model/event-payloads/contexts.mdx
mydea: After thinking about this, I think it makes more sense to split this out into https://github.com/getsentry/sentry-docs/pull/13205

develop-docs/sdk/data-model/event-payloads/contexts.mdx
```diff
@@ -778,6 +778,56 @@ envelope endpoint.
 }
 ```
 
+## Page Context
+
+Page context contains information about the page that the event occurred on.
+
+`full.url`
```
lforst, line 785: It feels a bit overkill for me to store everything in separate fields. What's wrong with just having a `url` field?

lforst:
Generally I am strongly in favor of this because request contexts make literally no sense on frontend events.

develop-docs/sdk/data-model/event-payloads/contexts.mdx
mydea: cc @cleptric had concerns about pii stripping etc!

mydea:
> I would recommend heavily analyzing what needs to be done for the entire stack before committing to just moving data around, just because the naming may not be 'optimal' for a frontend SDK.

The naming is not just not optimal, it is simply wrong. there are not really any headers, the http method is meaningless, nobody thinks about this as HTTP Request information.

In regards to backwards compatibility: We'll continue to send the data as `request` data for the time being until the next major or so, likely. However, relevant data is already inferred from this - e.g. there is a `url` tag that is inferred, etc - which would continue to be inferred, just from the new place.

As far as I can tell, neither on issues stream, nor on trace explorer, nor alerts, allow to filter by `request.xxx` data - you can already only filter on inferred data (e.g. the url tag, ...). So IMHO these should not really be impacted by moving this. I would assume the request itself is not indexed, as I can't filter it anywhere.

develop-docs/sdk/data-model/event-payloads/contexts.mdx
cleptric: https://github.com/getsentry/sentry-javascript/pull/7667 for context. We imo should not emit the full URL.

mydea: I updated this to be: `url`, `http.query` and `http.fragment`.

lforst:
Can we also put the user agent on the page context so that relay and the product can use it to display the device and browser?

develop-docs/sdk/data-model/event-payloads/contexts.mdx
lforst: This still feels like a mess because `url` sounds like it contains the full url.

How about we lean it on the `window.location` API, which I'd assume is something people can relate to?:

- `location.origin`
- `location.pathname`
- `location.search`
- `location.hash`

From that you can reconstruct the the entire URL (minus some PII redactions).

coolguyzone (approved):

mydea:
@cleptric are we good to merge this? Do we still need to adjust something around the url stuff?

PR #13204, opened 2025-04-02
feat(develop): Add `outgoing_request` context

mydea:
This should replace `event.request` on the client completely.

Draft on top of https://github.com/getsentry/sentry-docs/pull/13203

### DIFF
--- a/develop-docs/sdk/data-model/event-payloads/contexts.mdx
+++ b/develop-docs/sdk/data-model/event-payloads/contexts.mdx
@@ -778,6 +778,42 @@ envelope endpoint.
 }
 ```
 
+## Page Context
+
+Page context contains information about the page that the event occurred on.
+
+`url`
+
+: **Required.** The URL of the page that the event occurred on.
+
+- Example: `https://sentry.io/issues?page=1`
+
+`user_agent`
+
+: _Optional_. The user agent of the page that the event occurred on.
+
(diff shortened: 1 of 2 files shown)


develop-docs/sdk/data-model/event-payloads/contexts.mdx
```diff
@@ -814,6 +814,55 @@ Page context contains information about the page that the event occurred on.
...
+
+- Example: `{ "name": "John Doe", "email": "john.doe@example.com" }`
+
+### Example Outgoing Request Context
+
+```json
+{
+  "contexts": {
+    "outgoing_request": {
+      "url": "https://sentry.io/api/0/projects/sentry/sentry/issues/?page=1",
```
cleptric, line 852: We do not store query parameters this way. https://develop.sentry.dev/sdk/expected-features/data-handling/#spans

bruno-garcia:
> This should replace event.request on the client completely.

if the product will have UX based in the new outgoing request field (which makes sense) could we have something in Relay that moves event.request to the new field when we know it's a client event? the heuristics to be defined (sdk name? other fields etc)

develop-docs/sdk/data-model/event-payloads/contexts.mdx
mydea: Ah, I based this on https://develop.sentry.dev/sdk/data-model/event-payloads/request/#attributes:

> `url` Optional. The URL of the request if available. The query string can be
declared either as part of the url, or separately in query_string.

We can split it out for sure!

PR #13205, opened 2025-04-02
feat(develop): Allow to pass `user_agent` as browser context

mydea:
Related to https://github.com/getsentry/sentry-docs/pull/13203

Today, the browser context is inferred in relay from `event.request.headers['user-agent']`. This is slightly weird in the context of a browser event, as the page is not really the request - we want to replace this with something more explicit.

In order to make this possible, we propose to allow client SDKs to send a partial `browser` context with only user-agent, and have relay infer the rest of the data from this, if it exists (instead of from `event.request`).

### DIFF
--- a/develop-docs/sdk/data-model/event-payloads/contexts.mdx
+++ b/develop-docs/sdk/data-model/event-payloads/contexts.mdx
@@ -402,6 +402,12 @@ The `type` and default key is `"browser"`.
 
 : _Optional_. Version string of the browser.
 
+`user_agent`
+
+: _Optional_. The user agent of the browser. If this is set, `name` may also be omitted (as it will be derived from the user agent).
+
+- Example: `"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/123.0.0.0 Safari/537.36"`
+
 ## GPU Context
 
 GPU context describes the GPU of the device.


lforst (approved):
sgtm. Before we merge this we should actually implement this in Relay.

lforst:
> After talking with @Dav1dde we will actually _not_ do that, because it is hard for relay to partial-update contexts. Instead, relay can use the user-agent from the browser envelope request itself, which should already be here - which is great, because it means the SDK does not need to do/send anything 💪

Scary, in the case of tunneling this information is lost.

cleptric (approved):

cleptric (changes requested):
We actually need to figure out how this will be handled in Relay fist.
- If there is no browser context, Relay creates one
- How does the presence of a `user_agent` fit in in the current behavior
- How do we merge partial context

lforst:
I think the plan of action for relay is:

- Drop the requiredness of the `name` field in the Browser context.
- If `name` and `version` are not defined and `user_agent` is defined, backfill `name` and `version` in the browser context.
- Keep creating browser context when there is none and the Request interface contains a user agent.

coolguyzone (approved):

mydea:
@cleptric are we good with merging this? I think we agreed that this makes sense from a spec-perspective, or should we merge the relay change before we merge this in develop docs?

cleptric (approved):
I'm good with merging this, if we don't forgot to also add it in Relay 😬

PR #13227, opened 2025-04-04
fix(js): Fix span metrics examples

mydea:
This fixes some of the code examples around span metrics in JS.

### DIFF
--- a/docs/platforms/javascript/common/tracing/span-metrics/examples.mdx
+++ b/docs/platforms/javascript/common/tracing/span-metrics/examples.mdx
@@ -19,60 +19,50 @@ This guide provides practical examples of using span attributes and metrics to s
 **Solution:** Track the entire file processing pipeline with detailed metrics at each stage, from client-side upload preparation through server-side processing.
 
 **Frontend Instrumentation:**
+
 ```javascript
 // Client-side file upload handling
 Sentry.startSpan(
   {
-    name: 'Client File Upload',
-    op: 'file.upload.client',
+    name: "Client File Upload",
+    op: "file.upload.client",
     attributes: {
       // Static details available at the start
-      'file.size_bytes': 15728640,  // 15MB
-      'file.type': 'image/jpeg',
-      'file.name': 'user-profile.jpg',
-      'client.compression_applied': true
-    }
+      "file.size_bytes": 15728640, // 15MB
+      "file.type": "image/jpeg",
+      "file.name": "user-profile.jpg",
+      "client.compression_applied": true,
+    },
   },
-  async () => {
-    // Get the current active span to update during upload
-    const span = Sentry.getActiveSpan();
-    
+  async (span) => {
     try {
       // Begin upload process
       const uploader = new FileUploader(file);
-      
+
       // Update progress as upload proceeds
-      uploader.on('progress', (progressEvent) => {
-        if (span) {
-          span.setAttribute('upload.percent_complete', progressEvent.percent);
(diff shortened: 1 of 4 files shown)


docs/platforms/javascript/common/tracing/span-metrics/examples.mdx
```diff
@@ -19,86 +19,82 @@ This guide provides practical examples of using span attributes and metrics to s
...
-        span.setAttribute('upload.success', false);
-        span.setAttribute('upload.error_type', error.name);
-        span.setAttribute('upload.error_message', error.message);
-        span.setStatus({ code: 'ERROR' });
-      }
+      span.setAttribute("upload.success", false);
+      span.setAttribute("upload.error_type", error.name);
+      span.setAttribute("upload.error_message", error.message);
+
       throw error;
```
Lms24, line 66: m: I'd rather use our error monitoring product here instead of adding metrics. Maybe there's a case for `upload.error_type` if we want some kind of frequency/distribution gauge but given errors are trace-connected, capturing an error makes more sense to me.

Also, if we avoid rethrowing the error but capture here, we can associate the error with this very span and its scope data, rather than whatever outer scope is active where we'd catch it.

```suggestion
      span.setAttribute("upload.success", false);
      Sentry.captureException(error);
```

docs/platforms/javascript/common/tracing/span-metrics/examples.mdx
```diff
@@ -121,121 +117,116 @@ The frontend span initiates the trace and handles the file upload process. It pr
...
-        span.setStatus({ code: 'ERROR' });
-      }
+      span.setAttribute("error", true);
+      span.setAttribute("error.type", error.name);
+      span.setAttribute("error.message", error.message);
+      span.setAttribute(
+        "error.is_rate_limit",
+        error.code === "rate_limit_exceeded"
+      );
       throw error;
```
Lms24, line 230: m: same thought as above here:

```suggestion
      span.setAttribute("error", true);
      Sentry.captureException(error);
```

docs/platforms/javascript/common/tracing/span-metrics/index.mdx
```diff
@@ -74,42 +74,35 @@ To consistently add metrics across all spans in your application, you can use th
...
-  }
+    return span;
+  },
 });
 ```
 
 ## Best Practices for Span Metrics
 
 1. **Metric Naming**
+
```
Lms24, line 92: out of scope for this PR but as discussed yesterday, we should rethink this section.

docs/platforms/javascript/common/tracing/span-metrics/examples.mdx
mydea: re-throwing here should be fine and it should be correctly assigned to this span then anyhow. I agree on the error details being captured here, though, maybe we can just remove the whole `catch` block here 🤔

codyde: Thats a fair point - im fine wiht that. My logic was that if someone was reviewing the span metrics directly as part of troubleshooting - having the visibility close to line of sight would be useful, but if we're capturing it as part of product they are going to see the error. Im fine with the change.

docs/platforms/javascript/common/tracing/span-metrics/index.mdx
codyde: Yeah - im out of scope for this PR, but we're aligned that ill be removing this section.

codyde (approved):
Appreciate the help on this, LGTM!

PR #13233, opened 2025-04-04
feat(react-native): Add Browser replay init to the RN docs

krystofwoldrich:
RN SDK supports browser replay, but it was not part of the docs.

### DIFF
--- a/docs/platforms/react-native/session-replay/index.mdx
+++ b/docs/platforms/react-native/session-replay/index.mdx
@@ -33,7 +33,7 @@ pnpm add @sentry/react-native
 
 To set up the integration, add the following to your Sentry initialization.
 
-```javascript
+```javascript {tabTitle:Mobile}
 import * as Sentry from "@sentry/react-native";
 
 Sentry.init({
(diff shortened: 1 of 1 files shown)


docs/platforms/react-native/session-replay/index.mdx
```diff
@@ -44,6 +44,24 @@ Sentry.init({
...
+
+Sentry.init({
+  dsn: "___PUBLIC_DSN___",
+  replaysSessionSampleRate: 0.1,
+  replaysOnErrorSampleRate: 1.0,
+  integrations: (integrations) => {
+    if (Platform.OS === 'web') {
+      integrations.push(Sentry.browserReplayIntegration());
+    }
+    integrations.push(Sentry.mobileReplayIntegration());
```
kahest, line 59: should this be in an else block?

coolguyzone (approved):

docs/platforms/react-native/session-replay/index.mdx
krystofwoldrich: The mobile replay can be added on any platform. So for simplicity it doesn't have to be.

PR #13237, opened 2025-04-04
feat(js): Update enriching events page

mydea:
This updates the "Enriching Events" page in the JS docs to be more descriptive, and actually guide you through the different types of data you may add to your Sentry events.

It also:

1. Adds more details to `setUser` API
2. Adds more details to `setContext` API
3. Adds the missing `sendDefaultPii` option
4. Removes the "Identifying Users" page in favor of just using the `setUser` API page
5. Removes the "Contexts" page which was really confusing, as it mixed both explanation of contexts (which is now in `setContext` API page) as well as general "context" stuff like clearing scope data etc (which does not work anyhow because you need to clear the isolation scope etc... so just removing this for now).


docs/platforms/javascript/common/configuration/options.mdx
```diff
@@ -63,6 +63,12 @@ Sets the URL that will be used to transport captured events. This can be used to
 
 </SdkOption>
 
+<SdkOption name="sendDefaultPii" type='boolean' defaultValue='false'>
+
+Set this option to `true` to send default PII data to Sentry. Amongst other things, enabling this will enable automatic IP address collection on events.
```
s1gr1d, line 68: Apparently, "amongst" is more common in british english and also more formal. 
```suggestion
Set this option to `true` to send default PII data to Sentry. Among other things, enabling this will enable automatic IP address collection on events.
```

Lms24 (approved):

PR #13243, opened 2025-04-04
fix: platformSection spacing

### LINKED ISSUE #13134: `<PlatformSection>` adds line spacing in list
### Steps to Reproduce

- add `<PlatformSection>` to include/exclude specific platform in list
- navigate to the page
  - e.g. on https://docs.sentry.io/platforms/apple/guides/ios/features/experimental-features/, the bullet item for App Hangs V2 is excluded for some platforms

### Expected Result

- list renders without any additional spacing

### Actual Result

- list has additional spacing around the `<PlatformSection>`
Image

### Additional Info

This is because `<PlatformSection>` breaks the list (`<ul>`) and the list has additional space around it, while the list items don't. I hope there's an easy way to fix this :) 

Image

a-hariti:
closes #13134


kahest:
@a-hariti thanks for tackling this - seems like the preview fails to deploy, prob need to merge in main? in any case, I tested/previewed locally, looking good!

PR #13247, opened 2025-04-04
fix: All Sentry.feedbackIntegration examples include the Sentry.init() statement for clarity

ryan953:


platform-includes/user-feedback/manual-injection/javascript.mdx
```diff
@@ -1,9 +1,15 @@
...
+Sentry.init({
+  dsn: "___PUBLIC_DSN___",
+  integrations: [Sentry.feedbackIntegration({
+    // Disable the injection of the default widget
+    autoInject: false,
+  })],
 });
 
+// Get the instance returned by `feedbackIntegration()`
+const feedback = Sentry.getFeedback();
```
aliu39, line 11:
```suggestion
const feedback = Sentry.getFeedback()!;
```
Maybe?

aliu39 (approved):

platform-includes/user-feedback/manual-injection/javascript.mdx
ryan953: that's a bit of typescript syntax. It would depend on if someone wants to accept that the return is nullable, or force it not to be

if it's nullable then they could also refine it with an if-statement:
```
const feedback = Sentry.getFeedback();
if (feedback) {
  const widget = feedback.createWidget();
}

aliu39 (approved):

PR #13252, opened 2025-04-04
feat(feature flags): Add feature flag integrations to js frameworks with Browser Integrations sections

ryan953:
The change is on this page, for all the frameworks i found that have a specific Browser Integrations section
https://docs.sentry.io/platforms/javascript/configuration/integrations/

I've grouped all the FF integrations together. This is an improvement as-is because we're giving more reach to the FF integrations for JS frameworks with a browser component.

Pages will look like this now:

SCR-20250404-ndiv


aliu39:
Personally I like keeping them alphabetical, if theres no dedicated section to nest flags.

I think how it's structured now is auto enabled on top, otherwise alphabetical

ryan953:
> Personally I like keeping them alphabetical, if theres no dedicated section to nest flags.
> 
> I think how it's structured now is auto enabled on top, otherwise alphabetical

oh, i did notice that the whole thing wasn't sorted but not that it's in two parts. i can sort it to be auto-enabled vs. the-rest.

aliu39 (approved):

PR #13255, opened 2025-04-07
feat(js): Remove Performance metrics page

mydea:
This is not really needed anymore, folks should just use span metrics. I also noticed a redirect was missing for the non-guide variant of the page.


Dhrumil-Sentry (approved):

Lms24 (approved):
As discussed, I think some kind of note that `setMeasurement` users should ideally switch to `setAttribute` would be great but not a deal breaker for me.

Related, we should deprecate the APIs SDK-side as discussed today once we know the unit replacement strategy.

mydea:
I added a note at the bottom of the span metrics page explaining that measurements are deprecated and span metrics should be used instead :)

PR #13258, opened 2025-04-07
fix(platform): Replace incorrect usage of `<h6>`

mydea:
We use this quite a lot to just generically mean "heading", but this is incorrect - a h6 has to be structure below a h5, not just randomly. This PR changes this to use a different heading that makes more sense or just a div+class instead. Nothing should change visually.


a-hariti (approved):
Nice!

PR #13306, opened 2025-04-09
docs(sdks): Batch Processor

philipphofmann:
This PR migrates the [SpansAggregator RFC](https://github.com/getsentry/rfcs/blob/c76980940adeb3d6104da699c93885a0eba4c165/text/0126-sdk-spans-aggregator.md) to the develop docs. As logs also have to use some aggregation, we renamed the SpansAggregator to BatchProcessor so it can be used with any type of telemetry data.


AbhiPrasad (approved):
Nice! Really happy with where we're going with this.

Now that we have this, I feel like these are the next extension items. We can put them into an issue and tackle them one at a time though.

1. figuring out if we need priority queue in the processor
2. optimizing for trace completeness (related to 1)
3. detailing behaviour during crashes

cleptric (approved):

PR #13307, opened 2025-04-09
ref(platform): Update sidebar components structure

mydea:
This was a bit tricky to follow/understand, and should hopefully be clearer now.

I noticed this while trying to change some sidebar styling, that this is pretty hard to reason about.

Now, this PR:

* Removes options on `DynamicNav` that were not actually used anymore at all
* Combines `withChevron` and `collapse` into a single `collapsible` property, as we always use this together anyhow (except in one place in develop docs, where IMHO it is fine to also be consistent...)
* Removed this construct where we passed in the header class name - we can just import the style directly there, making this consistent and easier to understand/style.
* Removed one code path in `DynamicNav` handling if no `parentNode` exists - this is never the case as far as I can tell, we can just bail out in this theoretical case.

### DIFF
--- a/src/components/sidebar/dynamicNav.tsx
+++ b/src/components/sidebar/dynamicNav.tsx
@@ -4,9 +4,12 @@ import {serverContext} from 'sentry-docs/serverContext';
 import {sortPages} from 'sentry-docs/utils';
 import {getUnversionedPath, VERSION_INDICATOR} from 'sentry-docs/versioning';
 
-import {NavChevron} from './sidebar/navChevron';
-import {SidebarLink} from './sidebarLink';
-import {SmartLink} from './smartLink';
+import styles from './style.module.scss';
+
+import {SidebarLink} from '../sidebarLink';
+import {SmartLink} from '../smartLink';
+
+import {NavChevron} from './navChevron';
 
 type Node = {
   [key: string]: any;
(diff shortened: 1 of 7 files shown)


src/components/sidebar/dynamicNav.tsx
```diff
@@ -141,10 +132,8 @@ export function DynamicNav({
   rootBits.forEach(bit => {
     entity = currentTree.find(n => n.name === bit);
     if (!entity) {
-      if (!suppressMissing) {
-        // eslint-disable-next-line no-console
-        console.warn(`Could not find entity at ${root} (specifically at ${bit})`);
-      }
+      // eslint-disable-next-line no-console
+      console.warn(`Could not find entity at ${root} (specifically at ${bit})`);
```
chargome, line 136: we could theoretically also throw here as this is not a client component

src/components/sidebar/dynamicNav.tsx
```diff
@@ -155,44 +144,39 @@ export function DynamicNav({
...
     title = entity.node.context.sidebar_title || entity.node.context.title || '';
   }
-  const parentNode = entity.children
-    ? entity.children.find((n: EntityTree) => n.name === '')
-    : null;
+  const parentNode = entity.children?.find((n: EntityTree) => n.name === '');
+
+  if (!parentNode) {
+    // eslint-disable-next-line no-console
+    console.warn(`Could not find parentNode at ${root}`);
```
chargome, line 151: same here

chargome (approved):
Looks good, thank you!

PR #13329, opened 2025-04-10
Fix uglify example command

vanschelven:
uglify's `-o` should point to the minified output, not to the sourcemap


chargome (approved):

chargome:
@vanschelven thanks for the contribution! Could you please rebase on master so our CI is happy 🙏

Issue #13322, opened 2025-04-10
Something is broken with sourcemap uploading and hosting publicly


sfanahata:
### SDK

JavaScript SDK

### Description

I'm getting a [lint 404 failure on my PRs](https://sentry.slack.com/archives/C8H0BQRDJ/p1744238576577219) in the sentry-docs repo, and I think it's because the `notsupported` list was changed or messed up in a file or two. In [hosting publicly](https://github.com/getsentry/sentry-docs/blob/master/docs/platforms/javascript/common/sourcemaps/uploading/hosting-publicly.mdx?plain=1), this is the list: 
```
  - javascript.node
  - javascript.aws-lambda
  - javascript.azure-functions
  - javascript.connect
  - javascript.express
  - javascript.fastify
  - javascript.gcp-functions
  - javascript.hapi
  - javascript.hono
  - javascript.koa
  - javascript.nestjs
```

But in the [uploading index.mdx](https://github.com/getsentry/sentry-docs/blob/master/docs/platforms/javascript/common/sourcemaps/uploading/hosting-publicly.mdx?plain=1), this is the list

```
  - javascript.nextjs
  - javascript.remix
  - javascript.solidstart
  - javascript.sveltekit
  - javascript.astro
  - javascript.nuxt
```

I can't see the hosting publicly doc in the navigation for the `notsupported` list in the uploading index file, but I can still [navigate directly to it](https://docs.sentry.io/platforms/javascript/guides/astro/sourcemaps/uploading/hosting-publicly/). And then when I click on the `upload them` link in that doc, I get a `page not found`. 

Can someone help make the `notsupported` lists right?

mydea:
I guess when an index page is not accessible in a platform, we should make all child pages not accessible as well? Does that make sense @chargome ?

chargome:
Yeah I'll have a go at excluding child pages 👍 – potentially rising more 404s though

chargome:
https://github.com/getsentry/sentry-docs/pull/13339 fixed this for now

PR #13348, opened 2025-04-11
feat(platform): Inherit support decision from parent pages

chargome:
This change will update the logic for rendering pages based on the `supported` and `notSupported` frontmatter flags:

- If the page itself has a support decision this will always take precedence
- Else we **only** inspect the support decision of the direct parent (e.g. `content/index.mdx` is the direct parent of `content/subpage.mdx`) 

This way we prevent rendering child pages of pages that are already excluded.

ref https://github.com/getsentry/sentry-docs/issues/13322

### DIFF
--- a/src/mdx.ts
+++ b/src/mdx.ts
@@ -41,7 +41,7 @@ const isSupported = (
   frontmatter: FrontMatter,
   platformName: string,
   guideName?: string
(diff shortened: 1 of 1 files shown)


chargome:
Closing as stale for now

PR #13351, opened 2025-04-11
feat(js): Update product pages to drop `Set up` prefix in navigation

mydea:
Also make sure to use this in breadcrumbs instead of the actual page heading.

Note that the pages themselves keep their "Set up XXX" heading, which makes sense IMHO.

### DIFF
--- a/docs/platforms/javascript/common/crons/index.mdx
+++ b/docs/platforms/javascript/common/crons/index.mdx
@@ -1,5 +1,6 @@
 ---
 title: Set Up Crons
+sidebar_title: Crons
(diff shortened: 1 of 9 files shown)


sfanahata:
@mydea - I'm wondering about doing this for at least Python as well before merging this change. I should have some time next week to work on it if you don't get to it first.

coolguyzone (approved):
Thanks! This feature will be very useful. Code looks good to me, works on preview.

chargome (approved):

mydea:
Done fully here instead: https://github.com/getsentry/sentry-docs/pull/13392

PR #13352, opened 2025-04-11
feat(js): Update profiling pages to avoid unnecessary nesting

mydea:
For browser/node SDKs, we do not need to have a higher-level "Profiling" page that is empty with just a single sub-page "Browser Profiling" / "Node Profiling".

Instead, now those will simply have a single "Profiling" page with the respective content. Only meta SDKs have a landing page with sub pages.

Note that this may break some links, e.g. when somebody used to link to `express/profiling/node-profiling` 🤔 We'd have to manually add a redirect for every single guide here, not sure if we want this? 😬 as we cannot blanket redirect e.g. `/guides/:guide/profiling/profiling-node` because this page _does_ exist in some places.

Or alternatively we rename the new pages to something else (e.g. `profiling/node` and `profiling/browser`), then we can blanket-redirect again - WDYT?


chargome:
I would either rename the pages to have solid redirects or simply add hardcoded redirects for the guides that changed (e.g. https://docs.sentry.io/platforms/javascript/profiling/browser-profiling/ does have quite some traffic on plausible but would result in 404 now)

mydea:
> I would either rename the pages to have solid redirects or simply add hardcoded redirects for the guides that changed (e.g. https://docs.sentry.io/platforms/javascript/profiling/browser-profiling/ does have quite some traffic on plausible but would result in a 404 with this change)

Makes sense - I renamed them to `profiling/browser` and `profiling/node` and added general redirects to the `profiling` page!

chargome (approved):
🚀

PR #13354, opened 2025-04-11
fix(style): Ensure scrollbar is consistent

mydea:
This removes the "hack" to only show the scrollbar on hover. Which is nice in theory, but got really jumpy now.
IMHO it is ok to just always show this, def. better than the current jumpy version!

I also aligned the padding on top so it is consistent around for the sidebar.

## Previously

https://github.com/user-attachments/assets/f5fe550e-2775-4675-858d-2b256ee46f88

## With this change

https://github.com/user-attachments/assets/354034ec-64f5-43f8-aed1-8bee55ffcd6c

### DIFF
--- a/src/components/sidebar/index.tsx
+++ b/src/components/sidebar/index.tsx
@@ -81,11 +81,11 @@ export async function Sidebar({path, versions}: SidebarProps) {
       });
 
   return (
-    <aside className={styles.sidebar}>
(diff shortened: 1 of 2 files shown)


a-hariti (approved):

chargome (approved):
💅 ✨

Issue #13350, opened 2025-04-11
Platform search is buggy


mydea:
### Steps to Reproduce

1. Search for `hon` - you get a correctly filtered list
2. Now extend to `hono` - suddenly the filtered list becomes longer and includes weird stuff?

https://github.com/user-attachments/assets/359e266c-6c66-4712-b2c8-1ff7c4006448

### Expected Result

Search should work as expected.

### Actual Result

Weird results

PR #13358, opened 2025-04-11
feat(platform): Apply stricter matching in platform selector

chargome:
closes https://github.com/getsentry/sentry-docs/issues/13350


mydea (approved):
I don't know what the acronym matcher does, but it def. seems more correct 👍

PR #13408, opened 2025-04-16
docs(js): Add `sendDefaultPii: true` to all init calls

chargome:


docs/platforms/javascript/guides/astro/index.mdx
```diff
@@ -94,11 +94,15 @@ export default defineConfig({
...
 
-```javascript {filename:sentry.client.config.js} {"onboardingOptions": {"performance": "7,10-13", "session-replay": "8,14-21"}}
+```javascript {filename:sentry.client.config.js} {"onboardingOptions": {"performance": "11,14-17", "session-replay": "12,18-25"}}
 import * as Sentry from "@sentry/astro";
 
 Sentry.init({
   dsn: "___PUBLIC_DSN___",
+  
+  // Adds request headers and IP for users,
+  // visit: https://docs.sentry.io/platforms/javascript/data-management/data-collected/ for more info
```
mydea, line 104: we do not have this page yet 😢 I think I would just leave this link away for now, just leave the first line comment here?

chargome: Could we replace it with something else maybe? Otherwise ALL onoboarding options will change again and then once we have the link, possibly again

mydea: We could link to this: https://docs.sentry.io/platforms/javascript/guides/react/configuration/options/#sendDefaultPii

but we should make sure to use the correct guide link for each one, not use the generic `javascript` platform IMHO!

stephanie-anderson:
Checks are hanging, will bypass those roles. 

Screenshot 2025-04-16 at 21 23 53

docs/platforms/dart/guides/flutter/manual-setup.mdx
```diff
@@ -74,7 +74,7 @@ Future<void> main() async {
     await SentryFlutter.init(
       (options) {
         options.dsn = '___PUBLIC_DSN___';
-        // Adds request headers and IP for users,
+        // Adds request headers and IP for users, for more info visit:
         // visit: https://docs.sentry.io/platforms/dart/guides//data-management/data-collected/ for more info
```
bruno-garcia, line 78: This link 404s

buenaflor: surprised that CI didn't catch this

PR #13411, opened 2025-04-16
refactor: migrate php onboarding to new syntax

a-hariti:
related to #13361


cleptric:
Hmm, the problem in PHP is that `traces_sample_rate` by itself does nothing. You need to manually create a transaction. Same with profiling, it requires a transaction.

a-hariti:
I just migrated what was already there

do you think we should improve docs before merging @cleptric?

cleptric:
Ahh, good point. Yes, though I have to put this work back to in two weeks, as I'm OOO. We can merge this as is then.

cleptric (approved):

PR #13413, opened 2025-04-16
refactor: migrate Golang and related guides to inline product options syntax

a-hariti:
related to #13361

### DIFF
--- a/docs/platforms/go/guides/echo/index.mdx
+++ b/docs/platforms/go/guides/echo/index.mdx
@@ -29,7 +29,7 @@ go get github.com/getsentry/sentry-go/echo
 <Break />
 
 
-```go {"onboardingOptions": {"performance": "14-17"}}
+```go
 import (
 	"fmt"
 	"net/http"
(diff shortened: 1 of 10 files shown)


giortzisg (commented):
Cool stuff. Checked the docs locally and for some reason tabs break after the `___PRODUCT_OPTION_END___`: 
Screenshot 2025-04-17 at 09 55 19

Also we missed the update on zerolog [here](https://github.com/getsentry/sentry-docs/blob/a0856dac8758b8a93b1f76861c150558cde0725b/docs/platforms/go/guides/zerolog/index.mdx?plain=1#L29).

a-hariti:
1. can you elaborate on the issue?

2. good catch @giortzisg!
but the Zerolog performance range (12-16) doesn't make sense currently, and doesn't have any effect on production as the `<OnboardingOptionButtons />` component is lacking the `performance` option

3. Fiber doesn't seem to have product options specified, should we add the buttons and performance option range as well?

a-hariti:
so to cleanup Zerolog I think we should just remove any options related syntax as we are only offering error monitoring?
(remove the range syntax and the `<OnboardingOptionButtons />`) @giortzisg

giortzisg:
Yeap, that makes sense. Went through other pages that only support error monitoring and they don't have the button, so I guess it also helps with consistency. So let's remove from zerolog and slog. @a-hariti

giortzisg:
Also logrus 😅

coolguyzone (approved):
Looks great! Everything works as expected on vercel preview 🏄‍♂️

PR #13414, opened 2025-04-16
refactor: migrate Dart and Flutter sdks to inline product options syntax

a-hariti:
related to #13361


kahest (approved):
lgtm, thank you!

PR #13420, opened 2025-04-16
fix(platform): Avoid IntersectionObserver issues

mydea:
We have a lot of issues like this: https://sentry.sentry.io/issues/5543047544/?project=1267915

> SyntaxError: Failed to construct 'IntersectionObserver': rootMargin must be specified in pixels or percent.

TBH not sure how that happens, but I guess we can try to be defensive here. Maybe inner height is missing somehow?

### DIFF
--- a/src/components/sidebarTableOfContents/index.tsx
+++ b/src/components/sidebarTableOfContents/index.tsx
@@ -157,16 +157,17 @@ export function SidebarTableOfContents() {
 
   // Mark the active item based on the scroll position
   useEffect(() => {
-    if (!tocItems.length) {
+    const innerHeight = window.innerHeight;
+    if (!tocItems.length || !innerHeight) {
       return () => {};
     }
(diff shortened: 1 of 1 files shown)


src/components/sidebarTableOfContents/index.tsx
```diff
@@ -183,7 +184,7 @@ export function SidebarTableOfContents() {
     }, observerOptions);
     const headings = tocItems.map(item => item.element);
     headings.forEach(heading => observer.observe(heading));
-    return () => headings.forEach(heading => observer.unobserve(heading));
+    return () => observer.disconnect();
```
mydea, line 187: I think this is better, we can tear down the whole observer like this I believe? I can't even find `unobserve` here: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/disconnect 🤔

PR #13412, opened 2025-04-16
refactor: migrate python and related guides to the inline product option syntax

a-hariti:
related to #13361

PR #13449, opened 2025-04-17
refactor: migrate python platform-includes to inline options syntax

a-hariti:
missed during #13412


stephanie-anderson (approved):
looks good!

PR #13454, opened 2025-04-19
Feat: Add "Copy Page" button for docs markdown pages

codyde:
## DESCRIBE YOUR PR

Working on a feature that allows copying of the page content as markdown to use in LLMs. Similar to whats on [Cloudflares docs](https://developers.cloudflare.com/agents/guides/remote-mcp-server/)

## IS YOUR CHANGE URGENT?  


chargome (commented):
Nice, I think it is a great feature to have!

Some thoughts on the implementation:
- As discussed offline this will bump our build time quite a bit (adds about 3 minutes as of now and will grow with the amount of pages we have) 
- I'd think of moving the button somewhere else (and adapt it to light mode), feels a bit too busy wdyt?

@a-hariti we can look into optimizing this a bit and think of alternatives

PR #13456, opened 2025-04-20
Overhaul and expand API guidelines

dcramer:
The exposed interfaces in the API have unfortunately become far too varied over the years, and part of this can be blamed on our lack of intentional design behind them. It was always implicit: design it like it was before. Unfortunately implicit design doesn't scale as more people work on a system, so this change documents a set of design principles that we can rely on to future proof new APIs, as well as apply to legacy APIs down the road.

Some notable shifts from where we are today:

- All API endpoints must be designed as if its a public API.
- `PATCH` should not be used. It is redundant with `PUT` in our use cases. If you want to "replace a resource" (which is uncommon behavior), you would do it via `POST /resource/:id`.
- Responses should use noun-based nesting on the object vs being flat (e.g. `{project: {...}` vs `{...}` for a project response).
- Identifiers should accept both human-readable identifiers and numeric identifiers. This has regressed quite drastically over the years, both in terms of query params as well as route identifiers.
- Identifiers should use common names (`organization`) rather than ambiguous attributes (`organization_id`).

Refs #13461


dcramer:
I will note some of the API design was intentional, but not written down when I built it. Its understandable how its super fragmented today, but obviously as its a public API thats not a really good place to be.

There's a _lot_ of internal details exposed in the API, some that might lead to exploitable, and on top of that a number of the endpoints are actually quite difficult to use because of this.

One last point: even the existing guidelines arent totally followed, but this would be a first step in broader education around how we design our APIs. For example, if you look at some of our APIs, you will trivially see how they are not designed for public consumption as they expose cryptic terminology, hidden details, or generally speaking what I would describe as a _lack_ of design.

dcramer:
https://develop-docs-git-feat-api-docs.sentry.dev/backend/api/design/ quick link to review in a sane way

dcramer:
One last note:

I want to be clear about the recommendation for parameter names. We can either do `organization` or `organization_id`, but `organization_id_or_slug` isnt adding much value and just increases complexity. I prefer the former as it implies the type, vs creating confusion in the spec.

sentaur-athena:
Another good idea to standardize the way filters work within List APIs. Have a tech spec here for it but couldn't staff it: https://www.notion.so/sentry/API-List-View-Filter-e07a8eeedb0b4e4883a60e30629b2cc6

develop-docs/backend/api/design.mdx
```diff
@@ -1,57 +1,140 @@
...
 
-### Don't Exceed 3-level Nesting
-Nested resources format like `api/0/organizations/\{org}/projects/` is recommended over `api/0/projects/` for readability because it gives user an understanding of resource hierarchy. However, nesting can make URLs too long and hard to use. Sentry uses 3-level nesting as a hybrid solution.
+**Every endpoint in Sentry's API is public.**
 
-Here are some possible urls for values with this resource hierarchy: organization -> project -> tag -> value
-- 👍 `/projects/\{organization_slug}/\{project_slug}/tags/\{tag_id}/values`
-- 👎 `/organizations/\{organization_slug}/projects/\{project_slug}/tags/\{tag_id}/values/`
-- 👎 `/values/`
+You should design it as such. You should **not** expose internal designs, including parameters which control behaviors of the underlying system.
```
sentaur-athena, line 12: An example of not to do here would be useful.

develop-docs/backend/api/design.mdx
```diff
@@ -1,57 +1,140 @@
...
+
+Routes are one of the most important parts of your API structure, as they communicate relationships and intent. Its important to get them right and adhere to a set of guidelines to make it easier to understand and use the API.
+
+This is no different - and just as important - to how you think about the URL structure of your user interface.
+
+Use the following guidelines for naming resources and their collections:
+
+- **Do** use lowercase and hyphenated collection names, e.g. `commit-files`.
+- **Do** use plural collection names. Avoid using uncountable words because the user can't know whether the GET returns one item or a list.
+- **Do** use `snake_case` for route parameters. e.g. `organizations/\{organization}/projects/`.
```
sentaur-athena, line 32: Did you mean to include an example with snake_case?

develop-docs/backend/api/design.mdx
```diff
@@ -1,57 +1,140 @@
...
+- **Do** consistently shorten parameters that are excessively long when the term will unambiguous. e.g. `organization` -> `org`.
```
sentaur-athena, line 33: I think we should define this here in the docs rather than keeping it to the designer. Organization or org?

dcramer:
> `id_or_slug` as path parameter is something we did based on customer asks. They want API paths that are not mutable as project names changes. This is especially important in case of partnerships where partner doesn't even have access to the org to know about these changes but needs to call the org APIs.
> 
> I don't mind renaming them to `organization` but still be mindful that the type is `id_or_slug`

To be clear I'm saying that is a requirement going forward - im just commenting on the key name.

Originally these were all expected to be able to handle IDs or Slugs, but a lot of APIs now only accept internal IDs (which is a problem in a lot of things).

develop-docs/backend/api/design.mdx
dcramer: I think its worth defining ones that are common if we agree on this. I'd need feedback on if there are any besides org that we'd want to shorten and I can add them.

develop-docs/backend/api/design.mdx
```diff
@@ -66,25 +163,210 @@ Each API should be **stateless**, have a clear purpose, and do one specific thin
...
 Each response object returned from an API should be a serialized version of the Django model associated with the resource. You can see all the existing serializers [here](https://github.com/getsentry/sentry/tree/master/src/sentry/api/serializers/models).
 
- <Alert title="Note">
+Some guidelines around the shape of responses:
+
+- **Do** use `camelCase` for all response attributes. e.g. `\{numCount: "123"}`.
+- **Do** return a responses as a named resource (e.g. `\{"user": \{"id": "123"}}`).
+- **Do** indicate collections using plural nouns (e.g. `\{"users": []}`).
+- **Do not** return custom objects. **Do** use a `Serializer` to serialize the resource.
+- **Do** return the smallest amount of data necessary to represent the resource.
```
sentaur-athena, line 198: This is a tricky one. I it especially because we had security incidents caused by this. I'm worried that making this will lead into multiple unnecessary serializers for one object and make the code not readable. 

Don't have a good solution either. How about saying for new models/resources, the new serializer includes the least amount of data needed, but for existing models look back at existing serializer there's a high chance you need to add the field you need there.

develop-docs/backend/api/design.mdx
```diff
@@ -66,25 +163,210 @@ Each API should be **stateless**, have a clear purpose, and do one specific thin
...
+    "id": 5,
+    "name": "foo",
+    ...
+  }
+}
+```
+
+This is how the original API was designed, but due to the limitations we intend to move to nested objects in the future.
+
+### Expanding responses
```
sentaur-athena, line 234: I didn't review below this line because it seemed unchanged. Let me know if I missed something.

develop-docs/backend/api/design.mdx
```diff
@@ -1,57 +1,154 @@
 ---
-title: Designing a New API
+title: Design Principles
 sidebar_order: 2
 ---
 
-[Django REST framework](https://www.django-rest-framework.org/)(DRF) is a powerful and flexible toolkit
-for building Web APIs. Sentry APIs are built using DRF. Here are some considerations to make when designing APIs at Sentry.
+This document contains a set of design principles and requirements whish should be applied to all Sentry APIs. These are _requirements_ when designing public APIs, but internal APIs should attempt to adhere to these principles as well.
```
markstory, line 6: Should this mention 'internal APIs' if all APIs are intended to be public (as per the section starting on line 16).

develop-docs/backend/api/design.mdx
```diff
@@ -1,57 +1,154 @@
...
+
+Use the following guidelines for naming resources and their collections:
+
+- **Do** use lowercase and hyphenated collection names, e.g. `commit-files`.
+- **Do** use plural collection names. Avoid using uncountable words because the user can't know whether the GET returns one item or a list.
+- **Do** use `snake_case` for route parameters. e.g. `tags/\{tag_name}/`.
+- **Do** consistently shorten parameters that are excessively long when the term will unambiguous. e.g. `organization` -> `org`.
 
-In the above example we flattened `projects`. The table below shows the existing flattened collections which works out with our existing APIs.
+Standard parameters that should be shortened in route names:
```
markstory, line 37: Is it worth clarifying what route name means here? Is it the path segments, parameters in the route, or the internal name that is used for route resolution?

develop-docs/backend/api/design.mdx
```diff
@@ -1,57 +1,154 @@
...
+
+```
+POST /resources/{id}
+```
+
+### Batch Operations
+
+Resources can get complicated when you need to expose batch operations vs single resource operations.
+
+Here are some notes that can help you decide between similar methods. We use *Get* here as an example but the same applies to all the other methods in the parenthesis.
```
vgrozdanic, line 122:
```suggestion
Here are some notes that can help you decide between similar methods. We use *POST* here as an example but the same applies to all the other methods in the parenthesis.
```

vgrozdanic: > the same applies to all the other methods in the parenthesis.

Maybe it is just me, and being non-native english speaker, but what does this mean? 
I don't understand what "other methods in the parenthesis" is referring to

develop-docs/backend/api/design.mdx
```diff
@@ -1,57 +1,154 @@
...
+## Resource Isolation
 
-## Functionality
 Each API should be **stateless**, have a clear purpose, and do one specific thing. To achieve that, stick with the standard methods listed below. If your API needs to be more complicated, work with [owners-api](https://github.com/orgs/getsentry/teams/owners-api) on how to create it.
 
-- 👍 An API that updates project settings: PATCH for updating a field or PUT for updating settings
-- 👎 An API that creates a project, creates a team, and creates alerts for that team about that project
+- **Do** expose endpoints to operate on individual objects (such as a project)
+- **Do not** combine multiple operations in a single endpoint _unless it is necessary_ (e.g. requires transaction guarantees).
+- **Do** use explicit custom endpoints for complex operations, rather than overloading a single resource's endpoint.
```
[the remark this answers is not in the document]
dcramer: Ah good catch, yeah I'll update it. Was trying to not delete all previous docs, but i didn't do a good enough job reviewing the whole thing

develop-docs/backend/api/design.mdx
dcramer: yeah unchanged below this other than a couple words

develop-docs/backend/api/design.mdx
dcramer: fwiw we had something like this present before. I personally find the advice somewhat intangible. I understand the intent, but i think the bigger concern is "dont return infinite relationships when those arent going to be needed for most queries". Either way it always ends up being subjective

develop-docs/backend/api/design.mdx
```diff
@@ -1,62 +1,159 @@
...
-- 👎 `/projects/`
+### Prefixes and Scoping
+
+Information in Sentry is typically constrained by tenants. That is, almost all information is scoped to an organization. All endpoints which query customer data **must** be scoped to an organization:
+
+- **Do** prefix resource organizations collections with `organizations/\{org}`.
+- **Do** prefix resource project collections with `projects/\{org}/\{project}`.
+- **Do not** expose endpoints which require `org` as a query parameter (it should always be a path parameter).
+
+Knowing when to choose which constraint to couple an endpoint to will be based on the purpose of an endpoint. For example, if an endpoint is only ever going to be used \to query data for a single project, it should be prefixed with `/projects/{org}/{project}/things`. If an endpoint would need to exist to query multiple projects (which is common with cross-project queries), you likely should expose it as `/organizations/{org}/things`, and expose a query param to filter on the project(s).
```
markstory, line 49:
```suggestion
Knowing when to choose which constraint to couple an endpoint to will be based on the purpose of an endpoint. For example, if an endpoint is only ever going to be used to query data for a single project, it should be prefixed with `/projects/{org}/{project}/things`. If an endpoint would need to exist to query multiple projects (which is common with cross-project queries), you likely should expose it as `/organizations/{org}/things`, and expose a query param to filter on the project(s).
```

An extra slash, but otherwise this is great.

develop-docs/backend/api/design.mdx
```diff
@@ -66,25 +163,210 @@ Each API should be **stateless**, have a clear purpose, and do one specific thin
...
+  {
+    "id": 1,
+		"name": "Team 1",
+		"slug": "team1",
+  },
+	{
+    "id": 2,
+		"name": "Team 2",
+		"slug": "team2",
+  }
```
markstory, line 346:
```suggestion
    "id": 1,
    "name": "Team 1",
    "slug": "team1",
  },
  {
    "id": 2,
    "name": "Team 2",
    "slug": "team2",
  }
```

whitespace was off.

markstory (approved):

dcramer:
Going to ship this as-is. We can improve/make changes as things move forward, but I want to focus on incremental

PR #13463, opened 2025-04-22
fix(style): Add padding for dev docs sidebar

chargome:
simply adds some padding to the dev docs sidebar that got lost in https://github.com/getsentry/sentry-docs/pull/13354

### DIFF
--- a/src/components/sidebar/developDocsSidebar.tsx
+++ b/src/components/sidebar/developDocsSidebar.tsx
@@ -37,10 +37,10 @@ export function DevelopDocsSidebar({
     return toTree(apiNodes);
   };
   return (
-    <aside className={`${styles.sidebar} py-3`}>
+    <aside className={`${styles.sidebar} p-3`}>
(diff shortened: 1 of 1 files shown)


src/components/sidebar/developDocsSidebar.tsx
```diff
@@ -37,10 +37,10 @@ export function DevelopDocsSidebar({
     return toTree(apiNodes);
   };
   return (
-    <aside className={styles.sidebar}>
+    <aside className={`${styles.sidebar} py-3`}>
```
mydea, line 40: can this not just be

```suggestion
    <aside className={`${styles.sidebar} p-3`}>
```

or does this need to be spread out like this?

chargome: I simplified it by just moving the padding altogether to the `aside` but a a horizontal padding of `6` would be too much

mydea (approved):

PR #13467, opened 2025-04-22
docs(react-router): Add docs for distributed tracing

chargome:
documents https://github.com/getsentry/sentry-javascript/pull/16070


docs/platforms/javascript/guides/react-router/index.mdx
```diff
@@ -199,40 +199,110 @@ Sentry.init({
...
-  }
-}
-
-+ export default Sentry.sentryHandleRequest(handleRequest);
 
 // ... rest of your server entry
 ```
 
+<Expandable title="Do you need to customize your handleRequest function?">
+  If you need to update the logic of your `handleRequest` function you'll need to include the provided Sentry helper functions (`getMetaTagTransformer` and `wrapSentryHandleRequest`) manually:
```
Lms24, line 233: Do I understand this correctly that this more manual approach is necessary when users have their own `handleRequest` function already? If so, I think we should point this out more clearly because I didn't fully understand the "If you need to update the logic of your `handleRequest` function" part. (Maybe that's just me though, so feel free to disregard)

```suggestion
<Expandable title="Do you have your own handleRequest function?">
  If already defined your own `handleRequest` function you'll need to include the provided Sentry helper functions (`getMetaTagTransformer` and `wrapSentryHandleRequest`) manually:
```

(feel free to go with different wording as you see fit

Lms24 (approved):

Lms24:
Probably unrelated to this specific PR but I noticed that the `*` character in the diff code block (`entry.client.tsx`) has a bit too much right padding: 

image

docs/platforms/javascript/guides/react-router/index.mdx
chargome: yeah that is correct, BUT whenever a user creates this hook, the handleRequest function is already defined – so I only want to encourage this path when they need som special logic in their function .. not sure how to name this though

Lms24: Maybe I don't fully understand this yet but I guess what you mean is that when you run `react-router reveal` there already is a predefined `handleRequest` and this is the one users _don't_ need to add Sentry stuff to? 
So it's just for the case that users have a custom behaviour in the function already prior to adding Sentry? 

I don't think this is a blocker btw, so if you wanna merge this today, feel free. Maybe we have a minute tomorrow to chat about it this quickly in person

PR #13469, opened 2025-04-22
docs(contributing): Document new syntax for onboarding snippets

chargome:
documents https://github.com/getsentry/sentry-docs/pull/13361

### DIFF
--- a/docs/contributing/pages/components.mdx
+++ b/docs/contributing/pages/components.mdx
@@ -378,20 +378,19 @@ Attributes:
 
 ## Onboarding Options
 
-If you're writing product feature specific docs, you can specify code block `onboardingOptions` metadata:
+If you're writing product feature specific docs, you can specify markers within code blocks that enable or disable certain parts of snippets:
 
 ````markdown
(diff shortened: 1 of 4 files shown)


docs/contributing/pages/components.mdx
```diff
@@ -378,21 +378,20 @@ Attributes:
...
+```go
+  // ___product_option_start___ performance
+  // your code here
+  // ___product_option_end___ performance
 ```
 ````
 
-the general syntax is `{"onboardingOptions": {"feature": "range"}}` where `feature` is the feature id
-and `range` is the corresponding line range (similar to the line highlighting syntax).
+the general syntax is `// ___product_option_start___ feature` where `feature` is the feature id (e.g. `performance`, `profiling` or `session-replay`).
```
a-hariti, line 391: this should be uppercase, also we should highlight that the leading characters of the comment are language dependent, ie `#` for python, `//` for C like languages ...

a-hariti (changes requested):

chargome:
@a-hariti feel free to merge whenever 👍

PR #13478, opened 2025-04-22
ref(flutter): clean up init code snippets

buenaflor:
It's better to move the code snippet for custom zones to the troubleshooting page as it is not so common to run into this issue.


coolguyzone (approved):
Makes sense, thanks for the update!

PR #13497, opened 2025-04-24
docs(js): Review and update Nest.js Quick Start guide

inventarSarah:
## DESCRIBE YOUR PR
The Nest.js quick start guide already looked great -> I added and removed some content to be more consistent with our other guides.

I tried to streamline the CommonJS section, as we already have the details on [this page](https://docs.sentry.io/platforms/javascript/guides/nestjs/install/commonjs/) (I added a link to the page). Plus, the content seemed too detailed for a Quick Start guide IMO. 
Naturally, if you say the info I removed is vital for the QS guide, we can add it back in.

Additionally, I expanded the "Verify Your Setup" section so the example provided is like our [Node.js one](https://docs.sentry.io/platforms/javascript/guides/node/#step-4-verify-your-setup).

Closes: #13498

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

## EXTRA RESOURCES


platform-includes/getting-started-use/javascript.nestjs.mdx
```diff
@@ -33,28 +33,15 @@ import { AppService } from "./app.service";
 export class AppModule {}
 ```
 
-If you're using a global catch-all exception filter (which is either a filter registered with `app.useGlobalFilters()` or a filter registered in your app module providers annotated with a `@Catch()` decorator without arguments), add a `@SentryExceptionCaptured()` decorator to the filter's `catch()` method.
-This decorator will report all unexpected errors that are received by your global error filter to Sentry:
+By default, only unhandled exceptions that are not caught by an error filter are reported to Sentry. Additionally, `HttpException`s (including [derivatives](https://docs.nestjs.com/exception-filters#built-in-http-exceptions)) are not captured automatically as they mostly act as control flow vehicles.
```
inventarSarah, line 36: Here we start with the Error Capturing topic.
I was wondering if it would make sense to not have this in "Apply Instrumentation to Your App" and instead put it in a dedicated section, "Step x: Capture Nest.js Errors" -> similar to [how we have it for React ](https://docs.sentry.io/platforms/javascript/guides/react/#apply-instrumentation-to-your-app).

But then this only seems to be documented for CJS - I was wondering why we don't have this documented for ESM? Or is this not needed/possible? 

I don't see this as part of this PR -- we can address in a future issue/PR.

Lms24 (changes requested):
This looks good but I think we need to sort out some things:

1. h: we can't include nestjs specific code and instructions in the shared commonjs page
2. m: we should avoid linking to commonJS for the exception filter explanation (or IOW think about where we put this and then link there if applicable)

1 we need to solve in this PR, 2 we can also address in a follow-up.

docs/platforms/javascript/common/install/commonjs.mdx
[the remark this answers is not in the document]
chargome: Yeah sounds reasonable to just use a ts file here 👍

platform-includes/getting-started-verify/javascript.nestjs.mdx
```diff
@@ -1,6 +1,44 @@
+### Issues
+
+First, let's verify that Sentry captures errors and creates issues in your Sentry project. Add the following route to your application, which will call an undefined function, triggering an error that Sentry will capture:
+
 ```javascript
 @Get("/debug-sentry")
-getError() {
-  throw new Error("My first Sentry error!");
+  getError() {
+    setTimeout(() => {
```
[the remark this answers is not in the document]
inventarSarah: you're right - this isn't needed here 👍

inventarSarah:
Thanks @Lms24 ! 

> 1. h: we can't include nestjs specific code and instructions in the shared commonjs page
> 2. m: we should avoid linking to commonJS for the exception filter explanation (or IOW think about where we put this and then link there if applicable)

I restored the previous content of the commonjs and getting-started-use/nestjs pages. So, now the Nest.js specific info shouldn't surface on other platforms anymore -- lmk if I missed something 🙏 (I think this is the most complex structure I've worked with so far :D)

This way, we can keep this PR focused and small and tackle other issues in a new PR. -> I already created an issue here https://github.com/getsentry/sentry-docs/issues/13525
Please review and comment if you have any suggestions.

platform-includes/getting-started-use/javascript.nestjs.mdx
```diff
@@ -95,13 +99,11 @@ export class ExceptionFilter implements RpcExceptionFilter<RpcException> {
...
 ```
 
-
 </Expandable>
 
-
 If you have error filters for specific types of exceptions (for example `@Catch(HttpException)`, or any other `@Catch(...)` with arguments) and you want to capture errors caught by these filters, capture the errors in the `catch()` handler with `Sentry.captureException()`:
 
-```javascript {9}
+```javascript {4-9}
```
Lms24, line 106:
```suggestion
```javascript {4, 9}
```

Lms24 (approved):
Thanks for making the changes and for creating the follow-up issue.

platform-includes/getting-started-verify/javascript.nestjs.mdx
```diff
@@ -1,6 +1,42 @@
+### Issues
+
+First, let's verify that Sentry captures errors and creates issues in your Sentry project. Add the following route to your application, which will call an undefined function, triggering an error that Sentry will capture:
```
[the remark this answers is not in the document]
inventarSarah: Updated the examples -> hope this looks good now!

inventarSarah:
New updates in this PR, @chargome @Lms24 
- guide no longer inherits content from getting-started-node
- updated code snippets to Typescript (remove ESM tabs)
- created new step "Capturing Nest.js Errors" and moved content about exception filters there
- created a new include file for the exception filter content so I could use it in the quick start guide and the CommonJS page -> once the CommonJS page is updated, we can use the content from the include file directly in the QS guide and delete the include file
- removed "ESM" subsection in "Configure"

platform-includes/getting-started-capture-errors/javascript.nestjs.mdx
```diff
@@ -0,0 +1,84 @@
+By default, Sentry only captures unhandled exceptions that aren't caught by an error filter.
+Additionally, `HttpException`s (including [derivatives](https://docs.nestjs.com/exception-filters#built-in-http-exceptions)) aren't captured by default because they mostly act as control flow vehicles.
+
+To make sure Sentry captures all your app's errors, configure error handling based on how your application manages exceptions:
+
+### Using a Global Catch-All Exception Filter
```
chargome, line 6: This I would rather name something like "using a custom global filter" – and update the section below accordingly

platform-includes/getting-started-use/javascript.nestjs.mdx
```diff
@@ -14,9 +16,9 @@ async function bootstrap() {
 bootstrap();
 ```
 
-Afterwards, add the `SentryModule` as a root module to your main module:
+Afterward, add the `SentryModule` as a root module to your main module:
```
chargome, line 19:
```suggestion
Afterwards, add the `SentryModule` as a root module to your main module:
```
I think? 😅

chargome (approved):

platform-includes/getting-started-capture-errors/javascript.nestjs.mdx
inventarSarah: Updated the subheading to

Using a Custom Global Filter
and
Not Using a Custom Global Filter

chargome (changes requested):
@inventarSarah just realized that we have the global error filter sections now twice in the guide!

inventarSarah:
> @inventarSarah just realized that we have the global error filter sections now twice in the guide!

Fixed -> things got crazy with the includes!

Quick Start guide and CommonJS installation method page (which used the same includes) look okay to me now

inventarSarah:
> seems to be stuck, can you re-trigger a deployment?
@chargome now everything has worked out

chargome (approved):

PR #13511, opened 2025-04-25
Make the API Design doc a bit more consistent

Swatinem:
In particular, this removes a bunch of stray backslashes, like `\{`.
It also uses more consistent formatting of paths, related to slashes and placeholders.


sfanahata (approved):
Preview looks good.

PR #13538, opened 2025-04-28
chore(develop): Update frontend router testing guidance

malwilley:
Ref https://github.com/getsentry/frontend-tsc/issues/85

The internal development docs were referencing the old way of mocking router state in frontend tests. Updates with the new method of doing so.


scttcper (approved):
might be also helpful to mention how to avoid the usePageFilters mocks

PR #13561, opened 2025-05-01
Clarify db index docs

wedamija:


develop-docs/backend/application-domains/database-migrations/index.mdx
```diff
@@ -135,11 +135,12 @@ generally prefer that, since it averages the load out over a longer period of ti
 
 ### Indexes
 
-We prefer to create indexes on large existing tables with `CREATE INDEX CONCURRENTLY`. Our migration framework will do this automatically when creating
-a new index. Note that when `CONCURRENTLY` is used we can't run the migration in a transaction, so it's important to use `atomic = False` to run these.
+These are automatically handled by the migration framework, you can just create them on the Django model and generate the migration.
 
 When adding indexes to large tables you should use a `is_post_deployment` migration as creating the index could take longer than the migration statement timeout of 5s.
 
+Note: These are created using `CONCURRENTLY`, so it's important to not set `atomic = True` for migrations that contain indexes. This is disabled by default.
```
saponifi3d, line 142: I think it'd be a little clearer if the `atomic = True` was part of the first paragraph, like "these are things we should be doing when we want to generate a new index" 

Then in the note, it can explain that it's because we use concurrently.

saponifi3d (approved):
thanks for updating these!

develop-docs/backend/application-domains/database-migrations/index.mdx
wedamija: Generally no action is required about `atomic`. It's false by default, so it's just a side note in case someone tries to enable it for some reason.

PR #13594, opened 2025-05-05
feat(go): add sql queries documentation

giortzisg:
## DESCRIBE YOUR PR
adding docs for using sentrysql for db queries.


docs/platforms/go/common/tracing/instrumentation/sql-module.mdx
```diff
@@ -0,0 +1,64 @@
...
+## Automatic instrumentation
+
+The `sentrysql` package gives you the ability to either wrap a driver (recommended for `sql.Open`) or wrap a connector (recommended for `sql.OpenDB`).
+
+### Wrapping a driver
+
+```go
+dsn := "postgres://postgres:password@write.postgres.internal:5432/postgres"
+sql.Register("sentrysql-postgres", sentrysql.NewSentrySQL(
+  &pq.Driver{},
```
aldy505, line 21: We should mention that users should read the database driver's docs, since every database driver is different. Would be nice if we can put "This is an example for github.com/lib/pq package"

docs/platforms/go/common/tracing/instrumentation/sql-module.mdx
```diff
@@ -0,0 +1,64 @@
...
+```
+This allows you to connect like normal with `sql.Open`.
+
+### Wrapping a connector
+
+For more control (for example, for a read replica), you can wrap a connector:
+
+```go
+dsn := "postgres://postgres:password@read.postgres.internal:5432/postgres"
+connector, err := pq.NewConnector(dsn)
```
aldy505, line 46: The same with mentioning users should read their database driver's docs.

docs/platforms/go/common/tracing/instrumentation/sql-module.mdx
giortzisg: I would argue that a code comment would sometimes get overlooked by the users. However, I agree that it would be nice to add one more line in the actual documentation prompting the user to read the docs.

aldy505 (approved):

PR #13598, opened 2025-05-05
fix(platform): Add scrollable horizontal overflow

chargome:
Fixes UI bugs with overflowing tables etc.

### DIFF
--- a/src/components/docPage/index.tsx
+++ b/src/components/docPage/index.tsx
@@ -69,7 +69,7 @@ export function DocPage({
         <main className="main-content flex w-full mt-[var(--header-height)] flex-1 mx-auto">
           <div
             className={[
-              'mx-auto lg:mx-0 pt-6 px-6 prose dark:prose-invert max-w-full text-[var(--gray-12)] prose-a:no-underline hover:prose-a:underline',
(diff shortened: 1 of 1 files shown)


szokeasaurusrex (commented):
Hey @chargome, thanks for this PR, and sorry I missed the review notification.

Although this is probably better than what we had before, I don't think it's ideal.

Here's a screen recording to demonstrate (made on [this page](https://develop-docs-git-cg-overflow-x.sentry.dev/sdk/telemetry/traces/trace-propagation-cheat-sheet/) from the preview):

https://github.com/user-attachments/assets/bfa42f3b-b2ae-4dda-8610-c42e36ad5f19

Specifically, what I think should be improved:
1. It's a bit difficult to tell that this page is horizontally scrollable. Maybe, this could be fixed by adding a shadow border on the right side when the page is scrollable horizontally, which makes it look like the right sidebar is "above" the rest of the horizontally scrollable portion
2. Also, rather than having the whole docs page scroll horizontally, I think it would be better to have only the overflowing portion (in this case, the table) scroll horizontally.

I'll approve the PR if you think it's too much effort to address these suggestions here, but I think it would be best to create an issue for them, in that case.

sfanahata:
@chargome - are you still working on this one?

chargome:
This became obsolete with the layout changes, sorry that this one fell under the radar!

PR #13635, opened 2025-05-08
fix(js): Clarify scope setting docs

mydea:
Closes https://linear.app/getsentry/issue/FE-405/user-feedback-elaborate-on-current-scope-usage

### DIFF
--- a/docs/platforms/javascript/common/enriching-events/scopes/index.mdx
+++ b/docs/platforms/javascript/common/enriching-events/scopes/index.mdx
@@ -106,33 +106,32 @@ Sentry.captureException(new Error("my error"));
 
 ## Configuring the Scope
 
-There are two main ways to interact with the scope. You can access the current scope via `Sentry.getCurrentScope()` and use setters on the resulting scope, or you can use global methods like `Sentry.setTag()` directly, which will set on the respective scope under the hood (which will be the isolation scope).
+In most cases, you should use the top-level methods like `Sentry.setTag()` directly, which will set the data on the isolation scope, ensuring it is applied to all events related to the current request/process.
 
-You'll first need to import the SDK, as usual:
+If you want to apply data to a more narrow scope, you can use `withScope` to create a new scope that is only active for a specific part of your code. `withScope` receives the scope as callback argument, which you can use to set data on the scope.
 
 <PlatformContent includePath="enriching-events/import" />
 
 You can, for instance, add custom tags or inform Sentry about the currently authenticated user.
 
 ```javascript
-/// Usually, you don't want to write on the current scope, so use with care!
-const scope = Sentry.getCurrentScope();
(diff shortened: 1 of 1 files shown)


andreiborza (approved):
Just some suggestions, feel free to disregard.

docs/platforms/javascript/common/enriching-events/scopes/index.mdx
```diff
@@ -106,34 +106,33 @@ Sentry.captureException(new Error("my error"));
 
 ## Configuring the Scope
 
-There are two main ways to interact with the scope. You can access the current scope via `Sentry.getCurrentScope()` and use setters on the resulting scope, or you can use global methods like `Sentry.setTag()` directly, which will set on the respective scope under the hood (which will be the isolation scope).
+In most cases, you should use the top-level methods like `Sentry.setTag()` directly, which will set the data on the isolation scope, ensuring it is applied to all events related to the current request/process.
```
coolguyzone, line 109:
```suggestion
In most cases, you should use top-level methods like `Sentry.setTag()` directly, which will set the data on the isolation scope, ensuring it is applied to all events related to the current request/process.
```

docs/platforms/javascript/common/enriching-events/scopes/index.mdx
```diff
@@ -106,34 +106,33 @@ Sentry.captureException(new Error("my error"));
...
 
-You'll first need to import the SDK, as usual:
+If you want to apply data to a more narrow scope, you can use `withScope` to create a new scope that is only active for a specific part of your code. `withScope` receives the scope as callback argument, which you can use to set data on the scope.
```
coolguyzone, line 111:
```suggestion
If you want to apply data to a more narrow scope, you can use `withScope` to create a new scope that is only active for a specific part of your code. `withScope` receives the scope as a callback argument, which you can use to set data on the scope.
```

docs/platforms/javascript/common/enriching-events/scopes/index.mdx
```diff
@@ -106,34 +106,33 @@ Sentry.captureException(new Error("my error"));
...
   email: "john.doe@example.com",
 });
+
+// Alternatively, you can use `withScope` as documented below.
 ```
 
 To learn what useful information can be associated with scopes see
 [context](../context/), [tags](../tags), [users](../identify-user) and [breadcrumbs](../breadcrumbs/).
 
+Note that we generally recommend to avoid using `getCurrentScope()` and setting data on it, as the current scope can have an unreliable lifetime. Instead:
```
coolguyzone, line 130:
```suggestion
We recommend you avoid using `getCurrentScope()` and setting data on it, as the current scope can have an unreliable lifetime. Instead:
```

coolguyzone (approved):
🛳️

PR #13636, opened 2025-05-08
feat(js): Better document tag key/value requirements

mydea:
Closes https://linear.app/getsentry/issue/FE-394/user-feedback-tags-with-spaces-cause-problems


andreiborza (approved):

AbhiPrasad (approved):

coolguyzone (approved):
Thanks for updating!

PR #13669, opened 2025-05-12
docs(react-router): Add note for Windows users to set env var

RulaKhaled:
Users on Windows are seeing this error: 

> 'NODE_OPTIONS' is not recognized as an internal or external command,
> operable program or batch file.
> 

This PR should help users set up the env variable before running the command on Windows.

Screenshot 2025-05-12 at 12 34 52 PM

PS. I will be updating other occurrences in our docs for Windows users as well. 

## LEGAL BOILERPLATE

## EXTRA RESOURCES


src/components/WindowsNodeEnvSetup.tsx
chargome, line 1: This kind of reusable content can go into `/includes` as an mdx component, we don't really need a new component for this – you can then use this as `<Include name="your/path" />` 👍

src/components/WindowsNodeEnvSetup.tsx
```diff
@@ -0,0 +1,33 @@
+import {Alert} from './alert';
+import {CodeBlock} from './codeBlock';
+import {CodeTabs} from './codeTabs';
+
+export function WindowsNodeEnvSetup({nodeOption}: {nodeOption: string}) {
+  return (
+    <Alert>
```
chargome, line 7: WDYT about making this an expandable element?

chargome (changes requested):

src/components/WindowsNodeEnvSetup.tsx
RulaKhaled: Yah can do, is there a pattern for these kinds of stuff?

platform-includes/node-options-windows/javascript.react-router.mdx
```diff
@@ -0,0 +1,15 @@
+<Expandable title="Are you using Windows?">
+
+If you're on Windows, set `NODE_OPTIONS` environment variable manually before running your app.
```
coolguyzone, line 3:
```suggestion
If you're on Windows, set the `NODE_OPTIONS` environment variable manually before running your app.
```

coolguyzone (approved):
LGTM

chargome (approved):

PR #13685, opened 2025-05-13
fix(python): fix Python logs docs

lcian:
## DESCRIBE YOUR PR
Fixes some issues with the logs docs

### DIFF
--- a/platform-includes/logs/options/python.mdx
+++ b/platform-includes/logs/options/python.mdx
@@ -24,7 +24,7 @@ sentry_sdk.init(
 
 The `before_send_log` function receives a log object, and should return the log object if you want it to be sent to Sentry, or `None` if you want to discard it.
 
-The log dict has the following values:
+The log dict has the following keys:
 
 - `severity_text`: (`str` - one of `trace`, `debug`, `info`, `warning`, `error`, `fatal`) The log level.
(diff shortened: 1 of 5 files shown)


platform-includes/logs/options/python.mdx
```diff
@@ -1,29 +1,34 @@
...
 ```python
-def _before_log(log):
-    if log.level == "info":
-        # Filter out all info logs
-        return None
+import sentry_sdk
+from sentry_sdk.types import Log, Hint
+from typing import Optional
 
+def before_log(log: Log, _hint: Hint) -> Optional[Log]:
```
lcian, line 10: Should be correct right?

platform-includes/logs/options/python.mdx
```diff
@@ -1,29 +1,34 @@
...
-        "before_send_log": _before_log,
+        "before_send_log": before_log,
     },
 )
 ```
 
 The `before_send_log` function receives a log object, and should return the log object if you want it to be sent to Sentry, or `None` if you want to discard it.
 
-The log object has the following properties:
+The log dict has the following values:
```
lcian, line 27:
```suggestion
The log dict has the following keys:
```
maybe better?

AbhiPrasad (approved):
Thanks @lcian! Gonna go ahead and merge this PR.

PR #13705, opened 2025-05-14
fix(codeowners): Skip requested reviewers

Jeffreyhung:
Adding a check so that if a codeowner is already in the requested reviewer list then skip it, otherwise every new commit to the PR will trigger the workflow to readd the codeowner.


geoffg-sentry (approved):
Probably should wrap this into a lil bash function for reusability later

PR #13715, opened 2025-05-15
feat(native): update transaction docs

JoshuaMoelans:
## DESCRIBE YOUR PR
Part of https://github.com/getsentry/sentry-native/pull/1236

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

### DIFF
--- a/docs/platforms/native/common/configuration/filtering.mdx
+++ b/docs/platforms/native/common/configuration/filtering.mdx
@@ -31,3 +31,32 @@ Typically, a `hint` holds the original exception so that additional data can be 
 <PlatformContent includePath="configuration/before-send-hint" />
 
 When the SDK creates an event or breadcrumb for transmission, that transmission is typically created from some sort of source object. For instance, an error event is typically created from a log record or exception instance. For better customization, SDKs send these objects to certain callbacks (<PlatformIdentifier name="before-send" />, <PlatformIdentifier name="before-breadcrumb" /> or the event processor system in the SDK).
+
+
+## Filtering Transaction Events
+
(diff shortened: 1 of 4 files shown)


docs/platforms/native/common/configuration/filtering.mdx
```diff
@@ -31,3 +31,32 @@ Typically, a `hint` holds the original exception so that additional data can be
...
+### Using <PlatformIdentifier name="traces-sampler" />
+
+<Alert>
+The <PlatformIdentifier name="traces-sampler" /> and <PlatformIdentifier name="traces-sample-rate" /> config options are mutually exclusive. If you define a <PlatformIdentifier name="traces-sampler" /> to filter out certain transactions, you must also handle the case of non-filtered transactions by returning the rate at which you'd like them sampled.
+</Alert>
+
+
+In its simplest form, used just for filtering the transaction, it looks like this:
+
+<PlatformContent includePath="performance/traces-sampler-as-sampler" />
```
JoshuaMoelans, line 49: Other SDKs use a separate example (`traces-sampler-as-filter`) but for completeness sake this would need to look very similar to the sampler example, so I opted to just point to the same one we already use in our [Sampling](https://docs.sentry.io/platforms/native/configuration/sampling/#setting-a-sampling-function) docs.

(example on Android; [sampling](https://docs.sentry.io/platforms/android/configuration/sampling/#setting-a-sampling-function) vs [filtering](https://docs.sentry.io/platforms/android/configuration/filtering/#using-traces-sampler)

supervacuus: To be honest, I don't really see the difference in the Android example either. Both show that you change the rate based on some X. One can probably be dropped in, whereas the other drops the concrete conditionals for prose. 

Maybe I am blind, but sampling and filtering based on a dice roll seem like the same thing to me. What do the docs try to achieve by marking the distinction? Or is this only about the surrounding text?

In any case, I think linking this is fine.

supervacuus (approved):

PR #13717, opened 2025-05-15
docs: remove old DSC content

vgrozdanic:
Seems like it should have been TODO comment in a first place, 3 years ago.

Since in 3 years it hasn't been touched, seems to be good to deleted


AbhiPrasad (approved):
Some of them are still todos fwiw, but you're right that we can probably remove them.

cleptric (approved):

PR #13735, opened 2025-05-16
chore(dynamic-sampling): point out that dynamic sampling does not apply to errors

shellmayr:
- There has been some confusion around this - add a small callout to clarify that dynamic sampling only applies to transactions and spans
- Closes TET-462


vgrozdanic (approved):

priscilawebdev (approved):
image
lgtm!

PR #13755, opened 2025-05-19
feat(platform): Set `sentry.source` to `custom` for edge requests

chargome:
This is a temporary change to evaluate effects on grouping in Sentry.


dcramer:
Can we not resort to things like this? Customers have to maintain the mental burden of this kind of change, when all we need to do is fix the server side

dcramer:
Sorry ignore my comment - realized this is just internal to docs

cleptric (approved):

PR #13767, opened 2025-05-20
Remove docs related to the deprecated `sentry-cli files` commands

Swatinem:
Going forward, we only want to support uploading sourcemaps via the `sentry-cli sourcemaps upload` command.

This removes docs related to the deprecated `files upload` command, so it is less discoverable.


docs/cli/releases.mdx
```diff
@@ -120,22 +120,6 @@ sentry-cli releases set-commits "$VERSION" --auto --ignore-missing
...
-
-The first argument to `upload` is the path to the file, the second is an optional URL we should associate it with. Note that if you want to use an abbreviated URL (eg: `~/foo.js`) make sure to use single quotes to avoid the expansion by the shell to your home folder.
-
-```bash
-sentry-cli files upload --release "$VERSION" /path/to/file '~/file.js'
-```
-
 ### Upload Source Maps
 
 For source map upload, a separate command is provided which assists you in uploading and verifying source maps:
```
szokeasaurusrex, line 125:
```suggestion
You can upload sourcemaps via the `sentry-cli sourcemaps upload` command:
```

docs/cli/releases.mdx
```diff
@@ -120,22 +120,6 @@ sentry-cli releases set-commits "$VERSION" --auto --ignore-missing
 
 When you are working with JavaScript and other platforms, you can upload release artifacts to Sentry which are then considered during processing. The most common release artifact are [source maps](/platforms/javascript/sourcemaps/) for which `sentry-cli` has specific support.
```
szokeasaurusrex, line 121: This snippet should also be changed to reflect the deprecation, or perhaps, we should remove it since I don't think it makes a ton of sense.

```suggestion
```

szokeasaurusrex (approved):

PR #13768, opened 2025-05-20
Add logs docs for Dart/Flutter

buenaflor:
Adds docs for logs on Dart/Flutter


AbhiPrasad (commented):
let's make sure we also update https://github.com/getsentry/sentry-docs/blob/master/docs/product/explore/logs/getting-started/index.mdx

AbhiPrasad (approved):

PR #13770, opened 2025-05-20
remove version from provided url

bitsandfoxes:
Relates to https://github.com/getsentry/sentry-unity/issues/2162

Not adding the version by default allows users to make use of the `Update` button, since the package is no longer pinned to that specific version.

### DIFF
--- a/platform-includes/getting-started-install/unity.mdx
+++ b/platform-includes/getting-started-install/unity.mdx
@@ -1,8 +1,10 @@
 Install the package via the [Unity Package Manager using a Git URL](https://docs.unity3d.com/Manual/upm-ui-giturl.html) to Sentry's SDK repository:
 
 ```
-https://github.com/getsentry/unity.git#{{@inject packages.version('sentry.dotnet.unity', '0.0.5') }}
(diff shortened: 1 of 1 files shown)


platform-includes/getting-started-install/unity.mdx
```diff
@@ -1,9 +1,11 @@
 Install the package via the [Unity Package Manager using a Git URL](https://docs.unity3d.com/Manual/upm-ui-giturl.html) to Sentry's SDK repository:
 
 ```
-https://github.com/getsentry/unity.git#{{@inject packages.version('sentry.dotnet.unity', '0.0.5') }}
+https://github.com/getsentry/unity.git
 ```
 
+If you want to use a specific version of the SDK, you can append `#{{@inject packages.version('sentry.dotnet.unity', '0.0.5') }}` with the version you want to use to the URL.
```
coolguyzone, line 7:
```suggestion
To use a specific version of the SDK, append `#{{@inject packages.version('sentry.dotnet.unity', '0.0.5') }}` to the URL, replacing '0.0.5' with the desired version.
```

PR #13784, opened 2025-05-21
Fix Vercel AI instructions

BYK:
`Sentry.vercelAIIntegration` is not a constructor so drop the `new` keyword.

## IS YOUR CHANGE URGENT?  

None: Not urgent, can wait up to 1 week+


AbhiPrasad (approved):
Thanks!

PR #13787, opened 2025-05-21
fix: remove new method from koa docs

AbhiPrasad:


AbhiPrasad:
Going to open a big PR with all the changes - this happens in multiple integration docs.

PR #13809, opened 2025-05-22
Unreal Engine: Add note on supported platforms for Crash Reporter

tustanivsky:
This PR adds a note about the platforms where `Crash Reporter` is supported as an addition to the `CRC vs Sentry SDK` comparison.

### DIFF
--- a/docs/platforms/unreal/configuration/setup-crashreporter/index.mdx
+++ b/docs/platforms/unreal/configuration/setup-crashreporter/index.mdx
@@ -21,6 +21,12 @@ Below we'll break down each step in detail.
 The UE Crash Reporter is provided out of the box by Epic Games. It is able to collect relevant
 information about the crash and send it to a service like Sentry to process.
 
+<Alert>
+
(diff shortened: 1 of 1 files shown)


docs/platforms/unreal/configuration/setup-crashreporter/index.mdx
```diff
@@ -21,6 +21,12 @@ Below we'll break down each step in detail.
 The UE Crash Reporter is provided out of the box by Epic Games. It is able to collect relevant
 information about the crash and send it to a service like Sentry to process.
 
+<Alert>
+
+Crash Reporter is supported only on desktop (Windows, Mac, Linux). In order to capture crashes on mobile and consoles, you need to use the Sentry SDK.
```
bruno-garcia, line 26:
```suggestion
Epic Games only provides the Crash Reporter Client on desktop (Windows, Mac, Linux). In order to capture crashes on mobile and consoles, you need to use the Sentry SDK, which support all platforms, including desktop.
```

bruno-garcia (approved):

PR #13818, opened 2025-05-23
feat(platform): Capture plausible event for expandables

mydea:
This should help us to get some data on which expandables are actually read how often etc.


stephanie-anderson (approved):
Amazing, thank you!

PR #13832, opened 2025-05-26
feat(logs): Expand attributes section for logs spec and specify defaults for every platform

AbhiPrasad:
resolves https://linear.app/getsentry/issue/LOGS-28

After testing with some SDKs and getting feedback, we've decided on a list of default attributes the SDKs should try to attach to logs.

This includes
1. user attributes from https://develop.sentry.dev/sdk/data-model/event-payloads/user/ 
2. browser attributes, parsed from user agents (https://github.com/getsentry/relay/pull/4757)
3. os and device attributes attached to mobile, desktop, and native sdks, based on https://develop.sentry.dev/sdk/data-model/event-payloads/contexts/

I also re-organized this section to make things a bit more clear.

### DIFF
--- a/develop-docs/sdk/telemetry/logs.mdx
+++ b/develop-docs/sdk/telemetry/logs.mdx
@@ -371,6 +371,12 @@ For mobile, desktop, and native SDKs (Android, Apple, Electron, etc.), the SDKs 
 }
 ```
 
+#### Future Default Attributes
+
+The SDKs should aim to minimize the number of default attributes attached to a log. Logs are intended to be lightweight, and we want to try to keep the average byte size of a log as small as possible as users will be charged per byte size of logs sent.
+
+We are trying to settle on a balance of debuggability vs. smaller byte size for logs which is why new default attributes should only be added after significant feedback from users and discussion internally with the SDK and ingest teams. There is no hard rule about what exact attributes are allowed, every proposed new attribute will be evaluated on a case-by-case basis.
+
 ### Data Category and Rate Limiting
 
 A new data category for logs has been added to Relay, `log_item`. Both the `log` envelope and `otel_log` envelope is covered by this data category. This will need to implemented in the SDK. Rate limiting applies as usual, there is no special rate limit or rate limiting behaviour for logs. With this data category, client outcomes should be tracked by the SDKs to track how often logs are dropped by the SDK.


develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -268,33 +268,39 @@ An SDK should implement [Tracing without Performance](/sdk/telemetry/traces/trac
...
-6. `sentry.sdk.name`: The name of the SDK that sent the log
-7. `sentry.sdk.version`: The version of the SDK that sent the log
-8. [BACKEND SDKS ONLY] `server.address`: The address of the server that sent the log. Equivalent to `server_name` we attach to errors and transactions.
-
-Example:
+1. `sentry.environment`: The environment set in the SDK if defined.
+2. `sentry.release`: The release set in the SDK if defined.
+3. `sentry.trace.parent_span_id`: The span id of the span that was active when the log was collected. This should not be set if there was no active span.
+4. `sentry.sdk.name`: The name of the SDK that sent the log
+5. `sentry.sdk.version`: The version of the SDK that sent the log
```
philipphofmann, line 277: Do we also plan to support integrations, features, and settings that [already exist](https://develop.sentry.dev/sdk/data-model/event-payloads/sdk/)? These can be handy for investigating issues and tracking the adoption of integrations and features.

develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -303,7 +309,67 @@ If a log is generated by an SDK integration, the SDK should also set the `sentry
...
 ```
 
-Beyond these attributes, we are exploring if the SDK should also send OS, user, and device information automatically (via reading the appropriate contexts from the scope). Given this behaviour can easily be added as a new feature to the SDK, it does not have to be part of the initial SDK implementation until we make a finalized decision.
+#### User Attributes
+
+If `sendDefaultPii`/`send_default_pii` is set to `true` in the SDK, the SDK should attach the following user data if available:
+
+1. `user.id`: The user ID. Maps to `id` in the [User](/sdk/data-model/event-payloads/user/) payload.
+2. `user.name`: The username. Maps to `username` in the [User](/sdk/data-model/event-payloads/user/) payload.
+3. `user.email`: The email address. Maps to `email` in the [User](/sdk/data-model/event-payloads/user/) payload.
```
philipphofmann, line 318: What about the other existing attributes such as ip_address, geo, city, country_code and region? If we deliberately decided against having these, it would be good to explain here why.

develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -303,7 +309,67 @@ If a log is generated by an SDK integration, the SDK should also set the `sentry
...
+{
+  "user.id": "123",
+  "user.name": "john.doe",
+  "user.email": "john.doe@example.com"
+}
+```
+
+#### User Agent Parsing
+
+By default, relay should parse the user agent attached to an incoming log envelope to parse `browser` and `os` information for logs. These attributes should be attached by relay, but SDKs can attach them if they do not forward a user agent when sending logs to Sentry.
```
philipphofmann, line 330: I think we usually uppercase Relay
```suggestion
By default, Relay should parse the user agent attached to an incoming log envelope to parse `browser` and `os` information for logs. These attributes should be attached by Relay, but SDKs can attach them if they do not forward a user agent when sending logs to Sentry.
```

develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -303,7 +309,67 @@ If a log is generated by an SDK integration, the SDK should also set the `sentry
...
+  "browser.name": "Chrome",
+  "browser.version": "120.0"
+}
+```
+
+#### Backend SDKs
+
+For backend SDKs (Node.js, Python, PHP, etc.), the SDKs should attach the following:
+
+1. `server.address`: The address of the server that sent the log. Equivalent to `server_name` we attach to errors and transactions.
```
philipphofmann, line 346: A link to https://develop.sentry.dev/sdk/data-model/event-payloads/#optional-attributes would be nice
```suggestion
1. `server.address`: The address of the server that sent the log. Equivalent to [`server_name`](sdk/data-model/event-payloads/#optional-attributes) we attach to errors and transactions.
```

develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -303,7 +309,67 @@ If a log is generated by an SDK integration, the SDK should also set the `sentry
...
+
+```json
+{
+  "server.address": "foo.example.com"
+}
+```
+
+#### Mobile, Desktop, and Native SDKs
+
+For mobile, desktop, and native SDKs (Android, iOS, Electron, etc.), the SDKs should attach the following:
```
philipphofmann, line 356: It's also tvOS, macOS, watchOS, and visionOS 😄 
```suggestion
For mobile, desktop, and native SDKs (Android, Apple, Electron, etc.), the SDKs should attach the following:
```

philipphofmann, line 356: Again why is this list incomplete? What drove us to only add a subset? It's OK, but we should point out why here.

philipphofmann (approved):
Thanks for adding that. I asked a few questions on why we use a subset of available attributes. Apart from that, LGTM.

develop-docs/sdk/telemetry/logs.mdx
AbhiPrasad: At the current moment no. We want to avoid making the payload too large.

We can re-evaluate this based on internal and external feedback.

develop-docs/sdk/telemetry/logs.mdx
AbhiPrasad: I added a note as part of 9fc68094ccf8db1255c3e2c24e22a650b270a14b to address this.

develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -303,7 +309,73 @@ If a log is generated by an SDK integration, the SDK should also set the `sentry
...
+5. `device.family`: The family of the device. Maps to `family` in the [Contexts](/sdk/data-model/event-payloads/contexts/#device-context) payload.
+
+```json
+{
+  "os.name": "iOS",
+  "os.version": "17.0",
+  "device.brand": "Apple",
+  "device.model": "iPhone 15 Pro Max",
+  "device.family": "iPhone"
+}
```
romtsn, line 371: @AbhiPrasad do we plan to normalize logs too? Asking because e.g. for Android we backfill `device.name` to a human-readable name [here](https://github.com/getsentry/relay/blob/ac71f3590a6514e4e3f119d047a3e160eadc58ca/relay-event-normalization/src/normalize/contexts.rs#L345-L354), so it becomes e.g. `Galaxy Tab A7` instead of `SM-T500` or something like that. Would be great to have that for logs eventually, I suppose.

AbhiPrasad: Great point - I created https://linear.app/getsentry/issue/LOGS-155/enrich-devicename-attribute-to-logs-in-relay to track this issue.

PR #13853, opened 2025-05-27
deprecate(security-reports): Remove reference to Expect-CT and HPKP

armenzg:
These reports have been removed by Chrome a while ago:

- **Expect-CT** was removed in [Chrome 107](https://chromestatus.com/feature/6244547273687040), and Firefox never implemented support
- **HPKP** was removed in [Chrome 72](https://chromestatus.com/feature/5903385005916160) and [Firefox 72](https://groups.google.com/g/mozilla.dev.platform/c/AyMlrNHYepE/m/B5bgjjsiBwAJ) (not a typo)

We will need to follow up with a pull request to remove it from the code base.

### DIFF
--- a/docs/platforms/dotnet/common/security-policy-reporting/index.mdx
+++ b/docs/platforms/dotnet/common/security-policy-reporting/index.mdx
@@ -2,7 +2,7 @@
 title: Set Up Security Policy Reporting
 sidebar_title: Security Policy Reporting
 sidebar_order: 7500
-description: "Learn how Sentry can help manage Content-Security-Policy violations and CSP reports here."
+description: "Learn how Sentry can help manage Content-Security-Policy violations and reports here."
 ---
 
 <Include name="platforms/security-policy-reporting/content.mdx" />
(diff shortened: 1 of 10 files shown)


docs/platforms/dotnet/common/security-policy-reporting/index.mdx
```diff
@@ -2,7 +2,7 @@
 title: Set Up Security Policy Reporting
 sidebar_title: Security Policy Reporting
 sidebar_order: 7500
-description: "Learn how Sentry can help manage Content-Security-Policy violations, CSP reports, Expect-CT, and HTTP Public Key Pinning (HPKP) failures here."
+description: "Learn how Sentry can help manage Content-Security-Policy violations and CSP reports here."
```
mrduncan, line 5: Not sure why this ever called out CSP violations and CSP reports like that - what about something like this for all of these descriptions?

> Learn how Sentry can help manage Content-Security-Policy violations and reports here.

mrduncan (approved):
Just one question to make sure that json file isn't used somewhere that we need to keep around but LGTM otherwise

PR #13899, opened 2025-06-02
ref: add minimum required OS version for relay binary

Litarnus:
## DESCRIBE YOUR PR
Adds description with versions for relay binaries and how long we support each operating system.

I also added a description to the build config to update the documentation in case it gets updated: https://github.com/getsentry/relay/pull/4794

### DIFF
--- a/docs/product/relay/operating-guidelines.mdx
+++ b/docs/product/relay/operating-guidelines.mdx
@@ -31,6 +31,20 @@ The following recommendations assume that Relay is run in Docker.
 
 </p>
 
+## Relay Binary
+
+Every release of Relay also provides a set of binaries that come with a minimum required OS version.
+
+| Operating System  | Version   |
+| Ubuntu x86        | 20.04     |
+| Ubuntu ARM        | 22.04     |
+| MacOS             | macOS Sonoma (version 14)     |
+| Windows           | Windows Server 2019   |
+
+We support LTS versions of the listed Operating Systems as long as they receive updates from their vendors.
+The only exception is Ubuntu x86, which will be supported 2 years after LTS support has officially ended to allow
+customers with extended support to continue using Relay.
+
 ## Example Configuration
 
 This example configuration sets up basic logging and metrics settings, as well as changes the default concurrency level.
--- a/docs/product/relay/operating-guidelines.mdx
+++ b/docs/product/relay/operating-guidelines.mdx
@@ -33,18 +33,18 @@ The following recommendations assume that Relay is run in Docker.
 
 ## Relay Binary
 
-Every release of Relay also provides a set of binaries that come with a minimum required OS version.
+Every release of Relay also provides a set of binaries that are built for the Operating Systems listed below.
 
 | Operating System | Version                   |
 |------------------|---------------------------|
-| Ubuntu x86       | 20.04                     |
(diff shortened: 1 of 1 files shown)


docs/product/relay/operating-guidelines.mdx
```diff
@@ -31,6 +31,21 @@ The following recommendations assume that Relay is run in Docker.
...
+## Relay Binary
+
+Every release of Relay also provides a set of binaries that come with a minimum required OS version.
+
+| Operating System | Version                   |
+|------------------|---------------------------|
+| Ubuntu x86       | 20.04                     |
+| Ubuntu ARM       | 22.04                     |
+| MacOS            | macOS Sonoma (version 14) |
+| Windows          | Windows Server 2019       |
```
Dav1dde: How do we guarantee MacOS and WIndows?

docs/product/relay/operating-guidelines.mdx
```diff
@@ -31,6 +31,21 @@ The following recommendations assume that Relay is run in Docker.
 
 </p>
 
+## Relay Binary
+
+Every release of Relay also provides a set of binaries that come with a minimum required OS version.
+
+| Operating System | Version                   |
+|------------------|---------------------------|
+| Ubuntu x86       | 20.04                     |
```
Dav1dde: 32 bit?

docs/product/relay/operating-guidelines.mdx
```diff
@@ -31,6 +31,21 @@ The following recommendations assume that Relay is run in Docker.
...
+| Ubuntu ARM       | 22.04                     |
+| MacOS            | macOS Sonoma (version 14) |
+| Windows          | Windows Server 2019       |
+
+We support LTS versions of the listed Operating Systems as long as they receive updates from their vendors.
+The only exception is Ubuntu x86, which will be supported 2 years after LTS support has officially ended to allow
+customers with extended support to continue using Relay.
```
Dav1dde: I think we can phrase this differently and don't have to special case ARM/non-ARM, listed version +N years of support after LTS.

Dav1dde (changes requested):

docs/product/relay/operating-guidelines.mdx
Litarnus: We do have runners for those, or do you mean what would happen if Github discontinues those before the offical EOL?

Litarnus: Or are you suggesting to only mention Ubuntu in the table?

docs/product/relay/operating-guidelines.mdx
Litarnus: I'll rephrase it

docs/product/relay/operating-guidelines.mdx
Dav1dde: > We do have runners for those, or do you mean what would happen if Github discontinues those before the offical EOL?

Correct, how do we guarantee what you documented here, it's the same with ubuntu, github turned off the runners before the actual EOL. So if we promise it here, how do we make sure we can hold that promise.

sfanahata (approved):
👍 Looks good! (also, please tag the docs team when y'all are ready for review in future PRs)

PR #13919, opened 2025-06-04
feat(go): improve conflicting wording on page description

giortzisg:
## DESCRIBE YOUR PR

The page description using the word `manually` was conflicting with the first sentence :`automatically reports errors`.

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 


cleptric:
> Sentry's SDK hooks into your runtime environment and automatically reports errors, uncaught exceptions, and unhandled rejections as well as other types of errors depending on...

Should also update this then. This is from a time, where we shared this page across SDKs. But we neither have exceptions in Go nor do we do automatic error reporting, besides panics.

PR #13920, opened 2025-06-04
Update dynamic-sampling-context.mdx

cleptric:


giortzisg:
We also need to update the python pseudo code to match the fix

PR #13936, opened 2025-06-05
feat: Improve Hono docs to cover errors + Node.js

dcramer:
Update the hono include to be semi-correct (uncertain if used currently).


dcramer:
Shipping it. Tell me if somethings bad and well fix it in post.

PR #13940, opened 2025-06-06
Renaming to generalize AI editors; adding logger fmt details

codyde:
* Fixes titles for the AI snippets to be more generic 
* Adds clarity for logger.fmt which was missing from log snippets


nikolovlazar (approved):
Looks good! Although we should come back and fix the repeating content in the `llm-rules` directories.

PR #13951, opened 2025-06-06
Add Turborepo configuration and scripts for improved monorepo management

dcramer:
I implemented Turborepo support to optimize build times across the monorepo. I added `turbo@2.5.4` as a dev dependency and created a `turbo.json` configuration file with optimized task definitions for build, dev, lint, and test commands. In the root `package.json`, I added new turbo-specific scripts (`turbo:dev`, `turbo:build`, `turbo:lint`, `turbo:test`) and set the `packageManager` field to `yarn@1.22.22`. I configured caching by adding `.turbo` to `.gitignore`. The `turbo.json` includes task configurations with proper dependency chains, output caching for `.next/**` and `dist/**` directories, and environment variable handling. The setup enables parallel task execution and intelligent caching across workspaces, significantly improving build performance especially for unchanged code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,9 +138,7 @@
   "resolutions": {
     "dompurify": "3.2.4",
     "@types/dompurify": "3.0.5",
-    "@sentry/core": "9.27.0",
-    "@sentry/types": "9.27.0",
-    "@sentry/utils": "9.27.0"
+    "@sentry/core": "9.27.0"
   },
   "volta": {
     "node": "20.11.0",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -136,10 +136,12 @@
     "ws": "^8.17.1"
   },
   "resolutions": {
-    "dompurify": "3.2.4",
-    "@types/dompurify": "3.0.5",
     "@sentry/core": "9.27.0",
-    "@sentry/types": "9.27.0"
+    "@sentry/types": "9.27.0",
+    "@sentry/nextjs": "9.27.0",
```
dcramer, line 141: hrm why wasnt nextjs already in here? i think this package.json is actually for the main docs site?

PR #13968, opened 2025-06-09
Migrate from yarn to pnpm

dcramer:


chargome:
@dcramer can we close this one and continue in https://github.com/getsentry/sentry-docs/pull/13992?

PR #13972, opened 2025-06-09
Fix copy button output on logging page

dcramer:
The copy logic within the `Expandable` component in `src/components/expandable/index.tsx` was refined to prevent unwanted inline code from being copied.

Previously, the copy button selected all `<code>` elements, including inline code snippets like `.cursorrules` and `rules.md` from explanatory text.

The changes implement a more selective approach:
*   **Primary Selection**: The component now first attempts to copy content from `pre code` elements, which typically contain main code blocks.
*   **Fallback Filtering**: If no `pre code` blocks are found, a fallback mechanism filters all `<code>` elements. It excludes:
    *   Code elements found within paragraphs.
    *   Short code snippets (less than 100 characters), which are likely inline.
*   **Final Fallback**: If no suitable code blocks are identified, the entire text content of the expandable section is copied.

This ensures that only the intended code content is copied, omitting extraneous inline text.

### DIFF
--- a/src/components/expandable/index.tsx
+++ b/src/components/expandable/index.tsx
@@ -1,10 +1,10 @@
 'use client';
 
-import {ReactNode, useCallback, useEffect, useRef, useState} from 'react';
-import {ChevronDownIcon, ChevronRightIcon} from '@radix-ui/react-icons';
+import { ReactNode, useCallback, useEffect, useRef, useState } from 'react';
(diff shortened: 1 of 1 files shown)


evanpurkhiser:
Didn't see this and had a fix here too https://github.com/getsentry/sentry-docs/pull/13973

PR #13973, opened 2025-06-09
fix(ui): Only copy CodeBlock expandable content

evanpurkhiser:
Prior to this change all `code` elements would be copied, so if the
markdown content inside a expandable alert included single-backtick
style `<code>whatever</code>` nodes, those would end up in the copied
content.

The intention of this component is definitely intended to just copy the
CodeBlock content.


dcramer:
Dupe of #13972 - we could do both changes, though I'm not sure this change with the way its struct ured wouldn't break anything else

evanpurkhiser:
Cursor change seems fine for now

PR #13980, opened 2025-06-10
chore(dsn): update reference to dsn settings from settings -> client keys to settings -> sdk setup -> client keys

shellmayr:
- There was some feedback about not finding the client keys - looking at the settings, it makes sense that you would look only under Project > Settings > Project and not further down. 
- Not sure whether this is a conscious choice to not refer to the sub-group in settings. If this change is useful, we should probably also update all other project settings paths in the docs.


coolguyzone (approved):
Thanks for clarifying 🫡

PR #13981, opened 2025-06-10
fix(seer): fix broken link

lcian:
Fixes link that lead to a 404


lcian:
Hoped this would lead to a fix of the Vercel deployment but apparently that's unrelated to this

sl0thentr0py (approved):

PR #13984, opened 2025-06-10
feat(native): update before_send docs

JoshuaMoelans:
[Latest Vercel Preview](https://sentry-docs-git-joshua-nativebeforesenddocs.sentry.dev/platforms/native/configuration/filtering/#event-hints)

## DESCRIBE YOUR PR
Fixes https://github.com/getsentry/sentry-native/issues/1268

We fix some ancient docs which wrongly state that `before_send` uses the `hint` parameter in the Native SDK, by cleaning up the unnecessary 'Event Hints' section.

We move the example snippet into the corresponding `before_send` and `on_crash` sections, better showcasing how to use the passed-in values.

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST


platform-includes/configuration/before-send-hint/native.mdx
```diff
@@ -17,7 +17,7 @@ sentry_value_t before_send(sentry_value_t event, void *hint) {
   const char *type_str = sentry_value_as_string(type);
 
   /* use the data passed during initialization */
-  const char *custom_error = (const char *)hint;
+  const char *custom_error = (const char *)user_data;
```
supervacuus: I think this is still confusing, because it characterizes `user_data` as something that should contribute to error resolution, which is absolutely not what it is used for. `user_data` is about passing in context from the user, so they can call into their code or retrieve relevant data. 

`hint` and `user_data` are entirely unrelated. If `user_data` gets assigned to a hook variable called `custom_error`, that weakens the clear distinction we should accomplish in the docs.

platform-includes/configuration/before-send-hint/native.mdx
```diff
@@ -1,14 +1,14 @@
 <Alert>
 
-The Native SDK does not pass exception information into the hint, but instead allows to set a custom hint in SDK options. To detect the exception type, inspect the event instead.
+The Native SDK does not pass exception information into the hint, but instead allows to set custom `user_data` in the `before_send` SDK option. To detect the exception type, inspect the event instead.
```
supervacuus: We should note here that the Native SDK's solution to the problem of obtaining additional crash context is the `on_crash` hook, and that the usage of the `before_send` hook for handling crashes is deprecated and provided only as a compatibility fallback.

This is particularly confusing because the entire first section of this document clearly addresses this particular issue. I guess that this snippet has only "recently" been added as an included section below as part of a larger refactoring, because otherwise this confusing and contradictory explanation would have been evident in documenting the much later added `on_crash` hook.

platform-includes/configuration/before-send/native.mdx
```diff
@@ -28,7 +38,7 @@ static sentry_value_t
 crash_cleanup(
     const sentry_ucontext_t *uctx, // provides the user-space context of the crash
     sentry_value_t event,          // used the same way as in `before_send`
-    void *closure                  // user-data that you can provide at configuration time
+    void *user_data                  // user-data that you can provide at configuration time
```
JoshuaMoelans: I'm not quite sure what kind of cleanup one could do with the data one gets from the `uctx` and `event`; @supervacuus do you have any inspiration? I was thinking about using fields from `uctx->siginfo` to discard specific types of crashes, but maybe you know of something nicer.

platform-includes/configuration/before-send/native.mdx
```diff
@@ -1,14 +1,24 @@
 ```c
 #include <sentry.h>
 
-sentry_value_t strip_sensitive_data(sentry_value_t event, void *hint) {
+sentry_value_t strip_sensitive_data(sentry_value_t event, void *hint, void *user_data) {
   /* modify event here or return NULL to discard the event */
+  (void)hint;
```
supervacuus: I think this would be a sensible location to specify that `hint` is currently not expected to contain any data, and maybe to repeat that users who want more crash context should use the `on_crash` hook.

platform-includes/configuration/before-send/native.mdx
supervacuus: Clean-up is the wrong concept for `on_crash` to convey to users. This is precisely where `before_send` and `on_crash` dramatically diverge:

* the event we get in `on_crash` is essentially empty for all minidump generating backends
* `inproc` is the only backend that passes complete event data to either hook
* so, if you aren't using `inproc`, the `event` parameter is not a proper clean-up or filter discriminator
* the event can, up to a certain point, be _enriched_ inside the `on_crash` hook as the last client interaction before the crashed application terminates
* in the same vein, it can be used to act _at all_ before the crashed application terminates
* while the `ucontext` can provide a rich source of metadata, which can find its way back into the event or as a filter mechanism, the interactions would have to be very specific to a particular use case, and describing those might lead to more confusion than insight
* it is enormously rare, for any minidump generating backend to remove any events, solely by looking at the `ucontext`, since the biggest source of meta-data is the snapshot collection for the minidump which is currently not accessible in the client (and if we would make it accessible we probably wouldn't go indirectly via minidump)

Having said all that, a concrete example from the field was from a user who reported C++ exceptions via a `std::terminate` handler that would capture a manual exception event and didn't want the resulting crash of an unhandled C++ exception to generate an additional error report. To prevent sending C++ exceptions on Windows, I provided an `on_crash` approach for this particular scenario: https://github.com/getsentry/sentry-native/issues/860#issuecomment-2231670774. However, I question whether this approach falls into the "nice" category.

platform-includes/configuration/before-send/native.mdx
JoshuaMoelans: Adapted to
```c
sentry_value_t strip_sensitive_data(sentry_value_t event, void *hint, void *user_data) {
    (void)hint; // unlike in other SDKs, `hint` currently contains no data.
    // to get more crash context into a callback, use the `on_crash` hook instead.

```

platform-includes/configuration/before-send/native.mdx
JoshuaMoelans: Updated the wording in the [last commit](https://github.com/getsentry/sentry-docs/pull/13984/commits/de9309f920caeb054cfa0e4c40afcaa60ee2e671#diff-f0e10746d49e60c000ac2dfad2b50ae00f64bf45d370fdc105a8fc84b7fe4404R42-R53) , let me know if this better describes the use-case for `on_crash`.

PR #13994, opened 2025-06-11
feat(ai): Add .md extension to provide pages in markdown for LLMs

codyde:
Adds support for `.md` at the end of every pre-rendered path. Does this by a hack where we rewrite those paths to `public/md-exports/...`. The contents of this directory are generated _after_ `next build` by scraping all html files under `.next/server/app`

Not ideal but looks like the easiest path for now.

Also added a "View Markdown version" link to pages:

image


src/middleware.ts
```diff
@@ -62,6 +67,31 @@
...
+    const pathSegments = originalPath.split('/').filter(Boolean);
+    
+    // Rewrite to the API route with path segments
+    const apiPath = pathSegments.length > 0 
+      ? `/api/llms-txt/${pathSegments.join('/')}`
+      : '/api/llms-txt';
+    
+    const apiUrl = new URL(apiPath, request.url);
+    
+    return NextResponse.rewrite(apiUrl);
```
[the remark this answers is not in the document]
codyde: Bumped this up - currently testing!

BYK:
So this is a thing now, yay: https://sentry-docs-mq7fdge2k.sentry.dev/platforms/javascript/guides/react.md

next.config.ts
```diff
@@ -24,6 +24,7 @@ const outputFileTracingExcludes = process.env.NEXT_PUBLIC_DEVELOPER_DOCS
     };
 
 if (
+  !process.env.NODE_ENV &&
```
chargome: m: Is this one intended? This condition will always fail, even when NODE_ENV is `production`

src/middleware.ts
```diff
@@ -19,6 +19,7 @@ export const config = {
 
 // This function can be marked `async` if using `await` inside
 export function middleware(request: NextRequest) {
+  // Remove the llms.txt handling - it's now handled by Next.js redirects
```
chargome: Leftover?

next.config.ts
```diff
@@ -55,6 +56,12 @@ const nextConfig = {
     DEVELOPER_DOCS_: process.env.NEXT_PUBLIC_DEVELOPER_DOCS,
   },
   redirects,
+  rewrites: async () => [
+    {
+      source: '/:path*.md',
+      destination: '/md-exports/:path*.md',
+    },
+  ],
```
chargome, line 63: Nice idea 👍

chargome (commented):
Implementation looks pretty solid to me! 

The script adds more than 3min to every build now, which will only grow in the future.
@BYK just thinking loud but do you see any chance we could create these at runtime on request? I imagine performance is not the biggest concern for these pages?

And last point we probably don't want these pages to show up in google, can we add the according meta tag for these pages?

src/middleware.ts
codyde: Def leftover by me - crud from the AI cleanup

next.config.ts
BYK: Intended as otherwise `yarn build` fails locally, asking for these DSNs. That said I _think_ the negation at the front is incorrect. I wonder how this worked locally.

BYK:
@chargome 
>The script adds more than 3min to every build now, which will only grow in the future.

Yeah, we have 8k+ files so it takes a while. That said I do not think the build time will grow significantly as even if you added 100 more docs it's still about 10% of the total docs we have and need to walk through.

>@BYK just thinking loud but do you see any chance we could create these at runtime on request? I imagine performance is not the biggest concern for these pages?

As far as I'm aware Next.js does not allow the `*.md` routing. I think if we can figure out the routing, this should be doable. I also proposed adding a front-proxy, probably using Cloudflare, that can do all this stuff on the fly and cache it. Not sure if Vercel Functions may also help with this.

>And last point we probably don't want these pages to show up in google, can we add the according meta tag for these pages?

Where should I add these tags? These pages are not linked from anywhere (except from the link I just added and I'll add `rel=nofollow` to them right after posting this message) so not sure how they would end up in Google. Also I cannot add meta tags to MD files, do you mean some sort of a header or something?

next.config.ts
BYK: @codyde's or Cursor's 😅

next.config.ts
BYK: Okay reverted this as `next build` sets `NODE_ENV` to `production` by default if it was not set 😮 

Added a `yarn start:dev` command instead to do what I needed.

chargome:
> As far as I'm aware Next.js does not allow the *.md routing. I think if we can figure out the routing, this should be doable. I also proposed adding a front-proxy, probably using Cloudflare, that can do all this stuff on the fly and cache it. Not sure if Vercel Functions may also help with this.

Is the .md essential for LLMs? Otherwise we could possibly just add a search param `?md=true` and e.g. redirect to a dynamic route instead.

>  Also I cannot add meta tags to MD files, do you mean some sort of a header or something?

Nevermind, I thought the HTML came from the server here – ignore my comment 👍

BYK:
>Is the .md essential for LLMs? Otherwise we could possibly just add a search param ?md=true and e.g. redirect to a dynamic route instead.

Yeah we want the `.md` part without redirects as that's the best user experience and what seems to be established in the industry.

I'll work on a patch to speed things up using NextJS build cache.

PR #14009, opened 2025-06-11
Updates to Seer and AI docs

ItalyPaleAle:
1. Branding: Ensure "issue fix" and "issue scan" are lowercased (when not used in a title)
2. Moved "Seer Privacy and Security" to "AI Privacy and Security" in the higher-level directory, since it applies to all gen AI features (not just Seer)
3. Shortened the privacy and security details in the main Seer's page, which was duplicate with what's in the legal page (it's linking to that for more details)
4. Moved "Disabling gen AI features" to the "AI in Sentry" page since that applies to all AI features not just Seer
5. Updated Seer privacy/legal term: removed note that it's not available in the EU (it is available now), and specified that we do not train gen AI on customer data by default (before it said only "without express consent" but wasn't clear that consent is not given by default) - this was approved by legal for a separate whitepaper


ItalyPaleAle:
Merged with rebase because @trillville has a branch on top of this

PR #14011, opened 2025-06-12
:sparkles: feat(integrations): add github domain checker

iamrajjoshi:
folks are often confused about if they should use the Github Integration or the Github Enterprise Integration. to try to combat this, i added a component (based off of `CreateGitHubAppForm`) which parses the domain the user inputs and tells them what integration to install

https://github.com/user-attachments/assets/ed3890a4-26b4-40c4-ae8c-4fdba679105b

### DIFF
--- a/docs/organization/integrations/source-code-mgmt/github/index.mdx
+++ b/docs/organization/integrations/source-code-mgmt/github/index.mdx
@@ -12,12 +12,21 @@ There are two types of GitHub integrations, so make sure you're choosing the cor
   - `https://ghe.example.com/`
 
   For these accounts, follow the '[Installing GitHub Enterprise](#installing-github-enterprise)' instructions.
+  
+
+### To determine which integration to use, enter your GitHub domain below:
+
+<GitHubDomainChecker />
+
+<br />
 
 <Alert>
 
 Sentry owner, manager, or admin permissions, and GitHub owner permissions are required to install this integration.
 
 </Alert>
+
+## Installing GitHub
 
 1. Navigate to **Settings > Integrations > [GitHub](https://sentry.io/orgredirect/organizations/:orgslug/settings/integrations/github)**.
 
--- a/docs/organization/integrations/source-code-mgmt/github/index.mdx
+++ b/docs/organization/integrations/source-code-mgmt/github/index.mdx
@@ -14,7 +14,9 @@ There are two types of GitHub integrations, so make sure you're choosing the cor
   For these accounts, follow the '[Installing GitHub Enterprise](#installing-github-enterprise)' instructions.
   
 
-### To determine which integration to use, enter your GitHub domain below:
+### Check Your Domain
+
+To determine which integration to use, enter your GitHub domain below:
 
 <GitHubDomainChecker />
 
--- a/src/components/githubDomainChecker.tsx
+++ b/src/components/githubDomainChecker.tsx
@@ -1,0 +1,91 @@
+'use client';
+
+import {useState} from 'react';
+
+const MAX_COMPONENTS_ON_PAGE = 100;
+
+export function GitHubDomainChecker() {
+  const [domain, setDomain] = useState('');
+  const [isValidDomain, setIsValidDomain] = useState(false);
+  
+  const isGitHubCom = domain.toLowerCase().trim() === 'github.com' || domain.toLowerCase().trim() === 'https://github.com';
+  const hasInput = domain.trim().length > 0;
+  
+  // Validate domain format
+  const validateDomain = (inputDomain: string) => {
+    const trimmedDomain = inputDomain.trim();
+    if (!trimmedDomain) {
+      setIsValidDomain(false);
+      return;
+    }
+    
+    // Check if it's github.com (valid)
+    if (trimmedDomain.toLowerCase() === 'github.com' || trimmedDomain.toLowerCase() === 'https://github.com') {
+      setIsValidDomain(true);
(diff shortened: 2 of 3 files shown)


docs/organization/integrations/source-code-mgmt/github/index.mdx
```diff
@@ -12,13 +12,22 @@ There are two types of GitHub integrations, so make sure you're choosing the cor
   - `https://ghe.example.com/`
 
   For these accounts, follow the '[Installing GitHub Enterprise](#installing-github-enterprise)' instructions.
+  
+
+### To determine which integration to use, enter your GitHub domain below:
```
leeandher: I would change this to `Check your Domain` since these show up in the Table of Contents on the right.

You can add the full text underneath as a preamble to the component.

src/components/githubDomainChecker.tsx
leeandher, line 1: image

Text is hard to read on dark mode

coolguyzone:
Should the form input darken in dark mode, similar to the search input at the top of the page?

Screenshot 2025-06-12 at 1 37 06 PM

src/components/githubDomainChecker.tsx
```diff
@@ -0,0 +1,96 @@
...
+          value={domain}
+          placeholder="https://github.com or https://ghe.example.com"
+          className={inputClassName}
+          onChange={handleDomainChange}
+        />
+      </div>
+
+      {hasInput && (
+        <div className="mt-4 p-4 rounded-md border">
+          <div className="text-sm font-medium mb-2">Recommended Installation:</div>
```
leeandher: This probably only should be shown if it's a valid domain.

src/components/githubDomainChecker.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <div className="flex w-full">
+        <div className="flex items-center min-w-[16ch] px-4">
+          <label htmlFor={`gh-domain-${randomCounter}`} className="text-nowrap">
+            GitHub Domain
+          </label>
+        </div>
+        <input
+          id={`gh-domain-${randomCounter}`}
+          value={domain}
+          placeholder="https://github.com or https://ghe.example.com"
```
coolguyzone, line 104: should the first placeholder url be `https://github.example.com`?

src/components/githubDomainChecker.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const newDomain = ev.target.value;
+    setDomain(newDomain);
+    validateDomain(newDomain);
+  };
+
+  const inputClassName =
+    'form-input w-full rounded-md border-[1.5px] focus:ring-2 focus:ring-accent-purple/20 border-gray-200';
+
+  // This is to avoid in case multiple instances of this component are used on the page
+  const randomCounter = Math.round(Math.random() * MAX_COMPONENTS_ON_PAGE);
```
leeandher: This is an interesting pattern! could we avoid it by passing in a `key` or `id` prop when rendering, or is that like a limitation of MDX?

src/components/githubDomainChecker.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+const MAX_COMPONENTS_ON_PAGE = 100;
+
+export function GitHubDomainChecker() {
+  const [domain, setDomain] = useState('');
+  const [isValidDomain, setIsValidDomain] = useState(false);
+
+  const isGitHubCom =
+    domain.toLowerCase().trim() === 'github.com' ||
+    domain.toLowerCase().trim() === 'https://github.com';
```
leeandher: I think maybe we should relax this a bit, if I add my GH user it recommends the wrong installation. probably user error by adding my username after but still!

image

leeandher (commented):
this is cool! hopefully it makes things less confusing

src/components/githubDomainChecker.tsx
iamrajjoshi: gonna just remove it, they component is only gonna be used once

src/components/githubDomainChecker.tsx
iamrajjoshi: i relazed this just a bit so it checks if the input starts with `github.com`

src/components/githubDomainChecker.tsx
```diff
@@ -0,0 +1,140 @@
...
+      </div>
+
+      {hasInput && (
+        <div className="mt-4 p-4 rounded-md border dark:border-gray-600">
+          {isValidDomain ? (
+            <div>
+              <div className="text-sm font-medium mb-2">Recommended Installation:</div>
+              {isGitHubCom ? (
+                <div className="text-green-700 bg-green-50 dark:text-green-300 dark:bg-green-900/30 p-3 rounded-md">
+                  <div className="mb-2">
```
leeandher: nit: I dont think the margin here is necessary, it makes for uneven spacing.
image

src/components/githubDomainChecker.tsx
```diff
@@ -0,0 +1,140 @@
...
+                    <strong>GitHub</strong> - Use the standard GitHub integration for
+                    github.com
+                  </div>
+                </div>
+              ) : (
+                <div className="text-blue-700 bg-blue-50 dark:text-blue-300 dark:bg-blue-900/30 p-3 rounded-md">
+                  <div className="mb-2">
```
leeandher: here neither

leeandher (approved):

PR #14033, opened 2025-06-13
Logs for Laravel

cleptric:


platform-includes/logs/usage/php.laravel.mdx
```diff
@@ -0,0 +1,20 @@
...
+
+// Log to all channels in the stack (including Sentry)
+Log::info('This is an info message');
+Log::warning('This is a warning message');
+Log::error('This is an error message');
+
+// Log directly to the Sentry channel
+Log::channel('sentry')->error('This will only go to Sentry');
+
+// Log with additional context
```
AbhiPrasad: I'd make this a separate snippet and include:

> You can pass additional attributes directly to the logging functions. These properties will be sent to Sentry, and can be searched from within the Logs UI, and even added to the Logs views as a dedicated column.


ref: https://docs.sentry.io/platforms/php/logs/

PR #14037, opened 2025-06-16
Update index.mdx

InterstellarStella:
Clarifying the deprecation message to avoid any confusion. A customer was confused about the alert message explaining that Splunk Cloud is supported.


InterstellarStella:
@coolguyzone We've had another customer who was confused about the current wording in this docs page. Could you take a look and merge this PR? Thanks!

PR #14096, opened 2025-06-20
ci(md): Add caching to md-export script

BYK:
Should cut down build times significantly after the first run if the cache persists in Vercel

Locally my times went from 60 seconds down to 8 seconds.

### DIFF
--- a/scripts/generate-md-exports.mjs
+++ b/scripts/generate-md-exports.mjs
@@ -1,12 +1,12 @@
 #!/usr/bin/env node
 
-import {fileURLToPath} from 'url';
-
 import {selectAll} from 'hast-util-select';
-import {existsSync} from 'node:fs';
-import {mkdir, opendir, readFile, rm, writeFile} from 'node:fs/promises';
+import {createHash} from 'node:crypto';
+import {constants as fsConstants, existsSync} from 'node:fs';
+import {copyFile, mkdir, opendir, readFile, rm, writeFile} from 'node:fs/promises';
 import {cpus} from 'node:os';
 import * as path from 'node:path';
+import {fileURLToPath} from 'node:url';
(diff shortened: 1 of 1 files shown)


scripts/generate-md-exports.mjs
```diff
@@ -63,40 +80,54 @@ async function createWork() {
...
   console.log('✅ Markdown export generation complete!');
 }
 
-async function genMDFromHTML(source, target) {
+const md5 = data => createHash('md5').update(data).digest('hex');
+
+async function genMDFromHTML(source, target, {cacheDir, noCache}) {
   const text = await readFile(source, {encoding: 'utf8'});
+  const hash = md5(text);
+  const cacheFile = path.join(cacheDir, hash);
```
AbhiPrasad, line 121: We generate a lot of markdown files right? I wonder if there are issues caching them all in a single directory. Maybe we should split them up to be in different folders?

AbhiPrasad (approved):
I wonder if we can track cache hit/miss rate?

if we run sentry, we can emit logs and send them to sentry 😄 

Generally I think this is solid! We should try to cache more things in our build.

scripts/generate-md-exports.mjs
BYK: It's about 8.5k files. I think we're fine up to 10k but I'll split if I notice a slow down

BYK:
>I wonder if we can track cache hit/miss rate?
>Edit: If we track cache hit/miss rate if we can log them in the vercel build as well to see how well this is working.
>if we run sentry, we can emit logs and send them to sentry 😄

Happy to add this on afterwards. Should be easy enough.

>Generally I think this is solid! We should try to cache more things in our build.

If you have anything particular in mind, LMK. `next build` is _supposed to_ cache many things but no idea if it actually works or not.

PR #14104, opened 2025-06-23
feat(go): Structured logging support for slog

giortzisg:
## DESCRIBE YOUR PR

modifying the existing `slog` docs to add the log integration changes.

### DIFF
--- a/docs/platforms/go/common/logs/index.mdx
+++ b/docs/platforms/go/common/logs/index.mdx
@@ -120,9 +120,13 @@ slogger.Info("Implementing slog.Logger")
   Sentry UI.
 </Alert>
 
+### Integrations
+
+- <PlatformLink to="/guides/slog"> Slog </PlatformLink>
+
 ### Upcoming Integrations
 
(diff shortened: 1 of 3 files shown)


docs/platforms/go/guides/logrus/index.mdx
```diff
@@ -25,61 +27,147 @@ import (
...
 ```
 
 ## Configure
 
-`sentrylogrus` allows configuration via the `New` function, which accepts the levels to log and `sentry.ClientOptions`. The levels to log are the logrus levels that should be sent to Sentry. The `sentry.ClientOptions` are the same as the ones used in the `sentry.Init` function.
+`sentrylogrus` provides two types of hooks to configure the integration with Sentry. Both hooks accept these options:
+- **Levels**: A slice of `logrus.Level` specifying which log levels to capture
+- **ClientOptions**: Standard `sentry.ClientOptions` for configuration
+
+### LogHook
```
AbhiPrasad, line 116: IMO we should add minimum versions here (so users also know to upgrade).

giortzisg: added minimum version requirements for both logrus & slog.

AbhiPrasad (approved):

PR #14109, opened 2025-06-23
ci(build): Parallelize and cache mdx pipeline - fix md cache

BYK:
Follow up to #14096.

- Makes the entire `mdx.ts` and accompanying modules async
- Limits concurrency in `mdx.ts` async ops (otherwise we crash Vercel Functions / AWS Lambda) (check https://sentry-docs-git-byk-cimdx-cache.sentry.dev/platform-redirect/ -- should not crash)
- Adds compression to caches (as Vercel complained about the function size)
- Removes the `<script>` blocks from the HTML for cache key calculation and faster `md` generation. Script blocks are already ignored and they are not stable across builds even when nothing changes, causing cache misses.

Cuts down build times from ~21-22 minutes to ~13-14 minutes.


docs/product/explore/session-replay/web/index.mdx
```diff
@@ -4,8 +4,6 @@ sidebar_order: 10
 description: "Learn about Session Replay and its video-like reproductions of user interactions, which can help you see when users are frustrated and build a better web experience."
 ---
 
-<Include name="feature-stage-beta-session-replay.mdx" />
```
BYK, line 7: This file did not exist at all.

docs/product/sentry-basics/performance-monitoring.mdx
```diff
@@ -4,8 +4,6 @@ sidebar_order: 1
 description: "Understand and monitor how your application performs in production. Track key metrics, analyze bottlenecks, and resolve performance issues with distributed tracing, detailed transaction data, and automated issue detection."
 ---
 
-<Include name="performance-moving.mdx" />
```
BYK, line 7: This file did not exist at all.

BYK:
Cuts down build times quite notably (from ~21-22 mins down to ~~14-15~~ 13-14 mins)

chargome:
@BYK the dynamic pages are crashing, can you check that out: https://sentry-docs-bh9dt4677.sentry.dev/platform-redirect

platform-includes/sourcemaps/overview/javascript.cloudflare.mdx
```diff
@@ -19,9 +19,7 @@ If you want to configure source maps to upload manually, follow the guide for yo
 
 ### Guides for Source Maps
 
-- <PlatformLink to="/sourcemaps/uploading/typescript/">
```
BYK, line 22: @AbhiPrasad this was causing a visual glitch on https://docs.sentry.io/platforms/javascript/guides/cloudflare/sourcemaps/

image

platform-includes/sourcemaps/upload/primer/javascript.cloudflare.mdx
```diff
@@ -7,6 +7,6 @@ If you can't find the tool of your choice in the list below, we recommend you ch
 
 </Alert>
 
-<Include name="sourcemaps/overview/javascript.cloudflare.mdx" />
```
BYK, line 10: This include was simply being ignored as it could not be found. With the new system, this causes a build error so I fixed it.

Check [current](https://docs.sentry.io/platforms/javascript/guides/cloudflare/sourcemaps/uploading/) vs [fixed](https://sentry-docs-git-byk-cimdx-cache.sentry.dev/platforms/javascript/guides/cloudflare/sourcemaps/uploading/)

src/mdx.ts
```diff
@@ -334,25 +465,68 @@ export async function getFileBySlug(slug: string) {
...
-  const sourcePath =
-    [mdxPath, mdxIndexPath, mdPath, versionedMdxIndexPath].find(fs.existsSync) ??
-    mdIndexPath;
+  let source: string | undefined = undefined;
+  let sourcePath: string | undefined = undefined;
+  const sourcePaths = [mdxPath, mdxIndexPath, mdPath, versionedMdxIndexPath, mdIndexPath];
+  const errors: Error[] = [];
+  for (const p of sourcePaths) {
+    try {
+      source = await readFile(p, 'utf8');
```
AbhiPrasad, line 513: I wonder if calling `Promise.allSettled` with `access` then only running `readFile` on the found file is faster here.

src/files.ts
```diff
@@ -1,37 +1,13 @@
...
  */
-const getAllFilesRecursively = (folder: string): [string] => {
-  return pipe(
-    // yes, this arrow function is necessary to narrow down the readdirSync overload
-    (x: string) => fs.readdirSync(x),
-    map(pipe(pathJoinPrefix(folder), walkDir)),
-    // flattenArray
-    x => x.flat(Infinity)
-  )(folder) as [string];
+const getAllFilesRecursively = async (folder: string): Promise<string[]> => {
```
AbhiPrasad, line 7: should we cache this?

AbhiPrasad (approved):
just some theories about further improvements. Looks good overall!

src/mdx.ts
BYK: There is this common thinking of "check before using, avoid exceptions" which is sensible but the counter intuitive fact is reading/getting already includes a check where the exception arises from hence it is more efficient to just try reading the file. This is what NodeJS docs recommend:

https://nodejs.org/api/fs.html#fsexistspath-callback
>Using `fs.exists()` to check for the existence of a file before calling `fs.open()`, `fs.readFile()`, or `fs.writeFile()` is not recommended. Doing so introduces a race condition, since other processes may change the file's state between the two calls. Instead, user code should open/read/write the file directly and handle the error raised if the file does not exist.

>In general, check for the existence of a file only if the file won't be used directly, for example when its existence is a signal from another process.

(btw same thing goes for values in maps etc: don't check just get and fallback)

src/files.ts
BYK: Normally I'd say yes and even jumped on it when I saw your comment but then realized why I didn't do it: the only callsites for this helper are `getAllFilesFrontMatter` and `getDevDocsFrontMatter` which ~~both have a cache themselves~~.

Looks like only `getAllFilesFrontMatter` has one, adding one for the other before I resolve this comment and merge the patch. Thanks a lot for the pointer!

BYK: Didn't make a noticable difference in build times but still good practice 🚀

PR #14179, opened 2025-06-26
ref: Reduce bundle size for Vercel functions

BYK:


BYK:
checked `/platform-redirect` page but not other dynamic functions so we should get them checked

next.config.ts
```diff
@@ -5,22 +5,38 @@ import {redirects} from './redirects.js';
...
+        './.next/cache/mdx-bundler/**/*',
+        './.next/cache/md-exports/**/*',
         './apps/**/*',
         'develop-docs/**/*',
-        'node_modules/@esbuild/darwin-arm64',
+        'node_modules/@esbuild/*',
       ],
-      '/platform-redirect': ['**/*.gif', 'public/mdx-images/**/*', '*.pdf'],
+      '/platform-redirect': ['**/*.gif', 'public/mdx-images/**/*', '**/*.pdf'],
       '\\[\\[\\.\\.\\.path\\]\\]': [
```
codyde, line 28: @BYK - in my PR when I was looking at this I had some some indications from Claude that this wasn't actually a path - just want to double check if you if you see this one as correct? I refactored out of it in mine but if yours is good, than im good with it.

codyde: specifically `'\\[\\[\\.\\.\\.path\\]\\]'`

BYK: I think it's this one: https://docs.sentry.io/platform-redirect/?

It also broke on one of my earlier PRs 😅

BYK: oh you mean the grep?

codyde: Yeah - looks like it isnt breaking anyhing now, if it comes up - we can revert and figure it out.

codyde (approved):

next.config.ts
chargome: `\\[\\[\\.\\.\\.path\\]\\]` maps to [[...path]], which is the catchall route for every docs content page 👍

codyde:
Yeah - that's kinda what I was curious about if it should've just been
[[...path]] instead, but i assume we had it escaped out for a reason like
the existing way for a reason, and Claude yelled about it so it made
me 🤔 more.

PR #14193, opened 2025-06-27
feat(md): Use page title as the top level title

BYK:
context: markdown files with `# Tags` as the h1 are basically useless


dcramer:
context: markdown files with `# Tags` as the h1 are basically useless

BYK:
Will leave the "bug" Cursor found in as we wanna fail if the page doesn't have a title

dcramer:
@sentry review

dcramer:
its ugly but its an improvement

dcramer (approved):

PR #14196, opened 2025-06-27
feat(md): Rewrite URLs to be absolute and to .md versions

BYK:


dcramer (approved):

scripts/generate-md-exports.mjs
```diff
@@ -247,6 +250,17 @@ async function genMDFromHTML(source, target, {cacheDir, noCache}) {
           }),
         },
       })
+      .use(RemarkLinkRewrite, {
```
BYK, line 253:
```suggestion
      .use(RemarkLinkRewrite, {
      // There's a chance we might be changing absolute URLs here
      // We'll check the code base and fix that later
```

PR #14201, opened 2025-06-30
fix(native): Update Trace Boundaries documentation

JoshuaMoelans:
## DESCRIBE YOUR PR
Adding some more detailed explanation on the Native handling of trace boundaries & when/how to set a new trace.

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

### DIFF
--- a/platform-includes/performance/add-spans-example/native.mdx
+++ b/platform-includes/performance/add-spans-example/native.mdx
@@ -38,13 +38,21 @@ void perform_checkout() {
 
 This example will send a transaction named `checkout` to Sentry. The transaction will contain a `validation` span that measures how long `validate_shopping_cart()` took and a `process` span that measures `process_shopping_cart()`. Finally, the call to `sentry_transaction_finish()` will finish the transaction and send it to Sentry.
 
-By default, transactions will inherit the `trace_id` and `parent_span_id` from the `propagation_context` as generated during SDK initialization. These values can be overridden by calling `sentry_set_trace()`, for example from a downstream SDK or to define a manual trace boundary:
+## Trace Boundaries
+By default, transactions will inherit the `trace_id` and `parent_span_id` from the `propagation_context` as generated during SDK initialization. These values can be overridden by calling `sentry_set_trace()`, for example from a downstream SDK, or to define a manual trace boundary:
 
 ```c
 // Set the trace propagation_context which will be used
 //  for all events and transactions/spans
+
 const char *trace_id = "2674eb52d5874b13b560236d6c79ce8a";
 const char *parent_span_id = "a0f9fdf04f1a63df";
 
 sentry_set_trace(trace_id, parent_span_id);
+perform_checkout(); // trace set before checkout
+...
+sentry_set_trace(trace_id_2, parent_span_id_2);
 ```
+
+The format of the `trace_id` and `parent_span_id` should follow the ones defined in [the telemetry docs](https://develop.sentry.dev/sdk/telemetry/traces/#header-sentry-trace). The `trace_id` should consist of a 32 character hexadecimal string, and the `parent_span_id` should be a 16 character hexadecimal string. For downstream SDKs, these values should match the currently active trace (and span if there is one). For manual trace boundaries, these should be randomly generated.
+


platform-includes/performance/add-spans-example/native.mdx
```diff
@@ -38,13 +38,21 @@ void perform_checkout() {
...
 // Set the trace propagation_context which will be used
 //  for all events and transactions/spans
+
 const char *trace_id = "2674eb52d5874b13b560236d6c79ce8a";
 const char *parent_span_id = "a0f9fdf04f1a63df";
 
 sentry_set_trace(trace_id, parent_span_id);
+perform_checkout(); // trace set before checkout
+...
+sentry_set_trace(trace_id_2, parent_span_id_2);
```
JoshuaMoelans: Should we 
- provide an example on how to generate these IDs (e.g. using `sentry_uuid_new_v4`) 
- maybe even provide a `sentry_generate_trace/span_id()`?
- OR provide a `sentry_generate_set_trace()` function that just handles the trace/span-id generation behind the scenes?

supervacuus: noooo :-)

supervacuus: `sentry_set_trace()` is not a function for users, that is very explicitly mentioned (at least in the header docs). That is why we initially didn't expose it in the user docs. Now, this PR fixes why we didn't add it in the first place. :-D

supervacuus: If users should have the ability to invalidate traces manually, then a corresponding function should be available. I currently see no reason for users to do this, but if there is, then it should be

* either the trace information from another SDK
* or entirely generated by us

An interface where users have to generate their own IDs doesn't make sense to me, even if we expose the corresponding functionality in the public header. We should be doing it for them because otherwise, we bind the client code to the contract between SDKs and the global API.

supervacuus: so, `sentry_[refresh|invalidate|renew|regenerate|...]_trace()` without params.

platform-includes/performance/add-spans-example/native.mdx
```diff
@@ -38,13 +38,16 @@ void perform_checkout() {
...
+## Trace Boundaries
+By default, transactions will inherit the `trace_id` and `span_id` from the `propagation_context` as generated during SDK initialization. These values can be overridden by a downstream SDK, or by defining a manual trace boundary using `sentry_generate_trace()`:
 
 ```c
-// Set the trace propagation_context which will be used
-//  for all events and transactions/spans
-const char *trace_id = "2674eb52d5874b13b560236d6c79ce8a";
-const char *parent_span_id = "a0f9fdf04f1a63df";
-
-sentry_set_trace(trace_id, parent_span_id);
```
supervacuus: We should not delete this, but the comment should clarify that the required arguments originate from another SDK (and that another SDK would typically call the function). It is now more valuable in the docs because it acts as a counterpoint to `sentry_regenerate_trace()`.

supervacuus (approved):
Would now keep `sentry_set_trace()` and s/sentry_generate_trace/sentry_regenerate_trace

platform-includes/performance/add-spans-example/native.mdx
```diff
@@ -38,13 +38,25 @@ void perform_checkout() {
 
 This example will send a transaction named `checkout` to Sentry. The transaction will contain a `validation` span that measures how long `validate_shopping_cart()` took and a `process` span that measures `process_shopping_cart()`. Finally, the call to `sentry_transaction_finish()` will finish the transaction and send it to Sentry.
 
-By default, transactions will inherit the `trace_id` and `parent_span_id` from the `propagation_context` as generated during SDK initialization. These values can be overridden by calling `sentry_set_trace()`, for example from a downstream SDK or to define a manual trace boundary:
+## Trace Boundaries
+By default, transactions/spans and events will inherit the `trace_id` from the `propagation_context` as generated during SDK initialization. These values can be overridden by a downstream SDK which calls `sentry_set_trace(trace_id, parent_span_id)`, or by defining a manual trace boundary using `sentry_generate_trace()`:
```
supervacuus:
```suggestion
## Trace Boundaries
By default, transactions, spans, and events will inherit the `trace_id` from the `propagation_context` as generated during SDK initialization. These values can be overridden by a downstream SDK that calls `sentry_set_trace(trace_id, parent_span_id)`, or by a user who can manually create a trace boundary using `sentry_generate_trace()`:
```
I understand you want to signal that transactions and spans are the same thing, but they kinda aren't (yet), and it is easier to swallow if we enumerate them as equals.

supervacuus (approved):
minor nit, but otherwise good to go.

PR #14228, opened 2025-07-02
fix: Correct loader script sentence

priscilawebdev:
**Problem**

The docs says:

image

But when you create a JavaScript project and check the settings, everything is actually turned off. We [recently changed the defaults so all options start disabled](https://github.com/getsentry/sentry/pull/86395). Before, when they were all checked by default, the page would send a request on load to update the script loader configuration — but now it doesn’t. Basically, everything seems to be False (disabled) by default, so this PR updates the docs to reflect that.

### DIFF
--- a/docs/platforms/javascript/common/install/loader.mdx
+++ b/docs/platforms/javascript/common/install/loader.mdx
@@ -50,7 +50,7 @@ To use the loader, go in the Sentry UI to **Settings > Projects > (select projec
 ></script>
 ```
 
-By default, Tracing and Session Replay are enabled.
+By default, Tracing and Session Replay are disabled.
 
 ## Source Maps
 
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/common/install/loader.mdx
```diff
@@ -50,7 +50,7 @@ To use the loader, go in the Sentry UI to **Settings > Projects > (select projec
 ></script>
 ```
 
-By default, Tracing and Session Replay are enabled.
+By default, Tracing and Session Replay are disabled.
```
priscilawebdev, line 53: this is the only change, I intended to introduce, but there is something updating the indentation 😬

ArthurKnaus (approved):

PR #14231, opened 2025-07-02
chore: Refactor onboarding options for unique titles

### LINKED ISSUE #14221: MDX view does not respect onboarding options
### Steps to Reproduce

e.g. https://docs.sentry.io/platforms/javascript/guides/react.md has several duplicate steps in the markdown. This is because the steps are displayed dynamically depending on which onboarding options are selected.

Not entirely sure how a good solution would look like, maybe we just should get rid of the steps in the MD?

### Expected Result

No duplicate steps

### Actual Result

Duplicate steps

### Additional Info

_No response_

BYK:
Fixes #14221.


BYK:
@sentry review

app/globals.css
```diff
@@ -165,3 +165,32 @@ body {
 {
   color: rgb(134, 142, 150) !important;
 }
+
+/* CSS Counters for Onboarding Steps */
+.onboarding-steps {
+  counter-reset: onboarding-step;
```
chargome, line 171: TIL

chargome (commented):
I don't think this works as expected: 

e.g. the steps in https://sentry-docs-git-cursor-refactor-onboarding-options-for-u-4ebbf6.sentry.dev/platforms/javascript/guides/react/ are not correctly numbered anymore

app/globals.css
BYK: Yeah, magic!

BYK:
@chargome we should be good now. Ideally this whole `<OnboardingOption>` component will be refactored, getting rid of `hiddenForThisOption` and replacing that with `platforms={['!performance']}` or something similar. I'm also guessing the `isStep` option can be removed and moved into a CSS class by itself or a `<OnboardingStep>` component. That said I want to time box my changes and I think we are in a good-enough state with this patch.

BYK:
>Another thing I just noticed is that the TOC on the right side does not include the steps anymore (e.g. compared to nextjs) but I don't think this is a blocker.

I actually noticed that too but decided that it is a fix rather than a bug. This also "fixes" the anchor links as they used to have the `step-1` prefix in them which was not stable (when you change onboarding options). Referring to them without the step numbers make more sense IMO.

If we want them back, we may look into a way to get the full text content of the header.

PR #14259, opened 2025-07-04
feat(kmp): add `sendDefaultPii` to init

buenaflor:
Closes https://github.com/getsentry/sentry-kotlin-multiplatform/issues/331


kahest (approved):
thanks for adding!

PR #14262, opened 2025-07-05
playstation: clarify only on saas

bruno-garcia:


docs/platforms/playstation/index.mdx
```diff
@@ -12,6 +12,11 @@ categories:
 
 You can also use Sentry's [Unity](/platforms/unity/game-consoles/) and [Unreal Engine SDKs](/platforms/unreal/game-consoles/) with PlayStation.
 
+<Alert>
+
+PlayStation support is exclusive to our SaaS offering, as it depends on confidential components that cannot be distributed for self-hosted use.
```
coolguyzone, line 17:
```suggestion
PlayStation support is exclusive to [sentry.io](https://sentry.io/) customers, as it depends on confidential components that cannot be distributed for self-hosted use.
```
This is pretty opinionated but it felt clearer to me, your call though!

coolguyzone (approved):
🫡

PR #14283, opened 2025-07-07
chore(derived_code_mappings): Remove call out

armenzg:
It's always been a source of confusion since it would not make any mention of the automatic code mapping derivation (only available for GitHub).

There are also some automatic commit fixes.

### DIFF
--- a/docs/organization/integrations/source-code-mgmt/github/index.mdx
+++ b/docs/organization/integrations/source-code-mgmt/github/index.mdx
@@ -12,7 +12,7 @@ There are two types of GitHub integrations, so make sure you're choosing the cor
   - `https://ghe.example.com/`
 
   For these accounts, follow the '[Installing GitHub Enterprise](#installing-github-enterprise)' instructions.
-  
+
 
 ### Check Your Domain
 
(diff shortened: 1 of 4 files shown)


docs/organization/integrations/source-code-mgmt/github/index.mdx
```diff
@@ -328,7 +328,7 @@ By default, this feature is automatically enabled once your GitHub integration h
 
 <Alert>
 
-Sentry will automatically try to set up code mappings on C#, Go, JavaScript, Node.js, PHP, Python, and Ruby projects for organizations with the GitHub integration installed. If your project uses a different SDK, you can add code mappings manually. Support for other languages and other source code integrations is being worked on.
+Sentry will automatically try to set up code mappings on certain platforms for organizations with the GitHub integration installed ([more details](https://github.com/getsentry/sentry/blob/master/src/sentry/issues/auto_source_code_config/constants.py)). If your project uses a different SDK, you can add code mappings manually.
```
mrduncan:
```suggestion
Sentry will automatically try to set up code mappings on C#, Clojure, Go, Groovy, Java, JavaScript, PHP, Python, Ruby, and Scala for organizations with the GitHub integration installed. If your project uses a different SDK, you can add code mappings manually.
```

I know it's going to require some more maintenance but I'd personally prefer that we explicitly list those platforms here to save users an extra click plus having to understand that file.

coolguyzone (approved):
LGTM!

PR #14301, opened 2025-07-09
fix(md): Remove md-links and AI rules from md output

BYK:


BYK:
Confirmed that https://sentry-docs-git-byk-fixmd-removals.sentry.dev/platforms/javascript/guides/react.md is now clean

PR #14303, opened 2025-07-09
chore(dynamic-sampling): add hint about conditions in dynamic sampling rules

shellmayr:
There is an implementation detail in dynamic sampling rules: a condition needs to be added, otherwise the ruleset is not considered valid. The documentation does not contain any reference to this, and it's frustrating to figure out the correct way of specifying a catch-all yourself, therefore I'm adding this here.

### DIFF
--- a/develop-docs/application-architecture/dynamic-sampling/architecture.mdx
+++ b/develop-docs/application-architecture/dynamic-sampling/architecture.mdx
@@ -52,6 +52,18 @@ An example of rule encoded in JSON is the following:
 }
 ```
 
+<Alert title="✨ Note">
+Dynamic sampling rules must always include a `condition` field, otherwise the entire dynamic sampling ruleset will be ignored by Relay. If you want a rule to match every event, set the condition as follows:
+```json
+{
+  "condition": {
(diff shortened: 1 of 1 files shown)


develop-docs/application-architecture/dynamic-sampling/architecture.mdx
```diff
@@ -52,6 +52,18 @@ An example of rule encoded in JSON is the following:
 }
 ```
 
+<Alert title="✨ Note">
```
constantinius, line 55: Do we use emojies like these in our docs?

constantinius (approved):

develop-docs/application-architecture/dynamic-sampling/architecture.mdx
shellmayr: Haha beautiful catch, i copied this from another page - seems kinda weird. I'll remove it in both places

shellmayr: Oh you know what, we actually do have this in many places, now that I searched for it

PR #14310, opened 2025-07-09
docs(feedback): Update docs for onSubmitSuccess to include eventID 

ryan953:
Also added a note for how to use `eventID` to link back to sentry.io

SCR-20250709-ntqc

Related to https://github.com/getsentry/sentry-javascript/pull/16835
Related to https://github.com/getsentry/sentry-javascript/issues/16836

PR #14325, opened 2025-07-10
ref(feedback): rm escape chars from last onSubmitSucces update

aliu39:
Fix https://github.com/getsentry/sentry-docs/pull/14310/files#r2198457651

This isn't a template string, just md


aliu39:
Closed in favor of ryans

PR #14354, opened 2025-07-14
chore(ai): prefetch the markdown text to get around ios security blocking async call on gesture

NicoHinderling:


codyde (approved):
Looks good!

PR #14375, opened 2025-07-15
allow specifying docs pages as full width if their content requires it

armcknight:
This is one option for fixing the table formatting issue in https://github.com/getsentry/sentry-docs/pull/14227


coolguyzone:
@armcknight Thanks for putting this together! Looks like it's breaking because of typescript linting issue:

Screenshot 2025-07-17 at 2 40 15 PM

chargome:
Do you want to handle the according CSS also in this pr?

armcknight:
@chargome I may need help, I'm not sure what would be required.

chargome (approved):

PR #14376, opened 2025-07-15
provide more space for docs page content

armcknight:
This is another option for fixing the table formatting issue in https://github.com/getsentry/sentry-docs/pull/14227

### DIFF
--- a/src/components/docPage/index.tsx
+++ b/src/components/docPage/index.tsx
@@ -77,7 +77,7 @@ export function DocPage({
               'prose-blockquote:font-normal prose-blockquote:border-l-[3px] prose-em:font-normal prose-blockquote:text-[var(--gray-12)]',
               'prose-img:my-2',
               'prose-strong:text-[var(--gray-12)]',
-              fullWidth ? 'max-w-none w-full' : 'w-[85ch] xl:max-w-[calc(100%-220px)]',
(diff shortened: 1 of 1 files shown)


src/components/docPage/index.tsx
```diff
@@ -77,7 +77,7 @@ export function DocPage({
               'prose-blockquote:font-normal prose-blockquote:border-l-[3px] prose-em:font-normal prose-blockquote:text-[var(--gray-12)]',
               'prose-img:my-2',
               'prose-strong:text-[var(--gray-12)]',
-              fullWidth ? 'max-w-none w-full' : 'w-[75ch] xl:max-w-[calc(100%-250px)]',
+              fullWidth ? 'max-w-none w-full' : 'w-[85ch] xl:max-w-[calc(100%-220px)]',
```
armcknight: There's overlap, but in a different situation now. Put this back for the sidebar width to see if it fixes it

```suggestion
              fullWidth ? 'max-w-none w-full' : 'w-[85ch] xl:max-w-[calc(100%-250px)]',
```

armcknight:
Unfortunately this still doesn't fix the issue when the window is sufficiently narrow enough 😕 I'm sure there's a way to do it but I'm not familiar enough with this codebase or frontend generally to surgically fix the problem.

PR #14388, opened 2025-07-16
chore(ai): Hide copy llm button on mobile

NicoHinderling:
image
image


NicoHinderling:
I did https://github.com/getsentry/sentry-docs/pull/14354 in order to make sure the button worked correctly on mobile but now we're hiding it on mobile, so technically we could get rid of the prefetch... but I'm just going to leave this as is as I don't think it's a big deal (cc @BYK in case you're curious why I added the prefetch code)

PR #14461, opened 2025-07-25
fix(compose-multiplatform): remove unnecessary header

buenaflor:
`# Compose Multiplatform` is unnecessary since the page already displays a header by default


coolguyzone (approved):
Good catch, thanks for cleaning this up.

PR #14494, opened 2025-07-30
Update index.mdx

InterstellarStella:
Clarification on how transaction-based profiles are billed. Transaction-based profiles will be automatically converted to either UI or Continuous profile hours after ingestion depending on the SDK used. This is also pointed out in our help desk article [here](https://sentry.zendesk.com/hc/en-us/articles/36061384317339-Why-do-I-suddenly-see-UI-Continuous-profiling-hours).


rodolfoBee (approved):
LGTM

PR #14498, opened 2025-07-30
feat(develop): Update Response Context

lcian:
## DESCRIBE YOUR PR
Document the Response Context for events, including all fields, according to the implementation in [Relay](https://github.com/getsentry/relay/blob/1bdc463896fb6f1425fac10915fa49799ecc5d85/relay-event-schema/src/protocol/contexts/response.rs#L11). Note that we don't include `inferred_content_type` because that's always overridden by Relay anyways.

### DIFF
--- a/develop-docs/sdk/data-model/event-payloads/contexts.mdx
+++ b/develop-docs/sdk/data-model/event-payloads/contexts.mdx
@@ -793,7 +793,7 @@ Page context contains information about the page that the event occurred on.
 
 ## Response Context
 
-Response context contains information about the HTTP response related to the event.
+Response context contains information about the HTTP response associated with the event.
 
 This is mostly set on transactions in a web server environment where one transaction represents a HTTP request/response cycle.
 
--- a/develop-docs/sdk/data-model/event-payloads/contexts.mdx
+++ b/develop-docs/sdk/data-model/event-payloads/contexts.mdx
@@ -801,13 +801,13 @@ This is mostly set on transactions in a web server environment where one transac
 
 : _Optional_. The cookie values. Can be passed, unparsed, as a string, as a dictionary, or as a list of tuples.
 
-- Example: `["PHPSESSID=12345; csrftoken=1234567"]`
+- Example: `{ "PHPSESSID": "12345", "csrftoken": "1234567" }`
 
 `headers`
 
 : _Optional_. A dictionary of submitted headers. If a header appears multiple times it, needs to be merged according to the HTTP standard for header merging. Sentry treats header names as case-insensitive.
 
-- Example: `[["Content-Type", "text/plain"]]`
+- Example: `{ "Content-Type": "text/plain" }`
 
 `status_code`
 
--- a/develop-docs/sdk/data-model/event-payloads/contexts.mdx
+++ b/develop-docs/sdk/data-model/event-payloads/contexts.mdx
@@ -805,7 +805,7 @@ This is mostly set on transactions in a web server environment where one transac
 
 `headers`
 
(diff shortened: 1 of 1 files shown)


develop-docs/sdk/data-model/event-payloads/contexts.mdx
```diff
@@ -795,23 +795,54 @@ Page context contains information about the page that the event occurred on.
 
 Response context contains information about the HTTP response associated with the event.
 
-The only and required field is `status_code`.
-
 This is mostly set on transactions in a web server environment where one transaction represents a HTTP request/response cycle.
 
+`cookies`
+
+: _Optional_. The cookie values. Can be given unparsed as string, as dictionary, or as a list of tuples.
```
coolguyzone:
```suggestion
: _Optional_. The cookie values. Can be passed, unparsed, as a string, as a dictionary, or as a list of tuples.
```

develop-docs/sdk/data-model/event-payloads/contexts.mdx
```diff
@@ -795,23 +795,54 @@ Page context contains information about the page that the event occurred on.
...
+
+- Example: `["PHPSESSID=12345; csrftoken=1234567"]`
+
+`headers`
+
+: _Optional_. A dictionary of submitted headers. If a header appears multiple times it, needs to be merged according to the HTTP standard for header merging. Header names are treated case-insensitively by Sentry.
```
coolguyzone:
```suggestion
: _Optional_. A dictionary of submitted headers. If a header appears multiple times it, needs to be merged according to the HTTP standard for header merging. Sentry treats header names as case-insensitive.
```

coolguyzone (approved):
Looks good, thanks!

develop-docs/sdk/data-model/event-payloads/contexts.mdx
```diff
@@ -795,23 +795,54 @@ Page context contains information about the page that the event occurred on.
...
+
+: _Optional_. The cookie values. Can be passed, unparsed, as a string, as a dictionary, or as a list of tuples.
+
+- Example: `["PHPSESSID=12345; csrftoken=1234567"]`
+
+`headers`
+
+: _Optional_. A dictionary of submitted headers. If a header appears multiple times it, needs to be merged according to the HTTP standard for header merging. Sentry treats header names as case-insensitive.
+
+- Example: `[["Content-Type", "text/plain"]]`
```
antonpirker: it says this is a dictionary but the example is a list.

develop-docs/sdk/data-model/event-payloads/contexts.mdx
```diff
@@ -795,23 +795,54 @@ Page context contains information about the page that the event occurred on.
...
 
+`body_size`
+
+: _Optional_. HTTP response body size in bytes.
+
+- Example: `1000`
+
+`data`
+
+: _Optional_. Response data in any format that makes sense. SDKs should discard large and binary bodies by default. Can be given as a string or structural data of any format.
```
antonpirker, line 826: can this really be everything? is EAP not limited to only basic types like str, int, float, bool, null and lists of those types?

antonpirker (changes requested):

develop-docs/sdk/data-model/event-payloads/contexts.mdx
lcian: In practice it can be anything that you can serialize as a valid JSON.
I'm not sure how EAP is involved when considering event contexts, but I can see we send arbitrary JSON in this field at minimum in the Java and Dotnet SDKs, and it gets ingested just fine.

PR #14503, opened 2025-07-31
Add example image to Sentry iOS user feedback page

bruno-garcia:
> [!NOTE]
> Adds alt-text-based image sizing via a new remark plugin, updates image components to prefer manual dimensions, and inserts an example widget image in iOS user feedback docs.
> 
> - **Images/rendering**:
>   - **Manual sizing precedence**: `src/components/docImage` now prefers explicit `width`/`height` props over URL hash dimensions and passes `isManualDimensions` to the lightbox.
>   - **Lightbox styling**: `src/components/imageLightbox` respects `isManualDimensions`, applying only provided dimensions (other axis auto) for inline images.
> - **MDX pipeline**:
>   - Add `remark-image-resize.js` and wire it in `src/mdx.ts` to parse size hints from image ALT text (e.g., `=300x200`) and set `width`/`height` attributes.
> - **Docs**:
>   - Add example User Feedback Widget image to `docs/platforms/apple/common/user-feedback/index.mdx`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3e5008494be1e7487570b248bcd6b8af773a8395. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


philipphofmann (approved):
LGTM, but the linter complaints.

sfanahata:
@bruno-garcia looks like this is causing a 404 issue. Check it out and try to push a commit through again?

src/mdx.ts
```diff
@@ -655,6 +655,11 @@ export async function getFileBySlug(slug: string): Promise<SlugFile> {
       // for this specific slug easily
       options.outdir = assetsCacheDir || outdir;
 
+      // Ensure file-loader URLs are absolute from the public folder so
+      // imported image paths do not end up page-relative (which breaks <img src={imported}>).
+      // With this, imports will resolve like "/mdx-images/<name>-<hash>.<ext>".
+      options.publicPath = '/mdx-images';
```
romtsn: @BYK @chargome I figured I can't specify image size and use `` so cursor came up with this - is it plausible? Without this, it doesn't work apparently.

romtsn: here's the preview btw, looks legit: https://sentry-docs-git-cursor-add-example-image-to-sentry-ios-u-3f9de6.sentry.dev/platforms/apple/user-feedback/#user-feedback-widget

BYK: Why can't we use the markdown way of adding images?

romtsn: we can, but we can't specify the image size that way? I don't want it to stretch for the full window width, it looks gigantic in that case

BYK: Maybe this works: https://stackoverflow.com/a/21242579/90297

If not I found a bunch of rehype plugins around this which we may want to look into if this is a common need.

romtsn: yep, tried the stackoverflow link, didn't work either. I'll look into the rehype plugins, thanks

romtsn: @BYK ok I ended up implementing our own remark-resize plugin as I couldn't find anything legit 😅 but it's just a few lines - could you PTAL? The usage is `<alt> =300x100` and it looks like this for this PR:

Google Chrome 2025-09-29 16 19 26

romtsn: @jaffrepaul adding you since you recently touched these files. Thank you!

src/components/docImage/index.tsx
```diff
@@ -99,19 +99,32 @@ export default function DocImage({
...
 
   return (
     <ImageLightbox
       src={finalSrc}
       imgPath={imgPath}
       width={width}
       height={height}
+      isManualDimensions={isManual}
       alt={props.alt ?? ''}
       {...props}
```
jaffrepaul: Cool, this should be mostly fine. There is one small potential issue that we can safeguard. It's more from the original implementation but it caught my eye with this additional move. The prop spreading should just come before the height and width attributes. There's a world in which the string value MDX props would override the parsed Numbers. If that gets down to the Next.js `<Image>`, it'd be a problem. But if we just do it earlier, the parse will win when we need it to.

romtsn:
```suggestion
      isManualDimensions={isManual}
      alt={props.alt ?? ''}
      {...props}
      width={width}
      height={height}
```
I take whatever you say for truth 😅  you mean like this^ right?

jaffrepaul: Ha! It is a nuanced and maybe unlikely scenario, but safety first🥇 Let me grab the whole component for completeness.

romtsn: addressed in https://github.com/getsentry/sentry-docs/pull/14503/commits/3e5008494be1e7487570b248bcd6b8af773a8395

src/components/docImage/index.tsx
```diff
@@ -99,21 +99,34 @@ export default function DocImage({
...
+  const width = isManual
+    ? manualWidth
+    : hashDimensions[0] > 0
+      ? hashDimensions[0]
+      : undefined;
+  const height = isManual
+    ? manualHeight
+    : hashDimensions[1] > 0
+      ? hashDimensions[1]
+      : undefined;
```
BYK: Not really a fan of these nested ternaries but cannot think of a better way rn.

src/components/docImage/index.tsx
```diff
@@ -99,21 +99,34 @@ export default function DocImage({
...
 
   return (
     <ImageLightbox
       src={finalSrc}
       imgPath={imgPath}
+      {...props}
```
BYK: Why put `{...props}` in the middle? Typically you either place it to the top, meaning "okay first the props, and anything below are the hard-overrides we have" _or_ "first, our defaults, and then potential overrides from props". Somewhere in between is very confusing to me.

src/components/imageLightbox/index.tsx
```diff
@@ -15,6 +15,7 @@ interface ImageLightboxProps
   imgPath: string;
   src: string;
   height?: number;
+  isManualDimensions?: boolean;
```
BYK:
```suggestion
  hasManualDimensions?: boolean;
```

or possibly better: `hasDimensionOverrides`, `hasDimensions`

src/components/imageLightbox/index.tsx
```diff
@@ -97,21 +99,31 @@ export function ImageLightbox({
...
+
+    // Apply sizing:
+    // - If manual: set only provided dimension(s); missing one becomes 'auto'
+    // - Else: default responsive
     const imageStyle = isInline
-      ? {width: '100%', height: 'auto', ...style}
+      ? isManualDimensions
+        ? {
+            width: width != null ? `${width}px` : 'auto',
+            height: height != null ? `${height}px` : 'auto',
```
BYK, line 111: nit: you may wanna allow different units for the future too.

src/remark-image-resize.js
```diff
@@ -0,0 +1,52 @@
...
+  return tree =>
+    visit(tree, {type: 'mdxJsxTextElement', name: 'img'}, node => {
+      // Handle MDX JSX <img> produced by remark-mdx-images
+      const altIndex = node.attributes.findIndex(a => a && a.name === 'alt');
+      const altValue =
+        altIndex !== -1 && typeof node.attributes[altIndex].value === 'string'
+          ? node.attributes[altIndex].value
+          : null;
+      if (altValue) {
+        const m = altValue.match(SIZE_FROM_ALT_RE);
```
BYK: What is `m`? I'd refrain from single-letter variable names unless they are well-established or very obvious (like `i`).

src/remark-image-resize.js
```diff
@@ -0,0 +1,52 @@
+import {visit} from 'unist-util-visit';
+
+const SIZE_FROM_ALT_RE = /\s*=\s*(\d+)?x?(\d+)?\s*$/;
```
BYK: This regex is not ideal as it would allow matching `=100200` as `100` and `200`. The middle `x?` part should depend on the next part being matched I think.

If you are not already using https://regex101.com/ might help you fine tune it.

src/remark-image-resize.js
```diff
@@ -0,0 +1,52 @@
...
+        if (m) {
+          const [, wStr, hStr] = m;
+          const cleanedAlt = altValue.replace(SIZE_FROM_ALT_RE, '').trim();
+          // set cleaned alt
+          node.attributes[altIndex] = {
+            type: 'mdxJsxAttribute',
+            name: 'alt',
+            value: cleanedAlt,
+          };
+          // remove any pre-existing width/height attributes to avoid duplicates
```
BYK: How does this case happen?

src/remark-image-resize.js
romtsn: right, that's cursor's doing, meant to change this to `match` 🙈 thanks

src/components/docImage/index.tsx
romtsn: I guess it was influenced by this comment: https://github.com/getsentry/sentry-docs/pull/14503#discussion_r2388894382

so did I get it right that I can place it to the top considering the linked comment?

src/components/docImage/index.tsx
romtsn: will add a separate variable for inferred height/width (`isManual ? manualHeight : inferredHeight`)

src/components/docImage/index.tsx
BYK: >so did I get it right that I can place it to the top considering the linked comment?

I guess? Then why not place it at the top? And if there's a reason we cannot put it at the top, we should add a comment explaining what the comment you linked above in the code.

src/components/imageLightbox/index.tsx
romtsn: good point! I will skip it for now though, can always add it later. Will leave a todo

src/remark-image-resize.js
romtsn: great catch! will fix that

src/remark-image-resize.js
romtsn: I guess if there are third-party plugins adding those attrs, but probably this is unnecessary

PR #14554, opened 2025-08-05
docs(js): Add troubleshooting for client instrumentation hook warning

chargome:
ref https://github.com/getsentry/sentry-javascript/issues/16093


Lms24 (approved):
Nice follow-up, thanks!

PR #14578, opened 2025-08-07
ref(rust): Adjust getting started docs for tracing

lcian:
Adds getting started docs for the `tracing` integration.
This is our most popular integration, so we want to have a dedicated page to give it even more visibility.
Currently the only docs we have for this are [here](https://docs.sentry.io/platforms/rust/tracing/instrumentation/automatic-instrumentation/#tracing-integration) which means they're hard to find, and they only document the tracing functionality.

### DIFF
--- a/docs/platforms/rust/guides/tracing/index.mdx
+++ b/docs/platforms/rust/guides/tracing/index.mdx
@@ -90,9 +90,7 @@ async fn fail() {
 
 By default, error level events from `tokio-rs/tracing` are captured as Sentry events, while anything at or above info is added as a breadcrumb.
 
-{/* ___PRODUCT_OPTION_START___ logs */}
-
-To capture structured logs from `tokio-rs/tracing` events instead, you need to set up the Sentry layer with a custom event filter that maps to logs, like so:
+To capture structured logs from `tokio-rs/tracing` events instead, you need to pass `enable_logs: true` to `sentry::init`, and set up the Sentry layer with a custom event filter that maps to logs, like so:
 
 ```rust
 use sentry::integrations::tracing::EventFilter;
@@ -107,8 +105,6 @@ sentry::integrations::tracing::layer().event_filter(|md| match *md.level() {
 });
 ```
 
-{/* ___PRODUCT_OPTION_END___ logs */}
-
 ## Verify
 
 <Alert>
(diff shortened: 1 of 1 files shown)


docs/platforms/rust/guides/tracing/index.mdx
```diff
@@ -0,0 +1,120 @@
+---
+title: tracing
```
coolguyzone: I'm a little concerned that this could cause confusion since we also have our own 'Tracing' product. I think it's possible search might send people looking for /platforms/rust/tracing/ to this page or vice-versa. Would it make sense to update the title to something like 'tracing by tokio'? Or maybe add a little notification explaining that this is not about Sentry Tracing?

coolguyzone (approved):
See my note about the title, but overall looks good to me!

docs/platforms/rust/guides/tracing/index.mdx
```diff
@@ -0,0 +1,120 @@
...
+async fn fail() {
+    tracing::debug!("Doing work"); // Adds a breadrcumb
+    # ___PRODUCT_OPTION_START___ performance
+    let _span = tracing::info_span("Child span").entered(); // Captures a child span
+    # ___PRODUCT_OPTION_END___ performance
+    tracing::error!("Everything is on fire!");
+}
+```
+
+By default, error level events are captured as Sentry events, while anything at or above info is added as a breadcrumb.
```
alexander-alderman-webb: I think we should be explicit when talking about Tokio and Sentry events, even if it's arguably clear from context.

```suggestion
By default, error level events from Tokio are captured as Sentry events, while anything at or above info is added as a breadcrumb.
```

docs/platforms/rust/guides/tracing/index.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+{/* ___PRODUCT_OPTION_START___ logs */}
+
+To capture structured logs for `tracing` events instead, you need to set up the Sentry layer with a custom event filter that maps to logs, like so:
```
alexander-alderman-webb:
```suggestion
To capture structured logs for Tokio `tracing` events instead, you need to set up the Sentry layer with a custom event filter that maps to logs, like so:
```

lcian:
I mistakenly merged this as part of another commit.
Will update this one by addressing the feedback and then merge it.

lcian:
> Do we want to include "Tokio" in the name of the guide? From what I understand the tracing crate and tokio:tracing aren't the same thing. Not sure if all rust devs know the difference. I'm a bit confused on the differences and what they have in common.

For the record they're the same thing. The only confusion that could arise is from our own `tracing` product vs `tokio-rs/tracing`, therefore now I'm making it explicit what this is about.

lcian:
> /platforms/rust/tracing/ links to /platforms/rust/tracing/instrumentation/automatic-instrumentation/ here:
> ...
> Do we also want to link to this new guide?

I think we can remove that link now as docs for the tracing integration are now easily findable.

PR #14580, opened 2025-08-07
feat(agents-insights): update docs

obostjancic:
- Updates screenshots with hte new layout
- Removes beta label
- Adds descriptions for new widgets

Closes [TET-993: Update AI Agents docs](https://linear.app/getsentry/issue/TET-993/update-ai-agents-docs)


docs/product/insights/ai/agents/img/trace-table.png
shellmayr, line 1: The trace table screenshot does not include the selector for Traces / Models / Tools anymore, but the text continues to reference it - may be confusing if it's missing in the screenshot?

shellmayr: Ok nevermind, I just realized that segment moved 🤦‍♂️

obostjancic: We moved them up, so they are above one line of widgets, and the text was changed to also include widgets

PR #14627, opened 2025-08-12
fix: Fix broken AI banner due to invalid syntax

BYK:


andreiborza (approved):
Thanks for fixing this!

BYK:
Hat tip to @codyde who brought it to my attention!

PR #14672, opened 2025-08-15
docs(hydration errors): Add a FAQ entry about strange html diffs

ryan953:
The original ticket was something like "Sentry raises attribute order or empty spaces in values as Hydration errors that have no effect on the HTML rendering"

I wrote a response on the ticket, but since this is a common enough thing I think it should be in the FAQ too. It's a good question that we've gotten before.

Fixes https://github.com/getsentry/sentry/issues/97828

SCR-20250815-jnqv


docs/product/issues/issue-details/replay-issues/hydration-error.mdx
```diff
@@ -63,6 +63,22 @@ To make debugging easier, adjust the diff boundary to show more changes before o
...
+
+Sometimes you will see changes to the HTML like attribute being removed, order changing, spaces being removed, or strange classes like `:hover` appearing. These parts of the diff do not impact the HTML that the user sees, and often are not the cause of the hydration error.
+
+This is expected and a side effect of how we capture debugging data after a Hyration Error has occurred.
+
+Some common examples:
+
+![html diff tool](./img/hydration-diff-hover.png)
+
+An attribute `xmlns="http://www.w3.org/2000/svg"` was removed. This is an XML attribute and not needed within a valid HTML page, so react did not include it in the render.
```
michellewzhang:
```suggestion
An attribute `xmlns="http://www.w3.org/2000/svg"` was removed. This is an XML attribute and is not needed within a valid HTML page, so React did not include it in the render.
```

docs/product/issues/issue-details/replay-issues/hydration-error.mdx
```diff
@@ -63,6 +63,22 @@ To make debugging easier, adjust the diff boundary to show more changes before o
...
+
+![html diff tool](./img/hydration-diff-xmlns.png)
+
+The class name `:hover` was added Session Replay when it serialized the page, React didn't see this class and it has no impact on hydration. This means the user was hovering and interacting with the page before react was finished hydrating (this is the whole point of hydration; to make the page react for interaction sooner!). By the time hydration was complete the user had moved their mouse so the :hover class is changed.
```
michellewzhang, line 80:
```suggestion
The class name `:hover` was added by the Replay SDK when it serialized the page. React didn't see this class and it has no impact on hydration. This means the user was hovering and interacting with the page before React was finished hydrating (this is the whole point of hydration; to make the page react for interaction sooner!). By the time hydration was complete, the user had moved their mouse, so the `:hover` class is changed.
```
to? by?

ryan953: i think `by`, in the sentence it's doing the thing.
added to the serialized json.. like 'added to the [serialzied] replay [data]'

ryan953: could do 'by the replay sdk`

PR #14710, opened 2025-08-25
docs(replay): Bring replay nav item to all apple platforms

ryan953:
This puts all the Session Replay pages into the Apple platform section, not just iOS. 

`/platforms/apple/` Nav
| Before | After |
| --- | --- |
SCR-20250825-mjyo | SCR-20250825-mkbs

But since most of the Apple platforms are not specifically supported I put the two "Does Session Replay work with *" faq questions at the top of the index page whenever you're not looking specifically at iOS docs. 
Only the main Apple and iOS show the sub-pages in the nav bar, the other platform specific sections only show the FAQ stuff and no sub-pages.

| Apple Overall | iOS Specific | The others |
| --- | --- | --- |
| SCR-20250825-mird | SCR-20250825-misb | SCR-20250825-mlcq

### DIFF
--- a/docs/platforms/apple/common/session-replay/index.mdx
+++ b/docs/platforms/apple/common/session-replay/index.mdx
@@ -2,9 +2,27 @@
 title: Set Up Session Replay
 sidebar_title: Session Replay
 sidebar_order: 5500
-notSupported:
 description: "Learn how to enable Session Replay in your mobile app."
+notSupported:
 ---
+
+<PlatformSection notSupported={["apple.ios"]}>
+
+## FAQ
+
+Q: **Does Session Replay work with SwiftUI?**
+
+A: Yes. It works with both UIKit and SwiftUI.
+
+Q: **Does Session Replay work on macOS, watchOS, tvOS and visionOS?**
+
+A: We don't actively prevent you from using Session Replay on these platforms, but we only officially support it on iOS. As a consequence, we can't make guarantees about its functionality and performance on other platforms.
+
+# Session Replay for Mobile Apps
+
+</PlatformSection>
+
+<PlatformSection notSupported={["apple.macos", "apple.tvos", "apple.watchos", "apple.visionos"]}>
 
 <Alert level="warning">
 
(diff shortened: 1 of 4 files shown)


jas-kas:
_Sharing my feedback from Slack_

This is what I'd change:

- Add an Apple > Session Replay page (reasoning: since iOS is under Apple, this is fair game)
-- I would not lead with a FAQ on the homepage here, but instead simply add a banner or bolded phrase somewhere that this is for the iOS SDK only. Alternatively, link out to iOS docs for more information.

- I would not add a Session Replay page for macOS, tvOS, watchOS, visionOS simply because they aren't supported platforms. Why have a mostly empty page saying that? I think we should just exclude a Session Replay left-nav item altogether for these platforms imo

I think we can continue to massaging the Troubleshooting and mini-FAQ content in a follow up

docs/platforms/apple/common/session-replay/index.mdx
```diff
@@ -2,10 +2,28 @@
...
+
+## FAQ
+
+Q: **Does Session Replay work with SwiftUI?**
+
+A: Yes. It works with both UIKit and SwiftUI.
+
+Q: **Does Session Replay work on macOS, watchOS, tvOS and visionOS?**
+
+A: We don't actively prevent you from using Session Replay on these platforms, but we only officially [support it on iOS](/platforms/apple/guides/ios/session-replay/). As a consequence, we can't make guarantees about its functionality and performance on other platforms.
```
michellewzhang:
```suggestion
A: We don't actively prevent you from using Session Replay on these platforms, but we only officially [support it on iOS](/platforms/apple/guides/ios/session-replay/). As a result, we can't make guarantees about its functionality and performance on other platforms.
```

docs/platforms/apple/common/session-replay/troubleshooting.mdx
```diff
@@ -1,10 +1,32 @@
...
+notSupported:
+  - apple.macos
+  - apple.watchos
+  - apple.tvos
+  - apple.visionos
 ---
 
+<Expandable title="Why are parts of my replay not masked?" permalink>
+
+Text views, input views, images, video players and webviews are all masked by default. Images with bundled assets aren't masked because the likelihood of these assets containing PII is low. If you encounter a view that should be masked by default, consider opening a [GitHub issue](https://github.com/getsentry/sentry-cocoa/issues).
```
michellewzhang, line 34:
```suggestion
Text views, input views, images, video players, and webviews are all masked by default. Images with bundled assets aren't masked because the likelihood of these assets containing PII is low. If you encounter a view that should be masked by default, consider opening a [GitHub issue](https://github.com/getsentry/sentry-cocoa/issues).
```

docs/platforms/apple/common/session-replay/troubleshooting.mdx
```diff
@@ -1,10 +1,32 @@
...
+
+</Expandable>
+
+<Expandable title="What's the lowest version of iOS supported?" permalink>
+
+Session Replay recording happens even on the lowest version supported by the Sentry SDK, which is aligend with appstore support.
```
michellewzhang, line 40:
```suggestion
Session Replay recording happens even on the lowest version supported by the Sentry SDK, which is aligned with app store support.
```

PR #14713, opened 2025-08-25
fix: Fix breadcrumb + markdown button allignment

cleptric:
**Before**

Firefox 2025-08-26 01 48 44

**After**

Screenshot 2025-08-26 at 01 48 10

### DIFF
--- a/src/components/breadcrumbs/index.tsx
+++ b/src/components/breadcrumbs/index.tsx
@@ -24,7 +24,7 @@ export function Breadcrumbs({leafNode}: BreadcrumbsProps) {
   }
 
   return (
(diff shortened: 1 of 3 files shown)


src/components/breadcrumbs/index.tsx
```diff
@@ -24,7 +24,7 @@ export function Breadcrumbs({leafNode}: BreadcrumbsProps) {
   }
 
   return (
-    <ul className="list-none flex p-0 flex-wrap float-left" style={{margin: 0}}>
+    <ul className="not-prose list-none flex flex-wrap" style={{margin: 0}}>
```
cleptric, line 27: Removing prose here to get rid of the bottom margin.

codyde:
Out here doing Gods work. Merging!

PR #14756, opened 2025-08-28
docs(aws-integration): Remove alert that ESM is not supported

msonnb:
The AWS integration supports both ESM and CJS now


andreiborza (approved):
Good stuff!

PR #14759, opened 2025-08-28
feat: Update logs spec to mention message template constraint

AbhiPrasad:


develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -299,6 +299,8 @@ If the log was paramaterized the SDK should attach the following
 }
 ```
 
+If there are no `sentry.message.parameter.X` attributes included in the log, then the SDK MUST NOT attach a `sentry.message.template` attribute.
```
k-fish: Can you add that this is important for PII reasons, since lacking `parameter` means that template isn't a template but a clone of body?

AbhiPrasad:
```suggestion
If there are no `sentry.message.parameter.X` attributes included in the log, then the SDK MUST NOT attach a `sentry.message.template` attribute. This is important because it reduces duplicate content in the log and makes PII processing in Sentry much easier as only we don't have to duplicate processing between the log message and the `sentry.message.template` attribute.
```

PR #14776, opened 2025-09-01
docs: Introduce self hosted warning alert to Xbox and Nintendo Switch

priscilawebdev:


docs/platforms/nintendo-switch/index.mdx
```diff
@@ -12,6 +12,11 @@ categories:
 
 You can also use Sentry's [Unity](/platforms/unity/game-consoles/) and [Unreal Engine SDKs](/platforms/unreal/game-consoles/) with Nintendo Switch.
 
+<Alert>
+
+Nintendo Switch support is exclusive to our SaaS offering, as it depends on confidential components that cannot be distributed for self-hosted use.
+
+</Alert>
```
shellmayr, line 19: Is there an easy way we can we make this show only on self-hosted? I think it may confuse users on SaaS

bruno-garcia: Nintendo works on self hosted.

docs/platforms/xbox/index.mdx
```diff
@@ -12,6 +12,11 @@ categories:
 
 You can also use Sentry's [Unity](/platforms/unity/game-consoles/) and [Unreal Engine SDKs](/platforms/unreal/game-consoles/) with Xbox.
 
+<Alert>
+
+Xbox support is exclusive to our SaaS offering, as it depends on confidential components that cannot be distributed for self-hosted use.
```
bruno-garcia, line 17: Xbox also works on self hosted. Support is possible via the open source sentry native even though the bug tracker and CI is on a private repo

docs/platforms/nintendo-switch/index.mdx
priscilawebdev: > Is there an easy way we can we make this show only on self-hosted? I think it may confuse users on SaaS

Hm… @shellmayr could it be that you’re thinking of Sentry SaaS? As far as I know, we typically don’t check for self-hosted in our docs, do we?

priscilawebdev:
closing this as we need to re-think these alerts for Xbox and Nintendo-Switch

PR #14777, opened 2025-09-01
fix(platform): Adjust layout for wide screens in dev docs

chargome:
Adjusts css for devdocs on wide screens, also fixes the layout for dev-docs sidebar


stephanie-anderson (approved):
Thanks, @chargome 🙏

PR #14795, opened 2025-09-03
feat(js): Add log options

mydea:
Adds the log options for JS.

In a follow up, we should also add the `logger` to the APIs page.

### DIFF
--- a/docs/platforms/javascript/common/configuration/options.mdx
+++ b/docs/platforms/javascript/common/configuration/options.mdx
@@ -436,7 +436,7 @@ Please note that the `span` you receive as an argument is a serialized object, n
 
 A list of strings or regex patterns matching span names that shouldn't be sent to Sentry. When using strings, partial matches will be filtered out, so if you need to filter by exact match, use regex patterns instead. You can also provide an object with a `name` and `op` property to match both the span name and the operation name. Note that at least `name` or `op` must be provided.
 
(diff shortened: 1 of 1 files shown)


AbhiPrasad (approved):

docs/platforms/javascript/common/configuration/options.mdx
```diff
@@ -437,12 +437,26 @@ Please note that the `span` you receive as an argument is a serialized object, n
...
 
 By default, no spans are ignored.
 
 </SdkOption>
 
+## Logs Options
+
+<SdkOption name="enableLogs" type='boolean' defaultValue='false'>
+
+Set this option to `true` to enable log capturing in Sentry. Only when this is enabled will the `logger` APIs actually send logs to Sentry.
```
coolguyzone, line 470:
```suggestion
Set this option to `true` to capture logs in Sentry. The `logger` APIs will only send logs to Sentry when this option is enabled.
```

mydea: the reason I worded this slightly different is that by just setting this, you are not actually capturing any logs yet - you'll either need to manually call the log APIs or add the respective logging integrations 🤔 IMHO `"to capture logs" sounds as if when I flip this to true I will immediately get logs. But maybe I am overthinking this xD

coolguyzone: Ah, thanks for the context @mydea! That makes sense.

AbhiPrasad (approved):

PR #14807, opened 2025-09-04
docs(fix): remove extra angle bracket from docs link

jas-kas:
Removing extra angle bracket from docs link inside Searchable Properties page

### DIFF
--- a/docs/concepts/search/searchable-properties/session-replay.mdx
+++ b/docs/concepts/search/searchable-properties/session-replay.mdx
@@ -102,7 +102,7 @@ The `title` of an element that was clicked. For example, `Save this comment` wou
 
 ### `count_dead_clicks`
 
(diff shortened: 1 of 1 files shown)


docs/concepts/search/searchable-properties/session-replay.mdx
sergical, line 1: the link for `dead clicks` in this file also has wrong copy to highlight (probably updated in the past), it's now `/product/explore/session-replay/replay-page-and-filters/#:~:text=Dead%20Clicks%3A%20User%20clicks%20on%20a%20and%20button%20tags%20that%20do%20not%20result%20in%20any%20page%20activity%20after%207%20seconds%20(i.e.%20no%20HTML%20was%20added%2C%20removed%2C%20or%20updated%3B%20no%20visual%20changes%20were%20observed%20in%20the%20page).`

jas-kas: erm, can you clarify where?

the `dead clicks` link under `dead.selector` and `count_dead_clicks` highlight the right text when opening up the link :o

sergical: these highlights dont really work in NextJS on redirects, accessing the URL directly works sometimes, but in Safari it strips them out completely on navigation https://github.com/vercel/next.js/discussions/71970

sergical (approved):
:shipit:

PR #14910, opened 2025-09-12
fix(docs): fix typo on AI in Sentry page

jas-kas:
Removing asterisks below

Screenshot 2025-09-12 at 11 27 59 AM


jas-kas:
@sergical I thought about that too, but I couldn't find a spot that made sense to be bold, so I just dropped the stars 😛

PR #14927, opened 2025-09-14
feat: Github Action to linkify 404 links

a-hariti:
**Motivation**

make the links in user 404 reports clickable for easy verifiability

Not sure the ideal way to test this action aside from merging into master then observing its behavior


chargome (approved):

chargome:
Related: seems like the 404 linter has failed to detect some of the 404s lately, maybe we need to re-visit

sergical:
@a-hariti let me know if you can update this based on the bugbot feedback. one thing that would actually be really helpful is having the ticket report the previous page from where the 404 was accessed, not sure if we can do location.history or something and pass it off to the github issue

sergical (approved):

PR #14931, opened 2025-09-15
feat: Add thread pool info context to Contexts

priscilawebdev:
Contributes to https://linear.app/getsentry/issue/TET-1148/add-threadpoolinfo-context-to-docs

### DIFF
--- a/develop-docs/sdk/data-model/event-payloads/contexts.mdx
+++ b/develop-docs/sdk/data-model/event-payloads/contexts.mdx
@@ -622,6 +622,54 @@ The following example illustrates the contexts part of the <Link to="/sdk/data-m
 }
 ```
 
+## Thread Pool Info Context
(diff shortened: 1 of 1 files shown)


vgrozdanic:
I think this would be good to run by SDK folks, as those docs are mostly used by them when they implement things in SDKs. 

@cleptric might have some opinions here, and maybe others too

cleptric:
Is this typed in Relay?

priscilawebdev:
> Is this typed in Relay?

I couldn't find any mentions in the relay code

cleptric:
Ok, as a followup for all three PRs, it would be great to type this in Relay as well. Feel free to rope in SDK folks to help out.

coolguyzone (approved):
Thanks for updating, looks good

PR #14935, opened 2025-09-15
Remove outdated docs

noahsmartin:
## DESCRIBE YOUR PR
If users do this it would actually cause the build to fail. It hasn't been required for a while but only recently started making builds fail.

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST


alwx (approved):

philippsuess:
@noahsmartin, wouldn't it be better to keep this workaround and mention the versions it works for? Some users might still need this if they can't update to a more recent version.

noahsmartin:
@philipphofmann I considered that, but thought that would ultimately lead to more confusion because:

- this only applies to to the first time you set up the sdk, anyone setting up the sdk for the first time now would not be using a version where this mattered 
- Even if someone did set up an old version for the first time, the default compiler settings with cocoapods take care of it. A user would only have to change this if they are also overriding the default compiler settings
- This recommendation isn't just unnecessary now, it’s wrong. If a user does this accidentally they will break their build.

so this recommendation seems applicable to a very unlikely case, and it would be better for users if the docs contained the most relevant information, thats why I removed it. Given that what do you think? If you’d prefer to keep it I’m happy to

philipphofmann (approved):
Thanks for the explanation @noahsmartin. Ship it 🚀

PR #14947, opened 2025-09-16
Document catchall variant for protocol enums

jjbayer:
See for example https://github.com/getsentry/relay/pull/5144.

### DIFF
--- a/develop-docs/ingestion/relay/relay-best-practices.mdx
+++ b/develop-docs/ingestion/relay/relay-best-practices.mdx
@@ -16,6 +16,18 @@ Make sure changes to the event protocol and APIs are forward-compatible. Relay s
 or truncate data it does not understand. It is a supported use-case to have customers running
 outdated Relays but up-to-date SDKs.
 
+For example, `enum`s in the event protocol (or any other data type received from the SDK) need to have a catchall variant
+such that customer relays running an older version can still forward the data:
+
+```rust
+enum EventColor {
+    Red,
+    Green,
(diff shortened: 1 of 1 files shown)


develop-docs/ingestion/relay/relay-best-practices.mdx
```diff
@@ -16,6 +16,18 @@ Make sure changes to the event protocol and APIs are forward-compatible. Relay s
...
+For example, `enum`s in the event protocol (or any other data type received from the SDK) need to have a catchall variant
+such that customer relays running an older version can still forward the data:
+
+```rust
+enum EventColor {
+    Red,
+    Green,
+    Blue,
+    /// Unknown color, for forward compatibility.
+    Unknown(String),
```
Dav1dde, line 28: Sometimes we use `#[serde(other)] Unknown`, sometimes `Unknown(String)` and other times our derives have something builtin `#[metastructure(fallback_variant)]`.

Seems to be quite use-case specific and hard to phrase :/.

Dav1dde (approved):

coolguyzone (approved):

develop-docs/ingestion/relay/relay-best-practices.mdx
jjbayer: Yes I looked at `fallback_variant` and it seems we only use it with `Object`. I guess the important part is that we have some sort of fallback such that parsing does not fail and the data is forwarded.

PR #14955, opened 2025-09-17
fix(relay): Fix Relay docker commands, registry, nonroot

Dav1dde:
## DESCRIBE YOUR PR

- The permission change no longer works, due to Relay having a different base docker container. I opted to use busybox to make the permission change (with the new user/group id) to make it fully copy pastable, the alternative would be a `sudo chown` from the host.
- Fixed the `$(pwd)` quoting, this would fail on paths with spaces.
- Switched the registry to ghcr.io as the dockerhub image is seemingly deprecated
image

## PRE-MERGE CHECKLIST


docs/product/relay/getting-started.mdx
```diff
@@ -32,16 +32,15 @@ container and copying in the files.
...
 # See https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label.
-docker run --rm -it                  \
-  -v $(pwd)/config/:/work/.relay/:z  \
-  --entrypoint bash                  \
-  getsentry/relay                    \
-  -c 'chown -R relay:relay /work/.relay'
+docker run --rm -it                   \
+  -v "$(pwd)"/config/:/work/.relay/:z \
+  busybox                             \
+  sh -c "chown -R 65532:65532 /work"  \
```
jjbayer: I wonder, if this command can run within a single-use container, can't it simply run on the host machine instead? Or would that make to many assumptions about the user's host system?

Dav1dde: It can run on the host, I touched on this in the PR description.

> The permission change no longer works, due to Relay having a different base docker container. I opted to use busybox to make the permission change (with the new user/group id) to make it fully copy pastable, the alternative would be a sudo chown from the host.

I wasn't sure which one to use and opted to stay with the docker command. But I am open to changing this, it did feel a bit wrong to run a `chown` via docker...

jjbayer: The tutorial seems to implicitly assume a unix system so I would personally run it on the host. But no strong opinion and definitely not a blocker.

PR #14957, opened 2025-09-17
ref(develop/sdks): Clarify Exception Mechanism interface

Lms24:
This PR clarifies the exception `Mechanism` interface, specifically the purpose and naming conventions for `mechanism.type`:

- it now explicitly states that `type` should be used to indicate the "origin" or callsite of the exception, so that we can distinguish where an exception was captured. For instance, user vs. SDK but in case of the latter, type should be specific enough to know which instrumentation/integration within the SDK.
- it suggests to use [Trace Origin](https://develop.sentry.dev/sdk/telemetry/traces/trace-origin/)-esque naming for SDK-captured events with the main motivation of this being an already established naming scheme that works well enough for this use case. 

To be clear, SDKs don't have to change anything right now, but should they want to ([like we did in JS](https://github.com/getsentry/sentry-javascript/issues/17212)), this reworked doc should provide more clarity than the short descriptions previously.

(I have another PR coming up with some refactors changes all over this doc but wanted to get this reviewed and accepted first since it actually changes content)  

closes https://github.com/getsentry/sentry-javascript/issues/17266
closes https://github.com/getsentry/sentry-javascript/issues/17212 (finally)

### DIFF
--- a/develop-docs/sdk/data-model/event-payloads/exception.mdx
+++ b/develop-docs/sdk/data-model/event-payloads/exception.mdx
@@ -67,7 +67,7 @@ The identifier of the capturing mechanism that captured the exception.
 The chosen mechanism `type` MUST help users as well as Sentry employees determine the responsible mechanism for capturing the exception. 
 This is either instrumentation within our SDKs or users manually capturing exceptions (for example via `captureException`). 
 
-The `type` MUST be _reasonably unique_ so that identifying e.g. the integration or SDK API responsible for capturing is possible. 
+The `type` MUST be _reasonably unique_  to make it possible to identify the integration or SDK API that performed the capture.
 There's no strict uniqueness requirement as in certain situations, multiple paths exist within one instrumentation, in which case it is fine to use a common mechanism `type`.
 
 For user-invoked exception captures (e.g. via `captureException`), the `type` MUST be set to `'generic'`.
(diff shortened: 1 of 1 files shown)


cleptric (approved):

develop-docs/sdk/data-model/event-payloads/exception.mdx
```diff
@@ -51,89 +51,156 @@ created this exception.
...
-`type`
+#### `type`
+
+Required (`string`)
+
+The identifier of the capturing mechanism that captured the exception. 
+The chosen mechanism `type` MUST help users as well as Sentry employees determine the responsible mechanism for capturing the exception. 
+This is either instrumentation within our SDKs or users manually capturing exceptions (for example via `captureException`). 
+
+The `type` MUST be _reasonably unique_ so that identifying e.g. the integration or SDK API responsible for capturing is possible.
```
coolguyzone:
```suggestion
The `type` MUST be _reasonably unique_  to make it possible to identify the integration or SDK API that performed the capture.
```

develop-docs/sdk/data-model/event-payloads/exception.mdx
```diff
@@ -51,89 +51,156 @@ created this exception.
...
+Fully qualified URL to an online help resource, possibly interpolated with error parameters.
 
-: An optional string value describing the source of the exception. The SDK should populate this with the name of the property or attribute of the parent exception that this exception was acquired from. In the case of an array, it should include the zero-based array index as well.
+#### `synthetic`
+
+Optional (`boolean`)
+
+Flag indicating that this error is synthetic. 
+Synthetic errors are errors that carry little meaning by themselves. 
+This may be because they are created at a central place (like a crash handler), and are all called the same: `Error`, `Segfault` etc.
```
coolguyzone:
```suggestion
This may be because they are all created at a central place (like a crash handler), and share the same title: `Error`, `Segfault` etc.
```

coolguyzone (approved):
Thanks for updating, LGTM.

Issue #13749, opened 2025-05-19
Mermaid: pan & zoom


chargome:
### Problem Statement

Mermaid html diagrams [don't come with a zoom feature](https://github.com/mermaid-js/mermaid/issues/2162) ootb.

e.g. https://github.com/getsentry/sentry-docs/pull/13745 introduces a large mermaid diagram that would benefit from such a feature.

### Solution Brainstorm

Explore solutions in the PR linked above.

chargome:
@a-hariti fyi

yeminxuan:
How is the progress of this issue now?

PR #14967, opened 2025-09-18
feat(Mermaid) add pan and zoom via svg-pan-zoom

sergical:
Closes https://github.com/getsentry/sentry-docs/issues/13749

Example: https://develop-docs-7b5gl6fix.sentry.dev/self-hosted/data-flow/

| Before | After |
|--------|--------|
| CleanShot 2025-09-17 at 16 59 32@2x | mermaid-pan-zoom |


sergical:
> Nice! The only downside I see with this approach is that we have different zoom states for light/dark but I thinks thats not a big issue. Thanks for taking care of this!

Some say bug, others say feature 😅

chargome:
> Some say bug, others say feature 😅

PR #14996, opened 2025-09-19
docs(fix): Update Issues > Feature Flag docs To Clarify Feature Flag Change Tracking

jas-kas:
Update description on feature flag change tracking to make it clear that evaluation tracking needs to be enabled in order to see feature flag change annotations

### DIFF
--- a/docs/product/issues/issue-details/feature-flags/index.mdx
+++ b/docs/product/issues/issue-details/feature-flags/index.mdx
@@ -48,7 +48,7 @@ Change tracking enables Sentry to listen for additions, removals, and modificati
 
 ![Flag Change Tracking](./img/FlagChangeTracking.png)
 
-Within an issue, the audit log is represented in the "event" chart and presents itself as a "release" line. The lines will only appear for flags that were evaluated before the error event occurred; however, feature flag definition changes (audit log) can still occur after an error. This feature requires evaluation tracking to be enabled. 
+Within an issue, the audit log is represented in the "event" chart and presents itself as a "release" line. The lines will only appear for flags that were evaluated before the error event occurred; however, feature flag definition changes from the audit log can still occur after an error. This feature requires evaluation tracking to be enabled. 
 
 ![Flag Change Issue](./img/FlagChangeIssue.png)
 


michellewzhang:
> @jas-kas i think flag events can also be shown on the chart if they are after the event as well. as long as it's associated with the audit log (change tracking), the flag has the possibility of showing up
> 
> SCR-20250919-ktuw

cc @cmanallen

jas-kas:
@michellewzhang What I meant was the feature flag changes that we end up showing on the event volume chart are ones that were part of the evaluation tracking list to begin with (i.e., flags evaluated before an error event in the issue that are shown in the table view on an Issue Details and also part of the feature flag drawer). We did this as a way to filter the FF change annotations to be less noisy and show more relevant flags

michellewzhang:
> @michellewzhang What I meant was the feature flag changes that we end up showing on the event volume chart are ones that were part of the evaluation tracking list to begin with (i.e., flags evaluated before an error event in the issue that are shown in the table view on an Issue Details and also part of the feature flag drawer). We did this as a way to filter the FF change annotations to be less noisy and show more relevant flags

wait then why am i seeing lines after the error event happened

SCR-20250919-kwaw

docs/product/issues/issue-details/feature-flags/index.mdx
```diff
@@ -48,7 +48,7 @@ Change tracking enables Sentry to listen for additions, removals, and modificati
 
 ![Flag Change Tracking](./img/FlagChangeTracking.png)
 
-Within an issue, the audit log is represented in the "event" chart and presents itself as a "release" line.
+Within an issue, the audit log is represented in the "event" chart and presents itself as a "release" line. The lines will only appear for flags that were evaluated before the error event occurred. This feature requires evaluation tracking to be enabled.
```
michellewzhang:
```suggestion
Within an issue, the audit log is represented in the "event" chart and presents itself as a "release" line. The lines will only appear for flags that were evaluated before the error event occurred. This feature requires evaluation tracking to be enabled. 
```
makes sense to me now, but maybe we want to add a note here that we might still show flag changes after the error event to make the difference clear

jas-kas:
```suggestion
Within an issue, the audit log is represented in the "event" chart and presents itself as a "release" line. The lines will only appear for flags that were evaluated before the error event occurred; however, feature flag definition changes (audit log) can still occur after an error. This feature requires evaluation tracking to be enabled. 
```

jas-kas: ok, updated!

docs/product/issues/issue-details/feature-flags/index.mdx
```diff
@@ -48,7 +48,7 @@ Change tracking enables Sentry to listen for additions, removals, and modificati
 
 ![Flag Change Tracking](./img/FlagChangeTracking.png)
 
-Within an issue, the audit log is represented in the "event" chart and presents itself as a "release" line.
+Within an issue, the audit log is represented in the "event" chart and presents itself as a "release" line. The lines will only appear for flags that were evaluated before the error event occurred; however, feature flag definition changes (audit log) can still occur after an error. This feature requires evaluation tracking to be enabled.
```
michellewzhang:
```suggestion
Within an issue, the audit log is represented in the "event" chart and presents itself as a "release" line. The lines will only appear for flags that were evaluated before the error event occurred; however, feature flag definition changes from the audit log can still occur after an error. This feature requires evaluation tracking to be enabled. 
```

michellewzhang (approved):

cmanallen:
I think we should show flags regardless of if its before or after the event so long as they're related. It might show that toggling the flag off stopped the errors. That would be an interesting finding on it owns.

jas-kas:
@cmanallen 

> I think we should show flags regardless of if its before or after the event so long as they're related

Yes, exactly! As long as the flag is related to the errors in an issue, the feature flag definition changes will display in the event volume chart (and those could be before or after error events in the time-series)

PR #15017, opened 2025-09-23
feat(platform): Enable `useRunAfterProductionCompileHook`

chargome:
This PR enables source map uploads using an after production compile hook, potentially saving us some build time.


chargome:
stale

PR #15036, opened 2025-09-24
ref(rust): clarify logs defaults in new release

lcian:
We made some changes to remove friction in the setup for logs in the Rust SDK in versions `0.43.0` and `0.44.0`.
This updates the docs to reflect that.

Close https://github.com/getsentry/sentry-rust/issues/896
Close RUST-105


docs/platforms/rust/common/logs/index.mdx
```diff
@@ -70,7 +70,7 @@ As of today, array and object types will be converted to strings using their JSO
 
 ## Integrations
 
-### `tracing` Integration
+### `tokio-rs/tracing` Integration
```
szokeasaurusrex: Why the name change here?

docs/platforms/rust/common/logs/index.mdx
```diff
@@ -70,7 +70,7 @@ As of today, array and object types will be converted to strings using their JSO
...
 
 To use the `tracing` integration with logs, add the necessary dependencies to your `Cargo.toml`.
```
szokeasaurusrex, line 66: m: Assuming we keep the name change above, please also use the same name consistently throughout the doc

```suggestion
To use the `tokio-rs/tracing` integration with logs, add the necessary dependencies to your `Cargo.toml`.
```

docs/platforms/rust/common/logs/index.mdx
```diff
@@ -130,23 +119,25 @@ fn main() {
...
         tracing::Level::TRACE => EventFilter::Ignore,
-        // Capture everything else just as a log
-        _ => EventFilter::Log,
+        // Capture everything else as both a breadcrumb and a log
+        _ => EventFilter::Breadcrumb | EventFilter::Log,
     });
-```
 
-### `log` Integration
+### `log` crate Integration
```
szokeasaurusrex: Also why are we changing the name here?

l: I find it weird that we call this integration "`log` crate" but the other one just "`tokio-rs/tracing`" without the word "crate" – we should stay consistent.

docs/platforms/rust/guides/tracing/index.mdx
```diff
@@ -88,21 +88,25 @@ async fn fail() {
 }
 ```
 
-By default, error level events from `tokio-rs/tracing` are captured as Sentry events, while anything at or above info is added as a breadcrumb.
+By default, error level events from `tokio-rs/tracing` are captured as Sentry events, while anything at or above info level is added as a breadcrumb.
```
szokeasaurusrex: l: grammar/stylenits

```suggestion
By default, error-level events from `tokio-rs/tracing` are captured as Sentry events, and events at or above the info level are captured as breadcrumbs.
```

docs/platforms/rust/guides/tracing/index.mdx
```diff
@@ -88,21 +88,25 @@ async fn fail() {
...
 
-To capture structured logs from `tokio-rs/tracing` events instead, you need to pass `enable_logs: true` to `sentry::init`, and set up the Sentry layer with a custom event filter that maps to logs, like so:
+Additionally, if the `logs` feature flag of the `sentry` crate is enabled and the SDK is initialized with `enable_logs: true`, then events from `tokio-rs/tracing` at info level or above are also captured as [structured logs](https://docs.sentry.io/product/explore/logs/).
```
szokeasaurusrex: [question unrelated to this PR]: Needing to enable both the `logs` feature flag and `enable_logs: true` seems to be like a double-opt-in. Could we make it so that only the feature flag is needed?

lcian: Introducing the client option was mostly for the sake of having a uniform API across SDKs, but I agree it would make sense to make the feature flag be enough in this case.

szokeasaurusrex: As long as the feature is opt-in (not a default feature), I think it makes sense!

docs/platforms/rust/common/logs/index.mdx
lcian: I've changed the name because in the main docs we also use `tokio-rs/tracing` to avoid confusion between the Sentry tracing feature and the sentry crate.
But I guess this is not relevant here, reverting the change.

PR #15069, opened 2025-09-29
docs(react testing): Update testing docs to demo nested route fields

ryan953:
This documents some new nested-route fields that were added in https://github.com/getsentry/sentry/pull/100481

---

> [!NOTE]
> Extend RTL testing docs with outlet context configuration and nested routes examples, plus a clarified route param section.
> 
> - **Docs (Using React Testing Library)**:
>   - **Route Param Values**: Add a dedicated heading and example clarifying use of `route` for `useParams()`.
>   - **Outlet Context**: Document passing `initialRouterConfig.outletContext` and consuming via `useOutletContext()`.
>   - **Nested Routes**:
>     - Show configuring an in-memory route tree via `initialRouterConfig.children`.
>     - Add wrapper `Outlet` with typed context and custom hook, plus examples for index and param child routes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c1398546d63fafa6e9271c533ba335a8581a3e8b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/develop-docs/frontend/using-rtl.mdx
+++ b/develop-docs/frontend/using-rtl.mdx
@@ -77,6 +77,8 @@ router.navigate('/new/path/');
 router.navigate(-1); // Simulates clicking the back button
 ```
 
+### Route Param Values
+
 If you need to test route param values (as in `useParams()`), the `route` will need to be provided in the config:
 
 ```tsx
@@ -97,6 +99,8 @@ const {router} = render(<TestComponent />, {
 expect(screen.getByText('123')).toBeInTheDocument();
 ```
 
+### Outlet Context
+
 If you need to test outlet param values (as in `useOutletContext()`), you can pass them inside the `initialRouterConfig` object.
 
 ```tsx
@@ -113,6 +117,8 @@ const { router } = render(<TestComponent />, {
 
 expect(screen.getByText('123')).toBeInTheDocument();
 ```
+
+### Nested Routes
 
 If directly setting outlet context isn't enough, instead you want to render some nested routes you can do that too!
 
--- a/develop-docs/frontend/using-rtl.mdx
+++ b/develop-docs/frontend/using-rtl.mdx
@@ -169,7 +169,7 @@ expect(screen.getByText('Settings > Default')).toBeInTheDocument();
 render(<SettingsWrapper />, {
   initialRouterConfig: {
     location: {
-      pathname: '/settings/123',
+      pathname: '/settings/123/',
     },
     route: '/settings/:projectId',
     children: [
--- a/develop-docs/frontend/using-rtl.mdx
+++ b/develop-docs/frontend/using-rtl.mdx
@@ -138,7 +138,7 @@ We can configure the in-memory router to know about just the route tree that you
 ```tsx
 // settingsWrapper.tsx
 export interface OutletContext {
-  settings: {name: string}
+ name: string,
 }
 
 export function useCustomOutletContext() {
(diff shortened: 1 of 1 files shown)


develop-docs/frontend/using-rtl.mdx
```diff
@@ -97,6 +99,91 @@ const {router} = render(<TestComponent />, {
 expect(screen.getByText('123')).toBeInTheDocument();
 ```
 
+### Outlet Context
+
+If you need to test outlet param values (as in `useOutletContext()`), you can pass them inside the `initialRouterConfig` object.
+
+```tsx
+function TestComponent() {
+  const { id } = useOutletContext();
```
evanpurkhiser, line 108: We should suggest they use a typed hook no? These return `unknown`.

Really they should be typed as

```tsx
interface OutletContext {
  id: string
}

export function useWhateverOutletContext() {
  return useOutletContext<OutletContext>();
}

export function MyParentView() {
  return <Outlet id="whatever" />
}
```

ryan953: ah yeah. i can add that in so it's most copy+pasteable

develop-docs/frontend/using-rtl.mdx
```diff
@@ -97,6 +99,105 @@ const {router} = render(<TestComponent />, {
...
+    {path: ':projectId/', component: ProjectSettings},
+  ]
+}
+```
+
+We can configure the in-memory router to know about just the route tree that you need. Remember to render the wrapper component itself, and the router + route child components will render themsevles.
+```tsx
+// settingsWrapper.tsx
+export interface OutletContext {
+  settings: {name: string}
```
ryan953:
```suggestion
 name: string,
```

develop-docs/frontend/using-rtl.mdx
```diff
@@ -97,6 +99,105 @@ const {router} = render(<TestComponent />, {
...
+    {index: true, component: SettingsIndex},
+    {path: ':projectId/', component: ProjectSettings},
+  ]
+}
+```
+
+We can configure the in-memory router to know about just the route tree that you need. Remember to render the wrapper component itself, and the router + route child components will render themsevles.
+```tsx
+// settingsWrapper.tsx
+export interface OutletContext {
```
evanpurkhiser: This shouldn't need to be exported

develop-docs/frontend/using-rtl.mdx
```diff
@@ -97,6 +99,105 @@ const {router} = render(<TestComponent />, {
...
+ name: string,
+}
+
+export function useCustomOutletContext() {
+  return useOutletContext<OutletContext>();
+}
+
+export default function SettingsWrapper() {
+  return <Outlet context={{name: 'Default'} as OutletContext} />;
```
evanpurkhiser:
```suggestion
  return <Outlet context={{name: 'Default'} satisfies OutletContext} />;
```

this is probably better

or even better would be 
```suggestion
  const context: OutletContext = {name: "Default"};
  return <Outlet context={context} />;
```

develop-docs/frontend/using-rtl.mdx
```diff
@@ -97,6 +99,105 @@ const {router} = render(<TestComponent />, {
...
+}
+
+// settingsIndex.tsx
+function SettingsIndex() {
+  const {settings} = useCustomOutletContext();
```
evanpurkhiser: Shouldn't this be 
```suggestion
  const {name} = useCustomOutletContext();
```

evanpurkhiser (approved):

PR #15122, opened 2025-10-03
Update noProxy list in index.mdx

claeyswo:
With the new changes in 25.9.0 it seems we should add pgbouncer and seaweedfs to the no proxy list as well.


aldy505 (approved):
Yes! Thank you!

PR #15123, opened 2025-10-03
feat(go): refactor options page

giortzisg:
## DESCRIBE YOUR PR

Refactor the go options page to follow the format from other SDKs:
- add `SdkOption` element for config options, along with the default values.
-  include more details for each configuration option available.

### DIFF
--- a/docs/platforms/go/common/configuration/options.mdx
+++ b/docs/platforms/go/common/configuration/options.mdx
@@ -149,12 +149,6 @@ Used to customize the sampling of traces, overrides `TracesSampleRate`. This is 
 
 </SdkOption>
 
-<SdkOption name="ProfilesSampleRate" type="float64" defaultValue="0.0">
-
-A number between `0.0` and `1.0`, controlling the percentage chance a given sampled transaction will be profiled. (`0.0` represents 0% while `1.0` represents 100%.) Applies equally to all transactions created in the app. This is relative to the tracing sample rate - e.g. `0.5` means 50% of sampled transactions will be profiled.
-
-</SdkOption>
-
 <SdkOption name="MaxSpans" type="int" defaultValue="1000">
 
 Maximum number of spans that can be attached to a transaction.


docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+If you enable this option, be sure to manually remove what you don't want to send using our features for managing [_Sensitive Data_](../../data-management/sensitive-data/).
+
+</SdkOption>
+
+<SdkOption name="ServerName" type="string">
+
+This option can be used to supply a server name. When provided, the name of the server is sent along and persisted in the event. For many integrations, the server name actually corresponds to the device hostname, even in situations where the machine is not actually a server.
+
+Most SDKs will attempt to auto-discover this value.
```
lcian:
```suggestion
```

remove?

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+</SdkOption>
+
+<SdkOption name="IgnoreErrors" type="[]string">
+
+A list of regexp strings that will be used to match against event's message and if applicable, caught errors type and value. If the match is found, then a whole event will be dropped.
```
lcian:
```suggestion
A list of regexp strings that will be used to match against an event's message and, if applicable, the caught error's type and value. If a match is found, then the whole event will be dropped.
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+<SdkOption name="TracesSampleRate" type="float64" defaultValue="0.0" envVar="SENTRY_TRACES_SAMPLE_RATE">
+
+A number between `0.0` and `1.0`, controlling the percentage chance a given transaction will be sent to Sentry (`0.0` represents 0% while `1.0` represents 100%.) Applies equally to all transactions created in the app. Either this or `traces-sampler` must be defined to enable tracing.
+
+</SdkOption>
+
+<SdkOption name="TracesSampler" type="TracesSampler">
+
+Used to customize the sampling of traces, overrides TracesSampleRate. A function responsible for determining the percentage chance a given transaction will be sent to Sentry. It will automatically be passed information about the transaction and the context in which it's being created, and must return a number between `0` (0% chance of being sent) and `1` (100% chance of being sent). Can also be used for filtering transactions, by returning 0 for those that are unwanted. Either this or `TracesSampleRate` must be defined to enable tracing.
```
lcian:
```suggestion
Used to customize the sampling of traces, overrides `TracesSampleRate`. This is a function responsible for determining the percentage chance a given transaction will be sent to Sentry. It will automatically be passed information about the transaction and the context in which it's being created, and must return a number between `0` (0% chance of being sent) and `1` (100% chance of being sent). Can also be used for filtering transactions, by returning 0 for those that are unwanted. Either this or `TracesSampleRate` must be defined to enable tracing.
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+</SdkOption>
+
+<SdkOption name="ProfilesSampleRate" type="float64" defaultValue="0.0">
+
+A number between `0.0` and `1.0`, controlling the percentage chance a given sampled transaction will be profiled. (`0.0` represents 0% while `1.0` represents 100%.) Applies equally to all transactions created in the app. This is relative to the tracing sample rate - e.g. `0.5` means 50% of sampled transactions will be profiled.
+
+</SdkOption>
```
lcian:
```suggestion
```

Is this still applicable?

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+## Hooks
+
+These options can be used to hook the SDK in various ways to customize the reporting of events.
+
+<SdkOption name="BeforeSend" type="func(event *Event, hint *EventHint) *Event">
+
+This function is called with an SDK-specific message or error event object, and can return a modified event object, or `nil` to skip reporting the event. This can be used, for instance, for manual PII stripping before sending.
```
lcian:
```suggestion
This function is called with an event object, and can return a modified event object, or `nil` to skip reporting the event. This can be used, for instance, for manual PII stripping before sending.
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+By the time `BeforeSend` is executed, all scope data has already been applied to the event. Further modification of the scope won't have any effect.
+
+</SdkOption>
+
+<SdkOption name="BeforeSendTransaction" type="func(event *Event, hint *EventHint) *Event">
+
+This function is called with an SDK-specific transaction event object, and can return a modified transaction event object, or `nil` to skip reporting the event. One way this might be used is for manual PII stripping before sending.
```
lcian:
```suggestion
This function is called with a transaction object, and can return a modified transaction object, or `nil` to skip reporting the transaction. One way this might be used is for manual PII stripping before sending.
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+</SdkOption>
+
+<SdkOption name="BeforeBreadcrumb" type="func(breadcrumb *Breadcrumb, hint *BreadcrumbHint) *Breadcrumb">
+
+This function is called with an SDK-specific breadcrumb object before the breadcrumb is added to the scope. When `nil` is returned from the function, the breadcrumb is dropped. To pass the breadcrumb through, return the first argument, which contains the breadcrumb object.
+The callback typically gets a second argument (called a "hint") which contains the original object from which the breadcrumb was created to further customize what the breadcrumb should look like.
```
lcian:
```suggestion
This function is called with a breadcrumb object before the breadcrumb is added to the scope. When `nil` is returned from the function, the breadcrumb is dropped. To pass the breadcrumb through, return the first argument, which contains the breadcrumb object.
The callback gets a second argument (called a "hint") which contains the original object from which the breadcrumb was created to further customize what the breadcrumb should look like.
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+Integrations to be installed on the current Client, receives default integrations. This function can be used to add additional integrations or remove default integrations. For more information, please see our documentation for a specific integration.
+
+See the [Removing Default Integrations](#removing-default-integrations) section below for an example.
+
+</SdkOption>
+
+<SdkOption name="DebugWriter" type="io.Writer">
+
+io.Writer implementation that should be used with the Debug mode. This allows you to redirect debug output to a custom writer instead of stdout.
```
lcian:
```suggestion
`io.Writer` implementation that should be used with the Debug mode. This allows you to redirect debug output to a custom writer instead of stdout.
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+</SdkOption>
+
+## Transport Options
+
+Transports are used to send events to Sentry. Transports can be customized to some degree to better support highly specific deployments.
+
+<SdkOption name="Transport" type="Transport">
+
+The transport to use. Defaults to HTTPTransport. Switches out the transport used to send events. How this works depends on the SDK. It can, for instance, be used to capture events for unit-testing or to send it through some more complex setup that requires proxy authentication.
```
lcian:
```suggestion
The transport to use. Defaults to `HTTPTransport`. Switches out the transport used to send events. How this works depends on the SDK. It can, for instance, be used to capture events for unit-testing or to send it through some more complex setup that requires proxy authentication.
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+</SdkOption>
+
+<SdkOption name="HTTPClient" type="*http.Client">
+
+An optional pointer to http.Client that will be used with a default HTTPTransport. Using your own client will make HTTPTransport, HTTPProxy, HTTPSProxy and CaCerts options ignored.
```
lcian:
```suggestion
An optional pointer to `http.Client` that will be used with a default HTTPTransport. Using your own client will make HTTPTransport, HTTPProxy, HTTPSProxy and CaCerts options ignored.
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+<SdkOption name="HTTPTransport" type="http.RoundTripper">
+
+An optional pointer to http.Transport that will be used with a default HTTPTransport. Using your own transport will make HTTPProxy, HTTPSProxy and CaCerts options ignored.
+
+</SdkOption>
+
+<SdkOption name="HTTPProxy" type="string" envVar="HTTP_PROXY">
+
+When set, a proxy can be configured that should be used for outbound requests. This is also used for HTTPS requests unless a separate `HTTPSProxy` is configured. However, not all SDKs support a separate HTTPS proxy. SDKs will attempt to default to the system-wide configured proxy, if possible. For instance, on Unix systems, the `HTTP_PROXY` environment variable will be picked up.
```
lcian:
```suggestion
When set, a proxy can be configured that should be used for outbound requests. This is also used for HTTPS requests unless a separate `HTTPSProxy` is configured.
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
 
 <PlatformContent includePath="configuration/config-intro" />
 
+## Core Options
+
+Options that can be read from an environment variable (`SENTRY_DSN`, `SENTRY_ENVIRONMENT`, `SENTRY_RELEASE`) are read automatically.
+
+<SdkOption name="Dsn" type="string" envVar="SENTRY_DSN">
+
+The DSN tells the SDK where to send the events. If this value is not provided, the SDK will try to read it from the `SENTRY_DSN` environment variable. If that variable also does not exist, the SDK will just not send any events.
```
coolguyzone:
```suggestion
The DSN tells the SDK where to send the events. If this value is not provided, the SDK will try to read it from the `SENTRY_DSN` environment variable. If that's also missing, no events will be sent.
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+Learn more about [DSN utilization](/product/sentry-basics/dsn-explainer/#dsn-utilization).
+
+</SdkOption>
+
+<SdkOption name="Debug" type="bool" defaultValue="false">
+
+If debug is enabled, the SDK will attempt to print out useful debugging information if something goes wrong while sending the event. The default is always `false`. It's generally not recommended to turn it on in production, though turning `debug` mode on will not cause any safety concerns.
```
coolguyzone:
```suggestion
If debug is enabled, the SDK will attempt to print out useful debugging information if something goes wrong while sending the event. It's always `false` by default and generally not recommended in production environments, though it's safe to use.
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+</SdkOption>
+
+<SdkOption name="Release" type="string" envVar="SENTRY_RELEASE">
+
+Sets the release. Some Sentry features are built around releases, and, thus, reporting events with a non-empty release improves the product experience. See [the releases documentation](/product/releases/).
```
coolguyzone:
```suggestion
Sets the release. Some Sentry features are built around release info, so reporting releases can help improve the overall experience. See [the releases documentation](/product/releases/).
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+Environments tell you where an error occurred, whether that's in your production system, your staging server, or elsewhere. Sentry automatically creates an environment when it receives an event with the environment parameter set.
+
+</SdkOption>
+
+<SdkOption name="SampleRate" type="float64" defaultValue="1.0">
+
+Configures the sample rate for error events, in the range of `0.0` to `1.0`. The default is `1.0`, which means that 100% of error events will be sent. If set to `0.1`, only 10% of error events will be sent. Events are picked randomly.
+
+As a historical special case, the sample rate `0.0` is treated as if it was `1.0`. To drop all events, set the DSN to the empty string.
```
coolguyzone:
```suggestion
As a historical special case, the sample rate `0.0` is treated as if it was `1.0`. To drop all events, set the DSN to an empty string.
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+</SdkOption>
+
+<SdkOption name="MaxBreadcrumbs" type="int" defaultValue="100">
+
+This variable controls the total amount of breadcrumbs that should be captured. However, you should be aware that Sentry has a [maximum payload size](https://develop.sentry.dev/sdk/data-model/envelopes/#size-limits) and any events exceeding that payload size will be dropped.
+
+When MaxBreadcrumbs is negative then breadcrumbs are ignored.
```
coolguyzone:
```suggestion
If MaxBreadcrumbs is given a negative value, breadcrumbs are ignored.
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+If this flag is enabled, certain personally identifiable information (PII) is added by active integrations. By default, no such data is sent.
+
+If you enable this option, be sure to manually remove what you don't want to send using our features for managing [_Sensitive Data_](../../data-management/sensitive-data/).
+
+</SdkOption>
+
+<SdkOption name="ServerName" type="string">
+
+This option can be used to supply a server name. When provided, the name of the server is sent along and persisted in the event. For many integrations, the server name actually corresponds to the device hostname, even in situations where the machine is not actually a server.
```
coolguyzone:
```suggestion
Supplies a server name. When provided, the name of the server is sent along and persisted in the event. For many integrations, the server name actually corresponds to the device hostname, even in situations where the machine is not actually a server.
```

docs/platforms/go/common/configuration/options.mdx
coolguyzone:
```suggestion
A list of regular expression strings used to match an event’s message, and, when applicable, the type or value of a caught error. If a match is found, the entire event is dropped.
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+</SdkOption>
+
+
+
+<SdkOption name="MaxErrorDepth" type="int" defaultValue="100">
+
+This is the maximum number of errors reported in a chain of errors. This protects the SDK from an arbitrarily long chain of wrapped errors.
+
+An additional consideration is that arguably reporting a long chain of errors is of little use when debugging production errors with Sentry. The Sentry UI is not optimized for long chains either. The top-level error together with a stack trace is often the most useful information.
```
coolguyzone:
```suggestion
In practice, reporting very long chains usually provides little value when debugging production issues, as the Sentry UI isn’t optimized for them. The top-level error and its stack trace usually contain the most relevant information.
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+</SdkOption>
+
+## Logging Options
+
+<SdkOption name="EnableLogs" type="bool" defaultValue="false">
+
+Set this option to `true` to enable log capturing in Sentry. Only when this is enabled will the SDK capture log messages and send them to Sentry.
```
coolguyzone:
```suggestion
Set this option to `true` to enable log capturing in Sentry. The SDK will only capture and send log messages to Sentry if this option is enabled.
```

docs/platforms/go/common/configuration/options.mdx
```diff
@@ -6,125 +6,258 @@ description: "Learn more about how the SDK can be configured via options. These
...
+
+</SdkOption>
+
+## Tracing Options
+
+<SdkOption name="EnableTracing" type="bool" defaultValue="false">
+
+Enable performance tracing. When enabled, the SDK will create transactions and spans to measure performance.
```
coolguyzone:
```suggestion
Enables performance tracing. When enabled, the SDK will create transactions and spans to measure performance.
```

docs/platforms/go/common/configuration/options.mdx
coolguyzone:
```suggestion
The callback typically gets a second argument (called a "hint") that provides the original object used to create the breadcrumb. You can use this to further customize the breadcrumb’s content or appearance.
```

docs/platforms/go/common/configuration/options.mdx
giortzisg: Nice catch, thought I removed this. It doesn't apply.

PR #15128, opened 2025-10-03
feat: Add example collector config to otlp docs + extra cleanup

AbhiPrasad:
1. re-organize sections in otlp docs
2. add example collector config for both traces and logs
3. add android and flutter sdks to traceparent section (as per https://linear.app/getsentry/project/implement-traceparent-support-in-client-sdks-d842ae07e7d4/updates)
4. removes backend js platforms from traceparent section

### DIFF
--- a/docs/concepts/otlp/index.mdx
+++ b/docs/concepts/otlp/index.mdx
@@ -39,7 +39,7 @@ You can configure your OTEL collector instance to send traces to Sentry directly
 ```yaml {filename: otel-collector.yaml}
 exporters:
   otlphttp:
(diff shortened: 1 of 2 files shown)


mjq (approved):
LGTM. Should we mention the caveat that our OTLP URLs are project-specific, so using them (without routing by service name) from a mixed-service collector might not be appropriate?

AbhiPrasad:
> LGTM. Should we mention the caveat that our OTLP URLs are project-specific

Yes for sure! I added a quick note as part of 5da813993c0f615c56aae679b9b29065fa15d02c

PR #15131, opened 2025-10-04
docs(self-hosted): provide more insights on troubleshooting kafka

aldy505:
Turns out most of our self-hosted users has never touch Kafka before, so it's a good idea to introduce them regarding how Kafka works.

Also added how to increase consumers replica if they're lagging behind.

### DIFF
--- a/develop-docs/self-hosted/troubleshooting/kafka.mdx
+++ b/develop-docs/self-hosted/troubleshooting/kafka.mdx
@@ -4,19 +4,19 @@ sidebar_title: Kafka
 sidebar_order: 2
 ---
 
-## Offset Out Of Range Error
+## How Kafka Works
 
-```log
-Exception: KafkaError{code=OFFSET_OUT_OF_RANGE,val=1,str="Broker: Offset out of range"}
-```
+This section is aimed for those who has Kafka problems, yet not familiar with Kafka. At a high level, it is a message broker which stores message in a log (or in an easier language: very similar to array) format. It receives messages from producers that aimed to a specific topic, and then sends them to consumers that are subscribed to that topic. The consumers can then process the messages.
 
-This happens where Kafka and the consumers get out of sync. Possible reasons are:
+On the inside, when a message enters a topic, it would be written to a certain partition. You can think partition as physical boxes that stores messages for a specific topic, each topic will have their own separate & dedicated partitions. In a distributed Kafka setup, each partition might be stored on a different machine/node, but if you only have a single Kafka instance, then all the partitions are stored on the same machine.
 
-1. Running out of disk space or memory
-2. Having a sustained event spike that causes very long processing times, causing Kafka to drop messages as they go past the retention time
-3. Date/time out of sync issues due to a restart or suspend/resume cycle
+When a producer sends a message to a topic, it will either stick to a certain partition number (example: partition 1, partition 2, etc.) or it will randomly choose a partition. A consumer will then subscribe to a topic and will automatically be assigned to one or more partitions by Kafka. The consumer will then start receiving messages from the assigned partitions. One very important aspect to note is that the number of consumers must not exceed the number of partition for a given topic. If you have more consumers than number of partitions, then the consumers will be hanging with no messages to consume.
 
-### Visualize
+Each messages in a topic will then have an "offset" (number), this would easily translates to "index" in an array. The offset will be used by the consumer to track where it is in the log, and what's the last message it has consumed. If the consumer is not able to keep up with the producer, it will start to lag behind. Most of the times, we want "lag" to be as low as possible, meaning we don't want to have so many unprocessed messages. The easy solution would be adding more partitions and increasing the number of consumers.
+
+The differences with other types of queues or brokers like RabbitMQ or Redis is that Kafka has a concept called "retention time". Messages that are stored on Kafka and consumed by consumers won't be deleted immediately. Instead, they will be stored for a certain period of time. By default, self-hosted Sentry uses Kafka with a retention time of 24 hours. This means that messages that are older than 24 hours will be deleted. If you want to change the retention time, you can do so by modifying the `KAFKA_LOG_RETENTION_HOURS` environment variable in the `kafka` service.
+
+### Visualize Kafka
 
 You can visualize the Kafka consumers and their offsets by bringing an additional container, such as [Kafka UI](https://github.com/provectus/kafka-ui) or [Redpanda Console](https://github.com/redpanda-data/console) into your Docker Compose.
 
@@ -59,6 +59,20 @@ redpanda-console:
     - kafka
 ```
 
+It's recommended to put this on `docker-compose.override.yml` rather than modifying your `docker-compose.yml` directly. The UI will then can be accessed via `http://localhost:8080/` (or `http://<your-ip>:8080/` if you're using a reverse proxy).
+
+## Offset Out Of Range Error
+
+```log
+Exception: KafkaError{code=OFFSET_OUT_OF_RANGE,val=1,str="Broker: Offset out of range"}
+```
+
+This happens where Kafka and the consumers get out of sync. Possible reasons are:
+
+1. Running out of disk space or memory
+2. Having a sustained event spike that causes very long processing times, causing Kafka to drop messages as they go past the retention time
+3. Date/time out of sync issues due to a restart or suspend/resume cycle
+
 Ideally, you want to have zero lag for all consumer groups. If a consumer group has a lot of lag, you need to investigate whether it's caused by a disconnected consumer (e.g., a Sentry/Snuba container that's disconnected from Kafka) or a consumer that's stuck processing a certain message. If it's a disconnected consumer, you can either restart the container or reset the Kafka offset to 'earliest.' Otherwise, you can reset the Kafka offset to 'latest.'
 
 ### Recovery
@@ -77,19 +91,19 @@ The _proper_ solution is as follows ([reported](https://github.com/getsentry/sel
    ```
 2. Receive consumers list:
    ```shell
-   docker compose run --rm kafka kafka-consumer-groups --bootstrap-server kafka:9092 --list
+   docker compose exec kafka kafka-consumer-groups --bootstrap-server kafka:9092 --list
    ```
 3. Get group info:
    ```shell
-   docker compose run --rm kafka kafka-consumer-groups --bootstrap-server kafka:9092 --group snuba-consumers --describe
+   docker compose exec kafka kafka-consumer-groups --bootstrap-server kafka:9092 --group snuba-consumers --describe
    ```
 4. Watching what is going to happen with offset by using dry-run (optional):
    ```shell
-   docker compose run --rm kafka kafka-consumer-groups --bootstrap-server kafka:9092 --group snuba-consumers --topic events --reset-offsets --to-latest --dry-run
+   docker compose exec kafka kafka-consumer-groups --bootstrap-server kafka:9092 --group snuba-consumers --topic events --reset-offsets --to-latest --dry-run
    ```
 5. Set offset to latest and execute:
    ```shell
-   docker compose run --rm kafka kafka-consumer-groups --bootstrap-server kafka:9092 --group snuba-consumers --topic events --reset-offsets --to-latest --execute
+   docker compose exec kafka kafka-consumer-groups --bootstrap-server kafka:9092 --group snuba-consumers --topic events --reset-offsets --to-latest --execute
    ```
 6. Start the previously stopped Sentry/Snuba containers:
    ```shell
@@ -107,14 +121,16 @@ This option is as follows ([reported](https://github.com/getsentry/self-hosted/i
 
 1. Set offset to latest and execute:
    ```shell
-   docker compose run --rm kafka kafka-consumer-groups --bootstrap-server kafka:9092 --all-groups --all-topics --reset-offsets --to-latest --execute
+   docker compose exec kafka kafka-consumer-groups --bootstrap-server kafka:9092 --all-groups --all-topics --reset-offsets --to-latest --execute
    ```
 
 Unlike the proper solution, this involves resetting the offsets of all consumer groups and all topics.
 
 #### Nuclear option
 
-The _nuclear option_ is removing all Kafka-related volumes and recreating them which _will_ cause data loss. Any data that was pending there will be gone upon deleting these volumes.
+<Alert level="warning" title="Warning">
+  The _nuclear option_ is removing all Kafka-related volumes and recreating them which _will_ cause data loss. Any data that was pending there will be gone upon deleting these volumes.
+</Alert>
 
 1. Stop the instance:
    ```shell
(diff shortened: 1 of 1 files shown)


develop-docs/self-hosted/troubleshooting/kafka.mdx
```diff
@@ -77,19 +91,19 @@ The _proper_ solution is as follows ([reported](https://github.com/getsentry/sel
    ```
 2. Receive consumers list:
    ```shell
-   docker compose run --rm kafka kafka-consumer-groups --bootstrap-server kafka:9092 --list
+   docker compose exec kafka kafka-consumer-groups --bootstrap-server kafka:9092 --list
```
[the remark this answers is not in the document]
aldy505: No, we're not stopping Kafka container

develop-docs/self-hosted/troubleshooting/kafka.mdx
```diff
@@ -133,6 +149,33 @@ The _nuclear option_ is removing all Kafka-related volumes and recreating them w
...
+   docker compose exec kafka kafka-topics --bootstrap-server kafka:9092 --describe --topic ingest-events
+   ```
+3. Then, you need to increase the number of consumers for the consumer group. You can see on the `docker-compose.yml` that the container that consumes `ingest-events` topic using ` ingest-consumer` consumer group is `events-consumer` container. But we won't modify the `docker-compose.yml` directly, instead, we will create a new file called `docker-compose.override.yml` and add the following:
+   ```yaml
+   services:
+     events-consumer:
+       deploy:
+         replicas: 3
+   ```
+   This will increase the number of consumers for the `ingest-consumer` consumer group to 3.
```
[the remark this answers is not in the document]
aldy505: > However, the deploy key is only effective in Docker Swarm mode.

Wrong. It works on Docker Compose.

develop-docs/self-hosted/troubleshooting/kafka.mdx
```diff
@@ -4,19 +4,19 @@ sidebar_title: Kafka
 sidebar_order: 2
 ---
 
-## Offset Out Of Range Error
+## How Kafka Works
 
-```log
-Exception: KafkaError{code=OFFSET_OUT_OF_RANGE,val=1,str="Broker: Offset out of range"}
-```
+This section is aimed for those who has Kafka problems, yet not familiar with Kafka. At a high level, it is a message broker which stores message in a log (or in an easier language: very similar to array) format. It receives messages from producers that aimed to a specific topic, and then sends them to consumers that are subscribed to that topic. The consumers can then process the messages.
```
LordSimal:
```suggestion
This section is aimed for those who have Kafka problems, but are not yet familiar with Kafka. At a high level, it is a message broker which stores message in a log (or in an easier language: very similar to an array) format. It receives messages from producers that aimed to a specific topic, and then sends them to consumers that are subscribed to that topic. The consumers can then process the messages.
```

develop-docs/self-hosted/troubleshooting/kafka.mdx
```diff
@@ -4,19 +4,19 @@ sidebar_title: Kafka
...
 
-1. Running out of disk space or memory
-2. Having a sustained event spike that causes very long processing times, causing Kafka to drop messages as they go past the retention time
-3. Date/time out of sync issues due to a restart or suspend/resume cycle
+When a producer sends a message to a topic, it will either stick to a certain partition number (example: partition 1, partition 2, etc.) or it will randomly choose a partition. A consumer will then subscribe to a topic and will automatically be assigned to one or more partitions by Kafka. The consumer will then start receiving messages from the assigned partitions. One very important aspect to note is that the number of consumers must not exceed the number of partition for a given topic. If you have more consumers than number of partitions, then the consumers will be hanging with no messages to consume.
```
LordSimal:
```suggestion
When a producer sends a message to a topic, it will either stick to a certain partition number (example: partition 1, partition 2, etc.) or it will randomly choose a partition. A consumer will then subscribe to a topic and will automatically be assigned to one or more partitions by Kafka. The consumer will then start receiving messages from the assigned partitions. One very important aspect to note is that **the number of consumers must not exceed the number of partition for a given topic**. If you have more consumers than number of partitions, then the consumers will be hanging with no messages to consume.
```

BYK (approved):
This looks waaaay better than the old version but I'm not qualified to give proper feedback. Still unblocking as I think it is miles better that whatever we have currently.

develop-docs/self-hosted/troubleshooting/kafka.mdx
```diff
@@ -4,19 +4,19 @@ sidebar_title: Kafka
 sidebar_order: 2
 ---
 
-## Offset Out Of Range Error
+## How Kafka Works
 
-```log
-Exception: KafkaError{code=OFFSET_OUT_OF_RANGE,val=1,str="Broker: Offset out of range"}
-```
+This section is aimed for those who have Kafka problems, but are not yet familiar with Kafka. At a high level, it is a message broker which stores message in a log (or in an easier language: very similar to an array) format. It receives messages from producers that aimed to a specific topic, and then sends them to consumers that are subscribed to that topic. The consumers can then process the messages.
```
jjbayer:
```suggestion
This section is aimed for those who have Kafka problems, but are not yet familiar with Kafka. At a high level, Kafka is a message broker which stores messages in a log (or in an easier language: very similar to an array) format. It receives messages from producers that write to a specific topic, and then sends them to consumers that are subscribed to that topic. The consumers can then process the messages.
```

develop-docs/self-hosted/troubleshooting/kafka.mdx
```diff
@@ -4,19 +4,19 @@ sidebar_title: Kafka
...
 
-This happens where Kafka and the consumers get out of sync. Possible reasons are:
+On the inside, when a message enters a topic, it would be written to a certain partition. You can think partition as physical boxes that stores messages for a specific topic, each topic will have their own separate & dedicated partitions. In a distributed Kafka setup, each partition might be stored on a different machine/node, but if you only have a single Kafka instance, then all the partitions are stored on the same machine.
```
jjbayer: This paragraph has a bit of overlap with the next one and does not add much to understanding kafka IMO, so I think we could remove it completely.

develop-docs/self-hosted/troubleshooting/kafka.mdx
```diff
@@ -4,19 +4,19 @@ sidebar_title: Kafka
...
 
-1. Running out of disk space or memory
-2. Having a sustained event spike that causes very long processing times, causing Kafka to drop messages as they go past the retention time
-3. Date/time out of sync issues due to a restart or suspend/resume cycle
+When a producer sends a message to a topic, it will either stick to a certain partition number (example: partition 1, partition 2, etc.) or it will randomly choose a partition. A consumer will then subscribe to a topic and will automatically be assigned to one or more partitions by Kafka. The consumer will then start receiving messages from the assigned partitions. One very important aspect to note is that **the number of consumers must not exceed the number of partition for a given topic**. If you have more consumers than number of partitions, then the consumers will be hanging with no messages to consume.
```
jjbayer:
```suggestion
When a producer sends a message to a topic, it will either stick to a certain partition number (example: partition 1, partition 2, etc.) or it will randomly choose a partition. A consumer will then subscribe to a topic and will automatically be assigned to one or more partitions by Kafka. The consumer will then start receiving messages from the assigned partitions. One very important aspect to note is that **the number of consumers within a consumer group must not exceed the number of partition for a given topic**. If you have more consumers than number of partitions, then the consumers will be hanging with no messages to consume.
```

develop-docs/self-hosted/troubleshooting/kafka.mdx
aldy505: I don't mind leaving this here. At least I want to emphasize the existence of "partition"

develop-docs/self-hosted/troubleshooting/kafka.mdx
```diff
@@ -4,19 +4,19 @@ sidebar_title: Kafka
 sidebar_order: 2
 ---
 
-## Offset Out Of Range Error
+## How Kafka Works
 
-```log
-Exception: KafkaError{code=OFFSET_OUT_OF_RANGE,val=1,str="Broker: Offset out of range"}
-```
+This section is aimed for those who have Kafka problems, but are not yet familiar with Kafka. At a high level, Kafka is a message broker which stores messages in a log (or in an easier language: very similar to an array) format. It receives messages from producers that write to a specific topic, and then sends them to consumers that are subscribed to that topic. The consumers can then process the messages.
```
markstory: If you want to be pedantic, Kafka doesn't send messages to consumers. Consumers poll kafka and fetch  new messages.

develop-docs/self-hosted/troubleshooting/kafka.mdx
```diff
@@ -4,19 +4,19 @@ sidebar_title: Kafka
...
 
-1. Running out of disk space or memory
-2. Having a sustained event spike that causes very long processing times, causing Kafka to drop messages as they go past the retention time
-3. Date/time out of sync issues due to a restart or suspend/resume cycle
+When a producer sends a message to a topic, it will either stick to a certain partition number (example: partition 1, partition 2, etc.) or it will randomly choose a partition. A consumer will then subscribe to a topic and will automatically be assigned to one or more partitions by Kafka. The consumer will then start receiving messages from the assigned partitions. One very important aspect to note is that **the number of consumers within a consumer group must not exceed the number of partition for a given topic**. If you have more consumers than number of partitions, then the consumers will be hanging with no messages to consume.
```
markstory: If we use semantic partitioning, we don't generally assign partition numbers to messages. Instead we use 'keyed messages', which define how messages are grouped (by key) but how those keys map to partitions is up to Kafka.

develop-docs/self-hosted/troubleshooting/kafka.mdx
```diff
@@ -4,19 +4,19 @@ sidebar_title: Kafka
...
+
+Each messages in a topic will then have an "offset" (number), this would easily translates to "index" in an array. The offset will be used by the consumer to track where it is in the log, and what's the last message it has consumed. If the consumer is not able to keep up with the producer, it will start to lag behind. Most of the times, we want "lag" to be as low as possible, meaning we don't want to have so many unprocessed messages. The easy solution would be adding more partitions and increasing the number of consumers.
```
markstory: Do you want to mention that offsets are scoped to a partition, and that each partition in a topic will have the same offset numbers?

develop-docs/self-hosted/troubleshooting/kafka.mdx
```diff
@@ -59,6 +59,20 @@ redpanda-console:
...
+
+## Offset Out Of Range Error
+
+```log
+Exception: KafkaError{code=OFFSET_OUT_OF_RANGE,val=1,str="Broker: Offset out of range"}
+```
+
+This happens where Kafka and the consumers get out of sync. Possible reasons are:
+
+1. Running out of disk space or memory
```
bmckerry, line 72: I could be wrong, but I think if kafka runs out of disk space the service crashes (which wouldn't cause `offset out of range` on consumers)

develop-docs/self-hosted/troubleshooting/kafka.mdx
bmckerry: Adding on this, if messages don't have partition keys, kafka will assign the message to a partition via round-robin (so technically not randomly)

develop-docs/self-hosted/troubleshooting/kafka.mdx
aldy505: Yes I should be mentioning that

develop-docs/self-hosted/troubleshooting/kafka.mdx
aldy505: Too pedantic and not so beginner friendly 😅

develop-docs/self-hosted/troubleshooting/kafka.mdx
aldy505: Good point. Will update these.

develop-docs/self-hosted/troubleshooting/kafka.mdx
aldy505: Technically speaking, you're right. But after a disk out of space incident, there will potentially be a massive offset out of range error.

Perhaps this should be made clearer.

develop-docs/self-hosted/troubleshooting/kafka.mdx
sfanahata:
```suggestion
This section is aimed for those who have Kafka problems, but are not yet familiar with Kafka. At a high level, Kafka is a message broker which stores messages in a log, very similar to an array, format. It receives messages from producers that write to a specific topic, and then sends them to consumers that are subscribed to that topic. The consumers can then process the messages.
```

develop-docs/self-hosted/troubleshooting/kafka.mdx
```diff
@@ -4,19 +4,19 @@ sidebar_title: Kafka
...
-## Offset Out Of Range Error
+## How Kafka Works
 
-```log
-Exception: KafkaError{code=OFFSET_OUT_OF_RANGE,val=1,str="Broker: Offset out of range"}
-```
+This section is aimed for those who have Kafka problems, but are not yet familiar with Kafka. At a high level, Kafka is a message broker which stores messages in a log (or in an easier language: very similar to an array) format. It receives messages from producers that write to a specific topic, and then sends them to consumers that are subscribed to that topic. The consumers can then process the messages.
 
-This happens where Kafka and the consumers get out of sync. Possible reasons are:
+On the inside, when a message enters a topic, it would be written to a certain partition. You can think partition as physical boxes that stores messages for a specific topic, each topic will have their own separate & dedicated partitions. In a distributed Kafka setup, each partition might be stored on a different machine/node, but if you only have a single Kafka instance, then all the partitions are stored on the same machine.
```
sfanahata:
```suggestion
On the inside, when a message enters a topic, it will be written to a certain partition. You can think of a partition as physical a box that stores messages for a specific topic. In a distributed Kafka setup, each partition might be stored on a different machine/node, but if you only have a single Kafka instance, then all the partitions are stored on the same machine.
```

develop-docs/self-hosted/troubleshooting/kafka.mdx
sfanahata:
```suggestion
When a producer sends a message to a topic, it will either stick to a certain partition number (example: partition 1, partition 2, etc.) or it will randomly choose a partition. A consumer will then subscribe to a topic and will automatically be assigned to one or more partitions by Kafka. The consumer will then start receiving messages from the assigned partitions. **Important to note: the number of consumers cannot exceed the number of partitions**. If you have more consumers than partitions, the extra consumers will receive no messages.
```

develop-docs/self-hosted/troubleshooting/kafka.mdx
sfanahata:
```suggestion
Each message in a topic will have an "offset" (number). You can think of this like an "index" in an array. The offset will be used by the consumer to track where it is in the log, and what's the last message it has consumed. If the consumer is not able to keep up with the producer, it will start to lag behind. Most of the time, we want "lag" to be as low as possible. The easiest solution to lagging is adding more partitions and increasing the number of consumers.
```

sfanahata:
```suggestion
Each messages in a topic will then have an "offset" (number), this would easily translates to "index" in an array. The offset will be used by the consumer to track where it is in the log, and what's the last message it has consumed. If the consumer is not able to keep up with the producer, it will start to lag behind. Most of the times, we want "lag" to be as low as possible, meaning we don't want to have so many unprocessed messages. The easy solution would be adding more partitions and increasing the number of consumers. Learn more about [lagging](/self-hosted/troubleshooting/kafka/#consumers-lagging-behind).
```

sfanahata (commented):
There is some wording that could use cleaning up. I left a few suggestions. Overall, I love the additional information about partitions and consumers.

PR #15178, opened 2025-10-10
docs(dev-docs): Add metrics telemetry specs

chargome:
- Adds a new entry to `sdk/telemetry` for `metrics`.
- Adds a new entry to `/sdk/data-model/envelope-items/` for the trace metric item

Took the logs specs as a template.

ref https://github.com/getsentry/sentry-javascript/pull/17883

### DIFF
--- a/develop-docs/sdk/telemetry/metrics.mdx
+++ b/develop-docs/sdk/telemetry/metrics.mdx
@@ -124,7 +124,7 @@ Example with multiple attribute types:
 
 #### Counter
 
-Counters track the number of times an event occurs. They only increase and are useful for tracking things like requests, errors, or any other countable event. Note that the unit is defined by the `name` of the metric itself.
+Counters track the number of times an event occurs. They only increase and are useful for tracking things like requests, errors, or any other countable event.
 
 ```json
 {
@@ -212,7 +212,7 @@ The parameters for these methods should be the following, or an equivalent form 
 - `name` **String, required**: The name of the metric
 - `value` **Number, required**: The value of the metric
 - `options` **Object, optional**: An object containing the following properties:
-  - `unit` **String, optional**: The unit of measurement (only for `distribution` and `gauge`, for `counter` the unit is defined by `name`)
+  - `unit` **String, optional**: The unit of measurement (only used for `distribution` and `gauge`)
   - `attributes` **Object, optional**: A dictionary of attributes (key-value pairs with type information)
   - `scope` **Scope, optional**: The scope to capture the metric with.
 


develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+
+- For `counter` metrics: the count to increment by (should default to 1)
+- For `gauge` metrics: the current value
+- For `distribution` metrics: a single measured value
+
+Integers should be a 64-bit signed integer, while doubles should be a 64-bit floating point number.
+
+`trace_id`
+
+: **String, optional**. The trace id of the metric. The value should be 16 random bytes encoded as a hex string (32 characters long). The trace id should be grabbed from the current propagation context in the SDK.
```
Dav1dde:
```suggestion
: **String, required**. The trace id of the metric. The value should be 16 random bytes encoded as a hex string (32 characters long). The trace id should be grabbed from the current propagation context in the SDK.
```

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+
+#### Future Default Attributes
+
+The SDKs should aim to minimize the number of default attributes attached to a metric. Metrics cardinality can explode quickly with too many attributes, and we want to keep metrics lightweight and cost-effective.
+
+New default attributes should only be added after significant feedback from users and discussion internally with the SDK and ingest teams. There is no hard rule about what exact attributes are allowed, every proposed new attribute will be evaluated on a case-by-case basis.
+
+### Data Category and Rate Limiting
+
+TBD
```
Dav1dde: This is the same like for any other envelope item. Relay will emit rate limits in the `trace_metric` category. When the SDK receives such a rate limit it should (must?) not send any trace metrics until the limit is lifted (as known by the communicated end timestamp).

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+
+## Public API
+
+API wise the SDKs are required to expose metrics methods which are to be defined in a `metrics` module or namespace. The SDKs should also include some initialization options to configure the behavior of metrics in the SDK.
+
+### Initialization Options
+
+The SDKs must expose the following configuration options:
+
+- `enableMetrics`/`enable_metrics`: A boolean flag to control if metric envelopes will be generated and sent to Sentry via the Sentry SDK's Metrics APIs. If this flag is set to `false`, the SDK should not send metrics to Sentry. Defaults to `false`.
```
cleptric:
```suggestion
- `enableMetrics`/`enable_metrics`: A boolean flag to control if metric envelopes will be generated and sent to Sentry via the Sentry SDK's Metrics APIs. If this flag is set to `false`, the SDK should not send metrics to Sentry. Defaults to `true`.
```

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+
+- `beforeSendMetric`/`before_send_metric`: A function that takes a metric object and returns a metric object. This function is called before sending the metric to Sentry. It can be used to modify the metric object or to prevent the metric from being sent to Sentry. This function is optional.
+
+While the metrics functionality for an SDK is in an experimental state, SDKs should put these configuration options in an experimental namespace to avoid breaking changes. Also, since integrations will not send automatically generated metrics, the `enableMetrics` flag will not be needed after metrics are released and instead default to `true`.
+
+```js
+Sentry.init({
+  _experiments: { enableMetrics: true },
+});
```
cleptric, line 197: Please doc without the `_experiments`, we will only do this for the PoC phase in JS & Python.

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+  - `attributes` **Object, optional**: A dictionary of attributes (key-value pairs with type information)
+  - `scope` **Scope, optional**: The scope to capture the metric with.
+
+Below are some example SDK implementations to get you started. These are not finalized versions of the API and individual SDK authors should ensure the metrics APIs best fit their platforms.
+
+#### JavaScript
+
+```jsx
+// Increment a counter
+Sentry.metrics.increment("button.clicks", 1, {
```
cleptric:
```suggestion
Sentry.metrics.count("button.clicks", 1, {
```

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+Sentry.init({
+  _experiments: { enableMetrics: true },
+});
+```
+
+### Metrics Module
+
+At minimum the SDK needs to implement the following methods for each metric type:
+
+- `Sentry.metrics.increment(name, value, options)` - Increment a counter
```
cleptric:
```suggestion
- `Sentry.metrics.count(name, value, options)` - Increment a counter
```

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+    browser: "chrome",
+  },
+});
+```
+
+#### Python
+
+```python
+# Increment a counter
+Sentry.metrics.increment('button.clicks', 1, {
```
cleptric:
```suggestion
Sentry.metrics.count('button.clicks', 1, {
```

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+  'attributes': {'page': '/dashboard', 'browser': 'chrome'}
+})
+```
+
+#### PHP
+
+```php
+use function Sentry\metrics;
+
+metrics()->increment('button.clicks', 1, [
```
cleptric:
```suggestion
metrics()->count('button.clicks', 1, [
```

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+  'attributes' => ['page' => '/dashboard', 'browser' => 'chrome']
+]);
+```
+
+#### Java
+
+```java
+import io.sentry.Sentry;
+
+Sentry.metrics().increment("button.clicks", 1,
```
cleptric:
```suggestion
Sentry.metrics().count("button.clicks", 1,
```

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+  )
+);
+```
+
+#### Apple
+
+```swift
+// Swift
+SentrySDK.metrics()
+  .increment(
```
cleptric:
```suggestion
  .count(
```

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+
+### Threading and Concurrency Considerations
+
+The threading and concurrency considerations for metrics are the same as for logs. See the [Threading and Concurrency Considerations section in the Logs documentation](/sdk/telemetry/logs/#threading-and-concurrency-considerations) for more information.
+
+## SDK Behavior
+
+In general metric processing should follow this order:
+
+1. Capture metric via [Public APIs](#metrics-module) (e.g. `Sentry.metrics.increment`).
```
cleptric, line 363:
```suggestion
1. Capture metric via [Public APIs](#metrics-module) (e.g. `Sentry.metrics.count`).
```

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+1. (While experimental) Check if metrics are enabled as per `enableMetrics`/`enable_metrics` configuration - if not, skip the rest of the steps.
```
cleptric: Let's not add docs for this, we should focus on the end state when SDKs ship it.

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+
+#### User Attributes
+
+SDKs may optionally attach user information as attributes:
+
+1. `user.id`: The user ID. Maps to `id` in the [User](/sdk/data-model/event-payloads/user/) payload.
+2. `user.name`: The username. Maps to `username` in the [User](/sdk/data-model/event-payloads/user/) payload.
+3. `user.email`: The email address. Maps to `email` in the [User](/sdk/data-model/event-payloads/user/) payload.
+
+Note: Adding user information to metrics may significantly increase cardinality and should be done carefully.
```
cleptric: This is no longer an issue.

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+      "value": "36b75d9fa11f45459412a96c41bdf691",
+      "type": "string"
+    }
+  }
+}
+```
+
+#### User Attributes
+
+SDKs may optionally attach user information as attributes:
```
cleptric: Point out `send_default_pii`.

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+{
+  "attributes": {
+    "user.id": { "value": "123", "type": "string" }
+  }
+}
+```
+
+#### User Agent Parsing
+
+By default, Relay should parse the user agent attached to an incoming log envelope to parse `browser` and `os` information for logs. These attributes should be attached by Relay, but SDKs can attach them if they do not forward a user agent when sending logs to Sentry.
```
cleptric: This should only happen for client SDKs. And there are some logs left overs.

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+
+: **String, required**. The type of metric. One of:
+
+- `counter`: A metric that increments counts
+- `gauge`: A metric that tracks a value that can go up or down
+- `distribution`: A metric that tracks the statistical distribution of values
+
+`name`
+
+: **String, required**. The name of the metric. This should follow a hierarchical naming convention using dots as separators (e.g., `api.response_time`, `db.query.duration`).
```
AbhiPrasad, line 75: > This should follow a hierarchical naming convention using dots as separators

Will this be enforced by the backend? Or just recommended to users? Are spaces in the name valid?

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+
+Integers should be a 64-bit signed integer, while doubles should be a 64-bit floating point number.
+
+`trace_id`
+
+: **String, required**. The trace id of the metric. The value should be 16 random bytes encoded as a hex string (32 characters long). The trace id should be grabbed from the current propagation context in the SDK.
+
+`span_id`
+
+: **String, optional**. The span id for the metric.
```
AbhiPrasad: Is this is the span id of the parent span? Or does this represent the metric item itself? If it does, we can probably remove this from the client protocol.

develop-docs/sdk/telemetry/metrics.mdx
AbhiPrasad: No I think it's fine to have the experimental flag in other sdks as well. For example we did this in .NET for logs while they iterated on the API/integrations.

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+
+`value`
+
+: **Number, required**. The numeric value of the metric. The interpretation depends on the metric `type`:
+
+- For `counter` metrics: the count to increment by (should default to 1)
+- For `gauge` metrics: the current value
+- For `distribution` metrics: a single measured value
+
+Integers should be a 64-bit signed integer, while doubles should be a 64-bit floating point number.
```
AbhiPrasad, line 85: Can you supply both an integer and a float to a metric? So call `increment(3)` and then `increment(4.5)`?

develop-docs/sdk/telemetry/metrics.mdx
AbhiPrasad: Can count take a negative integer/float?

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+- `Sentry.metrics.distribution(name, value, options)` - Add a distribution value
+
+#### Method Signatures
+
+The parameters for these methods are:
+
+- `name` **String, required**: The name of the metric
+- `value` **Number, required**: The value of the metric
+- `options` **Object, optional**: An object containing the following properties:
+  - `unit` **String, optional**: The unit of measurement (distribution and gauge only)
```
AbhiPrasad: Why can't we put units on counts?

develop-docs/sdk/telemetry/metrics.mdx
k-fish: Also once it's public we'll be setting `enableMetrics` to be true so someone would only need to use this option to disable eg. `enableMetrics: false`

develop-docs/sdk/telemetry/metrics.mdx
AbhiPrasad: If we still expose `enableMetrics` as an option, we'll need to do this check anyway. I think we should have it, but just set it to be `true` by default.

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,551 @@
...
+
+New default attributes should only be added after significant feedback from users and discussion internally with the SDK and ingest teams. There is no hard rule about what exact attributes are allowed, every proposed new attribute will be evaluated on a case-by-case basis.
+
+### Data Category and Rate Limiting
+
+A new data category for metrics has been added to Relay, `trace_metric`. Rate limiting applies as usual. SDKs should track client outcomes to monitor how often metrics are dropped. If the SDK receives a rate limit, it should not send any trace metrics until the limit expires (based on the communicated end timestamp).
+
+### Buffering
+
+Metrics should be buffered and aggregated before being sent. SDKs should keep a buffer of metrics on the client that flushes out based on some kind of condition. We recommend following the [batch processor specification](/sdk/telemetry/spans/batch-processor/) outlined in the develop docs, but you should choose the approach that works best for your platform.
```
k-fish: Careful with the term `aggregated` here since generally trace metrics may have a span id, timestamp etc. that we want to preserve, making aggregation on the sdk side not desirable, although we may offer special aggregation functions / ability in the future.

Also as an aside, the metrics and log buffers are currently set to the same in the js sdk (100 items) but we may consider changing that if we keep the low metric size limit (~1-2kib)

develop-docs/sdk/telemetry/metrics.mdx
cleptric: @AbhiPrasad then let’s call it out specifically. By default, SDKs should not ship with the experimental option.

develop-docs/sdk/telemetry/metrics.mdx
chargome: I think counters in metrics should be always increasing – but there is e.g. a separate up/down-counter in otel. We don't prevent passing negative integers from the SDK though

develop-docs/sdk/telemetry/metrics.mdx
jjbayer: I suggest we stick to integers for counters. @k-fish I assume the value eventually ends up in an EAP attribute where integers and floats go into separate columns. Mixing both types might then make queries tricky, plus you'd have to deal with precision loss for very large numbers.

develop-docs/sdk/data-model/envelope-items.mdx
```diff
@@ -457,6 +457,48 @@ _None_
 
 _None_
 
+### Trace Metric
+
+Item type `"trace_metric"` contains an array of metric payloads encoded as JSON. This allows for multiple metric payloads to be sent in a single envelope item.
+
+Only a single trace_metric container is allowed per envelope. The `item_count` field in the envelope item header must match the amount of metrics sent, it's not optional. A `content_type` field in the envelope item header must be set to `application/vnd.sentry.items.trace_metric+json`.
```
coolguyzone:
```suggestion
Only a single trace_metric container is allowed per envelope. The `item_count` field in the envelope item header must match the amount of metrics sent.  A `content_type` field in the envelope item header must be set to `application/vnd.sentry.items.trace_metric+json`. Both the `item_count` and `content_type` fields are required.
```

develop-docs/sdk/data-model/envelope-items.mdx
```diff
@@ -457,6 +457,48 @@ _None_
...
+
+It's okay to mix metrics from different traces into the same trace_metric envelope item, but if you do, you MUST not attach a DSC (dynamic sampling context) to the envelope header.
```
coolguyzone:
```suggestion
It's okay to mix metrics from different traces into the same `trace_metric` envelope item, but if you do, you MUST NOT attach a DSC (dynamic sampling context) to the envelope header.
```

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,553 @@
...
+
+</Alert>
+
+## Metrics Protocol
+
+Metrics are sent to Sentry via the `trace_metric` envelope item. Each `trace_metric` envelope item contains a batch of metric payloads encoded as JSON, allowing for transmission of multiple metrics in a single envelope.
+
+### `trace_metric` Envelope Item
+
+The `trace_metric` envelope item is an object that contains an array of metric payloads encoded as JSON. This allows for multiple metric payloads to be sent in a single envelope item. See [Appendix A](#appendix-a-example-trace_metric-envelope) for an example `trace_metric` envelope.
```
alexander-alderman-webb: nitpick: These paragraphs could be consolidated.

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,553 @@
...
+  'unit': 'connection',
+  'attributes': {'pool_name': 'main_db', 'database': 'postgres'}
+})
+
+# Record a distribution
+Sentry.metrics.distribution('page.load_time', 245.7, {
+  'unit': 'millisecond',
+  'attributes': {'page': '/dashboard', 'browser': 'chrome'}
+})
+```
```
alexander-alderman-webb: Non-blocking.

```suggestion
```python
# Increment a counter
Sentry.metrics.count(
  'button.clicks',
  1,
  attributes={'button_id': 'submit', 'page': 'checkout'},
)

# Set a gauge value
Sentry.metrics.gauge(
  'db.connection_pool.active',
  42,
  unit='connection',
  attributes={'pool_name': 'main_db', 'database': 'postgres'},
)

# Record a distribution
Sentry.metrics.distribution(
  'page.load_time',
  245.7,
  unit='millisecond',
  attributes={'page': '/dashboard', 'browser': 'chrome'},
)
```
```

File: develop-docs/sdk/telemetry/metrics.mdx (hunk 18 above)
Comment (alexander-alderman-webb) [reply develop-docs/sdk/telemetry/metrics.mdx:85]: Tagging on to the discussion, is the SDK expected to truncate large integers? If so, how do we truncate, or do we just reject integers that are beyond 64 bits.

In particular, 64-bit floats are natural in Python, but integers are arbitrary precision (under the hood more memory is allocated if the user provides a large integer).

File: develop-docs/sdk/telemetry/metrics.mdx [hunk 25]
```diff
@@ -0,0 +1,553 @@
...
+});
+```
+
+### Metrics Module
+
+At minimum the SDK needs to implement the following methods for each metric type:
+
+- `Sentry.metrics.count(name, value, options)` - Increment a counter
+- `Sentry.metrics.gauge(name, value, options)` - Set a gauge value
+- `Sentry.metrics.distribution(name, value, options)` - Add a distribution value
```
alexander-alderman-webb, line 206: From an API perspective, would any of these changes make it clearer?

```suggestion
- `Sentry.metrics.increment(name, count, options)` - Increment a counter
- `Sentry.metrics.gauge(name, value, options)` - Set a gauge value
- `Sentry.metrics.distribution(name, sample, options)` - Add a distribution value
```

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,553 @@
...
+
+#### Method Signatures
+
+The parameters for these methods are:
```
alexander-alderman-webb: This sounds prescriptive right now.

```suggestion
The parameters for these methods should be the following, or an equivalent form that is more natural for SDK's language:
```

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,553 @@
...
+
+: **String, required**. The trace id of the metric. The value should be 16 random bytes encoded as a hex string (32 characters long). The trace id should be grabbed from the current propagation context in the SDK.
+
+`span_id`
+
+: **String, optional**. The span id for the metric.
+
+`unit`
+
+: **String, optional**. The unit of measurement for the metric value.
```
alexander-alderman-webb, line 97: Can we clarify in the spec who modifies the input or drops the metric if the user provides malformed arguments with the wrong type?

For example, I provide an integer in the unit field when calling `metrics.count()`. Does the SDK

- drop the metric; and/or
- coerce into a string; and/or
- send an integer and relies on downstream systems to discard the metric?

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,553 @@
...
+
+API wise the SDKs are required to expose metrics methods which are to be defined in a `metrics` module or namespace. The SDKs should also include some initialization options to configure the behavior of metrics in the SDK.
+
+### Initialization Options
+
+The SDKs must expose the following configuration options:
+
+- `enableMetrics`/`enable_metrics`: A boolean flag to control if metric envelopes will be generated and sent to Sentry via the Sentry SDK's Metrics APIs. If this flag is set to `false`, the SDK should not send metrics to Sentry. Defaults to `true`.
+
+- `beforeSendMetric`/`before_send_metric`: A function that takes a metric object and returns a metric object. This function is called before sending the metric to Sentry. It can be used to modify the metric object or to prevent the metric from being sent to Sentry. This function is optional.
```
alexander-alderman-webb, line 186: Since we have 3 different types of metrics, why do not have

- `beforeSendCounter`/`before_send_counter`; and
- `beforeSendGauge`/`before_send_gauge`; and
- `beforeSendDistribution` / `before_send_distribution`

options, especially if counters end up as integers or there are any other differences in the types. Is it ergonomic if we marshal everything into one type? Users then have to write if-conditions in all their callbacks.

See related integer discussion https://github.com/getsentry/sentry-docs/pull/15178/files#r2431749117

cleptric: We only want one `beforeSend` callback for metrics. Besides making it clunky to fiddle around with three callbacks, this pattern already worked well for us in the last Metrics implementation.

develop-docs/sdk/telemetry/metrics.mdx
cleptric: Coerce into a string is fine.

develop-docs/sdk/telemetry/metrics.mdx
cleptric: We want count for the top level APIs. Increment signals it being aggregated, and we don't do this anymore.

develop-docs/sdk/data-model/envelope-items.mdx
k-fish: `trace-metric` is correct vs trace_metric

k-fish:
```suggestion
Only a single trace_metric container is allowed per envelope. The `item_count` field in the envelope item header must match the amount of metrics sent, it's not optional. A `content_type` field in the envelope item header must be set to `application/vnd.sentry.items.trace-metric+json`.
```

develop-docs/sdk/data-model/envelope-items.mdx
```diff
@@ -457,6 +457,48 @@ _None_
...
+
+**Additional Item Headers:**
+
+`item_count`
+
+: **integer, required.** The number of metric entries in the envelope.
+
+`content_type`
+
+: **string, required.** The content type of the metric entries. Must be `application/vnd.sentry.items.trace_metric+json`.
```
k-fish:
```suggestion
: **string, required.** The content type of the metric entries. Must be `application/vnd.sentry.items.trace-metric+json`.
```

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -0,0 +1,553 @@
...
+
+`value`
+
+: **Number, required**. The numeric value of the metric. The interpretation depends on the metric `type`:
+
+- For `counter` metrics: the count to increment by (should default to 1)
+- For `gauge` metrics: the current value
+- For `distribution` metrics: a single measured value
+
+Integers should be a 64-bit signed integer, while doubles should be a 64-bit floating point number.
```
aldy505, line 85: Do we allow `0` as a value for metric? Or should we drop that from client-side?

develop-docs/sdk/telemetry/metrics.mdx
chargome: I don't think we need to enforce this pattern, rather recommend it. I'd restrict whitespaces both in the SDK (trim) and the backend.

develop-docs/sdk/telemetry/metrics.mdx
chargome: That would the be the current span atm

develop-docs/sdk/telemetry/metrics.mdx
chargome: Isn't count the unit itself?

develop-docs/sdk/telemetry/metrics.mdx
chargome: I'd not limit this on the SDK side, we can always decide to drop these during ingestion

Dav1dde: > Do we allow 0 as a value for metric? Or should we drop that from client-side?

A 0 is not relevant for a counter, but it most certainly is for gauges and distributions. So yes a `0` is allowed.

develop-docs/sdk/telemetry/metrics.mdx
cleptric: No, you can count 🥔 for example 😄

chargome: Yeah but that would be determined by the name right? `metrics.count('🥔', 2)`?

cleptric (approved):

PR #15217, opened 2025-10-14
Updating Java options with missing defaults

sfanahata:
## DESCRIBE YOUR PR
Adding more details to Java options since moving everything to SDKoption formatting

Preview: https://sentry-docs-git-java-options-updates.sentry.dev/platforms/java/configuration/options/

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

### DIFF
--- a/docs/platforms/java/common/configuration/options.mdx
+++ b/docs/platforms/java/common/configuration/options.mdx
@@ -28,7 +28,7 @@ Turns debug mode on or off. If debug is enabled SDK will attempt to print out us
 
 </SdkOption>
 
-<SdkOption name="diagnosticLevel" type="enum" defaultValue="debug">
+<SdkOption name="diagnosticLevel" type="enum" defaultValue="DEBUG">
 
 Enabling `debug` mode makes the SDK generate as much diagnostic data as possible. However, if you'd prefer to lower the verbosity of the Sentry SDK diagnostics logs, configure this option to set the appropriate level:
 
@@ -140,11 +140,11 @@ A list of exception types that will be filtered out before sending to Sentry.
 
 </SdkOption>
 
-<SdkOption name="maxRequestBodySize" type="enum" defaultValue="never">
+<SdkOption name="maxRequestBodySize" type="enum" defaultValue="NONE">
 
 This parameter controls whether integrations should capture HTTP request bodies. It can be set to one of the following values:
 
-- `never`: **default** Request bodies are never sent.
+- `none`: **default** Request bodies are never sent.
 - `small`: Only small request bodies will be captured. The cutoff for small depends on the SDK (typically 4KB).
 - `medium`: Medium and small requests will be captured (typically 10KB).
 - `always`: The SDK will always capture the request body as long as Sentry can make sense of it.


docs/platforms/java/common/configuration/options.mdx
```diff
@@ -140,11 +140,11 @@ A list of exception types that will be filtered out before sending to Sentry.
 
 </SdkOption>
 
-<SdkOption name="maxRequestBodySize" type="enum">
+<SdkOption name="maxRequestBodySize" type="enum" defaultValue="never">
```
lcian: Technically it's `RequestSize.NONE` in the code.
Not sure if we want to reflect that or we would rather keep a perhaps more nicer to read `never`, up to you.

docs/platforms/java/common/configuration/options.mdx
```diff
@@ -22,13 +22,13 @@ Learn more about [DSN utilization](/product/sentry-basics/dsn-explainer/#dsn-uti
...
 
-<SdkOption name="debug" type="bool" defaultValue="false">
+<SdkOption name="debug" type="boolean" defaultValue="false">
 
 Turns debug mode on or off. If debug is enabled SDK will attempt to print out useful debugging information if something goes wrong with sending the event. The default is always `false`. It's generally not recommended to turn it on in production, though turning `debug` mode on will not cause any safety concerns.
 
 </SdkOption>
 
-<SdkOption name="diagnosticLevel" type="enum">
+<SdkOption name="diagnosticLevel" type="enum" defaultValue="debug">
```
lcian: Same as above, technically in the code it's `SentryLevel.DEBUG`.

sfanahata: I think we can leave the casing alone, unless you think it would break without explicitly suggesting all caps.

docs/platforms/java/common/configuration/options.mdx
sfanahata: I would go with lower case still, but is `none` the proper customer-facing value that should be used?

docs/platforms/java/common/configuration/options.mdx
lcian: Yeah, my doubt was more about, do we want this to reflect the code exactly or should this be more of an indication of what to pass.
If you wanted this to be strict you would have to make it `SentryLevel.DEBUG` or something, but I think just `debug` is fine and indicates what to pass even if it's not always strictly correct.

If you're configuring this through code you will clearly see in your editor that you need to pass `SentryLevel.DEBUG`.
If you use other configuration methods we have in java (Spring's `application.properties`) then it's fine to specify it as `sentry.diagnostic-level=debug` as we convert the value to upper case when reading it 👍

docs/platforms/java/common/configuration/options.mdx
lcian: Yeah, the value should be `none`, not `never` in this case.
Same reasoning as for the other comment applies to the casing.

lcian:
On a side note @sfanahata we might want to figure out how to deal with the options in the Spring/Spring Boot cases as well.
They are the same options but in those cases you can also configure them through `application.properties`, which is the preferred configuration method in Spring/Spring Boot.
For example `dsn` is specified as `sentry.dsn=<something>` in `application.properties`.
This is not super clear from the docs on https://docs.sentry.io/platforms/java/configuration/options/ and I suspect it might not be super obvious that you can configure that way as well.
I wonder if we should come up with a better way to surface this, maybe extend the options component to handle this case as well? Obviously doesn't need to be done here but we can consider it as a future step.
It is only specific to Spring and Spring Boot though, so maybe we enable it there and automatically derive the property name (e.g. `diagnosticLevel` -> `sentry.diagnostic-level`)

docs/platforms/java/common/configuration/options.mdx
sfanahata: I'll make it upper case to make it clearer.

sfanahata:
> On a side note @sfanahata we might want to figure out how to deal with the options in the Spring/Spring Boot cases as well. They are the same options but in those cases you can also configure them through `application.properties`, which is the preferred configuration method in Spring/Spring Boot. For example `dsn` is specified as `sentry.dsn=<something>` in `application.properties`. This is not super clear from the docs on https://docs.sentry.io/platforms/java/configuration/options/ and I suspect it might not be super obvious that you can configure that way as well. I wonder if we should come up with a better way to surface this, maybe extend the options component to handle this case as well? Obviously doesn't need to be done here but we can consider it as a future step. It is only specific to Spring and Spring Boot though, so maybe we enable it there and automatically derive the property name (e.g. `diagnosticLevel` -> `sentry.diagnostic-level`)

Ohh, this is good information @lcian! Thanks for the call-out. Let me think on how we want to approach this, format wise, I'll start a separate PR for it.

PR #15279, opened 2025-10-21
Document global attributes

cleptric:


develop-docs/sdk/telemetry/scopes.mdx
```diff
@@ -51,6 +51,63 @@ Data from all three scope types MUST be merged in a specific order before being
...
+- `unit` (optional): The unit of measurement (e.g., `"ms"`, `"s"`, `"bytes"`, `"count"`, `"percent"`)
+
+#### Example Usage
+
+```javascript
+Sentry.getGlobalScope().setAttributes({
+  "app.feature_flag.enabled": {
+    type: "boolean",
+    value: true
+  },
```
AbhiPrasad: IMO we should allow for an API that allows for the SDK to infer the type for the user. Otherwise this becomes more annoying for the dynamic language sdks.

The attributes API currently in logs doesn't allow for users to specify the type.

develop-docs/sdk/telemetry/scopes.mdx
```diff
@@ -51,6 +51,63 @@ Data from all three scope types MUST be merged in a specific order before being
...
 
+## Scope Methods
+
+### Setting Attributes
+
+Users MUST be able to attach attributes to any scope using a dedicated method (e.g., `scope.setAttributes()` or `scope.setAttribute()`). These attributes follow the structure defined in the [Span Protocol](/sdk/telemetry/spans/span-protocol/#attribute-object-properties).
+
+Attributes are key-value pairs where each value is an object containing:
+
+- `type`: The data type (`"string"`, `"integer"`, `"double"`, or `"boolean"`)
```
AbhiPrasad: I think we can be less strict here. If we allow for more types, we can cast complex types to string and still allow them to be set. I think this makes it an easier API for users to use, and feels very easy to transition from contexts.

PR #15286, opened 2025-10-22
docs(nuxt): Add Nuxt's Nitro features

logaretm:
Adds entries on the newly added and available Nitro instrumented features released in `10.17.0` and `10.21.0`.

### DIFF
--- a/docs/platforms/javascript/guides/nuxt/features/nitro-server-features.mdx
+++ b/docs/platforms/javascript/guides/nuxt/features/nitro-server-features.mdx
@@ -7,18 +7,13 @@ The Sentry Nuxt SDK automatically instruments various Nitro server APIs to provi
 
 ## Server Middleware
 
-<Alert>
-  Server middleware instrumentation is available from version `@sentry/nuxt`
-  10.17.0 and up.
-</Alert>
+<Alert>Available from version `@sentry/nuxt` 10.17.0 and up.</Alert>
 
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/guides/nuxt/features/nitro-server-features.mdx
```diff
@@ -0,0 +1,51 @@
...
+  Server middleware instrumentation is available from version `@sentry/nuxt`
+  10.17.0 and up.
+</Alert>
+
+The Sentry Nuxt SDK automatically instruments [server middleware handlers](https://nuxt.com/docs/4.x/guide/directory-structure/server#server-middleware) in addition to server event handlers. This includes the handler functions and their `onBeforeResponse` and `onRequest` callbacks.
+
+## Cache
+
+<Alert>
+  Cache instrumentation is available from version `@sentry/nuxt` 10.21.0 and up.
```
JPeer264: l: I love that we add this info here. This reminds me of how [Kubernetes](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/) is doing these updates. However, I think less is more and this might be easier to digest:

```
Available from `@sentry/nuxt` 10.21.0 and up
```

Theoretically we can even write a component that is adding this info, like the `<PlatformContent />` we have (but that is just an idea for the future

JPeer264 (approved):

docs/platforms/javascript/guides/nuxt/features/nitro-server-features.mdx
logaretm: I ideally would like to put it next to the header. I agree on the suggested text, much easier to read, thanks!

The component idea is great! I will see if we can use it around the docs elsewhere.

s1gr1d (approved):

PR #15298, opened 2025-10-23
docs(js): Update Electron quick start guide

### LINKED ISSUE #15269: Create Electron Quick Start guide
### SDK

JavaScript SDK

### Description

Update the [Electron getting started guide](https://docs.sentry.io/platforms/javascript/guides/) into a quick start guide following our template.

~~This may result in two guides because it seems there's an installation wizard available, which we currently mention in the manual setup instructions.~~

### Suggested Solution

_No response_

inventarSarah:
## DESCRIBE YOUR PR
This PR focuses on transforming the existing Electron guide into a Quick Start guide.

**Notes:**
- I've removed the information about the Wizard. After testing it, I found that it created a file where I can add my project/org data, but it didn't set up the SDK (like the Next.js wizard, for example). This could be frustrating for users. Please let me know what you think and if I've missed something!
- I added a new section, verify > I tested the code snippets locally, and they worked in my setup, but please have a good look at them

**Next steps:**
You will notice that I haven't changed much for Steps number 3,4,5, and 7. The content in these sections seems to advanced for a quick start guide. 
I plan to create a **follow-up ticket to move this content** to other pages within the Electron docs (of course, we will link to them from the quick start guide). 
I noticed that @timfish handles a lot of the Electron SDK work, and you seem like the person to involve when changing the docs. Is that correct? If so, I will @ you in the follow-up ticket so we can align on the changes. 

Thanks!!

Closes #15269 

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

## EXTRA RESOURCES

### DIFF
--- a/docs/platforms/javascript/guides/electron/index.mdx
+++ b/docs/platforms/javascript/guides/electron/index.mdx
@@ -112,6 +112,17 @@ Sentry.init({
 });
 ```
 
+<Alert>
+If your app uses a preload script and `contextIsolation: true`, Sentry can't automatically capture errors that occur in the preload context. To include those, initialize Sentry in your preload script as well:
+
+```javascript
+import * as Sentry from "@sentry/electron/renderer";
+
+Sentry.init(); // don't forget to add your configuration options
+```
+
+</Alert>
+
 <Expandable title="Did you change the userData directory?" level="warning">
 If you change the `userData` directory used by your app, ensure this change is made before you configure the SDK as this path is used to cache scope and events between application restarts.
 


docs/platforms/javascript/guides/electron/index.mdx
```diff
@@ -31,23 +41,41 @@ yarn add @sentry/electron
...
+  _experiments: { enableLogs: true },
+  // ___PRODUCT_OPTION_END___ logs
 });
 ```
 
-In the Electron `renderer` process:
+### Configure the Renderer Process
 
-```javascript
+Initialize the SDK in your Electron renderer process. You can do this in either a preload script (`preload.(js|ts)`) or directly in your renderer script (`renderer.(js|ts)`):
```
timfish: With `contextIsolation: true`, you cannot configure in _either_ preload/renderer, you would need to configure the SDK in both. However see below for more explanation around this...

Basically, just instruct users to init in the renderer(s)
```
Initialize the SDK in your Electron renderer processes:
```

I wouldn't specifically mention file names (ie. `renderer.(js|ts)`) because these file names can vary depending on framework or whatever template the app was created from.

docs/platforms/javascript/guides/electron/index.mdx
```diff
@@ -31,23 +41,41 @@ yarn add @sentry/electron
...
+
+<Expandable title="Should you use a preload script or renderer script?">
+
+Choose based on your app's setup:
+
```
timfish: If `contextIsolation: true` (Electron default/recommended), the preload and renderer code run in different isolated contexts. This means _technically_, if you want to catch errors in both preload and renderer contexts, you should init the SDK in both. 

If `contextIsolation: false`, the preload and renderer code run in the same context so you could _technically_ just init in one of preload/renderer and it would capture errors from both.

However, many users starting out with Electron won't even know what a preload script is. Preload scripts need to be bundled so helping users configure them is non-trivial. We don't actually need the preload to be configured for the SDK to work and we don't want the SDK setup to vary by Electron configuration. 

Regardless of Electron configuration, the most reliable way to have the SDK just work in the majority of cases is to init in the main and init in all renderers and just ignore the preload entirely. 

Preload configuration is more of an advanced topic and won't be needed by most users.

docs/platforms/javascript/guides/electron/index.mdx
```diff
@@ -136,19 +146,19 @@ import { init as reactInit } from "@sentry/react";
...
+      /* integrations */
+    ],
     /* Other Electron and React SDK config */
   },
   reactInit
 );
 ```
 
-## Offline Support
+## Step 3: Offline Support (Optional)
```
timfish: I wouldn't even mention `Offline Support` in the basic intro docs.

I also wouldn't say that this is `(Optional)` either because it's enabled by default and most users don't need to worry about it.

timfish (changes requested):
Just a few comments around setup and preload.

Basically we don't want the SDK setup to vary by Electron config because it's not required and too complex.

docs/platforms/javascript/guides/electron/index.mdx
inventarSarah: 👍I agree -- let's do that in a separate PR, since I think there are a few other sections in here we can move to other pages (see the PR description).

For now, I removed the "(Optional)" for this PR.

docs/platforms/javascript/guides/electron/index.mdx
inventarSarah: Thanks for the detailed explanation!
I've now removed all content about the preload script, so that we're only writing about the renderer processes.

Do you think it would still be helpful to mention the preload script, as we do in the current guide? Something like:

> **Note**
>If your app uses a preload script and has `contextIsolation: true`, errors that occur in the preload context aren't captured automatically. To include those, call `Sentry.init()` in your preload script as well.

docs/platforms/javascript/guides/electron/index.mdx
inventarSarah: 👍 I've updated the description as you suggested and also removed the file names across the guide

docs/platforms/javascript/guides/electron/index.mdx
timfish: Yep good idea.

In this case you would call the renderer init specifically:
```ts
import * as Sentry from '@sentry/electron/renderer';

Sentry.init();
```

timfish (approved):

coolguyzone:
@inventarSarah This isn't a blocking issue, but I noticed that steps 4-7 are optional, I think this might be less daunting if we either collapse the optional steps into collapsibles or perhaps mention at the end of step 3 that the next steps are optional and offer a link to go straight to the verify step. What do you think?

inventarSarah:
> @inventarSarah This isn't a blocking issue, but I noticed that steps 4-7 are optional, I think this might be less daunting if we either collapse the optional steps into collapsibles or perhaps mention at the end of step 3 that the next steps are optional and offer a link to go straight to the verify step. What do you think?

@coolguyzone The plan is to clean up these steps (except source maps) in a new PR (will start working on it today). Ideally, I'd like to move the content to other pages within the Electron docs, since most of it is already too advanced for a quick-start guide. I will add you as a reviewer to this pr so you can also take a look 👍

PR #15317, opened 2025-10-25
feat(Vercel) Build cache improvements

sergical:
## DESCRIBE YOUR PR
* Caches release registry
* looking into md exports

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

## LEGAL BOILERPLATE

## EXTRA RESOURCES

### DIFF
--- a/scripts/generate-md-exports.mjs
+++ b/scripts/generate-md-exports.mjs
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+// testing cache
 /* eslint-disable no-console */
 import {ListObjectsV2Command, PutObjectCommand, S3Client} from '@aws-sdk/client-s3';
 import imgLinks from '@pondorasti/remark-img-links';


src/docTree.ts
```diff
@@ -39,9 +39,11 @@ export function getDocsRootNode(): Promise<DocNode> {
 }
 
 async function getDocsRootNodeUncached(): Promise<DocNode> {
-  return frontmatterToTree(
-    await (isDeveloperDocs ? getDevDocsFrontMatter() : getDocsFrontMatter())
-  );
+  const frontmatter = await (isDeveloperDocs
+    ? getDevDocsFrontMatter()
+    : getDocsFrontMatter());
+  const tree = frontmatterToTree(frontmatter);
```
BYK: You really don't need the tree variable here 😅

src/mdx.ts
```diff
@@ -65,6 +65,55 @@ if (process.env.CI) {
...
+  registryHits: 0,
+  registryMisses: 0,
+  uniqueRegistryFiles: new Set<string>(),
+};
+
+// Log summary periodically and at end
+let lastSummaryLog = Date.now();
+function logCacheSummary(force = false) {
+  const now = Date.now();
+  // Log every 30 seconds or when forced
```
BYK: This logic seems unnecessary? Why not just emit at the end?

src/mdx.ts
```diff
@@ -531,22 +582,37 @@ export async function getFileBySlug(slug: string): Promise<SlugFile> {
   const outdir = path.join(root, 'public', 'mdx-images');
   await mkdir(outdir, {recursive: true});
 
-  // If the file contains content that depends on the Release Registry (such as an SDK's latest version), avoid using the cache for that file, i.e. always rebuild it.
-  // This is because the content from the registry might have changed since the last time the file was cached.
-  // If a new component that injects content from the registry is introduced, it should be added to the patterns below.
-  const skipCache =
+  // Check if file depends on Release Registry
+  const dependsOnRegistry =
     source.includes('@inject') ||
```
BYK, line 612: Not sure if this `@inject` thing was related to the registry

src/mdx.ts
```diff
@@ -65,6 +65,55 @@ if (process.env.CI) {
   mkdirSync(CACHE_DIR, {recursive: true});
 }
 
+// Cache registry hash per worker to avoid recomputing for every file
+let cachedRegistryHash: string | null = null;
+async function getRegistryHash(): Promise<string> {
+  if (cachedRegistryHash) {
+    return cachedRegistryHash;
+  }
+  const [apps, packages] = await Promise.all([getAppRegistry(), getPackageRegistry()]);
```
BYK: There's a race condition here: if you call this function 3 times back to back, it would make 3 separate calls. 

What you need for proper caching is to change the type of `cachedRegistryHash` to `Promise<string>`, and do:

```
cachedRegistryHash = Promise.all(...). then(([apps, packages]) => md5(...));
return cachedRegistryHash;
```

src/mdx.ts
```diff
@@ -531,22 +582,37 @@ export async function getFileBySlug(slug: string): Promise<SlugFile> {
...
+    const sourceHash = md5(source);
+
+    // For files that depend on registry, include registry version in cache key
+    // This prevents serving stale content when registry is updated
+    if (dependsOnRegistry) {
+      try {
+        // Get registry hash (cached per worker to avoid redundant fetches)
+        const registryHash = await getRegistryHash();
+        cacheKey = `${sourceHash}-${registryHash}`;
+      } catch (err) {
```
BYK: This logic can and probably should be improved: the only way this can throw an exception should be a network related issue. In that case, pages depending on the registry will also have a problem so the try-catch is redundant. It's also wasteful as if it raises an exception, that means it will raise an exception for every single page. 

I'd rather add a retry mechanism into the cache key function and don't handle the exception if the retried fail, halting the build as we need the registry connection for the build.

vercel.json
```diff
@@ -1,4 +1,6 @@
 {
+  "$schema": "https://openapi.vercel.sh/vercel.json",
+  "buildCommand": "sh vercel.sh",
```
BYK: Is there no way to avoid an additional shell script for this?

sergical: i tired to be more explicit with it, we can always return early in `generate-md-exports` depending on [`VERCEL_ENV`](https://vercel.com/docs/environment-variables/system-environment-variables#VERCEL_ENV). This was their recommended approach https://vercel.com/guides/dynamic-build-commands

BYK: I prefer that much more than this as it's less abstraction. Up to you though

src/docTree.ts
sergical: idk why this refactor even exists, it doesn't change anything, i'll revert

vercel.json
sergical: chatted offline, i wasn't doing the right thing here

PR #15339, opened 2025-10-29
feat(caching) Release registry caching

sergical:
### Problem

- Registry-dependent files (~200 files) were never cached, requiring recompilation on every build
- Cache directory grew unbounded to 950MB, causing Vercel build cache upload failures

### Solution

1. Registry-aware caching (src/mdx.ts)
  - Include registry hash in cache key for registry-dependent files: ${sourceHash}-${registryHash}
  - Fetch registry once per worker instead of per file (~200+ fetches → 6 fetches)
  - Files now cache when registry unchanged, invalidate automatically when registry updates

2. Aggressive cache cleanup
  - scripts/generate-md-exports.mjs: Delete unused cache files after every build
  - src/mdx.ts: Delete cache files not accessed in last 24 hours
  - Prevents unbounded growth (950MB → ~650MB max)

### Impact
  - Incremental builds: ~10s faster when editing 1 file
  - Full builds with warm cache: Expected 45-75s savings (200 files skip compilation)
  - Cache size: Stays at ~650MB instead of growing to 950MB and breaking Vercel cache upload

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

## LEGAL BOILERPLATE

## EXTRA RESOURCES

### DIFF
--- a/scripts/generate-md-exports.mjs
+++ b/scripts/generate-md-exports.mjs
@@ -18,6 +18,7 @@ import {
   createBrotliCompress,
   createBrotliDecompress,
 } from 'node:zlib';
+import pLimit from 'p-limit';
 import rehypeParse from 'rehype-parse';
 import rehypeRemark from 'rehype-remark';
 import remarkGfm from 'remark-gfm';
@@ -64,7 +65,7 @@ let globalUsedCacheFiles = null;
 function taskFinishHandler({id, success, failedTasks, usedCacheFiles}) {
   // Collect cache files used by this worker into the global set
   if (usedCacheFiles && globalUsedCacheFiles) {
-    console.log(`🔍 Worker[${id}]: returned ${usedCacheFiles.length} cache files.`);
+    console.log(`🔍 Worker[${id}]: returned ${usedCacheFiles.size} cache files.`);
     usedCacheFiles.forEach(file => globalUsedCacheFiles.add(file));
   } else {
     console.warn(
@@ -103,11 +104,12 @@ async function createWork() {
   const CACHE_DIR = path.join(root, '.next', 'cache', 'md-exports');
   console.log(`💰 Cache directory: ${CACHE_DIR}`);
   const noCache = !existsSync(CACHE_DIR);
+  let initialCacheFiles = [];
   if (noCache) {
     console.log(`ℹ️ No cache directory found, this will take a while...`);
     await mkdir(CACHE_DIR, {recursive: true});
   } else {
-    const initialCacheFiles = await readdir(CACHE_DIR);
+    initialCacheFiles = await readdir(CACHE_DIR);
     console.log(
       `📦 Cache directory has ${initialCacheFiles.length} files from previous build`
     );
@@ -216,27 +218,27 @@ async function createWork() {
   // Clean up unused cache files to prevent unbounded growth
   if (!noCache) {
     try {
-      const allFiles = await readdir(CACHE_DIR);
-      const filesToDelete = allFiles.filter(file => !globalUsedCacheFiles.has(file));
-      const overlaps = allFiles.filter(file => globalUsedCacheFiles.has(file));
+      const filesToDelete = initialCacheFiles.filter(
+        file => !globalUsedCacheFiles.has(file)
+      );
+      const overlaps = initialCacheFiles.filter(file => globalUsedCacheFiles.has(file));
 
       console.log(`📊 Cache tracking stats:`);
-      console.log(`   - Files in cache dir (after build): ${allFiles.length}`);
+      console.log(`   - Files in cache dir (after build): ${initialCacheFiles.length}`);
       console.log(`   - Files tracked as used: ${globalUsedCacheFiles.size}`);
       console.log(`   - Files that existed and were used: ${overlaps.length}`);
       console.log(`   - Files to delete (old/unused): ${filesToDelete.length}`);
       console.log(`   - Expected after cleanup: ${overlaps.length} files`);
 
       if (filesToDelete.length > 0) {
+        const limit = pLimit(50);
         await Promise.all(
-          filesToDelete.map(file => rm(path.join(CACHE_DIR, file), {force: true}))
+          filesToDelete.map(file =>
+            limit(() => rm(path.join(CACHE_DIR, file), {force: true}))
+          )
         );
         console.log(`🧹 Cleaned up ${filesToDelete.length} unused cache files`);
       }
-
-      // Verify cleanup worked
-      const remainingFiles = await readdir(CACHE_DIR);
-      console.log(`✅ Cache directory now has ${remainingFiles.length} files`);
     } catch (err) {
       console.warn('Failed to clean unused cache files:', err);
     }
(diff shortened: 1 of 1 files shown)


scripts/generate-md-exports.mjs
```diff
@@ -58,7 +58,15 @@ async function uploadToCFR2(s3Client, relativePath, data) {
   return;
 }
 
-function taskFinishHandler({id, success, failedTasks}) {
+function taskFinishHandler(
+  {id, success, failedTasks, usedCacheFiles},
+  allUsedCacheFiles
```
BYK: The way these parameters are used doesn't make much sense. We never seem to pass `usedCacheFiles` in the `data` object.

I think this logic needs some adjustments where `allUsedCacheFiles` is a global in the main/controller worker. It would also be probably more efficient to either use `union` on all these sets or just do the look up in order as looping over each file is not efficient.

scripts/generate-md-exports.mjs
```diff
@@ -175,28 +186,51 @@ async function createWork() {
...
+    try {
+      const allFiles = await readdir(CACHE_DIR);
+      let cleanedCount = 0;
+
+      for (const file of allFiles) {
+        if (!usedCacheFiles.has(file)) {
+          await rm(path.join(CACHE_DIR, file), {force: true});
+          cleanedCount++;
+        }
+      }
```
BYK: Although this is async, it is still deleting the files in series. A better approach would be to delete all these files in parallel with a `Promise.all()` call. That said if the number of files to delete is too much, it may overwhelm the system. If that ever happens we can use [p-throttle](https://github.com/sindresorhus/p-throttle) to limit max concurrency to 50 files or less.

scripts/generate-md-exports.mjs
```diff
@@ -307,7 +344,7 @@ async function genMDFromHTML(source, target, {cacheDir, noCache}) {
   return {cacheHit: false, data};
 }
 
-async function processTaskList({id, tasks, cacheDir, noCache}) {
+async function processTaskList({id, tasks, cacheDir, noCache, usedCacheFiles}) {
```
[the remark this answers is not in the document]
BYK: This looks legit?

src/mdx.ts
```diff
@@ -63,10 +63,104 @@ const CACHE_COMPRESS_LEVEL = 4;
 const CACHE_DIR = path.join(root, '.next', 'cache', 'mdx-bundler');
 if (process.env.CI) {
   mkdirSync(CACHE_DIR, {recursive: true});
+
+  // Clean up old cache files in background to prevent unbounded growth
+  // Delete any file not accessed in the last 24 hours (meaning it wasn't used in recent builds)
+  // This runs once per worker process and doesn't block the build
```
BYK: I think we should remove this whole section as the files' `atime` will be modified upon cache restoration. So it is very unlikely that this code will ever delete anything.

I'd say incorporate the same logic from md generation and track the files we actually used and delete everything else?

src/mdx.ts
```diff
@@ -63,10 +63,104 @@ const CACHE_COMPRESS_LEVEL = 4;
...
+): Promise<string> {
+  let lastError: Error | null = null;
+
+  for (let attempt = 0; attempt <= maxRetries; attempt++) {
+    try {
+      const [apps, packages] = await Promise.all([
+        getAppRegistry(),
+        getPackageRegistry(),
+      ]);
+      return md5(JSON.stringify({apps, packages}));
```
BYK, line 90: Just curious if we can keep the app and package hashes separate as apps don't change as much?

src/mdx.ts
```diff
@@ -63,10 +63,104 @@ const CACHE_COMPRESS_LEVEL = 4;
...
+ * Get the registry hash, using cached value if available.
+ * This ensures we only fetch the registry once per worker process.
+ */
+function getRegistryHash(): Promise<string> {
+  if (!cachedRegistryHash) {
+    // eslint-disable-next-line no-console
+    console.info('Fetching registry hash for the first time in this worker');
+    cachedRegistryHash = getRegistryHashWithRetry().catch(err => {
+      // Reset cache on error so next call will retry
+      cachedRegistryHash = null;
```
BYK: This is not something we want?

src/mdx.ts
sergical: wed need to know which files depend on which registries, which is a bit more complex, no?

src/mdx.ts
sergical: +1, if registry fetching fails, we should fail once and skip caching for all files, not hammer the registry endpoint

src/mdx.ts
BYK: Fair, dropping.

scripts/generate-md-exports.mjs
```diff
@@ -190,13 +213,42 @@ async function createWork() {
 
   await Promise.all(workerPromises);
 
+  // Clean up unused cache files to prevent unbounded growth
+  if (!noCache) {
+    try {
+      const allFiles = await readdir(CACHE_DIR);
+      const filesToDelete = allFiles.filter(file => !globalUsedCacheFiles.has(file));
+      const overlaps = allFiles.filter(file => globalUsedCacheFiles.has(file));
```
BYK: You already read `initialCacheFiles` so all you need is (as a set operation) `initialCacheFiles - globalUsedCacheFiles`. No need to read the directory again.

scripts/generate-md-exports.mjs
```diff
@@ -190,13 +213,42 @@ async function createWork() {
...
+      console.log(`   - Files in cache dir (after build): ${allFiles.length}`);
+      console.log(`   - Files tracked as used: ${globalUsedCacheFiles.size}`);
+      console.log(`   - Files that existed and were used: ${overlaps.length}`);
+      console.log(`   - Files to delete (old/unused): ${filesToDelete.length}`);
+      console.log(`   - Expected after cleanup: ${overlaps.length} files`);
+
+      if (filesToDelete.length > 0) {
+        await Promise.all(
+          filesToDelete.map(file => rm(path.join(CACHE_DIR, file), {force: true}))
+        );
```
BYK, line 239: 🚀 

That said note that this _will_ overwhelm I/O if too many files are being deleted. I remember using a promise throttle library for mdx building so I'd use that and limit concurrency to 50.

scripts/generate-md-exports.mjs
```diff
@@ -190,13 +213,42 @@ async function createWork() {
...
+        console.log(`🧹 Cleaned up ${filesToDelete.length} unused cache files`);
+      }
+
+      // Verify cleanup worked
+      const remainingFiles = await readdir(CACHE_DIR);
+      console.log(`✅ Cache directory now has ${remainingFiles.length} files`);
```
BYK: Please no. Directory walking is very expensive. This is okay when you are testing locally but should not be done for each run :)

scripts/generate-md-exports.mjs
```diff
@@ -357,6 +430,7 @@ async function processTaskList({id, tasks, cacheDir, noCache}) {
     id,
     success,
     failedTasks,
+    usedCacheFiles: Array.from(usedCacheFiles),
```
BYK: Can we not pass `Set` objects directly?

sergical: done

PR #15381, opened 2025-11-03
feat(develop-docs): add Backend Telemetry Buffer

giortzisg:
## DESCRIBE YOUR PR
Add the docs for the backend telemetry buffer

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

### DIFF
--- a/develop-docs/sdk/telemetry/telemetry-buffer/backend-telemetry-buffer.mdx
+++ b/develop-docs/sdk/telemetry/telemetry-buffer/backend-telemetry-buffer.mdx
@@ -39,7 +39,7 @@ Introduce a `Buffer` layer between the `Client` and the `Transport`. This `Buffe
 │                                                                            │
 │   ┌──────────────────────┐  ┌──────────────────────┐  ┌──────────────────┐ │
 │   │  Error Store         │  │  Check-in Store      │  │  Log Store       │ │
-│   │  (CRITICAL)          │  │  (HIGH)              │  │  (MEDIUM)        │ │
+│   │  (CRITICAL)          │  │  (HIGH)              │  │  (LOW)           │ │
 │   │  Timeout: N/A        │  │  Timeout: N/A        │  │  Timeout: 5s     │ │
 │   │  BatchSize: 1        │  │  BatchSize: 1        │  │  BatchSize: 100  │ │
 │   └──────────────────────┘  └──────────────────────┘  └──────────────────┘ │


develop-docs/sdk/telemetry/telemetry-buffer/backend-telemetry-buffer.mdx
```diff
@@ -4,4 +4,288 @@ description: Detailed backend telemetry buffer design.
...
+
+### Architecture Overview
+
+Introduce a `Buffer` layer between the `Client` and the `Transport`. This `Buffer` wraps prioritization and scheduling and exposes a minimal API to the SDK:
+
+- Add(item).
+- Flush(timeout).
+- Close(timeout).
+
+```
```
sfanahata, line 29: Image

Text is being cut off in this diagram

develop-docs/sdk/telemetry/telemetry-buffer/backend-telemetry-buffer.mdx
```diff
@@ -4,4 +4,288 @@ description: Detailed backend telemetry buffer design.
...
+┌────────────────────────────────────────────────────────────────────────────┐
+│                             Transport                                      │
+│   - Single worker, disk cache, offline retry, client reports               │
+└────────────────────────────────────────────────────────────────────────────┘
+```
+
+#### How the Buffer works
+
+- **Smart batching**: Logs are batched into single requests; errors, transactions, and monitors are sent immediately.
+- **Pre-send rate limiting**: The scheduler checks rate limits before dispatching, avoiding unnecessary requests while keeping items buffered.
```
cleptric: We should clarify what "while keeping items buffered" means. Today, we drop data if are rate-limited, is the plan to keep them in the buffer?

develop-docs/sdk/telemetry/telemetry-buffer/backend-telemetry-buffer.mdx
```diff
@@ -4,4 +4,288 @@ description: Detailed backend telemetry buffer design.
...
+- **Graceful shutdown**: During client shutdown, force-drains all buffers to prevent data loss.
+
+#### Transport
+
+The transport layer handles HTTP communication with Sentry's ingestion endpoints:
+
+### Configuration
+
+#### Buffer Options
+- **Capacity**: 100 items for errors and check-ins, 10*BATCH_SIZE for logs, 1000 for transactions.
```
cleptric, line 153: Do we have multiple buffers per telemetry type or just one that changes its capacity based on which telemetry is configured to be emitted by the SDK?

develop-docs/sdk/telemetry/telemetry-buffer/backend-telemetry-buffer.mdx
```diff
@@ -4,4 +4,288 @@ description: Detailed backend telemetry buffer design.
...
+    }
+}
+```
+
+
+#### Flushing
+
+```go
+func (s *Scheduler) flush() {
+  // should process all store buffers and send to transport
```
cleptric, line 315:
```suggestion
   // should process all store buffers and send to transport
```

develop-docs/sdk/telemetry/telemetry-buffer/backend-telemetry-buffer.mdx
giortzisg: changed back from the `mermaid` chart to just an ascii chart when I saw the cut off, so should be ok now.

develop-docs/sdk/telemetry/telemetry-buffer/backend-telemetry-buffer.mdx
giortzisg: We should drop, to avoid filling up the buffer.

develop-docs/sdk/telemetry/telemetry-buffer/backend-telemetry-buffer.mdx
```diff
@@ -4,4 +4,300 @@ description: Detailed backend telemetry buffer design.
...
+- **Pre-send rate limiting**: The scheduler checks rate limits before serialization to avoid unnecessary processing. When a telemetry is rate-limited the selected batch should
+be dropped, to avoid filling up the buffers.
+- **Category isolation**: Separate ring buffers for each telemetry type prevent head-of-line blocking.
+- **Weighted scheduling**: High-priority telemetry gets sent more frequently via round-robin selection.
+- **Transport compatibility**: Works with existing HTTP transport implementations without modification.
+
+### Priorities
+- CRITICAL: Error, Feedback.
+- HIGH: Session, CheckIn.
+- MEDIUM: Transaction, ClientReport, Span.
```
lcian, line 76: Not sure if this is already set in stone, but I feel that logs should be at least as important as spans, thinking about it from a debuggability point of view.

develop-docs/sdk/telemetry/telemetry-buffer/backend-telemetry-buffer.mdx
```diff
@@ -4,4 +4,300 @@ description: Detailed backend telemetry buffer design.
...
+### Components
+
+#### Storage
+
+Each telemetry category maintains a store interface; a fixed-size circular array/ring buffer (not to be confused with the `Buffer` wrapper) that stores items before transmission:
+
+- **Bounded capacity**: Default to 100 items for errors, logs, and monitors; 1000 for transactions. This prevents unbounded memory growth regardless of telemetry volume and backpressure handling.
+- **Overflow policies**:
+  - `drop_oldest` (default): Evicts the oldest item when the buffer is full, making room for new data.
+  - `drop_newest`: Rejects incoming items when full, preserving what's already queued.
```
lcian, line 91: Maybe this can be made optional as I don't foresee a need to use this one.

develop-docs/sdk/telemetry/telemetry-buffer/backend-telemetry-buffer.mdx
```diff
@@ -4,4 +4,300 @@ description: Detailed backend telemetry buffer design.
...
+- **Batching configuration**:
+  - `batchSize`: Number of items to combine into a single batch (1 for errors, transactions, and monitors; 100 for logs).
+  - `timeout`: Maximum time to wait before sending a partial batch (5 seconds for logs).
+- **Bucketed Storage Support**: The storage interface should satisfy both bucketed and single-item implementations, allowing sending spans per trace id (required for Span First).
+- **Observability**: Each store tracks dropped item counts for client reports.
```
lcian: Does it just track them and then some time later create the client report or does it generate the client report immediately on each dropped envelope/item?

develop-docs/sdk/telemetry/telemetry-buffer/backend-telemetry-buffer.mdx
giortzisg: since logs are batched I don't think it makes lots of sense to prioritize them with the same importance. Also the main point of the design was to avoid starvation of other types by the overabundance of logs.

develop-docs/sdk/telemetry/telemetry-buffer/backend-telemetry-buffer.mdx
giortzisg: Actually this is a nice catch. The correct way to do this would be to add to the ClientReports Buffer and then send the reports when the scheduler selects the priority. I am planning to update both this and the client reports page when we implement Client Reports in Go, so will remove the line for now.

PR #15428, opened 2025-11-07
feat: Add "Open in ChatGPT/Claude" buttons for each page

nikolovlazar:
Besides the existing "Copy Page" and "View as Markdown", I've added two more buttons in the dropdown so users can open that page in ChatGPT or Claude and ask questions about it:

image

Both buttons are static links that open ChatGPT / Claude with predefined prompt.

Example: https://chatgpt.com/?hint=search&q=Read+from+https://docs.sentry.io/platforms/javascript/+so+I+can+ask+questions+about+its+contents


src/components/copyMarkdownButton.tsx
```diff
@@ -204,6 +207,44 @@ export function CopyMarkdownButton({pathname}: CopyMarkdownButtonProps) {
                   </div>
                 </div>
               </Link>
+
+              <a
```
sergical, line 221: not sure if using Link here is better since it's an external link

src/icons/chatgpt.tsx
```diff
@@ -0,0 +1,17 @@
+function ChatGPT({width = 16, height = 16, ...props}: React.SVGAttributes<SVGElement>) {
+  return (
+    <svg
+      fill="currentColor"
+      fillRule="evenodd"
+      height={height}
+      viewBox="0 0 24 24"
+      width={width}
+      xmlns="http://www.w3.org/2000/svg"
+      className="w-4 h-4"
```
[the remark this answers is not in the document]
KyleTryon: im fairly sure its the other way around? heigh and width props should override classes.

src/components/copyMarkdownButton.tsx
nikolovlazar: That's actually why I used `a` instead of `Link`. Isn't Next.js's `Link` for soft nav within the app?

src/icons/chatgpt.tsx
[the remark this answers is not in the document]
nikolovlazar: Redundant. I just removed the classname and it still looked fine.

src/components/copyMarkdownButton.tsx
sergical: it can be used for external links too as it'll render an anchor tag anyway, was just thinking since Link was already imported in the file, just for standardization but leave it as is :)

nikolovlazar: No I know. Here's what my thought was:
- we def have an external link and that's not going to change;
- we don't want the current behaviour to change as well (with the `a` tag);
- Next.js's Link component probably checks in with the router before navigating (extra steps for external navigation) and it has a preloading mechanism built in for perf optimization on soft navs. It's definitely built for local soft navigation;
- Next.js's Link component is third-party and prone to change;

The `a` approach seemed like the cleanest.

sergical: 100% aligned, great call

nikolovlazar:
@sentry review

src/components/copyMarkdownButton.tsx
```diff
@@ -204,6 +207,44 @@ export function CopyMarkdownButton({pathname}: CopyMarkdownButtonProps) {
...
+                </div>
+              </a>
+
+              <a
+                href={`https://claude.ai/new?q=${encodeURIComponent('Read from ' + window.location.href + ' so I can ask questions about its contents')}`}
+                target="_blank"
+                rel="noopener noreferrer"
+                className={`${dropdownItemClass} no-underline`}
+              >
+                <div className={iconContainerClass}>
```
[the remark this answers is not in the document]
nikolovlazar: @sentry no need in this case. The `copyMarkdownButton` component is marked with `'use client'` so it will always run in the browser. `window` is always going to have a value, so it is safe to use `window.location.href` directly in JSX.

src/components/copyMarkdownButton.tsx
```diff
@@ -204,6 +217,46 @@ export function CopyMarkdownButton({pathname}: CopyMarkdownButtonProps) {
                   </div>
                 </div>
               </Link>
+
+              <a
+                href={`https://chatgpt.com/?hints=search&q=${encodeURIComponent('Read from ' + window.location.href + ' so I can ask questions about its contents')}`}
```
nikolovlazar:
```suggestion
                href={`https://chatgpt.com/?hint=search&q=${encodeURIComponent('Read from ' + window.location.href + ' so I can ask questions about its contents')}`}
```

src/icons/external-link.tsx
```diff
@@ -0,0 +1,23 @@
...
+  height = 14,
+  ...props
+}: React.SVGAttributes<SVGElement>) {
+  return (
+    <span className="icon icon-external-link">
+      <svg
+        viewBox="0 0 24 24"
+        width={width}
+        height={height}
+        className="ml-1 inline"
```
sergical, line 12: id control this as props in case we want to use the external link as the first item

sergical (approved):

KyleTryon (approved):

PR #15440, opened 2025-11-10
feat: JavaScript Docs display order, tags, split-layout

codyde:
- Implementing new / beta tags for Logs + Metrics + OpenTelemetry  
- Adjusting ordering of content to reflect items that are higher priority for utilization / engagement 
- Utilize sidebar separator to breakup content 
- Implemement split layout for nextjs quickstart (text vs code)

Note: This change will create a difference in layout outside of the JavaScript SDKs. I will be fast following updates to the layout for other SDKs, as well as fast following the pill icons to the other SDKs as well for Logs / Metrics - i dont want to overload this PR with all of those changes. 

CleanShot 2025-11-10 at 9  59 37@2x


coolguyzone:
I think we may want to re-evaluate the breakpoints for stacking the two columns, if you look at it at 1040px width, the code sample becomes trick to read/navigate.

Screenshot 2025-11-17 at 11 21 48 AM

PR #15493, opened 2025-11-12
feat(dev): App reference Spotlight specs for SDKs

BYK:
Fixes getsentry/spotlight#545

### DIFF
--- a/develop-docs/sdk/expected-features/spotlight.mdx
+++ b/develop-docs/sdk/expected-features/spotlight.mdx
@@ -132,7 +132,7 @@ The interaction between configuration options and environment variables MUST fol
 3. **If using two-attribute approach:**
    - If `spotlightUrl` config and env var are both set → MUST use config value (`spotlightUrl`)
    - If `spotlight: false` is explicitly set → MUST ignore `spotlightUrl` value, the env var, and any URL configuration
-   - In either case the SDK MUST print a warning to the console explaining the reason for deactivation of Spotlight
+   - In either case the SDK MUST print a warning to the console explaining the reason for deactivation of Spotlight or that the env variable exists but not being used due to hard-coded configuration.
 
 **Precedence Examples:**
 


develop-docs/sdk/expected-features/spotlight.mdx
```diff
@@ -0,0 +1,414 @@
...
+
+### Single Attribute Approach (Recommended)
+
+This approach MUST be used if the SDK language allows for a single configuration attribute to have 2 different types (`boolean` and `string`).
+
+The SDK accepts a single `spotlight` configuration attribute that can be:
+
+- `false` or `undefined`/`null`: Spotlight is disabled
+- `true`: Spotlight is enabled using the default URL (`http://localhost:8969/stream`)
+- `string`: Spotlight is enabled using the provided URL
```
betegon, line 22: how will this work when someone runs `spotlight run` and it chooses a dynamic port?

BYK: See below? If this is explicitly set, we assume it's for a reason and ignore the env variabl.

develop-docs/sdk/expected-features/spotlight.mdx
```diff
@@ -0,0 +1,414 @@
...
+                self.error_logged = True
+            
+            # Wait before retry
+            time.sleep(self.retry_delay)
+            self.retry_delay = min(self.retry_delay * 2, self.max_retry_delay)
+            
+            # Retry once, then give up for this envelope
+            try:
+                self._send_envelope(envelope)
+                self.retry_delay = 1.0
```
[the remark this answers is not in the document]
BYK: This is intentional

develop-docs/sdk/expected-features/spotlight.mdx
```diff
@@ -0,0 +1,414 @@
+---
+title: Spotlight
+sidebar_order: 5
+---
+
+[Sentry Spotlight](https://spotlightjs.com/) is a local development tool that provides real-time observability for errors, traces, logs, and performance data during development. SDKs should implement support for Spotlight to unlock the power of Sentry for local development.
+
+## Configuration Options
+
+SDKs should support Spotlight configuration through one of two approaches:
```
philippsuess: This document already uses uppercase for a couple of occurrences or "MUST", "SHOULD", and "MAY". So it seems like we're following: https://www.ietf.org/rfc/rfc2119.txt. Therefore, I would write all occurrences uppercase
```suggestion
SDKs SHOULD support Spotlight configuration through one of two approaches:
```

develop-docs/sdk/expected-features/spotlight.mdx
```diff
@@ -0,0 +1,414 @@
...
+spotlightUrl: "http://localhost:3000/stream"
+
+# Explicitly disabled even with spotlightUrl set
+spotlight: False
+spotlightUrl: "http://localhost:3000/stream"  # Ignored
+```
+
+## Environment Variable Handling
+
+SDKs must support the `SENTRY_SPOTLIGHT` environment variable. The value should be parsed according to the following rules:
```
philippsuess: We don't have that yet on Cocoa. Environment variables are not that common for mobile platforms. I would change this to a SHOULD or MAY
```suggestion
SDKs SHOULD support the `SENTRY_SPOTLIGHT` environment variable. The value should be parsed according to the following rules:
```

develop-docs/sdk/expected-features/spotlight.mdx
```diff
@@ -0,0 +1,414 @@
...
+### Envelope Transmission
+
+- Send a copy of **every envelope** to the Spotlight server
+- This includes errors, transactions, sessions, profiles, replays, and all other envelope types
+- The Spotlight server HTTP semantics are the same as the Sentry server HTTP semantics (e.g. use POST to send envelopes)
+
+### Sampling
+
+- Enable **100% sample rate** for the Spotlight pipeline
+- This should **not affect** the upstream Sentry sample rates
```
philippsuess: This really must not impact the other sample rates, or do we have any uses cases or SDKs for which that is acceptable?
```suggestion
- This **MUST NOT affect** the upstream Sentry sample rates
```

develop-docs/sdk/expected-features/spotlight.mdx
```diff
@@ -0,0 +1,414 @@
...
+- The ideal implementation uses a **separate pipeline** that ignores sampling settings
+- Fallback: If separate pipeline is not feasible, enable sampling manually if no DSN is configured in development mode
+
+### PII Data Collection
+
+- Enable **all PII data collection** for Spotlight (equivalent to `sendDefaultPii: true`)
+- This should **not affect** upstream Sentry PII settings
+- Spotlight is intended for local development, so full data visibility is expected
```
philippsuess, line 183: This might be a bit tricky depending on the features sending PII and it could also be a bit risky. Is this a hard requirement for all SDKs? I think in most SDKs, we have two transports forwarding data to both spotlight and Sentry. All the PII collecting stuff happens before the envelope ends up in the transport.

develop-docs/sdk/expected-features/spotlight.mdx
```diff
@@ -0,0 +1,414 @@
...
+
+### Envelope Transmission
+
+- Send a copy of **every envelope** to the Spotlight server
+- This includes errors, transactions, sessions, profiles, replays, and all other envelope types
+- The Spotlight server HTTP semantics are the same as the Sentry server HTTP semantics (e.g. use POST to send envelopes)
+
+### Sampling
+
+- Enable **100% sample rate** for the Spotlight pipeline
```
philippsuess: I'm pretty sure some SDKs don't meet that requirement. For example, on Cocoa, we have two transports. One sends to Sentry, the other to Spotlight. Sampling happens before the transport. So, currently, the sampling behavior is the same for both Spotlight and Sentry envelopes. I'm unsure if having two different sampling behaviours is actually worth the complexity. This could be tricky to set up in some SDKs. It could also confuse users.

develop-docs/sdk/expected-features/spotlight.mdx
```diff
@@ -0,0 +1,414 @@
...
+```
+
+This URL should be used when:
+- `spotlight: true` is set (boolean, no URL specified)
+- `SENTRY_SPOTLIGHT` is set to a truthy value (not a URL string)
+
+### Default Behavior
+
+- Spotlight is **disabled by default**
+- SDKs should only enable Spotlight when explicitly configured or when the environment variable is set
```
philippsuess:
```suggestion
- SDKs MUST only enable Spotlight when explicitly configured or when the environment variable is set.
```

develop-docs/sdk/expected-features/spotlight.mdx
```diff
@@ -0,0 +1,414 @@
...
+
+## Error Handling
+
+SDKs must handle Spotlight server connectivity issues gracefully:
+
+### Unreachable Server
+
+- If the Spotlight server is unreachable, SDKs should:
+  - Log an error message (ideally once, not per-envelope)
+  - Implement exponential backoff retry logic
```
philippsuess: What drove you to add this? I'm unsure if it's worth adding this. Did we have any problems without the backoff retry logic? If yes, I would explain it a bit here, and clearly state it's optional if possible. I don't think we need this for example on mobile or desktop apps.

develop-docs/sdk/expected-features/spotlight.mdx
BYK: Unfortunately, no. Implementation of `SENTRY_SPOTLIGHT` is non-negotiable. How can we help support this in Cocoa?

develop-docs/sdk/expected-features/spotlight.mdx
BYK: This document is defining the ideal state we should be in, not what the current state of the SDKs are. We will need to try evolving the SDKs towards these goals.

develop-docs/sdk/expected-features/spotlight.mdx
BYK: We have the same issue with Python SDK and other SDKs. Ideally, we'd evolve the SDKs so we collect all the data upfront and then do some scrubbing before sending the envelope, allowing us to send the full data to Spotlight.

develop-docs/sdk/expected-features/spotlight.mdx
BYK: What's your concern here?

develop-docs/sdk/expected-features/spotlight.mdx
cleptric: Exposing local env vars into the Xcode simulator would still require a developer to do this manually. So mobile SDKs have other constrains that make this work differently than e.g. running a Flask app on your machine.

philippsuess: > Implementation of SENTRY_SPOTLIGHT is non-negotiable.

Please elaborate on the reasons.

It's not uncommon to use env vars on mobile in combination with simulators. We already do that for disabling spotlight in our UI tests here https://github.com/getsentry/sentry-cocoa/blob/aa43aea663aa0be90964af89e670f18dc22dc146/Samples/iOS-Swift/iOS-Swift-UITests/BaseUITest.swift#L34-L45

Adding adds value, but I wouldn't give it the highest priority on mobile. If you say non-negiotable, I'd like to know the reasons.

develop-docs/sdk/expected-features/spotlight.mdx
philippsuess: Yes, makes sense, thanks for the context. Maybe it's worth adding this info here as well.

develop-docs/sdk/expected-features/spotlight.mdx
philippsuess: It's some work and complexity that seems not worth the effort on mobile or JS frontend. TLDR; We can add it, but it's work. If it's a SHOULD then I'm OK with it. The logging an error message is a MUST, IMO.

develop-docs/sdk/expected-features/spotlight.mdx
```diff
@@ -0,0 +1,413 @@
...
+- Spotlight MUST be **disabled by default** and MUST only be enabled when explicitly configured or when the environment variable is set.
+
+## Error Handling
+
+SDKs MUST handle Spotlight server connectivity issues gracefully:
+
+### Unreachable Server
+
+- If the Spotlight server is unreachable, SDKs:
+  - SHOULD log an error message (ideally once, not per-envelope)
```
philippsuess: That could be a MUST, IMO. If you don't log an error it will confuse users.

philipphofmann (approved):
Thanks you @BYK , LGTM

develop-docs/sdk/expected-features/spotlight.mdx
```diff
@@ -62,7 +62,14 @@ spotlightUrl: "http://localhost:3000/stream"  # Ignored
 
 ## Environment Variable Handling
 
-SDKs MUST support the `SENTRY_SPOTLIGHT` environment variable. The value MUST be parsed according to the following rules:
+SDKs MUST support the `SENTRY_SPOTLIGHT` environment variable. This is to enable `spotlight run` to be used as a development server,
+which is the preferred way to run Spotlight. In this mode Spotlight runs your development command for you, with a fresh, unique Spotlight
+server running on a random port. To enable Spotlight on the target application, it sets the `SENTRY_SPOTLIGHT` environment variable to the
+URL of the Spotlight server. It also does some accommodations based on the target application, such as altering the host value for Docker Compose
+while keeping the front-end ones the same. It will also support emulator-based development environments and will set the host value accordingly in
+the near future.
```
philippsuess, line 70: Thanks for the context.

betegon (approved):

develop-docs/sdk/expected-features/spotlight.mdx
```diff
@@ -0,0 +1,424 @@
...
+   - If `spotlight: true` (boolean, no URL specified) AND `SENTRY_SPOTLIGHT` is set to a URL string → MUST use the env var URL
+     This allows developers to enable Spotlight via config but override the URL via environment variable
+
+2. **If `spotlight` is set to a string URL** → it MUST override the env var completely
+   When this happens, the SDK MUST print a warning to the console indicating that the config URL is taking precedence over the env var
+
+3. **If using two-attribute approach:**
+   - If `spotlightUrl` config and env var are both set → MUST use config value (`spotlightUrl`)
+   - If `spotlight: false` is explicitly set → MUST ignore `spotlightUrl` value, the env var, and any URL configuration
+   - In either case the SDK MUST print a warning to the console explaining the reason for deactivation of Spotlight
```
BYK:
```suggestion
   - In either case the SDK MUST print a warning to the console explaining the reason for deactivation of Spotlight or that the env variable exists but not being used due to hard-coded configuration.
```

PR #15520, opened 2025-11-14
docs(nextjs): Add vercel specific microfrontends guide

logaretm:
Adds micro frontends guide specific to [vercel's microfrontends feature](https://vercel.com/docs/microfrontends) since it has multi-zone application support which doesn't require the use of module federation.

### DIFF
--- a/docs/platforms/javascript/common/best-practices/micro-frontends.mdx
+++ b/docs/platforms/javascript/common/best-practices/micro-frontends.mdx
@@ -37,7 +37,7 @@ Below you'll find setup instructions for both an automatic and a manual way to r
 
 ## Vercel Multi-Zone Micro Frontends
 
-Vercel's Multi-zone applications let you slice a large app into smaller, independently managed projects. Each project operates as its own application, with its own UI and tech stack. For example, you might use Next.js for one project, SvelteKit for another, and React for a third.
+Vercel's multi-zone applications let you slice a large app into smaller, independently managed projects. Each project operates as its own application, with its own UI and tech stack. For example, you might use Next.js for one project, SvelteKit for another, and React for a third.
 
 To set up Sentry in a multi-zone setup, configure Sentry separately within each project. Follow the installation steps for the framework each project uses. Here are a few examples:
 


docs/platforms/javascript/guides/nextjs/best-practices/micro-frontends.mdx
```diff
@@ -0,0 +1,39 @@
+---
+title: Micro Frontends
+description: "Learn how to add Sentry instrumentation to your Vercel's micro frontend projects."
+---
+
+<PlatformContent includePath="getting-started-prerequisites" />
+
+You also need a [micro frontends project](https://vercel.com/docs/microfrontends) deployed to Vercel.
```
inventarSarah:
```suggestion
You also need a [micro frontend project](https://vercel.com/docs/microfrontends) deployed to Vercel.
```

docs/platforms/javascript/guides/nextjs/best-practices/micro-frontends.mdx
```diff
@@ -0,0 +1,39 @@
+---
+title: Micro Frontends
+description: "Learn how to add Sentry instrumentation to your Vercel's micro frontend projects."
```
inventarSarah: is this guide only for Vercel? If so, do we need the detailed content on module federation. As I understand it, Vercel only supports mult-zone apps - please lmk :)

inventarSarah (commented):
Hey 👋
I haven't reviewed everything yet, but I have a question about the scope of the guide that I need answered before continuing (see my comment). 
Thanks!

docs/platforms/javascript/guides/nextjs/best-practices/micro-frontends.mdx
logaretm: Yep, I tried setting one up and it worked, although they don't document it:

https://github.com/vercel-labs/microfrontends-single-spa-module-federation/

It's not in their docs or listed anywhere except their GitHub, so what do you think we should do here?

inventarSarah: I think we should stick to what's officially supported by Vercel. We can always update and revise if they support both approaches in the future.

I thought about your page a bit over the weekend, and I think the safest approach for users would be to keep the existing shared "Micro Frontends" page for Next.js (the one that is currently live) and add the Vercel multi-zone info as a new section rather than replacing the whole page.
That way, Vercel users and non-Vercel users can find everything in one place

My proposal:
Add [the original MF page ](https://docs.sentry.io/platforms/javascript/guides/nextjs/best-practices/micro-frontends/)back under nextjs and add a new section for Vercel multi-zone apps that only shows up for Next.js as the first subsection of the page. You can use the `PlatformSection` component for this:
```js
<PlatformSection supported={["javascript.nextjs"]}>
## Vercel Multi-Zone Micro Frontends
...
</PlatformSection>
```

In this new section, we can include a short intro to Vercel multi-zone apps and then reuse/adjust your existing “Multi-zone applications” content to explain how to set up Sentry in that environment.

---

Let me know if this makes sense!
And I’m more than happy to help with restructuring or integrating the Vercel content so the work you put in fits into the overall page. Just let me know!

chargome: +1 on adding this content to the existing page of microfrontends, you can keep a small note on module federation though (helps with search and module federation is well known).

logaretm: Yep makes sense! I will make the necessary changes.

EDIT: Done, is that what you folks had in mind?

docs/platforms/javascript/guides/nextjs/index.mdx
```diff
@@ -217,6 +217,7 @@ Our next recommended steps for you are:
 - Learn how to [manually capture errors](/platforms/javascript/guides/nextjs/usage/)
 - Continue to [customize your configuration](/platforms/javascript/guides/nextjs/configuration/)
 - Get familiar with [Sentry's product features](/product) like tracing, insights, and alerts
+- Learn how to [set up Sentry for Vercel's microfrontends](/platforms/javascript/guides/nextjs/best-practices/micro-frontends/)
```
inventarSarah:
```suggestion
- Learn how to [set up Sentry for Vercel's micro frontends](/platforms/javascript/guides/nextjs/best-practices/micro-frontends/)
```

docs/platforms/javascript/common/best-practices/micro-frontends.mdx
```diff
@@ -33,11 +33,39 @@ If your app uses micro frontends, it's very useful to be able to track which one
...
+- **[Next.js](/platforms/javascript/guides/nextjs/)**
+- **[React Router](/platforms/javascript/guides/react/)**
+- **[SvelteKit](/platforms/javascript/guides/sveltekit/)**
+
+We recommend using a different DSN for each project. This makes it easier to
+identify which project errors are coming from, helping you track issues and
+fix them faster in complex frontend architectures.
+
+<Alert>
+  With the exception of Vercel Multi-Zone Micro Frontends, you should call
```
inventarSarah:
```suggestion
  With the exception of Vercel multi-zone micro frontends, you should call
```

docs/platforms/javascript/common/best-practices/micro-frontends.mdx
```diff
@@ -33,11 +33,39 @@ If your app uses micro frontends, it's very useful to be able to track which one
...
+<PlatformSection supported={["javascript.nextjs"]}>
+
+## Vercel Multi-Zone Micro Frontends
+
+Vercel's Multi-zone applications let you slice a large app into smaller, independently managed projects. Each project operates as its own application, with its own UI and tech stack. For example, you might use Next.js for one project, SvelteKit for another, and React for a third.
+
+To set up Sentry in a multi-zone setup, configure Sentry separately within each project. Follow the installation steps for the framework each project uses. Here are a few examples:
+
+- **[Next.js](/platforms/javascript/guides/nextjs/)**
+- **[React Router](/platforms/javascript/guides/react/)**
```
inventarSarah:
```suggestion
- **[React](/platforms/javascript/guides/react/)**
```

Could be confusing since we have a React Router Framework SDK (?)

docs/platforms/javascript/common/best-practices/micro-frontends.mdx
```diff
@@ -33,11 +33,39 @@ If your app uses micro frontends, it's very useful to be able to track which one
 
 Below you'll find setup instructions for both an automatic and a manual way to route errors to different Sentry projects.
 
+<PlatformSection supported={["javascript.nextjs"]}>
+
+## Vercel Multi-Zone Micro Frontends
+
+Vercel's Multi-zone applications let you slice a large app into smaller, independently managed projects. Each project operates as its own application, with its own UI and tech stack. For example, you might use Next.js for one project, SvelteKit for another, and React for a third.
```
inventarSarah:
```suggestion
Vercel's multi-zone applications let you slice a large app into smaller, independently managed projects. Each project operates as its own application, with its own UI and tech stack. For example, you might use Next.js for one project, SvelteKit for another, and React for a third.
```

logaretm:
Thanks for the suggestions 🙏  I have applied them and will merge once the CI finishes.

PR #15546, opened 2025-11-18
ref(develop/scopes): Clarify unit type for attributes

Lms24:
For now we specify units on attributes as follows:
- accept all `MetricUnit`s (see [Relay](https://getsentry.github.io/relay/relay_metrics/enum.MetricUnit.html))
  - except for `''`, `'none'`
  - except for custom units (arbitrary strings). <-- we can revisit this if we want to widen the type here

We recognize that attribute units are therefore typed more strictly than e.g. units for metrics (they currently also accept custom units) but we can adjust both in either direction in the future.

### DIFF
--- a/develop-docs/sdk/telemetry/scopes.mdx
+++ b/develop-docs/sdk/telemetry/scopes.mdx
@@ -61,7 +61,7 @@ Users MUST be able to attach attributes to any scope using a dedicated method (e
 Attributes are key-value pairs where each value is either a an attribute value or an object containing:
 
 - `value`: The actual attribute value, which MUST match the specified type
-- `unit` (optional): The unit of measurement (e.g., `"millisecond"`, `"second"`, `"byte"`, `"ration"`).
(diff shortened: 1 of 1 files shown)


develop-docs/sdk/telemetry/scopes.mdx
```diff
@@ -61,7 +61,9 @@ Users MUST be able to attach attributes to any scope using a dedicated method (e
 Attributes are key-value pairs where each value is either a an attribute value or an object containing:
 
 - `value`: The actual attribute value, which MUST match the specified type
-- `unit` (optional): The unit of measurement (e.g., `"ms"`, `"s"`, `"bytes"`, `"count"`, `"percent"` or any other string value). SDKs MAY NOT add support for units for the moment, but MUST be able to do so at a later date without a breaking change.
+- `unit` (optional): The unit of measurement (e.g., `"millisecond"`, `"second"`, `"byte"`, `"ration"`).
```
Lms24: whoops
```suggestion
- `unit` (optional): The unit of measurement (e.g., `"millisecond"`, `"second"`, `"byte"`, `"ratio"`).
```

PR #15557, opened 2025-11-19
docs(spans): Create a dedicated attributes page

cleptric:


develop-docs/sdk/telemetry/attributes.mdx
```diff
@@ -0,0 +1,95 @@
...
+- `pebibyte`
+- `exabyte`
+- `exbibyte`
+
+### Fraction Units
+
+Defaults to `ratio`.
+
+- `ratio`
+- `percent`
```
buenaflor: when will it default to `milliseconds` / `byte` /`ratio`? can relay infer what type of unit it is if none is provided?

cleptric: Relay will default to `none` if no unit was provided. SDKs will not send it though, so you just omit the unit property.

develop-docs/sdk/telemetry/attributes.mdx
```diff
@@ -0,0 +1,95 @@
...
+}
+```
+
+### Properties
+
+| Property | Type | Required | Description |
+|----------|------|----------|-------------|
+| `type` | string | Yes | The data type of the attribute value. |
+| `value` | any | Yes | The actual attribute value, must match the specified type. |
+| `unit` | string | No | The unit of the attribute value. See [Units](#units) for a full list of supported units. |
```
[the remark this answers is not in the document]
cleptric: It's required in the protocol.

develop-docs/sdk/telemetry/attributes.mdx
```diff
@@ -0,0 +1,95 @@
...
+- `string[]`
+- `boolean[]`
+- `integer[]`
+- `double[]`
+
+**Note on Integers:** Integers should be 64-bit signed integers. For 64-bit unsigned integers, use the `string` type to avoid overflow issues until unsigned integers are natively supported.
+
+## Units
+
+The following units are supported by Relay.
```
Lms24: l: Good that we're writing them down but I'd still also link to Relay

```suggestion
The following units are [supported by Relay](https://getsentry.github.io/relay/relay_metrics/enum.MetricUnit.html).
```

develop-docs/sdk/telemetry/attributes.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+### Duration Units
+
+Defaults to `millisecond`.
```
Lms24: l: I'd probably omit the "defaults to" statements because SDKs can't default to any of these units (?)

develop-docs/sdk/telemetry/attributes.mdx
```diff
@@ -0,0 +1,95 @@
...
+The following types are supported:
+
+- `string`
+- `boolean`
+- `integer` (64-bit signed integer)
+- `double` (64-bit floating point number)
+- `string[]`
+- `boolean[]`
+- `integer[]`
+- `double[]`
```
Lms24, line 45: Reminder that we still have to figure out what should happen with mixed number arrays (integers and doubles).  Not a blocker though.

develop-docs/sdk/telemetry/attributes.mdx
cleptric: Ahh, good point!

cleptric:
I can add the API docs, but we have scopes, logs and metrics 😬

PR #15636, opened 2025-11-25
fix: Disable ISR to prevent ENOENT errors and reduce function invocations

sergical:
CleanShot 2025-11-25 at 18 43 00@2x

CleanShot 2025-11-25 at 18 44 20@2x

## Problem
ISR cache misses triggered runtime page regeneration, which called `getDocsFrontMatter()` attempting to access `docs/**/*`. Since `docs/**/*` is excluded from deployment to avoid the 250MB function limit, this caused ENOENT errors.

## Impact
- Error rate: **94.1%**
- Function duration: **5x increase** (40ms → 210ms)
- Started: October 29, 2025

## Solution
Disable ISR with `revalidate = false`. All pages are pre-rendered at build time and never need runtime regeneration.

## Expected Result
- Error rate: 0%
- Function duration: back to ~40ms baseline
- Function invocations reduced to static file serving only

### DIFF
--- a/app/[[...path]]/page.tsx
+++ b/app/[[...path]]/page.tsx
@@ -45,6 +45,7 @@ export async function generateStaticParams() {
 // Only render paths returned by generateStaticParams
 export const dynamicParams = false;
 export const dynamic = 'force-static';
+export const revalidate = false;
 
 const mdxComponentsWithWrapper = mdxComponents(
   {Include, PlatformContent},


sergical:
Closing this PR - the fix doesn't actually solve the problem. The real issue is that generateMetadata() tries to scan the docs/ directory at runtime even with force-static. Working on proper fix that pre-computes the doc tree at build time.

PR #15637, opened 2025-11-26
fix: Pre-compute doc tree at build time to avoid runtime ENOENT errors

sergical:
## Summary

This fixes the spike in function duration and 94.1% error rate on Vercel that started Oct 29, 2025.

### Before (Oct 29, 2025 spike):

Image_2025-11-27_11-36-52
Image_2025-11-27_11-41-35

Function invocations spiked to 20K errors

Error rate jumped from <0.1% to 89-94%

## Root Cause

1. **The Setup**: The `docs/**/*` directory was excluded from serverless function deployments in `next.config.ts` to stay under Vercel's 250MB limit
2. **Initial Success**: Pages worked fine after deployment because they were pre-built statically with access to the `docs/` folder
3. **The Trigger**: Around **Oct 29, 2025 at 16:40 UTC**, ISR caches started expiring across pages
4. **The Problem**: When Next.js 15 tried to verify metadata freshness on cache misses, it called:
   - `generateMetadata()` → `getDocsRootNode()` → `getAllFilesRecursively()` → attempted to scan `/var/task/docs`
5. **The Failure**: Since `docs/` was excluded from the bundle → **ENOENT errors**
6. **The Impact**: 
   - Errors spiked from ~100 per 4 hours to **20,000+ per 4 hours**
   - Function duration increased from **~40ms to ~210ms**
   - Error rate jumped from **<0.1% to 94.1%**

### Error logs from Vercel:
```
Error: ENOENT: no such file or directory, scandir '/var/task/docs'
```

## Solution

**Pre-compute doc tree at build time, load from JSON at runtime:**

1. **New build script** (`scripts/generate-doctree.ts`):
   - Runs during build when `docs/` is available
   - Generates complete doc tree for both regular and developer docs
   - Saves to `public/doctree.json` and `public/doctree-dev.json` (always deployed as static assets)
   - Strips circular `parent` references before serialization

2. **Runtime loading** (`src/docTree.ts`):
   - In development: scans filesystem normally (hot reload support)
   - In production: loads from pre-computed JSON
   - Reconstructs parent references after loading for pagination navigation
   - Checks `isDeveloperDocs` flag to load correct tree

3. **Build integration** (`package.json`):
   - Added `generate-doctree` step before `next build`
   - Uses esbuild to compile script, then node to execute

## Expected Impact

- ✅ Function duration: **210ms → ~40ms**
- ✅ Error rate: **94.1% → 0%** 
- ✅ Eliminates all ENOENT errors
- ✅ No changes to page rendering or user experience
- ✅ Doc tree still builds fresh in development mode

## Technical Details

**Files changed:**
- `scripts/generate-doctree.ts` - New build-time doc tree generator
- `src/docTree.ts` - Modified to load from JSON in production
- `package.json` - Added doctree generation to build pipeline
- `.gitignore` - Excluded generated JSON files from git

**Why this works:**
- `public/` files are always included in Vercel deployments (unlike `docs/` which we excluded)
- Pre-computing at build time means no runtime filesystem access needed
- JSON is ~5.2MB for regular docs, ~97KB for developer docs - well within limits

### DIFF
--- a/src/docTree.ts
+++ b/src/docTree.ts
@@ -62,6 +62,7 @@ async function getDocsRootNodeCached(): Promise<DocNode> {
   // Try public/ first (for serverless), then .next/ (for standalone)
   const paths = [path.join(root, 'public', filename), path.join(root, '.next', filename)];
 
+  let lastError: Error | undefined;
   for (const treePath of paths) {
     try {
       const treeData = await fs.readFile(treePath, 'utf-8');
@@ -70,13 +71,25 @@ async function getDocsRootNodeCached(): Promise<DocNode> {
       reconstructParentReferences(tree);
       return tree;
     } catch (error) {
-      // Try next path
-      continue;
+      // Only continue to next path if file doesn't exist
+      // Other errors (corrupt JSON, parse failures) should fail immediately
+      if (
+        error &&
+        typeof error === 'object' &&
+        'code' in error &&
+        error.code === 'ENOENT'
+      ) {
+        lastError = error as Error;
+        continue;
+      }
+      // Re-throw non-ENOENT errors (JSON parse errors, corruption, etc.)
+      throw error;
     }
   }
 
   throw new Error(
-    `Pre-computed doc tree not found (${filename}). Build may have failed or doctree was not generated with matching NEXT_PUBLIC_DEVELOPER_DOCS flag.`
+    `Pre-computed doc tree not found (${filename}). Build may have failed or doctree was not generated with matching NEXT_PUBLIC_DEVELOPER_DOCS flag.`,
+    {cause: lastError}
   );
 }
 


src/docTree.ts
```diff
@@ -34,11 +34,66 @@ export function getDocsRootNode(): Promise<DocNode> {
...
+  }
+  node.children.forEach(child => reconstructParentReferences(child, node));
+}
+
+async function getDocsRootNodeCached(): Promise<DocNode> {
+  // In development, scan filesystem for hot reloading
+  // In production (including CI builds), load from pre-computed JSON
+  if (process.env.NODE_ENV === 'development') {
+    return getDocsRootNodeUncached();
+  }
```
BYK, line 53: Not very efficient to check this at every invocation of the function as env variable value would never change across the lifetime of the app. A more efficient approch would be to define the function differently based on the env var value. Something along these lines:

```ts

const getDocsRootNodeCached = process.env.NODE_ENV === 'production' ? async () => {

...
...
} : getDocsRootNodeUncached;

```
Also note that I check for `=== 'production'` instead of `=== 'development'` as during development, `process.env.NODE_ENV` is usually empty or undefined.

PR #15678, opened 2025-12-01
docs: Simplify OTLP "Dedicated Integrations" section

szokeasaurusrex:
I am concerned that the idiom "make setting up OTLP a breeze" may not be immediately understood by non-native English speakers, especially those with a more basic proficiency level.

To ensure our docs are accessible to a global audience, I propose replacing the idiom with a direct statement that the integrations "make OTLP setup easy."


AbhiPrasad:
Gonna merge this in as I'm editing this page and I want to avoid merge conflicts

PR #15691, opened 2025-12-02
docs(nextjs): Move webpack build config under `webpack` namespace

logaretm:
We are moving the only-webpack related configurations under `webpack` namespace to better fit our goal of communicating the SDK readiness for Turbopack by default.

This involved moving several options to be nested under a `webpack` path in the build configuration. 

This is meant to be released once this https://github.com/getsentry/sentry-javascript/pull/18343/ goes live, so I will keep this open until then. Will keep it in draft as well.

The SDK PR adds warnings for those deprecated options and warns users about using them if detected so I don't think we need to do that here once it goes live.

### DIFF
--- a/docs/platforms/javascript/guides/nextjs/configuration/build/index.mdx
+++ b/docs/platforms/javascript/guides/nextjs/configuration/build/index.mdx
@@ -222,6 +222,21 @@ Enables the use of the [`runAfterProductionCompile` hook from Next.js](https://n
 
 </SdkOption>
 
+<SdkOption name="disableManifestInjection" type="boolean|undefined" defaultValue="false">
+
+Disables automatic injection of the route manifest into the client bundle.
+
+The route manifest is a build-time generated mapping of your Next.js App Router routes that enables Sentry to group transactions by parameterized route names (e.g., `/users/:id` instead of `/users/123`, `/users/456`, etc.).
+
+**Disable this option if:**
+
+- You want to minimize client bundle size
+- You're experiencing build issues related to route scanning
+- You prefer raw URLs in transaction names
+- You're only using the Pages Router (this feature is only supported in the App Router)
+
+</SdkOption>
+
 ## Next.js Webpack Options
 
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/guides/nextjs/configuration/build/index.mdx
```diff
@@ -182,25 +182,72 @@ Note: Enabling this option can lead to longer build times.
...
+- This option is set to `false` for Webpack as the default behavior is to upload sourcemaps during the build process using the [Sentry Webpack Plugin](https://github.com/getsentry/sentry-javascript-bundler-plugins).
+
+**Important:** Enabling this option will mutate your Next.js build output by injecting [Debug IDs](/platforms/javascript/guides/nextjs/sourcemaps/troubleshooting_js/debug-ids/) via the Sentry CLI. If you are relying on any sort of integrity hashes for your build artifacts, you will need to disable this option.
+
+</SdkOption>
+
+## Next.js Webpack Options
+
+<Alert level="info">
+  These options only take effect if you are using Webpack. If you are using
```
inventarSarah:
```suggestion
  These options only take effect if you're using Webpack. If you're using
```

docs/platforms/javascript/guides/nextjs/configuration/build/index.mdx
```diff
@@ -226,35 +273,25 @@ excludeServerRoutes: [
```
inventarSarah: I think we can remove the alert about Turbopack in the lines above. wdyt? 

This one:
```
<Alert level="info">
  If you're building your Next.js application with Turbopack, this option will
  have no effect. The Sentry SDK will no longer apply build-time instrumentation
  and will rely purely on Next.js telemetry features.
</Alert>
```

logaretm: Yep, absolutely 💯

PR #15710, opened 2025-12-03
chore(dynamic-sampling): add notice that we will remove the get samples button

shellmayr:
- As we will be removing the Get Samples button as part of our move to Span First, we want to shout this out in the docs

Closes TET-1479

### DIFF
--- a/docs/organization/dynamic-sampling/index.mdx
+++ b/docs/organization/dynamic-sampling/index.mdx
@@ -206,9 +206,9 @@ We deprioritize health check type transactions because while they're important f
 
 ### Get Samples
 
-<Info>
+<Alert>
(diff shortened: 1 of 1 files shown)


docs/organization/dynamic-sampling/index.mdx
```diff
@@ -206,6 +206,10 @@ We deprioritize health check type transactions because while they're important f
 
 ### Get Samples
 
+<Info>
+The "Get Samples" button and the investigation rules that go along with it will be retired in spring 2026. From that point on, you will no longer be able to create new investigation rules.
```
shellmayr:
```suggestion
The "Get Samples" button and the investigation rules that go along with it will be retired in Spring 2026. From that point on, you will no longer be able to create new investigation rules.
```

PR #15742, opened 2025-12-08
chore(docs): delete github deployment gate

cathteng:
This is no longer supported (we've removed all the infra). The app can be deleted too


leeandher (approved):
🔥

PR #15816, opened 2025-12-15
feat(lambda-layers): make ARN easier to see & copy

shellmayr:
- Pre-select a region to show an ARN without user interaction
- Put the ARN into a code block instead of plain text to make it more obvious that there is something to copy here

**Before** 
Screenshot 2025-12-15 at 09 58 52
Screenshot 2025-12-15 at 09 58 57

**After**
Screenshot 2025-12-15 at 09 59 05


shellmayr:
@andreiborza yeah that's a great idea, let me rehash this

shellmayr:
Just cleaning things up - I tried adapting it but it's not great, so closing this because I don't want to spend more time on it atm.

PR #15860, opened 2025-12-18
ref: Rename telemetry buffer pages to processor

philipphofmann:
Rename all the telemetry buffer pages to telemetry processor and clarify that the processor does more than buffering.

This is the first step of migrating the [Notion doc](https://www.notion.so/sentry/Telemetry-Processor-2b58b10e4b5d805e819be97fbfc0cd53) concept of the telemetry processor to the develop docs.

### DIFF
--- a/develop-docs/sdk/telemetry/telemetry-processor/index.mdx
+++ b/develop-docs/sdk/telemetry/telemetry-processor/index.mdx
@@ -23,7 +23,7 @@ flowchart LR
 Once an SDK adds the telemetry processor, its client **SHOULD** only forward data to the telemetry processor, and not directly to the transport. The TelemetryProcessor consists of two major components:
 
 - **TelemetryBuffer**: Focuses on buffering high-volume data, such as spans and logs, into batches to minimize HTTP requests.
-- **TelemetryDispatcher**: Takes buffered data from the TelemetryBuffer and manages prioritized sending, including potential offline caching and sending of client reports.
+- **TelemetryScheduler**: Takes buffered data from the TelemetryBuffer and manages prioritized sending, including potential offline caching and sending of client reports.
 
 Because telemetry workloads and platform constraints vary widely, the requirements differ across environments. For example, backend SDKs need high throughput and backpressure management to handle large data volumes. Mobile SDKs have lower throughput and don't need to worry much about backpressure, but they do need to minimize data loss in the event of abnormal process termination. Browser and GDX SDKs also have different requirements.
 
--- a/develop-docs/sdk/telemetry/telemetry-processor/backend-telemetry-processor.mdx
+++ b/develop-docs/sdk/telemetry/telemetry-processor/backend-telemetry-processor.mdx
@@ -50,7 +50,7 @@ Introduce a `TelemetryProcessor` layer between the `Client` and the `Transport`.
 │                                │                                           │
 │                                ▼                                           │
 │   ┌─────────────────────────────────────────────────────────────────────┐  │
-│   │             TelemetryDispatcher (Weighted Round-Robin)              │  │
+│   │             TelemetryScheduler (Weighted Round-Robin)               │  │
 │   │  - Priority weights: CRITICAL=5, HIGH=4, MEDIUM=3, LOW=2, LOWEST=1  │  │
 │   │  - Processes a batch of items based on BatchSize and/or Timeout     │  │
 │   │  - Builds envelopes from batch                                      │  │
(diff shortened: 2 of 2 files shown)


develop-docs/sdk/telemetry/telemetry-processor/index.mdx
```diff
@@ -0,0 +1,55 @@
...
+```
+
+Once an SDK adds the telemetry processor, its client **SHOULD** only forward data to the telemetry processor, and not directly to the transport. The TelemetryProcessor consists of two major components:
+
+- **TelemetryBuffer**: Focuses on buffering high-volume data, such as spans and logs, into batches to minimize HTTP requests.
+- **TelemetryDispatcher**: Takes buffered data from the TelemetryBuffer and manages prioritized sending, including potential offline caching and sending of client reports.
+
+Because telemetry workloads and platform constraints vary widely, the requirements differ across environments. For example, backend SDKs need high throughput and backpressure management to handle large data volumes. Mobile SDKs have lower throughput and don't need to worry much about backpressure, but they do need to minimize data loss in the event of abnormal process termination. Browser and GDX SDKs also have different requirements.
+
+Therefore, we recommend implementing different types of telemetry processors tailored to the platform's needs. As of Dec 18th, 2025, this page is under development, and we're currently refining the requirements for different platforms:
```
philippsuess, line 30: I'm sorry the git mv command somehow got messed up here. This is basically the new part. The rest already existed before.

develop-docs/sdk/telemetry/telemetry-processor/index.mdx
```diff
@@ -0,0 +1,55 @@
...
+```mermaid
+flowchart LR
+  Client -- send Data --> TelemetryProcessor
+  TelemetryProcessor -- sendEnvelope --> Transport
+```
+
+Once an SDK adds the telemetry processor, its client **SHOULD** only forward data to the telemetry processor, and not directly to the transport. The TelemetryProcessor consists of two major components:
+
+- **TelemetryBuffer**: Focuses on buffering high-volume data, such as spans and logs, into batches to minimize HTTP requests.
+- **TelemetryDispatcher**: Takes buffered data from the TelemetryBuffer and manages prioritized sending, including potential offline caching and sending of client reports.
```
giortzisg: I personally prefer TelemetryScheduler as a name, since for me it more clearly represents what the component actually does. But that's a nit, so leaving it as is, is fine.

giortzisg (approved):

develop-docs/sdk/telemetry/telemetry-processor/index.mdx
philippsuess: @giortzisg, I don't think that's a nit. It's an important point. I think we can also follow up on this later. The concept is still WIP, and we will get to a close to ideal solution step by step. @isaacs, what is your suggestion for a proper name for the TelemetryDispatcher?

develop-docs/sdk/telemetry/telemetry-processor/index.mdx
isaacs: `git mv` is basically just sugar around `git rm` + `git add`. If the file contents don't change, it's reflected as a rename; but if you add more than a certain percentage, it'll be interpreted as a new different file. (Ie, you did nothing wrong, this is just git's heuristics working as designed, but you can avoid this in the future by having separate commits: one to rename the file, then another to edit it. But that's usually more trouble than it's worth, to be honest.)

develop-docs/sdk/telemetry/telemetry-processor/index.mdx
isaacs: I would think of a "dispatcher" as the thing that manages the actual sending as well as scheduling. For example, if it had multiple different clients, and was deciding where to send different types of data.

In this case, we have exactly one client (the HTTP client that uploads to Sentry), and this component is just deciding which things get placed in its queue in which order, so I agree `Scheduler` is probably a slightly better term.

But I also agree with @giortzisg that this is a nit ;) I could easily see how it could grow into a full "dispatcher" if one had multiple endpoints, for example.

develop-docs/sdk/telemetry/telemetry-processor/index.mdx
philippsuess: Thanks for the explanation. TIL

develop-docs/sdk/telemetry/telemetry-processor/index.mdx
philippsuess: OK, then let's go for Scheduler 🙏

PR #15861, opened 2025-12-18
docs(tanstackstart-react): Add wrapFetchWithSentry

nicohrubec:
## DESCRIBE YOUR PR

We will release a new `wrapFetchWithSentry` API for the Tanstack Start SDK that adds support for server-side tracing. Later this will also handle server-side error monitoring, but that doesn't work yet.

PR should be merged after the next JS SDK release (10.32.0).

### DIFF
--- a/docs/platforms/javascript/guides/tanstackstart-react/index.mdx
+++ b/docs/platforms/javascript/guides/tanstackstart-react/index.mdx
@@ -140,7 +140,6 @@ Sentry.init({
 });
 ```
 
-{/* TODO: enable for error monitoring as well once that works  */}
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/guides/tanstackstart-react/index.mdx
```diff
@@ -140,6 +140,29 @@ Sentry.init({
 });
 ```
 
+{/* TODO: enable for error monitoring as well once that works  */}
```
andreiborza: I'd probably remove this, seems unlikely we'll forget to add this once we actually work on it right?

PR #15888, opened 2025-12-29
docs(encryption): Add docs for encrypted field usage

vgrozdanic:
Adds docs about usage of encryption fields at sentry.

it also includes a more in-detail section on how things work under the hood (for SRE and infra teams)

Closes [TET-523: Docs for adding encrypted field to the model](https://linear.app/getsentry/issue/TET-523/docs-for-adding-encrypted-field-to-the-model) and [TET-522: Docs for migrating existing fields to encrypted ones](https://linear.app/getsentry/issue/TET-522/docs-for-migrating-existing-fields-to-encrypted-ones)

### DIFF
--- a/develop-docs/backend/application-domains/encrypted-fields/index.mdx
+++ b/develop-docs/backend/application-domains/encrypted-fields/index.mdx
@@ -109,12 +109,13 @@ class MyModel(models.Model):
 
 ### Adding New Encrypted Fields
 
-Add the field to your model and generate a migration:
-
+Add the field to your model:
 ```python
 class MyModel(models.Model):
     api_key = EncryptedCharField(null=True, blank=True)
 ```
+
+and generate a migration:
 
 ```bash
 sentry django makemigrations


develop-docs/backend/application-domains/encrypted-fields/index.mdx
```diff
@@ -0,0 +1,303 @@
+---
+title: "Encrypted Fields"
+description: "Drop-in replacement Django fields for encrypting sensitive data in Sentry."
+categories:
+  - backend
+  - encryption
+  - django
+sidebar_order: 10
```
shellmayr: This puts the document at the top of Backend -> Application Domains - don't think we should be putting it there, since this will not be used by most people. Would rather put it after Metrics, which would be order 190.
```suggestion
sidebar_order: 190
```

develop-docs/backend/application-domains/encrypted-fields/index.mdx
```diff
@@ -0,0 +1,303 @@
...
+from sentry.db.models.fields.encryption import EncryptedCharField
+
+class MyModel(models.Model):
+    secret_token = EncryptedCharField()
+    api_key = EncryptedCharField(null=True, blank=True)
+```
+
+### [EncryptedJSONField](#encryptedjsonfield)
+
+A drop-in replacement for Django's `JSONField` that encrypts JSON data.
```
shellmayr: Does this replacement support the arguments that the original JSONField supports, especially special encoders & decoders?

shellmayr:
Reads well already! I think structurally, we can improve this by reordering a bit and grouping this slightly differently. It may pay off to have two sub-pages for using encrypted fields, and for all of the admin work around it:

User-facing:
1. Explanation about what it does (already exists) and when to use it ("What to encrypt") - this will be most important for developers who have never heard of it
2. Basic setup including querying
3. Field types
4. Troubleshooting
5. How it works under the hood

Admin-related:
1. Configuration
2. Key Rotation
3. Key Management

develop-docs/backend/application-domains/encrypted-fields/index.mdx
vgrozdanic: make sense

develop-docs/backend/application-domains/encrypted-fields/index.mdx
vgrozdanic: no, unfortunately it doesn't, i'll remove a term "drop-in" as it is not really a drop-in replacement.

develop-docs/backend/application-domains/encrypted-fields/index.mdx
```diff
@@ -0,0 +1,218 @@
...
+
+```python
+class MyModel(models.Model):
+    api_key = EncryptedCharField(null=True, blank=True)
+```
+
+```bash
+sentry django makemigrations
+sentry upgrade
+```
```
shellmayr, line 123: This should not be the same code-field - currently this is rendered next to each other, most people will not see it: 

Image

I think you need to move the text between the two for it to be rendered correctly.

develop-docs/backend/application-domains/encrypted-fields/administration.mdx
```diff
@@ -0,0 +1,92 @@
...
+Data will be gradually re-encrypted as records are updated.
+
+### Generating Keys
+
+Generate a Fernet key using Python:
+
+```python
+from cryptography.fernet import Fernet
+
+key = Fernet.generate_key()
```
shellmayr, line 82: Are there any cryptographic options we should put here? Or are these all centrally "managed"?

develop-docs/backend/application-domains/encrypted-fields/index.mdx
vgrozdanic: ohh damn, nice catch

develop-docs/backend/application-domains/encrypted-fields/administration.mdx
vgrozdanic: nothing, there are not special options for Fernet

PR #15899, opened 2026-01-05
docs(nextjs): Document cache components limitations

logaretm:
With more users adopting cache components and running into our current limitations with it, we have to document it for them with the possible workarounds.

None of these are ideal, but the way cache components interferes with our core SDK behavior is not straightforward to fix and may need fundamental changes to our core SDK.

For now I made sure we document the couple of wide cases that the issue appears in, and possible workarounds (more like compromises).

### DIFF
--- a/platform-includes/nextjs-cache-components-limitations/alert/javascript.nextjs.mdx
+++ b/platform-includes/nextjs-cache-components-limitations/alert/javascript.nextjs.mdx
@@ -1,4 +1,4 @@
-<Alert level="warning" title="Cache Components Limitation">
+<Alert level="warning" title="Cache Components Limitations">
 
 If you're using Next.js [Cache Components](https://nextjs.org/docs/app/getting-started/cache-components), there are known compatibility issues with server-side tracing. See [Cache Components Limitations](#cache-components-limitations) for details and workarounds.
 


docs/platforms/javascript/guides/nextjs/manual-setup/index.mdx
```diff
@@ -15,6 +15,12 @@ categories:
   installer](/platforms/javascript/guides/nextjs).
 </Alert>
 
+<Alert level="warning" title="Cache Components Limitation">
+
+If you're using Next.js [Cache Components](https://nextjs.org/docs/app/getting-started/cache-components), there are known compatibility issues with server-side tracing. See [Cache Components Limitations](#cache-components-limitations) for details and workarounds.
```
sergical: i dont think this page will have that content so the link won't work. maybe create both the alert and the section as platform includes components so we can reuse them if needed, maybe also link to a github ticket tracking the work / conversation around supporting cache components?

logaretm: Good catch! I already included the issues in the section description, do you think the links should be also in the alert?

logaretm: merging now but happy to change it later

sergical: i think its fine as is, sorry i missed it in the description

PR #15911, opened 2026-01-06
fix(build): Enable esbuild write in local production builds

a-hariti:
`yarn build` was failing with a weird MDX syntax error.

Upon closer inspection it was coming from the `esbuild` write option.


sergical (approved):
lgtm, thanks!

PR #15929, opened 2026-01-08
feat(stats): Update screenshots

priscilawebdev:


priscilawebdev:
> Looks good - for sake of performance - larger images here are on purpose, right?

right! I believe the sizes are fine

PR #15936, opened 2026-01-08
feat(nextjs): Add comprehensive error capturing documentation

sergical:
## Summary
- Create dedicated `/capturing-errors` page for Next.js with SplitLayout pattern
- Cover all Next.js error handling patterns: error boundaries, Server Actions, API routes, middleware
- Add best practices and troubleshooting sections
- Redirect `/usage` → `/capturing-errors`

## Test plan
- [ ] Verify page renders correctly at `/platforms/javascript/guides/nextjs/capturing-errors/`
- [ ] Verify redirect works from `/platforms/javascript/guides/nextjs/usage/`
- [ ] Verify only one "Capturing Errors" item in Next.js sidebar
- [ ] Check all internal links work

### DIFF
--- a/docs/platforms/javascript/guides/nextjs/capturing-errors/index.mdx
+++ b/docs/platforms/javascript/guides/nextjs/capturing-errors/index.mdx
@@ -98,7 +98,9 @@ export default function Error({
 
 ### Global Error Boundary
 
-`global-error.tsx` catches errors in your root layout. It must include `<html>` and `<body>` tags since it replaces the entire document.
+`global-error.tsx` is a last-resort safety net that only triggers when your root layout itself fails. This is rare—most errors are caught by route-level `error.tsx` files first.
+
+Include `<html>` and `<body>` tags since it replaces the entire document.
 
 </SplitSectionText>
 <SplitSectionCode>
@@ -172,7 +174,9 @@ export async function createPost(formData: FormData) {
 
 ### Adding Context
 
-Pass a second argument to `captureException` to add debugging context:
+Add metadata to help debug errors:
+- **Tags**: Searchable in Sentry's UI for filtering and grouping (e.g., by feature area or user tier)
+- **Extra**: Displayed in event details for debugging values like IDs (not searchable)
 
 ```tsx
 Sentry.captureException(error, {
@@ -180,6 +184,8 @@ Sentry.captureException(error, {
   extra: { orderId, userId: user.id },
 });
 ```
+
+See <PlatformLink to="/enriching-events/">Enriching Events</PlatformLink> for more options like breadcrumbs, user context, and attachments.
 
 For automatic tracing in Server Actions, see <PlatformLink to="/configuration/apis/#withServerActionInstrumentation">`withServerActionInstrumentation`</PlatformLink>.
 


docs/platforms/javascript/guides/nextjs/capturing-errors/index.mdx
```diff
@@ -0,0 +1,474 @@
...
+
+## API Routes
+
+<SplitLayout>
+<SplitSection>
+<SplitSectionText>
+
+### Route Handlers
+
+Errors in API Route Handlers are automatically captured. For custom error responses, catch and report manually.
```
sfanahata: Are we trying to demonstrate how to extend this functionality with another example of where to use `captureException`? 

I'm noticing you aren't adding more information for unhandled exceptions in client-side code, or unhandled promise rejections from the list of things Sentry automatically captures. 

If you do want to keep this section, I think it needs a bit more reasoning around why you're bringing it up.

docs/platforms/javascript/guides/nextjs/capturing-errors/index.mdx
```diff
@@ -0,0 +1,474 @@
...
+    return NextResponse.next();
+  }
+}
+```
+
+</SplitSectionCode>
+</SplitSection>
+</SplitLayout>
+
+## Manual Capture
```
sfanahata: How is this section different from the others? It seems like this whole guide is about manual capture?

docs/platforms/javascript/guides/nextjs/capturing-errors/index.mdx
```diff
@@ -0,0 +1,474 @@
...
+
+  Sentry.captureException(error);
+});
+```
+
+</SplitSectionCode>
+</SplitSection>
+</SplitLayout>
+
+## Best Practices
```
sfanahata: I like this, but it also feels a little repetitive. I wonder if this just gets added to the sections where you already talk about these things - for example global settings vs at a path level? 

Otherwise, this might be better as a TL;DR, instead of having more examples? Only because it elongates the page an awful lot. I love me some best practices though.

sfanahata (commented):
Meta comment: This is really good! I think we could do something similar to clean up the inherited common version of this page too, to streamline in some ways. 

I think we could be a bit more precise with when and why we're talking about using certain functions/options in order to further streamline this page.

docs/platforms/javascript/guides/nextjs/capturing-errors/index.mdx
sergical: i have this conviction that docs are more and more consumed by LLMs. And not so much in a way that is copy ai rules to an llm but literally copy and paste the url to an agent. we do the conversion to markdown already, and this structure imo gives the LLMs a better chance of parsing and setting things up properly. 

I could be completely wrong tho, definitely open to more feedback, cause yeah, i do understand that it extends the page in length, but for LLMs its a good thing (i think?lol) and for devs we have the sidebar with sections to skim and jump to the right part, if they are on this page, the assumption is they are here for a reason, and have it be a manual with more context than less. 

Just an opinion though, absolutely not set in stone

sfanahata: I don't know how to validate that, or that LLMs will consume this better, but I also think we should be writing as clearly and concisely as we can for people, with LLMs in mind. As a person, I'm not sure I'd be expecting this level of repetition, and still get to the end.

sergical: agreed, really great feedback, thank you! i'll update tomorrow!

docs/platforms/javascript/guides/nextjs/capturing-errors/index.mdx
```diff
@@ -0,0 +1,282 @@
...
+}
+```
+
+</SplitSectionCode>
+</SplitSection>
+
+<SplitSection>
+<SplitSectionText>
+
+### Global Error Boundary
```
sfanahata, line 99: In previous versions this was called out as more of a blunt force tool, kind of like a "porous mesh". Here, it's just listed as another place to add in captureException without much context on when to use it. Why did you take away the context? Was it incorrect?

docs/platforms/javascript/guides/nextjs/capturing-errors/index.mdx
```diff
@@ -0,0 +1,282 @@
...
+    return { success: false, error: "Failed to create post" };
+  }
+}
+```
+
+</SplitSectionCode>
+</SplitSection>
+</SplitLayout>
+
+### Adding Context
```
sfanahata, line 175: Funny enough, I don't think there's enough context here :wink: I'd like just one sentence about what's happening with that extra data. I get a clue with the "automatic tracing" reference right below, but I don't quite get it off the bat, other than looks like I can add "tags" and "extra" data.

sfanahata (approved):
Not gonna block this, but I do think that adding something more to those two sections I just commented on would make this doc 🔥

docs/platforms/javascript/guides/nextjs/capturing-errors/index.mdx
sergical: *Error boundary placement* section covers the placement. It's one of those pages that rarely should be hit since you would want to catch the error as early as possible with either a page specific `error.tsx` or the app specific one

PR #15948, opened 2026-01-09
fix(build): Invalidate MDX cache when referenced images change

sergical:
## Summary
- Fixes the issue where updating images doesn't invalidate the build cache
- Adds a function to hash referenced image files and include them in the MDX cache key
- When an image changes, the cache key changes, triggering a page rebuild

## Problem
When images are updated but their referencing MDX files aren't changed, Vercel serves stale images because the MDX cache key only depends on the MDX source content hash.

Discussed in [#docs-dev Slack thread](https://sentry.slack.com/archives/C8H0BQRDJ/p1767945779845739).

## Test plan
- [ ] Deploy to preview and verify images update correctly
- [ ] Verify build still works for pages without images
- [ ] Check that the cache key changes when an image is modified

### DIFF
--- a/src/mdx.ts
+++ b/src/mdx.ts
@@ -174,6 +174,22 @@ const isSupported = (
 let getDocsFrontMatterCache: Promise<FrontMatter[]> | undefined;
 
 export function getDocsFrontMatter(): Promise<FrontMatter[]> {
+  // Block filesystem scanning at Vercel runtime.
+  // Frontmatter should only be scanned during CI builds - the doc tree is pre-computed.
+  // See: DOCS-83Q, DOCS-9A5, DOCS-9RE
+  const isVercelRuntime =
+    process.env.VERCEL && !process.env.CI && process.env.NODE_ENV !== 'development';
+
+  if (isVercelRuntime) {
+    return Promise.reject(
+      new Error(
+        `[MDX Runtime Error] Attempted to scan docs frontmatter at Vercel runtime. ` +
+          `This should not happen - the doc tree should be pre-computed during CI. ` +
+          `If you're seeing this error, the requested path may not exist or was not included in generateStaticParams().`
+      )
+    );
+  }
+
   if (!getDocsFrontMatterCache) {
     getDocsFrontMatterCache = getDocsFrontMatterUncached();
   }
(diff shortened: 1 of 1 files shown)


src/mdx.ts
```diff
@@ -69,6 +69,35 @@ if (process.env.CI) {
...
+
+  const imageHashes: string[] = [];
+  for (const match of matches) {
+    const imagePath = path.join(cwd, match[1]);
+    try {
+      const imageContent = await readFile(imagePath);
+      imageHashes.push(md5(imageContent));
+    } catch {
+      // Image doesn't exist yet or can't be read, skip
+    }
```
BYK: You may benefit from doing a version of `Promise.all(...)` instead of this for loop. Don't forget to use [`p-limit`](https://github.com/sindresorhus/p-limit) when you do so to avoid overloading the FS though.

sergical: removing this in favor of adding has query param to the image itself, what do you think about that? should bust the vercel cache (and actually update the mdx hash because the content will be different

src/remark-image-size.js
```diff
@@ -1,13 +1,16 @@
...
- * the size is consumed by docImage.tsx and passed down to next/image
- * **this is a hack!**, there's probably a better way to set image node properties
- * but adding a hash to the url seems like a very low risk way to do it 🙈
+ * Appends image size and content hash to the URL.
+ * e.g. /img.png -> /img.png?v=abc12345#100x100
+ *
+ * The size is consumed by docImage.tsx and passed down to next/image.
+ * The content hash (?v=xxx) busts Vercel's CDN cache when images are updated.
  */
 export default function remarkImageSize(options) {
```
priscilawebdev: It looks like this function now does more than just handle image size. Would it make sense to update its name to better reflect its current responsibilities?

src/remark-image-size.js
```diff
@@ -1,13 +1,16 @@
...
 
 /**
- * appends the image size to the image url as a hash e.g. /img.png -> /img.png#100x100
- * the size is consumed by docImage.tsx and passed down to next/image
- * **this is a hack!**, there's probably a better way to set image node properties
- * but adding a hash to the url seems like a very low risk way to do it 🙈
+ * Appends image size and content hash to the URL.
+ * e.g. /img.png -> /img.png?v=abc12345#100x100
+ *
+ * The size is consumed by docImage.tsx and passed down to next/image.
```
priscilawebdev: Nit: why are we doing this? (That was my initial question) Would it be worth mentioning that this is done to avoid any extra image measurement?

src/mdx.ts
BYK: Not a fan of query params but you know Vercel platform better than I do so I have no preference here :)

src/remark-image-size.js
sergical: good point, thanks for flagging!

src/remark-image-size.js
sergical: The version hash will help bust the vercel cache so that new images with the same name don't end up being cached. this also updates the html that we use for content hashing and checking whether changes were made to a file or not to update that cache. 

why are we doing the image size? that was there before. The idea is when we render it in the lightbox container, `next/image` requires a width and a height prop to render the image properly, this puts it in the url of the image for the component to grab it

PR #15950, opened 2026-01-09
chore(platform): Enable default pii for Sentry

chargome:


chargome:
Not going forward with this

PR #15966, opened 2026-01-12
fix(mdx): Use /tmp for esbuild outdir on Vercel runtime

codyde:
## Summary
- Fixes esbuild "read-only file system" errors on Vercel Lambda runtime (DOCS-9K6 - 330k+ events)
- Redirects esbuild's `outdir` to `/tmp/mdx-images` when running on Vercel runtime since `/var/task` is read-only
- esbuild internally creates the outdir even when `options.write=false`, bypassing the existing try-catch

## Root Cause
esbuild's internal directory creation fails on Vercel's read-only filesystem (`/var/task/public/mdx-images`) during MDX compilation, despite `options.write=false` being set. The only writable directory in AWS Lambda is `/tmp`.

## Related Issue
https://sentry.sentry.io/issues/DOCS-9K6

### DIFF
--- a/src/mdx.ts
+++ b/src/mdx.ts
@@ -639,9 +639,13 @@ export async function getFileBySlug(slug: string): Promise<SlugFile> {
   let cacheFile: string | null = null;
   let assetsCacheDir: string | null = null;
 
-  // Always use public/mdx-images during build
-  // During runtime (Lambda), this directory is read-only but images are already there from build
-  const outdir = path.join(root, 'public', 'mdx-images');
+  // Use public/mdx-images during build (CI or local dev)
+  // During Vercel runtime (Lambda), /var/task is read-only but /tmp is writable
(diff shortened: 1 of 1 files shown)


codyde:
Closing this PR after deeper investigation.

**Analysis findings:**
- The existing guard at `getFileBySlug()` (added in commit d61e0aa20 on Jan 10, 2026) is working correctly
- **Zero errors** are occurring from releases after the guard was added
- All 330k+ errors are from **old Vercel preview deployments** from July-August 2024
- These old deployments are accessed via URLs like `sentry-docs-*.sentry.dev` (preview URLs)

**Why this fix wouldn't help:**
1. The guard throws before reaching the `outdir` configuration
2. Old deployments don't have this fix anyway
3. The real solution is cleaning up old preview deployments at the Vercel level

**Recommendation:** Delete old Vercel preview deployments to stop the errors, rather than adding code fixes that won't reach the affected deployments.

PR #15967, opened 2026-01-12
fix(components): Use useId for hydration-safe unique IDs

codyde:
## Summary
- Fixes React hydration errors (DOCS-7RD - 2.3M+ events) caused by dynamic ID generation during render
- Replaces `Math.random()` and `Date.now()` with React's `useId()` hook for stable, SSR-compatible unique IDs

## Changes
- **createGitHubAppForm.tsx**: Replace `Math.random() * MAX_COMPONENTS_ON_PAGE` with `useId()`
- **githubDomainChecker.tsx**: Replace `Date.now()` fallback with `useId()`

## Root Cause
When server renders a component with `Math.random()` or `Date.now()`, it produces a different value than the client during hydration, causing React to report a hydration mismatch. React's `useId()` hook generates deterministic IDs that are consistent between server and client.

## Related Issue
https://sentry.sentry.io/issues/DOCS-7RD

### DIFF
--- a/src/components/createGitHubAppForm.tsx
+++ b/src/components/createGitHubAppForm.tsx
@@ -1,9 +1,7 @@
 'use client';
 
-import {useState} from 'react';
+import {useId, useState} from 'react';
 import {Button} from '@radix-ui/themes';
-
-const MAX_COMPONENTS_ON_PAGE = 100;
 
 type AccountType = 'personal' | 'organization';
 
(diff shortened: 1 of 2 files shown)


codyde:
Closing in favor of #15970 which addresses the primary cause of hydration errors.

**Analysis findings:**
- The `createGitHubAppForm` and `githubDomainChecker` components fixed in this PR are only used on niche GitHub integration pages
- The **primary cause** of 2.3M hydration errors is the `PlatformSelector` component using localStorage to conditionally render a sidebar link
- PR #15970 fixes the `PlatformSelector` issue using the `hasMounted` pattern

The fixes in this PR are still technically valid (using `useId()` instead of `Math.random()`/`Date.now()`), but they address edge cases, not the main issue. They could be re-submitted separately as a cleanup if desired.

PR #15968, opened 2026-01-12
fix(page): Use pre-computed tree for version extraction

codyde:
## Summary
- Fixes ENOENT errors when scanning `/vercel/path0/docs` (DOCS-9RT - 6.5k events)
- Replaces `getDocsFrontMatter()` call with new `getVersionsFromTree()` function that extracts version info from the pre-computed doc tree

## Changes
- **src/docTree.ts**: Add `getVersionsFromTree()` function that collects version info by traversing the doc tree
- **app/[[...path]]/page.tsx**: Use `getVersionsFromTree(rootNode, pageNode.path)` instead of `getDocsFrontMatter()` + `getVersionsFromDoc()`

## Root Cause
The `Page` component was calling `getDocsFrontMatter()` which attempts to scan the filesystem recursively. On Vercel, the docs directory path (`/vercel/path0/docs`) may not exist during certain build/runtime states, causing ENOENT errors.

The doc tree (`doctree.json`) is pre-computed during build and always available, making it a reliable source for version information.

## Related Issue
https://sentry.sentry.io/issues/DOCS-9RT

### DIFF
--- a/app/[[...path]]/page.tsx
+++ b/app/[[...path]]/page.tsx
@@ -17,6 +17,7 @@ import {
   getDocsRootNode,
   getNextNode,
   getPreviousNode,
+  getVersionsFromTree,
   nodeForPath,
 } from 'sentry-docs/docTree';
 import {isDeveloperDocs} from 'sentry-docs/isDeveloperDocs';
@@ -24,7 +25,6 @@ import {
   getDevDocsFrontMatter,
   getDocsFrontMatter,
   getFileBySlugWithCache,
-  getVersionsFromDoc,
 } from 'sentry-docs/mdx';
 import {mdxComponents} from 'sentry-docs/mdxComponents';
 import {PageType} from 'sentry-docs/metrics';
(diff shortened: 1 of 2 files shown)


codyde:
Closing this PR after deeper investigation.

**Analysis findings:**
- All 6,500+ ENOENT errors are from **old Vercel preview deployments** (November 2024)
- Zero errors from releases after Jan 9, 2026 when the guard was added
- The combination of `dynamicParams=false` + `force-static` + the guard in `getDocsFrontMatter()` already prevents this issue on new deployments

**Why no errors from new deployments:**
1. `export const dynamicParams = false` - Next.js returns 404 for unknown paths without calling the Page component
2. `export const dynamic = 'force-static'` - All pages are pre-rendered at build time
3. Guard in `getDocsFrontMatter()` - Would throw a different error if somehow reached at runtime

**Why this fix wouldn't help:**
- Old deployments don't have this fix
- New deployments already don't have this problem

**Recommendation:** Clean up old Vercel preview deployments to stop these errors.

PR #15983, opened 2026-01-13
feat(telemetry-processor): Client Reports

philipphofmann:
## DESCRIBE YOUR PR

Add a concept of how the telemetry processor handles client reports.

Slighly related to https://github.com/getsentry/sentry-docs/pull/16027.

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

### DIFF
--- a/develop-docs/sdk/telemetry/telemetry-processor/index.mdx
+++ b/develop-docs/sdk/telemetry/telemetry-processor/index.mdx
@@ -91,6 +91,9 @@ The telemetry buffer batches high-volume data and forwards it to the telemetry s
 1. Before adding an item to a specific buffer, the telemetry buffer **SHOULD** drop rate-limited items to avoid overhead. If doing so, it **MUST** record client reports.
 2. When the telemetry buffer overflows and it drops data, it **MUST** record client reports.
 3. The telemetry buffer **MUST** forward low-volume data, such as normal events, session replays, or user feedback, directly to the telemetry scheduler.
+4. The telemetry buffer **MUST** start a timeout of 5 seconds when the first item is added. When the timeout expires, the telemetry buffer **MUST** forward all items to the telemetry scheduler.
+5. The telemetry buffer **MUST** define a size limit of x items. Each SDK MAY choose its own x based on its needs, but SDKs **MUST** ensure that the serialized envelope doesn't exceed the [envelope max sizes](/sdk/data-model/envelopes/#size-limits).
+6. When the size limit is reached, the telemetry buffer **MUST** forward all items to the telemetry scheduler. The buffer **MAY** forward items in batches.
 
 ## Data Forwarding Scenarios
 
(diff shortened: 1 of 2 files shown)


coolguyzone (approved):

develop-docs/sdk/telemetry/telemetry-processor/index.mdx
```diff
@@ -145,8 +145,31 @@ Here are the common requirements for all platforms:
 
 To be defined: How exactly does the telemetry scheduler interact with the transport to correctly remove items from the telemetry queue based on the received HTTP info and how to adhere to the [dealing with network failure](/sdk/expected-features/#dealing-with-network-failures) requirements.
 
+## Client Reports
 
-# Rate Limiting
+The telemetry scheduler **SHOULD** add [client reports](/sdk/telemetry/client-reports/) to outgoing envelopes or periodically send standalone client report envelopes to the transport, following the [SDK side recommendations](/sdk/telemetry/client-reports/#sdk-side-recommendations).
+
+Client report tracking **SHOULD NOT** be implemented in the telemetry scheduler. Instead, the SDK **SHOULD** have `ClientReports` accessible throughout the SDK, since data can be dropped in multiple places, as shown in the diagram below. The telemetry scheduler **SHOULD** request the current client reports from `ClientReports`.
```
giortzisg, line 152: This paragraph is quite vague and it lets many things up to interpretation. I want to make sure I understand this part, because this - to me - reads as, "the scheduler should implement extra logic for periodically sending client reports". 

 I think we need to make it a bit more clear how the `ClientReports` would interact with the scheduler:
- would `ClientReports` handle aggregation?
- would an added report be on a `Buffer` and the scheduler eventually flushes it
- would the scheduler have custom logic to handle a report separately? 

I think the problem originates from the previous Client Reports page being too vague, but it would be nice if we can clear this up a bit.

isaacs: I agree. It seems like this is suggesting the creation of another object in the mix here, a ClientReports system that the telemetry scheduler fetches reports from "periodically".

My approach in the JS SDK would be to keep the `Client.recordDroppedEvent`, as suggested, so that anyone anywhere can throw reports at it, but have the client be a thin pass-through exposing that functionality from the scheduler. So, the Telemetry Scheduler manages and defines how ClientReports are handled, but the Client exposes it to everywhere else (integrations, etc) to access.

philippsuess: > I think the problem originates from the previous Client Reports page being too vague, but it would be nice if we can clear this up a bit.

Yes, indeed. I also struggled a bit with the client reports spec being vague when I initially implemented them on the Cocoa SDK.

> would ClientReports handle aggregation?

I'm not 100% sure what you mean by aggregation, but I guess aggregating individual records of dropped events. So yes, client reports should do that. On Cocoa, we keep a dictionary with the key `reason-category` and then increment the count, see [code](https://github.com/getsentry/sentry-cocoa/blob/387fbe47543733916350d1b5597d38009a8d2325/Sources/Sentry/SentryHttpTransport.m#L158-L174). The [spec](https://develop.sentry.dev/sdk/telemetry/client-reports/#special-case-for-span-outcomes) doesn't clearly specify this, but it could. Do you think such a dict would work for the backend @giortzisg and also JS @isaacs?


> would an added report be on a Buffer and the scheduler eventually flushes it

I don't think it should be on a buffer. Instead, the telemetry scheduler can either periodically flush it or add it to outgoing envelopes. 

> would the scheduler have custom logic to handle a report separately?

Currently, yes, it would need to pull the client reports and prep an envelope. 

> My approach in the JS SDK would be to keep the Client.recordDroppedEvent, as suggested, so that anyone anywhere can throw reports at it, but have the client be a thin pass-through exposing that functionality from the scheduler. So, the Telemetry Scheduler manages and defines how ClientReports are handled, but the Client exposes it to everywhere else (integrations, etc) to access.

We can't only keep the client report / dropped events in the client. The telemetry scheduler also drops events due to queue overflow or rate limits and also the transport also needs to talk to it. So we could hold the object in the client, but we also need to pass it to the telemetry scheduler, telemtry buffer, and the transport. That's why I thought we should create an extra class / object that you can call from multiple places.

giortzisg:
Let's merge this for now, and we can expand later for a more detailed spec.

PR #16005, opened 2026-01-14
fix(metrics): Remove docs for non-existent feature

szokeasaurusrex:
Sentry CLI does not, and nor has it ever provided, a way to query metrics data from Sentry. I don't know how this line ended up in this document, but we should remove it.


szokeasaurusrex:
The mention of the non-existent feature was added in https://github.com/getsentry/sentry-docs/pull/15375. The line seems characteristic of an LLM hallucination; perhaps we should audit the rest of that PR to ensure no other inaccuracies were added there. cc @coolguyzone

PR #16008, opened 2026-01-14
feat(span-first): Remove `active` from `startSpan` options

buenaflor:
## DESCRIBE YOUR PR
Removes `active` from `startSpan`

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

## LEGAL BOILERPLATE

## EXTRA RESOURCES

### DIFF
--- a/develop-docs/sdk/telemetry/spans/span-api.mdx
+++ b/develop-docs/sdk/telemetry/spans/span-api.mdx
@@ -117,7 +117,7 @@ SDKs MUST allow specifying the following options to be passed to `startSpan`:
 Behaviour:
 - Spans MUST be started as active by default. This means that any span started, while the initial span is active, MUST be attached as a child span of the active span.
 - Inactive span behaviour is possible if users pass the parent of the currently active span as `parentSpan` which will attach the started span as a sibling of the currently active span.
-- If a `Span` is passed via `parentSpan`, the span will be started as the child of the passed parent span. This has precedence over the currently active span. 
+- If a `Span` is passed via `parentSpan`, the span will be started as the child of the passed parent span. This has precedence over the currently active span.
 - If `null` is passed via `parentSpan`, the new span will be started as a root/segment span.
 - SDKs MUST NOT end the span automatically. This is the responsibility of the user.
 - `startSpan` MUST always return a span instance, even if the started span's trace is negatively sampled.


develop-docs/sdk/telemetry/spans/span-api.mdx
```diff
@@ -115,11 +113,10 @@ SDKs MUST allow specifying the following options to be passed to `startSpan`:
 | `name`        | Yes      | The name of the span. MUST be set by users   |
 | `attributes`  | No       | Attributes to attach to the span.            |
 | `parentSpan`  | No       | The parent span. See description below for implications of allowed values |
-| `active`      | No       | Whether the started span should be _active_ (i.e. if spans started while this span is active should become children of the started span). |
 
 Behaviour:
-- Spans MUST be started as active by default. This means that any span started, while the initial span is active, MUST be attached as a child span of the active span. 
-- Only if users set `active: false`, the span will be started as inactive, meaning spans started while this span is not yet ended, will not become children, but siblings of the started span.
+- Spans MUST be started as active by default. This means that any span started, while the initial span is active, MUST be attached as a child span of the active span.
+- Inactive span behaviour is possible if users pass the parent of the currently active span as `parentSpan` which will attach the started span as a sibling of the currently active span.
```
buenaflor, line 119: pls double check in case I misunderstood

Lms24 (commented):
Before we merge this I would like to confirm that the replacement is sound. 

With this change, we put more work on users:
* previously, they could mark _one_ span as `active: false`, and all potential spans started while said span was running, would be attached to the active span (i.e. not the one marked)
* now, with the proposed change, _every_ span that might be started while the span that should be inactive is running, must explicitly set another parent. This gets especially tricky if we're talking about auto instrumentation. 

I would argue there are good reasons for still starting individual inactive spans without having to care about others. 

Pseudocode example

```ts
// Assume: fetch calls are instrumented automatically, creating `http.client` spans

const myParentSpan = Sentry.startSpan({name: 'parent'});

fetch('/resource-1').then(...);
fetch('/resource-2').then(...);

myParentSpan.end()
```

If the fetch spans are _not_ started as inactive, we end up with this hierarchy:
- parent
  - fetch resource-1
    - fetch resource-2
 
and there's little we can do about that, given fetch calls are auto instrumented.

While if auto instrumentation can start fetch spans with `active: false`, we get the "correct" hierarchy:
- parent
  - fetch resource-1
  - fetch resource-2

Fwiw, I think `http.client` spans are good examples for inactive spans, since there's rarely a reason to attach a child span to them (if ever?). We currently start `http.client` spans as inactive spans in JS.

develop-docs/sdk/telemetry/spans/span-api.mdx
Lms24: I think you're correct here in terms of how we discussed it. But I realized that this might not be what we want. See my comment below

cleptric:
Android/Java has `http.client` child spans.

develop-docs/sdk/telemetry/spans/span-api.mdx
sfanahata:
```suggestion
- Inactive span behaviour is possible if users pass the parent of the currently active span as `parentSpan`, which will attach the started span as a sibling of the currently active span.
```

sfanahata (approved):
Looks good from a docs perspective, but sounds like there's more work to do here?

cleptric:
Yes, sorry 😬 Moved back to draft, still needs to cook a bit longer.

PR #16011, opened 2026-01-14
feat(replay): Document skipRequestAnimationFrame option for WebGPU canvas snapshots

andreiborza:
Adds documentation for the `skipRequestAnimationFrame` option when using manual canvas snapshots with WebGPU.

WebGPU textures expire after the current task completes, so deferring snapshot capture to the next animation frame results in an empty canvas. This option captures immediately after rendering.

Sources:
- https://gpuweb.github.io/gpuweb/explainer/
- https://github.com/gpuweb/gpuweb/issues/2743
- https://github.com/gpuweb/gpuweb/issues/3295

Related: https://github.com/getsentry/sentry-javascript/issues/18804

### DIFF
--- a/docs/platforms/javascript/common/session-replay/index.mdx
+++ b/docs/platforms/javascript/common/session-replay/index.mdx
@@ -103,9 +103,21 @@ function paint() {
 
 #### WebGPU Canvases
 
-WebGPU works differently from WebGL in how it manages the lifetime of rendered frames. The canvas texture returned by `getCurrentTexture()` is only valid until the current task completes – after that, the texture expires, and its contents are no longer accessible. This means that by default, when `snapshot()` defers capture to the next animation frame, it will capture an empty or invalid canvas.
+WebGPU works differently from WebGL in how it manages the lifetime of rendered frames. The canvas texture returned by `getCurrentTexture()` is only valid until the current task completes – after that, the texture expires, and its contents are no longer accessible. This means that by default, when `snapshot()` defers capture to the next animation frame, it will capture an empty or invalid canvas. There are two steps to capturing WebGPU canvases:
 
-To fix this, pass `skipRequestAnimationFrame: true` to capture the snapshot immediately:
+**Step 1.**
+Enable manual snapshotting when initializing the `ReplayCanvas` integration.
+
+```javascript
+Sentry.replayCanvasIntegration({
+  // Enabling the following will ensure your canvas elements are not forced
+  // into `preserveDrawingBuffer`.
+  enableManualSnapshot: true,
+});
+```
+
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/common/session-replay/index.mdx
```diff
@@ -100,6 +100,24 @@ function paint() {
   Sentry.getClient().getIntegrationByName("ReplayCanvas").snapshot(canvasRef);
 }
 ```
+
+#### WebGPU Canvases
+
+WebGPU works differently from WebGL. The canvas texture returned by `getCurrentTexture()` is only valid until the current task completes - after that, the texture expires and its contents are no longer accessible. This means that by default, when `snapshot()` defers capture to the next animation frame, it will capture an empty or invalid canvas.
+
+To fix this, pass `skipRequestAnimationFrame: true` to capture the snapshot immediately:
```
inventarSarah: Does this also require setting `enableManualSnapshot: true`? If so, we could add the 2-step explanation as in the WebGL section to make sure users don't miss it.

docs/platforms/javascript/common/session-replay/index.mdx
```diff
@@ -100,6 +100,24 @@ function paint() {
   Sentry.getClient().getIntegrationByName("ReplayCanvas").snapshot(canvasRef);
 }
 ```
+
+#### WebGPU Canvases
+
+WebGPU works differently from WebGL. The canvas texture returned by `getCurrentTexture()` is only valid until the current task completes - after that, the texture expires and its contents are no longer accessible. This means that by default, when `snapshot()` defers capture to the next animation frame, it will capture an empty or invalid canvas.
```
inventarSarah:
```suggestion
WebGPU works differently from WebGL in how it manages the lifetime of rendered frames. The canvas texture returned by `getCurrentTexture()` is only valid until the current task completes – after that, the texture expires, and its contents are no longer accessible. This means that by default, when `snapshot()` defers capture to the next animation frame, it will capture an empty or invalid canvas.
```
I tried to add more context to the first sentence -- not sure what differences you want to highlight here

inventarSarah (commented):
Looks good to me overall. 
Just two comments from my side -- please have a look :)

docs/platforms/javascript/common/session-replay/index.mdx
andreiborza: Right, yea it does. Good catch!

docs/platforms/javascript/common/session-replay/index.mdx
andreiborza: Thanks, that makes it clearer!

PR #16013, opened 2026-01-14
fix(dsn): Hide double scrollbar in dsn picker

scttcper:
scroll always displays scroll bars when using windows or a mouse on mac. We can use auto and not display the bottom scroll bar.

Screenshot 2026-01-14 at 10 30 51 AM


sfanahata:
@scttcper - is there a way to preview this behavior? I can't seem to find a way to do it in the preview for DSN selector.

scttcper:
@sfanahata  i don't think it works in the preview unfortunately. but scroll generally always displays scroll bars and auto shows them when they're needed.

PR #16026, opened 2026-01-15
docs(js): Add more specific PII info on RequestDataIntegration

chargome:


includes/platforms/configuration/integrations/requestdata-v11-alert.mdx
```diff
@@ -0,0 +1,5 @@
+<Alert level="warning" title="Upcoming Default Changes in v11">
+
+In version 11, the default behavior of the RequestData integration will change to be more privacy-conscious. Fields like `cookies`, `data`, `headers`, `query_string`, and `user` will default to `false` instead of `true`. To continue capturing this data after upgrading to v11, you'll need to either explicitly configure the <PlatformLink to="/configuration/integrations/requestdata/">RequestData Integration</PlatformLink> or set <PlatformLink to="/configuration/options/#sendDefaultPii">`sendDefaultPii: true`</PlatformLink>.
```
JPeer264: Is this now the decision that it will be `false` by default? I would be more comfy if we don't add this info just yet, as I am personally not sure anymore what will be the default

chargome: I'd leave this in for now, so we have something public facing that this will eventually change. If some details change in v11 we can always adapt the docs.

JPeer264 (approved):

PR #16031, opened 2026-01-15
feat(electron): Add `startupTracingIntegration` to main process tracing

timfish:
This PR adds the `startupTracingIntegration` to the Electron `init` example options when users select tracing enabled.

image


timfish:
The lint-404 failures seem unrelated to this PR...

PR #16048, opened 2026-01-16
ref(docs): Restructure LangChain integration documentation

priscilawebdev:
- Separate server-side and browser-side usage sections for clarity
- Document langChainIntegration for Node.js platforms (auto-enabled by default)
- Document createLangChainCallbackHandler for browser and edge runtimes
- Add special handling for Cloudflare Workers and Next.js Edge runtime
- Simplify code examples with placeholder comments
- Restructure Configuration section with options first, then usage examples
- Improve platform-specific visibility using PlatformSection components

closes https://linear.app/getsentry/issue/TET-1740/langchain-integration-restructure-and-improve-docs

### DIFF
--- a/docs/platforms/javascript/common/configuration/integrations/langchain.mdx
+++ b/docs/platforms/javascript/common/configuration/integrations/langchain.mdx
@@ -115,7 +115,7 @@ To customize what data is captured (such as inputs and outputs), see the [Option
 
 ## Edge runtime
 
-This integration is automatically instrumented in the Node.js runtime. For Next.js applications using the Edge runtime, you need to manually instrument LangChain operations using the callback handler:
+For Next.js applications using the Edge runtime, you need to manually instrument LangChain operations using the callback handler:
 
 ```javascript
 import * as Sentry from "@sentry/nextjs";


priscilawebdev:
> Edge runtime should have it's own section: https://sentry-docs-rc5wqv0cx.sentry.dev/platforms/javascript/guides/nextjs/configuration/integrations/langchain/#browser-side-usage

Since we already have an alert in place for runtimes in general, I believe that should be sufficient for Edge as well. That said, I don’t have strong feelings on this, so I’m fine with adding it back.

Out of curiosity, is there a specific reason this section exists only for LangChain? I don’t see a similar section for OpenAI, for example.

RulaKhaled:
> > Edge runtime should have it's own section: https://sentry-docs-rc5wqv0cx.sentry.dev/platforms/javascript/guides/nextjs/configuration/integrations/langchain/#browser-side-usage
> 
> Since we already have an alert in place for runtimes in general, I believe that should be sufficient for Edge as well. That said, I don’t have strong feelings on this, so I’m fine with adding it back.
> 
> Out of curiosity, is there a specific reason this section exists only for LangChain? I don’t see a similar section for OpenAI, for example.

The alert is nested within the browser runtime so i think it is confusing for folks looking at Edge runtime as they are two separate concepts. We can either move the alert outside of the browser section or create it's own section (which is what i recommend). 

We do have it for other AI integrations: https://docs.sentry.io/platforms/javascript/guides/nextjs/configuration/integrations/google-genai/#edge-runtime

priscilawebdev:
> > > Edge runtime should have it's own section: https://sentry-docs-rc5wqv0cx.sentry.dev/platforms/javascript/guides/nextjs/configuration/integrations/langchain/#browser-side-usage
> > 
> > 
> > Since we already have an alert in place for runtimes in general, I believe that should be sufficient for Edge as well. That said, I don’t have strong feelings on this, so I’m fine with adding it back.
> > Out of curiosity, is there a specific reason this section exists only for LangChain? I don’t see a similar section for OpenAI, for example.
> 
> The alert is nested within the browser runtime so i think it is confusing for folks looking at Edge runtime as they are two separate concepts. We can either move the alert outside of the browser section or create it's own section (which is what i recommend).

Yes, I removed that previously. We now have a single general alert for other runtimes, and “edge” is considered one of these. It’s displayed under server-side, and for Next.js, it’s always shown since we display both server-side and browser-side information for meta frameworks. That said, I’m happy to keep a separate dedicated section if you feel it would improve discoverability. Thanks!

docs/platforms/javascript/common/configuration/integrations/langchain.mdx
```diff
@@ -25,76 +25,106 @@ supported:
...
+
 </PlatformSection>
 
-<PlatformSection supported={["javascript"]}>
+<PlatformSection supported={['javascript.nextjs']}>
+
+## Edge runtime
 
-For browser applications, you need to manually instrument LangChain operations using the `createLangChainCallbackHandler` helper:
+This integration is automatically instrumented in the Node.js runtime. For Next.js applications using the Edge runtime, you need to manually instrument LangChain operations using the callback handler:
```
RulaKhaled: you can remove `This integration is automatically instrumented in the Node.js runtime.` from this callout. Leaving this up to you.

PR #16076, opened 2026-01-19
docs(nextjs): document the route manifest injection option

logaretm:
This document outlines the new option introduced in https://github.com/getsentry/sentry-javascript/pull/18798, which was released in `10.34.0`.

This option replaces the `disableManifestinjection` option, providing users with the flexibility to either completely disable this behavior or selectively exclude specific routes from being included in the route manifest.

A common use case for this option is when some routes are private or contain unreleased features. Having these routes present in the route manifest posed challenges for certain users. To address this, an `exclude` option allows users to avoid the all-or-nothing compromise, ensuring that data quality is maintained and preventing the leakage of route information.

### DIFF
--- a/docs/platforms/javascript/guides/nextjs/configuration/build/index.mdx
+++ b/docs/platforms/javascript/guides/nextjs/configuration/build/index.mdx
@@ -242,20 +242,32 @@ You can set this option to `false` to disable the route manifest injection.
 - You're experiencing build issues related to route scanning
 - You prefer raw URLs in transaction names
 
-You can also pass in an object with an `exclude` property to control which routes should be excluded from the route manifest:
+You can also pass in an object with an `exclude` property to control which routes should be excluded from the route manifest. The `exclude` property accepts an array of strings or regular expressions, or a function that returns `true` to exclude a route.
 
 ```javascript
-routeManifestInjection: {
-  exclude: ["/api/health", "/api/excluded/[parameter]", /^\/internal\//],
-  // or a function that returns `true` to exclude the route
-  exclude: (route) => route.includes("/excluded"),
-},
+withSentryConfig(nextConfig, {
+  // Exclude specific routes using an array of strings or RegExps
+  routeManifestInjection: {
+    exclude: ["/api/health", "/api/excluded/[parameter]", /^\/internal\//],
+  },
+});
 ```
 
-**Use this if:**
+```javascript
+withSentryConfig(nextConfig, {
+  // Or use a function for dynamic exclusion
+  routeManifestInjection: {
+    exclude: (route) => route.includes("/excluded"),
+  },
+});
+```
 
-- You want to exclude routes that are not yet released or should not be publicly accessible.
-- You want to reduce the size of the route manifest by excluding routes that do not require parameterized grouping.
+Excluded routes will appear as raw URLs in transaction names instead of parameterized routes.
+
+**Use `exclude` if:**
+
+- You want to hide internal or unreleased routes from appearing in the client bundle
+- You want to reduce bundle size by excluding routes that don't benefit from parameterized grouping (for example, static routes with no dynamic segments)
 
 </SdkOption>
 


docs/platforms/javascript/guides/nextjs/configuration/build/index.mdx
```diff
@@ -222,19 +222,35 @@ Enables the use of the [`runAfterProductionCompile` hook from Next.js](https://n
...
 
-<SdkOption name="disableManifestInjection" type="boolean|undefined" defaultValue="false">
+<SdkOption name="routeManifestInjection" type="boolean|object" defaultValue="true">
 
-Disables automatic injection of the route manifest into the client bundle.
+<AvailableSince version="10.34.0" />
+
+Controls the behavior of the route manifest injection into the client bundle.
 
 The route manifest is a build-time generated mapping of your Next.js App Router routes that enables Sentry to group transactions by parameterized route names (e.g., `/users/:id` instead of `/users/123`, `/users/456`, etc.).
```
inventarSarah:
```suggestion
The route manifest is a build-time generated mapping of your Next.js App Router routes that enables Sentry to group transactions by parameterized route names (for example, `/users/:id` instead of `/users/123` or `/users/456`).
```
-> just a bit cleaner this way

docs/platforms/javascript/guides/nextjs/configuration/build/index.mdx
```diff
@@ -222,19 +222,35 @@ Enables the use of the [`runAfterProductionCompile` hook from Next.js](https://n
 
 </SdkOption>
 
-<SdkOption name="disableManifestInjection" type="boolean|undefined" defaultValue="false">
+<SdkOption name="routeManifestInjection" type="boolean|object" defaultValue="true">
 
-Disables automatic injection of the route manifest into the client bundle.
+<AvailableSince version="10.34.0" />
+
+Controls the behavior of the route manifest injection into the client bundle.
```
inventarSarah:
```suggestion
Controls injection and filtering of the route manifest in the client bundle.
```
-> Makes the description a bit more explicit

docs/platforms/javascript/guides/nextjs/configuration/build/index.mdx
```diff
@@ -222,19 +222,35 @@ Enables the use of the [`runAfterProductionCompile` hook from Next.js](https://n
...
+
+```javascript
+routeManifestInjection: {
+  exclude: ["/api/health", "/api/excluded/[parameter]", /^\/internal\//],
+  // or a function that returns `true` to exclude the route
+  exclude: (route) => route.includes("/excluded"),
+},
+```
+
+Excluding specific routes can be useful if you have routes with unreleased features or private routes that you don't want to be included in the route manifest.
```
inventarSarah: It would be great to also have a list here like above; for example, "Use this if:"

docs/platforms/javascript/guides/nextjs/configuration/build/index.mdx
```diff
@@ -222,19 +222,35 @@ Enables the use of the [`runAfterProductionCompile` hook from Next.js](https://n
...
+You can set this option to `false` to disable the route manifest injection.
+
 **Disable this option if:**
 
 - You want to minimize client bundle size
 - You're experiencing build issues related to route scanning
 - You prefer raw URLs in transaction names
 - You're only using the Pages Router (this feature is only supported in the App Router)
 
+You can also pass in an object to control which routes should be excluded from the route manifest:
```
inventarSarah: Maybe we could add here that the object supports `exclude` so that users know what's supported without relying on the example code.
Or, if there's more, we could list them like here: https://sentry-docs-git-awad-nextjs-route-manifest-injection-doc.sentry.dev/platforms/javascript/guides/nextjs/configuration/options/#transportOptions

docs/platforms/javascript/guides/nextjs/configuration/build/index.mdx
```diff
@@ -222,19 +222,35 @@ Enables the use of the [`runAfterProductionCompile` hook from Next.js](https://n
...
 The route manifest is a build-time generated mapping of your Next.js App Router routes that enables Sentry to group transactions by parameterized route names (e.g., `/users/:id` instead of `/users/123`, `/users/456`, etc.).
 
+You can set this option to `false` to disable the route manifest injection.
+
 **Disable this option if:**
 
 - You want to minimize client bundle size
 - You're experiencing build issues related to route scanning
 - You prefer raw URLs in transaction names
 - You're only using the Pages Router (this feature is only supported in the App Router)
```
inventarSarah: "this feature is only supported in the App Router" -> should we mention this earlier?

logaretm: It was part of the old description we had but yea I think it should be right under the section title, will update it. Thanks!

docs/platforms/javascript/guides/nextjs/configuration/build/index.mdx
logaretm: Right now there is only one option, it's an object for some future proofing, but yea "an object" doesn't really tell the user what to expect here.

docs/platforms/javascript/guides/nextjs/configuration/build/index.mdx
logaretm: Yea, I think I can come up with a couple of list items.

PR #16093, opened 2026-01-21
feat(develop): Add `new_trace()`

sentrivana:
Add API to clear the propagation context. Add some examples, too.

**Preview:** https://develop-docs-git-ivana-span-firstnew-trace.sentry.dev/sdk/telemetry/spans/span-trace-propagation/

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

## LEGAL BOILERPLATE

## EXTRA RESOURCES

### DIFF
--- a/develop-docs/sdk/telemetry/spans/span-trace-propagation.mdx
+++ b/develop-docs/sdk/telemetry/spans/span-trace-propagation.mdx
@@ -55,11 +55,11 @@ The SDK MUST offer a method to clear trace propagation data, allowing to create 
 
 ```js
 Sentry.startNewTrace(() => {
-  Sentry.startSpan('segment under trace 1');
+  Sentry.startSpan({ name: 'segment under trace 1' }, () => {...});
 })
 
 Sentry.startNewTrace(() => {
-  Sentry.startSpan('segment under trace 2');
+  Sentry.startSpan({ name: 'segment under trace 2' }, () => {...});
 })
 ```
 


develop-docs/sdk/telemetry/spans/span-trace-propagation.mdx
```diff
@@ -12,14 +12,21 @@ title: Span Trace Propagation
...
     baggage: request.headers.SENTRY_BAGGAGE,
 })
 ```
+
+```python
+import sentry_sdk
+
+sentry_sdk.continue_trace(request.headers)
+```
+
```
cleptric, line 34: I think we should align this across SDKs. The snippets here currently do not related to any implementation but we should pick one and document it via pseudo code.

sentrivana: @Lms24 Is the JS example up to date? I have two reservations regarding it:
- It's not top-level API, but scope API, which means users would need to be aware of the scope system and pick the correct scope to put the propagation context on. Given that even we ourselves struggle with this 🤣 I don't think that's a good idea.
- It's not immediately clear what it (a propagation context) does. Compare this with e.g. `continueTrace` which is intuitive.

Could we align on `continue_trace()`/`continueTrace()` or something similar?

sentrivana: Related, I didn't touch these in this PR but I have similar concerns about
```
traceparent = scope.getTraceparent()
baggage = scope.getBaggage()
```
Those should ideally also not have anything to do with a specific scope, but rather be top level, and we pick the correct scope to retrieve them from in the background.
```
traceparent = sentry_sdk.get_traceparent()
baggage = sentry_sdk.get_baggage()
```

cleptric: I'm ok with changing this to use global functions.

Lms24: Sorry, these snippets (at the time of writing) were meant to be low-level pseudocode (shouldn't have used the javascript langauge for this). In JS, today, we have `Sentry.continueTrace`:

```js
Sentry.continueTrace({
  sentryTrace: request.headers['sentry-trace'],
  baggage: request.headers['baggage'],
}, () => {
  Sentry.startSpan({ name: 'test' }, () => {
    // ....
  });
})
```

Which, besides the callback, is very similar to what is being proposed here IIUC.

develop-docs/sdk/telemetry/spans/span-trace-propagation.mdx
```diff
@@ -12,14 +12,21 @@ title: Span Trace Propagation
...
 scope.setPropagationContext({
     traceparent: request.headers.SENTRY_TRACE,
     baggage: request.headers.SENTRY_BAGGAGE,
 })
 ```
+
+```python
+import sentry_sdk
+
+sentry_sdk.continue_trace(request.headers)
```
Lms24, line 29: l: This might be too specific for the spec but I want to mention it anyway: I'd prefer defining the options to be passed in here more explicitly than a request object. We might also want to use this API in non-request contexts (e.g. queues) or when continuing traces via env variables, so I'd rather pass in sentry-trace and baggage as "neutral" options. 

We could still expose something more ergonomic for requests in addition, say `sentry_sdk.continue_trace_from_request(request)` but this doesn't have to be spec'd out IMHO.

develop-docs/sdk/telemetry/spans/span-trace-propagation.mdx
```diff
@@ -30,3 +37,20 @@ To propagate a trace to a downstream service, the SDK must expose methods to fet
...
+
+with sentry_sdk.start_span(name="segment under trace 1"):
+    ...
+
+sentry_sdk.new_trace()
+
+# The next span started after new_trace() will NOT continue the same trace:
+with sentry_sdk.start_span(name="segment under trace 2"):
+    ...
+```
```
Lms24, line 77: I can't suggest this directly due to the code block in the suggestion code block but if you want to, we can include the JS SDK's already available APIs as well. They're very similar, besides the slightly different name and the callback.

```javascript
Sentry.startNewTrace(() => {
  Sentry.startSpan('segment under trace 1');
})

Sentry.startNewTrace(() => {
  Sentry.startSpan('segment under trace 2');
})
```

develop-docs/sdk/telemetry/spans/span-trace-propagation.mdx
sentrivana: In Python, the existing (non-streaming) `continue_trace` simply accepts a dictionary which can either be request headers, queue headers or environment vars and the propagation context takes care of normalizing it and extracting propagation data from it.

I'd keep this for span first since it's what folks are used to and it's very convenient to have the SDK do the extraction for you.

I can include a blurb about making the args to this whatever makes sense canonically for the specific SDK.

develop-docs/sdk/telemetry/spans/span-trace-propagation.mdx
sentrivana: Awesome, will update.

develop-docs/sdk/telemetry/spans/span-trace-propagation.mdx
Lms24: One more thing:

> traceparent = sentry_sdk.get_traceparent()
baggage = sentry_sdk.get_baggage()

Sounds fine to me! We went down a slightly different route in JS and exported a `Sentry.getTraceData()` API which returns a dict/object with both headers (actually, 3, if `propagateTraceParent` is enabled). Our thinking back then was, there's no reason to only need one of these headers, so let's take care of them together. Not saying we should generalize this because the two separate ones can be combined and built upon. Just pointing out prior art 😅

sentrivana: @Lms24 Since I'm already updating all the examples, there's
```
traceparent = scope.getTraceparent()
baggage = scope.getBaggage()
```
How does this actually look in JS? I found `getTraceData()`, is that it?

sentrivana: Haha, commented about the same thing at the same time 😅

sentrivana: Ok thank you both, updated if anyone wants to give this a last pass.

develop-docs/sdk/telemetry/spans/span-trace-propagation.mdx
```diff
@@ -12,21 +12,66 @@ title: Span Trace Propagation
...
 
 ```js
-traceparent = scope.getTraceparent()
-baggage = scope.getBaggage()
+Sentry.startNewTrace(() => {
+  Sentry.startSpan('segment under trace 1');
+})
+
+Sentry.startNewTrace(() => {
+  Sentry.startSpan('segment under trace 2');
```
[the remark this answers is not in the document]
Lms24: argh seer is right, sorry about that 😬 

```js
Sentry.startNewTrace(() => {
  Sentry.startSpan({ name: 'segment under trace 1' }, () => {...});
})

Sentry.startNewTrace(() => {
  Sentry.startSpan({ name: 'segment under trace 2' }, () => {...});
})
```

cc @sentrivana (see https://github.com/getsentry/sentry-docs/pull/16093#discussion_r2713507657)

develop-docs/sdk/telemetry/spans/span-trace-propagation.mdx
```diff
@@ -12,21 +12,66 @@ title: Span Trace Propagation
...
+})
```
Lms24, line 63: at least this time I can suggest the edit. Sorry for the back and forth! (see https://github.com/getsentry/sentry-docs/pull/16093/changes#r2713471352)

```suggestion
Sentry.startNewTrace(() => {
  Sentry.startSpan({ name: 'segment under trace 1' }, () => {...});
})

Sentry.startNewTrace(() => {
  Sentry.startSpan({ name: 'segment under trace 2' }, () => {...});
})
```

sentrivana: No worries, ty!

PR #16101, opened 2026-01-21
fix(DOCS-A0Y): Handle missing MDX files gracefully during serverless cold starts

sfanahata:
## Summary

Fixes DOCS-A0Y - handles missing MDX files gracefully during Vercel serverless cold starts instead of throwing errors.

## The Issue & Impact

**What was happening:**
- During Vercel deployments and serverless cold starts, the server would try to read MDX doc files that aren't in the serverless bundle (they're excluded to stay under Vercel's 250MB size limit)
- This caused ~4,700 errors in Sentry

**User Experience Impact:** None
- Real users get pre-built static pages from the CDN
- The serverless function code path is only hit during infrastructure transitions
- 0 users were impacted by these errors

**Developer Experience Impact:** Significant noise
- Sentry Issues dashboard flooded with false-positive errors
- Made it harder to spot real problems

## The Solution

1. **Graceful error handling** - Instead of throwing an error, we now return Next.js's `notFound()` which shows a helpful 404 page with search (in the unlikely event a user ever hits this)

2. **Info-level logging via Sentry** - We log these occurrences as info (not errors) so you can:
   - Keep your Issues dashboard clean
   - Still have visibility in Sentry Logs if you want to monitor
   - See context like the path and reason (`serverless_bundle_exclusion`)

## Testing

- TypeScript compilation passes
- ESLint passes
- Error handling logic tested for various error types (ENOENT code, message matching, unrelated errors)

### DIFF
--- a/app/[[...path]]/page.tsx
+++ b/app/[[...path]]/page.tsx
@@ -123,19 +123,21 @@ export default async function Page(props: {params: Promise<{path?: string[]}>}) 
     try {
       doc = await getFileBySlugWithCache(`develop-docs/${params.path?.join('/') ?? ''}`);
     } catch (e: unknown) {
-      // Handle file not found errors gracefully - this can happen when:
-      // 1. Serverless function is invoked at runtime but docs files aren't in the bundle
-      // 2. Build-time issues with file resolution
-      // Check for ENOENT in multiple ways to be defensive
-      const isNotFound =
-        (e && typeof e === 'object' && 'code' in e && e.code === 'ENOENT') ||
+      // Handle file not found and runtime MDX compilation errors gracefully.
+      // This can happen when serverless function is invoked at runtime but docs files
+      // aren't in the bundle, or MDX compilation is attempted at Vercel runtime.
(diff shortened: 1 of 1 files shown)


app/[[...path]]/page.tsx
```diff
@@ -121,10 +122,22 @@ export default async function Page(props: {params: Promise<{path?: string[]}>})
...
+      // 2. Build-time issues with file resolution
+      // Check for ENOENT in multiple ways to be defensive
+      const isNotFound =
+        (e && typeof e === 'object' && 'code' in e && e.code === 'ENOENT') ||
+        (e instanceof Error && e.message.includes('Failed to find a valid source file'));
+      if (isNotFound) {
+        // Log as info (not error) for visibility without noise
+        // This typically happens during serverless cold starts when docs files
+        // aren't in the bundle - users are served static pages from CDN instead
+        Sentry.logger.info('MDX file not found at runtime, returning 404', {
```
Comment (sergical) [inline app/[[...path]]/page.tsx]: would probably do a warn here instead of info

sergical (approved):

PR #16116, opened 2026-01-22
fix(logs): Buffer size limit

philipphofmann:
Set the logs buffer size limit to 100 instead of 1000, as pointed out in https://github.com/getsentry/sentry-docs/pull/16114#discussion_r2716000375.

### DIFF
--- a/develop-docs/sdk/telemetry/logs.mdx
+++ b/develop-docs/sdk/telemetry/logs.mdx
@@ -400,7 +400,9 @@ A new data category for logs has been added to Relay, `log_item`. Both the `log`
 
 SDKs **MUST** buffer logs before sending them. SDKs should keep a buffer of logs and flush it when specific conditions are met. For initial implementation, a simple strategy is fine, for example: flushing logs if the buffer length exceeds 100 items or if 5 seconds have passed. To prevent data loss, the buffer SHOULD forward logs to the transport in the scenarios outlined in the [telemetry buffer data forwarding scenarios](/sdk/telemetry/telemetry-buffer/#data-forwarding-scenarios).
 
-SDKs **MUST** have a hard limit of 1000 log events queued up to avoid causing your customers' applications going out of memory. Logs added once this hard limit has been reached are dropped. SDKs **MAY** use lower values, but they **MUST NOT** exceed 1000. Relay is optimized for up to 100 logs per envelope. SDKs **MUST NOT** send more than 100 logs per envelope.
+Relay is optimized for up to 100 logs per envelope. SDKs **MUST NOT** send more than 100 logs per envelope.
+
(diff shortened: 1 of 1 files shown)


develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -400,7 +400,7 @@ A new data category for logs has been added to Relay, `log_item`. Both the `log`
...
 
-SDKs **MUST** also have a hard limit of 1000 log events queued up to avoid causing your customers' applications going out of memory. Logs added once this hard limit has been reached are dropped. Also, Relay has a hard limit of 1000 logs per envelope. SDKs **MUST** not exceed this limit. SDKs **SHOULD**  batch logs into envelopes of 100 or less.
+SDKs **MUST** have a hard limit of 100 log events queued up to avoid causing your customers' applications going out of memory. Logs added once this hard limit has been reached are dropped. SDKs **MAY** use lower values, but they **MUST NOT** exceed 100. Also, Relay is optimized for 100 logs or less per envelope. SDKs **MUST** not exceed this limit when sending log envelopes.
```
cleptric: This is not correct. The log buffer can hold 1000 logs, we just only emit 100 per envelope at a time.

JoshuaMoelans: @cleptric is the idea to have it hold a **maximum** of 1000 logs to avoid OOM, so if our SDK can do with less (e.g., a double-buffer approach with 2x100-log queues) without dropping logs, that is fine too?

cleptric: The 1000 limit is mostly in case we can't create the envelope fast enough prior receiving more logs. So instead of dropping them, we allow them to accumulate. And yes, this is purely a limit to avoid OOM. If native caps this at 200, I think it's ok.

develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -400,7 +400,7 @@ A new data category for logs has been added to Relay, `log_item`. Both the `log`
...
 
-SDKs **MUST** also have a hard limit of 1000 log events queued up to avoid causing your customers' applications going out of memory. Logs added once this hard limit has been reached are dropped. Also, Relay has a hard limit of 1000 logs per envelope. SDKs **MUST** not exceed this limit. SDKs **SHOULD**  batch logs into envelopes of 100 or less.
+SDKs **MUST** have a hard limit of 1000 log events queued up to avoid causing your customers' applications going out of memory. Logs added once this hard limit has been reached are dropped. SDKs **MAY** use lower values, but they **MUST NOT** exceed 1000. Relay is optimized for up to 100 logs per envelope. SDKs **MUST NOT** send more than 100 logs per envelope.
```
cleptric:
```suggestion
Relay is optimized for up to 100 logs per envelope. SDKs **MUST NOT** send more than 100 logs per envelope.

To prevent out-of-memory issues in customer applications, SDKs **MUST** enforce a hard limit of 1000 queued log events. Any logs added beyond this limit are dropped. SDKs **MAY** use a lower limit but **MUST NOT** exceed 1000.
```

PR #16137, opened 2026-01-23
chore: add claude skill for reviewing docs

shellmayr:
- This adds a skill that enables claude to review docs PRs for style & language using specific guidelines
- We tried this out and found that it was very correct in what it suggested, may be useful for conducting a local self-review before having other developers read through documentation

### DIFF
--- a/.claude/skills/docs-review/SKILL.md
+++ b/.claude/skills/docs-review/SKILL.md
@@ -1,0 +1,54 @@
+---
+name: docs-review
+description: Review documentation for Sentry style and voice. Use when reviewing or writing user-facing documentation to ensure it follows the Sentry documentation style guide.
+---
+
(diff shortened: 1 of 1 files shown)


dingsdax:
@shellmayr I found a related skill that’s an interesting application for ensuring a specific reading skill level: https://github.com/nateberkopec/dotfiles/blob/fa30f09f01fc50e6a49fd3db9a8e7626e440c2c1/files/home/.claude/skills/readme-writer/SKILL.md.

shellmayr:
@dingsdax that's also interesting! I guess we could also ask it to measure flesch score and optimize towards that, but I honestly wouldn't go that far. I've tried some of that in the past, and it's a bit of a rabbit hole to try to optimize for reading levels, especially because the models definitely don't one-shot that

PR #16147, opened 2026-01-26
ref(docs): Use dynamic version for Android Gradle plugin in upload instructions (EME-758)

runningcode:
Update the size analysis and build distribution documentation to use dynamic version fetching for the Sentry Android Gradle plugin instead of a hardcoded version.

Fixes EME-758

### DIFF
--- a/src/components/gradleUploadInstructions.tsx
+++ b/src/components/gradleUploadInstructions.tsx
@@ -19,7 +19,7 @@ export async function GradleUploadInstructions({feature}: Props) {
 
   const packageRegistry = await getPackageRegistry();
   const gradlePluginVersion =
-    packageRegistry.data?.['sentry.java.android.gradle-plugin']?.version || '7.12.3';
+    packageRegistry.data?.['sentry.java.android.gradle-plugin']?.version || '6.0.0';
 
   return (
     <Fragment>


src/components/gradleUploadInstructions.tsx
```diff
@@ -12,9 +14,13 @@ type Props = {
   feature: 'sizeAnalysis' | 'distribution';
 };
 
-export function GradleUploadInstructions({feature}: Props) {
+export async function GradleUploadInstructions({feature}: Props) {
   const featureName = feature === 'sizeAnalysis' ? 'size analysis' : 'distribution';
 
+  const packageRegistry = await getPackageRegistry();
```
runningcode, line 20: the mdx injection doesn't work here since this is a tsx file so this is the best solution i could come up with.

this code is used for size analysis and build distribution

chromy:
I think we need to add something else if we do this: https://github.com/getsentry/sentry-docs/blob/77e43a0beefef9cd2d1cc15482434e7005743a45/src/mdx.ts#L672

but maybe we can use: 
https://github.com/getsentry/sentry-docs/blob/77e43a0beefef9cd2d1cc15482434e7005743a45/src/components/platformSdkPackageName.tsx#L6 ? (or something very similar but that gets the version).

runningcode:
Nice catch @chromy! I've updated the PR

PR #16230, opened 2026-01-30
Add the 0.0.0.0 address to the proxy configuration

DragoonAethis:
## DESCRIBE YOUR PR

- `0.0.0.0` is the default socket bind address for multiple services (like relay, symbolicator, etc) in their template configuration files.
- When http(s)_proxy is set, health checks have issues for some people (errors like Backend Unreachable, etc).
- Health checks for those services are not integrated in the running service itself - they're separate processes (just executed from the same binary) that execute in the service container and connect to the running service over HTTP.
- Health checks defined in `docker-compose.yml` default to connecting to the service instance address from the service config.
- Which defaults to something like `0.0.0.0:3000`.
- Which is not in no_proxy.
- So the health check process tries to access the service via the proxy.
- Which fails.
- Which results in "Up 5 minutes (unhealthy)" despite the affected services working correctly.
- Except that `docker compose up --wait` waits for all services to be healthy. So that fails too.
- Also the `nginx` service depends on `web` (which works) and `relay` (which works, but reports itself as unhealthy), so the frontend never starts.

And to fix all that, just add `0.0.0.0` to `noProxy` here and we're gucci.

There are a couple of folks with weird health check issues on the self-hosted repo - it's likely they will benefit from this change too.

## PRE-MERGE CHECKLIST

### DIFF
--- a/develop-docs/self-hosted/index.mdx
+++ b/develop-docs/self-hosted/index.mdx
@@ -143,7 +143,7 @@ By default Sentry sends anonymous usage statistics to the Sentry team. It helps 
     "default": {
       "httpProxy": "http://proxy:3128",
       "httpsProxy": "http://proxy:3128",
-      "noProxy": "localhost,smtp,memcached,redis,postgres,pgbouncer,kafka,clickhouse,seaweedfs,snuba-api,symbolicator,web,worker,nginx,relay,vroom,taskbroker,172.17.0.0/16,127.0.0.0/8"
+      "noProxy": "localhost,smtp,memcached,redis,postgres,pgbouncer,kafka,clickhouse,seaweedfs,snuba-api,symbolicator,web,worker,nginx,relay,vroom,taskbroker,0.0.0.0,172.17.0.0/16,127.0.0.0/8"
     }
   }
 }


DragoonAethis:
Let me ping @Dav1dde and @aldy505 since you reworked the health checks in [relay](https://github.com/getsentry/relay/commits/master/relay/src/healthcheck.rs) and [symbolicator](https://github.com/getsentry/symbolicator/commits/master/crates/symbolicator/src/healthcheck.rs) - if you see issues around those two services, might be useful to ask if the affected setup has a proxy.

Dav1dde:
I wonder if we were better off to just rewrite `0.0.0.0` explicitly to `127.0.0.1` instead of relying on this somewhat platform specific behaviour. This would also eliminate the need to add it to the no proxy var. I assume adding it is fine, but I am definitely not certain if there aren't any unintended side-effects.

DragoonAethis:
> I wonder if we were better off to just rewrite `0.0.0.0` explicitly to `127.0.0.1` instead of relying on this somewhat platform specific behaviour.

That could work too, AFAIK health checks are executed like `docker exec -t service health-check-cmd`, so 127.0.0.1 should always point at the live process.

Dav1dde (approved):
Tried to research this a bit, seems like there isn't an obvious reason why this may have unintended side-effects.

PR #16261, opened 2026-02-04
feat(platform): Move docs to next 16

chargome:
- Update next and eslint-config-next to ^16.1.6
- Migrate middleware to proxy (renamed src/middleware.ts to src/proxy.ts per Next.js 16 convention)
- Migrate ESLint to flat config format (eslint.config.mjs)
- Add --webpack flag to dev/build scripts (We'll cover Turbopack at a later point)
- Move src/pages/ to root pages/ (Next.js 16 requires app and pages at same level)
- Add images.localPatterns config for query string support on local images
- Fix component type errors for React 19 / Next.js 16 compatibility

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,6 +8,7 @@ overrides:
   dompurify: 3.2.4
   '@types/dompurify': 3.0.5
   '@types/node': ^22
+  es-iterator-helpers: 1.2.1
 
 importers:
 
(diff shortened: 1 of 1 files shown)


chargome:
Leaving this here for posterity:
```
The [[...path]] route statically generates ~9,400 pages via generateStaticParams. In the Vercel build output, this produces 37,780 files (HTML fallbacks, RSC payloads, prerender configs) totaling 3.8GB. The deployment payload to Vercel's API exceeds its 75MB
   limit.

  This likely worked on Next.js 15 because the Vercel adapter produced a different (smaller) output format. Next.js 16's output structure creates more per-route artifacts.

  Size Breakdown

  - 9,439 HTML fallbacks: 2.4GB
  - 9,439 RSC fallbacks: 1.3GB
  - 18,896 JSON configs: 20MB
  - 4 .func bundles: ~3MB

  Solutions

  The main options are:

  Option 1: Switch to ISR (Incremental Static Regeneration)

  Change app/[[...path]]/page.tsx from full SSG to ISR - prerender a subset at build time, generate the rest on-demand:

  // Instead of:
  export const dynamicParams = false;
  export const dynamic = 'force-static';

  // Use:
  export const dynamicParams = true;
  export const revalidate = 300; // regenerate every 5 min

  And optionally limit generateStaticParams to only the most important pages (home, top-level sections).

  Option 2: Only prerender top-level pages

  Keep force-static but reduce what gets prerendered:

  export async function generateStaticParams() {
    const docs = await getDocsFrontMatter();
    // Only prerender top-level paths (depth ≤ 2), rest on-demand
    const paths = docs
      .filter(doc => doc.slug.split('/').length <= 2)
      .map(doc => ({ path: doc.slug.split('/') }));
    paths.push({ path: undefined });
    return paths;
  }
  export const dynamicParams = true; // allow unlisted paths



 ```

chargome:
closing this for now

PR #16283, opened 2026-02-06
docs(ai): add cost calculation docs

shellmayr:
- Add a new page in the agent monitoring documentation that outlines where we get the cost data from and how the cost is calculated
- Add a new include that is used in manual instrumentation docs for Python, Node and Ruby that points out some gotchas with cost and token accounting
- Point out which costs we don't account for, especially non-token-based costs

Closes TET-1868


docs/product/insights/ai/agents/costs.mdx
```diff
@@ -0,0 +1,71 @@
...
+            + reasoning_tokens × reasoning_rate
+
+total cost  = input cost + output cost
+```
+
+If a model doesn't have a specific reasoning token rate, the standard output rate is used for reasoning tokens as well.
+
+## Required Span Attributes
+
+Cost calculation relies on attributes set on AI spans:
```
alexander-alderman-webb, line 43: which AI span?

shellmayr: We talked about this offline - we currently apply cost calculation to all AI spans and possibly this should change to prevent double-accounting if attached to both request spans and invoke_agent spans. This will probably change in the future, but for now having slightly vague docs is an improvement over no docs.

shellmayr: I've just learned from @vgrozdanic that we actually do de-duplicate costs in the product, so we don't display too high costs if there are cost attributes on both the invoke_agent and request spans.

PR #16299, opened 2026-02-09
chore(devdocs): Clarify manually set data on scope and sendDefaultPii behavior

romtsn:
Clarified that user-defined User objects on the scope are not stripped.


romtsn:
@cleptric put an alert above stating that all manual data should be sent regardless of the flag, does it look better?

develop-docs/sdk/expected-features/data-handling.mdx
```diff
@@ -13,6 +13,12 @@ In the event that API returns data considered PII, we guard that behind a flag c
 This is an option in the SDK called [_send-default-pii_](https://docs.sentry.io/platforms/python/configuration/options/#send-default-pii)
 and is **disabled by default**. That means that data that is naturally sensitive is not sent by default.
 
+<Alert level="info">
+
+When a user manually sets the data on the scope (user, contexts, tags, data, request, response, etc.), this data should not be gated by the _Send Default PII_ flag and should always be attached to all outgoing telemetry.
```
cleptric: Can we clarify that this also applies to data a user manually sets on a span, log, metric etc.

romtsn:
```suggestion
When a user manually sets the data on the scope (user, contexts, tags, data, request, response, etc.), this data should not be gated by the _Send Default PII_ flag and should always be attached to all outgoing telemetry. This also applies to the data that the user manually sets on a span, log, metric and other types of telemetry (directly or, for example, via `BeforeSend`).
```

romtsn:
```suggestion
When a user manually sets the data on the scope (user, contexts, tags, data, request, response, etc.), this data should not be gated by the _Send Default PII_ flag and should always be attached to all outgoing telemetry. This also applies to the data that the user manually sets on a span, log, metric and other types of telemetry (directly or, for example, via `BeforeSend`).
```

romtsn: fbd363f74d59624bd69102e4d5b21860970c9602

PR #16327, opened 2026-02-10
feat(develop/spans): Add Span Buffer specification

Lms24:
As discussed yesterday, this PR adds a spec for a span buffer. This is less specified than the Telemetry Processor but more formal than the recommendation in the Span-First implementation guide (which I removed in favor of this page).

Most important change: We added weight-based flushing as an addition to the previous flushing behaviour

closes https://linear.app/getsentry/issue/FE-718/span-buffer-section-in-dev-docs

### DIFF
--- a/develop-docs/sdk/telemetry/spans/implementation.mdx
+++ b/develop-docs/sdk/telemetry/spans/implementation.mdx
@@ -4,13 +4,16 @@ sidebar_order: 10
 ---
 
 <Alert level="warning">
-  🚧 This document is work in progress.
-  The steps and suggestions in this document primarily serve as a means to document what SDKs so far have been doing when implementing Span-First.
-  This page also serves as a place to document (temporary) decisions, trade-offs, considerations, etc.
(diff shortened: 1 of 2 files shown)


cleptric:
The work on the telemetry processor is on hold. So having all things spans under one menu item is ok imo.

coolguyzone (approved):
LGTM!

Lms24:
> overall looks reasonable, however what is the long-term plan for the span buffer page? I'd imagine that it will easily gets out of sync with the telemetry processor docs and since both kinda have the same info it could be confusing which one is the source of truth for spans

My take for now: If SDKs already implement(ed) the telemetry processor (not sure if any), the span buffer specification is more or less in there, so I'd recommend they stick with that. For SDKs having not done this yet (which I assume is the majority), the new span buffer page should be all they care about at the moment for span streaming. Does this sound reasonable?

cleptric (approved):

PR #16419, opened 2026-02-17
chore: migrate from yarn to pnpm

BYK:
## Summary

Migrate the project's package manager from **Yarn Classic (1.22.22)** to **pnpm (10.30.0)** with strict hoisting.

### What changed

- **Package manager files**: Replaced `yarn.lock` + `.yarnrc.yml` with `pnpm-lock.yaml` + `.npmrc`
- **Version management**: Volta (`volta.pnpm`) + `packageManager` field (both)
- **Strict hoisting**: Uses pnpm's default strict `node_modules` layout — no `shamefully-hoist`
- **14 phantom dependencies declared**: Packages that were silently available via yarn's flat hoisting are now explicitly declared in `package.json`:
  - Runtime: `@radix-ui/react-popover`, `@radix-ui/react-select`, `@sentry/browser`, `classnames`, `hast-util-heading-rank`, `hast-util-to-string`, `mdast-util-from-markdown`, `mdast-util-to-markdown`, `mdast-util-to-string`, `refractor`, `remark-rehype`, `unist-util-visit`
  - Types: `@types/hast`, `@types/mdast`, `@types/unist`
  - Dev: `postcss-import`
- **pnpm config**: `resolutions` → `pnpm.overrides`, added `pnpm.onlyBuiltDependencies` for native build scripts
- **Type workaround**: Added `src/@types/remark-mdx-images.d.ts` + `tsconfig.json` paths override for `remark-mdx-images` (ships `.ts` source files that conflict under strict isolation)
- **CI**: All 4 GitHub Actions workflows updated (`pnpm/action-setup@v4`, `cache: 'pnpm'`, command changes)
- **Docs & tooling**: `Makefile`, `CONTRIBUTING.md`, `AGENTS.md`, `develop-docs/`, `.gitignore`, `_typos.toml`, `.claude/settings.json`

### Verification

- `pnpm install` — clean install, no warnings
- `pnpm test:ci` — 9 test suites, 100 tests pass
- `pnpm lint:ts` — clean (no type errors)

### Not changed

- `docs/` and `platform-includes/` MDX files — these contain `yarn add @sentry/*` as **end-user SDK install instructions**, not project tooling
- `public/_platforms/` JSON files — generated build artifacts

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -34,6 +34,9 @@ importers:
       '@mdx-js/loader':
         specifier: ^3.0.0
         version: 3.1.1(webpack@5.105.2(esbuild@0.25.12))
+      '@mdx-js/mdx':
+        specifier: ^3.0.0
+        version: 3.1.1
       '@mdx-js/react':
         specifier: ^3.0.0
         version: 3.1.1(@types/react@18.3.12)(react@19.2.4)


.github/workflows/algolia-index.yml
```diff
@@ -27,20 +32,13 @@ jobs:
...
 
-      - uses: actions/cache@v4
-        id: cache
-        with:
-          path: ${{ github.workspace }}/node_modules
-          key: node-${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
-
-      - run: yarn install --frozen-lockfile
-        if: steps.cache.outputs.cache-hit != 'true'
+      - run: pnpm install --frozen-lockfile
```
BYK, line 35: Why are we not skipping on cache hit?

.github/workflows/sync-labels.yml
```diff
@@ -21,27 +21,24 @@ jobs:
...
         with:
           node-version-file: 'package.json'
 
-      - name: Install github-label-sync
-        run: yarn global add github-label-sync@2.2.0
-
       - name: Run github-label-sync
         run: |
-          github-label-sync \
+          npx github-label-sync@2.2.0 \
```
BYK: Why not `pnpmx`?

develop-docs/frontend/upgrade-policies.mdx
```diff
@@ -52,13 +52,13 @@ You may not have to do this manually! See the below section on [Take advantage o
 ### Seeing what can be upgraded
 
 ```bash
-yarn outdated --color | sort
+pnpm outdated --color | sort
```
BYK, line 55: Can you ensure all these `pnpm` commands are legit?

.gitignore
```diff
@@ -1,8 +1,7 @@
 # Logs
 *.log
 npm-debug.log*
-yarn-debug.log*
-yarn-error.log*
```
BYK, line 5: If we keep `npm-debug.log*` why remove `yarn` ones? No harm there? Or remove the npm one too

.npmrc
```diff
@@ -0,0 +1 @@
+enable-pre-post-scripts=true
```
BYK: Do we really need this? It is quite dangerous

_typos.toml
```diff
@@ -23,7 +23,7 @@ extend-exclude = [
   # Data and configuration files
   "*.json",            # JSON files have many false positives
   "*.lock",            # Package lock files
-  "yarn.lock",         # Yarn lock file
+  "*.yaml",            # YAML files (includes pnpm-lock.yaml)
   "package-lock.json", # NPM lock file
```
BYK, line 27: Is this relevant at all?

.github/workflows/algolia-index.yml
BYK: The original yarn version of this workflow (algolia-index.yml) cached `node_modules/` directly and skipped `yarn install` on cache hit. With pnpm, `actions/setup-node` with `cache: 'pnpm'` caches the **pnpm store** (compressed tarballs), not `node_modules/`. So `pnpm install --frozen-lockfile` must always run to create the symlink structure in `node_modules/`, even with a warm store cache. The install is near-instant on cache hit since all packages are already in the store — it just links them.

This is the standard pnpm CI pattern recommended by pnpm docs and `pnpm/action-setup`.

.github/workflows/sync-labels.yml
BYK: Good call — `pnpx` is actually deprecated since pnpm v7 (replaced by `pnpm dlx`). Changed to `pnpm dlx github-label-sync@2.2.0` for consistency, and added `pnpm/action-setup@v4` so pnpm is available in the workflow.

develop-docs/frontend/upgrade-policies.mdx
BYK: Verified all three:

- `pnpm outdated` — lists packages with newer versions available ([docs](https://pnpm.io/cli/outdated)). Equivalent to `yarn outdated`.
- `pnpm update --latest` — updates packages to their latest versions, ignoring the ranges in `package.json` ([docs](https://pnpm.io/cli/update#--latest--l)). Equivalent to `yarn upgrade --latest`.
- `pnpm dedupe` (line 164) — deduplicates dependencies in the lockfile ([docs](https://pnpm.io/cli/dedupe)). Equivalent to `yarn dedupe` / `npx yarn-deduplicate`.

Also, `pnpm install` and `pnpm dev-ui` (lines 36, 153) are standard pnpm commands. All confirmed legit.

.gitignore
BYK: Good point — `*.log` on line 2 already covers `npm-debug.log*`, `yarn-debug.log*`, `yarn-error.log*`, and `pnpm-debug.log*`. Removed all the specific debug log entries since they are redundant. Also removed the separate `.pnpm-debug.log*` entry that was lower in the file.

.npmrc
BYK: You're right — removed it. `enable-pre-post-scripts=true` re-enables pnpm's execution of `pre*`/`post*` lifecycle scripts (disabled by default in pnpm v10 for security). But this project has **zero** pre/post lifecycle scripts in `package.json`, so it's not needed. Deleted `.npmrc` entirely.

_typos.toml
BYK: Good catch — `*.yaml` was too broad and would skip spell-checking all YAML files including CI workflows. The `*.lock` pattern on line 25 does not cover `pnpm-lock.yaml` (it only matches `*.lock` extensions). Changed to specifically `"pnpm-lock.yaml"` so only the lockfile is excluded.

PR #16492, opened 2026-02-20
feat(dev/logs/metrics): Add `sentry.timestamp.sequence` definition to the logs and metrics specs

logaretm:
This pull request updates the telemetry logs specification to introduce deterministic log ordering via a new attribute, `sentry.timestamp.sequence`. This change addresses issues in environments where multiple logs can share identical timestamps, ensuring logs can always be ordered correctly.

The attribute addresses an issue that came up in Cloudflare’s runtime which [freezes all timers and clock functions](https://developers.cloudflare.com/workers/reference/security-model/#step-1-disallow-timers-and-multi-threading), causing all timestamp calls to return the same value. This can lead to logs being displayed out of order if they don’t get sent in one request. Instead, they’ll appear in the order of arrival in ingest, which is subject to request latencies and out-of-order request completion.

This is a proposal as we have yet to see what product thinks about this.

### DIFF
--- a/develop-docs/sdk/telemetry/logs.mdx
+++ b/develop-docs/sdk/telemetry/logs.mdx
@@ -11,7 +11,7 @@ spec_depends_on:
     version: ">=1.0.0"
 spec_changelog:
   - version: 1.16.0
-    date: 2026-02-06
+    date: 2026-03-04
     summary: Add sentry.timestamp.sequence attribute for deterministic log ordering
   - version: 1.15.0
     date: 2026-02-03
--- a/develop-docs/sdk/telemetry/metrics.mdx
+++ b/develop-docs/sdk/telemetry/metrics.mdx
@@ -11,7 +11,7 @@ spec_depends_on:
     version: ">=1.0.0"
 spec_changelog:
   - version: 2.6.0
-    date: 2026-02-06
+    date: 2026-03-04
     summary: Add sentry.timestamp.sequence attribute for deterministic metric ordering
   - version: 2.5.0
     date: 2026-02-12


develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -291,6 +295,25 @@ Logs **SHOULD** be associated with replays. If a log is recorded during an activ
...
+
+Some runtimes (notably Cloudflare Workers) freeze timer APIs (`Date.now()`, `performance.now()`) during synchronous execution, causing multiple logs to share identical timestamps. To guarantee deterministic ordering, SDKs **MUST** attach a `sentry.log.sequence` integer attribute to every log.
+
+The sequence number **MUST**:
+- Start at `0` when the SDK initializes.
+- Increment by `1` for each log that is captured.
+- Be scoped to the SDK instance lifetime. It resets naturally on process restart.
+
+The sequence number **MUST NOT**:
+- Be intentionally reset during the process lifetime.
```
Dav1dde: You only need it as a discriminator per timestamp, so maybe that's enough as a constraint?

develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -554,7 +577,8 @@ A complete `log` envelope with six log entries at different severity levels:
         },
         "sentry.message.parameter.0": { "value": 120, "type": "integer" },
         "sentry.message.parameter.1": { "value": 85, "type": "integer" },
-        "sentry.message.parameter.2": { "value": 60, "type": "integer" }
+        "sentry.message.parameter.2": { "value": 60, "type": "integer" },
+        "sentry.log.sequence": { "value": 0, "type": "integer" }
```
Dav1dde: I wonder if this should be top-level, it feels heavily tied to the timestamp and a fundamental attribute of a log.

@k-fish wdyt?

develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -291,6 +295,25 @@ Logs **SHOULD** be associated with replays. If a log is recorded during an activ
...
+The sequence number **MUST**:
+- Start at `0` when the SDK initializes.
+- Increment by `1` for each log that is captured.
+- Be scoped to the SDK instance lifetime. It resets naturally on process restart.
+
+The sequence number **MUST NOT**:
+- Be intentionally reset during the process lifetime.
+- Be used to modify or synthesize sub-millisecond timestamp precision.
+
+The sequence provides deterministic ordering within a single SDK instance. It does not guarantee ordering across independent processes or workers, which have separate counters.
```
Dav1dde: But it does between separate threads?

develop-docs/sdk/telemetry/logs.mdx
logaretm: This might be a rare scenario, but If it does get reset during the lifetime of the SDK process then on environments like Cloudflare it will produce the same timestamps again since the time is frozen throughout their lifetime, so it won't be unique anymore. So, I wanted to close that door.

What do you think?

develop-docs/sdk/telemetry/logs.mdx
logaretm: I namespaced it initially to signal to the user that they should not mess with this (they shouldn't be able anyways).

But yea I agree, it is tied to the logs timestamps heavily and has little to no-use outside of that.

develop-docs/sdk/telemetry/logs.mdx
logaretm: Within the same process, yes the counter should be shared across threads. The limitation is only across separate processes or worker isolates where there's no shared memory.

Should I clarify that in the doc?

Dav1dde: I think it's good, I just wanted to make sure that was an explicit decision.

develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -291,6 +295,25 @@ Logs **SHOULD** be associated with replays. If a log is recorded during an activ
...
+Some runtimes (notably Cloudflare Workers) freeze timer APIs (`Date.now()`, `performance.now()`) during synchronous execution, causing multiple logs to share identical timestamps. To guarantee deterministic ordering, SDKs **MUST** attach a `sentry.log.sequence` integer attribute to every log.
+
+The sequence number **MUST**:
+- Start at `0` when the SDK initializes.
+- Increment by `1` for each log that is captured.
+- Be scoped to the SDK instance lifetime. It resets naturally on process restart.
+
+The sequence number **MUST NOT**:
+- Be intentionally reset during the process lifetime.
+- Be used to modify or synthesize sub-millisecond timestamp precision.
```
k-fish: Why? Adding a separate attribute means to do log ordering naturally we'll now have to either:

A. Include yet another column on every request in order to have logs ordered properly when a users sorts by time (log.sequence), we already are using timestamp and precise timestamp when doing log ordering. Adding a column adds additional query bytes to every request.
B. NOT sort properly by this monotonic counter, and have it as a free floating attribute that a user will have to know about and then sort by independently, with a count from beginning of the lifetime being hard to intuit. 

Is it possible we send it and have relay transform the precise timestamp into a synthesized version where we add an annotation (meta) to indicate the log sequence value it was modified by and why (because their sdk doesn't support precise timestamps). That way we can put a warning in the timestamp tooltip that explains it, and encode the sequence into the precise timestamp where we have an extra 32 bits being essentially unused at millisecond precision? Thoughts?

develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -291,6 +295,25 @@ Logs **SHOULD** be associated with replays. If a log is recorded during an activ
 
 </SpecSection>
 
+### Log Ordering
+
+<SpecSection id="log-ordering" status="candidate" since="1.16.0">
+
+Some runtimes (notably Cloudflare Workers) freeze timer APIs (`Date.now()`, `performance.now()`) during synchronous execution, causing multiple logs to share identical timestamps. To guarantee deterministic ordering, SDKs **MUST** attach a `sentry.log.sequence` integer attribute to every log.
```
k-fish: re: `sentry.log.sequence` it actually is probably necessary for our other telemetry too, at least for metrics I know it could be a problem. 

imo something like `sentry.timestamp.sequence` makes more sense to me, my $0.02

develop-docs/sdk/telemetry/logs.mdx
logaretm: > Is it possible we send it and have relay transform the precise timestamp into a synthesized version where we add an annotation (meta) to indicate the log sequence value it was modified by...

Nar suggested something similar about doing the synthesizing on the SDK level. I included that part to avoid lying to the user about the actual time. I think your suggestion is more transparent so we could do that, but it feels like it involves more changes downstream to the UI.

I have a few thoughts on this:

1. We can't dynamically tell if the clock is frozen or not so we can't synthesize conditionally unless we send in another flag to tell relay to do that, but then again users cross-deploy a lot and they can have a node SDK running on Cloudflare, so the synthesizing would need to happen all the time. I initially wanted us to use this for the off chance some SDK got its clock frozen for whatever reason.
2. We could skew the ms value but with 32 bit of space, it may not grow big enough to overflow into seconds range.


I would've liked for this to be completely transparent to the user, How considerable is the cost of adding an secondary sort integer to our queries?

develop-docs/sdk/telemetry/logs.mdx
logaretm: Actually the clockfreeze thing on cloudflare does affect all of our telemetry.

- Traces and spans always display 0ms since `startTime` and `endTime` are always the same, the counter wouldn't help with that tho.
- Metrics are probably affected, but I haven't been exposed to them yet. I can explore further there if we want to make this a general telemetry attribute.

logaretm: Talked with Kevan offline and we will do the same thing for metrics, so the attribute name will change to `sentry.timestamp.sequence`

develop-docs/sdk/telemetry/logs.mdx
logaretm: Talked with Kevan offline and we figured a good compromise is to do a reset of the counter whenever we detect that the clock did advance, this would keep the sequence numbers small for most cases which means it won't grow larger enough to skew the numbers.

We also will monitor how large the sequence numbers grows to see if we should modify our approach.

k-fish (approved):
Add one final piece about sequence _not_ being sent if not required (the sdk has better precision than 1ms or non frozen time) then 👍 from me

logaretm:
@Dav1dde and @cleptric, After going through it with Kevan, this is what we have right now. Any thoughts?

Dav1dde:
👍 from me

develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -290,6 +294,25 @@ Logs **SHOULD** be associated with replays. If a log is recorded during an activ
 
 </SpecSection>
 
+### Log Ordering
+
+<SpecSection id="log-ordering" status="candidate" since="1.16.0">
+
+Some runtimes (notably Cloudflare Workers) freeze timer APIs (`Date.now()`, `performance.now()`) during synchronous execution, causing multiple logs to share identical timestamps. SDKs that target runtimes where timestamps may be frozen or lack sub-millisecond precision **MUST** attach a `sentry.timestamp.sequence` integer attribute to every log. SDKs that only target runtimes with reliable sub-millisecond timestamps **MAY** omit it.
+
+When sent, the sequence number **MUST**:
```
coolguyzone:
```suggestion
When sent, the sequence integer **MUST**:
```

I would do this, or change them both to 'number', just so we are consistent.

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -187,6 +191,25 @@ Whenever possible, metrics **SHOULD** be linked to replays. If a metric is recor
 
 </SpecSection>
 
+### Metric Ordering
+
+<SpecSection id="metric-ordering" status="candidate" since="2.6.0">
+
+Some runtimes (notably Cloudflare Workers) freeze timer APIs (`Date.now()`, `performance.now()`) during synchronous execution, causing multiple metrics to share identical timestamps. SDKs that target runtimes where timestamps may be frozen or lack sub-millisecond precision **MUST** attach a `sentry.timestamp.sequence` integer attribute to every metric. SDKs that only target runtimes with reliable sub-millisecond timestamps **MAY** omit it.
+
+When sent, the sequence number **MUST**:
```
coolguyzone:
```suggestion
When sent, the sequence integer **MUST**:
```

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -114,6 +117,7 @@ SDKs **MUST** attach the following attributes to every metric by default:
 2. `sentry.release` — the release set in the SDK, if defined.
 3. `sentry.sdk.name` — the name of the SDK that sent the metric.
 4. `sentry.sdk.version` — the version of the SDK that sent the metric.
+5. `sentry.timestamp.sequence` — a monotonically incrementing integer counter used to determine correct ordering of metrics when timestamps are identical. **MUST** be sent on runtimes with frozen or low-precision timers, **MAY** be omitted otherwise. See [Metric Ordering](#metric-ordering).
```
cleptric: Let's please move this out of the default section. We do not need to add this to all SDKs.

develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -153,6 +156,7 @@ SDKs **MUST** attach the following attributes to every log:
 | `sentry.sdk.name` | 1.0.0 | The name of the SDK that sent the log. |
 | `sentry.sdk.version` | 1.0.0 | The version of the SDK that sent the log. |
 | `sentry.replay_id` | 1.8.0 | The replay ID of the active replay when the log was collected. **MUST NOT** be set if no replay is active. |
+| `sentry.timestamp.sequence` | 1.16.0 | A monotonically incrementing integer counter used to determine correct ordering of logs when timestamps are identical. **MUST** be sent on runtimes with frozen or low-precision timers, **MAY** be omitted otherwise. See [Log Ordering](#log-ordering). |
```
cleptric: Let's please move this out of the default section. We do not need to add this to all SDKs.

develop-docs/sdk/telemetry/metrics.mdx
logaretm: Done, thanks!

develop-docs/sdk/telemetry/logs.mdx
logaretm: Done, thanks!

develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -2,14 +2,17 @@
...
 spec_status: stable
 spec_depends_on:
   - id: sdk/foundations/transport/envelopes
     version: ">=1.0.0"
   - id: sdk/foundations/state-management/scopes/attributes
     version: ">=1.0.0"
 spec_changelog:
+  - version: 1.16.0
+    date: 2026-02-06
+    summary: Added sentry.timestamp.sequence default attribute for deterministic log ordering
```
cleptric:
```suggestion
    summary: Add sentry.timestamp.sequence attribute for deterministic log ordering
```

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -2,14 +2,17 @@
...
 spec_status: stable
 spec_depends_on:
   - id: sdk/foundations/transport/envelopes
     version: ">=1.0.0"
   - id: sdk/foundations/state-management/scopes/attributes
     version: ">=1.0.0"
 spec_changelog:
+  - version: 2.6.0
+    date: 2026-02-06
+    summary: Added sentry.timestamp.sequence default attribute for deterministic metric ordering
```
cleptric:
```suggestion
    summary: Add sentry.timestamp.sequence attribute for deterministic metric ordering
```

PR #16524, opened 2026-02-23
docs(span-first): Add Noop Span specification to Span API

buenaflor:
Adds noop span spec to span-first

### DIFF
--- a/develop-docs/sdk/telemetry/spans/span-api.mdx
+++ b/develop-docs/sdk/telemetry/spans/span-api.mdx
@@ -144,7 +144,7 @@ This is still TBD but the `parentSpan` option should suffice to serve this use c
 
 A Noop span is a span that implements the full [Span Interface](#span-interface) but does not record any data and has no influence on the trace hierarchy.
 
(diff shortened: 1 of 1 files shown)


develop-docs/sdk/telemetry/spans/span-api.mdx
```diff
@@ -140,6 +140,37 @@ Likewise, additional APIs MAY also adjust the span status based on errors thrown
 At this time, SDKs MUST NOT expose APIs like `Span::startChild` or similar functionality that explicitly creates a child span.
 This is still TBD but the `parentSpan` option should suffice to serve this use case.
 
+## Noop Span
+
+A Noop span is a span that implements the full [Span Interface](#span-interface) but does not record any data and has no influence on the trace hierarchy.
+
+`startSpan` **MUST** always return a valid `Span` instance. When a span should not be recorded (e.g., due to [sampling](../sampling/) or [filtering](../filtering/) decisions), the SDK **MUST** return a Noop span. This allows callers to use the span without null checks.
```
coolguyzone:
```suggestion
`startSpan` **MUST** always return a valid `Span` instance. If a span should not be recorded (for example due to [sampling](../sampling/) or [filtering](../filtering/) decisions), the SDK **MUST** return a Noop span instead. This allows callers to use the span without null checks.
```

PR #16529, opened 2026-02-23
docs(build-distribution): Add Install Groups documentation

runningcode:
## DESCRIBE YOUR PR

Add documentation for the new Install Groups feature in Build Distribution. Install groups let users tag builds with group names to control which builds are offered as over-the-air updates.

- Add product-level docs explaining install groups concept, OR-matching logic, and use cases
- Add Android platform guide covering CLI, Gradle plugin, and Fastlane plugin setup
- Add iOS platform guide covering CLI and Fastlane plugin setup
- Update upload metadata table with `install-group` field

## IS YOUR CHANGE URGENT?

- [ ] Other deadline:

## SLA

## PRE-MERGE CHECKLIST

EME-864

### DIFF
--- a/docs/product/build-distribution/index.mdx
+++ b/docs/product/build-distribution/index.mdx
@@ -33,12 +33,9 @@ This is useful when you have multiple distribution channels (for example, separa
 
 ### How It Works
 
-When you upload a build, you can assign one or more install groups to it. When the Auto-Update SDK checks for updates:
+When you upload a build, you can assign one or more install groups to it. If the Auto-Update SDK is provided install groups, it only returns builds that have overlapping groups.
 
-1. If install groups are provided when checking for updates, only builds with at least one matching group are considered.
-2. If no install groups are provided, all builds for that app and build configuration are eligible.
-
-Matching uses **OR logic** — a build only needs to share _one_ group to be considered a valid update. For example, a build tagged `["alpha", "staging"]` will see updates from a build tagged `["alpha", "beta"]` because both share `alpha`.
+Matching requires two builds to have a non-empty intersection of their install groups. For example, a build tagged `["alpha", "staging"]` will see updates from a build tagged `["alpha", "beta"]` because both share `alpha`.
 
 ### Use Cases
 


docs/product/build-distribution/index.mdx
```diff
@@ -25,6 +25,32 @@ Below is the metadata included in your build, regardless of the platform.
 
 <Include name="size-analysis/upload-metadata" />
 
+## Install Groups
+
+Install groups let you tag builds with one or more group names to control update visibility. When a device checks for updates, only builds that share at least one install group with the currently installed build will be returned as available updates.
```
noahsmartin: "When a device checks for updates you can provide a list of install groups, only builds that share at least one install group with this list will be returned as available updates."

docs/product/build-distribution/index.mdx
```diff
@@ -25,6 +25,32 @@ Below is the metadata included in your build, regardless of the platform.
...
+
+This is useful when you have multiple distribution channels (for example, separate branches, teams, or rollout stages) and want to prevent builds from one channel from being offered as updates to another.
+
+### How It Works
+
+When you upload a build, you can assign one or more install groups to it. When the Auto-Update SDK checks for updates:
+
+1. If the currently installed build has install groups, only builds with at least one matching group are considered.
```
noahsmartin: Same suggestion here, AFAIK on android there is no "currently installed build" sense right? Since there is no unique identifier. Even on iOS the unique identifier is best effort. It only changes when the binary changes so if two builds only have a difference in a non-binary file (like an image) we won't be able to tell which is which. So would suggest to just talk about providing the list of install groups and not mixing in the best-effort automatic "current" detection

docs/product/build-distribution/index.mdx
runningcode: ah that makes more sense

docs/product/build-distribution/index.mdx
runningcode: Good point! adjusted

docs/product/build-distribution/index.mdx
```diff
@@ -25,6 +25,32 @@ Below is the metadata included in your build, regardless of the platform.
...
+Install groups let you tag builds with one or more group names to control update visibility. When a device checks for updates, you can provide a list of install groups — only builds that share at least one install group with this list will be returned as available updates.
+
+This is useful when you have multiple distribution channels (for example, separate branches, teams, or rollout stages) and want to prevent builds from one channel from being offered as updates to another.
+
+### How It Works
+
+When you upload a build, you can assign one or more install groups to it. When the Auto-Update SDK checks for updates:
+
+1. If install groups are provided when checking for updates, only builds with at least one matching group are considered.
+2. If no install groups are provided, all builds for that app and build configuration are eligible.
```
noahsmartin: It's a bit ambiguous what "that app" is. Where does "that app's build configuration" come from? My point is I don't think this is a thing we actually do. So I would simplify this whole section:

```
When you upload a build, you can assign one or more install groups to it. If the auto-update SDK is provided install_groups it only returns builds that have overlapping groups.
```

docs/product/build-distribution/index.mdx
```diff
@@ -25,6 +25,32 @@ Below is the metadata included in your build, regardless of the platform.
...
+
+Matching uses **OR logic** — a build only needs to share _one_ group to be considered a valid update. For example, a build tagged `["alpha", "staging"]` will see updates from a build tagged `["alpha", "beta"]` because both share `alpha`.
```
noahsmartin: `OR` is probably not the right word? It's really a set intersection. I'd say "Matching requires two builds to have a non-empty intersection of their install_groups. For example... "

PR #16544, opened 2026-02-24
docs(nestjs): Isolate async background job scopes

nicohrubec:
Async background jobs leak breadcrumbs into HTTP request scopes in the `nestjs` SDK. This can be fixed by wrapping async jobs with `withIsolationScope`. We will investigate a more holistic fix but for now we should document the workaround.

[Context](https://github.com/getsentry/sentry-javascript/issues/17742#issuecomment-3950424831)

### DIFF
--- a/docs/platforms/javascript/guides/nestjs/index.mdx
+++ b/docs/platforms/javascript/guides/nestjs/index.mdx
@@ -78,7 +78,11 @@ export class AppModule {}
 
 <PlatformContent includePath="getting-started-capture-errors" />
 
-## Step 4: Add Readable Stack Traces With Source Maps (Optional)
+## Step 4: Isolate Async Context for Background Jobs (Optional)
+
+If your NestJS app uses background jobs (such as `@Cron()`, `@Interval()`, `@OnEvent()`, or BullMQ), breadcrumbs from those jobs can leak into unrelated HTTP request error events. Wrap your background job handlers with `Sentry.withIsolationScope()` to isolate them. See <PlatformLink to="/features/async-context">Isolate Background Job Scopes</PlatformLink> for details and code examples.
+
+## Step 5: Add Readable Stack Traces With Source Maps (Optional)
 
 The stack traces in your Sentry errors probably won't look like your actual code. To fix this, upload your source maps to Sentry.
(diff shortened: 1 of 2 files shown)


docs/platforms/javascript/guides/nestjs/index.mdx
```diff
@@ -78,7 +78,11 @@ export class AppModule {}
 
 <PlatformContent includePath="getting-started-capture-errors" />
 
-## Step 4: Add Readable Stack Traces With Source Maps (Optional)
+## Step 4: Isolate Async Context for Background Jobs (Optional)
+
+If your NestJS app uses background jobs (such as `@Cron()`, `@Interval()`, `@OnEvent()`, or BullMQ), breadcrumbs from those jobs can leak into unrelated HTTP request error events. Wrap your background job handlers with `Sentry.withIsolationScope()` to isolate them. See <PlatformLink to="/features/async-context">Isolate Background Job Scopes</PlatformLink> for details and code examples.
```
inventarSarah, line 83:
```suggestion
If your NestJS app uses background jobs (such as `@Cron()`, `@Interval()`, `@OnEvent()`, or BullMQ), breadcrumbs from those jobs can leak into unrelated HTTP request error events, making it harder to debug issues. 
To prevent this, wrap your background job handlers with `Sentry.withIsolationScope()` to isolate them:
```

inventarSarah: Suggestion: I would also show the code example on this page (maybe create an include to reuse the snippet). 
Since we don't have much additional information there, I wouldn't want to send users away from the guide; instead, I'd just show them here. 
However, if you plan to add more stuff to the feature page later, it could make sense to keep the link here as is.

inventarSarah (approved):
Looks good to me! 🎈

Just added a suggestion to the quick start guide :)

docs/platforms/javascript/guides/nestjs/index.mdx
nicohrubec: There is multiple cases where this is happening so I think maybe we can expand that page in the future in case more people run into this

PR #16578, opened 2026-02-25
docs: Enhance CI status check guidelines for PR reviews

szokeasaurusrex:
Expanded the CI status check to include conditions for prior reviews and comments.

### DIFF
--- a/develop-docs/sdk/getting-started/playbooks/development/reviewing-a-pr.mdx
+++ b/develop-docs/sdk/getting-started/playbooks/development/reviewing-a-pr.mdx
@@ -38,9 +38,15 @@ Related resources:
 
 Does it explain what and why? Is there a linked issue? If either is missing, request it before reviewing code.
 
-#### 2. Check CI status
+#### 2. Check CI status and prior AI/human reviews
 
-If checks are failing, comment on what's broken and wait. You **MUST NOT** review code that doesn't build or pass tests.
+A reviewer **SHOULD NOT** begin a full review unless all of the following conditions are met:
+
+- All CI checks, builds, and tests have completed successfully.
+- All requested AI code review agents have completed their reviews.
+- All comments from previous review cycles (AI or human) have been addressed, resolved, or explicitly marked as not applicable.
+
+If any of the above conditions are unmet, the reviewer **MAY** leave comments to help the author unblock CI failures or address outstanding feedback.
 
 #### 3. Review the code
 


develop-docs/sdk/getting-started/playbooks/development/reviewing-a-pr.mdx
```diff
@@ -38,9 +38,15 @@ Related resources:
...
+#### 2. Check CI status and prior AI/human reviews
 
-If checks are failing, comment on what's broken and wait. You **MUST NOT** review code that doesn't build or pass tests.
+A reviewer **SHOULD NOT** begin a full review unless all of the following conditions are met:
+
+- All CI checks, builds, and tests have completed successfully.
+- All requested AI code review agents have completed their reviews.
+- All comments from previous review cycles (AI or human) have been addressed, resolved, or explicitly marked as not applicable.
+
+If any of the above conditions are unmet, the reviewer **MAY** leave comments to help the author unblock CI failures or address outstanding feedback.
```
philippsuess, line 49: I think it is a bit cleaner if the two sentences before and after the conditions are merged into one paragraph. I think we should still mention that the reviewer can evaluate if a full review is appropriate, and not state that they should not begin a full review. 

```suggestion
If any of the following conditions are not met, the reviewer **SHOULD** evaluate whether a full review is appropriate. They **MAY** opt for a partial review and **MAY** provide comments to address the issues:

- All CI checks, builds, and tests have completed successfully.
- All requested AI code review agents have completed their reviews.
- All comments from previous review cycles (AI or human) have been addressed, resolved, or explicitly marked as not applicable.
```

develop-docs/sdk/getting-started/playbooks/development/reviewing-a-pr.mdx
```diff
@@ -38,9 +38,15 @@ Related resources:
...
 
-#### 2. Check CI status
+#### 2. Check CI status and prior AI/human reviews
 
-If checks are failing, comment on what's broken and wait. You **MUST NOT** review code that doesn't build or pass tests.
+A reviewer **SHOULD NOT** begin a full review unless all of the following conditions are met:
+
+- All CI checks, builds, and tests have completed successfully.
+- All requested AI code review agents have completed their reviews.
+- All comments from previous review cycles (AI or human) have been addressed, resolved, or explicitly marked as not applicable.
```
philippsuess, line 47: I'm not 100% sure about this bullet point, because this could strictly speaking stop you from reviewing PRs that were already reviewed. Sometimes it's totally fine to get multiple views from multiple maintainers at the same time, and you don't have to wait until all comments from previous review cycles have been addressed. I think maybe we can even remove it.

stephanie-anderson: I think with the change from `MUST NOT` to `SHOULD NOT`, we've covered this case. The rest is human judgement (if an additional review makes sense at that time or not).

develop-docs/sdk/getting-started/playbooks/development/reviewing-a-pr.mdx
stephanie-anderson: >  I think it is a bit cleaner if the two sentences before and after the conditions are merged into one paragraph.

I disagree on this one. The resulting statements are different:

original:
> Don't invest your time if the bare minimum hasn't been done for the PR to be reviewable.

the suggestion
> Invest at least some of your time.

develop-docs/sdk/getting-started/playbooks/development/reviewing-a-pr.mdx
philippsuess: Yep, that's also true.

szokeasaurusrex: That's also kinda why I used "previous cycles" here. If someone else reviewed the changes, and in the same cycle (i.e. before any changes are made), someone else wants to review and comment on things the initial review is missed, I think this should align with the playbook as written now.

develop-docs/sdk/getting-started/playbooks/development/reviewing-a-pr.mdx
philippsuess: I think now I get the point, thanks. The key difference is that the author should make sure that everything is green, and only then people should spend time on the full review. Yep, I get it.

szokeasaurusrex: I am open to restructuring if you find the original wording to be confusing, but I agree with @stephanie-anderson that your proposed wording changes the statement in a way which is potentially undesirable.

**SHOULD NOT** is a recommendation [per RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119#section-4), roughly equivalent to "NOT RECOMMENDED"; so it is already implied that reviewers can choose to do a full review despite unfulfilled conditions on a case-by-case basis

Edit: I wrote this before reading your most recent comment; I think we are all aligned now

stephanie-anderson: Yes, so the underlying principle here is: allow folks to not waste their time. it's the foundation for closing 1-shot PRs that simply are not good enough. and it also underlines that we can & should automate the obvious (CI, AI code review) and apply additional judgement when the time comes.

PR #16594, opened 2026-02-26
feat(tanstackstart-react): Add guide for setup without `--import`

nicohrubec:
For serverless environments where users can't set `--import` flag on startup we need a slightly different setup, where users need to import the `instrument.server.mjs` in the server entry file. So far we have no documentation for that, so I split the docs to set up server-side Sentry into two sections (one for normal node setup, one for serverless environments). The described setup should work starting from the next SDK release.

### DIFF
--- a/docs/platforms/javascript/guides/tanstackstart-react/index.mdx
+++ b/docs/platforms/javascript/guides/tanstackstart-react/index.mdx
@@ -180,9 +180,9 @@ To capture server-side errors and tracing data, you need to explicitly define a 
 
 Follow the section below that matches your deployment setup:
 
-#### Node.js Server
+#### With `--import` Flag (e.g. Node.js Server)
 
-Use this setup when running your application as a long-lived Node.js process (e.g. a traditional server or container).
+Use this setup when you can configure Node.js CLI flags.
 
 Create a `src/server.ts` file in your project:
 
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/guides/tanstackstart-react/index.mdx
```diff
@@ -227,6 +230,30 @@ Add a `--import` flag directly or to the `NODE_OPTIONS` environment variable whe
 }
 ```
 
+#### Serverless (e.g. Vercel, Netlify)
```
andreiborza: m: We should list limitations with this approach. You can look at how we did that for other SDKs.

nicohrubec: right good call, added!

andreiborza (approved):
LGTM!

docs/platforms/javascript/guides/tanstackstart-react/index.mdx
andreiborza: @nicohrubec also now that I think about it, we should not call this "Serverless", have a look at https://docs.sentry.io/platforms/javascript/guides/nuxt/install/cli-import/#scenarios-where---import-does-not-work

Maybe we can rephrase this heading similarly.

nicohrubec: sure I changed the section names to reflect that the deciding difference is whether users can set `--import` or not

andreiborza (approved):
LGTM, this makes it much clearer now!

PR #16641, opened 2026-03-02
fix(develop-docs): Restore old header and fix sidebar collapse on sub-pages

stephanie-anderson:
The nav refresh (fb3c1e192) added the new top navigation header unconditionally
to all doc pages, including developer docs which should use the simpler original
header. This also broke the sidebar on developer docs sub-pages because the
collapsible section `isActive` check used exact path equality instead of
prefix matching.

- Add DevelopDocsHeader component with the original header layout
- Conditionally render DevelopDocsHeader vs Header based on isDeveloperDocs
- Fix DynamicNav isActive to use startsWith so child pages expand their parent section
- Replace MobileMenu in develop docs header with an inline popover showing only external links (Changelog, Sandbox, Go to Sentry, Get Started), since MobileMenu hardcodes mainSections with user-facing doc routes that don't exist on developer docs
- Fix hamburger breakpoint to md:hidden (768px) to match sidebar visibility threshold
- Fix mobile menu breakpoint to lg-xl:hidden (1130px) to match desktop nav, closing the 768-1130px gap where neither menu was visible
- Fix hamburger aria-label from hardcoded "Close" to "Toggle sidebar navigation" and remove aria-hidden


chargome (approved):
🚀

PR #16666, opened 2026-03-04
feat(envelopes): Add envelope item constraints summary

### LINKED ISSUE #16188: Telemetry Processor: What telemetry data can be combined
Define more clearly either in the [telemetry processor develop docs](https://develop.sentry.dev/sdk/telemetry/telemetry-processor/) or somewhere else in the [telemetry docs](https://develop.sentry.dev/sdk/telemetry/) and link to this info from the telemetry processor, to clearly inform SDK developes which types of telemtery data SDKs can combine in envelopes, because this impacts the design of the telemetry processor as it needs to know what it can put into an envelope. For example, an envelope with spans can only contain spans from a single traceID, while this isn't true for logs and metrics.

philipphofmann:
## DESCRIBE YOUR PR

Add an "Envelope Item Constraints" section to the envelope items spec documenting which envelope item types can be combined in a single envelope. This addresses a gap in SDK developer guidance for telemetry processor design.

Closes #16188

## IS YOUR CHANGE URGENT?

- [ ] Other deadline:

## SLA

## PRE-MERGE CHECKLIST

### DIFF
--- a/develop-docs/sdk/foundations/transport/envelope-items.mdx
+++ b/develop-docs/sdk/foundations/transport/envelope-items.mdx
@@ -22,7 +22,7 @@ The headers described in this section are **in addition to the common headers**.
 
 ## Envelope Item Constraints
 
-SDKs **MUST NOT** combine different types of telemetry data in the same envelope, because Relay enforces [rate limits](/sdk/foundations/transport/rate-limiting/) per [data category](/sdk/foundations/transport/rate-limiting/#definitions), and mixing types makes this enforcement more expensive. Exceptions to this rule are listed below (e.g., SDKs **MAY** combine sessions, client reports, and attachments with other items).
+SDKs **MUST NOT** combine different types of telemetry data in the same envelope, because Relay enforces [rate limits](/sdk/foundations/transport/rate-limiting/) per [data category](/sdk/foundations/transport/rate-limiting/#definitions), and mixing types makes this enforcement more expensive. The exceptions to this rule are listed below.
 
 The following constraints apply:
 
@@ -31,7 +31,7 @@ The following constraints apply:
 3. **Logs**: An envelope **MUST** contain at most one envelope item containing logs. SDKs **MAY** mix logs from different traces in the same envelope item. The envelope **MUST NOT** include a [DSC](/sdk/foundations/trace-propagation/dynamic-sampling-context/) header; see [Logs](/sdk/telemetry/logs/#log-envelope-item).
 4. **Metrics**: An envelope **MUST** contain at most one envelope item containing metrics. SDKs **MAY** mix metrics from different traces in the same envelope item. The envelope **MUST NOT** include a [DSC](/sdk/foundations/trace-propagation/dynamic-sampling-context/) header; see [Metrics](/sdk/telemetry/metrics/#trace_metric-envelope-item).
 5. **Sessions**: Multiple envelope items containing sessions **MAY** appear in a single envelope. SDKs **SHOULD** send session updates in the same envelope as their associated error or crash event; see [Sessions](/sdk/telemetry/sessions/#session-update-payload).
-6. **Attachments**: Multiple envelope items containing attachments **MAY** appear in a single envelope. SDKs **SHOULD** send attachments in the same envelope as their associated event, so that if the event is filtered, the attachments are also filtered. SDKs **MUST** send minidump and Apple crash report attachments in the same envelope as their event; see [Attachments](/sdk/telemetry/attachments/#standard-attachments).
+6. **Attachments**: Multiple envelope items containing attachments **MAY** appear in a single envelope. SDKs **SHOULD** send attachments in the same envelope as their associated event, so that if the event is filtered, the attachments are also filtered. For minidump and Apple crash report attachments, this is a **MUST**; see [Attachments](/sdk/telemetry/attachments/#standard-attachments).
 7. **Check-ins**: An envelope **MUST** contain at most one envelope item containing a check-in; see [Check-Ins](/sdk/telemetry/check-ins/#envelope-constraints).
 8. **Client reports**: Multiple envelope items containing client reports **MAY** appear in a single envelope; see [Client Reports](/sdk/telemetry/client-reports/#wire-format).
 9. **Replays**: SDKs **MUST** send the `replay_event` and `replay_recording` envelope items together in the same envelope; see [Replay](/sdk/telemetry/replays/#replay-event-payload).


develop-docs/sdk/foundations/transport/envelope-items.mdx
```diff
@@ -14,6 +20,27 @@ The type of an Item is declared in the `type` header, as well as the payload
...
+## Envelope Item Constraints
+
+SDKs **SHOULD NOT** combine different types of telemetry data in the same envelope, because Relay enforces [rate limits](/sdk/foundations/transport/rate-limiting/) per [data category](/sdk/foundations/transport/rate-limiting/#definitions), and mixing types makes this enforcement more expensive. Exceptions to this rule are listed below (e.g., sessions, client reports, and attachments **MAY** be combined with other items).
+
+The following constraints apply:
+
+1. **Mutual exclusivity**:
+    - [`event`](#event) and [`transaction`](#transaction) envelope items are mutually exclusive.
+    - [`feedback`](/sdk/telemetry/feedbacks/#envelope-item-constraints) and [`transaction`](#transaction) envelope items are mutually exclusive.
+2. **Spans**: An envelope **MUST** contain at most one envelope item containing spans, and all spans **MUST** share a single traceID; see [Span Buffer](/sdk/telemetry/spans/span-buffer/).
```
philippsuess: @cleptric I'm not 100% sure if that's actually the case that you can only send one envelope item containing spans per envelope. If not, I will adjust the docs here accordingly.

develop-docs/sdk/foundations/transport/envelope-items.mdx
```diff
@@ -14,6 +20,27 @@ The type of an Item is declared in the `type` header, as well as the payload
 size in `length`. See Serialization Format for a list of common Item headers.
 The headers described in this section are **in addition to the common headers**.
 
+## Envelope Item Constraints
+
+SDKs **SHOULD NOT** combine different types of telemetry data in the same envelope, because Relay enforces [rate limits](/sdk/foundations/transport/rate-limiting/) per [data category](/sdk/foundations/transport/rate-limiting/#definitions), and mixing types makes this enforcement more expensive. Exceptions to this rule are listed below (e.g., sessions, client reports, and attachments **MAY** be combined with other items).
```
philippsuess: @cleptric, I added a comment for this on the initial GitHub issue https://github.com/getsentry/sentry-docs/issues/16188#issuecomment-3815965519, but I'm not 100% sure if this is actually the case.

And maybe we should even change this to: SDKs **MUST NOT** combine different types of telemetry data in the same envelope. Please let me know.

develop-docs/sdk/foundations/processing/telemetry-processor/index.mdx
```diff
@@ -84,6 +84,8 @@ This exact flow is not enforced for all SDKs. SDKs **MAY** adapt the architectur
 
 We aim to standardize requirements so SDKs share consistent logic across platforms. This benefits multi-platform SDKs like Java. However, requirements **MAY** differ where necessary. This page outlines common requirements; platform-specific requirements are listed [below](#platform-specific-requirements).
 
+For rules on which items can be combined in a single envelope, see [Envelope Item Constraints](/sdk/foundations/transport/envelope-items/#envelope-item-constraints). These constraints impact the design — for example, the [span buffer](#span-buffer) organizes spans per traceID because each envelope can only contain spans from a single trace, while logs and metrics can be batched across traces.
```
coolguyzone:
```suggestion
We aim to standardize requirements so SDKs share consistent logic across platforms. This benefits multi-platform SDKs like Java. However, requirements **MAY** differ where necessary. This page outlines common requirements; platform-specific requirements are listed [below](#platform-specific-requirements).

For rules about which items can be combined in a single envelope, see [Envelope Item Constraints](/sdk/foundations/transport/envelope-items/#envelope-item-constraints). These constraints impact the design — for example, the [span buffer](#span-buffer) organizes spans per traceID because each envelope can only contain spans from a single trace, while logs and metrics can be batched across traces.
```

develop-docs/sdk/foundations/transport/envelope-items.mdx
cleptric: You can only emit one span container (the envelope item).

develop-docs/sdk/foundations/transport/envelope-items.mdx
cleptric: With all the size issues we are fighting lately, I would just explicitly state that everything is their own envelope.
Though if we are concerned about the impact that could have, e.g. on mobile, let's chat.

philippsuess: >  I would just explicitly state that everything is their own envelope.

@cleptric, is that also true for sessions? I know that we have always had this requirement that sessions should be in the same envelope as their corresponding crash event or error event, because otherwise it gets quite expensive. 

And what about attachments and user feedback? I also know that attachments and user feedback should also be in the same envelope as their related events.

cleptric: Sessions are also sent without an error, so this makes it even simpler. Attachments today do not work standalone, hence we can't change it. But trace attachments can be send on their own. User Feedback is such low volume that I wouldn't bother with all this merging logic.

jjbayer (approved):

develop-docs/sdk/foundations/transport/envelope-items.mdx
Dav1dde: I would like to make this `must not` with a defined list of exceptions what is allowed. For example, sessions may be mixed into envelopes (like you suggested). 

Size limits are usually a problem per type, you rarely run into overall envelope limits (that is 200MiB -> attachments).

develop-docs/sdk/foundations/transport/envelope-items.mdx
```diff
@@ -14,7 +20,28 @@ The type of an Item is declared in the `type` header, as well as the payload
...
-## Event
+## Envelope Item Constraints
+
+SDKs **SHOULD NOT** combine different types of telemetry data in the same envelope, because Relay enforces [rate limits](/sdk/foundations/transport/rate-limiting/) per [data category](/sdk/foundations/transport/rate-limiting/#definitions), and mixing types makes this enforcement more expensive. Exceptions to this rule are listed below (e.g., sessions, client reports, and attachments **MAY** be combined with other items).
+
+The following constraints apply:
+
+1. **Mutual exclusivity**:
+    - [`event`](#event) and [`transaction`](#transaction) envelope items are mutually exclusive.
+    - [`feedback`](/sdk/telemetry/feedbacks/#envelope-item-constraints) and [`transaction`](#transaction) envelope items are mutually exclusive.
```
Dav1dde: `feedback` is also mutually exlcusive with `event`/`error`. Every type of event is mutually exclusive with a different type of event (feedback is an event).

Since, for example, attachments are implicitly scoped to the event in the envelope -> there can only be one. Also the envelope header carries an event_id which must match the one in the envelope.

develop-docs/sdk/foundations/transport/envelope-items.mdx
```diff
@@ -14,7 +20,28 @@ The type of an Item is declared in the `type` header, as well as the payload
...
+2. **Spans**: An envelope **MUST** contain at most one envelope item containing spans, and all spans **MUST** share a single traceID; see [Span Buffer](/sdk/telemetry/spans/span-buffer/).
```
Dav1dde: Only one span container per envelope, the envelope must have a DSC (in the envelope header) -> therefor all spans in the envelope must belong that that DSC

develop-docs/sdk/foundations/transport/envelope-items.mdx
```diff
@@ -14,7 +20,28 @@ The type of an Item is declared in the `type` header, as well as the payload
...
+3. **Logs**: An envelope **MUST** contain at most one envelope item containing logs. Logs from different traces **MAY** be mixed in the same envelope item, but if they are, the envelope **MUST NOT** include a [DSC](/sdk/foundations/trace-propagation/dynamic-sampling-context/) header; see [Logs](/sdk/telemetry/logs/#log-envelope-item).
+4. **Metrics**: An envelope **MUST** contain at most one envelope item containing metrics. Metrics from different traces **MAY** be mixed in the same envelope item, but if they are, the envelope **MUST NOT** include a [DSC](/sdk/foundations/trace-propagation/dynamic-sampling-context/) header; see [Metrics](/sdk/telemetry/metrics/#trace_metric-envelope-item).
```
Dav1dde: We can simplify this to: Logs and metrics must never have a DSC attached on the envelope.

develop-docs/sdk/foundations/transport/envelope-items.mdx
```diff
@@ -14,7 +20,28 @@ The type of an Item is declared in the `type` header, as well as the payload
...
+5. **Sessions**: Multiple envelope items containing sessions **MAY** appear in a single envelope; see [Sessions](/sdk/telemetry/sessions/#session-update-payload).
+6. **Attachments**: Multiple envelope items containing attachments **MAY** appear in a single envelope. Minidump and Apple crash report attachment envelope items **MUST** be sent in the same envelope as their event; see [Attachments](/sdk/telemetry/attachments/#standard-attachments).
```
Dav1dde: Attachments should be sent with their event.

E.g. while you can send standalone attachments, you should send them with the error, in case the error is filterd, the attachments will also be filtered.

philippsuess:
Thanks a lot @Dav1dde for the review. I think I addressed all your valid concerns. Please have another look.

PR #16787, opened 2026-03-05
docs(build-distribution): Document install groups for auto-update SDK

runningcode:
## DESCRIBE YOUR PR

Document the dual role of install groups in Build Distribution: tagging builds at upload time and filtering available updates at runtime via the Auto-Update SDK.

- Expand "How It Works" in product docs to explain both upload tagging and runtime filtering
- Add "How Install Groups Work With the Auto-Update SDK" section to Android install-groups page
- Add "Install Groups" section and `installGroups` config option to auto-update page
- Document programmatic override via `SentryOptions`

## IS YOUR CHANGE URGENT?

- [ ] Other deadline:

## SLA

## PRE-MERGE CHECKLIST

### DIFF
--- a/docs/product/build-distribution/index.mdx
+++ b/docs/product/build-distribution/index.mdx
@@ -51,7 +51,7 @@ Install groups are configured in two places, which can be set independently:
 When the SDK checks for updates, the API returns the single latest build (highest semver version, with build number as tiebreaker) whose install groups overlap with the filter. Install groups for the filter are determined as follows:
 
 - If the SDK **provides groups explicitly**, the API uses those.
-- If the SDK **doesn't provide groups**, the API looks up the uploaded build matching the device's current version and uses that build's upload groups. On iOS, the lookup uses the UUID of the app binary, which is unique per binary. On Android, it uses `buildVersion` and `buildNumber` — if multiple builds share the same values, the API picks the most recently uploaded one, which may lead to unexpected results.
+- If the SDK **doesn't provide groups**, the API looks up the uploaded build matching the device's current version and uses that build's upload groups. On iOS, the lookup uses the UUID of the app binary. The UUID only changes when the main binary changes, so uploads that differ only in other files (such as images) won't be differentiated. On Android, it uses `buildVersion` and `buildNumber` — if multiple builds share the same values, the API picks the most recently uploaded one, which may lead to unexpected results.
 
 We recommend configuring the SDK to send install groups explicitly rather than relying on API-side inheritance. On Android, the [Gradle plugin](/platforms/android/build-distribution/install-groups/#gradle-plugin) `installGroups` config handles this automatically — it tags the upload **and** embeds the groups in the app so the SDK sends them with every update check.
 
--- a/docs/platforms/apple/guides/ios/build-distribution/auto-update.mdx
+++ b/docs/platforms/apple/guides/ios/build-distribution/auto-update.mdx
@@ -79,7 +79,7 @@ struct MyView: View {
 
 Install groups let you control which build the Auto-Update SDK returns when checking for updates. The API always returns a single build — the latest (by semver version, then build number) whose install groups overlap with the filter. Tag your builds with install groups at upload time using sentry-cli or the Fastlane plugin, and the API automatically filters update checks — a device running a build uploaded with `["beta"]` will only see updates from other `["beta"]` builds.
 
-Unlike Android, iOS builds are identified by the UUID of the app binary, which is unique per binary. This means the API can always reliably look up the correct build and its install groups, so no additional SDK-side configuration is needed.
+Unlike Android, iOS builds are identified by the UUID of the app binary. However, UUID matching is not guaranteed to identify the exact build you uploaded — the UUID only changes when the main binary changes, so uploads that differ only in other files (such as images) won't be differentiated. For deterministic filtering, configure the SDK to [provide install groups explicitly](/platforms/apple/guides/ios/build-distribution/install-groups/#sdk-configuration).
 
 See the [Install Groups guide](/platforms/apple/guides/ios/build-distribution/install-groups/) for upload instructions and full details on how filtering works.
 


docs/platforms/android/build-distribution/auto-update.mdx
```diff
@@ -125,6 +126,38 @@ suspend fun checkForUpdate() {
 }
 ```
 
+## Install Groups
+
+Install groups let you control which builds the Auto-Update SDK considers as available updates. When configured, the SDK only sees builds that share at least one install group with the current app. See the [Install Groups guide](/platforms/android/build-distribution/install-groups/) for full upload instructions.
```
mtopo27: "When configured, the SDK only sees builds that share at least one install group with the current app"

Does the programmatic override section below mean that they can set the logic to change the matching logic? i.e. they can configure to update to builds that share >1 || all install groups?

If so, I think it makes sense to change "When configured" -> "By default" or something like that and then mention the logic can be changed and hyperlink to the section below

docs/platforms/android/build-distribution/install-groups.mdx
```diff
@@ -58,3 +58,26 @@ sentry_upload_build(
...
 ```
+
+## How Install Groups Work With the Auto-Update SDK
+
+When you use the Gradle plugin's `installGroups` config, it does two things:
+
+1. **Upload**: Passes `--install-group` flags to sentry-cli so the build is tagged on the server.
+2. **Embed**: Writes the groups to `sentry-distribution.properties` inside the APK (as `io.sentry.distribution.install-groups-override`). The [Auto-Update SDK](/platforms/android/build-distribution/auto-update/) reads this value at runtime and sends the groups as query parameters when checking for updates.
+
+The server then only returns builds whose install groups overlap with the ones sent by the SDK.
```
mtopo27: "The server then only returns builds whose install groups overlap with the ones sent by the SDK."

Seems somewhat vague and reading this I'm not sure if it's match for 1 / some / all. Needs to be clarified or linked to somewhere with the answer

docs/product/build-distribution/index.mdx
```diff
@@ -43,9 +43,18 @@ This is useful when you have multiple distribution channels (for example, separa
 
 ### How It Works
 
-When you upload a build, you can assign one or more install groups to it. If the Auto-Update SDK is provided install groups, it only returns builds that have overlapping groups.
+Install groups serve two purposes:
 
-Matching requires two builds to have a non-empty intersection of their install groups. For example, a build tagged `["alpha", "staging"]` will see updates from a build tagged `["alpha", "beta"]` because both share `alpha`.
+1. **At upload time**: Tag each build with one or more group names (for example, `"alpha"`, `"staging"`).
+2. **At runtime**: The Auto-Update SDK sends the device's install groups when checking for updates. The server only returns builds that share at least one group with the request.
```
mtopo27: Same comments as above: 
"The server only returns builds that share at least one group with the request." -> can this be customized? not clear and still not sure how this plays with the "programmatic override"

docs/platforms/apple/guides/ios/build-distribution/install-groups.mdx
```diff
@@ -36,3 +36,12 @@ sentry_upload_build(
   install_groups: ['alpha', 'staging']
 )
 ```
+
+## How Install Groups Work With the Auto-Update SDK
+
+When the [Auto-Update SDK](/platforms/apple/guides/ios/build-distribution/auto-update/) checks for updates, the server returns the single latest build (highest semver version, with build number as tiebreaker) whose install groups overlap with the filter:
```
noahsmartin: I would replace all "server" with "API" just sounds friendlier

docs/platforms/apple/guides/ios/build-distribution/install-groups.mdx
```diff
@@ -36,3 +36,12 @@ sentry_upload_build(
...
+
+- If the SDK **sends groups explicitly**, the server uses those to filter.
```
noahsmartin: sends -> provides on this line and the next

docs/platforms/apple/guides/ios/build-distribution/install-groups.mdx
```diff
@@ -36,3 +36,12 @@ sentry_upload_build(
...
+- If the SDK **doesn't send groups**, the server looks up the uploaded build using the device's Mach-O UUID of the app binary and uses that build's upload groups for filtering.
```
noahsmartin: delete "device's Mach-O", should just be "using the UUID of the app binary"

docs/platforms/apple/guides/ios/build-distribution/install-groups.mdx
```diff
@@ -36,3 +36,12 @@ sentry_upload_build(
...
+
+Because the Mach-O UUID is unique per binary, the server can reliably identify the correct build. This means tagging builds with install groups at upload time is sufficient on iOS — the server inherits the correct groups automatically when checking for updates.
```
noahsmartin: This isn't quite the case. UUID lookup is best effort. Change it to say:

```
UUID matching is not guaranteed to identify the build you uploaded. The UUID is only unique to an app's main binary, so an upload that only changes other files such as images won't be differentiated.
```

docs/product/build-distribution/index.mdx
```diff
@@ -37,15 +37,25 @@ By default, Build Distribution processes all uploaded builds. You can configure
 
 ## Install Groups
 
-Install groups let you tag builds with one or more group names to control update visibility. When a device checks for updates, you can provide a list of install groups — only builds that share at least one install group with this list will be returned as available updates.
+Install groups let you tag builds with one or more group names to control update visibility. When the Auto-Update SDK checks for updates, the API returns the single latest build (by semver version, then build number) that shares at least one install group with the request.
```
mtopo27, line 40: Might need to wrap around, but install groups will also be available to use in the API + the UI. For now this is fine but will need to update when that's wired up

docs/platforms/android/build-distribution/auto-update.mdx
```diff
@@ -125,6 +127,38 @@ suspend fun checkForUpdate() {
...
+    installGroups.set(setOf("alpha", "staging"))
+  }
+}
+```
+
+The Gradle plugin tags the uploaded build with these groups **and** embeds them in the app so the SDK sends them explicitly when checking for updates. This is the recommended approach — it ensures deterministic filtering regardless of how many builds share the same version number.
+
+### Programmatic Override
+
+You can override the groups the SDK sends at runtime to filter by different groups than what the build was uploaded with:
```
coolguyzone:
```suggestion
You can override the groups the SDK sends at runtime to filter by groups different from those included in the uploaded build:
```

PR #16788, opened 2026-03-05
feat(dev): add client reports architecture guidance

giortzisg:
## DESCRIBE YOUR PR
As per this [discussion](https://github.com/getsentry/sentry-docs/pull/15983#discussion_r2705136789), this PR clarifies architecture questions for how client reports should work along with the Telemetry Processor.

Resolves: ENG-6381 and GO-100

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

## LEGAL BOILERPLATE

## EXTRA RESOURCES

### DIFF
--- a/develop-docs/sdk/telemetry/client-reports.mdx
+++ b/develop-docs/sdk/telemetry/client-reports.mdx
@@ -223,7 +223,7 @@ SDKs **SHOULD** use a single, centralized `ClientReports` component per client t
 
 ### Telemetry Processor Integration
 
-Early SDK implementations tracked client reports inside the transport, but as the number of telemetry data types has grown and data can be dropped at multiple points throughout the SDK, this approach no longer scales well. The transport only sees items at send time and has no visibility into earlier drops.
+The original specification suggested tracking client reports directly within the transport. However, due to the increase in telemetry data types and the possibility of data loss at various stages in the SDK, this approach no longer scales well. Client reports must be accessible across different parts of the SDK. Thus, we now recommend developing a separate `ClientReports` component rather than integrating it into the transport.
 
 SDKs **SHOULD** own the `ClientReports` component at the client level, regardless of whether they implement the full [telemetry processor](/sdk/foundations/processing/telemetry-processor/) architecture. This keeps the transport focused on delivery.
 


develop-docs/sdk/foundations/processing/telemetry-processor/index.mdx
```diff
@@ -154,15 +154,17 @@ Here are the common requirements for all platforms:
...
+The telemetry scheduler **SHOULD** add [client reports](/sdk/telemetry/client-reports/) to outgoing envelopes or periodically send standalone client report envelopes to the transport, following the [SDK recommendations](/sdk/telemetry/client-reports/#sdk-recommendations).
 
-Client report tracking **SHOULD NOT** be implemented in the telemetry scheduler. Instead, the SDK **SHOULD** have `ClientReports` accessible throughout the SDK, since data can be dropped in multiple places, as shown in the diagram below. The telemetry scheduler **SHOULD** request the current client reports from `ClientReports`.
+Client report aggregation **SHOULD NOT** be implemented in the telemetry scheduler. Instead, the SDK **SHOULD** have a centralized `ClientReports` component per client, accessible throughout the SDK, since data can be dropped in multiple places, as shown in the diagram below. `ClientReports` handles aggregation internally. The telemetry scheduler **SHOULD** call an extraction method on `ClientReports` to atomically retrieve and reset the current aggregated report.
```
philippsuess: `l`: Maybe slightly more specific
```suggestion
Client report aggregation **SHOULD NOT** be implemented in the telemetry scheduler. Instead, the SDK **SHOULD** have a centralized `ClientReports` component per client, accessible throughout the SDK, since data can be dropped in multiple places, as shown in the diagram below. `ClientReports` handles aggregation internally. The telemetry scheduler **SHOULD** call an extraction method on `ClientReports` to atomically retrieve and reset the current aggregated client report.
```

develop-docs/sdk/foundations/processing/telemetry-processor/index.mdx
```diff
@@ -154,15 +154,17 @@ Here are the common requirements for all platforms:
...
 
-To be defined: How exactly does the telemetry scheduler interact with the transport to correctly remove items from the telemetry queue based on the received HTTP info and how to adhere to the [dealing with network failure](/sdk/foundations/transport/offline-caching/#dealing-with-network-failures) requirements.
+To be defined: How exactly does the telemetry scheduler interact with the transport to correctly remove items from the telemetry queue based on the received HTTP info and how to adhere to the [dealing with network failure](/sdk/expected-features/#dealing-with-network-failures) requirements.
 
 ## Client Reports
 
-The telemetry scheduler **SHOULD** add [client reports](/sdk/telemetry/client-reports/) to outgoing envelopes or periodically send standalone client report envelopes to the transport, following the [SDK side recommendations](/sdk/telemetry/client-reports/#sdk-side-recommendations).
+The telemetry scheduler **SHOULD** add [client reports](/sdk/telemetry/client-reports/) to outgoing envelopes or periodically send standalone client report envelopes to the transport, following the [SDK recommendations](/sdk/telemetry/client-reports/#sdk-recommendations).
```
philippsuess: `m`: I think that should now move down to the sending part, or we remove it here. It's now duplicate.

develop-docs/sdk/foundations/processing/telemetry-processor/index.mdx
philippsuess: `l`: 
```suggestion
Client reports aggregation **SHOULD NOT** be implemented in the telemetry scheduler. Instead, the SDK **SHOULD** have a centralized `ClientReports` component per client, accessible throughout the SDK, since data can be dropped in multiple places, as shown in the diagram below. `ClientReports` handles aggregation internally. The telemetry scheduler **SHOULD** call an extraction method on `ClientReports` to atomically retrieve and reset the current aggregated report.
```

develop-docs/sdk/foundations/processing/telemetry-processor/index.mdx
```diff
@@ -154,15 +154,17 @@ Here are the common requirements for all platforms:
...
 
-Client report tracking **SHOULD NOT** be implemented in the telemetry scheduler. Instead, the SDK **SHOULD** have `ClientReports` accessible throughout the SDK, since data can be dropped in multiple places, as shown in the diagram below. The telemetry scheduler **SHOULD** request the current client reports from `ClientReports`.
+Client report aggregation **SHOULD NOT** be implemented in the telemetry scheduler. Instead, the SDK **SHOULD** have a centralized `ClientReports` component per client, accessible throughout the SDK, since data can be dropped in multiple places, as shown in the diagram below. `ClientReports` handles aggregation internally. The telemetry scheduler **SHOULD** call an extraction method on `ClientReports` to atomically retrieve and reset the current aggregated report.
 
-Here is a diagram showing how `ClientReports` interacts with the telemetry scheduler.
+See the [client reports spec](/sdk/telemetry/client-reports/#telemetry-processor-integration) for the full architecture details, per-client scoping requirements, and recording points.
+
+**Recording:** Multiple components call `Record` on `ClientReports` when they drop telemetry.
```
philippsuess: `l`: I think we can make this a heading so it's a bit better structured. 
```suggestion
### Recording

Multiple components call `Record` on `ClientReports` when they drop telemetry items.
```

develop-docs/sdk/foundations/processing/telemetry-processor/index.mdx
```diff
@@ -177,7 +179,20 @@ flowchart LR
   TelemetryBuffer -- "records: buffer_overflow, ..." --> ClientReports
   TelemetryScheduler -- "records: queue_overflow, ..." --> ClientReports
   Transport -- "records: send_error, network_error" --> ClientReports
-  ClientReports -- "aggregated reports" --> TelemetryScheduler
+```
+
+**Sending:** The component on the sending path extracts the aggregated report and attaches it to outgoing envelopes, or periodically sends it as a standalone envelope.
```
philippsuess: `l`: For better structure 
```suggestion
### Sending 

The component on the sending path extracts the aggregated report and attaches it to outgoing envelopes, or periodically sends it as a standalone envelope.
```

develop-docs/sdk/telemetry/client-reports.mdx
```diff
@@ -201,15 +204,31 @@ If certain spans are dropped in `beforeSendTransaction`, an event processor, etc
 
 </SpecSection>
 
-<SpecSection id="telemetry-processor-integration" status="candidate" since="1.18.0">
+<SpecSection id="log-byte-outcomes" status="stable" since="1.20.0">
 
-### Telemetry Processor Integration
+### Log Byte Outcomes
+
+When a log is dropped, SDKs **MUST** record an additional `log_byte` category entry containing the serialized byte size of the log. This provides visibility into how much log data is being discarded, complementing the item count reported under the `log` category. Counting the log size is best effort, and the SDKs can send an approximation of the log size (eg. by counting as if serialized).
```
philippsuess: `l`: That doesn't have much to do with the telemetry processor. I think this could be in its own pull request, but if that's too much work for you, we can also keep it.

philippsuess: `m`: To me, the approximation is a bit unclear about what exactly it does by counting as if it were a serialized mean.

This is an argument for me to maybe discuss in a separate PR, so we can keep the discussion here focused on telemetry processors, but up to you!

develop-docs/sdk/telemetry/client-reports.mdx
```diff
@@ -201,15 +204,31 @@ If certain spans are dropped in `beforeSendTransaction`, an event processor, etc
...
+
+</SpecSection>
+
+<SpecSection id="aggregation" status="candidate" since="1.20.0">
+
+### Aggregation
 
-SDKs **MAY** keep track of discarded events in the transport. Once SDKs implement the [telemetry processor](/sdk/foundations/processing/telemetry-processor/), they **SHOULD** implement client reports tracking outside the transport, since data can be dropped in multiple places throughout the SDK.
+SDKs **SHOULD** use a single, centralized `ClientReports` component per client, accessible throughout the SDK. Client reports **MUST** be scoped per client — if an SDK supports multiple client instances, each **MUST** have its own `ClientReports` instance. When client reports are disabled, SDKs **SHOULD** use a no-op implementation to avoid overhead.
```
philippsuess: `l`: I think we could also explain here more clearly why we wanna scope this per client. The reason is that clients are already scoped per DSN, and client reports are also scoped per DSN.

develop-docs/sdk/telemetry/client-reports.mdx
```diff
@@ -201,15 +204,31 @@ If certain spans are dropped in `beforeSendTransaction`, an event processor, etc
...
+
+`ClientReports` handles aggregation internally — it maintains a map of `(reason, category)` to count, and components call a `Record` method to increment counters directly. The component **MUST** be safe for concurrent access, since multiple SDK components record into it simultaneously. The extraction operation **SHOULD** atomically read and reset all counters to prevent double-counting.
```
philippsuess: `m`: This should be a must, I think. 
```suggestion
`ClientReports` handles aggregation internally — it maintains a map of `(reason, category)` to count, and components call a `Record` method to increment counters directly. The component **MUST** be safe for concurrent access, since multiple SDK components record into it simultaneously. The extraction operation **MUST** atomically read and reset all counters to prevent double-counting.
```

develop-docs/sdk/telemetry/client-reports.mdx
```diff
@@ -201,15 +204,31 @@ If certain spans are dropped in `beforeSendTransaction`, an event processor, etc
...
+
+</SpecSection>
+
+<SpecSection id="telemetry-processor-integration" status="candidate" since="1.20.0">
+
+### Telemetry Processor Integration
 
-The [telemetry scheduler](/sdk/foundations/processing/telemetry-processor/#telemetry-scheduler) is then responsible for passing aggregated client reports to the transport. The telemetry scheduler still **MUST** adhere to all requirements in this spec.
+SDKs **MAY** keep track of discarded events in the transport. Once SDKs implement the [telemetry processor](/sdk/foundations/processing/telemetry-processor/), they **SHOULD** track client reports outside the transport, since data can be dropped at multiple points throughout the SDK. Multiple components record into `ClientReports` and the [telemetry processor](/sdk/foundations/processing/telemetry-processor/) is responsible for extracting and sending the aggregated report.
```
philippsuess: > since data can be dropped at multiple points throughout the SDK.

`m`: I think that is already true right now. The key here is that we don't want to expose the transport to various parts of the SDK, and we want to keep it more focused on sending. The client shouldn't talk to the transport directly anymore, but only to the telemetry processor, once the telemetry processor architecture is fully implemented. I think we can already be bold and say, "Hey, no matter if you implement a Telemetry Process Architecture or not, the SDK should consider also keeping track of the client reports in the client."  We can state that the initial implementation said, "Hey, the transport should keep track of it," but now that we have more of telemetry data item types and drop data in various parts of the SDK, this approach doesn't work really well anymore. Basically, we could recommend the client keeping track of the client reports component.

develop-docs/sdk/telemetry/client-reports.mdx
philippsuess: `l`: I even forgot that you can disable this with an option. Maybe we should add it here to clearly state how you can disable it. 
```suggestion
SDKs **SHOULD** use a single, centralized `ClientReports` component per client, accessible throughout the SDK. Client reports **MUST** be scoped per client — if an SDK supports multiple client instances, each **MUST** have its own `ClientReports` instance. When client reports are disabled, via the `sendClientReports` option, SDKs **SHOULD** use a no-op implementation to avoid overhead.
```

develop-docs/sdk/telemetry/client-reports.mdx
philippsuess: `l`: I think we could split this up into bullet points, so basically having one requirement per bullet point.

philippsuess: `m`: I think it would be great to explain a bit more clearly how this map works, so that the key is a string combined with a reason category, and the value is basically the integer for counting. So something like this:
```suggestion
`ClientReports` handles aggregation internally. It SHOULD use a map where the key is `reason- category` and the value is the number of dropped items to count. Components dropping data call a `Record` method to increment counters directly. The `ClientReports` **MUST** be safe for concurrent access, since multiple SDK components record into it simultaneously. The extraction operation **SHOULD** atomically read and reset all counters to prevent double-counting.
```

develop-docs/sdk/telemetry/client-reports.mdx
philippsuess: `m`: It's not clear to me how this translates to the JSON. I think it would be great to give a JSON example.

philippsuess: `m`: How does this work with the aggregation? I think below we defined that we have a map of string to ints, but how does this then work with the map when we also need to record the byte size? Should we add a specific map just for the logs, or how should we guide users to implement this? It would be great to add this information to give guidance.

develop-docs/sdk/foundations/processing/telemetry-processor/index.mdx
philippsuess: `m`: I think we have to be a bit more specific about what exactly the component on the sending path is. I thought that the telemetry scheduler is responsible for this.

develop-docs/sdk/foundations/processing/telemetry-processor/index.mdx
```diff
@@ -177,7 +179,20 @@ flowchart LR
...
+```mermaid
+flowchart LR
+  ClientReports["ClientReports"]
+  TelemetryQueue["Telemetry Queue"]
+  Sender["Telemetry Scheduler / Transport"]
+
+  TelemetryQueue -- "envelope" --> Sender
+  ClientReports -- "TakeReport()" --> Sender
+  Sender -- "envelope + client report attached" --> Sentry
+  Sender -- "standalone client report envelope\n(periodic, to avoid starvation)" --> Sentry
```
philippsuess: `m`: I'm not sure if we should put the sender here in the diagram. I think we should be specific about what exactly the sender is, and I think that should be the telemetry scheduler.

Right now, it seems to me that we want to communicate with this diagram that it's still okay that the transport might send out the client reports, but ideally that shouldn't be the case. I think we could maybe mention that there can be a migration phase in between when the client reports component has not fully been converted over, but the diagram should show the ideal state. We could maybe also create a diagram for the migration phase, but I find this a bit confusing now. 

I had something like this in my mind, let me know if that would work or not. 

Image

```suggestion
```mermaid
flowchart LR
  ClientReports["ClientReports"]

  subgraph TelemetrySchedulerSubGraph["TelemetryScheduler"]
    TelemetryQueue["**TelemetryQueue**"]
    TelemetryScheduler["**TelemetryScheduler**"]
    TelemetryQueue -- "envelope" --> TelemetryScheduler
    TelemetryScheduler -- "add or remove envelope" --> TelemetryQueue
  end

  Transport["Transport"]

  ClientReports -- "TakeReport()" --> TelemetryScheduler
  TelemetryScheduler -- "sendEnvelope" --> Transport

**Review by philipphofmann — COMMENTED** [review]
Thanks a lot for tackling this, @giortzisg. Sorry, I added plenty of comments. We can also hop on a call to discuss these if you want to speed things up. I also added a couple of nitpicks.

File: develop-docs/sdk/foundations/processing/telemetry-processor/index.mdx (hunk 2 above)
Comment (giortzisg) [reply develop-docs/sdk/foundations/processing/telemetry-processor/index.mdx]: Makes sense to remove it.

File: develop-docs/sdk/telemetry/client-reports.mdx (hunk 5 above)
Comment (giortzisg) [reply develop-docs/sdk/telemetry/client-reports.mdx]: Can separate it.

Comment (giortzisg) [reply develop-docs/sdk/telemetry/client-reports.mdx]: https://github.com/getsentry/sentry-docs/pull/16822

Comment (giortzisg) [reply develop-docs/sdk/telemetry/client-reports.mdx]: Tried to keep the same phrasing as here https://develop.sentry.dev/sdk/telemetry/logs/#client-reports, but let's discuss on the separate PR.

File: develop-docs/sdk/foundations/processing/telemetry-processor/index.mdx (hunk 4 above)
Comment (giortzisg) [reply develop-docs/sdk/foundations/processing/telemetry-processor/index.mdx]: wanted to keep it vague, since other sdks might not have fully implemented the telemetry processor, but since it's the processor spec it makes sense to be more specific.

File: develop-docs/sdk/telemetry/client-reports.mdx [hunk 10]
```diff
@@ -201,15 +204,32 @@ If certain spans are dropped in `beforeSendTransaction`, an event processor, etc
...
+
+### Aggregation
+
+SDKs **SHOULD** use a single, centralized `ClientReports` component per client that handles aggregation internally:
+
+- **MUST** be scoped per client — if an SDK supports multiple client instances, each **MUST** have its own `ClientReports` instance. This is required because clients are already scoped per DSN, so scoping `ClientReports` per client ensures report data is attributed to the correct DSN and avoids mixing data across instances.
+- **SHOULD** be accessible throughout the SDK, so any component that drops telemetry can record into it.
+- **SHOULD** use a no-op implementation when client reports are disabled via the `sendClientReports` option, to avoid overhead.
+- **SHOULD** maintain a map where the key is a `reason-category` string and the value is the number of dropped items. Components call a `Record` method to increment counters directly.
+- **MUST** be safe for concurrent access, since multiple SDK components record into it simultaneously.
```
philippsuess: `l`: A bit more specific.
```suggestion
- **MUST** be thread safe for concurrent read and write access, since multiple SDK components record into it simultaneously.
```

develop-docs/sdk/telemetry/client-reports.mdx
```diff
@@ -201,15 +204,32 @@ If certain spans are dropped in `beforeSendTransaction`, an event processor, etc
...
+- **MUST** atomically read and reset all counters during extraction to prevent double-counting.
+
+</SpecSection>
+
+<SpecSection id="telemetry-processor-integration" status="candidate" since="1.20.0">
 
 ### Telemetry Processor Integration
 
-SDKs **MAY** keep track of discarded events in the transport. Once SDKs implement the [telemetry processor](/sdk/foundations/processing/telemetry-processor/), they **SHOULD** implement client reports tracking outside the transport, since data can be dropped in multiple places throughout the SDK.
+Early SDK implementations tracked client reports inside the transport, but as the number of telemetry data types has grown and data can be dropped at multiple points throughout the SDK, this approach no longer scales well. The transport only sees items at send time and has no visibility into earlier drops.
```
philippsuess: `m`: Well, Cocoa solved this problem by basically forwarding client reports to the client and then to the transport, actually. So that's how the transport had visibility into earlier drops.

I think it doesn't make sense now to basically pass dropped data items from somewhere in the SDK to the client to the telemetry processor down to the transport, and then again back up to the telemetry scheduler. I think that doesn't work well. The key here is that the telemetry scheduler component now adds client reports and not the transport anymore, because it needs to balance out the offline cache and priority queues and client reports. The transport should focus mostly on sending events, with little knowledge of client reports and such. It still would work; it just doesn't seem like a great approach.

So the main reason to put it on clients is that we want a component for client reports that's easier to access throughout the SDK. 

So, all that said, now I gave a lot of context, sorry 😅 . I think we should rephrase this slightly and do a very short and precise summary of what I just said. Why it should be scoped per client and keeping in the transport doesn't make sense. It should be a component on its own. Those are basically the two most important points. 

```suggestion
The original specification suggested tracking client reports directly within the transport. However, due to the increase in telemetry data types and the possibility of data loss at various stages in the SDK, this approach no longer scales well. Client reports must be accessible across different parts of the SDK. Thus, we now recommend developing a separate `ClientReports` component rather than integrating it into the transport.
```

PR #16796, opened 2026-03-05
docs(sdks): Add spec for `dataCollection` option to supersede `sendDefaultPii`

s1gr1d:
This PR extends the Data Collection spec so it is the single place for what SDKs collect and how they scrub it. It adds concrete denylist behavior, request-body and cookie rules, and pulls in the relevant scrubbing behavior from the Data Handling doc.
- Data Collection spec (`/sdk/foundations/client/data-collection/`)
	- https://develop-docs-git-sig-data-collection-options.sentry.dev/sdk/foundations/client/data-collection/
	
	

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST


develop-docs/sdk/expected-features/data-handling.mdx
stephanie-anderson, line 1: This page does not exist anymore - we moved the PII related content to 
https://develop.sentry.dev/sdk/foundations/data-scrubbing/

cleptric:
We should include some more general information, such as how we expect the default snippet too like now, which is very minimal and in-line with the behavior of `sendDefualtPii`.
```
sentry.init({
  dsn: '...',
})

// or with user information

sentry.init({
  dsn: '...',
  dataCollection: {
    includeUserInfo: true,     
  }
})
```
and also mention the reasoning of this change. In particular, we should highlight that we now include more context by default, without a change in our position to be privacy first.

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -0,0 +1,327 @@
...
+
+#### Matching Rule
+
+SDKs **MUST** perform a **partial, case-insensitive match** when comparing key names against the denylist. A key is treated as sensitive if any denylist term appears as a substring in the key name (e.g. the term `auth` matches `Authorization` and `X-Auth-Token`).
+
+#### Base Denylist (Sensitive Data)
+
+The following terms **MUST** be included in the default denylist for headers, and **SHOULD** be applied to cookies and query params where applicable:
+
+`["auth", "token", "secret", "password", "passwd", "pwd", "key", "jwt", "bearer", "sso", "saml", "csrf", "xsrf", "credentials", "session", "sid", "identity"]`
```
itaybre, line 120: `m`: We have some additional filtered headers on cocoa that may be relevant here (https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Swift/Core/Tools/HTTPHeaderSanitizer.swift#L8): `X-REAL-IP` and `REMOTE-ADDR`

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -0,0 +1,327 @@
...
+- Any key ending with or containing **`-user`** (e.g. `x-user-id`, `remote-user`) — often carries user identifiers
+
+Effective denylist when PII is disabled: base list + `["x-forwarded-", "-user"]` (partial match, case-insensitive).
+
+#### Cookies and Cookie Headers
+
+- SDKs **SHOULD** maintain a default denylist of cookie names using the same matching rule (e.g. `session`, `auth`, `identity`). Values for matching cookie names **MUST** be replaced with `"[Filtered]"`.
+- **When individual cookie key-value pairs cannot be extracted** (e.g. malformed or opaque cookie string), the entire `Cookie` or `Set-Cookie` header value **MUST** be replaced with `"[Filtered]"`. Unfiltered raw cookie header values **MUST NOT** be sent. When in doubt, treat the whole cookie header as sensitive.
+
+#### Request Bodies
```
itaybre: `m`: Should the same apply for response bodies? This is (or will be, depends on the SDK) being recorded now for Session Replay

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -0,0 +1,327 @@
...
+
+No built-in option scrubs **keys**; users who need to hide header or cookie names **MUST** use `beforeSend` (or equivalent).
+
+</SpecSection>
+
+<SpecSection id="user-set-data-scrubbing" status="draft" since="1.0.0">
+
+### User-Set Data and Scrubbing
+
+When the user **explicitly** sets data on the scope (user, request, response, tags, contexts, etc.) or on a span, log, or other telemetry, that data is **not** gated by `dataCollection`. It **MUST** always be attached to outgoing telemetry. The same applies to data the user provides via `beforeSend` or event processors.
```
itaybre: Thanks for the clarification 👍

develop-docs/sdk/foundations/client/data-collection/index.mdx
itaybre: This configuration is set in SessionReplay [configuration](https://github.com/getsentry/sentry-docs/blob/46871b782ff6874c2dab3a41c237e47822bc1310/docs/platforms/android/session-replay/configuration.mdx#L26), it may be worth aligning there

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -0,0 +1,327 @@
...
+This tier is **not** gated by the data collection configuration. SDKs **MAY** collect it by default.
+
+#### 2. PII Data
+
+Personally identifiable or user-linked data:
+
+- User identifiers (user ID, username, email)
+- IP address
+- Cookies and headers that identify the user or session
+- HTTP request data (TBD)
```
itaybre, line 64: `h`: What about request paths? 
Some requests may be identifiable, like `/user/USER_ID`
Should we have a denylist/allowlist for url paths?

itaybre (commented):
This make sense, but got some questions here

develop-docs/sdk/foundations/client/data-collection/index.mdx
s1gr1d: Good question 🤔  
What we currently do in JS: When we either know it's a param route, we use the appropriate parametrized route name (e.g. `user/:id`) as the transaction name but the full URL (e.g. `user/123`) is still added in the attributes. @cleptric Any opinions on that?

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -0,0 +1,327 @@
...
+  },
+});
+```
+
+### Migration from `sendDefaultPii`
+
+- **`sendDefaultPii: true`** (legacy) → `dataCollection: { includeUserInfo: true, collect: { aiAgentMessages: false } }`, keep most `collect` defaults
+- **`sendDefaultPii: false`** (legacy) → `dataCollection: { includeUserInfo: false }` (or omit entirely — same as default)
+
+SDKs **SHOULD** document this mapping and **MAY** implement `send_default_pii` as a compatibility shim that sets `includeUserInfo`.
```
alexander-alderman-webb: Thanks for this, global options sound great!
What's the plan to migrate away from integration-level options? Such as `recordInputs` and `recordOutputs` in JavaScript AI integrations.

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -0,0 +1,327 @@
+---
+title: Data Collection
+description: Configuration for what data SDKs collect by default — technical context, PII, and sensitive data.
+spec_id: sdk/foundations/client/data-collection
+spec_version: 1.0.0
```
philippsuess: `m`: I mean, if it's still a draft and we're trying to validate it, I would actually start with a 0.1.0 version or something to clearly communicate that this still might change.

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -0,0 +1,327 @@
...
+#### 1. Technical Context Data
+
+Non-identifying context used for debugging and performance:
+
+- Device and environment context (OS, runtime, non-PII identifiers)
+- Performance and error context (stack frames, breadcrumbs, span metadata)
+- Framework/routing context where it does not contain PII or secrets
+- AI agent messages (input, output, metadata)
+
+This tier is **not** gated by the data collection configuration. SDKs **MAY** collect it by default.
```
philippsuess: `m`: I think you can change this, actually, to a should. I mean, every SDK adds some technical context data.

```suggestion
This tier is **not** gated by the data collection configuration. SDKs **SHOULD** collect it by default.
```

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -0,0 +1,327 @@
...
+- [Client](/sdk/foundations/client/) — client lifecycle and event pipeline
+- [Configuration](/sdk/foundations/client/configuration/) — top-level init options including `send_default_pii` (deprecated in favor of this spec)
+
+---
+
+## Concepts
+
+<SpecSection id="data-tiers" status="draft" since="1.0.0">
+
+### Data Tiers
```
philippsuess: `l`: I was a bit confused at first by the term "data tier" since to me it usually means architecture layers or pricing plans or something.

What about "Data Sensitivity Levels"? It feels more natural here — it immediately tells you the grouping is about privacy risk, which maps directly to the three levels: 
1. technical context (collect freely)
2. PII (opt-in)
3. secrets (always scrubbed).

```suggestion
### Data Sensitivity Levels
```

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -0,0 +1,327 @@
...
+Collected data is grouped into three tiers. SDKs **MUST** treat these tiers consistently when applying defaults and user configuration.
+
+#### 1. Technical Context Data
+
+Non-identifying context used for debugging and performance:
+
+- Device and environment context (OS, runtime, non-PII identifiers)
+- Performance and error context (stack frames, breadcrumbs, span metadata)
+- Framework/routing context where it does not contain PII or secrets
+- AI agent messages (input, output, metadata)
```
philippsuess: `h`: I think AI Agent Messages can actually contain PII. I think that should be under PII. We have that under PII in the user-facing docs, see for example [Python](https://docs.sentry.io/platforms/python/data-management/data-collected/#llm-inputs-and-responses).

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -0,0 +1,327 @@
...
+
+<SpecSection id="data-tiers" status="draft" since="1.0.0">
+
+### Data Tiers
+
+Collected data is grouped into three tiers. SDKs **MUST** treat these tiers consistently when applying defaults and user configuration.
+
+#### 1. Technical Context Data
+
+Non-identifying context used for debugging and performance:
```
philippsuess: `m`: I think it would be great to clearly state that these here are just examples, but not a complete list. I think it would be great to add links to references where you can find a complete list, or we add a complete list here and then link from the other places in the doc that this is now here, the spec for the complete list. This applies to all three data tiers.

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -0,0 +1,327 @@
...
+<SpecSection id="configuration-surface" status="draft" since="1.0.0">
+
+### Configuration Requirements
+
+All data-collection options live under a single top-level key: `dataCollection`. SDKs **MUST** support at least `includeUserInfo` and the `collect` object. SDKs **MAY** omit options that do not apply to the platform (e.g. no `outgoingRequestBody` on a browser-only SDK).
+
+`dataCollection` accepts two fields:
+
+- **`includeUserInfo`** — the primary toggle for Personally Identifiable Information (PII). Controls whether user-identity fields are included in automatic collection, and sets the default for PII-heavy `collect` options (such as HTTP request bodies - TBD). Defaults to `false`.
+- **`collect`** — controls which categories of request/response and runtime data are gathered. See [`collect` Option Behavior](#collect-option-behavior) and [How Defaults Cascade](#how-defaults-cascade).
```
philippsuess: `h`: The nested config with `includeUserInfo` cascading into `collect` defaults confuses me a bit. When I look at `cookies: true`, I can't tell what actually gets sent without also checking `includeUserInfo` — the same option behaves differently depending on another flag. For a privacy-sensitive config, I think it's important that users can look at their config and immediately know what's being collected.

I wonder if a flatter approach would work better:

**1. A flat list of explicit options** — each boolean or string list independently controls exactly what gets sent. No flag changes the behavior of other flags.

**2. Presets / factory constructors** that return a complete, resolved config:
```typescript
const config = DataCollection.default();
// { userInfo: false, incomingRequestBody: false,
//   aiAgentMessages: true, stackFrameVariables: true,
//   cookies: ['locale', 'theme'],
//   httpHeaders: ['content-type', 'accept', 'x-request-id'],
//   queryParams: true, ... }

// Tweak individual fields — no surprises
config.userInfo = true;
config.incomingRequestBody = true;
config.httpHeaders.push('x-custom-trace');

init({ dsn: "...", dataCollection: config });

// Or start from a PII-inclusive preset
init({ dsn: "...", dataCollection: DataCollection.withPii() });
```

Why I think this could be better:

- **Transparency:** Each option means exactly one thing. No implicit cascades to reason about.
- **Debuggability:** SDK can log the resolved config at init. Support can ask "what's your config?" and get an unambiguous answer.
- **Simpler implementation:** SDK devs check one value per data type — no "if `includeUserInfo` is false, then the effective default for `incomingRequestBody` is false, unless explicitly overridden" logic.
- **Presets handle the common cases:** Most users want "default" or "with PII" — the factory gives them that in one call, and they can still tweak individual fields.

The tradeoff is slightly more verbose config for custom setups, but I think the clarity is worth it — especially when "I can't easily tell what's being sent" is a real problem.

philipphofmann (commented):
Thanks a lot for tackling this, @s1gr1d. I have one major, high-level, general concept question we should address before moving forward (https://github.com/getsentry/sentry-docs/pull/16796#discussion_r2923042918).

develop-docs/sdk/foundations/client/data-collection/index.mdx
cleptric: No, we explicitly want to emit those by default, else our product becomes useless. Also, this is not Pii, this is maybe Pii, and everything is maybe Pii.

develop-docs/sdk/foundations/client/data-collection/index.mdx
cleptric: What is missing?

develop-docs/sdk/foundations/client/data-collection/index.mdx
philippsuess: Okay, I get that. I'm fine with that, but I think we also need to align our docs because that confused me a bit that it was under PII for Python, for example. But that is out of scope of this PR.

develop-docs/sdk/foundations/client/data-collection/index.mdx
philippsuess: The event context payload has way more properties: https://develop.sentry.dev/sdk/foundations/transport/event-payloads/contexts/

For example: culture context, GPU context, app context (version, permissions, view names), cloud resource context, memory info context, ...

develop-docs/sdk/foundations/client/data-collection/index.mdx
s1gr1d: As talked offline, this is a good suggestion that makes things more clear. I agree, the approach with "silently" overwriting the collection options can be confusing. I'll try to incorporate this into the spec 👍

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -0,0 +1,327 @@
+---
+title: Data Collection
+description: Configuration for what data SDKs collect by default — technical context, PII, and sensitive data.
```
coolguyzone:
```suggestion
description: Configuration for what data SDKs collect by default, including technical context, PII, and sensitive data.
```

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -0,0 +1,327 @@
...
+- HTTP request data (TBD)
+
+This tier **MUST** be off by default unless the user opts in via `includeUserInfo` and/or explicit `collect` allowlists. See [`includeUserInfo`](#include-user-info-behavior), [`collect` options](#collect-option-behavior), and [Default Denylist](#default-denylist).
+
+#### 3. Sensitive Data
+
+Credentials and secrets that **MUST** never be sent by default:
+
+- Passwords, tokens, API keys, bearer tokens
+- Header or cookie values that match known sensitive names (auth, token, secret, password, key, jwt, etc.)
```
coolguyzone:
```suggestion
- Header or cookie values that match known sensitive names (auth, token, secret, password, key, etc.)
```

Is jwt redundant here since it's a type of token?

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -0,0 +1,327 @@
...
+
+When user data is set **explicitly** on the scope (or equivalent), it is **always** attached regardless of this setting. See [User-Set Data and Scrubbing](#user-set-data-and-scrubbing).
+
+</SpecSection>
+
+<SpecSection id="collect-options" status="draft" since="1.0.0">
+
+### `collect` Option Behavior
+
+Each key under `collect` maps to a category of automatically collected data and uses one of two option types, depending on whether the data is structured as key-value pairs.
```
coolguyzone:
```suggestion
Each key under `collect` maps to a category of automatically collected data types and uses one of two option types, depending on whether the data is structured as key-value pairs.
```

coolguyzone (approved):
LGTM!

develop-docs/sdk/foundations/data-scrubbing.mdx
```diff
@@ -3,63 +3,21 @@ title: Data Scrubbing
...
-- `http.request.header.cookie: "[Filtered]"` (Used as a fallback if the cookie header cannot be parsed)
+- SDKs should not include PII or other sensitive data in the payload by default. The legacy option [_send-default-pii_](https://docs.sentry.io/platforms/python/configuration/options/#send-default-pii) is **disabled by default**; the replacement is `dataCollection.includeUserInfo` and `dataCollection.collect` (see [Data Collection](/sdk/foundations/client/data-collection/)).
+- Certain sensitive data must never be sent through SDK instrumentation: header/cookie/query values matching the default denylist are replaced with `"[Filtered]"`. User-set data is always attached; only automatically gathered data is scrubbed. Users can use `beforeSend` / event processors to remove or redact any data.
+- For the exact default denylist (partial, case-insensitive match), PII denylist (`x-forwarded-`, `-user`), cookies when unparseable, and raw request bodies, see [Data Collection — Default Denylist](/sdk/foundations/client/data-collection/#default-denylist) and [User-Set Data and Scrubbing](/sdk/foundations/client/data-collection/#user-set-data-scrubbing).
```
coolguyzone:
```suggestion
- For the exact default denylist (partial, case-insensitive match), PII denylist (`x-forwarded-`, `-user`), cookies when unparsable, and raw request bodies, see [Data Collection — Default Denylist](/sdk/foundations/client/data-collection/#default-denylist) and [User-Set Data and Scrubbing](/sdk/foundations/client/data-collection/#user-set-data-scrubbing).
```

This should fix the failing spell-check build.

develop-docs/sdk/foundations/client/data-collection/index.mdx
s1gr1d: I would keep it explicitly in this list. Auth could also be considered a token 🤔

develop-docs/sdk/foundations/client/data-collection/index.mdx
s1gr1d: I initially had the idea of using `presets` that can be overwritten. It's the same concept but still a bit clumsy. Could look like this:
```ts
init({
  dataCollection: {
    preset: "default", // or "withPii"
    overrides: {
      userInfo: true,
      incomingRequestBody: true,
    },
  },
});
```

The approach you mentioned (have pre-configured objects) is very straight-forward and easily testable. Users would also be able to re-use the pre-configured options and e.g. filter on them.

Should also be possible with other languages Examples:
- JS/TS: static factory funcs returning plain objects
- Java/C#: DataCollectionConfig.defaultConfig() + builder mutation/copy
- Python: dataclass + classmethods
- Ruby/PHP: hash/array factories

develop-docs/sdk/foundations/client/data-collection/index.mdx
s1gr1d: Thanks, as `x-real-ip` and `remote-addr` are not considered "sensitive" but PII, I added those to the list of user-sensitive header snippets 👍

develop-docs/sdk/foundations/client/data-collection/index.mdx
s1gr1d: I included them, but made the distinction that `outgoing` bodies are only applicable in a server-environment.

develop-docs/sdk/foundations/client/data-collection/index.mdx
s1gr1d: Right now, I specified one option `aiAgentMessages` for all messages. Do you think it would make sense to split this into `aiAgentInputMessages` and `aiAgendOutputMessages`?

alexander-alderman-webb: to summarize in-person discussion:
- Once the spec is merged and implemented the integration level options are deprecated, and then removed in the next major of the respective SDKs. For finer controls users must use hooks. cc @nicohrubec 
- According to the LLM monitoring RFC we will have two options. See [this](https://github.com/getsentry/rfcs/pull/153/changes/d330c77385c6e055bf27e7ae37912242a13e2084) commit. They are called `record_inputs` and `record_outputs`. Translating to global options (and camel case), I propose `recordGenerativeAIInputs` and `recordGenerativeAIOutputs`.

s1gr1d: Thanks for the suggestion. I'll call them `generativeAIInputs` and `generativeAIOutputs` - the `record` is a bit repetitive as all the options are about "recoding" something.

s1gr1d:
I addressed the feedback, renamed some options and restructured the document a bit. I put a summary of the changes in the description.

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -0,0 +1,466 @@
...
+
+</SpecSection>
+
+---
+
+## Public API
+
+<SpecSection id="option-types" status="draft" since="0.1.0">
+
+### Option Types
```
giortzisg, line 216: I feel that the Option Types are heavily optimized for Javascript. For statically typed languages this is really hard to implement. I would consider for this section to be kept as a semantic model for how the option types work and leave the implementation details to each SDK.

giortzisg (commented):
Thanks for working on this @s1gr1d. The spec makes a lot of sense. Just left a high-level comment.

develop-docs/sdk/foundations/client/data-collection/index.mdx
s1gr1d: What are the options that would be hard to implement in Go for example?

giortzisg: Union and nullable types in general. In Go we need something like this:
```go
KeyValueCollectionPolicy {
  mode: Off | DenyList | AllowList
  terms: string[]
}
```

giortzisg (approved):
Thanks for the changes, LGTM 🚀

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -0,0 +1,599 @@
...
+  dataCollection: {
+    userInfo?: boolean,                            // default: false
+    cookies?: KeyValueCollectionBehavior,          // default: { mode: "denyList" }
+    httpHeaders?: {
+      request?: KeyValueCollectionBehavior,        // default: { mode: "denyList" }
+      response?: KeyValueCollectionBehavior,       // default: { mode: "denyList" }
+    },
+    queryParams?: KeyValueCollectionBehavior,      // default: { mode: "denyList" }
+    generativeAIInputs?: boolean,                  // default: true
+    generativeAIOutputs?: boolean,                 // default: true
```
cleptric:
```suggestion
    genAIInputs?: boolean,                  // default: true
    genAIOutputs?: boolean,                 // default: true
```

cleptric (approved):
Thanks for working on this!

develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -467,11 +467,11 @@ init({
       request?: KeyValueCollectionBehavior,        // default: { mode: "denyList" }
       response?: KeyValueCollectionBehavior,       // default: { mode: "denyList" }
     },
+    httpBodies?: string[],                         // default: [] (off)
```
cleptric, line 470: Hmm, what about GraphQL or gRPC? 😬

PR #16797, opened 2026-03-05
feat(logs,metrics,spans): Add `settings` object spec for envelope item container

Lms24:
This PR adds a specification for ingestion settings of what data Relay should infer from incoming requests. This object shall be the same across our new telemetry items (logs, metrics and spans (v2)) and it lives on the container object:

```json
{
  "version": 2,
  "items": [...],
  "ingest_settings": {
    "infer_ip:": "auto",
    "infer_useragent:": "auto",
  }
}
```

This shall serve as a replacement for the `settings` property on the `sdk` object on events as [defined here](https://develop.sentry.dev/sdk/foundations/transport/event-payloads/sdk/).
I went with the same `'auto'` and `'never'` values but omitted `'legacy'` since it doesn't seem like we need it. If reviewers prefer, we can also use a `boolean` instead, assuming we never need more than two states.

For backwards compatibility, this also introduces a `version` field which must be set to at least `2` for `ingest_settings` to take effect.

Prior to this PR, @cleptric @Dav1dde and I discussed this format. Happy to adjust as reviewers see fit!

ref https://linear.app/getsentry/issue/FE-714/how-to-send-sdksettings-infer-ip

### DIFF
--- a/develop-docs/sdk/telemetry/logs.mdx
+++ b/develop-docs/sdk/telemetry/logs.mdx
@@ -361,7 +361,7 @@ An envelope **MUST** contain at most one `log` envelope item — all log entries
   "items": [{..log..}, {..log..}, {..log..}, {..log..}, {..log..}],
   "settings": {
     "infer_ip:": "auto",
-    "infer_useragent:": "auto",
+    "infer_useragent:": "auto"
   }
 }
 ```
--- a/develop-docs/sdk/telemetry/logs.mdx
+++ b/develop-docs/sdk/telemetry/logs.mdx
@@ -10,6 +10,9 @@ spec_depends_on:
   - id: sdk/foundations/state-management/scopes/attributes
     version: ">=1.0.0"
 spec_changelog:
+  - version: 1.17.0
+    date: 2026-03-05
+    summary: Added `settings` to log envelope item
   - version: 1.16.0
     date: 2026-03-04
     summary: Add sentry.timestamp.sequence attribute for deterministic log ordering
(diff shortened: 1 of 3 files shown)


develop-docs/sdk/telemetry/spans/span-protocol.mdx
```diff
@@ -42,6 +42,35 @@ The envelope item header must contain the following properties:
...
+{
+  "items": [{..span..}, {..span..}, {..span..}],
+  "settings": {
+    "infer_ip:": "auto",
+    "infer_useragent:": "auto",
+  }
+}
+```
+
+<SpecSection id="span-envelope-item-settings" status="proposal" since="0.0.0">
```
Lms24: spans docs are not yet converted to the changelog-based SpecSection format. Hence for now I added the component but set version 0.0.0.

cleptric (approved):

Dav1dde (commented):
Looks good to me, but maybe we wanna bike-shed some more on the namings.

We also need to figure out how this works for logs and the existing extractions, I want to make sure we don't build more special cases.

It also has potential billing impact for logs, as these are billed by size (before extraction).

develop-docs/sdk/telemetry/logs.mdx
```diff
@@ -355,12 +358,31 @@ An envelope **MUST** contain at most one `log` envelope item — all log entries
...
+<SpecSection id="log-envelope-item-settings" status="proposal" since="1.17.0">
+
+### Envelope Item `settings`
+
+The `log` envelope item `settings` property is an optional JSON object that contains ingestion settings for Relay to infer certain data from the incoming request.
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `infer_ip` | String | **OPTIONAL** | The setting to infer the IP address from the incoming request. One of `auto` or `never` (default). |
+| `infer_useragent` | String | **OPTIONAL** | The setting to infer the user agent from the incoming request. One of `auto` or `never` (default). |
```
jjbayer, line 381: Discussed with @Dav1dde : We should introduce a top level `version` field in the item JSON that informs relay what to do with logs:

- `version: 1` is the default. For logs, Relay automatically assumes `infer_ip: auto` (legacy behavior).
- New SDK versions set `version: 2`. Relay assumes `infer_ip: never` as the default.

Without the version field, we'd have to make `infer_ip: auto` the default, which would force new SDKs to explicitly send `infer_ip: never` to disable it.

Lms24: @jjbayer to confirm, you'd envision this? 

```json
{
  "version": 2,
  "items": [...],
  "settings": {
    "infer_ip:": "auto",
    "infer_useragent:": "auto",
  }
}
```

Happy to add it!

Lms24:
@Dav1dde 

> but maybe we wanna bike-shed some more on the namings.

sure, anything that comes to mind? I think `ingest_settings` was on the table yesterday, which is more explicit and communicates intent more clearly. Happy to change it to that. 

Also,

> I went with the same 'auto' and 'never' values but omitted 'legacy' since it doesn't seem like we need it. If reviewers prefer, we can also use a boolean instead, assuming we never need more than two states.

we can "do better" with this second attempt, though tbh this is LOGAF-L for me (bundle size differences are negligible).

develop-docs/sdk/telemetry/logs.mdx
jjbayer: Yes please!

coolguyzone (approved):
Looks good! Thanks for updating

Lms24:
@cleptric @jjbayer apologies for the delay! I made the following changes as discussed:
- added `version` property and specified that SDKs must set it to 2 or higher for Relay to apply the settings
- renamed `settings` -> `ingest_settings` to more clearly communicate the intent

Happy to make more adjustments if necessary!

cleptric (approved):

jjbayer (approved):

PR #16802, opened 2026-03-05
fix(theme-toggle): show moon in light mode, sun in dark mode

HazAT:
The icon should represent the action (what you switch TO), not the current state. In light mode, show the moon to switch to dark. In dark mode, show the sun to switch to light.


cleptric (approved):

HazAT:
For the record: We shipped this during happy hour, having a beer, @cleptric starting Claude from his phone in the cloud, and me on the laptop telling my clanker to find the branch and open a PR.
We truely live in crazy times.

PR #16597, opened 2026-02-26
docs: Make commit and PR attribution agent-agnostic

armenzg:
Update AGENTS.md so AI agents are instructed to use whatever Co-Authored-By line their tool provides (e.g. Cursor, Claude Code) instead of mandating Claude attribution. This avoids incorrect attribution when using other AI coding assistants.

Made with [Cursor](https://cursor.com)

PR #16803, opened 2026-03-05
meta(agents): Remove Claude co-author guidance

szokeasaurusrex:
## DESCRIBE YOUR PR

These lines cause AI agents, other than Claude Code, to incorrectly attribute Claude in commit messages.

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -74,6 +74,6 @@ When writing requirements in `develop-docs/`:
 
 ## Commit and Pull Request generation 
 
-When committing changes, use the Sentry commit skill when available.
+When committing changes, use the Sentry commit skill if available.
 
-For PRs, use .github/PULL_REQUEST_TEMPLATE.md. Also, use the Sentry pull request skill when available. 
+For PRs, use .github/PULL_REQUEST_TEMPLATE.md. Also, use the Sentry pull request skill if available. 


AGENTS.md
```diff
@@ -4,14 +4,6 @@
 
 Use **pnpm**: `pnpm install`, `pnpm dev`, `pnpm build`, `pnpm test`
 
-## Commit Attribution
```
stephanie-anderson, line 7: There is an explicit guideline to add this to every commit message, see:
https://github.com/getsentry/skills/blob/main/plugins/sentry-skills/skills/commit/SKILL.md#ai-generated-changes

We should simply improve it, so the note is not hardcoded to Claude

szokeasaurusrex: Not all AI agents have a specific commit email address, e.g. Codex lacks one as far as I am aware. I guess this suggestion would work for Cursor, though. 

Maybe we can register a bot account to indicate commits generated by other AI agents? Something like `ai-agent-<xyz>@noreply.sentry.io`?

sergical:
There's already https://github.com/getsentry/sentry-docs/pull/16597 @armenzg are you okay with merging that?

szokeasaurusrex:
@sergical I commented on #16597

My main complaint with that PR is it's adding a lot of text, which means the AGENTS.md will consume more context. And, this info is potentially redundant with the commit skill, as I discussed with @stephanie-anderson in the [inline comment above](https://github.com/getsentry/sentry-docs/pull/16803#discussion_r2891199170)

AGENTS.md
szokeasaurusrex: Also, more fundamentally @stephanie-anderson -- if we already have this line in the skill (as you noted), then why should it also be included in `AGENTS.md`?

I think it is actually a bit redundant, and this guidance should just belong in the skill, and not be repeated here. Repeating here also means additional tokens/context are being used. 

I don't feel super strongly about this, just want to question fundamental assumptions 😄

stephanie-anderson: Yeah, I think the AGENTS.md should hint to use the commit skill instead

sergical:
@szokeasaurusrex i agree that we want to minimize the default context loaded in every agent session, but removing all of attribution guidelines breaks it for Cursor, and Claude Code which are the defacto supported agents at Sentry. 

Maybe the silver line is just having the 3 lines that are in https://github.com/getsentry/sentry-for-ai/blob/main/AGENTS.md#commit-attribution

AGENTS.md
```diff
@@ -80,6 +72,8 @@ When writing requirements in `develop-docs/`:
 - Make the plan extremely concise. Sacrifice grammar for the sake of concision.
 - At the end of each plan, give me a list of unresolved questions to answer, if any.
 
-## Pull Request generation 
+## Commit and Pull Request generation
```
szokeasaurusrex, line 75: @stephanie-anderson does this look good?

szokeasaurusrex:
@sergical I have updated this PR to reference using the Sentry commit and PR skills which already contain the attribution line guidelines -- as folks should already be using that, I think this may be sufficient?

szokeasaurusrex:
@sergical see my comment on that PR, but in general, I think that's a sensible direction 👍

PR #16818, opened 2026-03-06
docs(nextjs): Add workaround for multiple transactions inside after

### LINKED ISSUE #16815: Document `after()` trace context behavior for Next.js
### SDK

JavaScript SDK

### Description

Add documentation explaining how to properly instrument deferred work scheduled with Next.js's `after()` function to maintain trace context.

### Background

When using `after()` from `next/server`, spans created inside the deferred callback become detached root transactions instead of children of the request span. This happens because:

1. The SDK flushes the request transaction when the response completes
2. Deferred work in `after()` continues running afterward
3. Spans created during deferred work have no in-memory parent to attach to

This results in dozens of small "db" or other transactions instead of one cohesive trace. See [getsentry/sentry-javascript#19529](https://github.com/getsentry/sentry-javascript/issues/19529) for the original issue and investigation.

### Suggested Solution

Users should wrap their deferred work with `startSpan` **inside** the `after()` callback:

```js
// ✓ Correct
after(
  Sentry.startSpan({ name: "background.task", op: "task" }, async () => {
    await doWork();
  })
);

// ✗ Wrong - task starts before after() runs
const task = doWork();
after(task);
```

JPeer264:
## DESCRIBE YOUR PR

closes #16815
closes [DOCS-2548](https://linear.app/getsentry/issue/DOCS-2548/document-after-trace-context-behavior-for-nextjs)

I wasn't sure where to add this, but decided to move it into the Next.js specific troubleshootings inside the tracing section. I was also not sure how technical it should be, but was then going into a mix of being technical enough to use "transaction", but not going into much detail why this happens exactly.	

Would it make sense to link to the original discussion: https://github.com/getsentry/sentry-javascript/issues/19529#issuecomment-3973460550 ?

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

## LEGAL BOILERPLATE

## EXTRA RESOURCES

### DIFF
--- a/docs/platforms/javascript/guides/nextjs/tracing/index.mdx
+++ b/docs/platforms/javascript/guides/nextjs/tracing/index.mdx
@@ -375,7 +375,7 @@ Sentry.init({
 
 ### Detached spans with `after()`
 
-If you're using Next.js's [`after()`](https://nextjs.org/docs/app/api-reference/functions/after) and seeing many small transactions (like individual `db` spans) instead of one cohesive trace, your spans are becoming detached.
+If you're using Next.js's [`after()`](https://nextjs.org/docs/app/api-reference/functions/after) and you see multiple root spans (like individual `db` spans) instead of one cohesive trace, your spans are becoming detached.
 
 This happens when `after()` runs after the response is sent. By then, the request's root span has already ended.
 


docs/platforms/javascript/guides/nextjs/tracing/index.mdx
```diff
@@ -367,6 +367,41 @@ Sentry.init({
...
 
+</SplitSectionCode>
+</SplitSection>
+
+<SplitSection>
+<SplitSectionText>
+
+### Detached spans with `after()`
+
+If you're using Next.js's [`after()`](https://nextjs.org/docs/app/api-reference/functions/after) and seeing many small transactions (like individual `db` spans) instead of one cohesive trace, your spans are becoming detached.
```
chargome: The term transaction might be confusing?

JPeer264: Do you have any alternatives in mind? Here I was really in between of being non-technical and staying technical. As the transactions are actually the main issue here it is tough to have a different term for it.

docs/platforms/javascript/guides/nextjs/tracing/index.mdx
```diff
@@ -367,6 +367,41 @@ Sentry.init({
...
+
+This happens because `after()` runs after the response is sent. By then, the request's root span has already ended.
```
coolguyzone:
```suggestion
This happens when `after()` runs after the response is sent. By then, the request's root span has already ended.
```

docs/platforms/javascript/guides/nextjs/tracing/index.mdx
Lms24: yeah we should try to phrase this without "transaction". Maybe something like 

```suggestion
If you're using Next.js's [`after()`](https://nextjs.org/docs/app/api-reference/functions/after) and you see multiple root spans (like individual `db` spans) instead of one cohesive trace, your spans are becoming detached.
```

docs/platforms/javascript/guides/nextjs/tracing/index.mdx
```diff
@@ -367,6 +367,41 @@ Sentry.init({
...
+import * as Sentry from "@sentry/nextjs";
+
+export async function POST(request: Request) {
+  const data = await request.json();
+
+  after(
+    Sentry.startSpan(
+      { name: "webhook.process", op: "task" },
+      () => processWebhook(data)
+    )
```
Lms24, line 398: I guess seer is flagging this because it doesn't know that `processWebHook` returns a promise? Wondering if we could somehow make this clearer but not a big deal overall. Feel free to disregard

JPeer264: Actually Seer is not far off because `after` also takes a callback. It is just `waitUntil` that only takes a promise. I think this needs to be added in the docs as well. I'll move it in a quick follow up PR, so this is getting out and people know.

PR #17014, opened 2026-03-18
docs(sdk): Rewrite API and Architecture spec for readability

stephanie-anderson:
Rewrites the API and Architecture standards page to be more human-friendly:

- Add intro paragraph framing the core philosophy
- Replace rigid Rule/Enforcement/Per-SDK override template with natural prose
- Merge breaking changes and deprecation lifecycle into one cohesive section
- Remove enforcement boilerplate and StatusBadge components
- Add contextual "why" before rules instead of dropping straight into requirements
- Thin out RFC keyword density — keep MUST/REQUIRES for hard rules, plain language elsewhere
- Promote spec status to stable

### DIFF
--- a/docs/platforms/android/user-feedback/configuration/index.mdx
+++ b/docs/platforms/android/user-feedback/configuration/index.mdx
@@ -99,6 +99,7 @@ You can customize which form elements are shown, whether they are required, and 
 | `showName`                    | `Bool`   | `true`                                   | Displays the name field on the feedback form. Ignored if `isNameRequired` is `true`.                                      |
 | `isEmailRequired`             | `Bool`   | `false`                                  | Requires the email field on the feedback form to be filled in.                                                            |
 | `showEmail`                   | `Bool`   | `true`                                   | Displays the email field on the feedback form. Ignored if `isEmailRequired` is `true`.                                    |
+| `useShakeGesture`             | `Bool`   | `false`                                  | Opens the feedback form when the user shakes the device. Uses the accelerometer and requires no additional permissions.    |
 | `useSentryUser`               | `Bool`   | `true`                                   | Sets the `email` and `name` fields to the corresponding Sentry SDK user fields that were called with `SentrySDK.setUser`. |
 
 Example:
@@ -111,6 +112,7 @@ Example:
     <meta-data android:name="io.sentry.feedback.show-email" android:value="false" />
     <meta-data android:name="io.sentry.feedback.use-sentry-user" android:value="false" />
     <meta-data android:name="io.sentry.feedback.show-branding" android:value="false" />
+    <meta-data android:name="io.sentry.feedback.use-shake-gesture" android:value="true" />
 </application>
 ```
 ```java
@@ -121,6 +123,7 @@ SentryAndroid.init(context, options -> {
     options.getFeedbackOptions().setShowEmail(false);
     options.getFeedbackOptions().setUseSentryUser(false);
     options.getFeedbackOptions().setShowBranding(false);
+    options.getFeedbackOptions().setUseShakeGesture(true);
 });
 ```
 ```kotlin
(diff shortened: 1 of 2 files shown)


develop-docs/sdk/getting-started/standards/api-architecture.mdx
```diff
@@ -16,179 +16,102 @@ sidebar_order: 1
...
+These standards guide how SDKs are designed and evolved. The core philosophy: prefer server-side logic, protect public API stability, and give users smooth upgrade paths. SDKs often stay installed for years after release, so decisions made today have long-lasting consequences.
 
-## Server-side (Relay) vs. SDK-side decision gate
+<SpecSection id="relay-decision-gate" status="stable" since="1.0.0">
 
-#### Rule
+## Server-side (Relay) vs. SDK implementation
 
-New processing or transformation logic **MUST** default to server-side (Relay).
+When deciding where new logic should live, default to server-side (Relay). Processing data in Relay keeps behavior consistent across SDK versions and avoids duplicating logic in clients that may remain deployed indefinitely.
```
dingsdax, line 25: Is Relay our only entry point? If not we should make this a little bit more open imho.

develop-docs/sdk/getting-started/standards/api-architecture.mdx
```diff
@@ -16,179 +16,102 @@ sidebar_order: 1
...
+<SpecSection id="breaking-and-deprecation" status="stable" since="1.0.0">
 
-## Breaking change process
+## Breaking changes and deprecations
 
-#### Rule
+Breaking changes and deprecations are closely linked — every breaking change goes through a deprecation period first. Both follow a lifecycle designed to give users time and guidance to migrate.
 
-Breaking changes **MUST** follow the [breaking changes playbook](/sdk/getting-started/playbooks/sdk-lifecycle/breaking-changes).
+### Deprecation lifecycle
```
dingsdax, line 96: I wonder if we really need this here or just in the playbook.

develop-docs/sdk/getting-started/standards/api-architecture.mdx
stephanie-anderson: It is, yes - we only send envelopes to the /envelope endpoint in relay

develop-docs/sdk/getting-started/standards/api-architecture.mdx
```diff
@@ -16,179 +16,102 @@ sidebar_order: 1
...
 
-## Semantic conventions process
+## Semantic conventions
 
-#### Rule
+New attributes **MUST** be defined in [sentry-conventions](https://github.com/getsentry/sentry-conventions) before SDK implementation. The process:
 
-New attributes **MUST** first be defined in [sentry-conventions](https://github.com/getsentry/sentry-conventions).
-Process:
+1. Open a PR to sentry-conventions
```
dingsdax, line 77: I wonder if we should move this out of standard, it reads like process, maybe extracting into a Sentry conventions playbook? I would add that changes to attributes are considered a breaking change and that conventions should be treated like public API, wdyt?

develop-docs/sdk/getting-started/standards/api-architecture.mdx
```diff
@@ -16,179 +16,102 @@ sidebar_order: 1
...
 
 </SpecSection>
 
 ---
 
-<SpecSection id="breaking-change-process" status="candidate" since="1.0.0">
+<SpecSection id="breaking-and-deprecation" status="stable" since="1.0.0">
 
-## Breaking change process
+## Breaking changes and deprecations
```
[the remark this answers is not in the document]
stephanie-anderson: This isn't a spec that is implemented against. Changing this now before we roll it out is as safe as it gets.

develop-docs/sdk/getting-started/standards/api-architecture.mdx
stephanie-anderson: Do you mean the intro?

stephanie-anderson: I'll do the playbooks review after the standards - so I might do this in a separate PR in a few hours

PR #17024, opened 2026-03-18
docs(fastapi, flask): Remove feature configuration options from integ…

KyleTryon:
Before:
image

After:
image


KyleTryon:
> do we know if we have other places that could be duplicated or this was it? why was it there in the first place,im curious

This appears to be the only 2 places. Looks like it was an AI oopsie from a few months ago.

PR #17146, opened 2026-03-26
feat(ACI): Mention issue stream monitor

ceorourke:
Add a mention of the issue stream detector and what it is for. This will be linked to in the workflow POST and PUT docs.

### DIFF
--- a/docs/product/new-monitors-and-alerts/monitors/index.mdx
+++ b/docs/product/new-monitors-and-alerts/monitors/index.mdx
@@ -8,7 +8,7 @@ sidebar_order: 70
  New Monitors and Alerts is currently in <strong>beta</strong>. Beta features are still a work in progress and may have bugs. We recognize the irony. Help improve this feature by providing feedback on our [GitHub discussion](https://github.com/getsentry/sentry/discussions/101960).
 </Alert>
 
-Sentry's Monitors are used to customize when to turn errors and performance problems into issues. They can be configured using conditional rules to create issues when specific criteria are met. For example, you could monitor when the duration of a transaction takes longer than 3 seconds on a particular browser. Further automate your issue creation experience using Monitors to automatically add assignees and set different priorities based on severity. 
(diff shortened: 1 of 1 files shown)


docs/product/new-monitors-and-alerts/monitors/index.mdx
```diff
@@ -8,7 +8,7 @@ sidebar_order: 70
...
 </Alert>
 
-Sentry's Monitors are used to customize when to turn errors and performance problems into issues. They can be configured using conditional rules to create issues when specific criteria are met. For example, you could monitor when the duration of a transaction takes longer than 3 seconds on a particular browser. Further automate your issue creation experience using Monitors to automatically add assignees and set different priorities based on severity. 
+Sentry's Monitors are used to customize when to turn errors and performance problems into issues. They can be configured using conditional rules to create issues when specific criteria are met. For example, you could monitor when the duration of a transaction takes longer than 3 seconds on a particular browser. Further automate your issue creation experience using Monitors to automatically add assignees and set different priorities based on severity.
```
ceorourke, line 11: All of this was auto-linting 🫤

docs/product/new-monitors-and-alerts/monitors/index.mdx
```diff
@@ -22,8 +22,9 @@ You can use Custom Monitors to track errors based on span attributes and custom
 
 ### Default Monitors
 
-Sentry also provides default monitors that are automatically created for you when you create a new project. 
+Sentry also provides default monitors that are automatically created for you when you create a new project.
 
+- **Issue Stream Monitor**: The default monitor tracking new issues of all types created for a project
```
ceorourke, line 27: This is the only actual change. @saponifi3d suggested potentially calling it the "issue feed monitor" but I checked the [product docs for issues](https://docs.sentry.io/product/issues/#how-to-sort-issues) and it says stream so I am leaning towards sticking with "stream" because API users will need to hit the detector GET endpoint and filter to "Issue stream" so calling it "feed" would necessitate a data migration for parity.

souredoutlook: I'm fine with this - we use stream elsewhere externally AFAICT

saponifi3d: 👍 sounds good to me! i think we should probably come up with a single term for it, but we don't need to do that now / for this. thanks for diving in further to find those docs too!

saponifi3d (approved):

PR #17164, opened 2026-03-28
fix(electron): Detail Electron process differences in SDK options

### LINKED ISSUE #1333: docs: Separate main and renderer options
Currently the docs just display the base JavaScript options. Ideally we should display separate options for the main and renderer process because they differ:

https://docs.sentry.io/platforms/javascript/guides/electron/configuration/options/

timfish:
- Closes https://github.com/getsentry/sentry-electron/issues/1333

I also removed `docs/platforms/javascript/guides/electron/configuration/environments.mdx` because this resulted in an extra menu item that didn't work when clicked and this exact text is already in the [`environment` options section](https://docs.sentry.io/platforms/javascript/guides/electron/configuration/options/#environment).


Lms24 (approved):
thanks!

PR #17196, opened 2026-03-31
docs(cli): Add code mappings documentation page

romtsn:
## Summary

- Add new CLI docs page at `/cli/code-mappings/` covering the `sentry-cli code-mappings upload` command
- Document JSON format, options, batching behavior, and CI/CD integration
- Use real sentry-java monorepo as example (including duplicate stack roots for opentelemetry modules)
- Add backlinks from:
  - Suspect commits page
  - Java source context page
  - GitHub integration page
  - CLI index page

## Test plan

- [ ] Verify page renders correctly at `/cli/code-mappings/`
- [ ] Verify backlinks work from all updated pages


docs/cli/code-mappings.mdx
```diff
@@ -0,0 +1,114 @@
...
+description: "Upload code mappings to Sentry via the CLI. Code mappings link stack trace paths to source code paths in your repository, enabling source context, suspect commits, and code owners."
+---
+
+Code mappings link stack trace paths to source code paths in your repository. They enable features like [source context](/platforms/java/source-context/), [suspect commits](/product/issues/suspect-commits/), [stack trace linking](/organization/integrations/source-code-mgmt/github/#stack-trace-linking), and [code owners](/product/issues/ownership-rules/).
+
+You can manage code mappings through the Sentry web UI (see [Setting Up Code Mappings](/product/issues/suspect-commits/#set-up-code-mappings)), or upload them in bulk using `sentry-cli`.
+
+## Permissions
+
+The `sentry-cli` requires an [Organization Token](https://sentry.io/orgredirect/organizations/:orgslug/settings/auth-tokens/) with the `org:ci` scope to upload code mappings.
```
szokeasaurusrex: Would a user token with `org:write` or `org:admin` also work?
- m: If yes, we should document it.
- l: If not, I would suggest we enable this feature in the backend (as a separate PR of course), then document it.

docs/cli/code-mappings.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+## Upload Code Mappings
+
+<Alert>
+
+You need to specify the organization and project you are working with because code mappings work on projects. For more information about this refer to [Working with Projects](/cli/configuration/#sentry-cli-working-with-projects).
+
+</Alert>
```
szokeasaurusrex: I think we can leave this out. The command should error if any required arguments are missing. 

```suggestion
```

docs/cli/code-mappings.mdx
```diff
@@ -0,0 +1,114 @@
...
+```
+
+### Options
+
+| Option | Required | Description |
+|--------|----------|-------------|
+| `--org` | Required | The organization slug. Can also be set via `SENTRY_ORG` or in `.sentryclirc`. |
+| `--project` | Required | The project slug. Can also be set via `SENTRY_PROJECT` or in `.sentryclirc`. |
+| `--repo` | Optional | The repository name (e.g. `getsentry/sentry`). Defaults to the git remote origin. Required if not running from within a git repository. |
+| `--default-branch` | Optional | The default branch name. Defaults to the git remote HEAD or `main`. Required if not running from within a git repository. |
```
szokeasaurusrex: m: Let's keep the example invocation minimal (the only strictly required argument is the path to the mappings; everything else can be auto-detected or configured via environment variables that folks may already be setting). Also, let's not document the options here, and instead point folks towards `--help`. Documenting here increases risk of divergence in the future.

````suggestion
Use the `sentry-cli code-mappings upload` command to upload code mappings:

```bash
sentry-cli code-mappings upload /path/to/code-mappings.json
```

See `sentry-cli code-mappings upload --help` for all options.
````

File: docs/cli/code-mappings.mdx [hunk 4]
```diff
@@ -0,0 +1,114 @@
...
+
+### JSON File Format
```
szokeasaurusrex, line 26: Is there a standardized way to generate such files? If so, I think we should mention it.

docs/cli/code-mappings.mdx
```diff
@@ -0,0 +1,114 @@
...
+    "stackRoot": "io/sentry",
+    "sourceRoot": "sentry/src/main/java/io/sentry"
+  }
+]
+```
+
+- **stackRoot**: The path prefix as it appears in stack traces (e.g. `io/sentry/android/core` for Java, `src/` for Python/JS).
+- **sourceRoot**: The corresponding path in your repository where the source code lives.
+
+Multiple mappings can share the same `stackRoot` if they point to different `sourceRoot` paths. In the example above, `io/sentry/opentelemetry` maps to two different modules. Sentry evaluates mappings from most specific to least specific, using the first one that resolves to a real file in the repository.
```
szokeasaurusrex: l: I think it makes sense to reorder this. Right now, we start the description, have an example, then continue the description afterwards. I would do the full description first, then the example under a subheading, like so:


````suggestion
The input file should be a JSON array of mapping objects, each with a `stackRoot` and `sourceRoot`:
- **stackRoot**: The path prefix as it appears in stack traces (e.g. `io/sentry/android/core` for Java, `src/` for Python/JS).
- **sourceRoot**: The corresponding path in your repository where the source code lives.

Multiple mappings can share the same `stackRoot` if they point to different `sourceRoot` paths. In the example below, `io/sentry/opentelemetry` maps to two different modules. Sentry evaluates mappings from most specific to least specific, using the first one that resolves to a real file in the repository.

#### Example

This example is based on the [getsentry/sentry-java](https://github.com/getsentry/sentry-java) monorepo, which has multiple modules sharing the `io/sentry` package prefix:

```json {filename:mappings.json}
[
  {
    "stackRoot": "io/sentry/android/core",
    "sourceRoot": "sentry-android-core/src/main/java/io/sentry/android/core"
  },
  {
    "stackRoot": "io/sentry/opentelemetry",
    "sourceRoot": "sentry-opentelemetry/sentry-opentelemetry-core/src/main/java/io/sentry/opentelemetry"
  },
  {
    "stackRoot": "io/sentry/opentelemetry",
    "sourceRoot": "sentry-opentelemetry/sentry-opentelemetry-bootstrap/src/main/java/io/sentry/opentelemetry"
  },
  {
    "stackRoot": "io/sentry",
    "sourceRoot": "sentry/src/main/java/io/sentry"
  }
]
```
````

File: docs/cli/code-mappings.mdx [hunk 6]
```diff
@@ -0,0 +1,114 @@
...
+
+### Batching
+
+Mappings are uploaded in batches of up to 300 per request. If your file contains more than 300 mappings, the CLI will automatically split them into multiple batches.
```
szokeasaurusrex: m: This is an implementation detail, I see no reason to document it.

```suggestion
```

docs/cli/code-mappings.mdx
```diff
@@ -0,0 +1,114 @@
...
+```
+Uploading 47 code mapping(s)...
+Created: 42, Updated: 5, Errors: 0
+```
+
+- **Created**: New mappings that were added.
+- **Updated**: Existing mappings (matched by project, stack root, and source root) whose metadata was updated.
+- **Errors**: Mappings that failed to save.
+
+If any errors occur, the CLI exits with a non-zero exit code and prints a table showing which mappings failed.
```
szokeasaurusrex: l: I would also remove this section; I think the output format is pretty self-explanatory. And, we again risk divergence by including this


```suggestion
```

docs/cli/code-mappings.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+## Stack Trace Root and Source Code Root
```
szokeasaurusrex: m: This sections seems at least partially redundant with the JSON file format description above.

I think we should consider removing this section. If we keep it, we should ensure that no redundant info is included here, which is also included there (or vice-versa)

docs/cli/code-mappings.mdx
```diff
@@ -0,0 +1,114 @@
+---
+title: "Code Mappings"
+sidebar_order: 5
+description: "Upload code mappings to Sentry via the CLI. Code mappings link stack trace paths to source code paths in your repository, enabling source context, suspect commits, and code owners."
+---
+
+Code mappings link stack trace paths to source code paths in your repository. They enable features like [source context](/platforms/java/source-context/), [suspect commits](/product/issues/suspect-commits/), [stack trace linking](/organization/integrations/source-code-mgmt/github/#stack-trace-linking), and [code owners](/product/issues/ownership-rules/).
+
+You can manage code mappings through the Sentry web UI (see [Setting Up Code Mappings](/product/issues/suspect-commits/#set-up-code-mappings)), or upload them in bulk using `sentry-cli`.
```
szokeasaurusrex: m: Let's mention that you can use the CLI command in CI here. I think we can then eliminate the CI section at the bottom. Also, let's use "Sentry CLI" rather than `sentry-cli`.

l: Also, we can integrate the link to the web UI more cleanly in the text.

```suggestion
You can [manage code mappings through the Sentry web UI](/product/issues/suspect-commits/#set-up-code-mappings)), or upload them in bulk using Sentry CLI. It is also possible to upload code mappings automatically in your CI process using Sentry CLI.
```

docs/cli/code-mappings.mdx
szokeasaurusrex: m:

```suggestion
Sentry CLI requires an [Organization Token](https://sentry.io/orgredirect/organizations/:orgslug/settings/auth-tokens/) with the `org:ci` scope to upload code mappings.
```

docs/cli/code-mappings.mdx
```diff
@@ -0,0 +1,114 @@
...
+
+```bash
+sentry-cli code-mappings upload ./mappings.json \
+  --org $SENTRY_ORG \
+  --project $SENTRY_PROJECT \
+  --repo $GITHUB_REPOSITORY \
+  --default-branch $GITHUB_REF_NAME
+```
+
+For authentication, set the `SENTRY_AUTH_TOKEN` environment variable. See [Configuration](/cli/configuration/) for details.
```
szokeasaurusrex: m: This section is unnecessary imo. Instead, we can [mention the idea of using CLI in CI at the top](https://github.com/getsentry/sentry-docs/pull/17196/changes#r3021101060).

```suggestion
```

docs/cli/index.mdx
```diff
@@ -18,6 +18,6 @@ These docs cover `sentry-cli`, used in CI/CD pipelines and build processes. If y
 
 </Alert>
 
-For certain actions, you can use the `sentry-cli` command line executable. It can connect to the Sentry API and manage some data for your projects. It’s primarily used for managing debug information files for iOS, Android as well as release and source maps management for other platforms.
+For certain actions, you can use the `sentry-cli` command line executable. It can connect to the Sentry API and manage some data for your projects. It’s primarily used for managing debug information files for iOS, Android, release and source maps management, as well as [code mappings](/cli/code-mappings/) for other platforms.
```
szokeasaurusrex: l: I think this change is unnecessary. This list is only for some examples of use cases, it is not meant to be exhaustive

```suggestion
For certain actions, you can use the `sentry-cli` command line executable. It can connect to the Sentry API and manage some data for your projects. It’s primarily used for managing debug information files for iOS, Android as well as release and source maps management for other platforms.
```

docs/platforms/java/common/source-context/index.mdx
```diff
@@ -10,7 +10,7 @@ og_image: /og-images/platforms-java-common-source-context.png
 
 ## Setting Up Code Mappings
 
-To use suspect commits and stack trace linking, you'll need to set up a code mapping. This is a mapping between the source code in your repository and the source code in your stack traces. You can find information on how to set up code mappings in our [Set Up Code Mappings](/product/issues/suspect-commits/#2-set-up-code-mappings) docs.
+To use suspect commits and stack trace linking, you'll need to set up a code mapping. This is a mapping between the source code in your repository and the source code in your stack traces. You can find information on how to set up code mappings in our [Set Up Code Mappings](/product/issues/suspect-commits/#2-set-up-code-mappings) docs. For projects with many modules, you can also [upload code mappings in bulk via the CLI](/cli/code-mappings/).
```
szokeasaurusrex: m: The link for "Set Up Code Mappings" is broken (I guess it was already broken before your PR, but it makes sense to fix here or in a separate PR alongside this one)

l: Also, I think it makes more sense to present the options of Sentry UI and CLI as alternatives. The existing page primarily describes the Sentry web UI path

```suggestion
To use suspect commits and stack trace linking, you'll need to set up a code mapping. This is a mapping between the source code in your repository and the source code in your stack traces. You can [set up code mappings manually in the Sentry UI](/product/issues/suspect-commits/#set-up-code-mappings) or [upload code mappings in bulk via Sentry CLI](/cli/code-mappings/). We recommend using the CLI if your project has many modules.
```

szokeasaurusrex (commented):
Left some suggestions. I think we can and should omit several of the items that were added here; please see comments.

docs/cli/code-mappings.mdx
romtsn: Yep, those would work. However, I'm not sure I want to promote user tokens here, since the primary usecase is to call this in Ci -- at least we don't do that in the other sections (Release Manager or Debug Files, which also [allow](https://github.com/getsentry/sentry/blob/21ac14e8bc375f7a71797ca9aa71305585ad53b4/src/sentry/api/bases/organization.py#L140-L146) using user tokens).

romtsn: If we go for that, we should probably align and call this out in the other sections as well

docs/cli/code-mappings.mdx
romtsn: This was also a copy from [Debug Files](https://docs.sentry.io/cli/dif/#uploading-files) and [Release Management](https://docs.sentry.io/cli/releases/), so if you want to drop this we should align all of them, I believe

szokeasaurusrex: Let's drop it here; we can align on the other pages in a separate PR

docs/cli/code-mappings.mdx
szokeasaurusrex: Fair enough, I am okay with leaving this as is, in that case

docs/cli/index.mdx
romtsn: I think I'd keep it for discoverability purpose

szokeasaurusrex: I understand, but it is also awkward that code mappings is the only feature we mention there, which we also link to. So I'd consider removing the link, or adding a link for everything else, to keep it consistent 😄

romtsn: sure, I can drop the link 👍

docs/cli/code-mappings.mdx
romtsn: nope, not really. I was thinking to add a sentry.io/changelog entry where I'd provide an example snippet on how to do this in Gradle, but I could also add it here if you prefer. Although, the next step would be to do this automatically in the Sentry Gradle plugin so this would make the snippet redundant.

docs/platforms/java/common/source-context/index.mdx
romtsn:
```suggestion
To use suspect commits and stack trace linking, you'll need to set up a code mapping. This is a mapping between the source code in your repository and the source code in your stack traces. You can [set up code mappings manually in the Sentry UI](/product/issues/suspect-commits/#set-up-code-mappings) or [upload code mappings in bulk via Sentry CLI](/cli/code-mappings/). We recommend using the CLI if your project has many modules.
```

docs/cli/code-mappings.mdx
```diff
@@ -0,0 +1,58 @@
+---
+title: "Code Mappings"
+sidebar_order: 5
+description: "Upload code mappings to Sentry via the CLI. Code mappings link stack trace paths to source code paths in your repository, enabling source context, suspect commits, and code owners."
+---
+
+Code mappings link stack trace paths to source code paths in your repository. They enable features like [source context](/platforms/java/source-context/), [suspect commits](/product/issues/suspect-commits/), [stack trace linking](/organization/integrations/source-code-mgmt/github/#stack-trace-linking), and [code owners](/product/issues/ownership-rules/).
```
romtsn, line 8:
```suggestion
_Available in version 3.3.4 of Sentry CLI_

Code mappings link stack trace paths to source code paths in your repository. They enable features like [source context](/platforms/java/source-context/), [suspect commits](/product/issues/suspect-commits/), [stack trace linking](/organization/integrations/source-code-mgmt/github/#stack-trace-linking), and [code owners](/product/issues/ownership-rules/).
```
@szokeasaurusrex not sure if we should mention the min supported version here?

coolguyzone (approved):
LGTM

docs/cli/code-mappings.mdx
szokeasaurusrex: Sure, sounds good!

szokeasaurusrex (approved):
lgtm

PR #17228, opened 2026-04-02
Add setup step for Azure SSO

rodolfoBee:
Add a configuration step to ensure the Unique User Identifier format is set to unspecified.

Sentry sets the format as unspecified in the SAMLRequest and some users have reported that this caused issues while configuring SSO.

Adding this step to ensure the nameID format is set to unspecified to avoid any related setup issues.


InterstellarStella (approved):
LGTM

PR #17234, opened 2026-04-02
docs: Update logic placement guidelines in API architecture

szokeasaurusrex:
This section was using RFC 2119 requirement levels in a strange way. These levels are meant to designate requirement levels, but here, they were using to describe technical details of data processing features. 

My change removes the strangely-used "**MUST**" specifiers from the examples of changes that need to live in the SDK, instead using generic "needs to" phrasing


coolguyzone (approved):
Good catch 👍

PR #17320, opened 2026-04-13
docs(react-native): Update useDispatchedActionData description

alwx:
Update the React Navigation integration docs to reflect the new behavior of `useDispatchedActionData`:

- **Immediate span naming** from action payload (`NAVIGATE`, `PUSH`, `REPLACE`, `JUMP_TO`)
- **Skip spans** for actions without a known route name (`GO_BACK`, `POP`, `POP_TO_TOP`, `RESET`)
- **Filter non-navigation actions** (`SET_PARAMS`, drawer actions)
- Clarified that `goBack` transactions are skipped when the option is enabled

Ref: getsentry/sentry-react-native#5982


antonis (approved):
LGTM

PR #17374, opened 2026-04-16
chore(code-mappings): Update integration settings instructions for clarity

romtsn:
Make it clearer which integration do we mean here


coolguyzone (approved):
Thanks for clarifying!

PR #17588, opened 2026-05-02
fix(mobile): Lock body scroll when sidebar opens on mobile

rahulchhabria:
On iOS Safari, the mobile sidebar overlay wasn't scrollable because
touch events were being captured by the scrollable body underneath.
This adds body scroll locking when the sidebar opens on mobile, matching
the existing behavior for the mobile search overlay.

Slack thread: https://sentry.slack.com/archives/C8H0BQRDJ/p1777747105228359

https://claude.ai/code/session_01XpiMadpiudFwxMEpwPDpPE

### DIFF
--- a/src/components/header.tsx
+++ b/src/components/header.tsx
@@ -124,6 +124,27 @@ export function Header({
     return () => checkbox.removeEventListener('change', handleChange);
   }, [isHomePage]);
 
+  // Lock body scroll when sidebar is open on mobile (fixes iOS Safari touch scrolling)
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


sergical (approved):
Gonna unblock this because it solves half the issue, the other half i think is the fact that we have a list of items that are laid out horizontally on desktop and get stacked in the mobile menu but aren't scrollable, only the small part at the bottom becomes the "sidebar" that scrolls, making it a poor experience - i recommend that we collapse all the horizontal menu items into an expandable mobile menu item rather than rendering all of them all the time.

Image

PR #17723, opened 2026-05-11
chore: force vite >=7.3.2 to patch CVE-2026-39364

coolguyzone:
Vite is a transitive dependency (via vitest) so dependabot can't auto-bump it. Adding a pnpm override to resolve to a patched version.

## DESCRIBE YOUR PR

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

## LEGAL BOILERPLATE

## EXTRA RESOURCES

### DIFF
--- a/package.json
+++ b/package.json
@@ -177,7 +177,8 @@
       "rollup": "^4.59.0",
       "fast-xml-parser": "^5.5.7",
       "undici": "^6.24.0",
-      "minimatch": "^9.0.0"
+      "minimatch": "^9.0.0",
+      "vite": ">=7.3.2 <8.0.0"
     },
     "onlyBuiltDependencies": [
       "@parcel/watcher",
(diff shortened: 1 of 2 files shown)


sergical:
is there a version of vitest that makes sure it's vite version is good? i am not sure adding an override is the right approach long term but i could also be wrong

additionally the override is a range, is that right? no other packages do this in the file

coolguyzone:
@sergical to give some more context, I'm trying to address this issue: https://github.com/getsentry/sentry-docs/security/dependabot/288

Talking it through with Claude led me to the override approach, although I think we only use vite for vitest tests, so I don't think the vulnerability is a real issue for us, if you think this could introduce changes we could also table it.

sergical (approved):
i think in terms of best practices it makes more sense to pin vite to a specific version that we know is patched vs this range of versions, otherwise LGTM (plus a conflict resolution now)

PR #17752, opened 2026-05-13
feat(node): Document SEA setup

logaretm:
A user asked how to get their instrumentations working in a [Node.js Single Applications](https://nodejs.org/api/single-executable-applications.html) setup, I tried a few things and settled on this as it seems to work best and simplest way to set it up.

Not sure if needs its own dedicated page, but the user did follow the trail of not being able to use `--import`, so it might be an appropriate place for it.

### DIFF
--- a/docs/platforms/javascript/common/install/esm-without-import.mdx
+++ b/docs/platforms/javascript/common/install/esm-without-import.mdx
@@ -18,7 +18,7 @@ supported:
 </Alert>
 
 
-When running your application in ESM mode, you will most likely want to <PlatformLink to="/install/esm">follow the ESM instructions</PlatformLink>. However, if you can't use the `--import` command line option, you can either use [direct imports](#direct-import-setup) or [SEA bootstrap setup](#nodejs-single-executable-applications) if you are using a Node.js Single Executable Application (SEA).
+When running your application in ESM mode, you will most likely want to <PlatformLink to="/install/esm">follow the ESM instructions</PlatformLink>. However, if you can't use the `--import` command line option, you can either use [direct imports](#direct-imports) or [SEA bootstrap setup](#nodejs-single-executable-applications) if you are using a Node.js Single Executable Application (SEA).
 
 ## Direct Imports
 
--- a/docs/platforms/javascript/common/install/esm-without-import.mdx
+++ b/docs/platforms/javascript/common/install/esm-without-import.mdx
@@ -17,7 +17,11 @@ supported:
   [installation methods](../).
 </Alert>
 
-When running your application in ESM mode, you will most likely want to <PlatformLink to="/install/esm">follow the ESM instructions</PlatformLink>. However, if you want to avoid using the `--import` command line option, for example if you have no way of configuring a CLI flag, you can also follow an alternative setup that involves importing the `instrument.mjs` file directly in your application.
+
+When running your application in ESM mode, you will most likely want to <PlatformLink to="/install/esm">follow the ESM instructions</PlatformLink>. However, if you can't use the `--import` command line option, you can either use [direct imports](#direct-import-setup) or [SEA bootstrap setup](#nodejs-single-executable-applications) if you are using a Node.js Single Executable Application (SEA).
+
+## Direct Imports
+
 
 <Alert level='warning' title='Restrictions of this installation method'>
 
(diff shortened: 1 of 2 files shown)


docs/platforms/javascript/common/install/esm-without-import.mdx
```diff
@@ -55,3 +60,49 @@ import http from "http";
...
+Node.js Single Executable Applications (SEA) may not load your Sentry instrumentation early enough, so you need to package a small bootstrap file as the SEA main instead of packaging your app entrypoint directly.
+
+The bootstrap imports Sentry first, then imports your real app entrypoint:
+
+```javascript {filename: sea-main.cjs}
+(async () => {
+  await import("./instrument.mjs");
+  await import("./app.mjs");
+})();
+```
```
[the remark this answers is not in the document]
logaretm: It only breaks in Node 26, there is a workaround with `createRequire`.

docs/platforms/javascript/common/install/esm-without-import.mdx
```diff
@@ -55,3 +60,68 @@ import http from "http";
...
+  "output": "sea-prep.blob",
+  "disableExperimentalSEAWarning": true,
+  "useSnapshot": false,
+  "useCodeCache": false
+}
+```
+
+Keep `sea-bootstrap.cjs`, `instrument.mjs`, and `app.mjs` available on the
+filesystem next to the executable. If you want a fully self-contained
+executable, bundle your instrumentation and app into the SEA main instead.
```
inventarSarah:
```suggestion
bundle your instrumentation and app into the SEA main instead.
```
It would be nice to make this more actionable -- e.g., link to docs that explain how to do this.

docs/platforms/javascript/common/install/esm-without-import.mdx
```diff
@@ -55,3 +60,68 @@ import http from "http";
...
+```javascript {tabTitle:ESM} {filename: instrument.mjs}
+import * as Sentry from "@sentry/node";
+
+Sentry.init({
+  dsn: "___PUBLIC_DSN___",
+  tracesSampleRate: 1.0,
+});
+```
+
+Then configure SEA to use `sea-main.cjs` as its main script and disable code
```
inventarSarah: q: is it clear why users need to disable code cache or do we need to mention why?

docs/platforms/javascript/common/install/esm-without-import.mdx
inventarSarah, line 25: Then create a new section for the original setup instructions. Add an H2 and name it something like "Static Import" or "Direct Import" (or whatever you want to call it)

docs/platforms/javascript/common/install/esm-without-import.mdx
```diff
@@ -25,7 +25,12 @@ This installation method has the fundamental restriction that only native Node.j
 
 As a result, the Sentry SDK will not capture data from database calls, queues, ORMs, third-party libraries, or other framework-specific data.
 
-We recommend using this only if the `--import` flag is not an option for you.
+We recommend using these setups only if the `--import` flag is not an option for you.
```
inventarSarah: You can remove your changes from this alert because it now lives in the "Static import" section

docs/platforms/javascript/common/install/esm-without-import.mdx
inventarSarah: Since this page now covers 2 options, we need to update this intro.
We could write something like:

If you can't use the `--import` flag, choose the setup that matches your situation: static import for regular ESM apps, or the SEA bootstrap setup if you're building a Single Executable Application.

inventarSarah (commented):
Thank you, the content looks good overall 👍

I left some suggestions and questions for you.

**Regarding Structure:**
I suggest restructuring the "ESM without import" page into two sections (add a new section for the original setup instructions).
That would help users find what they need more quickly and, maybe, make it easier for you to decide where to place which content.

JPeer264 (approved):
amazing thank you

coolguyzone (approved):
Looks good, thanks for updating 🫡

docs/platforms/javascript/common/install/esm-without-import.mdx
logaretm: So when investigating this earlier we did need it to force node to resolve the files at runtime, but looks like we don't anymore with `createRequire` approach I settled on last. I will remove it!

thanks!

docs/platforms/javascript/common/install/esm-without-import.mdx
logaretm: Settled on "Direct imports"

PR #17755, opened 2026-05-13
docs(platforms): Add missing query string clarifications

adinauer:
Add the request query string `sendDefaultPii` clarification to Android, React Native, Rust, and Unity data-collected docs.

The parent PR adds this clarification across other platform pages, but these pages already had Request Query String sections and were left out. This keeps the existing query string documentation consistent across the affected SDKs while using the parent PR as the base.


0xadam-brown (approved):
👍

coolguyzone (approved):
Looks good, thanks for adding these warnings.

PR #17756, opened 2026-05-13
docs(platforms): Generalize query string collection wording

adinauer:
Generalize the Apple and Native request query string descriptions so they do not imply query strings are only sent through tracing span attributes.

The data-collected pages should describe the data category rather than a single transport path. Query strings can also appear through other data paths such as breadcrumbs or event request data, so this switches those sections to the same path-agnostic wording used by the other SDK docs.


philprime (approved):
LGTM

philprime (approved):
LGTM

coolguyzone (approved):
Much simpler, thanks!

PR #17779, opened 2026-05-15
feat(api): Use region-specific URLs in API doc curl examples

mwarkentin:
## DESCRIBE YOUR PR

API doc pages (e.g. `/api/explore/query-explore-events-in-table-format/`) currently render curl examples with `https://sentry.io/api/0/...` as the base URL. This ignores the OpenAPI spec's `servers` field which defines region-specific URLs (`https://{region}.sentry.io`).

This PR:

- **Reads the global `servers` field** from the OpenAPI spec instead of hardcoding `sentry.io`. Per-operation server overrides are still respected first, then the global servers block, then `https://sentry.io` as a safety fallback.
- **Adds a region picker** (US | DE tabs) above the curl code block on every API endpoint page, so users can switch between `https://us.sentry.io/api/0/...` and `https://de.sentry.io/api/0/...`.
- **Auto-detects the region** for logged-in users by reading their project's `API_URL` from `CodeContext`, pre-selecting the correct tab.

This steers customers (especially those on the DE region) toward the correct regional endpoint, avoiding cross-Atlantic RPCs.

Refs INC-2335

## IS YOUR CHANGE URGENT?

- [ ] Urgent deadline (GA date, etc.)
- [ ] Other deadline

## PRE-MERGE CHECKLIST

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/build/open-api/types.ts
+++ b/src/build/open-api/types.ts
@@ -36,12 +36,20 @@ type Tag = {
   'x-sidebar-name': string;
 };
 
-type ServerMeta = {
-  description: string;
+export type ServerVariable = {
+  default: string;
+  description?: string;
+  enum?: string[];
+};
+
+export type ServerMeta = {
+  description?: string;
   url: string;
+  variables?: Record<string, ServerVariable>;
 };
 
 export type DeRefedOpenAPI = {
+  servers?: ServerMeta[];
   paths: {
     [key: string]: {
       [key: string]: {
--- a/src/build/resolveOpenAPI.ts
+++ b/src/build/resolveOpenAPI.ts
@@ -2,7 +2,11 @@
 
 import {promises as fs} from 'fs';
 
-import {type DeRefedOpenAPI, type ServerMeta, type ServerVariable} from './open-api/types';
+import {
+  type DeRefedOpenAPI,
+  type ServerMeta,
+  type ServerVariable,
+} from './open-api/types';
 
 // SENTRY_API_SCHEMA_SHA is used in the sentry-docs GHA workflow in getsentry/sentry-api-schema.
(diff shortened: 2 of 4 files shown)


sergical:
CleanShot 2026-05-19 at 11 11 20@2x

Just thinking out loud, would it make more sense to have it be 2 separate code tabs in the snippet vs additional tabs above? @mwarkentin

mwarkentin:
@sergical do you mean tabs where we have "bash"? Or underneath that?

sergical:
> @sergical do you mean tabs where we have "bash"? Or underneath that?

yeah where it says Bash. I am not sure if we can create those programmatically but _feels_ like something that can be done? 

And then the UX will be more close to for example 
CleanShot 2026-05-19 at 16 19 36@2x

mwarkentin:
@sergical seems like that would get a bit weird - would we have a us and de tab for each language that we support? Seems like a different "level" of hierarchy compared to language.

Edit: Note that this will get worse as new localities are added in the future.

sergical:
> @sergical seems like that would get a bit weird - would we have a us and de tab for each language that we support? Seems like a different "level" of hierarchy compared to language.
> 
> Edit: Note that this will get worse as new localities are added in the future.

I was thinking since it's for api docs, there isn't anything language specific and it's just curl commands with different urls. 

maybe the right approach is on API routes to have a region dropdown where on SDK pages we have the SDK dropdown? 

Sorry, im just trying to figure out a way to have existing patterns serve the need. since most of the time a user looking at docs will only care about 1 config for the region vs each code block example, thoughts?
CleanShot 2026-05-19 at 16 27 48@2x
CleanShot 2026-05-19 at 16 27 44@2x

PR #17782, opened 2026-05-18
docs(self-hosted): introduce troubleshooting snuba page

### LINKED ISSUE #4157: Clarify behavior of Snuba subscription consumers
### Problem Statement

I’m trying to understand the behavior of **Snuba subscription consumers** in a self-hosted Sentry deployment.

While monitoring Kafka consumer groups, the subscription consumers appear to **not advance the current offset**, and in some cases the **end offset is not shown**, even though:
- The application is working correctly
- Events and subscriptions are functioning as expected

I have a basic understanding of Kafka and consumer groups, and the general Kafka documentation is clear, but it’s not obvious:
- What these subscription consumers are expected to consume
- Whether static offsets or missing end offsets are expected behavior
- How to correctly interpret their state when troubleshooting

This makes it difficult to determine whether their behavior is healthy or misconfigured.

This is example of a consumer in kafka-ui:

Image

### Solution Brainstorm

Add documentation that explains:
- What **Snuba subscription consumers** are responsible for
- Which topics they consume from and how
- Why offsets may appear static or missing
- What “healthy” behavior looks like for these consumers
- Any differences between subscription consumers and other Snuba consumers (for example, event consumers)

aldy505:
Closes https://github.com/getsentry/self-hosted/issues/4157

Most self-hosted users most likely won't read Snuba's documentation (https://getsentry.github.io/snuba/) and it mostly only includes technical stuff related to ClickHouse, but not the thoughts behind them.


sfanahata (approved):
🤘

PR #17791, opened 2026-05-18
feat(develop-docs): Add timestamp to Metrics API

philprime:
### Problem

The current Metrics API (`Sentry.metrics.count`, `.gauge`, `.distribution`) always timestamps each metric at the moment the SDK method is called. 
There is no way for the caller to supply a timestamp.

This is fine when application code emits metrics inline (e.g. incrementing a counter inside a request handler), but it breaks down for an important class of use cases: **scraper-style collection**, where an external process reads a batch of metrics from a source and then forwards them to Sentry.

#### Concrete example: Prometheus scraping

A Prometheus-compatible application exposes a `/metrics` endpoint that returns all current metric values in a single response:

```
# HELP go_memstats_heap_alloc_bytes Number of heap bytes allocated and still in use.
# TYPE go_memstats_heap_alloc_bytes gauge
go_memstats_heap_alloc_bytes 7.261032e+06

# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 43

# HELP process_cpu_seconds_total Total user and system CPU time spent in seconds.
# TYPE process_cpu_seconds_total counter
process_cpu_seconds_total 11.32

# HELP promhttp_metric_handler_requests_in_flight Current number of scrapes being served.
# TYPE promhttp_metric_handler_requests_in_flight gauge
promhttp_metric_handler_requests_in_flight 1
```

A scraper fetches this snapshot at time **T**, then iterates through the metrics and calls `Sentry.metrics.gauge(...)`, `Sentry.metrics.count(...)`, etc. for each one.

**Without a user-supplied timestamp**, each SDK call captures `Date.now()` independently.
Even on a fast machine, iterating through dozens of metrics introduces sub-millisecond to low-millisecond drift between calls. 
This means metrics that were observed at the exact same instant end up with slightly different timestamps.

This change is necessary if we want to extend our [sentry-kubernetes](https://github.com/getsentry/sentry-kubernetes) integration to support Prometheus-style scraping of metrics in a Kubernetes cluster.

### Why this matters: bucket boundaries

Sentry aggregates metrics into time buckets. 
When two metrics that logically belong to the same observation are recorded with timestamps close to the bucket boundary, they land in different buckets. 
This causes:

Metrics from the same scrape can be close to a bucket boundary, splitting a single observation into two partial data points and causing artificial jitter or gaps on dashboards and correlations.

The probability of hitting a bucket boundary increases with the number of metrics per scrape and the scrape frequency.

### Solution

Add an optional `timestamp` parameter to the `options` object of all three metrics methods (`count`, `gauge`, `distribution`).

```javascript
const scrapedAt = new Date();
const metrics = scrapePrometheusEndpoint(url);

for (const m of metrics) {
  Sentry.metrics.gauge(m.name, m.value, {
    timestamp: scrapedAt,  // all metrics share the same observation time
    attributes: m.labels,
  });
}
```

When provided, the SDK uses the caller's timestamp for the metric payload instead of `Date.now()`.
When omitted, behavior is unchanged (current time).

When a timestamp is set, we need to ensure the ordering of metrics as described in [Metric Ordering](https://develop.sentry.dev/sdk/telemetry/metrics/#metric-ordering).

### Type

SDKs should accept the language's native date/time type for SDK user ergonomics:

| Language   | Native type          |
|------------|----------------------|
| JavaScript | `Date`               |
| Python     | `datetime`           |
| Java       | `Instant`            |
| Swift      | `Date`               |
| PHP        | `DateTimeInterface`  |

The SDK converts to seconds-since-epoch for the wire format, which already has a required `timestamp` field. 
This change should be aligned with the current implementation of the SDKs.

#### Batching

The wire format already supports batching (the `items` array), but the public API is per-metric (`count`, `gauge`,`distribution`). 
A batch API would be a much larger surface area change than necessary to resolve the issue and raises a couple of concerns:

- A generic `SentrySDK.metrics.capture(items: MetricItem[])` would introduce an object-based public API. At this point the users only need to call the API with the values and options, so this is a conceptual change.
- Allowing multiple values, e.g. `SentrySDK.metrics.gaugeBatch(values: number[], options)` would capture all of the values with the same options, which is not always the right choice.
- Allowing multiple values, e.g. `SentrySDK.metrics.gaugeBatch(items: { value: number, options }[])` is the same as if we introduce an external item type.

All of these changes bring in additional complexity compared to just allowing the same timestamp to be set.

#### Use `beforeSendMetric` to override the timestamp

This callback receives the metric object before sending, so it could theoretically mutate the timestamp. 
But it runs at flush time (potentially seconds later), not at capture time, and it doesn't have access to the caller's intended timestamp without out-of-band state.

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST


develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -377,6 +380,7 @@ SDKs **MUST** expose metrics methods in a `metrics` module or namespace. At mini
   - `unit` **String, optional** — the unit of measurement (only used for `distribution` and `gauge`). SDKs **MUST NOT** restrict what can be sent in (e.g. by using an enum) but **SHOULD** offer constants or similar that help customers send in <Link to="/sdk/foundations/state-management/scopes/attributes/#units">units we support</Link> (since 2.4.0).
   - `attributes` **Object, optional** — a dictionary of attributes (key-value pairs with type information).
   - `scope` **Scope, optional** — the scope to capture the metric with.
+  - `timestamp` **Timestamp, optional** — indicates when the metric was recorded (since 2.9.0). SDKs **SHOULD** accept the language's native date/time type (e.g. `Date` in JavaScript/Swift, `datetime` in Python, `Instant` in Java, `DateTimeInterface` in PHP). When provided, the SDK **MUST** convert the value to seconds since epoch and use it as the `timestamp` field in the [metric payload](#metric-payload). When not provided, the SDK **MUST** default to the current time at the moment the metric is captured. If no such type exists, SDKs **MUST** accept a numeric Unix timestamp in seconds (epoch time) as a fallback instead.
```
giortzisg: > When provided, the SDK **MUST** convert the value to seconds since epoch and use it as the `timestamp`

IIRC relay accepts both unix and RFC 3339 strings, so this shouldn't really be a MUST or at least mention that both are supported.

giortzisg (commented):
Nice change @philprime. Though of adding a batch metric API, but this seems like the simplest way to solve the problem currently.

develop-docs/sdk/telemetry/metrics.mdx
philprime: In the section "Metric Payload" the `timestamp` is defined as a number in seconds (epoch time), without mentioning RFC 3339 strings.

giortzisg: I think this is a general misconception on our develop docs. Relay accepts both last time I checked.

dingsdax (approved):

develop-docs/sdk/telemetry/metrics.mdx
Dav1dde: Relay does accept both and there is no plan to change that.

coolguyzone (approved):
Looks good, thanks for updating.

philprime:
From internal discussion with @logaretm, the following was proposed:

---

1. Multiple metric items can use the same timestamp. 
  **Proposal:** we keep using different sequence numbers to disambiguate
2. Right now the sequence counter in the SDK is resetting when the last used timestamp changes, this breaks when user collects a metric with timestamp `A`, then one with timestamp `B` and then one with `A` again. 
  **Proposal:** the sequence counter is an Int64, continuously incrementing from SDK start, never resets. Will be sent with every metric item
3. When log sequence counter hits Int64 max it would overflow. 
  **Proposal:** We do a reset to 0 and accept the blip, but probability of this happening is super very low (requires a ton of metrics in one SDK lifetime)
4. There is a chance multiple SDK instances on multiple devices/replicas emit metrics at the same timestamp with the same sequence number. There is no shared sequence logic available. 
  **Proposal:** At SDK init we store a timestamp, and this one is included as an attribute too. Chances of collision of timestamp + sequence number + exact same SDK start time is super low and can be considered out-of-scope. 
  **Alternative:** maybe we have another value sent with envelopes that can be used, must be an monotonic growing number to allow ordering

_Note:_

- Combining the timestamp in milliseconds with the sequence number as nanoseconds to synthesize a "virtual" timestamp in the backend, limits the amount of metrics per timestamp to 1,000. This could be established as a known limitation and the upper limit of metrics per timestamp in our docs. No further handling in the SDK.

---

Based on internal feedback, ordering is not as important for Metrics as they are for Logs. Therefore I am going to update the PR to incorporate proposals 1-3 but proposal 4 is out-of-scope due to YAGNI. If necessary it can be added at a later stage.

develop-docs/sdk/telemetry/metrics.mdx
philprime: Updated to wording to reflect both options

philprime:
Waiting for final approval of @logaretm to confirm wording and details based on the proposed changes.

develop-docs/sdk/telemetry/metrics.mdx
```diff
@@ -202,16 +205,22 @@ Whenever possible, metrics **SHOULD** be linked to replays. If a metric is recor
...
+- Reset to `0` only when the SDK is re-initialized, or when the counter reaches the maximum `Int64` value (`2^63 - 1`). Overflow to `0` is acceptable because the probability of reaching this limit within a single SDK lifetime is negligible.
 
-The sequence provides deterministic ordering within a single SDK instance. It does not guarantee ordering across independent processes or workers, which have separate counters. The reset behavior ensures the sequence only increments for consecutive metrics that share the same timestamp.
+The sequence provides deterministic ordering within a single SDK instance. It does not guarantee ordering across independent processes or workers, which have separate counters.
+
+<Alert title="Known Limitation" level="info">
+
+The backend synthesizes a virtual timestamp from the metric's integer millisecond and the sequence number. This limits the number of metrics that share the same integer millisecond timestamp to **1,000** per SDK instance. SDKs do not need to enforce this limit. It is a backend constraint that users should be aware of.
+
```
logaretm: I think we need to adjust the wording here a bit, because from SDK POV, we shouldn't suggest what the sequence number be used for.

Maybe a simple "If the backend synthesizes..." would both document this caveat without suggesting how it would be used on the BE service.

WDYT?

philprime: I was considering to just remove this alert in general, I just rewrote it because it was added to the page before. If you believe this is out-of-scope, let's keep it clean.

logaretm: I think we can remove it, Which gives the BE service flexibility on how to use it.

PR #17796, opened 2026-05-18
Coolguyzone/chore/ai rules robots

coolguyzone:
## DESCRIBE YOUR PR

This PR addresses some of the issue found here: https://www.mintlify.com/score/sentry

These updates to robots.txt add AI bot rules and content signals to help agents better navigate our docs.

More on content signals: https://contentsignals.org/

Some details of this change:

  - All bots get Allow: / — appropriate for public documentation
  - The wildcard User-agent: * block now also has an explicit Allow: /
  (previously it had no rule at all, which is technically ambiguous)
 - If we ever wants to block a specific crawler, it's a one-line change to Disallow: /.
  - isDeveloperDocs is now evaluated at module load time rather than inside the
   handler, which is fine since it's a build-time env var

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

## LEGAL BOILERPLATE

## EXTRA RESOURCES

### DIFF
--- a/app/robots.txt/route.ts
+++ b/app/robots.txt/route.ts
@@ -1,10 +1,43 @@
 import {isDeveloperDocs} from 'sentry-docs/isDeveloperDocs';
 
+const sitemap = isDeveloperDocs
+  ? 'https://develop.sentry.dev/sitemap.xml'
+  : 'https://docs.sentry.io/sitemap.xml';
+
 export const GET = (_request: Request) => {
   return new Response(
     `
-Sitemap: ${isDeveloperDocs ? 'https://develop.sentry.dev/sitemap.xml' : 'https://docs.sentry.io/sitemap.xml'}
+Sitemap: ${sitemap}
+
 User-agent: *
+Allow: /
+
+User-agent: GPTBot
+Allow: /
+
+User-agent: OAI-SearchBot
+Allow: /
+
+User-agent: Claude-Web
+Allow: /
+
+User-agent: anthropic-ai
+Allow: /
+
+User-agent: Google-Extended
+Allow: /
+
+User-agent: Amazonbot
+Allow: /
+
+User-agent: Bytespider
+Allow: /
+
+User-agent: CCBot
+Allow: /
+
+User-agent: Applebot-Extended
+Allow: /
 `.trim(),
     {headers: {'content-type': 'text/plain'}}
   );
--- a/app/robots.txt/route.ts
+++ b/app/robots.txt/route.ts
@@ -12,42 +12,6 @@ Sitemap: ${sitemap}
 User-agent: *
 Allow: /
 Content-Signal: ai-train=yes, search=yes, ai-input=yes
-
-User-agent: GPTBot
-Allow: /
-Content-Signal: ai-train=yes, search=yes, ai-input=yes
-
-User-agent: OAI-SearchBot
-Allow: /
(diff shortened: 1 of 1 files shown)


sergical (approved):
noice

app/robots.txt/route.ts
```diff
@@ -1,10 +1,53 @@
...
+
+User-agent: GPTBot
+Allow: /
+Content-Signal: ai-train=yes, search=yes, ai-input=yes
+
+User-agent: OAI-SearchBot
+Allow: /
+Content-Signal: ai-train=yes, search=yes, ai-input=yes
+
+User-agent: Claude-Web
```
sfanahata:
```suggestion
User-agent: ClaudeBot
```

app/robots.txt/route.ts
```diff
@@ -1,10 +1,53 @@
...
+Allow: /
+Content-Signal: ai-train=yes, search=yes, ai-input=yes
+
+User-agent: Bytespider
+Allow: /
+Content-Signal: ai-train=yes, search=yes, ai-input=yes
+
+User-agent: CCBot
+Allow: /
+Content-Signal: ai-train=yes, search=yes, ai-input=yes
```
sfanahata: Do we want to explicitly allow trainers to train on our docs? I kind of feel like no.

sergical: i had the same feeling but then i thought if they don't train on our current docs, then LLMs will keep showing people old ways of using Sentry which might not be ideal 😅

app/robots.txt/route.ts
```diff
@@ -1,10 +1,53 @@
...
+  ? 'https://develop.sentry.dev/sitemap.xml'
+  : 'https://docs.sentry.io/sitemap.xml';
+
 export const GET = (_request: Request) => {
   return new Response(
     `
-Sitemap: ${isDeveloperDocs ? 'https://develop.sentry.dev/sitemap.xml' : 'https://docs.sentry.io/sitemap.xml'}
+Sitemap: ${sitemap}
+
 User-agent: *
```
sfanahata, line 12: @coolguyzone since we're wildcard allowing every bot, I'm not sure it makes sense to add in more specific bots that we're also carte blanche allowing? Beside the trainers, which we might want to update in what they're allowed to do, I wonder if we remove the explicits of each bot? Otherwise, we're still missing some, like 

- ChatGPT-User 
- PerplexityBot
- Perplexity-User
- Meta-ExternalAgent
- cohere-ai
- Diffbot

And that list will continue to get longer/need updating as more bots are created.

app/robots.txt/route.ts
sfanahata: Ah, good point. I guess train is a wide range of things. It's too early in content-signal land to be explicit. Sounds like `content-signal` is all experimental and not adopted anyway, so not sure it'll actually prevent any bot from taking an action.

coolguyzone: Yeah, I had this discussion with Matt, it feels like are direction so far has been to optimize the docs for LLM usage, so training seemed like the right call. I'm open to hearing any objections though.

app/robots.txt/route.ts
coolguyzone: Yeah, in the future it might make sense if we are adding different rules for different bots but you're right, for now it makes sense just to collapse everything to a wildcard.

PR #17803, opened 2026-05-19
fix: documentation on code call is wrong

michaelmairegger:
The original code sample for the call to `StartTransaction` won't compile. There's no overload of that method that takes both a transaction context and the name/operation parameters. 

This PR fixes the code sample so that it compiles correctly.


docs/platforms/dotnet/common/tracing/instrumentation/custom-instrumentation/queues-module.mdx
jamescrosswell, line 1: @michaelmairegger I don't see any problem with the original code.

The [underlying method signature](https://github.com/getsentry/sentry-dotnet/blob/0418c99956d495dde6912a429fb189a8df2bde3f/src/Sentry/SentrySdk.cs#L774-L778) is:
```csarp
    public static TransactionContext ContinueTrace(
        SentryTraceHeader? traceHeader,
        BaggageHeader? baggageHeader,
        string? name = null,
        string? operation = null)
```
So the last two parameters are optional. They can be used to override the name/operation of the transaction, but I think for the samples/demos it simplest to omit them.

michaelmairegger: StartTransaction is the method that is incompatible with the method parameters shown in documentation

jamescrosswell: Ah I see - nice catch - thank you!

jamescrosswell (approved):

jamescrosswell:
@Flash0ver I think someone from your office needs to authorise the Vercel checks:
image

jamescrosswell:
> If this PR is still relevant, please:
> 
> * Push new commits, or
> * Leave a comment to keep it open

@Flash0ver @dingsdax this one is waiting for us (it was kindly contributed by @michaelmairegger).

dingsdax:
@michaelmairegger thanks for contributing 🙏 🎉 !

Issue #17799, opened 2026-05-19
Update SDK tracing docs for span-first (JS, Python, Flutter)


inventarSarah:
### SDK

Other

### Description

Update the tracing documentation to introduce the new span-first tracing mode and help users (mainly users using manual instrumentation) opt in and update their code where needed. Ensure existing tracing docs and new span-first docs are consistent and work together.

Basis: #17621

### Background and context
Sentry is moving from a transaction-based tracing model to a span-first model (aligning with how OTel handles it). In the new model, transactions are phased out and everything is a span.

There will be a new (opt-in) tracing mode, currently called "stream spans": spans get sent to Sentry in batches as they finish, rather than being held in memory until the transaction ends. This removes the 1,000-span limit and makes trace data visible sooner.

Impact by sdk:
- JS: already on the new Span API; stream spans is a new opt-in. One or more functions work differently for stream spans so users may need to  update code when switching (even if they already use Span API)
- Python & Flutter: A new Span API is coming  and stream spans will also be available as an opt-in.

The changes mainly affect users who do manual instrumentation and want to opt in to streaming mode, or those hitting the limitations of the current transaction model.
#### Naming recommendations: 

- New spans
  - Can describe how they differ from transactions. These are stream mode spans…
- Stream mode - the way spans are formatted:
  - is a tracing mode, where spans are sent to Sentry in batches as they finish, rather than being held in memory until the transaction ends. This removes the 1k-span limit, reduces memory usage, and makes trace data visible in Sentry sooner.
  - This allows for describing stream mode vs transaction mode
- Root span - the topmost span connected to a trace ID
- Child spans - any span nested under a parent, connected to the same trace ID
- Service span - a parent-level span at top of a service, nested under the same trace ID as the root span

#### What is a Service Span?
[Developer Docs Span API](https://develop.sentry.dev/sdk/telemetry/spans/span-api/)

The topmost span within a service boundary is called the service span.  

For example, a distributed trace from frontend to backend, would have a service span for the backend, and a service span for the frontend. The frontend service span is also the root span of the entire span tree.
Visualization:
```
Trace (trace_id)
│
└── Root span  [frontend]       ← also a service span
     ├── Child span
     │    └── Child span
     ├── Child span
     │    └── ...
     └── Service span  [backend]   ← remote parent in frontend
          ├── Child span
          └── Child span
               └── Child span
```

#### Hierarchy
Transaction-based:
```
Trace
 └── Root Span
	  └── Child span
[...]

PR #17834, opened 2026-05-21
docs(js): Add Stream Spans guide

### LINKED ISSUE #17835: Document stream mode in JS SDKs
### SDK

JavaScript SDK

### Description

Create a new page under /Tracing to document the new stream mode/new spans.

Based on this draft: #17621

### Suggested Solution

_No response_

inventarSarah:
## DESCRIBE YOUR PR
Adds a new page documenting stream mode for all JS SDKs, covering:
- How to enable stream mode
- Migration path from transaction mode
- Manual instrumentation (start a span, add attributes)
- Distributed tracing 
- Extended configuration (`beforeSendSpan`, `ignoreSpans`)
- Verify setup

Based on this draft: #17621
Michi's draft includes sections on distributed tracing, which I removed in this version. The APIs work identically in both modes; the topics are advanced enough that users are better served by the existing page, where all the details live, and users who already have distributed tracing set up don't need to change anything when switching to stream mode. Happy to discuss if you disagree.

**Questions/Considerations for you**
- Filter Spans and Drop Spans currently use a single snippet per environment (browser/server) rather than framework-specific snippet file -> is that sufficient or should we create framework-specific snippet files, instead?
-  Does the reasoning for removing the Distributed Tracing sections make sense to you? (see explanation above)
- please leave a comment when the guide uses a wrong (or not the ideal) span name (eg root span vs service span, service span vs child span, ...)

Closes: #17835

## IS YOUR CHANGE URGENT?  

- [x] Other deadline: end of May

## SLA

## PRE-MERGE CHECKLIST

## EXTRA RESOURCES


docs/platforms/javascript/common/tracing/new-spans/index.mdx
```diff
@@ -0,0 +1,373 @@
...
+  ignoreSpans: [
+    // Drop spans whose name contains "healthcheck"
+    "healthcheck",
+    // Drop spans whose name matches a pattern
+    /^GET \/api\/v1\/internal/,
+    // Drop spans matching name and attribute conditions
+    {
+      name: /^GET \//,
+      attributes: {
+        "http.response.status_code": 200,
```
Lms24: m: I think we should remove the response status code attribute or use another example here because this one suggests that something is possible that actually isn't: `ignoreSpans` is only applied at span start. At this time, the span will not yet have the response status code attribute. Usually this attribute is added at or close to span end. At which point we already have decided whether to keep the span or not.

```suggestion
```

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
Lms24, line 1: we mix "service span" and "root span" in this doc. A couple of thoughts:

- I really like the "service span" name (as opposed to "segment span" how we call it in SDK code). I've probably been quite out of the loop here: Is this the public name we ended up choosing for it? Just want to confirm.
- It's probably fine to use "root span" when we talk about how things used to be with transactions but further below, we explain the `ignoreSpans` behaviour differences for child vs. root spans. Should we adjust to "child vs. service spans" there?

docs/platforms/javascript/common/tracing/new-spans/index.mdx
```diff
@@ -0,0 +1,373 @@
...
+
+</PlatformCategorySection>
+
+</SplitSectionCode>
+</SplitSection>
+</SplitLayout>
+
+If a matching span is a root span, all of its child spans are dropped as well. If a child span matches, only that span is dropped and its children are reparented to the nearest ancestor.
+
+In stream mode, `ignoreSpans` is evaluated at span start, so only the span name and attributes available at that point are taken into account. Any name updates or additional attributes added while the span is active won't invluence whether the span is dropped.
```
Lms24: (nit)
```suggestion
In stream mode, `ignoreSpans` is evaluated at span start, so only the span name and attributes available at that point are taken into account. Any name updates or additional attributes added while the span is active won't influence whether the span is dropped.
```

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
```diff
@@ -0,0 +1,373 @@
...
+
+## Verify Your Setup
+
+To make sure you've enabled stream mode successfully:
+
+<PlatformCategorySection supported={["server", "serverless"]}>
+
+<PlatformSection notSupported={["javascript.nextjs", "javascript.nuxt", "javascript.sveltekit", "javascript.astro", "javascript.solidstart", "javascript.capacitor", "javascript.tanstackstart-react", "javascript.remix", "javascript.react-router"]}>
+
+- **Check the Sentry dashboard**: Spans should appear in the Traces view shortly after they complete. Traces look the same as in transaction mode, but without transactions.
```
Lms24, line 384: > Traces look the same as in transaction mode, but without transactions

This is true but unless I missed something, there are very few visual indications that actually point out that a span is a (new) span and not a transaction 😬 The only thing that comes to mind is that it's not showing a JSON button anymore. I don't think there's harm in leaving in this sentence but users might get confused at what they should be looking in the product.

This is of course not a primary docs issue but rather a product/UI issue. We might want to think about some kind of indicator (and, separate topic, please bring back JSON for streamed spans lol).  That is assuming, I didn't miss any obvious indication 😅 Will ask product folks about this.

docs/platforms/javascript/common/tracing/new-spans/index.mdx
inventarSarah: Makes sense - I removed it 👍

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
inventarSarah: - "service span" is a naming recommendation provided by Shannon -- I've collected the full list [here](https://github.com/getsentry/sentry-docs/issues/17799)
- I would really like to get use all the different span correctly -- it would be great if you (reviewers) leave a comment when the guide uses the wrong (or not best fitting) name

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
inventarSarah: In that case, I will update this to something like 
"Traces look almost the same as in transaction mode. The main difference is that they have no transactions and only consist of spans."

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
```diff
@@ -0,0 +1,370 @@
+---
+title: New Spans
+description: "Learn how to use stream mode to send spans to Sentry as they finish, removing the 1,000-span limit and making trace data visible sooner."
```
sfanahata, line 3: @cleptric - should we be calling out the 1k span limit as a benefit? Seems really valuable, but want to make sure we're not locking ourselves into something unnecessarily.

docs/platforms/javascript/common/tracing/new-spans/index.mdx
```diff
@@ -0,0 +1,370 @@
+---
+title: New Spans
+description: "Learn how to use stream mode to send spans to Sentry as they finish, removing the 1,000-span limit and making trace data visible sooner."
+sidebar_order: 35
+new: true
+notSupported:
+  - javascript.cordova
+---
+
+<AvailableSince version="10.53.1" />
```
sfanahata: This is SDK version? Would be good to clarify.

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
sfanahata: I think it could be useful to add two parts to the index to help: 
1. Definitions (root span, service span, child span)
2. A chart like you had in our notes doc, Sarah, to show where each span would live in a trace

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
sfanahata: In the new experience, what do we call those attributes that we currently call `transaction`?

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
```diff
@@ -0,0 +1,370 @@
...
+</SplitLayout>
+
+## Distributed Tracing (Optional)
+
+Distributed tracing works out of the box when tracing is enabled and works the same way in stream mode. If you need to manually propagate trace context, for example,
+when the SDK can't instrument automatically, see <PlatformLink to="/tracing/distributed-tracing/custom-instrumentation/">Custom Trace Propagation</PlatformLink>.
+
+## Extended Configuration (Optional)
+
+### Filter Spans
```
sfanahata, line 206: I think this may be the right place for it, but somewhere on this page I believe we should mention with a slight bit more detail in the description that you also need to modify `span.op` to `span.attributes?.["sentry.op"]` and that `description` becomes `name` in this new paradigm.

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
```diff
@@ -0,0 +1,370 @@
...
+<Alert level="warning" title="Migrating from transaction mode?">
+
+If you're using `beforeSendSpan`, wrap it with `Sentry.withStreamedSpan()` as shown above, otherwise the SDK falls back to transaction mode.
+The callback signature and properties of the span object are also shaped differently. See <PlatformLink to="/configuration/options/#beforesendspan">`beforeSendSpan`</PlatformLink> for the full reference.
+
+If you're using `beforeSendTransaction` to drop spans, use [`ignoreSpans`](#drop-spans) instead, since `beforeSendTransaction` is not available in stream mode.
+
+</Alert>
+
+### Drop Spans
```
sfanahata, line 292: @Lms24 - just wanting to verify that this behavior is true: 

In transaction mode, ignoreSpans is evaluated at transaction end (so all attributes accumulated during the span's lifetime are available). In stream mode, it's evaluated at span start (so only initial attributes are considered).

If so, @inventarSarah, it may be useful to add that detail or heads up here.

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
sfanahata: Let me know if this is true, though @Lms24 - 

- Root span - the topmost span connected to a trace ID
- Child spans - any span nested under a parent, connected to the same trace ID
- Service span - a parent-level span at the top of a service, nested under the same trace ID as the root span

docs/platforms/javascript/common/tracing/new-spans/index.mdx
inventarSarah: Actually, let me remove that because we have the required SDK version in the Prerequisites section.

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
inventarSarah: Yes, you can find this down below at line 339+
It will render below the split layout section here because the info isn't relevant to the code snippet. 
But let me know if you think it would make more sense, or help users avoid missing this info, if we pull it into the split layout.

inventarSarah: I moved the stream mode-specific explanation under the section header and left the stream mode vs transaction mode explanations in the migration callout below - wdyt?

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
inventarSarah: Updated this section to mention these changes in the text accompanying the code snippet. And then I referred to the "migration" note, where I've added the table with all name changes.
wdyt?

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
inventarSarah: That's a good idea -- I added the definitions and the graph to the intro of the page

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
inventarSarah: @Lms24 Have you received a response from the product team yet?

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
Lms24: Yes this is correct @sfanahata and @inventarSarah. I like the explainer paragraph of the terms and the graph as well - good ideas! 

>  would really like to get use all the different span correctly -- it would be great if you (reviewers) leave a comment when the guide uses the wrong (or not best fitting) name

Will do in a bit!

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
Lms24: No response so far :( I suggest for now we go with the more generic sentence. If we decide to mark them specifically in the product, we can update this section. (I'm hacking on a small visual indication but no guarantees that this will get approved/merged).

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
Lms24: Don't wanna answer for Michi (so feel free to overrule me) but removing the hard 1k limit is the key feature of span streaming. I think we can/should call it out. The 1k limit exists but isn't documented publicly a lot (besides [this](https://sentry.zendesk.com/hc/en-us/articles/25471509279515-Transaction-with-many-spans-doesn-t-appear-in-Sentry) zendesk page). I don't think it's a secret though, so no harm in mentioning it.

docs/platforms/javascript/common/tracing/new-spans/index.mdx
```diff
@@ -0,0 +1,398 @@
...
+- **Lower memory usage.** Spans are flushed periodically and don't need to be held in memory until the root span ends. This is especially useful for long-running processes like queue consumers or cron jobs.
+- **Faster visibility.** Span data arrives in Sentry as your application runs, instead of only after the entire operation completes.
+- **No data loss from crashes.** If your process terminates unexpectedly, spans that were already flushed are preserved. In transaction mode, a crash before the root span ends means all span data is lost.
+
+</Expandable>
+
+You can find the following span types mentioned throughout this page:
+
+- **Root span**: The topmost span in a trace. It has no parent span and is always a service span.
+- **Service span**: A parent-level span at the entry of a service. In transaction mode, this is called transaction.
```
sfanahata:
```suggestion
- **Service span**: A parent-level span at the entry of a service. In transaction mode, this is called a transaction.
```

docs/platforms/javascript/common/tracing/new-spans/index.mdx
```diff
@@ -0,0 +1,398 @@
...
+</SplitSectionText>
+<SplitSectionCode>
+
+```javascript
+Sentry.startSpan(
+  {
+    name: "process-order",
+    op: "queue.process",
+    attributes: {
+      "order.id": "abc-123",
```
Lms24: l: Setting top-level `op` is indeed still possible in the SDK but to align with how we then _send_ op, I suggest we document setting the `sentry.op` attribute right away. Fwiw, it's totally fine that this example is the only one that sets an op, because usually we don't require users to think about ops in their manually started spans. Some still set one, for manual instrumentation that feeds a specific Sentry feature, so it's fine.

(I only caught this now, sorry for the late notice)

```suggestion
    attributes: {
      "sentry.op": "queue.process",
      "order.id": "abc-123",
```

docs/platforms/javascript/common/tracing/new-spans/index.mdx
```diff
@@ -0,0 +1,398 @@
...
+  ignoreSpans: [
+    // Drop spans whose name contains "healthcheck"
+    "healthcheck",
+    // Drop spans whose name matches a pattern
+    /^GET \/api\/v1\/internal/,
+    // Drop spans matching name and attribute conditions
+    {
+      name: /^GET \//,
+      attributes: {
+        "http.response.status_code": 200,
```
Lms24: l: let's remove this attribute as it isn't available at span start time
```suggestion
```

docs/platforms/javascript/common/tracing/new-spans/index.mdx
```diff
@@ -0,0 +1,398 @@
...
+
+</SplitSectionCode>
+</SplitSection>
+</SplitLayout>
+
+If a matching span is a service span, all of its child spans are dropped as well. If a child span matches, only that span is dropped and its children are reparented to the nearest ancestor.
+
+<Alert level="warning" title="Migrating from transaction mode?">
+
+In transaction mode, `ignoreSpans` is evaluated at transaction end rather than at span start. Review your existing rules to make sure the attributes you're matching on are passed when the span is created.
```
Lms24: l: just to be extra clear
```suggestion
In transaction mode, `ignoreSpans` is evaluated at transaction end rather than at span start. Review your existing rules to make sure the attributes and names you're matching on are passed when the span is created.
```

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
Lms24: Update: Didn't find any confusing places anymore from my original review. All good from my end!

Lms24 (commented):
Second pass: Just had some minor remarks, other than that this looks good to me!

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
Lms24: I opened https://github.com/getsentry/sentry/pull/116386 which proposes a visual indication. Can't promise that this gets merged though yet. Will keep you posted!

Lms24 (approved):
Thanks for making the changes!

docs/platforms/javascript/common/tracing/streamed-spans/index.mdx
cleptric: Yes, it's a nice benefit.

PR #17746, opened 2026-05-13
chore(platform): Enable span streaming

chargome:
Enables Sentry span streaming for all runtimes.

PR #17863, opened 2026-05-22
chore(deps): Bump to latest Sentry SDKs

Lms24:
Bump SDKs to 
1. test `no_parent_span` client outcome
2. eventually enable span streaming (though not yet)

(Also noticed we seem to explicitly declare `@sentry/browser` as a dep but we only have one use case which I think should be covered by the NextJS SDK as well. Will investigate if we can remove the explicit dep).


chargome (approved):
fyi there is https://github.com/getsentry/sentry-docs/pull/17746

Lms24:
oh lol, sorry! Gonna close my PR then. I'll find something else to test the outcomes then.

PR #17875, opened 2026-05-26
feat(tanstackstart-react): Add component tracking

nicohrubec:
Adds `reactComponentAnnotation` docs to the `sentryTanstackStart` Vite plugin page. Enabled in the SDK via https://github.com/getsentry/sentry-javascript/pull/21149. This should be merged after the next SDK release (`10.55.0`).

Also changed the structure to have a separate section for each TSS specific option and put them all one level below the `Configuration Options` section.

### DIFF
--- a/docs/platforms/javascript/guides/tanstackstart-react/features/sentryTanstackStart.mdx
+++ b/docs/platforms/javascript/guides/tanstackstart-react/features/sentryTanstackStart.mdx
@@ -53,7 +53,9 @@ export default defineConfig({
 
 The plugin passes through all options to the underlying [`@sentry/vite-plugin`](/platforms/javascript/sourcemaps/uploading/vite/). See the [Sentry Vite Plugin documentation](/platforms/javascript/sourcemaps/uploading/vite/) for the full list of available options.
 
-Additionally, the plugin automatically instruments all TanStack Start middlewares for tracing by default. To disable this behavior:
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/guides/tanstackstart-react/features/sentryTanstackStart.mdx
```diff
@@ -62,15 +64,40 @@ sentryTanstackStart({
 }),
 ```
 
-## Tunnel Route
+### Component Name Annotation
+
+The `reactComponentAnnotation` option annotates your React components with `data-sentry-component` and `data-sentry-source-file` attributes at build time. This lets you see component names instead of generic selectors in Session Replay, breadcrumbs, and performance monitoring.
```
s1gr1d, line 69: I am not 100% sure how we handle capitalization in the docs but we should be consistent here (as Replay is capitalized as it's a feature name but the others are not capitalized).

s1gr1d (approved):

docs/platforms/javascript/guides/tanstackstart-react/features/sentryTanstackStart.mdx
nicohrubec: From what I've seen this is the capitalization that we use in the docs everywhere so should be consistent

PR #17876, opened 2026-05-26
feat(tanstackstart-react): Move client-side init to client entry point

nicohrubec:
Moves the recommended client-side `Sentry.init()` from `src/router.tsx` to `src/client.tsx`. SDK change in https://github.com/getsentry/sentry-javascript/pull/21161.


coolguyzone (approved):
Makes sense, thanks!

PR #17880, opened 2026-05-26
docs(unreal): Add Session Replay feature

tustanivsky:
This PR adds a dedicated "Session Replay" docs page for Unreal SDK.

Related items:
- https://github.com/getsentry/sentry-unreal/pull/1404
- https://github.com/getsentry/sentry-unreal/pull/1386
- https://github.com/getsentry/sentry-xbox/pull/136

### DIFF
--- a/docs/platforms/unreal/session-replay/index.mdx
+++ b/docs/platforms/unreal/session-replay/index.mdx
@@ -1,0 +1,97 @@
+---
+title: Set Up Session Replay
+sidebar_title: Session Replay
+description: "Learn how to capture gameplay video and attach it to crash reports with Sentry's Unreal Engine SDK."
+sidebar_order: 5500
+sidebar_section: features
+---
+
+<Alert level="warning">
+
+Session Replay for Unreal Engine is **experimental** and currently limited to a small set of platforms (see [Platform Support](#platform-support)). The API and behavior may change in future releases.
+
+</Alert>
+
+Session Replay records what was happening on screen in the moments leading up to a crash, giving you a visual reproduction to pair with the captured event.
+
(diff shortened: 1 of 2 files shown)


docs/platforms/unreal/session-replay/index.mdx
```diff
@@ -0,0 +1,97 @@
+---
+title: Set Up Session Replay
+sidebar_title: Session Replay
+description: "Learn how to capture gameplay video and attach it to crash reports with Sentry's Unreal Engine SDK."
+sidebar_order: 5500
+sidebar_section: features
```
JoshuaMoelans, line 6: we could mark it with the green `new` icon (like we used for metrics/logs) maybe? not sure if allowed if the product itself isn't new, but only the platform is 🤔 
```suggestion
sidebar_section: features
new: true
```

tustanivsky: Given that Session Replay is still an experimental feature I think it might be too early to highlight it as “New”. We can probably do that later once we’ve gathered more customer feedback and the configuration steps and limitations are clearer.

JoshuaMoelans: okay that makes sense. We do have other markers ([`BETA`](https://docs.sentry.io/ai/dotagents/), [`EA`](https://docs.sentry.io/platforms/android/snapshots/)) that we could use instead if we still want it to show as being experimental? If not, I think just adding the "new" once actually released makes sense too.

JoshuaMoelans (approved):

coolguyzone (approved):

mujacica (approved):

PR #17898, opened 2026-05-27
docs(react-native): document .env.local for Expo auth token

betegon:
The wizard listed `.env.local` as a file it creates, but never showed what goes in it. The Expo plugin snippet also showed a bare `export` shell command, which doesn't work for EAS builds.

**What changed:**

- `react-native-expo-plugin-code-snippet.mdx`: replaced the shell `export` command with a proper `.env.local` code block (format that actually works for local builds), and added a note directing EAS users to add `SENTRY_AUTH_TOKEN` as an EAS secret instead
- `expo.mdx` Notes: replaced vague "use an environment variable like `SENTRY_AUTH_TOKEN`" with explicit `.env.local` and EAS secret guidance
- `index.mdx` + `expo.mdx`: fixed typo `env.local` → `.env.local` in the wizard task expandables

Feedback from Antonis Lilis testing `sentry init` on Expo — the missing `.env.local` meant EAS builds couldn't upload source maps.


antonis (approved):
LGTM! Thank you for fixing that 🙇

PR #17918, opened 2026-05-28
macuzi/fix/ask ai button

### LINKED ISSUE #17351: Ask AI control is not keyboard-focusable (div instead of button)
The "Ask AI" control in the docs header (`docs.sentry.io`) is rendered as a styled `div` instead of a native `<button>`, which excludes it from the tab order and makes it unreachable via keyboard navigation or assistive technologies.

Replacing it with a native `<button>` (or adding `role="button"` + `tabindex="0"` + keyboard event handlers as a fallback) restores standard focus behavior.

- **Reproduction**: tab through the docs header — the Ask AI control is skipped entirely
- **Quick verify**: open DevTools, temporarily swap the `div` to a `button` → element becomes keyboard-focusable as expected
- **Fix**: use a native `<button>` element; avoid ARIA role workarounds when a semantic element fits

Action taken on behalf of Lamberto Asghedom.

macuzi:
## Describe Your PR

Fixes [#17351](https://github.com/getsentry/sentry-docs/issues/17351).

The "Ask AI" was rendered through Radix `Button asChild`, but its child was a [`<div>`](https://github.com/getsentry/sentry-docs/blob/00c76b0f7669711663e1e7b6fa956afb8d941655/src/components/search/index.tsx#L348-L351). Which won't behave like a native button and was being skipped during keyboard navigation.

https://github.com/user-attachments/assets/145251d7-6fab-4bbc-81c4-7c2b398354e3

Proposing a change from a `div` to a native `<button type="button">`, which puts "Ask AI" in the normal tab order and gives it standard keyboard behavior. It keeps the Radix `Button asChild` wrapper so the existing button styling, spacing, and icon layout stay the same.

For transparency, I tried replacing the Radix wrapper with a plain native button in [7175020](https://github.com/getsentry/sentry-docs/commit/7175020cad03b4212e1818079170d7b0fb8391a9), but that introduced small styling differences. For this fix, changing the rendered child element is a smaller and safer path.

https://github.com/user-attachments/assets/c58c931b-8aae-45f2-a8ee-52ca8a4e64b5


sergical (approved):
semantics and a11y ftw! thank you!

PR #17921, opened 2026-05-28
docs(logs): clarify attribute value types for Python logs

sergical:
## Summary

Clarifies which value types are supported for Python log attributes, after a user reported that array-valued attributes (e.g. `list[int]` like `workflow_ids`/`detector_ids`) silently show up empty in the Logs UI.

## Findings (validated against a real project)

Sentry's log storage (EAP) only retains **scalar** attribute types — string, number, boolean. The Python SDK (`sentry_sdk.utils.format_attribute`) sends a homogeneous scalar `list`/`tuple` as a real **array**, which the backend does not store, so it appears empty. Mixed lists and `dict`s are `repr`'d to strings and *do* survive. (For contrast, `@sentry/core` stringifies all non-primitives via `serializeAttributes(attrs, true)`, so JS users see arrays as JSON strings — the two SDKs diverge here.)

## Change

Adds an `<Alert>` to the Python logs usage include documenting that attribute values should be primitives, explaining the array behavior, and recommending serializing complex values (e.g. `json.dumps(...)`) before logging.

A separate `sentry-python` issue tracks the SDK-level inconsistency.


KyleTryon (approved):

platform-includes/logs/usage/python.mdx
```diff
@@ -30,3 +30,9 @@ sentry_sdk.logger.error(
     }
 )
 ```
+
+<Alert title="Attribute value types">
+
+Attribute values should be primitives — strings, numbers, or booleans. Sentry's log storage only retains these scalar types, so other values are handled inconsistently: a `list` or `tuple` of same-typed scalars (for example `[1, 2, 3]`) is sent as an array, which is **not currently stored** and shows up empty in the Logs UI, while mixed lists and `dict`s are stored as their string representation. To keep a complex value searchable, serialize it to a string yourself before logging — for example `json.dumps(workflow_ids)`.
```
sfanahata:
```suggestion
Attribute values should be primitives — strings, numbers, or booleans. Sentry's log storage only retains these scalar types, so other values are handled inconsistently. A `list` or `tuple` of same-typed scalars (for example `[1, 2, 3]`) is sent as an array, which is **not currently stored** and shows up empty in the Logs UI. Mixed lists and `dict`s are stored as their string representation. To keep a complex value searchable, serialize it to a string before logging, for example `json.dumps(workflow_ids)`.
```

sfanahata (approved):
@sergical  - not going to block you, but can we update the run-on sentence alert with my suggestions, or something similar?

PR #17927, opened 2026-05-29
docs(snapshots): Document SnapshotPreviews metadata

cameroncooke:
Document SnapshotPreviews metadata sidecars and modifiers in the Apple snapshots docs.

This keeps the generic Uploading Snapshots page platform-agnostic while adding the new `tags` and `context` fields to the shared JSON metadata schema. SnapshotPreviews-specific details now live on a dedicated Apple page with examples for `.snapshotTags`, `.snapshotAdditionalContext`, and `.snapshotDiffThreshold`.

The Apple setup page stays focused on getting snapshots exported, with a short link to the detailed metadata page for users who want to customize sidecars.


docs/platforms/apple/common/snapshots/snapshotpreviews-metadata.mdx
```diff
@@ -0,0 +1,103 @@
...
+      "device_name": "iPhone 17 Pro",
+      "model_identifier": "iPhone18,1"
+    }
+  }
+}
+```
+
+## Generated Context Keys
+
+SnapshotPreviews can generate these `context` keys:
```
mtopo27: can or does? if this is automatic, think it should be more like "SnapshotPreviews automatically generates..."

mtopo27 (approved):

docs/platforms/apple/common/snapshots/snapshotpreviews-metadata.mdx
cameroncooke: @cursor fix

PR #18275, opened 2026-06-05
docs(dev-docs): Document the spec lifecycle and maturity framework

chargome:
Adds a meta-spec documenting the SDK spec lifecycle framework, which until now existed only in code (`specConstants.ts`) with no written definition of what each status means or when a spec may advance.

The new page defines the status model (proposal, draft, candidate, stable, deprecated), the frontmatter/section structure, version semantics, and—most importantly—the implementation-based promotion criteria between statuses, aligned with the W3C Process maturity stages.

ref https://github.com/getsentry/sentry-docs/pull/16461


develop-docs/sdk/getting-started/standards/spec-lifecycle.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+- There **MUST** be at least **one** SDK implementation demonstrating the spec is feasible as written.
+- The spec version **MUST** be bumped to `1.0.0` (see [Version semantics](#version-semantics)).
+- Open questions that block implementation **MUST** be resolved or explicitly deferred.
+
+### candidate → stable
+
+- There **SHOULD** be at least **two** independent, interoperable SDK implementations of each feature — the signal that the spec is implementable consistently across SDKs, not just in the one that shaped it.
+- Where two implementations don't yet exist, the gap **SHOULD** be recorded (in the changelog summary or an open-questions note) rather than silently promoted.
+- No breaking changes **SHOULD** remain outstanding.
```
dingsdax, line 64: "There SHOULD be at least two independent, interoperable SDK implementations" could be read as a gate on promotion (you need two before promoting), but I think the intent was that there MUST be at least one (from the draft → candidate gate) and a second is preferable?

A small rewording could make the intent clearer. Something like:
* A second independent SDK implementation SHOULD exist to confirm the spec is
  implementable consistently across SDKs, not just in the one that shaped it.
* If only one implementation exists at promotion time, the gap MUST be recorded
  in the changelog summary or an open-questions note.

I would use MUST on the second bullet point to strengthen the transition criteria

dingsdax (commented):
one small comment for the candidate → stable gate, otherwise big ❤️ 

Love the W3C alignment and the clearly defined transition criteria.

It describes very well the "implementation/evidence promotion model" that SDKs have been following implicitly anyways (more or less). Great to see that codified now 👏 and hopefully even better prevents any future specs from drifting into aspirational territory 😅. Good stuff!

develop-docs/sdk/getting-started/standards/spec-lifecycle.mdx
```diff
@@ -0,0 +1,87 @@
...
+</Alert>
+
+### proposal → draft
+
+A proposal becomes a draft once it's written up as a spec page and the team agrees the problem is worth specifying. No implementation is required.
+
+### draft → candidate
+
+- There **MUST** be at least **one** SDK implementation demonstrating the spec is feasible as written.
+- The spec version **MUST** be bumped to `1.0.0` (see [Version semantics](#version-semantics)).
```
cleptric: 1.0.0 for candidate seems weird to me.

cleptric (approved):

s1gr1d (approved):
Thanks for adding this!

dingsdax (approved):

PR #18277, opened 2026-06-05
docs(snapshots): Document SnapshotPreviews selective testing

cameroncooke:
Add an Apple Snapshots guide for using SnapshotPreviews with Sentry selective testing.

**What Changed**

This adds a dedicated selective testing page under the Apple Snapshots docs and links to it from the Apple Snapshots setup guide.

The new page explains how to:

- organize SnapshotPreviews into selectable `SnapshotTest` groups;
- generate image-name manifests for the groups that participate in selective testing;
- merge those manifests into one complete tracked image-name file;
- render only the group selected for the current CI run;
- upload the rendered subset with `--all-image-file-names-file`.

**Why This Exists**

Selective Sentry uploads need to distinguish images that were skipped in the current run from images that were removed from the snapshot suite. SnapshotPreviews provides the image-name manifest; Sentry uses that manifest to treat listed-but-missing images as skipped rather than removed.

**Workflow Constraints**

The page documents the constraints users need to account for:

- name generation and image rendering are separate, mutually exclusive SnapshotPreviews modes;
- manifest generation must run once per participating `SnapshotTest` group;
- CI is responsible for merging the per-group manifest files before upload;
- inclusion filters define selectable groups;
- exclusion filters remove snapshots from the tracked set entirely.

**Examples**

The examples use placeholder project, scheme, and test class names so readers can adapt the workflow to their own app. The page links to the SnapshotPreviews MultiModuleDemo workflow as a complete reference implementation.

Refs EME-1182

### DIFF
--- a/docs/platforms/apple/common/snapshots/selective-testing.mdx
+++ b/docs/platforms/apple/common/snapshots/selective-testing.mdx
@@ -4,9 +4,8 @@ sidebar_title: Selective Testing
 sidebar_order: 4952
 sidebar_section: features
 description: Render only selected SnapshotPreviews groups in CI while keeping Sentry aware of the full tracked image set.
+early_access: true
 ---
-
-<Include name="feature-available-for-user-group-early-adopter" />
 
 Selective testing is an advanced workflow for large suites where rendering every preview on every pull request is too expensive. On each run you render and upload only a subset of your previews, and Sentry diffs that subset against the baseline.
 
@@ -15,7 +14,7 @@ There are two approaches:
 - **Approach A — upload just the subset.** Simplest to set up. Any image you don't upload counts as skipped, so Sentry can't detect deletions or renames.
 - **Approach B — upload the subset plus the full list of tracked image names.** Sentry can then tell apart snapshots that were rendered, skipped, or removed/renamed.
 
-Either way, you first organize snapshots into selectable groups and upload a baseline build, covered in the next two sections. The approaches differ only in the final render-and-upload step.
+For both implementations, you first organize snapshots into groups and upload a baseline build, covered in the next two sections. The approaches differ only in the final render-and-upload step.
 
 ## Organize Snapshots Into Selectable Groups
 
--- a/docs/platforms/apple/common/snapshots/selective-testing.mdx
+++ b/docs/platforms/apple/common/snapshots/selective-testing.mdx
@@ -1,0 +1,148 @@
+---
+title: Selective Testing with SnapshotPreviews
+sidebar_title: Selective Testing
+sidebar_order: 4952
+sidebar_section: features
+description: Render only selected SnapshotPreviews groups in CI while keeping Sentry aware of the full tracked image set.
+early_access: true
+---
+
+<Include name="feature-available-for-user-group-early-adopter" />
+
+Selective testing is an advanced SnapshotPreviews workflow for large suites. Use it when rendering every preview on every pull request is too expensive, but you still want Sentry to tell apart:
+
+- snapshots rendered in this run;
+- snapshots skipped because their group wasn't selected;
+- snapshots intentionally removed from the tracked set.
+
+The idea is to separate **name generation** from **image rendering**. First, write the complete list of image names for the full tracked suite. Then render only the groups selected for this CI run and upload those images with the complete image-name file.
+
+## Organize snapshots into selectable groups
+
+Create one `SnapshotTest` subclass per group CI can select. Inclusion filters, such as `snapshotPreviewModules()` or `snapshotPreviews()`, define what a group renders. Exclusion filters, such as `excludedSnapshotPreviews()`, remove snapshots from the tracked set entirely.
+
+```swift
+import SnapshottingTests
+
+final class ModuleASnapshots: SnapshotTest {
+  override class func snapshotPreviewModules() -> [String]? {
+    ["ModuleA"]
+  }
+
+  override class func excludedSnapshotPreviews() -> [String]? {
+    ["ModuleA/LoadingView.swift:Loading"]
+  }
+}
+
+final class ModuleBCSnapshots: SnapshotTest {
+  override class func snapshotPreviewModules() -> [String]? {
+    ["ModuleB", "ModuleC"]
+  }
+}
+```
+
+Together, these groups cover every tracked snapshot in Module A, Module B, and Module C. `ModuleA/LoadingView.swift:Loading` is excluded, so it is not part of the tracked set. If that snapshot existed in the base build, Sentry can report it as removed instead of skipped.
+
+Use class-level `-only-testing:<test-target>/<class-name>` selectors. Do not shard by generated preview method selectors; those selectors are runtime-generated implementation details.
+
+## Step 1: Upload a complete baseline build
+
+On your base branch, render and upload the full snapshot suite. This is a normal, complete upload. Do not pass an image-name file on the baseline upload.
+
+```bash
+TEST_RUNNER_SNAPSHOTS_EXPORT_DIR="/private/tmp/base_snapshots/iphone" \
+xcodebuild test \
+  -project MyApp.xcodeproj \
+  -scheme MyApp \
+  -destination 'platform=iOS Simulator,name=iPhone 17 Pro Max'
+
+sentry-cli build snapshots "/private/tmp/base_snapshots/iphone" \
+  --app-id com.example.MyApp
+```
+
+## Step 2: Build the pull request test bundle
+
+On the pull request, build the test bundle once and reuse it for name generation and rendering.
+
+```bash
+xcodebuild build-for-testing \
+  -project MyApp.xcodeproj \
+  -scheme MyApp \
+  -destination 'platform=iOS Simulator,name=iPhone 17 Pro Max' \
+  -derivedDataPath /private/tmp/snapshotpreviews_root/DerivedData
+```
+
+Use the `.xctestrun` file generated under the derived data directory for the following `test-without-building` commands:
+
+```bash
+/private/tmp/snapshotpreviews_root/DerivedData/Build/Products/<generated>.xctestrun
+```
+
+## Step 3: Generate image names for every selectable group
+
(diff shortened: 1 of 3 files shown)


docs/platforms/apple/common/snapshots/selective-testing.mdx
```diff
@@ -0,0 +1,148 @@
...
+description: Render only selected SnapshotPreviews groups in CI while keeping Sentry aware of the full tracked image set.
+early_access: true
+---
+
+<Include name="feature-available-for-user-group-early-adopter" />
+
+Selective testing is an advanced SnapshotPreviews workflow for large suites. Use it when rendering every preview on every pull request is too expensive, but you still want Sentry to tell apart:
+
+- snapshots rendered in this run;
+- snapshots skipped because their group wasn't selected;
```
NicoHinderling: nit: remove the semi colons on these two lines

docs/platforms/apple/common/snapshots/selective-testing.mdx
```diff
@@ -0,0 +1,148 @@
...
+sidebar_title: Selective Testing
+sidebar_order: 4952
+sidebar_section: features
+description: Render only selected SnapshotPreviews groups in CI while keeping Sentry aware of the full tracked image set.
+early_access: true
+---
+
+<Include name="feature-available-for-user-group-early-adopter" />
+
+Selective testing is an advanced SnapshotPreviews workflow for large suites. Use it when rendering every preview on every pull request is too expensive, but you still want Sentry to tell apart:
```
NicoHinderling: nit: I'd put the word "subset" or something somewhere in these two sentences. Just make it dumb obvious  that only a subset of the snapshots are being generated

docs/platforms/apple/common/snapshots/selective-testing.mdx
```diff
@@ -0,0 +1,148 @@
...
+
+- snapshots rendered in this run;
+- snapshots skipped because their group wasn't selected;
+- snapshots intentionally removed from the tracked set.
+
+The idea is to separate **name generation** from **image rendering**. First, write the complete list of image names for the full tracked suite. Then render only the groups selected for this CI run and upload those images with the complete image-name file.
```
NicoHinderling: Important point to make: You can totally use selective testing just by uploading a subset of images **without the list of image names**. 

We just won't be able to flag any removals or renames. Including the `--all-image-file-names` field is an optional add-on to get visibility of the removals and renames.

I think the current wording presents the image file names as a requirement

docs/platforms/apple/common/snapshots/selective-testing.mdx
```diff
@@ -0,0 +1,148 @@
...
+    ["ModuleA/LoadingView.swift:Loading"]
+  }
+}
+
+final class ModuleBCSnapshots: SnapshotTest {
+  override class func snapshotPreviewModules() -> [String]? {
+    ["ModuleB", "ModuleC"]
+  }
+}
+```
```
NicoHinderling, line 37: nice!!

docs/platforms/apple/common/snapshots/selective-testing.mdx
```diff
@@ -0,0 +1,148 @@
...
+
+Together, these groups cover every tracked snapshot in Module A, Module B, and Module C. `ModuleA/LoadingView.swift:Loading` is excluded, so it is not part of the tracked set. If that snapshot existed in the base build, Sentry can report it as removed instead of skipped.
+
+Use class-level `-only-testing:<test-target>/<class-name>` selectors. Do not shard by generated preview method selectors; those selectors are runtime-generated implementation details.
+
+## Step 1: Upload a complete baseline build
+
+On your base branch, render and upload the full snapshot suite. This is a normal, complete upload. Do not pass an image-name file on the baseline upload.
```
NicoHinderling: `Do not pass an image-name file on the baseline upload.`

I feel like this is another example of why we should make it clear near the top that the image name generation stuff is optional ONLY for the selective testing case.

If so, I feel like it might be implicit here that it shouldn't be included for full (non selective) uploads 🤔 

idk, curious your thoughts

docs/platforms/apple/common/snapshots/selective-testing.mdx
```diff
@@ -0,0 +1,148 @@
...
+
+```bash
+TEST_RUNNER_SNAPSHOTS_EXPORT_DIR="/private/tmp/base_snapshots/iphone" \
+xcodebuild test \
```
NicoHinderling, line 45: should this also be `xcodebuild build-for-testing`?

docs/platforms/apple/common/snapshots/selective-testing.mdx
```diff
@@ -0,0 +1,148 @@
...
+  -derivedDataPath /private/tmp/snapshotpreviews_root/DerivedData
+```
+
+Use the `.xctestrun` file generated under the derived data directory for the following `test-without-building` commands:
+
+```bash
+/private/tmp/snapshotpreviews_root/DerivedData/Build/Products/<generated>.xctestrun
+```
+
+## Step 3: Generate image names for every selectable group
```
NicoHinderling: again, would flag this as an optional section/step

also perhaps this step could be after the selected snapshot generation step?

docs/platforms/apple/common/snapshots/selective-testing.mdx
```diff
@@ -0,0 +1,148 @@
...
+TEST_RUNNER_SNAPSHOTS_EXPORT_DIR="/private/tmp/head_snapshots/iphone" \
+xcodebuild test-without-building \
+  -xctestrun /private/tmp/snapshotpreviews_root/DerivedData/Build/Products/<generated>.xctestrun \
+  -destination 'platform=iOS Simulator,name=iPhone 17 Pro Max' \
+  -only-testing:MyAppTests/ModuleASnapshots
+```
+
+## Step 5: Upload the selective build
+
+Upload the rendered image directory with the complete image-name file. Passing `--all-image-file-names-file` tells Sentry this is a selective upload, so listed images that were not uploaded are treated as skipped rather than removed.
```
NicoHinderling: `Passing `--all-image-file-names-file` tells Sentry this is a selective upload`

`--selective` tells sentry it's a selective build, but i then made it so that including `--all-image-file-names-file` implicitly implies that it's selective so that you can just do that rather than `--selective --all-image-file-names-file`

might be worth making sure this is clear here 🤷

NicoHinderling: nit: 

"so listed images that were not uploaded are treated as skipped rather than removed."

-> 

"so listed images that were not uploaded are correctly identified as skipped rather than removed."

docs/platforms/apple/common/snapshots/selective-testing.mdx
cameroncooke: No, the only reason for the two stage, `build-for-testing`  and `test-without-building` is be able to run the head all-image-file-names filename generation and then run the the selective tests for .png generation from the same existing test build artefact.

docs/platforms/apple/common/snapshots/selective-testing.mdx
cameroncooke: I will do, the reason I didn't treat it as optional was because this entire page is specifically about selective-testing when using SnapshotPreviews. If the user chooses not to upload an image filenames .txt file then they don't need to do anything different with this library, the exercise would then just become add a flag when uploading images which has nothing to do with SnapshotPreviews docs.

If the user is using SnapshotPreviews which is likely if they are reading the docs, then I can't see a world where the user would prefer to use `--selective` over the library automatically generating the image filenames .txt file and pass the `--all-image-file-names-file` flag. It would appear to me `--selective` is more a connivence where generating the image filenames .txt file is non-trivial.

docs/platforms/apple/common/snapshots/selective-testing.mdx
```diff
@@ -0,0 +1,260 @@
+---
+title: Selective Testing with SnapshotPreviews
+sidebar_title: Selective Testing
+sidebar_order: 4952
+sidebar_section: features
+description: Render only selected SnapshotPreviews groups in CI while keeping Sentry aware of the full tracked image set.
+---
+
+<Include name="feature-available-for-user-group-early-adopter" />
```
NicoHinderling: I think this can be beta instead of early adopter

docs/platforms/apple/common/snapshots/selective-testing.mdx
```diff
@@ -0,0 +1,260 @@
...
+
+Selective testing is an advanced workflow for large suites where rendering every preview on every pull request is too expensive. On each run you render and upload only a subset of your previews, and Sentry diffs that subset against the baseline.
+
+There are two approaches:
+
+- **Approach A — upload just the subset.** Simplest to set up. Any image you don't upload counts as skipped, so Sentry can't detect deletions or renames.
+- **Approach B — upload the subset plus the full list of tracked image names.** Sentry can then tell apart snapshots that were rendered, skipped, or removed/renamed.
+
+Either way, you first organize snapshots into selectable groups and upload a baseline build, covered in the next two sections. The approaches differ only in the final render-and-upload step.
```
NicoHinderling: u-nit: `Either way` -> `For both implementations`

NicoHinderling: u-nit: `selectable groups` -> `groups`

NicoHinderling (approved):

docs/platforms/apple/common/snapshots/selective-testing.mdx
cameroncooke: Good catch. The feature is in the Early Adopter program rather than open beta, and there's no generic beta banner, so I removed the banner and kept the `early_access` sidebar tag instead. Done in 575bbded2.

docs/platforms/apple/common/snapshots/selective-testing.mdx
cameroncooke: Done — changed to "For both implementations" in 575bbded2.

cameroncooke: Done — simplified to "groups" in 575bbded2.

PR #18287, opened 2026-06-05
fix(md-exports): Return 200 for agent not-found pages so the body survives

sergical:
The `.md` catch-all (`app/md-exports/[...path]/route.ts`) serves a Markdown "page not found" helper that links to real nearby pages so AI agents can recover from a wrong URL. It was returned as HTTP 404 — but agent fetchers, including Claude Code's WebFetch (User-Agent `Claude-User`), **discard the response body on any non-2xx status**. The agent received "0 bytes" instead of the recovery content, so a guessed URL became a dead end even though the server had useful Markdown to hand back.

This returns the helper with HTTP `200` so the body reaches the agent, and adds a metric so we can see which agents invent which URLs.

**Behavior change**

Before/after for a non-existent page fetched as Markdown (e.g. `/platforms/javascript/guides/vercel-edge.md`):

```text
Before:  404  text/markdown  (agent fetchers drop the body -> "0 bytes")
After:   200  text/markdown  X-Robots-Tag: noindex  X-Sentry-Docs-Not-Found: 1
```

The body is unchanged — same "Page Not Found" heading and the same list of real sibling pages.

**Not-found is still detectable**

`X-Robots-Tag: noindex` keeps crawlers from treating the soft-200 as real content, and `X-Sentry-Docs-Not-Found: 1` lets monitoring distinguish these soft misses from genuine hits even though the status is now 200. This only affects the agent-facing `.md` route; the HTML route still returns a true 404.

**Metric: which agents make up which URLs**

Adds `DocMetrics.mdExportNotFound` → emits `docs.md_export.not_found` with attributes:

- `requested_path` — the **full** invented path (the signal we want). Kept in full because trace metrics are EAP-backed (high cardinality is fine) and an agent-invented docs path isn't PII.
- `agent` — normalized agent name (`claude`, `gptbot`, …) via the same `matchPattern(ua, AI_AGENT_PATTERN)` convention used in `tracesSampler.ts`; `other` when unmatched.
- `has_suggestions` — whether the soft-404 had sibling pages to offer.

## IS YOUR CHANGE URGENT?

- [ ] Other deadline: 

### DIFF
--- a/app/md-exports/[...path]/route.test.ts
+++ b/app/md-exports/[...path]/route.test.ts
@@ -61,9 +61,18 @@ describe('md-exports 404 catch-all route', () => {
     return GET(request, {params: Promise.resolve({path: pathSegments})});
   }
 
-  it('returns 404 status', async () => {
+  // Intentionally 200, not 404: agent fetchers (e.g. Claude Code's WebFetch) drop the
+  // response body on non-2xx, which would strip the helpful not-found content. The
+  // not-found signal lives in the body and the X-Sentry-Docs-Not-Found header instead.
+  it('returns 200 status so agent fetchers keep the body', async () => {
     const res = await callRoute(['platforms', 'javascript', 'ai-monitoring.md']);
-    expect(res.status).toBe(404);
+    expect(res.status).toBe(200);
+  });
+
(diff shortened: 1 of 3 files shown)


sfanahata (approved):
👏  This is slick! 🚀

PR #18324, opened 2026-06-09
docs(tanstackstart-react): Initialize Sentry from a dedicated client init file

nicohrubec:
Moves the client `Sentry.init()` into a dedicated `src/instrument.client.ts` file that is imported first in the client entry point. Importing it first ensures Sentry initializes before any other import's side effects (and before hydration).

SDK PR: getsentry/sentry-javascript#21397


andreiborza (approved):
Let's also update onboarding, lgtm!

PR #18328, opened 2026-06-09
docs(azure-devops): correct auth mechanism from PAT to OAuth

souredoutlook:
The install note incorrectly states the integration uses personal access tokens. It actually authenticates via OAuth — the bearer token stored at install time is the OAuth access token from whoever authorized the integration, not a PAT.

The attribution behavior (actions attributed to the authorizing user) is unchanged; only the mechanism description is corrected.

---
[View Session in Sentry](https://sentry.sentry.io/traces/?project=4510944073809921&query=gen_ai.conversation.id%3A%22slack%3ACD4NYFD34%3A1780965528.550929%22)


coolguyzone (approved):
Thanks for catching this and updating 🏄‍♂️

PR #18354, opened 2026-06-10
docs(seer): Fix repository settings links for both Seer billing models

mrduncan:
The Autofix and Code Review setup steps linked to a Seer SCM settings page that is not a real route, and the Code Review step used a broken `#autoEnableCodeReview` anchor. Point repository setup to the canonical `/settings/repos/` page (ungated, works for legacy and seat-based Seer) and the default toggle to the `/settings/seer/repos/defaults/` drawer where the field actually lives.


coolguyzone (approved):
Thanks for the fix!

PR #18372, opened 2026-06-11
feat(develop-docs): Link spec status badge to lifecycle page

chargome:
Adds an info (ⓘ) icon next to the page-level **Status** badge in `SpecMeta`, linking to the spec-lifecycle page.


dingsdax (approved):
nice 👏

PR #18401, opened 2026-06-12
fix(seer): correct pr_created provider value in webhook docs

joseph-sentry:
The `pr_created` (and the new `iteration_completed`) Seer webhook payloads emit `"provider": "unknown"`, not `"github"`.

The PR-state object that backs the payload (`RepoPRState`) carries no provider field, so the broadcaster hardcodes `"unknown"`. The docs example claimed `"github"`, which never matches what is actually delivered.

This updates the `pr_created` example to reflect the real value.

## PRE-MERGE CHECKLIST
- [x] None: Not urgent


coolguyzone (approved):
🤙

PR #18456, opened 2026-06-17
docs(python): Draft New Spans and Migration Guide pages

inventarSarah:
## DESCRIBE YOUR PR
This PR adds the initial docs for Python span streaming: a New Spans page (intro to stream mode) and a Migration Guide (how to update existing custom instrumentation).

**Context**
For Python, the new Span API and stream mode ship together, so migrating to stream mode means migrating to the new Span API at the same time abd there's no way to use the new API while staying in transaction mode. 
Because of that, putting the migration steps into the New Spans was way too much content and I split that content out into a new Migration guide. This guide is basically a human-readable version of the migration content from the engineering skill file.

**Considerations**
- Content here may still change once I go through the existing Tracing, Option, Filtering, etc. pages
- The Migration Guide is currently nested under New Spans in the nav. But it could also live under the main "Migration Guides" section instead. Let me know if it makes sense to move the guide.
- I reused content from the JS New Spans page -> I will create includes for these later when the content is fixed :)

## IS YOUR CHANGE URGENT?  

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST

## LEGAL BOILERPLATE

## EXTRA RESOURCES

### DIFF
--- a/docs/platforms/python/tracing/new-spans/index.mdx
+++ b/docs/platforms/python/tracing/new-spans/index.mdx
@@ -95,7 +95,7 @@ Each flush sends only the spans accumulated since the last flush, grouped into e
 
 </Expandable>
 
-## Manual Instrumenation (Optional)
+## Manual Instrumentation (Optional)
 
 ### Start a Span
 
--- a/docs/platforms/python/tracing/new-spans/index.mdx
+++ b/docs/platforms/python/tracing/new-spans/index.mdx
@@ -13,7 +13,7 @@ Stream mode changes this by sending spans to Sentry in batches as they finish. S
 - **No 1,000-span limit.** In transaction mode, transactions are capped at 1,000 spans. Stream mode has no upper limit since spans are sent in batches. This is especially beneficial for complex generative AI pipelines that easily exceed standard span limits.
 - **Lower memory usage.** Spans are flushed periodically and don't need to be held in memory until the root span ends. This is especially useful for long-running processes like queue consumers or cron jobs.
 - **Faster visibility.** Span data arrives in Sentry as your application runs, instead of only after the entire operation completes.
-- **No data loss from crashes.** If your process terminates unexpectedly, spans that were already flushed are preserved. In transaction mode, a crash before the root span ends means all span data is lost.
+- **Spans are emitted in more failure scenarios.** If your process terminates unexpectedly, spans that were already flushed are preserved. In transaction mode, a crash before the transaction ends means all span data is lost.
 
 </Expandable>
 
@@ -142,7 +142,8 @@ with sentry_sdk.traces.start_span(name="parent-operation"):
 <SplitSection>
 <SplitSectionText>
 
-To start a service span (the equivalent of a transaction), set `parent_span=None`:
+A span is automatically promoted to a service span (the equivalent of a transaction) if no other parent is currently active.
+If you want to force a new service span, regardless of whether it has a parent span, set `parent_span=None`:
 
 </SplitSectionText>
 <SplitSectionCode>
--- a/docs/platforms/python/tracing/new-spans/index.mdx
+++ b/docs/platforms/python/tracing/new-spans/index.mdx
@@ -10,7 +10,7 @@ Stream mode changes this by sending spans to Sentry in batches as they finish. S
 
 <Expandable title="Why use stream mode?">
 
-- **No 1,000-span limit.** In transaction mode, transactions are capped at 1,000 spans. Stream mode has no upper limit since spans are sent in batches.
+- **No 1,000-span limit.** In transaction mode, transactions are capped at 1,000 spans. Stream mode has no upper limit since spans are sent in batches. This is especially beneficial for complex generative AI pipelines that easily exceed standard span limits.
 - **Lower memory usage.** Spans are flushed periodically and don't need to be held in memory until the root span ends. This is especially useful for long-running processes like queue consumers or cron jobs.
 - **Faster visibility.** Span data arrives in Sentry as your application runs, instead of only after the entire operation completes.
 - **No data loss from crashes.** If your process terminates unexpectedly, spans that were already flushed are preserved. In transaction mode, a crash before the root span ends means all span data is lost.
@@ -181,6 +181,12 @@ For more details on span creation, see <PlatformLink to="/tracing/instrumentatio
 ### Add Span Attributes
 
 Attach structured metadata to spans using `attributes`, which can be `str`, `int`, `float`, or `bool`, as well as arrays of these types.
+
+<Alert level="info">
+
+Sentry automatically sets several standard attributes on spans. To avoid accidentally overwriting these, refer to our <a href="https://getsentry.github.io/sentry-conventions/attributes/">Sentry Attribute Conventions</a>.
+
+</Alert>
 
 <SplitLayout>
 <SplitSection>
--- a/docs/platforms/python/tracing/new-spans/index.mdx
+++ b/docs/platforms/python/tracing/new-spans/index.mdx
@@ -1,0 +1,408 @@
+---
+title: New Spans
+description: "Learn how to use stream mode to send spans to Sentry as they finish, removing the 1,000-span limit and making trace data visible sooner."
+sidebar_order: 45
+new: true
+---
+
+By default, the Sentry Python SDK collects all spans in memory and sends them to Sentry as a single transaction once the root span ends. This is called transaction mode.
+Stream mode changes this by sending spans to Sentry in batches as they finish. Service spans, which represent a service's entry point, replace transactions as the main grouping for each service.
+
+<Expandable title="Why use stream mode?">
+
+- **No 1,000-span limit.** In transaction mode, transactions are capped at 1,000 spans. Stream mode has no upper limit since spans are sent in batches.
+- **Lower memory usage.** Spans are flushed periodically and don't need to be held in memory until the root span ends. This is especially useful for long-running processes like queue consumers or cron jobs.
+- **Faster visibility.** Span data arrives in Sentry as your application runs, instead of only after the entire operation completes.
+- **No data loss from crashes.** If your process terminates unexpectedly, spans that were already flushed are preserved. In transaction mode, a crash before the root span ends means all span data is lost.
+
+</Expandable>
+
+You can find the following span types mentioned throughout this page:
+
+- **Root span**: The topmost span in a trace. It has no parent span and is always a service span.
+- **Service span**: A parent-level span at the entry of a service. In transaction mode, this is called a transaction.
+- **Child span**: Any span nested under a parent span within the same trace.
+
+This graph shows how these span types relate to each other within a trace:
+
+```
+Trace
+│
+└── Root span [service A]
+     ├── Child span
+     │    └── Child span
+     └── Service span [service B]
+          ├── Child span
+          └── Child span
+```
+
+<Alert level="warning" title="Migrating from transaction mode?">
+
+Stream mode requires the new Span API, so migrating to stream mode and migrating to the new Span API are the same step. If you have existing custom instrumentation, see the <PlatformLink to="/tracing/new-spans/migration-guide/">Migration Guide</PlatformLink> for a full list of changes.
+
+</Alert>
+
+## Prerequisites
+
+You need:
+
+- <PlatformLink to="/tracing/#configure">Tracing configured</PlatformLink> in
+  your app
+- Sentry SDK `>=2.62.0`
+
+## Enable Stream Mode
+
+<SplitLayout>
+<SplitSection>
+<SplitSectionText>
+
+Opt in by adding `trace_lifecycle` to the `_experiments` option when initializing the SDK:
+
+</SplitSectionText>
+<SplitSectionCode>
+
+```python
+import sentry_sdk
+
+sentry_sdk.init(
+    dsn="___PUBLIC_DSN___",
+    traces_sample_rate=1.0,
+    _experiments={
+        # enables stream mode
+        "trace_lifecycle": "stream",
+    },
+)
+
+```
+
+</SplitSectionCode>
+</SplitSection>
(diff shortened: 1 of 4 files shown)


docs/platforms/python/tracing/streamed-spans/index.mdx
```diff
@@ -0,0 +1,408 @@
+---
+title: New Spans
+description: "Learn how to use stream mode to send spans to Sentry as they finish, removing the 1,000-span limit and making trace data visible sooner."
+sidebar_order: 45
+new: true
+---
+
+By default, the Sentry Python SDK collects all spans in memory and sends them to Sentry as a single transaction once the root span ends. This is called transaction mode.
+Stream mode changes this by sending spans to Sentry in batches as they finish. Service spans, which represent a service's entry point, replace transactions as the main grouping for each service.
```
ericapisani, line 9: > Service spans, which represent a service's entry point, replace transactions as the main grouping for each service.

This is my first time seeing the term "service" used for what we've been calling "segments". I'm not sure if this is a term that we plan on using for user-facing comms going forward, but wanted to highlight the discrepency.

I'm also not sure that the main grouping of spans is always a service's entry point. If we refer to transactions in the same way, then leave as-is, but would want to double check with @sentrivana or @alexander-alderman-webb

docs/platforms/python/tracing/streamed-spans/index.mdx
```diff
@@ -0,0 +1,408 @@
...
+
+<Expandable title="Why use stream mode?">
+
+- **No 1,000-span limit.** In transaction mode, transactions are capped at 1,000 spans. Stream mode has no upper limit since spans are sent in batches.
+- **Lower memory usage.** Spans are flushed periodically and don't need to be held in memory until the root span ends. This is especially useful for long-running processes like queue consumers or cron jobs.
```
ericapisani, line 14: I think it could be worth calling out the benefits for AI monitoring since generative AI spans emitted by our SDKs are one of the main beneficiaries of this change

docs/platforms/python/tracing/new-spans/index.mdx
```diff
@@ -0,0 +1,408 @@
...
+
+</SplitSectionText>
+<SplitSectionCode>
+
+```python
+import sentry_sdk
+
+with sentry_sdk.traces.start_span(name="task-name", parent_span=None) as span:
+    do_work()
+```
```
ericapisani: This example is why I think referring to what are currently called "segment spans" as "service entry points" [above](https://github.com/getsentry/sentry-docs/pull/18456/changes#r3429192698) will be a bit confusing since this code snippet could be placed anywhere, not just at an "entry point" in a web framework/service.

docs/platforms/python/tracing/streamed-spans/index.mdx
```diff
@@ -0,0 +1,408 @@
...
+        "http.route": "/api/users",
+        "user.id": "user-42",
+    })
+```
+
+</SplitSectionCode>
+</SplitSection>
+</SplitLayout>
+
+Find more examples in our <PlatformLink to="/tracing/span-metrics/">Sending Span Metrics</PlatformLink> documentation.
```
ericapisani, line 245: I wonder if it's worth mentioning our [Sentry Convention/Attribute docs](https://getsentry.github.io/sentry-conventions/attributes/) to users here. 

These are attributes that we may set within the SDK that users will want to be aware of so they don't accidentally overwrite them when setting their own attributes (unless that's what they want 😄 )

ericapisani (commented):
Looking good so far! 

I have questions about the use of the term "service span" vs "segment span" as the latter is what we've been using thus far.

Additionally, should we keep the term "service", referring to these spans as "service entry points" could be a point of confusion because I can see this being mixed up with entry points for a request in the context of a web framework, even though these could be placed/started anywhere (which one of the examples in the documentation show).

docs/platforms/python/tracing/streamed-spans/index.mdx
cleptric: I thought I brought this up in some sync (😇), but service span is the canonical, user facing term of these kind of spans.

docs/platforms/python/tracing/new-spans/migration-guide.mdx
```diff
@@ -0,0 +1,240 @@
...
+- scope = sentry_sdk.get_current_scope()
+- transaction = scope.transaction
++ service_span = sentry_sdk.traces.get_current_span()._segment
+
+# Getting the service span from a child span
+- transaction = span.containing_transaction
++ service_span = span._segment
+```
+
+`_segment` returns the service span of the current trace (the equivalent of what used to be the transaction). It is a private API — prefer restructuring the code to avoid needing the service span where possible.
```
sfanahata: This part is confusing. A private API is returning service span data? Or will this data become unavailable. Let's reword this to make it clearer, and remove reference to private APIs if we can, since customers won't be able to use it.

docs/platforms/python/tracing/new-spans/migration-guide.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+# Getting the current span via the scope
+- scope = sentry_sdk.get_current_scope()
+- current_span = scope.span
++ current_span = sentry_sdk.traces.get_current_span()
+
+# Getting the service span (formerly the transaction)
+- scope = sentry_sdk.get_current_scope()
+- transaction = scope.transaction
++ service_span = sentry_sdk.traces.get_current_span()._segment
```
alexander-alderman-webb: There should be **no references** to underscored things from our SDK (single leading underscore).
These are private, which means we can change them whenver and however we want. If users are relying on `_` prefixed stuff, it's a world of pain for them in the future.

alexander-alderman-webb: That doesn't include `_experiments`, which is fine to advertise to users.

docs/platforms/python/tracing/new-spans/migration-guide.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+```python diff
+- ctx = span.get_trace_context()
+- trace_id = ctx["trace_id"]
+- span_id = ctx["span_id"]
++ trace_id = span.trace_id
++ span_id = span.span_id
+```
+
+If you genuinely need the full trace context dict, for example to pass it to an external system, use the private `span._get_trace_context()` method instead. Prefer the direct properties where possible.
```
alexander-alderman-webb: Same here, please don't advertise `_get_trace_context()`.

For many of the wacky things possible in transaction land there is no alternative.

docs/platforms/python/tracing/new-spans/migration-guide.mdx
```diff
@@ -0,0 +1,240 @@
...
++           "/health",
++           re.compile(r"^GET /api/v1/internal"),
++           {"attributes": {"service.id": "15def9a"}},
++       ],
++       "before_send_span": my_span_processor,
++   },
+)
+```
+
+Both `ignore_spans` and `before_send_span` only have access to the span name and attributes set at creation time — not attributes added later in the span's lifetime, like an HTTP status code set after the request completes. If your `before_send_transaction` logic depended on that kind of late-set data, it can't be replicated in stream mode. Consider server-side filtering with Sentry inbound data filters or Relay rules instead.
```
sfanahata: "Consider server-side filtering with Sentry inbound data filters or Relay rules instead." <-- let's link to these, if we have documentation on them.

docs/platforms/python/tracing/new-spans/migration-guide.mdx
inventarSarah: removed this paragraph

docs/platforms/python/tracing/streamed-spans/index.mdx
inventarSarah: 👍Added the link in a callout at the top of the section

docs/platforms/python/tracing/new-spans/index.mdx
```diff
@@ -0,0 +1,414 @@
...
+Stream mode changes this by sending spans to Sentry in batches as they finish. Service spans, which represent a service's entry point, replace transactions as the main grouping for each service.
+
+<Expandable title="Why use stream mode?">
+
+- **No 1,000-span limit.** In transaction mode, transactions are capped at 1,000 spans. Stream mode has no upper limit since spans are sent in batches. This is especially beneficial for complex generative AI pipelines that easily exceed standard span limits.
+- **Lower memory usage.** Spans are flushed periodically and don't need to be held in memory until the root span ends. This is especially useful for long-running processes like queue consumers or cron jobs.
+- **Faster visibility.** Span data arrives in Sentry as your application runs, instead of only after the entire operation completes.
+- **No data loss from crashes.** If your process terminates unexpectedly, spans that were already flushed are preserved. In transaction mode, a crash before the root span ends means all span data is lost.
```
alexander-alderman-webb: A root span starts the trace. This is a different concept to a transaction / segment span.

Also, not all application crashes result in data loss.
It'd be good to rephrase to say that spans are emitted (not lost) in more cases if span streaming is enabled.

```suggestion
- **No data loss from crashes.** If your process terminates unexpectedly, spans that were already flushed are preserved. In transaction mode, a crash before the transaction ends means all span data is lost.
```

docs/platforms/python/tracing/new-spans/index.mdx
```diff
@@ -0,0 +1,414 @@
...
+    with sentry_sdk.traces.start_span(name="child-step-2"):
+        step_two()
+```
+
+</SplitSectionCode>
+</SplitSection>
+<SplitSection>
+<SplitSectionText>
+
+To start a service span (the equivalent of a transaction), set `parent_span=None`:
```
alexander-alderman-webb: I'd note that even if `parent_span` is not provided, spans can be promoted to service spans.

docs/platforms/python/tracing/streamed-spans/migration-guide.mdx
```diff
@@ -0,0 +1,228 @@
...
+def traces_sampler(sampling_context):
+    if sampling_context["span_context"]["name"] in IGNORED_SPAN_NAMES:
+        return 0.0
+    return 1.0
+
+sentry_sdk.init(
+    traces_sampler=traces_sampler,
+    _experiments={"trace_lifecycle": "stream"},
+)
+```
```
alexander-alderman-webb, line 197: It's good practice for code snippets to be self contained and include all required imports.
Could you make sure the snippets all have `import sentry_sdk` if relevant?

```suggestion
```python
import sentry_sdk

def traces_sampler(sampling_context):
    if sampling_context["span_context"]["name"] in IGNORED_SPAN_NAMES:
        return 0.0
    return 1.0

sentry_sdk.init(
    traces_sampler=traces_sampler,
    _experiments={"trace_lifecycle": "stream"},
)
```
```

**Review by alexander-alderman-webb — APPROVED** [review]
Looks good overall! See comments

File: docs/platforms/python/tracing/new-spans/index.mdx (hunk 10 above)
Comment (inventarSarah) [reply docs/platforms/python/tracing/new-spans/index.mdx]: 👍Rewrote the sentence

File: docs/platforms/python/tracing/streamed-spans/migration-guide.mdx (hunk 11 above)
Comment (inventarSarah) [reply docs/platforms/python/tracing/streamed-spans/migration-guide.mdx:197]: 👍added missing imports

File: docs/platforms/python/tracing/new-spans/index.mdx (hunk 9 above)
Comment (inventarSarah) [reply docs/platforms/python/tracing/new-spans/index.mdx]: 👍updated this list item

File: docs/platforms/python/tracing/streamed-spans/index.mdx (hunk 2 above)
Comment (inventarSarah) [reply docs/platforms/python/tracing/streamed-spans/index.mdx:14]: sounds good -- added to the 1k span limit bullet point

File: docs/platforms/python/tracing/streamed-spans/migration-guide.mdx [hunk 12]
```diff
@@ -0,0 +1,238 @@
...
+
+```python diff
+import sentry_sdk
+
+- with sentry_sdk.start_span(
+-     name="handle request",
+-     custom_sampling_context={"asgi_scope": asgi_scope},
+- ):
+-     ...
++ sentry_sdk.Scope.set_custom_sampling_context({"asgi_scope": asgi_scope})
```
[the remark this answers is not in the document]
inventarSarah: See comment on the same issue here: https://github.com/getsentry/sentry-docs/pull/18533#discussion_r3467073310

docs/platforms/python/tracing/streamed-spans/migration-guide.mdx
```diff
@@ -0,0 +1,238 @@
...
+
+# Starting what used to be a transaction: pass parent_span=None to force a service span
+- with sentry_sdk.start_transaction(name="flow.checkout") as transaction:
++ with sentry_sdk.traces.start_span(name="flow.checkout", parent_span=None) as span:
+      ...
+
+# Starting a child span: just start a span while the parent is active
+- with parent.start_child(op="db", description="SELECT") as child:
++ with sentry_sdk.traces.start_span(name="SELECT", attributes={"sentry.op": "db"}):
+      ...
```
sentrivana, line 50: Might be worth mentioning that if this automatic parenting behavior is not desired, e.g. because the user wants to parent to a span that's currently not active, they can use the `parent_span` argument here too

```suggestion
# Starting a child span: just start a span while the parent is active
- with parent.start_child(op="db", description="SELECT") as child:
+ with sentry_sdk.traces.start_span(name="SELECT", attributes={"sentry.op": "db"}):
      ...

# Starting a child span: if the span's parent should be a span that's currently not active, you can provide it explicitly
- with parent.start_child(op="db", description="SELECT") as child:
+ with sentry_sdk.traces.start_span(name="SELECT", attributes={"sentry.op": "db"}, parent_span=parent):
      ...
```

sentrivana (approved):

PR #18479, opened 2026-06-18
docs(ssl): add Google Trust Services (pki.goog) root CAs to certificate pinning list

BojanOro:
Adds Google Trust Services (pki.goog) root CAs to the certificate pinning documentation.

## Why

Sentry is deploying the US2 region using GCP-managed TLS certificates. GCP can issue certs from either `letsencrypt.org` or `pki.goog`. The GTS roots were missing from our pinning docs, which could break customers who pin certificates and end up being served a pki.goog cert.

All four GTS root CAs (R1–R4, covering both RSA and EC) are included. Fingerprints verified directly from the live `https://pki.goog/roots.pem` bundle via openssl.

Also updates the introductory sentence in the Certificate Pinning section to mention Google Trust Services alongside Digicert and Let's Encrypt.

**GlobalSign ECC Root CA - R4 is not included.**
GlobalSign ECC Root CA - R4 shows up in pki.goog/roots.pem because that bundle contains every root GTS has a trust relationship with — 20 certs total, including GlobalSign, DigiCert, GoDaddy, COMODO, USERTrust, etc. It's not a root GTS issues from.

GCP-managed certificates chain to GTS Root R1–R4 only (R1/R2 for RSA, R3/R4 for EC). GlobalSign R4 was historically used for cross-signing GTS intermediates for broader device compatibility, but modern GTS-issued certs go straight to their own roots.

## Context

https://sentry.slack.com/archives/C1G612XHC/p1779718624966249

---
[View Session in Sentry](https://sentry.sentry.io/traces/?project=4510944073809921&query=gen_ai.conversation.id%3A%22slack%3AC1G612XHC%3A1779718624.966249%22)

### DIFF
--- a/docs/security-legal-pii/security/ssl.mdx
+++ b/docs/security-legal-pii/security/ssl.mdx
@@ -44,7 +44,7 @@ This suite of protocols and cipher suites represents Sentry's official support; 
 
 ## Certificate Pinning
 
-At the moment, all Sentry TLS certificates use either Digicert or Let's Encrypt as certificate authorities. If this list of roots needs to be changed, we will publish an announcement on [status.sentry.io](https://status.sentry.io/).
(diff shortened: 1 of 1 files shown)


coolguyzone (approved):
LGTM, thanks for updating!

PR #18554, opened 2026-06-25
docs(dart): Add New Spans page and stream mode migration guide

buenaflor:
## DESCRIBE YOUR PR

Introduces the Dart/Flutter **stream mode** (span streaming) documentation — the base of a 4-PR stack. Mirrors the Python New Spans docs.

- New **New Spans** page: enabling stream mode (`traceLifecycle: SentryTraceLifecycle.stream`), manual instrumentation (`startSpan`/`startSpanSync`/`startInactiveSpan`), typed `SentryAttribute` attributes (including array factories), span status, `beforeSendSpan`/`ignoreSpans`, sampling, and verification
- New **Migrate to Stream Mode** guide mapping the transaction API to the new span API
- Minimum SDK version documented as `9.23.0`

This is the bottom of the stack — merge this first. Stacked children: #18609, #18610, #18611.

Ref: https://github.com/getsentry/sentry-dart/issues/3886

## IS YOUR CHANGE URGENT?

- [ ] Other deadline: 

## SLA

## PRE-MERGE CHECKLIST


docs/platforms/dart/common/tracing/streamed-spans/index.mdx
```diff
@@ -0,0 +1,249 @@
...
+<PlatformContent includePath="performance/span-streaming-enable" />
+
+To revert to transaction mode, remove the option or set `traceLifecycle` to `SentryTraceLifecycle.static` (the default).
+
+You can only use one tracing system at a time:
+
+- In `stream` mode, the transaction APIs (`Sentry.startTransaction`, `ISentrySpan.startChild`) do nothing and log a warning.
+- In `static` mode, the new span APIs (`Sentry.startSpan`) do nothing.
+- Auto-instrumentations switch to the correct API automatically based on this setting.
+
```
inventarSarah, line 101: I'd like to add the Expandable "How does span flushing work?" we have in the other guides here too. Or is there a specific reason why it's not included on this page?


This is the content I'd add (copied from Python):
```
<Expandable title="How does span flushing work?">

When stream mode is enabled, the SDK maintains an internal buffer that groups spans by trace ID.

Spans are flushed:

- On a regular interval (every 5 seconds by default).
- When a trace's buffer reaches 1,000 spans.
- When the SDK shuts down.

Each flush sends only the spans accumulated since the last flush, grouped into envelopes by trace ID.

</Expandable>
```

docs/platforms/dart/common/tracing/new-spans/index.mdx
```diff
@@ -0,0 +1,249 @@
...
+void onPaymentComplete() {
+  paymentSpan.end();
+}
+```
+
+#### Control Parenting
+
+By default, a span inherits the currently active span as its parent. To change this, pass `parentSpan`:
+
+- `parentSpan: null` forces a root span with no parent.
```
inventarSarah: does this force a root span or a service span?
iirc, in python, it's a service span

docs/platforms/dart/common/tracing/new-spans/index.mdx
```diff
@@ -0,0 +1,249 @@
...
+- `parentSpan: someSpan` parents the new span under a specific `SentrySpanV2`.
```
inventarSarah: we mention `SentrySpanV2` twice in this document -> will this be documented somewhere we can link to (I haven't looked at the other prs yet)? If not, I propose to just write "span"

docs/platforms/dart/common/tracing/new-spans/index.mdx
```diff
@@ -0,0 +1,249 @@
...
+  if (spanContext.name == 'health-check') {
+    return 0.0;
+  }
+  return 0.2;
+};
+```
+
+Only **root spans** are sampled; child spans inherit the root span's decision. When a root span isn't sampled, its callback still executes with a no-op span, so all span operations remain safe to call.
+
+## Flutter Auto-Instrumentation
```
inventarSarah: Does it make sense to only show this on the Flutter page and not the Dart page?

docs/platforms/dart/common/tracing/streamed-spans/migration-guide.mdx
```diff
@@ -0,0 +1,101 @@
...
+- final transaction = Sentry.startTransaction('checkout', 'task');
+- try {
+-   final child = transaction.startChild('db.query', description: 'load cart');
+-   final cart = await loadCart();
+-   await child.finish();
+-   transaction.setData('cart.item_count', cart.items.length);
+- } finally {
+-   await transaction.finish();
+- }
++ await Sentry.startSpan('checkout', (span) async {
```
inventarSarah, line 29: We should add more examples here like we did for Python:
- start a span
- Start what used to be a transaction (parent none)
- Start a child span while a parent is active
- Start a child span with a parent that's currently not active

let me know if you have any thoughts on this

docs/platforms/dart/common/tracing/streamed-spans/index.mdx
buenaflor: Sounds good, I'd also add:

`When the internal buffer size exceeds 1 MiB` or something like that. This is specific to mobile because we have tighter memory constraints

docs/platforms/dart/common/tracing/new-spans/index.mdx
buenaflor: Service span should be more accurate here 👍

docs/platforms/dart/common/tracing/new-spans/index.mdx
buenaflor: Not explictly no, it's the class name for the Span you use in Dart code. I'm also fine with just calling it span

docs/platforms/dart/common/tracing/new-spans/index.mdx
buenaflor: Ah that's my bad. yes definitely

docs/platforms/dart/common/tracing/streamed-spans/migration-guide.mdx
buenaflor: Makes sense, here are some examples - feel free to adapt these as you see fit

### 1. Start a span

If there is no active parent, this becomes a service span. If there is an active span, it becomes a child automatically.

```dart
await Sentry.startSpan('load-user', (span) async {
  span.setAttribute('user.id', SentryAttribute.string(userId));

  await loadUser(userId);
});
```

### 2. Start what used to be a transaction

Explicitly pass `parentSpan: null` to force a service span.

```dart
await Sentry.startSpan(
  'checkout-flow',
  (span) async {
    await runCheckout();
  },
  parentSpan: null,
);
```

### 3. Start a child span while a parent is active

Nested `startSpan` calls parent automatically through zones.

```dart
await Sentry.startSpan('checkout-flow', (checkoutSpan) async {
  await Sentry.startSpan('charge-card', (chargeSpan) async {
    await paymentService.charge();
  });

  await Sentry.startSpan('create-order', (orderSpan) async {
    await orderService.create();
  });
});
```

### 4. Start a child span with a parent that is not active

Keep the parent span reference and pass it explicitly.

```dart
final parent = Sentry.startInactiveSpan(
  'background-sync',
  parentSpan: null,
);

try {
  await someOtherAsyncBoundary();

  await Sentry.startSpan(
    'fetch-page',
    (child) async {
      await api.fetchPage();
    },
    parentSpan: parent,
  );
} finally {
  parent.end();
}
```

docs/platforms/dart/common/tracing/new-spans/index.mdx
```diff
@@ -0,0 +1,249 @@
...
+final config = Sentry.startSpanSync('parse-config', (_) {
+  return Config.parse(raw);
+});
+```
+
+If a span isn't sampled, the callback still runs and receives a no-op span, so all span operations remain safe to call.
+
+#### Spans That Outlive a Callback
+
+Use `Sentry.startInactiveSpan` when the work can't be wrapped in a single callback — widget lifecycles, stream subscriptions, or platform channel round-trips. You must call `end()` manually, and other spans do **not** automatically become its children:
```
inventarSarah:
```
.... and other spans do **not** automatically become its children:
```
How do you create child spans in that case?

inventarSarah: Do you have to set the parent explicitly?

buenaflor: Yes, then the 'inactive span' needs to be set as parent specifically

inventarSarah (approved):

PR #18615, opened 2026-07-01
chore(plugins): Remove plugin docs

ceorourke:
Remove plugin docs and references now that we've nuked plugins.


saponifi3d (approved):
👍 lgtm!

PR #18647, opened 2026-07-06
docs(java): Document OpenTelemetry BOM alignment

adinauer:
Document the new `io.sentry:sentry-opentelemetry-bom` for Java OpenTelemetry agentless and OTLP setups.

Sentry's OpenTelemetry integrations rely on a compatible OpenTelemetry core, instrumentation, and incubator dependency set. Spring Boot dependency management, Maven BOM imports, or other build-level dependency management can downgrade part of that set and leave users with classpath errors. These docs show how to opt in to Sentry's OpenTelemetry BOM without adding that behavior to Java agent setup pages.

The new troubleshooting page gives users a place to start when they hit OpenTelemetry classpath errors and links back to the BOM setup guidance for the relevant integration.

This should land with the sentry-java change that publishes `sentry-opentelemetry-bom`.


docs/platforms/java/common/opentelemetry/setup/otlp.mdx
```diff
@@ -47,6 +47,76 @@ This includes `sentry-opentelemetry-otlp` and the OpenTelemetry Spring Boot star
...
+
+```groovy {tabTitle:Gradle}{filename:build.gradle}
+dependencyManagement {
+    imports {
+        mavenBom 'io.sentry:sentry-opentelemetry-bom:{{@inject packages.version('sentry.java.opentelemetry-bom', '8.46.0') }}'
+    }
+}
+```
+
+When using Maven, import Sentry's OpenTelemetry BOM before `spring-boot-dependencies`:
```
lbloder, line 93: Should gradle and maven be tabs like for the java view or do we want to have them split due to the additional info outside of snippet?

adinauer: In this case I think it makes sense to have it split since there's additional info that feels odd in a comment or mixing instructions around the snippets.

lbloder (approved):
LGTM 👍

coolguyzone (approved):
Looks good!

PR #18693, opened 2026-07-10
docs(sdk): Clarify denylist source for parsed cookies/query params

ericapisani:
Clarifies that the `httpHeaders` denylist rules don't apply directly to parsed cookie names and query param keys. Instead, the `cookies` and `queryParams` configuration options each have their own dedicated denylist rules/terms, per key type.

Related to PY-2581


cleptric (commented):
Duplicate the expectation, this will be easier to read.

chargome (approved):

PR #18744, opened 2026-07-16
docs(sdk): Adjust and clarify strict trace continuation spec

szokeasaurusrex:
This PR proposes clarifying strict trace continuation so its behavior is only specified when the SDK knows its own organization ID. When both the SDK and incoming trace lack an organization ID, the trace may have come from a poorly configured or older third-party SDK, not the same organization. Strict trace continuation therefore has no useful defined behavior without SDK organization awareness. Users who enable the option likely intend to verify the trace's organization, so SDKs should warn when the organization ID is missing and may prevent that configuration where appropriate.

Separate the baseline organization-ID continuation rules from strict trace continuation. The default rules continue traces unless both known organization IDs differ; strict mode additionally rejects traces that omit an organization ID when the SDK has one. Keeping these concepts separate makes the specification easier to reason about and avoids treating an SDK without organization awareness as a valid strict-continuation use case.

> [!TIP]
> View the [relevant portion of the docs preview](https://develop-docs-git-szokeasaurusrex-update-strict-trace-con-7901ea.sentry.dev/sdk/foundations/trace-propagation/#trace-continuation).

Resolves [SDK-1365](https://linear.app/getsentry/issue/SDK-1365/update-develop-docs-for-strict-trace-continuation-proposal)

### DIFF
--- a/develop-docs/sdk/foundations/trace-propagation/index.mdx
+++ b/develop-docs/sdk/foundations/trace-propagation/index.mdx
@@ -2,7 +2,7 @@
 title: Trace Propagation
 description: How SDKs propagate trace context between services via headers, metadata, and environment variables.
 spec_id: sdk/foundations/trace-propagation
-spec_version: 1.9.0
+spec_version: 1.10.0
 spec_status: stable
 spec_depends_on:
   - id: sdk/foundations/transport/envelopes
@@ -10,6 +10,9 @@ spec_depends_on:
   - id: sdk/foundations/state-management/scopes
     version: ">=1.0.0"
 spec_changelog:
+  - version: 1.10.0
+    date: 2026-07-16
+    summary: Clarified trace continuation and strict trace continuation behavior
   - version: 1.9.0
     date: 2026-02-24
     summary: Consolidated into foundations/trace-propagation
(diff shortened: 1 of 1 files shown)


develop-docs/sdk/foundations/trace-propagation/index.mdx
```diff
@@ -2,14 +2,17 @@
...
 spec_status: stable
 spec_depends_on:
   - id: sdk/foundations/transport/envelopes
     version: ">=1.0.0"
   - id: sdk/foundations/state-management/scopes
     version: ">=1.0.0"
 spec_changelog:
+  - version: 1.10.0
+    date: 2026-07-16
+    summary: Clarified trace continuation and strict trace continuation behavior
```
szokeasaurusrex, line 15: The clanker added these changes, is it necessary to write this manually, or is there some automation to update the changelog here?

coolguyzone (approved):
Looks good!

develop-docs/sdk/foundations/trace-propagation/index.mdx
dingsdax: that's fine

develop-docs/sdk/foundations/trace-propagation/index.mdx
```diff
@@ -126,59 +129,80 @@ This option replaces the non-standardized `tracingOrigins` option which was prev
...
+When `strictTraceContinuation` is `true` but the SDK organization ID is missing, trace continuation behavior is unspecified. SDKs **SHOULD** warn users about this configuration. SDKs **MAY** prevent users from enabling `strictTraceContinuation` without an organization ID in a platform-appropriate way.
+
+The following table compares the default trace continuation behavior (with `strictTraceContinuation` set to `false`) against the strict trace continuation behavior.
+
+| Incoming `sentry-org_id` | SDK organization ID | Default Trace Continuation | Strict Trace Continuation |
+| --- | --- | --- | --- |
+| `1` | `1` | ✅ Continue trace | ✅ Continue trace |
+| none | `1` | ✅ Continue trace | ❌ Start new trace |
+| `1` | `2` | ❌ Start new trace | ❌ Start new trace |
+| any | none | ✅ Continue trace | ❓ Unspecified |
```
cleptric, line 199: We can't have docs with `Unspecified` behaviour. We start a new trace.

develop-docs/sdk/foundations/trace-propagation/index.mdx
```diff
@@ -126,59 +129,80 @@ This option replaces the non-standardized `tracingOrigins` option which was prev
...
+When `strictTraceContinuation` is `true` and the SDK has an organization ID:
+
+- If the incoming `sentry-org_id` matches the SDK organization ID, the SDK **SHOULD** continue the trace.
+- If the incoming `sentry-org_id` is missing or differs from the SDK organization ID, the SDK **MUST NOT** continue the trace.
+
+When `strictTraceContinuation` is `true` but the SDK organization ID is missing, trace continuation behavior is unspecified. SDKs **SHOULD** warn users about this configuration. SDKs **MAY** prevent users from enabling `strictTraceContinuation` without an organization ID in a platform-appropriate way.
+
+The following table compares the default trace continuation behavior (with `strictTraceContinuation` set to `false`) against the strict trace continuation behavior.
+
+| Incoming `sentry-org_id` | SDK organization ID | Default Trace Continuation | Strict Trace Continuation |
```
cleptric, line 194: There is already a table above that specifies the default behaviour.

develop-docs/sdk/foundations/trace-propagation/index.mdx
```diff
@@ -126,59 +129,80 @@ This option replaces the non-standardized `tracingOrigins` option which was prev
...
+
 - If `orgId` is present, the SDK's organization ID is the value passed to `orgId`.
 - Otherwise, the SDK's organization ID is the organization ID [in the DSN](https://develop.sentry.dev/sdk/foundations/transport/authentication/#parsing-the-dsn).
-- If `orgId` is absent and the DSN cannot be parsed, then the organization ID is missing.
+- If `orgId` is absent and the DSN cannot be parsed, the SDK organization ID is missing.
 
 If the SDK has an organization ID, the SDK **MUST** propagate the organization ID in baggage as `sentry-org_id`.
 
-#### Continuing Traces
+When the SDK organization ID is missing, SDKs **MAY** warn users and recommend configuring `orgId`.
```
cleptric, line 151: *emit a debug log that informs the user.

develop-docs/sdk/foundations/trace-propagation/index.mdx
```diff
@@ -126,59 +129,80 @@ This option replaces the non-standardized `tracingOrigins` option which was prev
...
+</Alert>
+
+`strictTraceContinuation` **MUST** be a boolean value. The default is `false`.
+
+When `strictTraceContinuation` is `true` and the SDK has an organization ID:
+
+- If the incoming `sentry-org_id` matches the SDK organization ID, the SDK **SHOULD** continue the trace.
+- If the incoming `sentry-org_id` is missing or differs from the SDK organization ID, the SDK **MUST NOT** continue the trace.
+
+When `strictTraceContinuation` is `true` but the SDK organization ID is missing, trace continuation behavior is unspecified. SDKs **SHOULD** warn users about this configuration. SDKs **MAY** prevent users from enabling `strictTraceContinuation` without an organization ID in a platform-appropriate way.
```
cleptric, line 190: We parse the org ID from the DSN, how can we prevent setting an option?

cleptric (commented):
Just a first pass, this does not look correct.

develop-docs/sdk/foundations/trace-propagation/index.mdx
szokeasaurusrex: I agree with you in general that our specs should not allow for a supported input to have an arbitrary result. However, my intention with labelling this row as "unspecified" was narrower: strict trace continuation is inapplicable when the SDK cannot determine its own org ID because we cannot properly validate the incoming org ID in that state.

If we go with your proposal of changing this row to "Start new trace," then we would have to change all the SDKs that already implemented strict trace continuation according to the old spec. That seems to me like a significant amount of effort to change how we handle a case which we should not support.

Could we instead describe this state as a non-applicable, unsupported configuration instead, with a note that the exact SDK behavior is implementation-defined? I'm also open to other ways of communicating that intent

develop-docs/sdk/foundations/trace-propagation/index.mdx
szokeasaurusrex: I mean that we can error loudly on initialization when strict trace continuation is enabled and neither `org_id` is passed nor can the org ID be parsed from the DSN.

There are several precedents for erroring when invalid states are encountered upon init:

- Java's core `Sentry.init` propagates DSN validation errors: [source](https://github.com/getsentry/sentry-java/blob/796047a6054c10556286d8b161e9634fa9880bf5/sentry/src/main/java/io/sentry/Sentry.java#L598-L616)
- Python's `sentry_sdk.init` propagates `ValueError` for invalid option values: [source](https://github.com/getsentry/sentry-python/blob/3db66a6b2a58539c36d83188466851b1bb570623/sentry_sdk/client.py#L676-L682)
- Go's `sentry.Init` returns an error when initialization options are invalid, including a malformed DSN: [source](https://github.com/getsentry/sentry-go/blob/fb9acd84b1dfc91073c046e9c5612faf3422d749/sentry.go#L18-L28)

My suggestion is to allow SDKs to follow such precedent to prevent initializing the SDK with `strict_trace_continuation=true` and no org ID

develop-docs/sdk/foundations/trace-propagation/index.mdx
szokeasaurusrex: Yeah, I intentionally included both columns so that there could be a clear side-by-side comparison, but we can also deduplicate the two if you think that is better.

szokeasaurusrex (commented):
Replied to some of the comments

PR #18766, opened 2026-07-20
docs(js): Document tracing setup for sveltekit 3

chargome:
Documents how tracing is set up with sveltekit 3

ref https://github.com/getsentry/sentry-javascript/issues/21502


inventarSarah (approved):
👍

PR #18797, opened 2026-07-22
chore: Stop tracking dotagents-managed artifacts

szokeasaurusrex:
dotagents now manages agents.lock and .agents/.gitignore as local state and expects them gitignored. Remove the tracked leftovers, drop the stale comment in agents.toml, and document in AGENTS.md that skills must be installed via 'make develop'.

The .gitignore and agents.toml changes are per 'dotagents doctor --fix' recommendations.

PR #18809, opened 2026-07-23
chore(dotagents): Manage .claude with dotagents

szokeasaurusrex:
PR #18797 left stale skill files in the `.claude` directory.

By slightly adjusting the `dotagents` configuration, we can also manage Claude skills using `dotagents`. This has the added benefit of keeping `.claude` skills in sync with the vendor-neutral `.agents` standard, which Claude Code unfortunately refuses to support.


.claude/skills/commit
szokeasaurusrex, line 1: These are stale symlinks, which are replaced now by the overarching `.claude/skills` symlink to `.agents/skills`

agents.toml
szokeasaurusrex, line 1: The changes to this file were generated by running `dotagents init --force`, which sets up `dotagents` from scratch, with support configured for all dotagents-supported agents.

As `init --force` clears the config, I then added back all the pre-existing skills manually.

skills-lock.json
szokeasaurusrex, line 1: This file is also made redundant by `dotagents`

sfanahata (approved):
Not blocking, but still curious to know how this will work in the wild? It looks like people will now need to know to manually add skills?

PR #18866, opened 2026-07-29
fix(platform-redirect): Restrict redirects to same-origin paths

chromy:
Restricts the `/platform-redirect` endpoint to same-origin destinations. Ref VULN-2173.

- Redirect to the sanitized path rather than the raw `next` query param
- Rebuild `sanitizeNext` on the WHATWG `URL` parser so target validation matches browser URL resolution semantics
- Add test coverage for same-origin enforcement

### DIFF
--- a/app/platform-redirect/page.tsx
+++ b/app/platform-redirect/page.tsx
@@ -124,9 +124,19 @@ export default async function Page(props: {
     }
   }
 
+  // make the Sidebar aware of the current path
+  setServerContext({rootNode, path: ['platform-redirect']});
+
+  // Only redirect to doctree-resolved URLs. Never redirect based on `next` alone
   if (platformOrGuideList.length === 0) {
-    // Never redirect to the raw, attacker-controlled `next`
-    return redirect(pathname || '/');
+    return (
+      <DocPage frontMatter={{title: defaultTitle, description: ''}}>
+        <Alert level="warning">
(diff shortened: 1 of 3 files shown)


Jeffreyhung:
replaced devco.re with example.com, probably not the best to use a vendor's domain in there :P

app/platform-redirect/page.tsx
```diff
@@ -124,9 +124,19 @@ export default async function Page(props: {
...
-    // try to redirect the user to the page directly, might result in 404
-    return redirect(next);
+    return (
+      <DocPage frontMatter={{title: defaultTitle, description: ''}}>
+        <Alert level="warning">
+          We couldn't find documentation for this path on any platform.{' '}
+          <SmartLink to="/">Go to the docs home</SmartLink>.
+        </Alert>
+      </DocPage>
+    );
```
Jeffreyhung, line 139: @sergical let me know if you have any concerns with this (mainly around user experience)
This is a bit more aggressive but also more future proof around redirects problems

sergical: id be surprised if this ever got rendered tbqh

sergical (approved):

app/platform-redirect/page.tsx
Jeffreyhung: hopefully we will never need it :)

PR #18867, opened 2026-07-29
docs(data-collection): Clarify Session Replay is not gated by `dataCollection`

chargome:
Documents that Session Replay is intentionally not wired into `dataCollection` (or `sendDefaultPii`). Because Replay uses an opt-in privacy model while `dataCollection` is opt-out, coupling them would flip privacy-sensitive defaults for existing Replay users. This clarifies the exception in both the spec and the JS user docs.


s1gr1d (approved):

ericapisani (approved):

philprime (approved):
LGTM

PR #18880, opened 2026-07-30
docs(dataCollection): Extend url example to encoded query params

chargome:
Add spec that URL query keys/values must be stored percent-encoded, exactly as they appear in the URL

### DIFF
--- a/develop-docs/sdk/foundations/client/data-collection/index.mdx
+++ b/develop-docs/sdk/foundations/client/data-collection/index.mdx
@@ -204,7 +204,7 @@ URL strings in span descriptions follow the format defined in [Structuring Data]
 
 The `urlQueryParams` option controls query parameter filtering wherever query strings appear — including `url.full`, `url.query`, and [`request.query_string`](/sdk/foundations/envelopes/event-payloads/request/). `url.path` and the span description are unaffected since they never contain query strings. For general data scrubbing rules (variable size limits, structuring for server-side scrubbing), see [Data Scrubbing](/sdk/foundations/data-scrubbing/).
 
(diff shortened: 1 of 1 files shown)


develop-docs/sdk/foundations/client/data-collection/index.mdx
```diff
@@ -201,15 +204,17 @@ URL strings in span descriptions follow the format defined in [Structuring Data]
 
 The `urlQueryParams` option controls query parameter filtering wherever query strings appear — including `url.full`, `url.query`, and [`request.query_string`](/sdk/foundations/envelopes/event-payloads/request/). `url.path` and the span description are unaffected since they never contain query strings. For general data scrubbing rules (variable size limits, structuring for server-side scrubbing), see [Data Scrubbing](/sdk/foundations/data-scrubbing/).
 
-**Example:** Given a request to `https://example.com/api/users?token=abc123&page=5`:
+Query keys and values **MUST** be stored percent-encoded, exactly as they appear in the URL. SDKs **MUST NOT** decode them before attaching them, since decoding makes delimiters ambiguous and loses the original request.
```
s1gr1d:
```suggestion
Encoded query keys and values **MUST** be stored exactly as they appear in the URL.
```

I would just keep it very simple as this already goes beyond this spec's purpose. wdyt?

ericapisani: This might be a bit of a nit pick (so feel free to ignore), but I think we can say "query parameters" instead of "keys and values".

This would help avoid potential drift in terminology over time, such as what's highlighted by the [bot above](https://github.com/getsentry/sentry-docs/pull/18880/changes#r3682792506).
```suggestion
Query parameters **MUST** be stored percent-encoded, exactly as they appear in the URL. SDKs **MUST NOT** decode them before attaching them, since decoding makes delimiters ambiguous and loses the original request.
```

ericapisani (approved):
One comment a food for thought but otherwise LGTM

PR #18922, opened 2026-08-03
docs(sdk): Make the client event pipeline the authoritative step ordering

buenaflor:
## DESCRIBE YOUR PR

The Client and Sessions specs disagreed on the event pipeline order: the Client spec sampled at step 2 and had no session update step, while the Sessions spec put the session update after `before_send` with sampling last. We ratified the Sessions ordering, so this aligns the Client spec to it and makes that the only place the order is stated — Sessions now defers to it instead of restating it.

As discussed internally:
Sampling runs late only so that sampled-out events still count toward release health, so SDKs without session support **MAY** keep sampling immediately after the disabled check.

Client spec bumped to `3.0.0` as a breaking reorder — happy to drop it to `2.3.0` if you read this as correcting a spec that never matched real SDK behavior.

## IS YOUR CHANGE URGENT?

- [ ] Other deadline:

## PRE-MERGE CHECKLIST

### DIFF
--- a/develop-docs/sdk/foundations/client/index.mdx
+++ b/develop-docs/sdk/foundations/client/index.mdx
@@ -137,7 +137,7 @@ Steps 2 through 4 are developer-controlled filters: an event they discard **MUST
 
 SDKs that implement sessions **MUST** apply sampling after the session update, as ordered above.
 
-SDKs that do **not** implement sessions **MAY** apply sampling immediately after the disabled check, before scope application and `before_send`. With no session to update, moving sampling earlier has no effect on release health and avoids the cost of running event processors and `before_send` on events that will be discarded. These SDKs **MUST** still record the `sample_rate` client report.
+SDKs that do **not** implement sessions **MAY** apply sampling immediately after the disabled check. With no session to update, moving sampling earlier has no effect on release health and avoids the cost of running event processors and `before_send` on events that will be discarded. These SDKs **MUST** still record the `sample_rate` client report.
 
 </SpecSection>
 
--- a/develop-docs/sdk/foundations/client/index.mdx
+++ b/develop-docs/sdk/foundations/client/index.mdx
@@ -133,7 +133,7 @@ When `captureEvent`, `captureException`, or `captureMessage` is called, the Clie
 
 Ignored exception types, event processors, and `before_send` are developer-controlled filters: an event they discard **MUST NOT** update the session, because the developer chose to ignore it. Sampling and the transport discard events to save quota, so they **MUST NOT** prevent the session update — a sampled-out event still counts toward release health. This is the only reason sampling runs after `before_send`.
 
-SDKs that do not implement sessions therefore **MAY** apply sampling immediately after the disabled check, discarding events before event processors and `before_send` run. They **MUST** still record the `sample_rate` client report.
+SDKs without session support **MAY** instead apply sampling immediately after the disabled check. With no session to update, where sampling runs does not affect release health, and sampling early skips event processors and `before_send` for events that will never be sent. The `sample_rate` client report is **REQUIRED** either way.
 
 </SpecSection>
 
--- a/develop-docs/sdk/foundations/client/index.mdx
+++ b/develop-docs/sdk/foundations/client/index.mdx
@@ -124,20 +124,16 @@ See [Scopes: Client Resolution](/sdk/foundations/state-management/scopes/#client
 When `captureEvent`, `captureException`, or `captureMessage` is called, the Client processes the event through the following pipeline. The event **MAY** be discarded at any stage, stopping further processing.
 
 1. **Disabled check**: If the SDK is disabled (no transport / no DSN), discard immediately.
-2. **Ignored exception types**: Discard events matching `ignoreErrors` (client report reason: `event_processor`). See [Configuration](/sdk/foundations/client/configuration/#event-sampling).
+2. **Ignored exception types**: Discard events matching [`ignoreErrors`](/sdk/foundations/client/configuration/#event-sampling) (client report reason: `event_processor`).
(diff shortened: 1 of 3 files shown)


develop-docs/sdk/foundations/client/index.mdx
```diff
@@ -114,17 +117,23 @@ See [Scopes: Client Resolution](/sdk/foundations/state-management/scopes/#client
 
 </SpecSection>
 
-<SpecSection id="event-pipeline" status="stable" since="1.0.0">
+<SpecSection id="event-pipeline" status="stable" since="3.0.0">
```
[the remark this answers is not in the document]
buenaflor: Keeping `since="3.0.0"` here — the convention is the opposite of this.

From [`spec-lifecycle.mdx`](https://github.com/getsentry/sentry-docs/blob/master/develop-docs/sdk/getting-started/standards/spec-lifecycle.mdx):

> A section's `since` **MUST** be the `spec_version` in which that section's current requirements were introduced. Bump it only when the section's requirements actually change.

So `since` tracks the current requirements, not when the section first appeared. Reordering the mandated pipeline steps does change this section's requirements, so bumping it is what the rule prescribes.

The surrounding code agrees: `specSection.tsx` renders the value as "specified since {since}", and `client-resolution` already sits at `since="2.0.0"` in this same file, which started at 1.0.0. The inverse case is visible in the sessions spec, where `session-update-filtering` stayed at `1.4.0` when 1.4.1 only clarified it.

develop-docs/sdk/foundations/client/index.mdx
```diff
@@ -2,14 +2,17 @@
...
 spec_status: stable
 spec_depends_on:
   - id: sdk/foundations/state-management/scopes
     version: ">=1.0.0"
   - id: sdk/foundations/transport/envelopes
     version: ">=1.0.0"
 spec_changelog:
+  - version: 3.0.0
+    date: 2026-08-03
+    summary: Breaking — event pipeline is now the authoritative ordering; sampling runs after before_send and the session update for SDKs that support sessions; added ignored exception types and session update steps
```
Lms24: l: this sounds a bit clankered (sorry in case it wasn't). I think we can shorten it a bit:

```suggestion
    summary: Breaking: Sampling runs after before_send and the session update for SDKs that support sessions; added ignored exception types and session update steps
```

develop-docs/sdk/foundations/client/index.mdx
```diff
@@ -114,17 +117,23 @@ See [Scopes: Client Resolution](/sdk/foundations/state-management/scopes/#client
...
-2. **Sampling**: Apply the configured `sample_rate`. Events **MAY** be randomly discarded. When discarded, the Client **MUST** record a [client report](/sdk/telemetry/client-reports/) with discard reason `sample_rate`.
-3. **Scope application**: Obtain the current, isolation, and global scopes. Merge scope data onto the event in the [defined order](/sdk/foundations/state-management/scopes/#applying-scope-data-to-events). Invoke event processors in registration order. Any processor returning `null` discards the event (client report reason: `event_processor`).
+2. **Ignored exception types**: Discard events matching [`ignoreErrors`](/sdk/foundations/client/configuration/#event-sampling) (client report reason: `event_processor`).
+3. **Scope application**: Obtain the current, isolation, and global scopes. Merge scope data onto the event in the [defined order](/sdk/foundations/state-management/scopes/#applying-scope-data-to-events). Invoke scoped event processors first, then global event processors, each in registration order. Any processor returning `null` discards the event (client report reason: `event_processor`).
```
Lms24: m: The ordering here might still depend a bit on SDK implementation. In JS, `ignoreErrors` is added via an event processor on client-level. I think it's good that we call it out in this pipeline but maybe we can merge these two steps? In general, I think there are 3 things going on in these two steps and we could probably point this out more clearly: 

- applying scope data onto the event
- applying event processors in the written down order
- applying `ignoreErrors` (which might very well just be an event processor)
 
technically, the ordering here has influence on the behaviour. In JS, we apply data before we apply event processors which I think makes sense but implementations might vary.

develop-docs/sdk/foundations/client/index.mdx
```diff
@@ -114,17 +117,23 @@ See [Scopes: Client Resolution](/sdk/foundations/state-management/scopes/#client
...
 4. **`before_send` hook**: Invoke the user-configured callback. Returning `null` discards the event (client report reason: `before_send`). This hook **MUST** only apply to error events. For transactions, use `before_send_transaction`. For logs, use `before_send_log`. See [Hooks](/sdk/foundations/client/hooks/) for details.
-5. **Transport**: Pass the event to the transport (or TelemetryProcessor). The transport **MAY** discard due to missing DSN, full queue, or [rate limiting](/sdk/foundations/transport/rate-limiting/).
+5. **Session update**: SDKs that implement [Sessions](/sdk/telemetry/sessions/) **MUST** update the current session here — incrementing its `errors` count or setting its status. SDKs without session support skip this step.
```
Lms24: l:
```suggestion
5. **Session update**: SDKs that implement [Sessions](/sdk/telemetry/sessions/) **MUST** update the current session by incrementing its `errors` count or setting the appropriate status. SDKs without session support skip this step.
```

Lms24 (commented):
thanks for updating. This looks good to me and reflects what we discussed in the last weeks. Just had some minor remarks regarding ordering.

coolguyzone (approved):

develop-docs/sdk/foundations/client/index.mdx
buenaflor: thx! updated it, lmk what you think

Lms24 (approved):
thank you, looks good now!

cleptric (approved):

PR #18944, opened 2026-08-05
docs(cloudflare): clarify Vite entry instrumentation

sergical:
Clarifies that the Cloudflare Vite plugin locates `instrument.server.*` next to the Worker entry configured by wrangler's `main`, with a concrete nested-entry example.

Documents that configured Durable Object, Workflow, and Agents SDK classes imported or re-exported from another module cannot be wrapped automatically, and shows how to export a manually wrapped class from the entry with the same options callback.

### DIFF
--- a/docs/platforms/javascript/guides/cloudflare/features/vite-plugin.mdx
+++ b/docs/platforms/javascript/guides/cloudflare/features/vite-plugin.mdx
@@ -85,18 +85,35 @@ export default defineConfig({
 });
 ```
 
-With auto-instrumentation, you can optionally provide Sentry options via a co-located `instrument.server.*` file (`.ts`, `.mts`, `.js`, `.mjs`, or `.cjs`) next to your Worker entry. Use `defineCloudflareOptions` for full type-checking:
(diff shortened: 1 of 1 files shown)


docs/platforms/javascript/guides/cloudflare/features/vite-plugin.mdx
```diff
@@ -85,19 +85,36 @@ export default defineConfig({
...
 If no `instrument.server.*` file exists, the SDK reads all configuration (DSN, release, environment, sample rate, etc.) from the Worker's `env` bindings at runtime.
 
+Configured Durable Object, Workflow, and Agents SDK classes must be declared in the Worker entry for the plugin to wrap them automatically. The plugin cannot rewrite a class that the entry only imports or re-exports from another module. In that case, wrap the imported class in the entry with its matching helper and pass it the options callback from `instrument.server.*`:
+
+```typescript {filename:src/worker/index.ts}
+import * as Sentry from "@sentry/cloudflare";
+import { sentryOptions } from "./instrument.server";
+import { MyAgent as MyAgentBase } from "./my-agent";
+
+export const MyAgent = Sentry.instrumentAgentWithSentry(
```
JPeer264, line 110: That should actually work I did an entire hack to make it work. I wonder why this doesn't work correctly.

JPeer264: We can land this for now, as it doesn't seem to work properly

JPeer264 (approved):

PR #19043, opened 2026-08-13
add bulk delete limitation

rodolfoBee:
Adding two limitations to the bulk delete replay option that were not described on the page


gggritso (approved):
👍🏻

PR #19045, opened 2026-08-13
docs(js): Move Mastra under integrations

matejminar:
Right now, Mastra lives right under Agent Tracing making it look like it's the only thing we support.

This PR moves the Mastra exporter guide from Agent Tracing into the JavaScript integrations section, alongside Vercel AI, OpenAI, Anthropic, and the other supported AI stacks. The Agent Tracing overview now includes Mastra in its AI stack list, and existing internal links point directly to the new location.

Permanent redirects preserve the current and older Mastra page URLs, including framework-specific and Markdown export paths. Legacy URLs resolve directly to the new page without redirect chains.

Likely reason for why Mastra stood standalone is that it's technically exporter, not integration, but it's less confusing this way.


shellmayr (approved):

sfanahata (approved):
Looks much better. Thanks!

PR #19070, opened 2026-08-16
docs(javascript): clarify HTTP headers require sendDefaultPii

MHJahanbakhsh:
## Summary
- Correct the JavaScript **Data Collected** docs: HTTP response headers (and most request headers) are **not** sent by default.
- Clarify that collecting them requires the deprecated `sendDefaultPii: true`, or an explicit `dataCollection.httpHeaders` setting (aligned with the Cookies section and maintainer comments on the issue).
- Link `sendDefaultPii` option docs to the HTTP Headers section on Data Collected.

Fixes getsentry/sentry-javascript#20706

## Test plan
- [ ] Preview the Data Collected page HTTP Headers section
- [ ] Confirm `sendDefaultPii` option text links to `#http-headers`
- [ ] Cross-check wording against Cookies / `dataCollection` defaults on the same page

### DIFF
--- a/docs/platforms/javascript/common/configuration/options.mdx
+++ b/docs/platforms/javascript/common/configuration/options.mdx
@@ -110,7 +110,7 @@ Set to `false` to opt out, for example, when using a custom `tunnel` or when run
   (v11). Use [`dataCollection`](#dataCollection) instead.
 </Alert>
 
-Set this option to `true` to send default PII data to Sentry. Among other things, enabling this will enable automatic IP address collection on events.
+Set this option to `true` to send default PII data to Sentry. Among other things, enabling this will enable automatic IP address collection on events, and it is required to collect HTTP response headers (and most request headers) when you are not using [`dataCollection`](#dataCollection). See <PlatformLink to="/data-management/data-collected/#http-headers">Data Collected: HTTP Headers</PlatformLink> for details.
 
 For backwards compatibility, `sendDefaultPii: true` behaves like enabling all [`dataCollection`](#dataCollection) categories.
 If both options are set, `dataCollection` takes precedence.
--- a/docs/platforms/javascript/common/data-management/data-collected/index.mdx
+++ b/docs/platforms/javascript/common/data-management/data-collected/index.mdx
@@ -22,9 +22,9 @@ Regardless of these options, you can always scrub any data before it's sent to S
 
 ## HTTP Headers
 
-By default, the Sentry SDK sends HTTP request and response headers.
+By default, the Sentry SDK does **not** send HTTP response headers, and most request headers are omitted as well (for example, browser SDKs may only include limited request headers such as `User-Agent`). When using `dataCollection`, however, request and response headers are collected with sensitive values scrubbed. Opt out by setting `dataCollection: { httpHeaders: false }`.
 
-Use the `dataCollection.httpHeaders` option to control this. For example, set `dataCollection: { httpHeaders: false }` to disable it, or use `{ allow: [...] }` or `{ deny: [...] }` to restrict which header values are sent. Values whose keys match Sentry's built-in sensitive denylist (such as `auth`, `token`, or `password`) are automatically scrubbed, while the keys are kept.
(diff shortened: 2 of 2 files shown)


docs/platforms/javascript/common/data-management/data-collected/index.mdx
```diff
@@ -22,9 +22,9 @@ Regardless of these options, you can always scrub any data before it's sent to S
...
 
-Use the `dataCollection.httpHeaders` option to control this. For example, set `dataCollection: { httpHeaders: false }` to disable it, or use `{ allow: [...] }` or `{ deny: [...] }` to restrict which header values are sent. Values whose keys match Sentry's built-in sensitive denylist (such as `auth`, `token`, or `password`) are automatically scrubbed, while the keys are kept.
+To collect HTTP headers when not using `dataCollection`, set the deprecated `sendDefaultPii: true` in `Sentry.init()`. Response headers in particular require this opt-in (or an explicit `dataCollection.httpHeaders` setting). You can also restrict which header values are sent using `{ allow: [...] }` or `{ deny: [...] }` on `dataCollection.httpHeaders`. Values whose keys match Sentry's built-in sensitive denylist (such as `auth`, `token`, or `password`) are automatically scrubbed, while the keys are kept.
```
s1gr1d, line 27: "most request headers are omitted" isn't right for current versions: on the server, request headers are sent by default and only sensitive values are `[Filtered]` ([httpServerSpansIntegration.ts#L194](https://github.com/getsentry/sentry-javascript/blob/10.73.0/packages/node-core/src/integrations/http/httpServerSpansIntegration.ts#L194)). The part about response headers is correct for the browser `httpClient` integration though.


```suggestion
When `dataCollection` is configured, the SDK collects HTTP request and response headers by default, with sensitive values (such as `auth`, `token`, or `password`) replaced by `[Filtered]` while the keys are kept. Opt out with `dataCollection: { httpHeaders: { request: false, response: false } }`, or restrict values with `{ allow: [...] }` / `{ deny: [...] }` on `request` and `response`.

When `dataCollection` is not configured, the SDK sends request headers with the same filtering, plus additional filtering of IP-related headers. Response headers from browser requests (`httpClient` integration) additionally require the deprecated `sendDefaultPii: true`.
```

docs/platforms/javascript/common/configuration/options.mdx
```diff
@@ -110,7 +110,7 @@ Set to `false` to opt out, for example, when using a custom `tunnel` or when run
   (v11). Use [`dataCollection`](#dataCollection) instead.
 </Alert>
 
-Set this option to `true` to send default PII data to Sentry. Among other things, enabling this will enable automatic IP address collection on events.
+Set this option to `true` to send default PII data to Sentry. Among other things, enabling this will enable automatic IP address collection on events, and it is required to collect HTTP response headers (and most request headers) when you are not using [`dataCollection`](#dataCollection). See <PlatformLink to="/data-management/data-collected/#http-headers">Data Collected: HTTP Headers</PlatformLink> for details.
```
s1gr1d, line 113: Request headers don't need `sendDefaultPii`, and the response-header part only applies to the browser `httpClient` integration when `dataCollection` is unset. I'd keep just the link to the Data Collected page here.
```suggestion
See <PlatformLink to="/data-management/data-collected/#http-headers">Data Collected: HTTP Headers</PlatformLink> for details.
```

s1gr1d (changes requested):
Thanks for digging into this! The behavior on current SDK versions depends on whether `dataCollection` is set (at least in v10), so neither the old text nor this version is fully accurate:

- In v10, with `dataCollection` set (even `{}`): request and response headers are collected by default, sensitive values replaced by `[Filtered]` ([resolveDataCollectionOptions.ts#L4-L15](https://github.com/getsentry/sentry-javascript/blob/10.73.0/packages/core/src/utils/data-collection/resolveDataCollectionOptions.ts#L4-L15)).
- Without `dataCollection` (legacy v10 behavior): request headers are still sent with filtered values ([defaultPiiToCollectionOptions.ts#L28](https://github.com/getsentry/sentry-javascript/blob/10.73.0/packages/core/src/utils/data-collection/defaultPiiToCollectionOptions.ts#L28)). Only the browser `httpClient` integration is fully gated on `sendDefaultPii` ([httpclient.ts#L430-L438](https://github.com/getsentry/sentry-javascript/blob/10.73.0/packages/browser/src/integrations/httpclient.ts#L430-L438)).

So "response headers require `sendDefaultPii`" is only true for `httpClient`, and "most request headers are omitted" isn't true at all. Suggestions inline. The link from `sendDefaultPii` to the Data Collected page is a good addition, let's keep that.

In v11, `dataCollection` will be the only option and it's used by default.

I left suggestions inline.

s1gr1d:
Closing this in favor of this PR (which adds separate docs for v10 and v11): https://github.com/getsentry/sentry-docs/pull/19583

PR #19075, opened 2026-08-18
docs(python): clarify gauge vs distribution metric usage

VedantMadane:
Clarify when to use `gauge` vs `distribution` in the Python metrics docs: add a comparison table, use distinct examples (connections/memory vs latency/page load), and align guidance with the JavaScript metrics docs.

Related to getsentry/sentry-python#7211


ericapisani (approved):
Thank you so much for your contribution, these are a fantastic addition to the docs! 🙏🏻

ericapisani (approved):

PR #19152, opened 2026-08-27
ref(js): Move AI library pages under agent-tracing

chargome:
Each AI library (Vercel AI, OpenAI, Anthropic, Google Gen AI, LangChain, LangGraph) now has its own page under Agent Tracing instead of living in `configuration/integrations`. Old URLs redirect.

Merging `agent-tracing-browser` into `agent-tracing` is a prerequisite — the library pages are supported on browser platforms, which had no `agent-tracing` index. Without a parent page the sidebar drops the whole subtree, leaving the pages reachable by URL but invisible in nav.


stephanie-anderson:
@chargome looks great! The only thing missing is the integration also showing up in the sidebar as it does today. 
Screenshot 2026-08-28 at 12 16 15
Screenshot 2026-08-28 at 12 16 00

I think we can solve this by creating new files for the moved integrations, that simply have some frontmatter so they get rendered in the sidebar, but the redirect will anyways kick in then and redirect users to the agent tracing subpage

cc @ale-cota

stephanie-anderson (approved):
approving as the rest of the change is already as intended, and the rest is additive 👍

andreiborza (approved):

PR #19176, opened 2026-08-28
docs(attachments): Update attachments docs

rodolfoBee:
Multiple platforms have the same text for accessing and viewing attachments. 

This PR moves the text into a general page and imports this page into all platforms that support attachments. This makes maintaining this section of the docs easier.

Updating screenshots to the current UI at the same time.


coolguyzone (approved):
This is great! Thank you so much for helping keep the docs organized and up-to-date!

PR #19189, opened 2026-08-31
docs(sdk): Specify callback error isolation for user callbacks

dingsdax:
## DESCRIBE YOUR PR

Defines what an SDK must do when a user-provided callback throws. The hooks spec already defined each callback's `null`/return contract but said nothing about exceptions, so behavior diverged across SDKs — some drop the item, some keep it, some let the exception reach the host.

- New `Callback Error Isolation` section: recovery boundary, error-level log, no re-throw, no self-capture, plus a per-callback fallback table — filters drop, `before_send_span` emits the span, samplers fall back as if no sampler were configured.
- Client report obligations for callback failures: a dropped item is reported exactly as on an explicit drop, and a failure never produces a billed event.
- New `error_sampler` hook, which SDKs already ship but which was undocumented here.
- Spec bumped to 1.1.0; both new sections are `candidate`.

Refs https://linear.app/getsentry/project/wrap-all-sdk-user-callbacks-in-a-try-catch-6bfdbe9f7268

## IS YOUR CHANGE URGENT?

## SLA

## PRE-MERGE CHECKLIST


develop-docs/sdk/foundations/client/hooks/index.mdx
```diff
@@ -100,6 +109,53 @@ Hook called with the breadcrumb (and on some platforms the hint) that allow the
...
+| `before_send_log` | Drop the log | `before_send`, category `log_item` |
+| `before_send_metric` | Drop the metric | `before_send`, category `trace_metric` |
+| `before_send_feedback` | Drop the feedback | `before_send`, category `feedback` |
+| `before_send_check_in` | Drop the check-in | `before_send`, category `monitor` |
+| `before_breadcrumb` | Drop the breadcrumb | None |
+| `before_send_span` | Emit the span, discarding the return value | None |
+| Event processor | Drop the event | `event_processor`, category of the item |
+| `traces_sampler` | Fall back as if unconfigured | `sample_rate` if sampled out |
+| `profiles_sampler` | Fall back as if unconfigured | `sample_rate` if sampled out |
+| `error_sampler` | Fall back to `sample_rate` | `sample_rate` if sampled out |
```
Lms24, line 147: m: If we emit the same client report discard reasons for both of these cases

- a user callback fails/throws and we therefore drop an event
- a user callback filtered out an event

users can't distinguish between these cases. The alternative would be to add a new discard reason ([which is fairly easy](https://develop.sentry.dev/sdk/telemetry/client-reports/#adding-a-new-discard-reason)) like `callback_error`. Though with the caveat that we couldn't distinguish which callback threw. Unless we add a new reason per callback which is probably overkill.

I'd tend towards sending the `callback_error` discard reason for simplicity but I might be missing some context why we went with the same discard reason.

develop-docs/sdk/foundations/client/hooks/index.mdx
```diff
@@ -100,6 +109,53 @@ Hook called with the breadcrumb (and on some platforms the hint) that allow the
...
+
+This hook **MUST** only apply to `error` events. Unlike `before_send`, it is a sampling mechanism rather than a developer-controlled filter, so it **MUST** run at the sampling step of the [event pipeline](/sdk/foundations/client/#event-pipeline). An event it samples out **MUST** still update the session, and SDKs **MUST** record it with the `sample_rate` discard reason.
+
+</SpecSection>
+
+<SpecSection id="callback-error-isolation" status="candidate" since="1.1.0">
+
+### Callback Error Isolation
+
+A failing user callback **MUST NOT** break the host application. Where the language can recover from an error raised in a called function (`try`/`catch`, `rescue`, `recover`, `catch_unwind`), SDKs **MUST** invoke every user-provided callback inside that boundary, covering every non-local exit the language permits. A rejected promise, future, or task **MUST** be handled like a synchronous raise.
```
szokeasaurusrex, line 128: h: The section currently requires isolating every non-local exit a language permits, and names Rust's `catch_unwind` explicitly. I would like to narrow the requirement to failures the language defines as recoverable, and let intentionally fatal exits propagate.

The spec, as written, would produce unidiomatic and surprising results in the Rust and Python SDKs, and likely in other SDKs too, which is why I believe it should be changed.

### Why only handle "recoverable" errors?

- **Python already draws this line.** [sentry-python#6853](https://github.com/getsentry/sentry-python/pull/6853), the reference implementation for this spec, catches only `except Exception` in its callback isolation; intentionally fatal exceptions like `KeyboardInterrupt` and `SystemExit` propagate, and it would be surprising to users if we were to change that behavior. The spec should therefore encode that existing boundary rather than mandate more.
- **In Rust, a panic is not assessable as recoverable by the SDK.** Panics are Rust's mechanism for non-recoverable errors — including errors the user chose to treat as non-recoverable. Some panics may in principle be recoverable, but the SDK has no way to tell. Per the [`catch_unwind` documentation](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html), it is not intended as a general try/catch and is not guaranteed to catch all panics; additionally the documentation warns that handling a panic can itself cause a panic when the object wrapping the error goes out of scope.
- **Infallible callback types.** In Rust, callbacks are not typed as returning `Result`, so recoverable errors have no channel to bubble up through the callback — users must handle them within the callback. Languages with a similar mechanism could enforce this at compile time, and compile-time enforcement should be accepted as an alternative to runtime isolation where available.

### My proposed rewording

- SDKs **MUST** recover from recoverable errors that are raised by user callbacks (what is considered "recoverable" depends on the language). 
- SDKs **SHOULD NOT** attempt to recover from intentionally fatal exits, e.g. Python `BaseException` subclasses, Rust panics, C aborts 
- In typed languages, compile-time enforcement via infallible callback types is an accepted alternative to runtime isolation. SDKs which type callbacks as infallible, such that recoverable errors cannot escape the signature at compile time, do not need to handle anything at runtime: the only remaining failure modes are the intentionally fatal ones, which per the previous proposed point SHOULD NOT be recovered anyway.

develop-docs/sdk/foundations/client/hooks/index.mdx
```diff
@@ -100,6 +109,53 @@ Hook called with the breadcrumb (and on some platforms the hint) that allow the
...
+
+### Error-Sampler Hook
+
+SDKs **MAY** implement an `error_sampler` callback that decides the sample rate of an individual error event, mirroring `traces_sampler` for transactions.
+
+The callback receives the event and, on platforms that support it, the hint. It returns either a boolean, or a float between `0.0` and `1.0`. When `error_sampler` is configured, `sample_rate` **MUST** be ignored.
+
+This hook **MUST** only apply to `error` events. Unlike `before_send`, it is a sampling mechanism rather than a developer-controlled filter, so it **MUST** run at the sampling step of the [event pipeline](/sdk/foundations/client/#event-pipeline). An event it samples out **MUST** still update the session, and SDKs **MUST** record it with the `sample_rate` discard reason.
+
+</SpecSection>
```
szokeasaurusrex, line 122: m: I would suggest removing this and adding in a separate PR, if at all.

@cleptric told me a few weeks ago when we were considering adding this to the Rust SDK that we do not intend to add this hook to any more SDKs beyond the Python SDK, so I would lean towards not specifying it at all in the spec.

```suggestion
```

develop-docs/sdk/foundations/client/hooks/index.mdx
szokeasaurusrex: I agree, I like the idea of a separate reason

msonnb: [`callback_error`](https://develop.sentry.dev/sdk/telemetry/client-reports/#discard-reasons-1:~:text=callback_error,threw%20an%20error.) is now added as a dedicated discard reason 👍

develop-docs/sdk/foundations/client/hooks/index.mdx
```diff
@@ -100,6 +109,53 @@ Hook called with the breadcrumb (and on some platforms the hint) that allow the
...
+| `before_send` | Drop the event | `before_send`, category `error` |
+| `before_send_transaction` | Drop the transaction | `before_send`, category `transaction` (plus [span outcomes](/sdk/telemetry/client-reports/#span-outcomes)) |
+| `before_send_log` | Drop the log | `before_send`, category `log_item` |
+| `before_send_metric` | Drop the metric | `before_send`, category `trace_metric` |
+| `before_send_feedback` | Drop the feedback | `before_send`, category `feedback` |
+| `before_send_check_in` | Drop the check-in | `before_send`, category `monitor` |
+| `before_breadcrumb` | Drop the breadcrumb | None |
+| `before_send_span` | Emit the span, discarding the return value | None |
+| Event processor | Drop the event | `event_processor`, category of the item |
+| `traces_sampler` | Fall back as if unconfigured | `sample_rate` if sampled out |
```
szokeasaurusrex, line 145: "Fall back as if unconfigured" is a bit vague, as it could either mean we fall back to a `traces_sample_rate`, to a `0.0` sample rate, or to TwP deferred-sampling-decision behavior.

Imo we should fall back to disabled-tracing/TwP behavior in this case. This should happen regardless of whether `traces_sample_rate` is set, [because our spec states that `traces_sample_rate` is only used when `traces_sampler` is not defined](https://develop.sentry.dev/sdk/telemetry/traces/#sampling). Falling back to TwP allows other SDKs in the trace potentially to choose to sample the trace, which is not possible if we fall back to a sample rate of `0.0`.

develop-docs/sdk/foundations/client/hooks/index.mdx
szokeasaurusrex, line 147: Going along with my other suggestion, I would remove this, as this is not a callback we intend to add to other SDKs besides the Python SDK.

```suggestion
```

develop-docs/sdk/foundations/client/hooks/index.mdx
runningcode: +1 to this. This is also very true for java.
For example, we should not attempt to recover from `OutOfMemoryException` or `StackoverflowError` and other non-recoverable errors. Our codebase currently does this and it hides deeper underlying problems and makes it harder for customers to actually understand what the error was.

develop-docs/sdk/foundations/client/hooks/index.mdx
```diff
@@ -100,6 +109,53 @@ Hook called with the breadcrumb (and on some platforms the hint) that allow the
...
+
+</SpecSection>
+
+<SpecSection id="callback-error-isolation" status="candidate" since="1.1.0">
+
+### Callback Error Isolation
+
+A failing user callback **MUST NOT** break the host application. Where the language can recover from an error raised in a called function (`try`/`catch`, `rescue`, `recover`, `catch_unwind`), SDKs **MUST** invoke every user-provided callback inside that boundary, covering every non-local exit the language permits. A rejected promise, future, or task **MUST** be handled like a synchronous raise.
+
+On failure, SDKs **MUST** emit an error-level [internal log](/sdk/foundations/client/configuration/#debug-mode) naming the callback, **MUST NOT** re-throw, and **MUST NOT** capture the error as an event or attach it to the item. See [Never capture your own exceptions](/sdk/getting-started/principles/#never-capture-your-own-exceptions). SDKs **MAY** re-throw in a non-default test mode documented to surface SDK failures.
```
runningcode, line 130: > emit an error-level internal log

The docs there say "SDKs may log fatal error messages even when the debug mode is disabled." We should be clear if these are considered loggable or not.

Forgive me if I misinterpret here, but if these logs aren't being sent to Sentry then they are effectively invisible in production which begs the question what is the value of doing this?

develop-docs/sdk/foundations/client/hooks/index.mdx
runningcode: When a customer sees `callback_error`, what is the story on discovering the `callback_error` is happening and then understanding where it is happening?

runningcode (commented):
Thanks for putting this together and defining this behavior. I just discovered the existence of this!

develop-docs/sdk/foundations/client/hooks/index.mdx
dingsdax: agreed, changed the language around this

develop-docs/sdk/foundations/client/hooks/index.mdx
dingsdax: updated

develop-docs/sdk/foundations/client/hooks/index.mdx
dingsdax: Good question. We keep the log behind debug mode like every other internal log, and I changed the wording to clarify that. The production signal is the `callback_error` discard reason in Stats. Its category narrows down the callback, and enabling debug then shows exactly which one failed. 
We probably should follow up with user-facing docs that explain this path. Always-on internal logging feels wrong in this case.

develop-docs/sdk/foundations/client/hooks/index.mdx
```diff
@@ -100,6 +108,44 @@ Hook called with the breadcrumb (and on some platforms the hint) that allow the
...
 
+<SpecSection id="callback-error-isolation" status="candidate" since="1.1.0">
+
+### Callback Error Isolation
+
+A failing user callback **MUST NOT** break the host application. SDKs **MUST** catch every recoverable error a user-provided callback raises, for example with `try`/`catch` or `rescue`. A rejected promise, future, or task **MUST** be handled like a synchronous raise.
+
+SDKs **SHOULD NOT** recover from intentionally fatal exits, such as Python's `KeyboardInterrupt` and `SystemExit`, Java `Error` subclasses like `OutOfMemoryError`, Rust panics, or C aborts. Recovering from them hides the underlying failure. SDKs whose callback types can't raise recoverable errors, checked at compile time, don't need a runtime boundary.
+
+On failure, SDKs **MUST** emit an error-level [internal log](/sdk/foundations/client/configuration/#debug-mode) naming the callback, **MUST NOT** re-throw, and **MUST NOT** capture the error as an event or attach it to any telemetry. Like any internal log, it follows debug mode. See [Never capture your own exceptions](/sdk/getting-started/principles/#never-capture-your-own-exceptions). SDKs **MAY** offer an opt-in test mode that re-throws.
```
runningcode: ```SDKs **MAY** offer an opt-in test mode that re-throws.```

I agree but I think this should be a required part of the spec. I also think we should rename this to something like "strict mode". I could imagine that many teams would want this in production. For example, Square, has a policy to Crash Fast. Here's a talk about it: https://www.youtube.com/watch?v=9lwUYxTIb8k

runningcode (commented):
Thanks for the update, this looks a lot better.
This sounds to me like a breaking change and therefore we should wait until a next major version to implement it.

develop-docs/sdk/foundations/client/hooks/index.mdx
dingsdax: good point, added strict-mode wording, but did not make it a requirement yet, we can decide per SDK basis, feel free to implement for Java already

develop-docs/sdk/foundations/client/hooks/index.mdx
```diff
@@ -100,6 +108,46 @@ Hook called with the breadcrumb (and on some platforms the hint) that allow the
 
 </SpecSection>
 
+<SpecSection id="callback-error-isolation" status="candidate" since="1.1.0">
+
+### Callback Error Isolation
+
+A failing user callback **MUST NOT** break the host application. SDKs **MUST** catch every recoverable error a user-provided callback raises, for example with `try`/`catch` or `rescue`. A rejected promise, future, or task **MUST** be handled like a synchronous raise.
```
szokeasaurusrex: m: The "**MUST NOT** break" requirement is still stated a bit too broadly, I would specify that the requirement not to break applies for recoverable errors

```suggestion
A user callback failing due to a recoverable error **MUST NOT** break the host application. SDKs **MUST** catch every recoverable error a user-provided callback raises, for example with `try`/`catch` or `rescue`. A rejected promise, future, or task **MUST** be handled like a synchronous raise.
```

develop-docs/sdk/foundations/client/hooks/index.mdx
```diff
@@ -100,6 +108,46 @@ Hook called with the breadcrumb (and on some platforms the hint) that allow the
...
+
+SDKs **SHOULD NOT** recover from intentionally fatal exits, such as Python's `KeyboardInterrupt` and `SystemExit`, Java `Error` subclasses like `OutOfMemoryError`, Rust panics, or C aborts. Recovering from them hides the underlying failure. SDKs whose callback types can't raise recoverable errors, checked at compile time, don't need a runtime boundary.
+
+On failure, SDKs **MUST** emit an error-level [internal log](/sdk/foundations/client/configuration/#debug-mode) naming the callback, **MUST NOT** re-throw, and **MUST NOT** capture the error as an event or attach it to any telemetry. Like any internal log, it follows debug mode. See [Never capture your own exceptions](/sdk/getting-started/principles/#never-capture-your-own-exceptions).
+
+SDKs **SHOULD** offer an opt-in strict mode, disabled by default. In strict mode, a callback failure **MUST** propagate as an unhandled error of the host application instead of being isolated. SDKs **MUST NOT** pass the event reporting that error through the callback that failed.
```
szokeasaurusrex: l: Why are we adding this opt-in strict mode? I feel it makes the configuration a lot more complex from a user perspective.

I would either remove this entirely or make it an optional "MAY" rather than a recommended "SHOULD."

develop-docs/sdk/foundations/client/hooks/index.mdx
```diff
@@ -100,6 +108,46 @@ Hook called with the breadcrumb (and on some platforms the hint) that allow the
...
+| `before_send_span` | Emit the original span | None |
+| User-registered event processor | Drop the event | Category of the item |
+| `traces_sampler` | Use the parent's decision, otherwise defer | None |
+| `profiles_sampler` | Don't profile the transaction | None |
+
+Filters drop the item because a callback that failed part-way may have skipped a redaction. `before_send_span` keeps the span because it enriches spans; [`ignore_spans`](/sdk/telemetry/spans/filtering/) is what filters them.
+
+For a callback not listed here, SDKs **MUST** choose the outcome that sends less data. For example, if `before_capture_screenshot` fails, SDKs **MUST NOT** attach the screenshot.
+
+If `traces_sampler` fails, SDKs **MUST** use the parent's sampling decision. Without a parent, they **MUST** leave the decision to downstream SDKs, as with tracing disabled, and **MUST NOT** fall back to `traces_sample_rate`. If `profiles_sampler` fails, SDKs **MUST NOT** profile the transaction.
```
szokeasaurusrex: m:

> If `traces_sampler` fails, SDKs **MUST** use the parent's sampling decision.

I am not sure I agree with this principle. It contradicts what I think should be the overarching principle that "SDKs **MUST** choose the outcome that sends less data."

The reason is that, if the `traces_sampler` was written in a way that intentionally ignored the parent sampling decisions some or all of the time, in order to prevent some data from being sent, falling back to respecting the sampling decision due to an error might mean we send data that the user did not want us to send.

Therefore, I would suggest that a failing `traces_sampler` should result in us falling back to TwP mode: the sampling decision would be propagated to downstream SDKs, but we would not send any spans/transactions for that trace from the SDK where the sampler failed.

develop-docs/sdk/foundations/client/hooks/index.mdx
```diff
@@ -100,6 +108,46 @@ Hook called with the breadcrumb (and on some platforms the hint) that allow the
...
+
+SDKs **SHOULD** pass `before_send_span` a copy of the span, so they can emit the original if the callback fails. Otherwise, they **MUST** emit the span as it stands and skip the [`before_send_span` merge](/sdk/telemetry/spans/scrubbing-data/#scrubbing-data-with-beforesendspan).
```
szokeasaurusrex: m: This also contradicts the overall principle of not sending data when a callback fails. It seems especially risky to send the original span after a `before_send_span` failure, if that callback is meant to be a reliable place for folks to add data scrubbing.

develop-docs/sdk/foundations/client/hooks/index.mdx
```diff
@@ -100,6 +108,46 @@ Hook called with the breadcrumb (and on some platforms the hint) that allow the
...
+
+A failure **MUST NOT** produce a billed event of any type. SDKs **MUST** record every item a failure drops with the [`callback_error`](/sdk/telemetry/client-reports/#discard-reasons) discard reason and the dropped item's category, so users can tell a failure apart from an explicit drop.
```
szokeasaurusrex: m: I think we can greatly simplify this section in a way that will also address my previous comments:

```suggestion
When a user-provided callback fails due to a recoverable error, SDKs:
  - **MUST** emit an error-level [internal log](/sdk/foundations/client/configuration/#debug-mode) naming the callback.
  - **MUST NOT** re-throw or capture the error as an event.
  - **MUST** assume that the callback would have dropped/scrubbed as much data as possible.
  - **SHOULD** record a client report with reason `callback_error` for all data dropped as a result of the callback error.
```

Note that I also changed the client report from "MUST" to "SHOULD" as client reports are generally a best-effort thing (but "MUST" is also fine by me)

develop-docs/sdk/foundations/client/hooks/index.mdx
sentrivana: `before_send_span` is a special case. In span streaming, we don't want to allow users to drop individual spans from a trace -- `before_send_span` can only be used to scrub data from spans. You can't return `None` from it by design, so it follows you also can't work around the restriction by raising an error.

develop-docs/sdk/foundations/client/hooks/index.mdx
dingsdax: see the previous comment from @runningcode, some companies might prefer or need that; I think SHOULD is fine, it's not MUST

develop-docs/sdk/foundations/client/hooks/index.mdx
szokeasaurusrex: Got it, but in that case, I still feel we should then scrub everything from the span, because we otherwise may send data the user intended to remove.

develop-docs/sdk/foundations/client/hooks/index.mdx
```diff
@@ -100,6 +108,46 @@ Hook called with the breadcrumb (and on some platforms the hint) that allow the
...
```
szokeasaurusrex: m: I think we can greatly simplify this section in a way that will also address my previous comments:

```suggestion
When a user-provided callback fails due to a recoverable error, SDKs:
  - **MUST** emit an error-level [internal log](/sdk/foundations/client/configuration/#debug-mode) naming the callback.
  - **MUST NOT** re-throw or capture the error as an event.
  - **MUST** assume that the callback would have dropped/scrubbed as much data as possible.
  - **SHOULD** record a client report with reason `callback_error` for all data dropped as a result of the callback error.
```

Note that I also changed the client report from "MUST" to "SHOULD" as client reports are generally a best-effort thing (but "MUST" is also fine by me)

develop-docs/sdk/foundations/client/hooks/index.mdx
dingsdax: a `callback_error` count shows up under "Stats > Usage > Client Discard", broken down by project and data category. No alert or issue is created, so they only see it if they go looking.

the client report only carries the reason, category and count, not the callback; the data category narrows it down; to see which callback failed and why, customers have to turn on debug and read the SDK's internal log

develop-docs/sdk/foundations/client/hooks/index.mdx
```diff
@@ -100,6 +108,22 @@ Hook called with the breadcrumb (and on some platforms the hint) that allow the
...
+### Callback Error Isolation
+
+A user callback failing due to a recoverable error **MUST NOT** break the host application. SDKs **MUST** catch every recoverable error a user-provided callback raises, for example with `try`/`catch` or `rescue`. A rejected promise, future, or task **MUST** be handled like a synchronous raise.
+
+SDKs **SHOULD NOT** recover from intentionally fatal exits, such as Python's `KeyboardInterrupt` and `SystemExit`, Java `Error` subclasses like `OutOfMemoryError`, Rust panics, or C aborts. Recovering from them hides the underlying failure. SDKs whose callback types can't raise recoverable errors, checked at compile time, don't need a runtime boundary.
+
+When a user-provided callback fails due to a recoverable error, SDKs:
+  - **MUST** emit an error-level [internal log](/sdk/foundations/client/configuration/#debug-mode) naming the callback.
+  - **MUST NOT** re-throw or capture the error as an event.
+  - **MUST** assume that the callback would have dropped/scrubbed as much data as possible.
```
szokeasaurusrex: m: Per our discussion with @sentrivana, we should change this as follows:

```suggestion
  - **SHOULD** drop the data passed to the callback. For callbacks that do not support dropping data (e.g. `before_send_span), SDKs **SHOULD** instead send the data to Sentry unchanged.
```

PR #19194, opened 2026-08-31
ref(php): Trim the Laravel AI integration page to a pointer

chargome:
PHP does not need the restructure from #19152 / #19191 — there is only one AI page and the real content already lives under Agent Tracing. That page just restated it, so it is trimmed to a pointer.


chargome:
Closing — PHP doesn't need the restructure from #19152 / #19191, and the remaining cleanup here isn't worth a review on its own.

PR #19217, opened 2026-09-01
docs(javascript): Add v10 to v11 migration guide

andreiborza:
## DESCRIBE YOUR PR

Fills in the one-page v10 to v11 migration guide for the JavaScript SDKs, which was a placeholder before. The content follows the SDK's [`MIGRATION.md`](https://github.com/getsentry/sentry-javascript/blob/develop/MIGRATION.md).

- Long tables (span names, span ops, attribute renames) and optional details are in expandable sections
- Framework and runtime specific changes are filtered per guide, so each platform page only shows what applies to it

This is the bottom of a stack. #19076 (interactive guide) goes on top of it and adds the link to the interactive version.

## IS YOUR CHANGE URGENT?

## SLA

## PRE-MERGE CHECKLIST


docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1394 @@
...
+### Node.js
+
+The minimum supported Node.js version is **20.19.0**. Node.js 18 is no longer supported, and the supported range has gaps:
+
+| Node.js version  | Supported |
+| ---------------- | --------- |
+| `<20.19.0`       | No        |
+| `20.19.0` - `21` | Yes       |
+| `22.0` - `22.11` | No        |
+| `22.12` - `22.x` | Yes       |
```
JPeer264: m: `22.x` would also mean `22.9`, but we don't support it. This chart has a mix of version ranges and a range with `-` in it. Maybe we just add our actual ranges from the `package.json` and give couple of examples or we just keep the range. I guess the majority people (and all LLMs) are capable of reading the version ranges

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1394 @@
...
+
+Make sure to select your framework in the dropdown in the top left corner of the page.
+
+</Alert>
+</PlatformSection>
+
+Version 11 of the Sentry JavaScript SDK focuses on better OpenTelemetry interoperability, more flexible instrumentation, and better out-of-the-box defaults. The biggest changes are:
+
+- **OpenTelemetry interoperability:** The SDK no longer takes over your OpenTelemetry setup.
+- **Instrumentation:** Instrumentation now runs through diagnostics channels, which unlocks tracing on platforms like Vercel and Netlify, and on Cloudflare, Bun, and Deno.
```
JPeer264: Adding a link to diagnostics channels wouldn't harm, also the second sentence is a bit off

```suggestion
- **Instrumentation:** The instrumentations of the SDK now runs through [diagnostics channels](https://nodejs.org/api/diagnostics_channel.html#diagnostics-channel), which unlocks tracing on platforms like Vercel, Netlify and Cloudflare, but also runtimes like Bun and Deno.
```

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1394 @@
...
+- **Span streaming:** Stream mode is the new default, so spans no longer hit the size and volume limits of transactions.
```
JPeer264:
```suggestion
- **Span streaming:** [Stream mode](https://docs.sentry.io/platforms/javascript/tracing/streamed-spans/) is the new default, so spans no longer hit the size and volume limits of transactions.
```

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1394 @@
...
+- **Data collection:** `sendDefaultPii` is replaced by `dataCollection`, which controls each category of data separately and collects more by default.
```
JPeer264: l: Maybe a backlink to the blog post wouldn't harm

```suggestion
- **Data collection:** `sendDefaultPii` is replaced by `dataCollection`, which controls each category of data separately and collects more by default ([Read more](https://blog.sentry.io/datacollection-control-panel/) about `dataCollection`)
```

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1394 @@
...
+
+### Browsers
+
+Safari 14 is no longer supported. The minimum version is Safari 15. See [Browser Support](/platforms/javascript/#browser-support) for the full matrix.
+
+Browser projects also need Node.js 20.19.0 or higher if their build steps run on Node.js.
+
+</PlatformCategorySection>
+
+<PlatformSection supported={["javascript.deno"]}>
```
JPeer264, line 64: l: Do we really want to have `PlatformSection`s for the migration guide? Seems a bit odd to fetch multiple pages to just have one guide. I'd rather remove it to have one migration guide?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1394 @@
...
+
+### Deno
```
JPeer264, line 66: q: Did Bun change too?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1394 @@
...
+| Astro                         | 4               |
+| React Router (framework mode) | 7.15            |
+| Remix (`@remix-run/node`)     | 2               |
+| Fastify                       | 3.21            |
+
+<PlatformSection supported={["javascript.aws-lambda"]}>
+
+### AWS Lambda Layer
+
+The version 11 layer is published as `SentryNodeServerlessSDKv11`. Its ARN will be listed in the <PlatformLink to="/install/layer/">Lambda Layer docs</PlatformLink> once it's available.
```
JPeer264: l: Not sure if there is a "version 11 layer"

```suggestion
The layer for v11 is published as `SentryNodeServerlessSDKv11`. Its ARN will be listed in the <PlatformLink to="/install/layer/">Lambda Layer docs</PlatformLink> once it's available.
```

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1394 @@
...
+
+The layer supports the `nodejs20.x`, `nodejs22.x`, and `nodejs24.x` runtimes. Move functions that still run on `nodejs18.x` to a newer runtime before you upgrade.
+
+Updates and fixes for version 10 are published as `SentryNodeServerlessSDKv10`.
```
JPeer264: q: Should that be part of a migration guide?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1394 @@
...
+
+</PlatformSection>
+
+## Data Collection
+
```
JPeer264, line 127: l: Wouldn't harm to link again to the blog post in form of an alert or so.

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+We raised the minimum supported versions of several frameworks and libraries:
+
+| Framework                     | Minimum version |
+| ----------------------------- | --------------- |
+| Next.js                       | 14              |
+| React                         | 17              |
+| Astro                         | 4               |
+| React Router (framework mode) | 7.15            |
+| Remix (`@remix-run/node`)     | 2               |
```
chargome: We already dropped remix v1 in a prev major, this line can be removed

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+  );
+```
+
+</PlatformSection>
+
+<PlatformSection supported={["javascript.nextjs"]}>
+
+## Next.js Changes
+
+The build-time configuration moved and dropped its long-deprecated options.
```
chargome: long-deprecated seems off

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+// After
+import { withSentryConfig } from "@sentry/nextjs/config";
+```
+
+The `SentryBuildOptions` type moved with it, and the no-op `withSentryConfig` passthroughs of the client and edge builds were removed.
+
+### Removed Build Options
+
+The long-deprecated top-level options of `withSentryConfig` were removed. Most of them moved under `webpack` in v10:
```
chargome: Same here, would get rid of long-deprecated

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+});
+```
+
+The bundler-specific `webpack.reactComponentAnnotation` and `_experimental.turbopackReactComponentAnnotation` options still work, but are deprecated and will be removed in v12. If both are set, the bundler-specific one wins for that bundler.
+
+On Turbopack, component annotation requires Next.js 16 or higher. The SDK warns at build time when annotation is enabled on an older version, where it previously did nothing.
+
+### Other Next.js Changes
+
+- Tracing was removed from the generated Pages Router API handler, Edge API handler, and Middleware wrapper templates. Route handlers and middleware are still instrumented automatically, so no action is required for most setups.
```
chargome: I think we can omit this line here

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1394 @@
...
+
+## Tracing
+
+Span streaming changes how spans are sent, named, and filtered. Most of this isn't caught by TypeScript, so check your filters, dashboards, and alerts.
+
+### Span Streaming Is the Default
+
+Spans are now sent in batches as they finish (<PlatformLink to="/tracing/streamed-spans/">stream mode</PlatformLink>), instead of being buffered until the root span ends. Spans are no longer capped at 1000 per transaction, and individual payload limits are higher.
+
+Because no transaction events are produced, `beforeSendTransaction` and `ignoreTransactions` no longer do anything, and `beforeSendSpan` receives a different payload.
```
JPeer264: l: I feel like that this should rather go into some alert box.

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+The minimum supported Deno version is **2.8.3**.
+
+</PlatformSection>
+
+### TypeScript
+
+The minimum required TypeScript version is **5.0.4**. The SDK no longer ships down-leveled types. Older TypeScript versions _may_ continue to work, but no guarantees apply.
+
+### Frameworks and Libraries
```
mydea: should we show this section here so generically? What about, instead of this, just showing a line in the respective framework, e.g. in Astro

> Raised minimum support of Astro to 4

so you see what actually applies to the current guide?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
JPeer264: l: 

> receives a different payload

We should link to the payload down below.

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+</PlatformSection>
+
+## Data Collection
+
+`sendDefaultPii` is replaced by `dataCollection`, which controls each category of collected data separately.
+
+<Alert level="warning">
+
+This is a behavior change, not a rename. In v10, an unset `sendDefaultPii` was restrictive. In v11, an unset `dataCollection` collects most categories.
```
mydea: I love this message, but would reword it slightly:


```suggestion
This is a behavior change, not a rename. In v10, an unset `sendDefaultPii` was restrictive. In v11, an unset `dataCollection` collects everything by default.
```

plus mentioning somewhere the sensitive information like auth tokens are stripped on a best-effort basis ?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+</Expandable>
+
+See the <PlatformLink to="/configuration/options/#dataCollection">`dataCollection` option</PlatformLink> for the full list of categories.
+
+<PlatformCategorySection supported={['server', 'browser']}>
+
+## Tracing
+
+Span streaming changes how spans are sent, named, and filtered. Most of this isn't caught by TypeScript, so check your filters, dashboards, and alerts.
```
mydea: before this, maybe add a sentence that we are moving to span streaming, and what does that mean?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1394 @@
...
+| `queue.receive`                                                          | The kafkajs operation name (`poll my-topic`)                                        | The operation type and destination (`receive my-topic`)                                                                                   |
+| `cache.get`, `cache.put`, `cache.remove`                                 | The cache keys (`user:123`), or for dataloader the operation and loader name        | The cache operation (`cache.get`, `cache.put`, `cache.remove`)                                                                            |
+| `db`, `db.query` (SQL)                                                   | The statement the driver ran (`SELECT * FROM "User" WHERE id = $1`)                 | A summary of it (`SELECT "User"`), or, without a statement, the operation and table, the namespace, or the database system (`postgresql`) |
```
JPeer264: l: If we already have `"User"` in quotes then we should put `id` also in quotes

```suggestion
| `db`, `db.query` (SQL)                                                   | The statement the driver ran (`SELECT * FROM "User" WHERE "id" = $1`)                 | A summary of it (`SELECT "User"`), or, without a statement, the operation and table, the namespace, or the database system (`postgresql`) |
```

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1394 @@
...
+
+</PlatformSection>
+
+<PlatformSection supported={["javascript.cloudflare"]}>
+
+## Cloudflare Changes
+
+Workers need a new compatibility flag, and a few APIs moved or became automatic.
+
+### `nodejs_compat` Is Required
```
JPeer264: q: It appears that having a compatibility date of `2026-08-04` we don't need to set any compatibility flag as `nodejs_compat` is then on by default. Is it worth to mention in the guide?

https://developers.cloudflare.com/workers/runtime-apis/nodejs/#get-started

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+// Applied to a single operation
+Sentry.withScope((scope) => {
+  scope.setAttribute("checkout.step", "payment");
+});
+```
+
+#### `beforeSendSpan` Receives the Streamed Span Format
+
+The callback now runs as each span finishes and receives a `StreamedSpanJSON` object. Its fields were renamed:
```
mydea, line 258: missing here is that this is the case for streamed spans but not for trace lifecycle, which still get the old behvaior but now need a function wrapper instead for type safety, right?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+      span.attributes["db.query.text"] = scrub(
+        span.attributes["db.query.text"]
+      );
+    }
+    return span;
+  },
+});
+```
+
+Streamed spans always have a status, and finer-grained error statuses are mapped to `'error'`. Additional error information is available in span attributes such as `sentry.status.message`.
```
mydea: what is `error` and where is this?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+A `beforeSendSpan` callback that does not match your `traceLifecycle` is never invoked. Turn on debug logging to see a warning about the mismatch.
+
+#### Replacing `beforeSendTransaction` and `ignoreTransactions`
+
+Move data changes to `beforeSendSpan` and guard on `is_segment` to target what used to be the transaction. Use `ignoreSpans` to drop spans, because `beforeSendSpan` can't drop them.
```
mydea:
```suggestion
Move data changes to `beforeSendSpan` and guard on `is_segment` to target what used to be the transaction. You can use `ignoreSpans` to drop spans - `beforeSendSpan` can only mutate spans, not drop them.
```

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+Move to a supported version everywhere the SDK is installed or your app is built: Dockerfiles, CI images, `engines`, `.nvmrc`, and serverless runtime settings.
+
+</PlatformSection>
+</PlatformCategorySection>
+
+<PlatformCategorySection supported={['browser']}>
+
+### Browsers
+
+Safari 14 is no longer supported. The minimum version is Safari 15. See [Browser Support](/platforms/javascript/#browser-support) for the full matrix.
```
msonnb: `#browser-support` doesn't exist anymore on this page. maybe we should link to (and update) [this](https://sentry-docs-git-ab-js-v11-migration-guide.sentry.dev/platforms/javascript/troubleshooting/supported-browsers/) instead?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+Sentry.init({
+  ignoreSpans: [
+    { name: "GET /health", attributes: { "sentry.op": "http.server" } },
+  ],
+});
+```
+
+Scope `tags` and `extra` [aren't carried over to streamed spans](#scope-tags-and-extra-arent-applied-to-spans) either.
+
+#### `spanToJSON` Returns the Streamed Span Format
```
mydea: this seems pretty internal, IMHO this is not an API I would expect users to use/care about?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+`Sentry.spanToJSON()` now returns a `StreamedSpanJSON` object, the same shape `beforeSendSpan` receives. The `spanToStreamedSpanJSON` helper was removed, so replace calls to it with `spanToJSON`.
+
+<Expandable title="Staying on transaction mode">
```
mydea: this should IMHO move to the very top -  I would structure it as:

```
## Tracing

[short description of the changes]

[this expandable here]

then the rest
```

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+Transaction mode only exists for backwards compatibility and will be removed in a future major version. Treat it as a temporary measure.
+
+</Expandable>
+
+### Span Names Are Low Cardinality
+
+In stream mode, span names follow the [Sentry span name conventions](https://getsentry.github.io/sentry-conventions/names/). URLs, route parameters, and other client-supplied values are no longer part of the name. They stay available as span attributes.
+
+- `http.client`: `GET https://api.example.com/users/123` becomes `GET api.example.com`. The full URL stays on `url.full`, the domain on `url.domain`.
```
mydea: this is a bit hard to read xD  maybe we can reformat this a bit

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+- `http.server`: unchanged when the SDK resolves a route (`GET /users/:id`). Requests without a route are named after the method alone (`GET`).
+- `pageload` and `navigation`: the parameterized route, or `Pageload` and `Navigation` when there is none.
+- `resource.*`: the resource domain, or `Resource`.
+- `graphql`: `GraphQL query`, `GraphQL parse`. The operation name stays on `graphql.operation.name` and the phase on the new `graphql.processing.type` attribute.
+- Messaging spans read `<operation type> <destination>` in every integration, for example `process my-queue`.
```
mydea: what does messaging spans mean, and the "in every integration" part is IMHO more confusing than helpful. I'd rephrase this to e.g. "this type of span XX becomes YY, for example ZZZ

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+- Database spans are named after a summary of the query (`SELECT "User"`), not the statement. The sanitized statement stays on `db.query.text`.
```
mydea: this should state the fallback too, because in many cases this does not exist I believe? (Or did we change this, not sure)

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+<Expandable title="Full list of span name changes">
+
+| Span op                                                                  | Before                                                                              | After                                                                                                                                     |
+| ------------------------------------------------------------------------ | ----------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
+| `pageload`, `navigation`                                                 | The route, or the raw URL path (`/users/123`)                                       | The route, or `Pageload` and `Navigation`                                                                                                 |
```
mydea: let's go over this table and make sure that breaks are applied in a reasonable way. right now the first column is super long which makes the table hard to read.

mydea: esp. this: mcp.notification.client_to_server, mcp.notification.server_to_client

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+- The `pg`, `postgres.js`, `mysql`, `mysql2`, `knex`, `tedious`, Prisma, Nitro `db0`, and Cloudflare D1 instrumentations name their query spans the same way now. A statement that touches no table (`SELECT NOW()`) summarizes to the bare operation. Connect and pool spans (`pg.connect`, `mysql2.connect`, `redis-connect`, `generic-pool.acquire`) keep their names.
+- Supabase query spans drop the builder call from the name, and the modifiers stay on `db.query`. Auth spans are named after the method they call (`auth.admin.createUser`).
+- AWS SQS `SendMessage`, `SendMessageBatch`, and `ReceiveMessage`, and SNS `Publish`, are messaging spans now instead of `rpc` spans. Every other command on those clients stays `rpc`.
+- Child spans carry their segment's name in `sentry.segment.name`, so that changes with it.
+
+</Expandable>
+
+<Alert level="warning">
+
+`ignoreSpans` and `tracesSampler` both run when a span **starts**, so a span may not have its final name yet. Name-based rules stop matching silently. Match on attributes instead:
```
mydea: > Name-based rules stop matching silently. 

what does that mean? 🤔 I think we can just omit this sentence?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+  tracesSampler: ({ name, inheritOrSampleWith }) =>
+    inheritOrSampleWith(name === "GET /health" ? 0 : 1),
+
+  // After
+  ignoreSpans: [
+    { attributes: { "sentry.op": "pageload", "url.path": "/health" } },
+    { attributes: { "db.query.text": "SELECT * FROM health_check" } },
+  ],
+  tracesSampler: ({ attributes, inheritOrSampleWith }) =>
+    inheritOrSampleWith(attributes?.["url.path"] === "/health" ? 0 : 1),
```
mydea: just reading this, can we change this so this always passes attributes, would make API usage easier!

mydea: https://github.com/getsentry/sentry-javascript/pull/24153

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+  ],
+});
+```
+
+`browserTracingIntegration` no longer captures `performance.mark()` and `performance.measure()` spans by default, and no longer accepts an `_experiments` object at all. The `idleTimeout`, `finalTimeout`, and `childSpanTimeout` options of interaction spans are configured on `interactionsIntegration` now, with the same defaults as before.
+
+Web vitals also changed:
+
+- CLS and LCP are recorded as measurements on the pageload span, or as dedicated spans in stream mode.
+- INP is always sent as a web vital span, carrying its value in the `browser.web_vital.inp.value` attribute instead of as a span measurement. Update dashboards and alerts that read it as a measurement.
```
mydea:
```suggestion
- INP is always sent as a web vital span, carrying its value in the `browser.web_vital.inp.value` attribute instead of as a span measurement. Update custom dashboards and alerts that read it as a measurement (built-in dashboards do not need adjustments).
```

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+- `http.server`: unchanged when the SDK resolves a route (`GET /users/:id`). Requests without a route are named after the method alone (`GET`).
+- `pageload` and `navigation`: the parameterized route, or `Pageload` and `Navigation` when there is none.
+- `resource.*`: the resource domain, or `Resource`.
+- `graphql`: `GraphQL query`, `GraphQL parse`. The operation name stays on `graphql.operation.name` and the phase on the new `graphql.processing.type` attribute.
+- Messaging spans read `<operation type> <destination>` in every integration, for example `process my-queue`.
+- Database spans are named after a summary of the query (`SELECT "User"`), not the statement. The sanitized statement stays on `db.query.text`.
+
+<Expandable title="Full list of span name changes">
+
+| Span op                                                                  | Before                                                                              | After                                                                                                                                     |
```
msonnb: can we control the width of these columns? "Span op" is so wide that it isn't obvious that there's a third column. See:

Image

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+<PlatformCategorySection supported={['server']}>
+
+## OpenTelemetry
+
+Sentry and OpenTelemetry are separate pipelines now, and you choose how they connect.
+
+### The SDK No Longer Sets Up OpenTelemetry
+
+By default, the SDK no longer registers an OpenTelemetry tracer provider. It owns the full span lifecycle, produces native Sentry spans, and ignores spans created through `@opentelemetry/api`. That leaves `@opentelemetry/api` as the only OpenTelemetry dependency, so you can run your own setup next to Sentry without Sentry spans leaking into your pipeline, and without using Sentry components for export, context management, and trace propagation.
```
mydea:
```suggestion
By default, the SDK no longer registers an OpenTelemetry tracer provider. It owns the full span lifecycle, produces native Sentry spans, and ignores spans created through `@opentelemetry/api`. You can now run your own OpenTelemetry setup next to Sentry without Sentry spans leaking into your pipeline, and without using Sentry components for export, context management, and trace propagation.
```

IMHO not necessary to mention this here, may be confusing why this still exists as a dependency?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+If you only use the Sentry SDK, tracing works as before and there's nothing to do.
+
+`skipOpenTelemetrySetup` is replaced by `enableOpenTelemetrySetup`. Turn it on and the SDK registers a minimal OpenTelemetry-compatible tracer provider, context manager, and propagator, which turns spans from `@opentelemetry/api` into Sentry spans. It's off by default, except on `@sentry/nextjs` and `@sentry/sveltekit`, which need it to capture the spans those frameworks emit.
```
mydea:
```suggestion
`skipOpenTelemetrySetup` is replaced by `enableOpenTelemetrySetup`. Turn it on and the SDK registers a custom OpenTelemetry tracer provider, context manager, and propagator, which turns spans from `@opentelemetry/api` into Sentry spans. It's off by default, except on `@sentry/nextjs` and `@sentry/sveltekit`, which need it to capture the spans those frameworks emit.
```

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+```js
+// Before
+Sentry.init({ skipOpenTelemetrySetup: true });
+
+// After
+Sentry.init({ enableOpenTelemetrySetup: false });
+```
```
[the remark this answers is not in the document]
mydea: IMHO the example is weird because it infers you should set this to `false` to replace the previous `skipOpenTelemetrySetup` which is not what you need to do. I think this example can probably be omitted, a single example does not explain anything here 😅

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+```js
+Sentry.init({
+  dsn: "___PUBLIC_DSN___",
+  tracesSampleRate: 1.0,
+  enableOpenTelemetrySetup: true,
+});
+```
+
+This is not a general OpenTelemetry pipeline: no exporter, no OTLP output. Sentry also refuses to register its provider if you already registered one of your own, and the `@sentry/cloudflare/request` entry point doesn't support the option at all.
```
mydea, line 640: we should explain here when you'd want to do this. E.g. "This is meant to be used if you are using other libraries or code that natively emits open telemetry spans which you'd like to capture in Sentry. It is not meant to be used if you are using your own OTEL setup already" or something along these lines

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+Sentry.init({
+  integrations: [
+    Sentry.httpIntegration({
+      ignoreIncomingRequests: (urlPath) => urlPath.startsWith("/admin"),
+    }),
+  ],
+});
+```
+
+### `onUnhandledRejectionIntegration` Prints Less in `strict` Mode
```
JPeer264: l: Not really a PR issue (but could be fixed by changing it), but when a headline starts with a inline code block then it is hard to see that this is a new headline. Especially after a code block

Image

JPeer264 (commented):
Now `@sentry/node-core/light` is only mentioned once. Should we add more info about it in [Package Changes](https://sentry-docs-git-ab-js-v11-migration-guide.sentry.dev/platforms/javascript/guides/nextjs/migration/v10-to-v11/#package-changes) as well?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+</Expandable>
+
+## Instrumentation
+
+How the SDK hooks into your app changed, along with the way some frameworks are set up.
+
+### Channel-Based Instrumentation Is the Default
+
+Instrumentation now runs through diagnostics channels (`orchestrion`) instead of `import-in-the-middle`. This was opt-in in v10. It allows instrumentation at run time and build time, which enables tracing on deployment targets like Vercel and Netlify, and on non-Node.js runtimes like Cloudflare, Bun, and Deno. Most setups need no changes.
```
mydea:
```suggestion
Instrumentation now runs through diagnostics channels (`orchestrion`) instead of `import-in-the-middle`. It allows instrumentation at run time and build time, which enables tracing on deployment targets like Vercel and Netlify, and on non-Node.js runtimes like Cloudflare, Bun, and Deno. Most setups need no changes.
```

IMHO that was not really public before?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+The `vercelAIIntegration` is the exception. It no longer works on Vercel Edge, which doesn't support diagnostics channels, and it no longer enhances the native OpenTelemetry spans the `ai` package emits. Those spans are captured as they are, so some agent monitoring capabilities are lost.
+
+<PlatformSection notSupported={["javascript.deno", "javascript.bun", "javascript.cloudflare", "javascript.wasm"]}>
+
+### Initializing With `--require` Is No Longer Supported
+
+Node.js re-runs `--require` preloads on the internal module loader thread that the SDK spawns when it installs its instrumentation hooks, so `Sentry.init()` ran a second time on a thread that never executes your code. The SDK now skips initialization on that thread and warns when it was loaded through `--require`.
+
+Use [`--import`](https://nodejs.org/api/cli.html#--importmodule) instead. It also works for CommonJS apps, because the instrument file's extension decides how it loads.
```
mydea:
```suggestion
Use [`--import`](https://nodejs.org/api/cli.html#--importmodule) instead - it works for both ESM as well as CommonJS apps.
```

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+        return Number(error.statusCode ?? 500) >= 400;
+      },
+    }),
+  ],
+});
+```
+
+By default, Express captures 5xx errors and errors without a resolvable status, but not 3xx and 4xx errors. To capture errors yourself instead, set `expressIntegration({ shouldHandleError: false })` and call `Sentry.captureException` in your own error-handling middleware.
+
+`fastifyIntegration` is a single, channel-based plugin now that instruments Fastify 3.21 through 5, including error capture.
```
mydea: not sure, is this relevant? IMHO this should not matter to users?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+### Incoming HTTP Span Hooks Moved to `onSpanCreated`
```
mydea: I'd have thought this is not soo important, I would de-prioritize this - I think this can just be in e.g. "other changes"...?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+| Removed option                           | Replacement          |
+| ---------------------------------------- | -------------------- |
+| `trackIncomingRequestsAsSessions`        | `sessions`           |
+| `maxIncomingRequestBodySize`             | `maxRequestBodySize` |
+| `ignoreIncomingRequestBody`              | `ignoreRequestBody`  |
+| `dropSpansForIncomingRequestStatusCodes` | `ignoreStatusCodes`  |
+| `incomingRequestSpanHook`                | `onSpanCreated`      |
+
+`ignoreStatusCodes` is deprecated on `httpIntegration` and `httpServerSpansIntegration`, and will be removed in v12 without a replacement. It filters finished transaction events, so it only has an effect with `traceLifecycle: 'static'`. In stream mode, child spans are sent before the response status is known. To keep requests out of Sentry, decide before they're instrumented, with `tracesSampler` or `ignoreIncomingRequests`:
```
mydea:
```suggestion
`ignoreStatusCodes` is deprecated on `httpIntegration` and `httpServerSpansIntegration`, and will be removed in v12 without a replacement. It filters finished transaction events, so it only has an effect with `traceLifecycle: 'static'`. In stream mode, child spans are sent before the response status is known. To keep requests out of Sentry, decide before they're instrumented, with `tracesSampler`, `ignoreSpans` or `ignoreIncomingRequests`:
```

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+
+## Build Options
+
+The bundler plugin options of the meta-framework SDKs are all first-class build options now.
+
+### `sourceMapsUploadOptions` Was Removed
+
+Move its fields to the top level of the SDK's build options. A few fields changed on the way:
+
+- `url` is `sentryUrl` (Nuxt, SvelteKit).
```
mydea: we should only show this in relevant places/guides

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+- AI integrations are no longer available in the browser SDK. They remain available in the server-side SDKs.
+
+</PlatformCategorySection>
+
+<PlatformCategorySection supported={['server']}>
+<PlatformSection notSupported={["javascript.deno", "javascript.cloudflare", "javascript.wasm"]}>
+
+### Node.js and Server SDKs
+
+- The `@sentry/node/init` and `@sentry/node/preload` entry points were removed. Create your own instrument file that calls `Sentry.init()` and preload it with `node --import ./instrument.mjs app.js`.
```
mydea: I think you can also just import `@sentry/node/import` ? That should work like the former preload, but better (no config needed etc)

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+- The `loader` entry points were removed from every server SDK. Use `node --import @sentry/<sdk>/import` instead, for example `node --import @sentry/node/import`.
```
mydea: what was the loader entrypoint? 🤔 I don't see this anywhere in docs.

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+- `preloadOpenTelemetry()` was removed. Channel-based instrumentation is set up when the instrumented module loads, so preloading is no longer needed.
```
mydea: this is replaced by build-time instrumentation _or_ `@sentry/node/import` import, actually, I think?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1379 @@
...
+- The `registerEsmLoaderHooks` option was removed. The SDK no longer registers `import-in-the-middle` ESM loader hooks.
+- `generateInstrumentOnce` and `SentryContextManager` are no longer exported.
+- The deprecated `SentryHttpInstrumentation` and `SentryNodeFetchInstrumentation` exports were removed. Use `instrumentHttpOutgoingRequests()` and `nativeNodeFetchIntegration` instead.
+- The deprecated `honoIntegration` was removed. Use the [`@sentry/hono`](https://www.npmjs.com/package/@sentry/hono) SDK instead.
+- The `connect` instrumentation and the deprecated `prismaInstrumentation` option were removed. Prisma works out of the box.
+- The `OpenTelemetryServerRuntimeOptions` type was removed. `enableOpenTelemetrySetup` is part of the SDK-specific options types, such as `NodeOptions`.
+- OpenTelemetry resources are no longer collected. `contexts.otel.resource` was dropped from events, and the SDK no longer reads `OTEL_SERVICE_NAME` and `OTEL_RESOURCE_ATTRIBUTES`.
+- `getTraceContextForScope` and `getSentryResource` were removed from `@sentry/opentelemetry`, along with the `@opentelemetry/core` peer dependency.
```
mydea, line 1373: this is a bit misleading because the opentelemetry package does not exist anymore at all xD which is actually not mentioned at all in the guide, should be somewhere instead of this.e g.g "this package was removed, the following exports are still available from somewhere else: x, y, the others have been removed"?

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in c0697b7e02

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in d6ef11e36b

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in d6ef11e36b

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in 0b7552226a

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Yep, updated in e32dbc5450.

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in 42af812114

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in 5071d02db4. I also made sure we don't use the term segment span to match how we communicate it in https://docs.sentry.io/platforms/javascript/tracing/streamed-spans/

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in 87d21cd5e2

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in 87d21cd5e2

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in 04bf54f76f

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in 05b19493f2

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: It's in our public API reference (`configuration/apis.mdx`: `spanToJSON(span: Span): SpanJSON`), and the v7 → v8 guide pointed people at it for reading span fields (`span.status` → `spanToJSON(span).status`, `span.data` → `spanToJSON(span).data`, …). So anyone who followed that advice gets a different shape back in v11, and TypeScript won't always catch it at the call site.

Happy to drop the section if you think that audience is small, but I'd lean towards keeping it.

Unrelated to this PR: `apis.mdx` still documents the `SpanJSON` return type and doesn't mention `spanToStaticSpanJSON`. I'll fix that separately.

andreiborza: We used to tell users to use `spanToJSON` to read data from spans, since the shape changed, I think this is still relevant for them.

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: I reformatted this into a table, updated in 8766466a5b

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Dropped that phrasing. The bullets became a table (from your other comment), and the row now names the ops with examples instead. The per-integration detail (amqplib, kafkajs, NestJS BullMQ) stays in the expandable, where naming them is actually useful.

Updated in 9958ec7811.

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in 75951281ef

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in 071162a55e

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in 071162a55e

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in dda1381578

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in b037fdaf29

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in c225255656 and added a code example by which users can identify it. Let me know if you think it's too much and I can drop it.

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Removed the anchor, but the matrix itself documents Safari 14 which is current v10 so linking there is not quite right. We can update it once v11 is current. Updated in ad2de72b4b.

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in 3f5f9f747b

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: See https://github.com/getsentry/sentry-docs/pull/19217#discussion_r3948252708. I can see a point for both cases. I think it's fine, if you are selecting deno you see deno specific stuff.

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Nope

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Dropped the sentence in e841fc4a44

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Added in e6b243f740

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in fc47948cfe

andreiborza: Updated in ec5a6fa115

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in f9b3de59ce

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Added in d5c83647cf

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Good point, reworded every heading that started with inline code so they all start with a word now. Updated in 1b63eb3881

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in d32b90bbf3

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in 8fb2830943

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in ab8412eb5d

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in 42614bd7e0

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in 42614bd7e0

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Updated in 42614bd7e0

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Which package was removed? We still have `@sentry/opentelemetry.

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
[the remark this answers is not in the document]
andreiborza: Removed in 8092ef2971

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1463 @@
...
+## Version Support Changes
+
+Version 11 has new compatibility ranges for runtimes, frameworks, and libraries.
+
+<PlatformCategorySection supported={['server']}>
+<PlatformSection notSupported={["javascript.deno", "javascript.bun", "javascript.cloudflare", "javascript.wasm"]}>
+
+### Node.js
+
+Node.js 18 is no longer supported. The supported versions are **`>=20.19.0`**, **`>=22.12.0`**, and **`>=23.2.0`**.
```
JPeer264: m: Tbh that reads a bit odd. With `>=20.19.0` we mean "everything in the 20 range", but `>=23.2.0` we mean "everything in the 23 range but also later". So being explicit wouldn't harm either


```suggestion
Node.js 18 is no longer supported. The supported versions are **`>=20.19.0 <22.0.0 || >=22.12.0 <23.0.0 || >=23.2.0`**.
```

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1463 @@
...
+
+## Cloudflare Changes
+
+Workers need a new compatibility flag, and a few APIs moved or became automatic.
+
+### Node.js Compatibility Is Required
+
+The SDK requires the `nodejs_compat` compatibility flag instead of `nodejs_als`. Update your `wrangler.toml` or `wrangler.jsonc`:
+
+```diff
```
JPeer264: the normal `diff` seems to be broken it needs curly braces

```suggestion
```json {diff} {filename:wrangler.jsonc}
```
```

File: docs/platforms/javascript/common/migration/v10-to-v11/index.mdx [hunk 51]
```diff
@@ -0,0 +1,1463 @@
...
+
+```js
+sentryCloudflareVitePlugin({ autoInstrumentation: false });
+```
+
+### RPC Trace Propagation Is Explicit
+
+The `enableRpcTracePropagation` option was removed, and trace context is no longer appended to every RPC call on `env`. List the bindings you call in `rpcTracePropagationBindings` instead:
+
+```diff
```
JPeer264: the normal `diff` seems to be broken it needs curly braces

```suggestion
```js {diff}
```
```

File: docs/platforms/javascript/common/migration/v10-to-v11/index.mdx [hunk 52]
```diff
@@ -0,0 +1,1463 @@
...
+### Other Cloudflare Changes
+
+- `wrapRequestHandler` moved to the `@sentry/cloudflare/request` subpath.
+- `instrumentD1WithSentry` was removed. `withSentry()` instruments all D1 bindings on `env` automatically.
+- The `instrumentPrototypeMethods` option of `instrumentDurableObjectWithSentry` was removed. Prototype methods are always wrapped now, so remove the option.
+- `honoIntegration` was removed. Use the [`@sentry/hono`](https://www.npmjs.com/package/@sentry/hono) middleware instead.
+- Rate limiter bindings (`env.MY_RATE_LIMITER.limit()`) no longer create a span. Remove any dashboard, alert, or `ignoreSpans` entry that references the `rpc` span with `rpc.service: cloudflare.rate_limit`.
+- The `env` types and the generics on `withSentry` and `instrumentDurableObjectWithSentry` were reworked. If you don't pass explicit generic type parameters, nothing changes.
+
+```diff
```
JPeer264: the normal `diff` seems to be broken it needs curly braces

```suggestion
```js {diff}
```
```

File: docs/platforms/javascript/common/migration/v10-to-v11/index.mdx [hunk 53]
```diff
@@ -0,0 +1,1463 @@
...
+
+Workers enable `nodejs_compat` by default from a `compatibility_date` of `2026-08-04`, so you only need the flag on earlier dates.
+
+Because Node.js compatibility is required for everyone, the `@sentry/cloudflare/nodejs_compat` subpath export was removed. Import from `@sentry/cloudflare` instead.
+
+### The Vite Plugin Auto-Instruments Your Worker
+
+`sentryCloudflareVitePlugin()` wraps your Worker entry at build time, along with any Durable Object, Workflow, or WorkerEntrypoint class listed in your wrangler config. Entries you wrapped yourself are left untouched, so most setups need no changes. The experimental `_experimental.autoInstrumentation` opt-in was removed, and you can opt out with the top-level `autoInstrumentation` option:
+
+```js
```
JPeer264:
```suggestion
```js {filename:vite.config.mjs}
```

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -0,0 +1,1463 @@
...
+sentryCloudflareVitePlugin({ autoInstrumentation: false });
+```
+
+### RPC Trace Propagation Is Explicit
+
+The `enableRpcTracePropagation` option was removed, and trace context is no longer appended to every RPC call on `env`. List the bindings you call in `rpcTracePropagationBindings` instead:
```
JPeer264, line 971: l: Maybe worth to mention that users with our Vite plugin can simply remove `enableRpcTracePropagation: true` and only allow list external services, or simply link to the option: https://docs.sentry.io/platforms/javascript/guides/cloudflare/tracing/distributed-tracing/#rpc-trace-propagation

andreiborza: Added in cd1a650b24

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Technically true but nobody uses these uneven node versions and they go EOL way quicker anyway. I think it's fine as is.

JPeer264: > nobody uses these uneven node versions 

famous last words

JPeer264 (approved):

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: trust

msonnb (approved):

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
andreiborza: Went with naming the gaps plus the raw range, since Node 21 is supported (the SDK falls back to patching `node:http` below 22.12): "The minimum is 20.19.0, and Node.js 22 needs 22.12 or higher while Node.js 23 needs 23.2 or higher (`>=20.19.0 <22.0.0 || >=22.12.0 <23.0.0 || >=23.2.0`)". Updated in fca41d4044

PR #19259, opened 2026-09-03
chore(deps): Update @sentry/browser and @sentry/nextjs to 11.0.0-beta.0

andreiborza:
## What

Updates `@sentry/browser` and `@sentry/nextjs` to 11.0.0-beta.0 and adapts the app to the v11 build options: `withSentryConfig` now comes from `@sentry/nextjs/config`, `enableLogs` is gone, and `unstable_sentryWebpackPluginOptions` is replaced by a top-level `applicationKey`.

Reaching `@sentry/nextjs/config` needs `moduleResolution: "bundler"`, since `"node"` ignores package exports maps. That pulls in two unrelated follow-ons: the `hastscript` deep import moves to its canonical home in `hast`, and `ts-node` gets a commonjs override so `scripts/` keeps resolving extensionless imports.

## Why

Keep the docs site on the latest v11 prerelease so we catch breaking changes before it goes stable.


src/components/highlightCode.tsx
```diff
@@ -1,5 +1,5 @@
+import type {Nodes} from 'hast';
```
chargome: Why?

tsconfig.json
```diff
@@ -52,7 +52,7 @@
     "allowJs": true,
     "strict": false,
     "incremental": true,
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
chargome: Why?

chargome (approved):

tsconfig.json
andreiborza, line 55: > Reaching @sentry/nextjs/config needs moduleResolution: "bundler", since "node" ignores package exports maps. That pulls in two unrelated follow-ons: the hastscript deep import moves to its canonical home in hast, and ts-node gets a commonjs override so scripts/ keeps resolving extensionless imports.

PR #19267, opened 2026-09-03
ref(mdx): Read image dimensions with sharp instead of image-size

oioki:
## DESCRIBE YOUR PR

Drops the `image-size` dependency and reads doc image dimensions with `sharp` instead.

`image-size` has three unpatched infinite-loop DoS advisories against its ICNS, JXL and HEIF parsers ([GHSA-w3rx-r6r6-pgpr](https://github.com/advisories/GHSA-w3rx-r6r6-pgpr), [GHSA-5p2g-fcmc-qvqq](https://github.com/advisories/GHSA-5p2g-fcmc-qvqq)). There is nothing to upgrade to — the advisories cover every version through 2.0.2, which is the latest release, and upstream has been dormant since April 2025. The fix PR the advisory points at no longer exists on GitHub.

Our only call site is `src/remark-image-processing.js`, the build-time remark plugin that stamps `?v=<hash>#<W>x<H>` onto doc image URLs. It only ever sees files committed to this repo, so there is no request path an attacker can reach, and the practical exposure was a crafted image in a pull request hanging the docs build until CI times out. Rather than accept that indefinitely, this removes the package: `sharp` is already a direct dependency (required by `next/image`) and reads the same headers.

This closes Dependabot alerts 396, 397, 403 and 404 by removing the package rather than dismissing them as accepted risk.

### Changes

- Remove `image-size` from `package.json` (and `queue`, its only dependency, from the lockfile)
- Rewrite the remark transformer to use `sharp(buffer).metadata()`
- Add tests for the plugin, which previously had none

### Notes for reviewers

**Output is unchanged.** Every image in the repo — 1015 files across `public/`, `docs/`, `develop-docs/`, `includes/` and `platform-includes/` — was compared across both libraries, with **zero differences**. That includes all 10 SVGs and one with fractional dimensions (`1071.046875 × 507.42999629748056` → `1071x507` from both), so the SVG special case I expected to need turned out to be unnecessary. Re-verified against `sharp` 0.35.0 after it landed on master.

**The transformer is now async.** `sharp` has no sync metadata API, so it collects the image nodes during `visit` and resolves them with `Promise.all`. Async remark transformers already work in this pipeline (`src/remark-variables.js:30`), and I confirmed it end-to-end through real `bundleMDX` rather than relying on that precedent.

**Small perf cost.** Reading all 1015 images takes 91ms with `sharp` versus 4ms with `image-size` (~0.09ms per image). Immaterial against a full docs build, and it's now parallel per page where it used to be serial — but flagging it rather than burying it.

## IS YOUR CHANGE URGENT?

## SLA

## PRE-MERGE CHECKLIST

---

Fixes VULN-2750
Fixes VULN-2753

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,6 @@
     "hast-util-to-jsx-runtime": "^2.3.2",
     "hast-util-to-string": "^3.0.1",
     "hastscript": "^8.0.0",
-    "image-size": "^1.2.1",
     "js-cookie": "^3.0.7",
     "js-yaml": "^4.3.1",
     "match-sorter": "^6.3.4",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -147,9 +147,6 @@ importers:
       hastscript:
         specifier: ^8.0.0
         version: 8.0.0
-      image-size:
-        specifier: ^1.2.1
-        version: 1.2.1
       js-cookie:
         specifier: ^3.0.7
         version: 3.0.8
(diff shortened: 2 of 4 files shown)


chargome (approved):
LGTM!

PR #19302, opened 2026-09-08
fix(seo): canonicalize dotted documentation routes

sfanahata:
## DESCRIBE YOUR PR

Fixes non-indexable canonical targets for documentation routes whose final segment contains a dot, including Java and Python migration guides such as `7.x-to-8.0` and `1.x-to-2.x`.

How it was fixed:
- Normal documentation pages use a trailing slash.
- Pages with filename-like version numbers, such as 7.x-to-8.0, do not.
- The home page remains /.

We then applied that same rule everywhere Sentry tells search engines about official URLs:
- The canonical tag inside each page
- The sitemap submitted to search engines
- The canonical header attached to Markdown versions of pages
- Manually configured canonical paths

This PR adds one shared `canonicalPath` helper that mirrors Next.js route normalization and uses it for:

- HTML canonical metadata
- Sitemap URLs
- Canonical `Link` headers on `.md` responses
- Custom canonical paths

Ordinary documentation routes continue to use trailing slashes. Dotted migration and versioned routes use the direct no-slash form. 

### Verification

- `pnpm test:ci`: 377 tests passed
- `pnpm lint:eslint`: 0 errors (9 existing warnings)
- `pnpm lint:prettier`: passed
- `pnpm lint:ts`: passed
- Next production compilation completed locally; the full Vercel production build passed
- Local HTTP verification confirmed:
  - Java and Python dotted routes return `200` and are self-canonical without `/`
  - Their slash forms return `308` to the no-slash URL
  - An ordinary docs route returns `200` and is self-canonical with `/`
  - Sitemap and `.md` canonical headers use the same route-aware forms

## IS YOUR CHANGE URGENT?

- [ ] Urgent deadline (GA date, etc.)
- [ ] Other deadline

## PRE-MERGE CHECKLIST

- [ ] PR was reviewed and approved by a member of the Sentry docs team

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -1,6 +1,7 @@
 import * as Sentry from '@sentry/nextjs';
 import type {NextRequest} from 'next/server';
 import {NextResponse, userAgent} from 'next/server';
+import {canonicalPath} from 'sentry-docs/canonical';
 import {
   AI_AGENT_PATTERN,
   matchPattern,
@@ -21,7 +22,11 @@ const CANONICAL_HOST = new URL(BASE_URL).hostname;
 // Production domains whose content should be indexable by search engines.
 // All other hostnames (Vercel preview/deployment URLs, old production deployments)
 // get X-Robots-Tag: noindex to prevent search engines from indexing stale content.
-const INDEXABLE_HOSTNAMES = new Set(['docs.sentry.io', 'develop.sentry.dev', 'localhost']);
+const INDEXABLE_HOSTNAMES = new Set([
+  'docs.sentry.io',
+  'develop.sentry.dev',
+  'localhost',
+]);
 
 export const config = {
   // learn more: https://nextjs.org/docs/pages/building-your-application/routing/middleware#matcher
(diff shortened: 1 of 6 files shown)


middleware.ts
```diff
@@ -21,7 +22,11 @@ const CANONICAL_HOST = new URL(BASE_URL).hostname;
 // Production domains whose content should be indexable by search engines.
 // All other hostnames (Vercel preview/deployment URLs, old production deployments)
 // get X-Robots-Tag: noindex to prevent search engines from indexing stale content.
-const INDEXABLE_HOSTNAMES = new Set(['docs.sentry.io', 'develop.sentry.dev', 'localhost']);
+const INDEXABLE_HOSTNAMES = new Set([
+  'docs.sentry.io',
+  'develop.sentry.dev',
+  'localhost',
```
sergical: why is localhost indexable?

sergical (approved):
small comment otherwise LGTM

middleware.ts
sfanahata: This is pre-existing behavior from #18889; this PR only caused Prettier to wrap the existing line. `localhost` is exempt from the `X-Robots-Tag: noindex` header so local development and SEO testing behave like production. It does not make localhost publicly crawlable or indexable, because external search crawlers cannot access a developer's loopback address.

sergical: can we just remove it then?

sfanahata: Yeah, I think so. I don't think anyone even does "local SEO testing". Not sure that's a thing.

sfanahata: Agreed. We do not rely on production-like indexing behavior for local SEO testing, so I removed `localhost` from the allowlist. Local responses now receive `X-Robots-Tag: noindex`, with a middleware test covering the behavior.

sfanahata: It's been removed.

PR #19367, opened 2026-09-11
docs(javascript): Fix browser tracing verification

msonnb:
Make the browser tracing verification handler async so the example runs instead of failing to parse at `await`. Clarify which error and span to look for in Sentry.

Fixes SDK-1503

## IS YOUR CHANGE URGENT?

### DIFF
--- a/platform-includes/getting-started-verify/javascript.mdx
+++ b/platform-includes/getting-started-verify/javascript.mdx
@@ -42,9 +42,9 @@ To verify that Sentry captures errors and creates issues in your Sentry project,
 <SplitSection>
 
 <SplitSectionText>
-To test your tracing configuration, update the previous code to simulate a longer operation and start a trace.
+To test your tracing configuration, update the previous code to measure a simulated operation with a span.
 
-Open the page in a browser and click the button to throw an error and create a trace.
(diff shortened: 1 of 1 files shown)


platform-includes/getting-started-verify/javascript.mdx
```diff
@@ -42,26 +42,26 @@ To verify that Sentry captures errors and creates issues in your Sentry project,
 <SplitSection>
 
 <SplitSectionText>
-To test your tracing configuration, update the previous code to simulate a longer operation and start a trace.
+To test your tracing configuration, update the previous code to measure a simulated operation with a span.
 
-Open the page in a browser and click the button to throw an error and create a trace.
+Open the page in a browser and click the button. In Sentry, look for the `Sentry Test Error` issue and the `Example Frontend Span` span in the trace.
```
JPeer264: l: I know it was in the ticket, but I'm not sure if we should be that specific

JPeer264 (approved):

platform-includes/getting-started-verify/javascript.mdx
msonnb: updated the wording slightly to not include the specific names but still tell the user to look in Sentry

JPeer264: Cool thx

PR #19456, opened 2026-09-17
docs(javascript): Align tracing attributes and behavior with v11

msonnb:
Update JavaScript tracing docs and examples to use v11 span attributes, including HTTP, database, Gen AI, and the root-span source constant. Explain low-cardinality span names as the default behavior.

Clarify that `useOperationNameForRootSpan` controls `sentry.graphql.operation` on the enclosing root span in both modes, but only renames it in static mode. The GraphQL span's own operation attributes are recorded regardless of the option.

Fixes https://linear.app/getsentry/issue/SDK-1475

## IS YOUR CHANGE URGENT?


platform-includes/performance/automatic-instrumentation-custom-routing/javascript.mdx
```diff
@@ -19,7 +19,7 @@ const client = Sentry.init({
 let pageLoadSpan = Sentry.startBrowserTracingPageLoadSpan(client, {
   name: window.location.pathname,
   attributes: {
-    [Sentry.SEMANTIC_ATTRIBUTE_SENTRY_SOURCE]: "url",
+    [Sentry.SENTRY_SEGMENT_NAME_SOURCE]: "url",
```
Lms24: m: I just saw we actually export this attribute constant from the SDK packages but I'm not yet sure we want to do this. Can we switch to the string instead here?

Lms24 (approved):

platform-includes/performance/automatic-instrumentation-custom-routing/javascript.mdx
msonnb: fixed here and in the migration guide 👍

PR #19466, opened 2026-09-17
docs(nuxt): Update docs for v11

s1gr1d:
I added the changes in two separate commits so it's easier to review. The first commit (add v10 versioned docs) is just a copy-paste.

- Adds versioned docs for v10 (just a copy-paste)
- Updates the current docs to mention the bundling on the server-config (no `--import`)
- Renames "Installation Methods" in the sidebar to "Server-Side Initialization" so it works for both versions and we have a fallback

Linear: https://linear.app/getsentry/issue/SDK-1510/nuxt-getting-started-manual-setup


docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
```diff
@@ -1238,12 +1238,33 @@ The generated client snippet always includes the Replay integration with default
...
+The SDK bundles `sentry.server.config.ts` into the Nitro server build and initializes Sentry when the server starts. Instrumentation happens at build time, so you no longer preload the config file. Remove the `--import` flag from your start commands:
 
 ```bash
+# Before
 node --import ./.output/server/sentry.server.config.mjs .output/server/index.mjs
+NODE_OPTIONS='--import ./.nuxt/dev/sentry.server.config.mjs' nuxt dev
+
+# After
+node .output/server/index.mjs
+nuxt dev
```
chargome, line 1254: If we remove `--import` from `dev` we'll lose all auto-instrumentation right? We should either leave it or document this caveat that build time instrumentation will only work in prod

docs/platforms/javascript/common/troubleshooting/index.mdx
```diff
@@ -596,14 +600,27 @@ shamefully-hoist=true
...
+    Read more about this in <PlatformLink to="/install">Server-Side Initialization</PlatformLink>.
+
+    If the server side doesn't send events, check the following:
 
-    When running the build output, Sentry needs to be initialized before running the rest of the application. This is done automatically, but might not work for your use case.
-    Read more about this in <PlatformLink to="/install">installation methods</PlatformLink>.
+    - The `sentry.server.config.ts` file is in the root of your project, next to `nuxt.config.ts`.
+    - Enable `debug` in the `sentry` options of your `nuxt.config.ts`. The SDK then prints a message during the build when it bundles the server config.
+    - Remove the deprecated `autoInjectServerSentry` option, so that the SDK uses the bundled initialization.
+    - With a Cloudflare Nitro preset, the SDK doesn't use `sentry.server.config.ts`. Use the [Cloudflare Nitro plugin](/platforms/javascript/guides/cloudflare/frameworks/nuxt/) instead.
```
inventarSarah, line 611:
```suggestion
    - If you're using a Cloudflare Nitro preset, the SDK doesn't use `sentry.server.config.ts`. Use the [Cloudflare Nitro plugin](/platforms/javascript/guides/cloudflare/frameworks/nuxt/) instead.
```

docs/platforms/javascript/common/migration/v10-to-v11/index.mdx
s1gr1d: We don't emit this file anymore so we can't `--import` it but we loose **some** instrumentation as there's no server bundle to instrument at build time (and we cannot inject the tracing channels). But libraries with native tracing channels work.

I'll add a comment.

s1gr1d: Done in: [`a507bcd` (this PR)](https://github.com/getsentry/sentry-docs/pull/19466/changes/a507bcdfa63559645d834453a625163a179ffdd2)

docs/platforms/javascript/guides/nuxt/install/index.mdx
```diff
@@ -1,12 +1,111 @@
...
+
+### Prerendering
+
+The SDK doesn't initialize Sentry while Nitro prerenders your pages at build time. As a result, builds don't send events to Sentry, and prerendered HTML doesn't contain trace meta tags.
+
+### Cloudflare
+
+With a Cloudflare Nitro preset, the SDK doesn't use `sentry.server.config.ts`. Set up the SDK with the [Cloudflare Nitro plugin](/platforms/javascript/guides/cloudflare/frameworks/nuxt/) instead.
+
+## Upgrade From Version 10
```
inventarSarah, line 56:
```suggestion
## Upgrade From SDK Version 10
```

docs/platforms/javascript/guides/nuxt/install/index.mdx
```diff
@@ -1,12 +1,111 @@
...
+
+Because a preload is no longer necessary, the following `sentry` module options are deprecated. They continue to work in version 11 and will be removed in the next major version.
+
+| Option                                                  | Replacement                                 |
+| ------------------------------------------------------- | ------------------------------------------- |
+| `autoInjectServerSentry: "top-level-import"`            | Remove the option. The default replaces it. |
+| `autoInjectServerSentry: "experimental_dynamic-import"` | Remove the option. The default replaces it. |
+| `experimental_entrypointWrappedFunctions`               | Remove the option. It has no replacement.   |
+
+If you set `autoInjectServerSentry`, the SDK uses that mode instead of the bundled initialization and prints a deprecation warning during the build. Remove the option to use the bundled initialization.
```
inventarSarah, line 96:
```suggestion
If you set `autoInjectServerSentry`, the SDK uses that mode instead of the bundled initialization and prints a deprecation warning during the build. Remove the option to use the bundled initialization:
```

docs/platforms/javascript/guides/nuxt/install/index.mdx
```diff
@@ -1,12 +1,111 @@
...
 
-<PageGrid />
+The SDK bundles your `sentry.server.config.ts` file into the Nitro server build. A Nitro plugin that the SDK generates imports this file, so `Sentry.init()` runs when the server starts.
+
+The SDK adds its instrumentation at build time. Because of this, Sentry doesn't have to load before the rest of your application, and you don't need the Node `--import` flag.
+
+This has the following effects:
+
+- You start your application without additional flags, in production and in development mode.
+- You don't need a platform-specific setup on hosts where you can't set Node CLI flags, such as Vercel or Netlify.
```
inventarSarah: I wonder if this shouldn't go below into the "Upgrade" section since this talks about how things used to be -> this could be confusing for users who've never used the old version. Let me know if that makes sense

Instead, I'd add a description for the commands below - something like , "Use the following commands to start your app:"

docs/platforms/javascript/guides/nuxt/install/index.mdx
```diff
@@ -1,12 +1,111 @@
...
+
+```bash
+# Production: start your app after building your project with `nuxi build`
+node .output/server/index.mjs
+
+# Development
+nuxt dev
+```
```
inventarSarah, line 30: Is this relevant for users who don't/didn't need to migrate versions and start with 11? If no, I'd suggest moving this part to the "Upgrade" section

docs/platforms/javascript/guides/nuxt/install/index.mdx
inventarSarah: Or maybe "Upgrade to SDK Version 11" because in the alert at the top we say that this page only applies to V11 and it seems more logical to me to refer to v11 here

docs/platforms/javascript/guides/nuxt/install/index.mdx
```diff
@@ -1,12 +1,111 @@
 ---
-title: Installation Methods
+title: Server-Side Initialization
 sidebar_order: 1.5
-description: "Review our alternate installation methods."
+description: "Learn how the Nuxt SDK initializes Sentry on the server side, and which installation options are deprecated."
```
inventarSarah, line 4: I suggest phrasing that more positively since this is the page description. Something like this maybe?
->

Learn how the Nuxt SDK initializes Sentry on the server side, and how to upgrade to version 11.

platform-includes/getting-started-complete/javascript.nuxt.mdx
```diff
@@ -234,16 +234,16 @@ dotenv.config();
 <SplitSection>
 <SplitSectionText>
 
-**Sentry's server-side monitoring doesn't work in development mode**. To enable it, you first need to build your application and then load the Sentry server-side config using the `--import` flag when running your application.
+The SDK bundles `sentry.server.config.ts` into the Nitro server build and initializes Sentry when the server starts. This works in production and in development mode, and you don't need the Node `--import` flag.
```
inventarSarah, line 237: Is the mention of the flag needed for new users?

inventarSarah (commented):
Looks good!
I added a couple of suggestions and questions.
My main question: Do users - especially new Nuxt SDK users - need to be informed that they don't need to use the import flag? 
The info is def. needed in the migration guides but otherwise we usually document what users need to do and not what they don't need to do. Or maybe this info is helpful to a Nuxt dev -- please let me know :)

platform-includes/getting-started-complete/javascript.nuxt.mdx
s1gr1d: True, we can remove this and only keep it for the "upgrading" path.

docs/platforms/javascript/guides/nuxt/install/index.mdx
```diff
@@ -1,12 +1,98 @@
 ---
-title: Installation Methods
+title: Server-Side Initialization
```
chargome: I think this title does not make sense with the slug?

chargome (approved):

docs/platforms/javascript/guides/nuxt/index__v10.x.mdx
```diff
@@ -0,0 +1,140 @@
...
+
+<PlatformContent includePath="getting-started-prerequisites" />
+
+<StepConnector selector="h2" showNumbers={true}>
+
+## Install
+
+<PlatformContent includePath="sentry-init/experimental-warning" />
+
+<PlatformContent includePath="sentry-init/install-command" />
```
sfanahata, line 19: Looks like this include resolves to `npx sentry@latest init`. Does that flow know to install and configure SDK v10, or does it default to the latest major? Would that be a compatibility issue?

sfanahata (approved):
Overall looks good. I have one compatibility question, and looks like some conflicts need to be resolved.

inventarSarah (approved):

docs/platforms/javascript/guides/nuxt/index__v10.x.mdx
s1gr1d: That's a good observation! But our wizard currently does not provide the option to select the version. It will always install the latest major (soon v11). When setting up a new project, only the latest version matters. I think we can keep the snippet just like that because v10 users probably go to the v10 docs for information, not for a new install.

PR #19472, opened 2026-09-17
docs(ember): Document explicit addon setup and tracing

msonnb:
The Ember setup guide now initializes Sentry in `app/app.ts` and enables tracing through an instance initializer. It also calls `loadInitializers()` so Ember loads that initializer. The options page documents performance settings on `instrumentAppInstancePerformance()`, including component and runloop thresholds, instead of the old addon config.

Verified against the v10-to-v11 migration guide and Ember SDK implementation.

Fixes [SDK-1484](https://linear.app/getsentry/issue/SDK-1484/rewrite-ember-setup-for-the-v2-embroider-addon-in-docs).

## IS YOUR CHANGE URGENT?


docs/platforms/javascript/guides/ember/index.mdx
```diff
@@ -51,16 +51,16 @@ Choose the features you want to configure, and this guide will show you how:
...
 </SplitSectionText>
 <SplitSectionCode>
 
-```javascript {filename:app.js}
+```typescript {filename:app/app.ts}
 import Application from "@ember/application";
 import Resolver from "ember-resolver";
 import loadInitializers from "ember-load-initializers";
-import config from "./config/environment";
+import config from "my-app/config/environment";
```
s1gr1d: Is this the way this should be imported? Hm, looks like this does not work for everyone on copy/paste.

s1gr1d (approved):

docs/platforms/javascript/guides/ember/index.mdx
msonnb: agreed, reverted to the relative import.

PR #19445, opened 2026-09-16
docs(javascript): rework Mastra guide to SDK-based setup, extend Node guide

mydea:
## DESCRIBE YOUR PR

Reworks the standalone **Mastra** JavaScript guide to the Sentry SDK-based setup, mirroring the Eve guide (#19310). Based on the [`node-mastra`](https://github.com/getsentry/sentry-javascript/tree/develop/dev-packages/e2e-tests/test-applications/node-mastra) and [`cloudflare-mastra`](https://github.com/getsentry/sentry-javascript/tree/develop/dev-packages/e2e-tests/test-applications/cloudflare-mastra) e2e test apps.

- **Extends the Node guide** (`fallbackGuide: javascript.node`) instead of `inheritCommonContent: false`, so the Mastra guide inherits the Node common pages. Adds the `mastra` icon.
- **Replaces the `@mastra/sentry` exporter** with a plain `Sentry.init()`. Sentry's Mastra integration is enabled by default and auto-bootstraps observability through `@mastra/observability`. Adds a warning against running the community `@mastra/sentry` exporter alongside it.
- **Node setup (primary):** `instrument.mjs` in `src/mastra/public/`, preloaded via `mastra dev/start --custom-args="--import=./instrument.mjs"`.
- **Cloudflare Workers (install variant):** `@sentry/cloudflare` + `sentryCloudflareVitePlugin()` + `src/instrument.server.ts` + `nodejs_compat`.
- Requires `@sentry/node` / `@sentry/cloudflare` `>=11.0.0-rc.0` and `@mastra/core` `>=1.63.2`; updates the captured-span mapping to the built-in integration.
- Adds `javascript.mastra` to every `supported`/`notSupported` list in JS common content wherever `javascript.eve` appears (112 files), so inherited Node sections render and browser-only content is hidden from the Mastra guide. `PlatformSection` doesn't resolve `fallbackGuide`, so `javascript.mastra` must be enumerated alongside `javascript.node`/`javascript.eve` in both directions.

## IS YOUR CHANGE URGENT?

Help us prioritize incoming PRs by letting us know when the change needs to go live.
Select exactly one option. For deadlines, replace `YYYY-MM-DD` with the due date. You can update this information later by editing the PR description.

- [ ] Urgent deadline (GA date, etc.): YYYY-MM-DD
- [ ] Other deadline: YYYY-MM-DD
- [x] No deadline: Not urgent, can wait up to 1 week+

## SLA

[...]

PR #19528, opened 2026-09-21
docs(javascript): Give Flue its own guide with SDK-based setup

RulaKhaled:
## DESCRIBE YOUR PR

Gives **Flue** its own JavaScript guide on the Sentry SDK-based setup, mirroring the Mastra guide (#19445). Based on the [`node-flue`](https://github.com/getsentry/sentry-javascript/tree/develop/dev-packages/e2e-tests/test-applications/node-flue) and `cloudflare-flue` e2e test apps.

- **Extends the Node guide** (`fallbackGuide: javascript.node`), so the Flue guide inherits the Node common pages. Adds the `flue` icon. Deletes `common/agent-tracing/flue.mdx` and redirects it, Cloudflare page first.
- **Replaces the `flue add tooling sentry` blueprint**, which wires Sentry up through `@flue/opentelemetry`. The SDK instruments Flue directly now. Adds a warning to remove the blueprint's `createOpenTelemetryInstrumentation` call, which double-counts `gen_ai` spans.
- **Node setup (primary):** `Sentry.init()` plus `instrument(Sentry.createFlueInstrumentation())`, imported as the first line of `src/app.ts`. Flue is registered rather than patched, so that call is the wiring.
- **Cloudflare Workers:** `sentryCloudflareVitePlugin()` registers the instrumentation at build time, so there's no `instrument()` call. Each agent runs in its own Durable Object, so the guide leads with the `extend({ wrap })` wrapper that gets `Sentry.init()` into that isolate — without it the agent runs fine, turns settle, nothing errors, and nothing is captured.
- **Documents that `instrument.server.ts` does nothing for Flue.** That convention needs a worker entry in Wrangler's `main`; Flue supplies its own virtual entry instead. Verified against `cloudflare-flue`.
- Requires `@sentry/node` / `@sentry/cloudflare` `>=11.0.0-rc.0` and `@flue/runtime` `>=2.0.0 <3.0.0`. Corrects the old Correlation Tags section: the SDK emits `gen_ai.conversation.id`, `gen_ai.agent.name`, and `flue.turn.purpose`, not the blueprint's `flue.instance.id` / `flue.agent.name` / `flue.submission.id`.
- Adds `javascript.flue` to every `supported`/`notSupported` list in JS common content wherever `javascript.eve` appears (106 files), so inherited Node sections render and browser-only content is hidden from the Flue guide. `PlatformSection` doesn't resolve `fallbackGuide`, so `javascript.flue` must be enumerated alongside `javascript.node`/`javascript.eve` in both directions.

Follow-ups: Flue needs a `STANDALONE_AGENT_TRACING_GUIDES` entry (`sidebar_order: 28`) once #19445 lands; the version pins need revisiting once getsentry/sentry-javascript#24476 ships; and the double-counting warning is reasoned from both instrumentations registering through `instrument()` under different keys, not run against the blueprint.

## IS YOUR CHANGE URGENT?

## SLA

## PRE-MERGE CHECKLIST


docs/platforms/javascript/guides/flue/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+<OnboardingOptionButtons
+  options={["error-monitoring", "performance"]}
+  defaults={["performance"]}
+/>
+
+Install the Sentry Node SDK:
+
+```bash {tabTitle:npm}
+npm install @sentry/node@^11.0.0-rc.0
```
JPeer264:
```suggestion
npm install @sentry/node@^11.0.0-rc.1
```

Maybe even stable v11 in case we release tomorrow

docs/platforms/javascript/guides/flue/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+- **Token and cost values are doubled.** Remove the `createOpenTelemetryInstrumentation` call left over from the `flue add tooling sentry` blueprint. Flue emits its own model spans through it, so both it and the SDK count the same turn.
+- **No AI spans at all.** Confirm `instrument(Sentry.createFlueInstrumentation())` actually runs. Importing `sentry-init.ts` at the top of `src/app.ts` runs it during app startup, before any agent takes a turn.
+
+<Include name="flue/next-steps" />
+
+<Include name="flue/what-gets-captured" />
+
+## Supported Versions
+
+- `@sentry/node`: `>=11.0.0-rc.0`
```
JPeer264: l: Maybe it's better to keep just one RC version in the docs, even if `rc.0` is already supporting it. Or just move to v11: https://github.com/getsentry/sentry-docs/pull/19528/changes#r4071017479

```suggestion
- `@sentry/node`: `>=11.0.0-rc.1`
```

docs/platforms/javascript/guides/flue/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+Deploying Flue agents to Cloudflare Workers? Follow the [Cloudflare Quick Start](./cloudflare/) instead.
+
+## Prerequisites
+
+Before you begin, you need:
+
+- A Sentry [account](https://sentry.io/signup/) and [project](/product/projects/). The project's DSN tells the SDK where to send data.
+- A Flue application using `@flue/runtime` version `2.0.0` or later. Flue requires Node.js `22.19` or later.
+- `@sentry/node` version `11.0.0-rc.0` or later, or [`@sentry/cloudflare@^11.0.0-rc.1` when running on Cloudflare](./cloudflare). Browser runtimes are not supported.
```
JPeer264: l: Maybe we should wait 1 day and just move it to v11 right away? 

```suggestion
- `@sentry/node` or `@sentry/cloudflare` version `11.0.0` or later. Browser runtimes are not supported.
```

docs/platforms/javascript/guides/flue/index.mdx
```diff
@@ -0,0 +1,122 @@
+---
```
JPeer264, line 1: q: Do we already have `flue` prepared to be part of `platformicons`? Would be nice to have a dedicated logo there

Image

docs/platforms/javascript/guides/flue/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+```bash {tabTitle:pnpm}
+pnpm add @sentry/node@^11.0.0-rc.0
+```
+
+## Configure
+
+Create a file that calls `Sentry.init()` and registers the Flue instrumentation. Flue is instrumented by registering with it rather than by patching, so this `instrument()` call is what connects the two:
+
+```typescript {filename:sentry-init.ts}
```
JPeer264: l/q: Shouldn't this rather be `instrumentation.ts` or `instrument.server.ts` or something like that? Never seen `sentry-init.ts` before

docs/platforms/javascript/guides/flue/cloudflare.mdx
```diff
@@ -0,0 +1,150 @@
...
+
+Each Flue agent runs in its own Durable Object, which is a separate isolate from your Worker entry. `Sentry.init()` has to run inside it, so create a module that wraps the generated agent Durable Object:
+
+```typescript {filename:src/sentry.ts}
+import { extend } from "@flue/runtime/cloudflare";
+import * as Sentry from "@sentry/cloudflare";
+
+export const cloudflare = extend({
+  wrap: (Final) =>
+    Sentry.instrumentDurableObjectWithSentry(
```
JPeer264, line 75: q: Do we need to wrap this? The Vite plugin should automatically instrument it actually. We would only need our Vite plugin and a `instrument.server.ts` - if it doesn't work we have to fix it, but could leave it as a first step.

JPeer264 (commented):
Added couple of points.

docs/platforms/javascript/guides/flue/index.mdx
RulaKhaled: oh true renamed to instrument.ts

docs/platforms/javascript/guides/flue/index.mdx
RulaKhaled: yah will open it separately

docs/platforms/javascript/guides/flue/cloudflare.mdx
RulaKhaled: yah flue instrumentation automatically registers, but instrument.server.ts can't work here, flue's main never lands in the file we read. leaving this as is for now, i'll look into the best approach this week getsentry/sentry-javascript#24601

mydea:
note: this will need rebasing once the v11 docs are merged (that introduces a bunch of changes to the agent tracing pages as well)

docs/platforms/javascript/guides/flue/index.mdx
mydea: I think there should be one already, see https://docs.sentry.io/platforms/javascript/guides/node/agent-tracing/#instrumentation

includes/flue/sdk-setup-alert.mdx
```diff
@@ -0,0 +1,5 @@
+<Alert>
+
+This guide covers the Sentry SDK-based setup, which replaces the `flue add tooling sentry` blueprint. If you previously ran the blueprint, remove its `createOpenTelemetryInstrumentation` call — it emits its own `gen_ai` spans and will double-count against the ones the SDK creates.
```
mydea, line 3:
```suggestion
This guide covers the Sentry SDK-based setup, which replaces the `flue add tooling sentry` blueprint. If you previously ran the blueprint, remove its `createOpenTelemetryInstrumentation` call — otherwise, you will get duplicate spans.
```

docs/platforms/javascript/guides/flue/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+- `@sentry/node` or `@sentry/cloudflare` version `11.0.0` or later. Browser runtimes are not supported.
+
+<StepConnector selector="h2" showNumbers={true}>
+
+## Install
+
+Choose the features you want to configure, and this guide will show you how:
+
+<OnboardingOptionButtons
+  options={["error-monitoring", "performance"]}
```
mydea, line 38: just wondering, no strong feelings, but is there a specific reason we do not show profiling here? That _should_ work with node I suppose?

docs/platforms/javascript/guides/flue/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+  // ___PRODUCT_OPTION_START___ performance
+
+  // Set tracesSampleRate to 1.0 to capture 100%
+  // of spans for tracing.
+  // We recommend adjusting this value in production.
+  tracesSampleRate: 1.0,
+  // ___PRODUCT_OPTION_END___ performance
+});
+
+instrument(Sentry.createFlueInstrumentation());
```
mydea, line 75: is this still necessary? I though we shipped something that this can be skipped?

docs/platforms/javascript/guides/flue/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+```
+
+Import it at the top of your app entry, so it runs before your agents take a turn:
+
+```typescript {filename:src/app.ts}
```
mydea, line 80: no strong feeling, but is there a reason  we do not put the instrument.ts file inside of src? Seems more "common" to me that way?

docs/platforms/javascript/guides/flue/index.mdx
```diff
@@ -0,0 +1,122 @@
...
+import "../instrument.ts";
+```
+
+Call `instrument()` once, at module scope.
```
mydea, line 84: missing here (looking at our node-flue e2e test app at least) is a section on running this with `--import`, which I guess is required to actually get auto instrumentation for other things? We state that auto instrumentation works here but I think not quite? IMHO we can add a sub-step to configure that shows how to run this with `NODE_OPTIONS='--import=@sentry/node/import'` or similar in order to make this work?

docs/platforms/javascript/guides/flue/index.mdx
RulaKhaled: no reason, just copied mastra's structure

docs/platforms/javascript/guides/flue/index.mdx
RulaKhaled: not really, no real constraint either way, will move it

docs/platforms/javascript/guides/flue/index.mdx
RulaKhaled: still needed on node, what we shipped is the vite plugin auto registration for cloudflare only

PR #19530, opened 2026-09-21
docs(eve): Move Sentry.init to the agent/instrumentation/ directory

sergical:
## DESCRIBE YOUR PR

Eve `0.62.0` removed `agent/instrumentation.ts`. When the file exists, `eve build` fails with this error:

```
Found removed instrumentation file ".../agent/instrumentation.ts". Move it into the "instrumentation/" directory as one file per provider.
```

The Eve guide still told users to create that file. A user who follows the guide on a current Eve version cannot build their agent.

Changes to `docs/platforms/javascript/guides/eve/index.mdx`:

- **Configure:** `Sentry.init` moves to `agent/instrumentation/sentry.ts`. Eve requires each file in that directory to default-export a provider, so the file exports `defineInstrumentation({ flush })`, which also flushes Sentry when Eve flushes telemetry.
- **Upgrade alert:** explains the removal, the build failure, and that every file in `agent/instrumentation/` is a provider, so helper code belongs in a different directory.
- **Prompt and output recording:** the guide said that prompts and outputs are recorded by default. That is true only under `eve dev`. Outside development, Eve passes `recordInputs: false` and `recordOutputs: false` on each model call, and those values take priority over `dataCollection.genAI`. A new subsection shows the fix: `Sentry.vercelAIIntegration({ recordInputs: true, recordOutputs: true })`.
- **OpenTelemetry ownership:** warns against Eve's `otel()` and `otelIntegration()` providers next to the SDK setup. Eve refuses to start when the global tracer provider is already registered.
- **Verify and Supported Versions:** updated for the new path. The tested Eve version is now `0.63.0`.

How I tested: `eve` `0.63.0` with `@sentry/node` `11.0.0-rc.0`, two-turn sessions under `eve dev` and `eve start`. I checked the `gen_ai.*` spans, the conversation id, and the recorded inputs and outputs in Sentry. The example app is in [getsentry/sentry-agent-tracing-examples](https://github.com/getsentry/sentry-agent-tracing-examples).

## IS YOUR CHANGE URGENT?

- [ ] Other deadline:

## PRE-MERGE CHECKLIST

Created with Claude Code


sergical:
Closing. #19529 landed first and documents the new `Sentry.eveInstrumentation()` provider from getsentry/sentry-javascript#24539, which covers the same eve 0.62+ changes (the `agent/instrumentation/` layout, and the per-call `recordInputs: false` that dropped prompt content outside `eve dev`). That is the better fix.

PR #19549, opened 2026-09-22
docs(javascript): Note that tags and extra don't apply to streamed spans

chargome:
## DESCRIBE YOUR PR

With span streaming enabled, spans only carry attributes, so scope tags and extra (`setTag`, `setTags`, `setExtra`, `setExtras`) no longer reach any span. They still apply to errors, and in transaction mode (`traceLifecycle: "static"`) they keep landing on the transaction. This adds a note about that to the three JavaScript enriching-events pages where readers set that data, and points them to `setAttribute`/`setAttributes` for anything that should be searchable on spans.

- `enriching-events/tags/` — extend the existing "not applied to logs or metrics" warning to cover spans
- `enriching-events/scopes/` — callout after the "How Scope Data is Applied to Events" example
- `enriching-events/attributes/` — callout in the intro, noting attributes keep their type

Refs SDK-1474


docs/platforms/javascript/common/enriching-events/attributes/index.mdx
```diff
@@ -11,6 +11,12 @@ Unlike [tags](../tags/), which only accept `string` values, attributes support `
 Common uses include subscription tier, feature flags, or any business context that helps you filter and query your telemetry.
 Check out [Sending Span Metrics](../../tracing/span-metrics/) for an example of how to use attributes to enrich your spans.
 
+<Alert>
+
+If you're using <PlatformLink to="/tracing/streamed-spans/" fallbackPlatform="javascript">span streaming</PlatformLink>, use attributes for any data you want to search on spans. Streamed spans only carry attributes, so tags and extra (`Sentry.setTag()`, `Sentry.setExtra()`, and similar) aren't added to any span. They still apply to errors, so you don't need to remove them. Unlike tags, attributes keep their type, so you don't need to convert numbers and booleans to strings.
```
Lms24: The "if you're using..." sentence might be understood by users that they don't use it because they set up the SDK without any further adjustments. I think we should reword this a bit

```suggestion
Use attributes for any data you want to search on spans. <PlatformLink to="/tracing/streamed-spans/" fallbackPlatform="javascript">Streamed</PlatformLink> spans only carry attributes, so tags and extra (`Sentry.setTag()`, `Sentry.setExtra()`, and similar) aren't added to any span. They still apply to errors, so you don't need to remove them. Unlike tags, attributes keep their type, so you don't need to convert numbers and booleans to strings.
```

Lms24 (approved):

